PR #5830, opened 2025-04-28
[PD] Fix Assertion failed: /DeepEP/csrc/kernels/internode.cu:483, condition: ibgda_get_state()->num_rc_per_pe >= num_channels #134

whybeyoung:
## Motivation

Fix https://github.com/deepseek-ai/DeepEP/issues/134

PR #6832, opened 2025-06-03
[FIX]Assertion failed: /DeepEP/csrc/kernels/internode.cu:483, condition: ibgda_get_state()->num_rc_per_pe >= num_channels

pengchengneo:
related issue
https://github.com/deepseek-ai/DeepEP/issues/134

reproduce command

`python3 -m sglang.launch_server --model-path=/models/DeepSeek-R1-BF16 --tp-size 32 --attention-backend flashinfer --trust-remote-code --nnodes 4 --node-rank 0 --host 0.0.0.0 --port 30000 --dist-init-addr epic0528-decode-0.epic0528-decode.inference-system.svc.cluster.local:20000 --mem-fraction-static 0.80 --enable-metrics --speculative-algorithm EAGLE --speculative-num-steps 3 --speculative-eagle-topk 1 --speculative-num-draft-tokens 4 --chunked-prefill-size -1 --served-model-name deepseek-r1 --disable-cuda-graph --log-level debug --enable-deepep-moe`

 num_qps_per_rank in Buffer must be greater or equal than comm_sms // 2 to avoid 
`Assertion failed: /DeepEP/csrc/kernels/internode.cu:483, condition: ibgda_get_state()->num_rc_per_pe >= num_channels`


pengchengneo:
@zhyncs hi, Could you help me review this bug fix, thanks !

pengchengneo:
@Ying1123 hello, Could you take a look with this PR ? Thanks

whybeyoung:
strange..     i have ever fixed it  .  emmm> https://github.com/sgl-project/sglang/pull/5830

pengchengneo:
> strange.. i have ever fixed it . emmm> #5830

I think the code of DeepEP auto mode import this bug again @whybeyoung 

 @fzyzcjy  @zhyncs @Ying1123  Could anyone can review this bugfix ?

fzyzcjy:
I think I have made that if-else o/w it errored in another case... I may check later what I met

fzyzcjy:
Found it: https://github.com/sgl-project/sglang/pull/6673 (iirc it is because it failed another assertion; anyway the most naive way may be add more if-else)

pengchengneo:
a bit confused, so how can I modify this pr ?  @fzyzcjy

pengchengneo:
@fzyzcjy hi, Could you share more about how can I modify this PR ?

fzyzcjy:
Have no time now to dig deeper :( but IIRC why I did that is b/c qwen3 moe + H20 fails some assertions in deepep.

ch-wan:
@whybeyoung According to DeepEP's doc (https://github.com/xai-org/DeepEP-private/blob/6f40d7494c2ddea0f8c8c887c8704a7c166943c6/deep_ep/buffer.py#L45), `num_qps_per_rank` should be equal to the number of local experts. Maybe we should reduce `num_sms` to resolve this error?

zhyncs:
Thanks for your contribution!

PR #6852, opened 2025-06-04
feat: add dp-rank to KV events

ishandhanani:
## Motivation

In DP attention - multiple DP ranks will maintain their own KV cache. In order to have efficient routing - we want to ensure that we're able to route to the DP rank that contains the best KV cache match. This PR now annotates KV events with the DP attention rank so it can be used by external consumers (ie Dynamo)

## Modifications

- KV events are now annotated with dp rank
- New test case that uses 2.3B version of DSR1 for testing

## Test Case

```bash
[2025-06-04 00:47:54] INFO:     127.0.0.1:34760 - "POST /generate HTTP/1.1" 200 OK
DP Rank 0 - EventBatch: ts=1748998067.7508266, attn_dp_rank=0
  DP0 - BlockStored(block_hashes=[2355870434279254203], parent_block_hash=5740354900026072187, token_ids=[0, 671, 6102, 4593, 294, 8760, 344], block_size=7, lora_id=None)
  DP0 - BlockStored(block_hashes=[2736904391167510675], parent_block_hash=2355870434279254203, token_ids=[22979, 27851, 128031, 59823, 33739, 113928, 105988], block_size=7, lora_id=None)
DP Rank 1 - EventBatch: ts=1748998067.750748, attn_dp_rank=1
  DP1 - BlockStored(block_hashes=[2355870434279254203], parent_block_hash=5740354900026072187, token_ids=[0, 671, 6102, 4593, 294, 8760, 344], block_size=7, lora_id=None)
  DP1 - BlockStored(block_hashes=[2736904391167510675], parent_block_hash=2355870434279254203, token_ids=[22979, 27851, 128031, 59823, 33739, 113928, 105988], block_size=7, lora_id=None)
DP Rank 0 - EventBatch: ts=1748998068.962058, attn_dp_rank=0
  DP0 - BlockRemoved(block_hashes=[2355870434279254203])
  DP0 - BlockStored(block_hashes=[-8753497827991233192], parent_block_hash=5740354900026072187, token_ids=[0], block_size=1, lora_id=None)
  DP0 - BlockStored(block_hashes=[-3043180041460246154], parent_block_hash=-8753497827991233192, token_ids=[671, 6102, 4593, 294, 8760, 344], block_size=6, lora_id=None)
DP Rank 1 - EventBatch: ts=1748998068.3989677, attn_dp_rank=1
  DP1 - BlockRemoved(block_hashes=[2355870434279254203])
[...]

PR #6884, opened 2025-06-05
feat: add direct routing strategy to DP worker

ishandhanani:
## Motivation

Because each DP attention worker contains its own KV cache we can have smarter optimization strategies when routing to a specific rank. Now that we have [KV events annotated with the proper DP rank](https://github.com/sgl-project/sglang/pull/6852), we can enable a KV based routing strategy. 

## Modifications

- Direct DP Routing: Adds the capability to specify a target data parallel rank for a request, allowing it to be routed directly to that specific DP worker.
- Parameter Propagation: Introduces a data_parallel_rank parameter/field and propagates it through various request and input structures (GenerateReqInput, TokenizedGenerateReqInput, Req) and function calls across the system.
- Routing Logic Update: Modifies the DataParallelController's round_robin_scheduler to check for the presence of data_parallel_rank and use it for direct routing if provided, falling back to the existing scheduling logic otherwise.
- Input Validation: Adds validation in the engine entrypoints (generate, async_generate) to ensure the provided data_parallel_rank is within the valid range if DP attention is enabled.

## Checklist


python/sglang/srt/managers/data_parallel_controller.py
```diff
@@ -248,12 +248,20 @@ def launch_tensor_parallel_group(
 
     def round_robin_scheduler(self, req: Req):
         if self.server_args.disaggregation_mode == "null":
-            self.workers[self.round_robin_counter].send_pyobj(req)
-            self.round_robin_counter = (self.round_robin_counter + 1) % len(
-                self.workers
-            )
+            if req.data_parallel_rank is not None:
+                logger.info(f"Direct routing to DP rank {req.data_parallel_rank}")
```
trevor-m: Should be a debug print

trevor-m (commented):
Overall looks good!

python/sglang/srt/managers/data_parallel_controller.py
ishandhanani: fixed

trevor-m (approved):

python/sglang/srt/entrypoints/EngineBase.py
```diff
@@ -23,6 +23,12 @@ def generate(
         token_ids_logprob: Optional[Union[List[List[int]], List[int]]] = None,
         lora_path: Optional[Union[List[Optional[str]], Optional[str]]] = None,
         custom_logit_processor: Optional[Union[List[str], str]] = None,
+        return_hidden_states: Optional[bool] = None,
+        stream: Optional[bool] = None,
+        bootstrap_host: Optional[Union[List[str], str]] = None,
+        bootstrap_port: Optional[Union[List[int], int]] = None,
+        bootstrap_room: Optional[Union[List[int], int]] = None,
+        data_parallel_rank: Optional[int] = None,
```
ByronHsu, line 31: should we unifiy the name `target_dp_rank` and `data_parallel_rank`?

ShangmingCai (commented):
Should prefill and decode instances have different params for target DP ranks? If we want to achieve load-aware DP routing, I think this should be considered. Also, what if the DP size of the decode instance is way larger than the prefill instance? If the DP rank of decode is dynamically adjusted, we need a mechanism to notify the decode instance which prefill rank its requests should be sent to.

python/sglang/srt/entrypoints/EngineBase.py
ishandhanani: Yea good idea - let me unify this

ishandhanani: Done

ishandhanani:
> Should prefill and decode instances have different params for target DP ranks? If we want to achieve load-aware DP routing, I think this should be considered. Also, what if the DP size of the decode instance is way larger than the prefill instance? If the DP rank of decode is dynamically adjusted, we need a mechanism to notify the decode instance which prefill rank its requests should be sent to.

This is a good idea but I think it might require some deeper changes to ensure that the communication between p/d for transfer does not break down. I want to make sure that I'm not just appending new fields into the various request APIs. As we run benchmarks - I can work on a design for this and then make a follow up PR @ShangmingCai

ByronHsu (approved):

PR #3822, opened 2025-02-24
[feat] add small vocab table for eagle's draft model[1].

Zhou-sx:
## Motivation
In speculative decoding, while the draft model's decode layers are drastically reduced, the computational bottleneck shifts to its lm_head (proportional to vocabulary size). In our case, the profile file shows that the small model decode takes approximately 1.8ms at a time, with lm-head occupying 860us, which is about half of the small model. By leveraging the target model's validation to ensure correctness, the draft model can safely use a truncated vocabulary—retaining only high-frequency tokens. This optimization reduces lm_head overhead, further accelerating the Eagle speculative decoding pipeline without compromising quality.
Todo: further enhance efficiency, I  propose dynamically updating the reduced vocabulary (token_map) during inference. This ensures the draft model adapts to shifting token distributions in real-time workloads.

## Modifications
Add  speculative-token-map in ServerArgs.
Change the the lm-head part in the LlamaEagle Model.

## Checklist

- [x] Format your code according to the [Code Formatting with Pre-Commit](https://docs.sglang.ai/references/contribution_guide.html#code-formatting-with-pre-commit).
- [ ] Add unit tests as outlined in the [Running Unit Tests](https://docs.sglang.ai/references/contribution_guide.html#running-unit-tests-adding-to-ci).
- [ ] Update documentation / docstrings / example tutorials as needed, according to [Writing Documentation](https://docs.sglang.ai/references/contribution_guide.html#writing-documentation-running-docs-ci).
- [ ] Provide throughput / latency benchmark results and accuracy evaluation results as needed, according to [Benchmark and Profiling](https://docs.sglang.ai/references/benchmark_and_profiling.html) and [Accuracy Results](https://docs.sglang.ai/references/accuracy_evaluation.html).
- [ ] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co-author when merging the PR.
- [ ] Please feel free to join our Slack channel at https://slack.sglang.ai to discuss your PR.

## How to use
Set `--speculative-token-map` to use this optimization. 
+ You can get the high-frequency token in FR-Spec from https://huggingface.co/thunlp/LLaMA3-Instruct-8B-FR-Spec. 

Or you can obtain high-frequency token by yourself.
1. Execute inference on your dataset using sglang's standard inference mode and persist the outputs.
[...]

PR #6903, opened 2025-06-05
[Bugfix] add small vocab table for eagle qwen2

Swipe4057:
I am following the original PR [https://github.com/sgl-project/sglang/pull/3822/files ](https://github.com/sgl-project/sglang/pull/3822/files), which introduces the small vocab table and modifies the script for llama models in python/sglang/srt/models/llama_eagle.py, similarly adjusting the code for the qwen2 series of models. This should also resolve the issue I recently opened: https://github.com/sgl-project/sglang/issues/6863.


hebiao064:
@Swipe4057 would you please some testing result after this change? Maybe something like before and after?

The code has only one line and it make sense to me, after the test results being attached we can merge it, thanks!

PR #6907, opened 2025-06-05
Minor Optimizations in Radix Cache and Schedule Batch

sushah13:
## Motivation

This PR introduces performance optimizations to two key components of the SGLang serving stack: `radix_cache.py` and `schedule_batch.py`. The primary goal is to reduce latency and improve throughput by eliminating inefficient data structures and unnecessary computations during request processing and KV cache eviction.

### Optimizations in radix_cache.py
**1. Reduced latency via optimized eviction logic**
Before: Average latency over 12 requests: 0.577s
After: Average latency over 12 requests: 0.495s

**Evict Method Improvements (~60% speedup, from 31ms → 12ms):**

- Avoid rebuilding the heap with `heapq.heapify()` on every eviction
    - Previously, the eviction heap was rebuilt from scratch each time (O(n) cost).
    - ✅ Now: Maintain a persistent min-heap (self.evict_heap) and incrementally update it during insertions, lock/unlock updates, and deletions.

- Avoid full-tree traversal in `_collect_leaves()`
    - Tree traversal for collecting evictable leaves was costly and unnecessary on each eviction.
    - ✅ Now: The persistent heap contains all evictable leaves, eliminating the need for traversal.

- Avoid repeated heap rebalancing
    - Each heapq.heappop() call triggered rebalancing.
    - ✅ Now: evict() loop pops only unlocked and valid nodes using a guard, minimizing operations.

**Implementation Details:**
- Added self.evict_heap: List[Tuple[int, TreeNode]] to RadixCache.
- Heap is updated incrementally:
    - insert(): Add leaf if evictable.
    - inc_lock_ref(): Remove node from heap if it becomes locked.
    - dec_lock_ref(): Re-add node if it becomes unlockable.
    - _delete_leaf(): No-op for heap; eviction logic handles parent-child promotion.

**2. Faster leaf deletion via key tracking**
Previously, _delete_leaf() required a linear search in the parent’s children dict to find the key corresponding to the child node.
✅ Now: Each TreeNode stores its child_key upon insertion, eliminating the need to search during deletion.

### Optimizations in schedule_batch.py
Before: Average latency over 12 requests: 0.495s
After: Average latency over 12 requests: 0.475s

**Improvements:**
1. Flattening input IDs efficiently
❌ Before: `sum(input_ids, [])` (O(n²) operation due to repeated list copying).
✅ Now: `list(chain.from_iterable(input_ids))` for linear-time flattening.

2. Avoid redundant tensor construction
❌ Before: `torch.tensor(seq_lens, ...)` may create unnecessary copies.
✅ Now: `torch.as_tensor(seq_lens, dtype=torch.int64, device=self.device)` avoids extra copy if input is already a tensor-like object.

These optimizations contribute to overall reduced latency and improved batch throughput, especially under high request volume and large prompt size settings.

## Checklist

- [X] Format your code according to the [Code Formatting with Pre-Commit](https://docs.sglang.ai/references/contribution_guide.html#code-formatting-with-pre-commit).
- [X] Add unit tests as outlined in the [Running Unit Tests](https://docs.sglang.ai/references/contribution_guide.html#running-unit-tests-adding-to-ci).

##Profiling Screenshots:
Radix Cache:
BEFORE:
BEFORE - Radix Cache Profiler Screenshot
AFTER:
AFTER - Radix Cache Profiler Screenshot

Schedule Batch Logic:
BEFORE:
BEFORE - Schedule Batch Profiler Screenshot
AFTER:
AFTER - Schedule Batch Profiler Screenshot

### DIFF
--- a/python/sglang/srt/mem_cache/hiradix_cache.py
+++ b/python/sglang/srt/mem_cache/hiradix_cache.py
@@ -82,6 +82,9 @@ class HiRadixCache(RadixCache):
         return height
 
     def write_backup(self, node: TreeNode, write_back=False):
+        # Nothing to copy if the node is already evicted.
+        if node.value is None:
+            return 0
         host_indices = self.cache_controller.write(
             device_indices=node.value,
             node_id=node.id,
@@ -104,7 +107,15 @@ class HiRadixCache(RadixCache):
         return len(host_indices)
 
     def inc_hit_count(self, node: TreeNode):
-        if node.backuped or self.cache_controller.write_policy == "write_back":
+        # Skip if:
+        # - the node is already on host (backuped)
+        # - write-back policy is in use
+        # - the node has no live device data (evicted)
+        if (
+            node.backuped
+            or node.evicted
+            or self.cache_controller.write_policy == "write_back"
+        ):
             return
         node.hit_count += 1
         if node.hit_count >= self.write_through_threshold:
@@ -198,9 +209,15 @@ class HiRadixCache(RadixCache):
 
     def _evict_regular(self, node: TreeNode):
         # evict a node not initiated write to host
-        self.cache_controller.mem_pool_device_allocator.free(node.value)
         num_evicted = len(node.value)
-        self._delete_leaf(node)
+        self.cache_controller.mem_pool_device_allocator.free(node.value)
+
+        self.evictable_size_ -= num_evicted  # keep GPU counters in sync
+        node.value = None  # mark as evicted
+
+        # Remove the node only if it has no (even evicted) children
+        if len(node.children) == 0:
+            self._delete_leaf(node)
         return num_evicted
 
     def evict_host(self, num_tokens: int):
@@ -288,7 +305,15 @@ class HiRadixCache(RadixCache):
             len(prefix_indices) == 0 or prefix_indices.is_cuda
         ), "indices of device kV caches should be on GPU"
         if last_node.evicted:
+            # No host copy: fall back to recomputation
+            if not last_node.backuped:
+                # Climb to the nearest ancestor that *does* have live data.
+                while last_node.evicted and last_node != self.root_node:
+                    last_node = last_node.parent
+                return last_node, prefix_indices
+            # Host copy exists: try to pull it back
             loading_values = self.load_back(last_node, mem_quota)
+
             if loading_values is not None:
                 prefix_indices = (
                     loading_values
--- a/python/sglang/srt/mem_cache/radix_cache.py
+++ b/python/sglang/srt/mem_cache/radix_cache.py
@@ -277,6 +277,9 @@ class RadixCache(BasePrefixCache):
             node_value_len = len(node.value)
             self.token_to_kv_pool_allocator.free(node.value)
             node.value = None  # Mark node as evicted
+            # Update evictable size here since value is freed;
+            # do not update again in _delete_leaf()
+            self.evictable_size_ -= node_value_len
             num_evicted += node_value_len
             self._delete_leaf(node)
             self._record_remove_event(node)
@@ -447,7 +450,6 @@ class RadixCache(BasePrefixCache):
     def _delete_leaf(self, node):
         if node.child_key in node.parent.children:
             del node.parent.children[node.child_key]
-        self.evictable_size_ -= len(node.key)
 
     def _total_size_helper(self):
         total_size = 0
--- a/python/sglang/srt/mem_cache/radix_cache.py
+++ b/python/sglang/srt/mem_cache/radix_cache.py
@@ -395,6 +395,7 @@ class RadixCache(BasePrefixCache):
     def _insert_helper(self, node: TreeNode, key: List, value):
         now = time.monotonic()
         node.last_access_time = now
+
         if len(key) == 0:
             return 0
 
--- a/python/sglang/srt/mem_cache/radix_cache.py
+++ b/python/sglang/srt/mem_cache/radix_cache.py
@@ -52,6 +52,7 @@ class TreeNode:
         self.value = None
         self.lock_ref = 0
         self.last_access_time = time.monotonic()
+        self.child_key = None
 
         self.hit_count = 0
         # indicating the node is loading KV cache from host
@@ -110,6 +111,7 @@ class RadixCache(BasePrefixCache):
         self.disable = disable
         self.enable_kv_cache_events = enable_kv_cache_events
         self.kv_event_queue = []
+        self.evict_heap = []
 
         if self.token_to_kv_pool_allocator:
             self.device = self.token_to_kv_pool_allocator.device
@@ -134,6 +136,7 @@ class RadixCache(BasePrefixCache):
         self.evictable_size_ = 0
         self.protected_size_ = 0
         self._record_all_cleared_event()
+        self.evict_heap.clear()
 
     def match_prefix(self, key: List[int], **kwargs) -> Tuple[torch.Tensor, int]:
         """Find the matching prefix from the radix tree.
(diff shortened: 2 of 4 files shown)


python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -264,26 +267,29 @@ def evict(self, num_tokens: int):
...
-            if len(x.parent.children) == 0:
-                heapq.heappush(leaves, x.parent)
-
-            self._record_remove_event(x)
+            # Free memory and Update state
+            node_value_len = len(node.value)
+            self.token_to_kv_pool_allocator.free(node.value)
+            node.value = None  # Mark node as evicted
+            num_evicted += node_value_len
+            self._delete_leaf(node)
```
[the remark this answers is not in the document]
sushah13: Addressed comments in follow up commit

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -429,10 +445,8 @@ def _print_helper(self, node: TreeNode, indent: int):
                 ), f"{key=}, {self.get_child_key_fn(child.key)=}"
 
     def _delete_leaf(self, node):
-        for k, v in node.parent.children.items():
-            if v == node:
-                break
-        del node.parent.children[k]
+        if node.child_key in node.parent.children:
+            del node.parent.children[node.child_key]
         self.evictable_size_ -= len(node.key)
```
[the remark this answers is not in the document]
sushah13: Addressed comments in follow up commit

xiezhq-hermann:
Thank you for this PR, this is indeed something we wanted to accelerate the eviction. Would you mind making the same changes to hiradixtree as well? It's sure ok if you don't have the bandwidth then I can incorporate it on that side.

sushah13:
> Thank you for this PR, this is indeed something we wanted to accelerate the eviction. Would you mind making the same changes to hiradixtree as well? It's sure ok if you don't have the bandwidth then I can incorporate it on that side.

Sure thing - let me update the PR tonight with changes in Hiradix cache as well. Thank you!

sushah13:
> Thank you for this PR, this is indeed something we wanted to accelerate the eviction. Would you mind making the same changes to hiradixtree as well? It's sure ok if you don't have the bandwidth then I can incorporate it on that side.

Let me do the hiradixtree optimizations in the different PR. I want to do some more detailed analysis and comparison before I publish those changes (can do it right after this is merged). For now, I updated hiradix_cache to work along with radix_cache optimizations.

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -264,26 +267,32 @@ def evict(self, num_tokens: int):
...
-            if len(x.parent.children) == 0:
-                heapq.heappush(leaves, x.parent)
-
-            self._record_remove_event(x)
+            # Free memory and Update state
+            node_value_len = len(node.value)
+            self.token_to_kv_pool_allocator.free(node.value)
+            node.value = None  # Mark node as evicted
+            # Update evictable size here since value is freed;
+            # do not update again in _delete_leaf()
```
xiezhq-hermann, line 281: No need to update node.value to None as it is to be deleted

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -264,26 +267,32 @@ def evict(self, num_tokens: int):
...
+            self.evictable_size_ -= node_value_len
```
xiezhq-hermann, line 282: I think we can keep this in _delete_leaf still which is a clear semantic.

python/sglang/srt/mem_cache/hiradix_cache.py
```diff
@@ -82,6 +82,9 @@ def get_height(self, node: TreeNode):
         return height
 
     def write_backup(self, node: TreeNode, write_back=False):
+        # Nothing to copy if the node is already evicted.
+        if node.value is None:
```
xiezhq-hermann, line 86: I think there is no need for this.

python/sglang/srt/mem_cache/hiradix_cache.py
```diff
@@ -198,9 +209,15 @@ def _evict_backuped(self, node: TreeNode):
...
-        self.cache_controller.mem_pool_device_allocator.free(node.value)
         num_evicted = len(node.value)
-        self._delete_leaf(node)
+        self.cache_controller.mem_pool_device_allocator.free(node.value)
+
+        self.evictable_size_ -= num_evicted  # keep GPU counters in sync
+        node.value = None  # mark as evicted
+
+        # Remove the node only if it has no (even evicted) children
+        if len(node.children) == 0:
```
xiezhq-hermann, line 219: this should be determined already

xiezhq-hermann:
@sushah13 thanks again for the PR. I think we should revoke the change about hicache for now and the rest on the base radix_cache looks good to me.

sushah13:
> @sushah13 thanks again for the PR. I think we should revoke the change about hicache for now and the rest on the base radix_cache looks good to me.

The hiradix_cache changes aren't optimizations. They are to compliment the changes in radix cache so that the functionality doesn't break.

python/sglang/srt/mem_cache/radix_cache.py
sushah13: Agreed. If the node is immediately deleted and garbage-collected, setting node.value = None is redundant. However, if _delete_leaf() doesn’t fully nullify the node (or deletion is conditional), it may be safer to clear explicitly to avoid stale references.

python/sglang/srt/mem_cache/radix_cache.py
sushah13: I would disagree. The evictable_size_ tracks memory currently freeable. Adjusting it before _delete_leaf() prevents double-decrement bugs if _delete_leaf() also triggers eviction logic or tracking.

I would suggest leaving the decrement where it is (immediately after freeing). Moving it to _delete_leaf() risks coupling unrelated concerns (memory vs. tree structure).

python/sglang/srt/mem_cache/hiradix_cache.py
sushah13: I agree (conditionally) — If evict() guarantees a node is a leaf before calling _delete_leaf, the check is redundant. But if eviction can happen preemptively on non-leaf nodes, this check is necessary. I am not confident in evict() ensuring leaf-ness, would prefer keeping the check.

python/sglang/srt/mem_cache/hiradix_cache.py
sushah13: This check is necessary to prevent redundant or incorrect backup logic:

- If node.value is None, it’s already evicted. There's literally nothing on device to copy to host.
- Continuing without this check would likely lead to: 
    - Unnecessary allocation on host
    - Potential crash in self.cache_controller.write() expecting valid device indices

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -264,26 +267,32 @@ def evict(self, num_tokens: int):
...
         num_evicted = 0
-        while num_evicted < num_tokens and len(leaves):
-            x = heapq.heappop(leaves)
-
-            if x == self.root_node:
-                break
-            if x.lock_ref > 0:
+        while self.evict_heap and num_evicted < num_tokens:
+            node = heapq.heappop(self.evict_heap)
+            if node.lock_ref > 0 or node == self.root_node or node.evicted:
```
hanming-lu, line 273: suggest to be 
```
if node.lock_ref > 0:
  continue
assert node != self.root_node
```

`node.evicted` is undefined.

hanming-lu: after looking at the logic more, I think you are referring to `node.value is None` here instead of `node.evicted`

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -429,11 +449,8 @@ def _print_helper(self, node: TreeNode, indent: int):
                 ), f"{key=}, {self.get_child_key_fn(child.key)=}"
 
     def _delete_leaf(self, node):
-        for k, v in node.parent.children.items():
-            if v == node:
-                break
-        del node.parent.children[k]
-        self.evictable_size_ -= len(node.key)
+        if node.child_key in node.parent.children:
```
hanming-lu, line 452: this if is not needed, instead it should be an assert, after checking `node.value is None` properly.

python/sglang/srt/mem_cache/radix_cache.py
hanming-lu: IIUC, there's an implicit assumption that child's last_access_time is always older than parent here because the heap also contains internal nodes. the assumption doesn't seem to hold? should also check if the node is a leaf node, if not, continue.

hanming-lu:
Could you please provide the details on how you evaluated the performance gain for radix cache so we can repro? thanks.

merrymercy:
Thanks for the contribution, but we will close this one. Feel free to reopen if you think it is still necessary.

- Please try this c++ implementation. It should be much faster https://github.com/sgl-project/sglang/pull/7369
- @hanming-lu tested this on our benchmark but did not found noticeable perf difference between this and main. Could you tell us the exact instructions on how to reproduce?
- The change in schedule_batch.py is picked by https://github.com/sgl-project/sglang/pull/8724

PR #5113, opened 2025-04-07
Support MHA with chunked prefix cache for DeepSeek chunked prefill

Fridge003:
## Motivation

The current implementation of MLA is slow when when handling long prefix lengths, such as sequences with 32k input length, as highlighted in issues like #5031.

Profiling revealed that the attention kernel's slow runtime is the primary bottleneck.  Currently, we perform absorption during chunked prefilling, but this approach is inefficient for long prefix lengths, as MLA with absorption has overly large computational intensity when `q` contains many tokens. Conversely, while multi-head attention (MHA) is computationally efficient, materializing key-value (KV) heads often leads to out-of-memory (OOM) errors. 

we propose dividing long prefix KV caches into multiple chunks. The output tensors and log-sum-exp (LSE) states for each chunk can be merged into a single accumulated output and LSE state. This approach enables MHA during chunked prefilling while avoiding OOM issues.

Inspired by solutions in [vllm](https://github.com/vllm-project/vllm/blob/main/vllm/v1/attention/backends/mla/common.py) and [flashinfer](https://docs.flashinfer.ai/tutorials/recursive_attention.html), we propose dividing long prefix KV caches into multiple chunks.The output tensors and log-sum-exp (LSE) states for each chunk can be merged into a single accumulated output and LSE state. This approach enables MHA during chunked prefilling while avoiding OOM issues.

In our design, the chunked prefix cache is enabled by default for DeepSeek models (unless page_size > 1). Users can disable it by passing the server argument `--disable-chunked-prefix-cache`. Currently, the chunked prefix cache is supported only on the FA3 backend.

For short input cases (e.g., 128 or 256 tokens), this optimization introduces overhead that degrades performance. Therefore, it is activated only when the total prefix cache length exceeds a threshold (currently set to 8,192 tokens). In the future, a more refined trigger condition should be developed.

Really want to thank @ispobock for helping with debugging, and thank @DefTruth for implementing merge_state kernel~

## Modifications

- Add prefix chunk preparing and management features for `ForwardBatch`.
- Change the logic of forward function dispatching in deepseek models.
- Add a new forward method for deepseek model, which handles chunked prefilling with chunked prefix caches.
- Add MHA logic in FlashAttention backend with `flash_attn_with_varlen_func` API.
- Add a test for checking chunking logic.

## TODO List
- [x] Add relevant tests and run through the pipeline
- [x] Align accuracy
- [x] Profile and optimize performance
- [x] Benchmarking
- [x] Add a server argument that turns this feature on/off
[...]

PR #6991, opened 2025-06-09
[server arg] better arg help. disable chunked prefix cache.

jimmy-evo:
disable chunked prefix cache when disable-radix-cache is True

## Motivation

when disable radix prefix cache, I found chunked-prefix-cache is True. should be disabled.

related PR: #5113 

## Modifications

set disable_chunked_prefix_cache when radix prefix cache is disabled.

## Checklist

### DIFF
--- a/python/sglang/srt/model_executor/model_runner.py
+++ b/python/sglang/srt/model_executor/model_runner.py
@@ -406,8 +406,9 @@ class ModelRunner:
                     f"Automatically turn of --chunked-prefill-size as it is not supported for "
                     f"{self.model_config.hf_config.model_type}"
                 )
-
-        if not self.use_mla_backend:
+        if server_args.disable_radix_cache:
+            server_args.disable_chunked_prefix_cache = True
+        elif not self.use_mla_backend:
             server_args.disable_chunked_prefix_cache = True
         elif self.page_size > 1:
             logger.info("Disable chunked prefix cache when page size > 1.")
--- a/python/sglang/srt/server_args.py
+++ b/python/sglang/srt/server_args.py
@@ -1411,7 +1411,7 @@ class ServerArgs:
         parser.add_argument(
             "--disable-chunked-prefix-cache",
             action="store_true",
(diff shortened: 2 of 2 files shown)


python/sglang/srt/model_executor/model_runner.py
```diff
@@ -406,8 +406,9 @@ def model_specific_adjustment(self):
                     f"Automatically turn of --chunked-prefill-size as it is not supported for "
                     f"{self.model_config.hf_config.model_type}"
                 )
-
-        if not self.use_mla_backend:
+        if server_args.disable_radix_cache:
```
Fridge003, line 409: This is not true.. Chunked prefix cache can be used when radix cache is disabled.

jimmy-evo: but why? there is no prefix cache.

jimmy-evo: > This is not true.. Chunked prefix cache can be used when radix cache is disabled.

when radix cache is disabled, is there any prefix cache?

Fridge003: Here prefix cache means kv cache, which is different from radix cache.

jimmy-evo: > Here prefix cache means kv cache, which is different from radix cache.

```python
if (
                forward_batch.forward_mode.is_extend()
                and not self.disable_chunked_prefix_cache
                and not forward_batch.forward_mode.is_target_verify()
                and not forward_batch.forward_mode.is_draft_extend()
                and (
                    sum_extend_prefix_lens >= self.chunked_prefix_cache_threshold
                    or sum_extend_prefix_lens == 0
                )
            ):
                return AttnForwardMethod.MHA_CHUNKED_KV
```
but here use when forward_mode.is_extend()  
no kv cache if disable radix-cache.
or you mean there is any other way to prefix caching?

Fridge003: KV cache will still be used if disabling radix-cache. It's just not managed by radix tree, and there will be many repeatedly computed tokens.

PR #7019, opened 2025-06-10
Add `CMakeLists.txt` binary_dir

EduardDurech:
Fixes error related to out of tree src https://stackoverflow.com/questions/33896389/cmake-add-subdirectory-error-when-specifying-an-out-of-tree-source-a-binary-di


Alcanderian:
@zyksir

zyksir:
Thank you for pointing it out! I will run the build command to check it. Just previously I didn't see error while building, which build command did you use?

EduardDurech:
> Thank you for pointing it out! I will run the build command to check it. Just previously I didn't see error while building, which build command did you use?

```shell
CMAKE_BUILD_PARALLEL_LEVEL=16 python -m build --wheel \
                                        -Ccmake.define.SGL_KERNEL_ENABLE_BF16=1 \
                                        -Ccmake.define.SGL_KERNEL_ENABLE_SM90A=1 \
                                        -Ccmake.define.SGL_KERNEL_ENABLE_FA3=1 \
                                        --no-isolation --outdir=<foo>
```

Didn't used to get this on older versions with
```shell
CMAKE_BUILD_PARALLEL_LEVEL=16 pip wheel . --no-deps --no-cache-dir --wheel-dir <foo>
```

Not sure if it's `python -m build --wheel` or new version, but regardless it's good practice to have an explicit `binary_dir`, mandatory if there are any sources out of tree https://cmake.org/cmake/help/latest/command/add_subdirectory.html

zyksir (commented):
The original CMakeLists.txt works well while `make build` under sgl-kernel and build.sh as well. @EduardDurech if you have time you can use `make build` to build whl.
In addition, I have tested this change and it worked as well. Therefore I will Approve this PR.

zyksir (approved):

EduardDurech:
@zyksir yea I know, I just have it for consistency in Dockerfile 😃 the changes shouldn't affect the documented method, it will make it more robust for others

Alcanderian (approved):

EduardDurech:
@zhyncs this is a very minimal change which makes sgl-kernel build more robust

EduardDurech:
can someone merge please @zhyncs @merrymercy

EduardDurech:
@zyksir @Alcanderian could one of you guys contact the maintainers please, it's been 2 months, there's also other hardware-related errors that no one's maintaining #8813

mickqian (approved):

PR #4836, opened 2025-03-28
Introduce moe_dense_tp_size to fix dense layer errors in DeepSeek V3 + 4x8xH100

fzyzcjy:
This code depends on https://github.com/sgl-project/sglang/pull/5190 and https://github.com/sgl-project/sglang/pull/5205 and please subtract diff from there

## Speed tests

4x8xH100 cannot run without this PR, thus 2x8xH100 is tested.

Before PR:

```
python -m sglang.launch_server --model-path deepseek-ai/DeepSeek-V3-0324 --trust-remote-code --tp 16 --dp 16 --enable-dp-attention --enable-deepep-moe --disable-cuda-graph --enable-flashmla --disable-radix-cache --decode-log-interval 1 --host 0.0.0.0 --port 33221 --nnodes 2 --dist-init-addr 10.10.38.10:23456 --chunked-prefill-size 16384 --node-rank 0
```

Run bench_serving 5 times and get `Total token throughput (tok/s)` as:

* 15715
* 19388
* 19873
* 20031
* 19922

After PR:

```
python -m sglang.launch_server --model-path deepseek-ai/DeepSeek-V3-0324 --trust-remote-code --tp 16 --dp 16 --enable-dp-attention --enable-deepep-moe --disable-cuda-graph --enable-flashmla --disable-radix-cache --decode-log-interval 1 --host 0.0.0.0 --port 33445 --nnodes 2 --dist-init-addr 10.10.38.10:34567 --chunked-prefill-size 16384 --node-rank 0 --moe-dense-tp-size 1
```

Results

* 7900
* 19884
* 20083
* 20318
* 17505

GSM8K: 93.8

benchmark and test commands

```
python3 -m sglang.bench_serving --backend sglang --port 32123 --dataset-name random --num-prompt 1024 --random-input 1000 --random-output 1 --random-range-ratio 1 --max-concurrency 1024 --host 10.10.38.10 --port 33221

(cd /host_home/primary_synced/sglang && python3 benchmark/gsm8k/bench_sglang.py --port 20000 --parallel 1400 --num-questions 1400)
```

UPDATE: more gsm8k tests

```
# baseline
[...]

PR #7037, opened 2025-06-10
Clean up server_args.py

merrymercy:
- Move all EP related things into a single group
- Sort them correctly


python/sglang/srt/server_args.py
```diff
@@ -149,6 +148,26 @@ class ServerArgs:
...
+    eplb_algorithm: str = "auto"
+    eplb_rebalance_num_iterations: int = 1000
+    eplb_rebalance_layers_per_chunk: Optional[int] = None
+    expert_distribution_recorder_mode: Optional[
+        Literal["stat", "per_pass", "per_token"]
+    ] = None
+    expert_distribution_recorder_buffer_size: Optional[int] = None
+    enable_expert_distribution_metrics: bool = False
+    deepep_config: Optional[str] = None
+    moe_dense_tp_size: Optional[int] = None
```
fzyzcjy, line 169: not sure whether this is ep-related

fzyzcjy (approved):

python/sglang/srt/server_args.py
zhyncs: @ch-wan 😓

fzyzcjy: this is implemented by me https://github.com/sgl-project/sglang/pull/4836

I said "not sure" b/c theoretically speaking it is about how dense layers in deepseek behaves (its TP size) and is unrelated to EP, but on the other hand currently its primary usage is use with EP flags. So I personally think maybe consider it as non EP, but I am ok for whatever choice.

PR #7080, opened 2025-06-11
internvl huggingface version

hehesangsj:
## Motivation

This branch implements support for deploying the Hugging Face version of InternVL models. Previously, deployment was only based on the dev version. Now it supports deploying models hosted on Hugging Face, such as https://huggingface.co/OpenGVLab/InternVL3-8B-hf.

## Modifications

- Added support for loading and deploying InternVL models from the Hugging Face repository.
- Updated model loading logic to be compatible with the Hugging Face model structure and checkpoint format.
- Tested adaptation with InternVL3-8B-h to ensure proper functionality.


python/sglang/srt/managers/multimodal_processors/internvl_hf.py
```diff
@@ -0,0 +1,245 @@
+# Adapted from https://huggingface.co/OpenGVLab/InternVL2-4B/blob/main/modeling_intern_vit.py
```
mickqian: both processor and model file seems to have a lot of duplication with `2_5`. Can you try merging this into the previous one?

python/sglang/srt/models/internvl_hf.py
```diff
@@ -0,0 +1,691 @@
...
+                self.position_embeddings[:, :1, :],
+                self._get_pos_embed(self.position_embeddings[:, 1:, :], height, width),
+            ],
+            dim=1,
+        )
+        embeddings = embeddings + position_embedding.to(target_dtype)
+        return embeddings
+
+
+class InternRMSNorm(nn.Module):
```
mickqian: you can try using `RMSNorm` of sglang, they are basically the same

python/sglang/srt/managers/multimodal_processors/internvl_hf.py
mickqian: You can reference [here](https://github.com/mickqian/sglang/commit/d6eea011ce2b4b5f680e8810c37b8c31a0e6b902#diff-11d038483a1b74766062e87e820e767b96aa5920142f20c27fe4604021658fed)

mickqian (changes requested):

python/sglang/srt/models/internvl_hf.py
mickqian: Also for FlashAttention part, you can reference the `2_5`. My suggestion is try to reuse as much existing code as possible

mickqian (approved):

PR #7113, opened 2025-06-12
[FIX] make return type as  EagleVerifyOutput

pengchengneo:
This function declares to return the torch.tensor type, but actually returns EagleVerifyOutput, which can be very confusing to read.


pengchengneo:
@Ying1123

PR #4058, opened 2025-03-04
Let `bench_one_batch` support `enable_dp_attention`

fzyzcjy:
## Motivation

Before fix, enable_dp_attention=false, it works:

<details>
<summary>Details</summary>

```
python3 -m sglang.bench_one_batch --model-path deepseek-ai/DeepSeek-Coder-V2-Lite-Instruct --tp 2 --correct --trust-remote-code
[2025-03-04 00:23:46 TP1] MLA optimization is turned on. Use triton backend.
[2025-03-04 00:23:46 TP1] Init torch distributed begin.
[2025-03-04 00:23:47 TP0] MLA optimization is turned on. Use triton backend.
[2025-03-04 00:23:47 TP0] Init torch distributed begin.
[2025-03-04 00:23:47 TP0] sglang is using nccl==2.21.5
[2025-03-04 00:23:47 TP1] sglang is using nccl==2.21.5
[2025-03-04 00:23:48 TP0] Load weight begin. avail mem=77.18 GB
[2025-03-04 00:23:48 TP1] Load weight begin. avail mem=77.18 GB
[2025-03-04 00:23:48 TP0] The following error message 'operation scheduled before its operands' can be ignored.
[2025-03-04 00:23:48 TP1] The following error message 'operation scheduled before its operands' can be ignored.
[2025-03-04 00:23:49 TP0] Using model weights format ['*.safetensors']
[2025-03-04 00:23:49 TP1] Using model weights format ['*.safetensors']
Loading safetensors checkpoint shards:   0% Completed | 0/4 [00:00<?, ?it/s]
Loading safetensors checkpoint shards:  25% Completed | 1/4 [00:00<00:01,  1.57it/s]
Loading safetensors checkpoint shards:  50% Completed | 2/4 [00:01<00:01,  1.16it/s]
Loading safetensors checkpoint shards:  75% Completed | 3/4 [00:02<00:00,  1.05it/s]
Loading safetensors checkpoint shards: 100% Completed | 4/4 [00:03<00:00,  1.01s/it]
Loading safetensors checkpoint shards: 100% Completed | 4/4 [00:03<00:00,  1.05it/s]

[2025-03-04 00:23:53 TP0] Load weight end. type=DeepseekV2ForCausalLM, dtype=torch.bfloat16, avail mem=62.40 GB
[2025-03-04 00:23:53 TP1] Load weight end. type=DeepseekV2ForCausalLM, dtype=torch.bfloat16, avail mem=62.40 GB
[...]

PR #7169, opened 2025-06-14
adapt bench_one_batch on dp-attention

zyksir:
## Motivation

Previous [this PR](https://github.com/sgl-project/sglang/pull/4058) support made bench_one_batch support enable_dp_attention on correctness test. Recently bench_one_batch has some issues about dp-attention
1. the API of `prepare_dp_attn_batch_raw` is changed, while the arguments on `bench_one_batch.py` does not change. This can cause error if we run `bench_one_batch.py` with dp-attention cc @f
2. If we run latency test, the decode does not use cuda graph. This is because we set `cuda_graph_max_bs=max(batch_size)`, while the actually batch_size is `dp_size * batch_size` since we are running `dp_size` batches at the same time. This will cause the actual batch_size to be larger than `cuda_graph_max_bs`.

## Modifications

1. change the arguments of `prepare_dp_attn_batch_raw`
2. modify the batch size so that the comparison of `--tp` and `--tp --dp --enable-dp-attention` is fair.

This is the command I used. The following result does not mean that dp-attention is worse since in bench_one_batch, the advantage of saving KVCache is not tested. But it does show some feature work if we want to improve dp-attention.
```shell
python3 -m sglang.bench_one_batch --model Qwen/Qwen3-30B-A3B --batch 32 --input-len 256 --output-len 32 --tp 4
python3 -m sglang.bench_one_batch --model Qwen/Qwen3-30B-A3B --batch 32 --input-len 256 --output-len 32 --tp 4 --dp 4 --enable-dp-attention 
```
This is the result of tp4 on H100:
image
This is the result of tp4 and dp4 on H100:
image

By adding `--profile`, we can see that the drop of performance comes from:
1. The addition memory bound kernels added before and after all-reduce
image
2. The GEMM in Attention Layer is slower since the weight is larger.
This is qkv_proj in dp-attention
image

This is qkv_proj in baseline.
image

## Checklist

### DIFF
--- a/python/sglang/bench_one_batch.py
+++ b/python/sglang/bench_one_batch.py
@@ -68,6 +68,7 @@ from sglang.srt.sampling.sampling_params import SamplingParams
 from sglang.srt.server_args import PortArgs, ServerArgs
 from sglang.srt.speculative.spec_info import SpeculativeAlgorithm
 from sglang.srt.utils import (
+    DeepEPMode,
     configure_logger,
     get_bool_env_var,
     kill_process_tree,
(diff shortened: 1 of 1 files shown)


yhyang201:
This is very useful. Can this be merged? cc @zhyncs

PR #7188, opened 2025-06-14
feat(server): Add Windows support for http_server

celsowm:
sglang.server needs to supports windows (because uvloop is linux and macOs only)

## Motivation

uvloop is linux and macOs only, very sad :(

## Modifications

just few lines on http_server.py


celsowm:
hi @zhaochenyang20 ! Its a very simple PR but would be very nice to run server on Windows too

hnyls2002:
Thanks @celsowm! Closing - sglang targets Linux/CUDA and we don't currently support Windows.

PR #7213, opened 2025-06-15
[EAGLE] Refactor code for page size > 1 & more simplifications

merrymercy:
- Unify the kv cache layout during the decode of the draft model for page size > 1 and page size =1. It changed the draft kv cache layout in both `forward_batch.out_cache_loc` and `forward_batch.req_to_token_pool`. This only affects the draft decode phase. If any attention backend builds the page table based on `forward_batch.out_cache_loc` or `forward_batch.req_to_token_pool.req_to_token`, they need to be updated accordingly.
  -  The layout of forward_batch.out_cache_loc is changed from (bs, num_speculative_steps, topk) to (bs, topk, num_speculative_steps) 
  - The layout of forward_batch.req_to_token_pool for a single request is changed as follows
    ```
        # Before: Layout of the req_to_token for a single request.
        # [       iter 0         ] [       iter 1         ]
        # [topk=0, topk=1, topk=2] [topk=0, topk=1, topk=2]
    ```
    ```
        # After: Layout of the req_to_token for a single request.
        # [       topk 0         ] [       topk 1         ]
        # [iter=0, iter=1, iter=2] [iter=0, iter=1, iter=2]
    ```
- Simplify the verify function.
- Now it supports page size > 1 and top-k > 1 for the flashinfer backend correctly. This combination is still not fully optimized due to some extra page operations and device sync.
- There is still one remaining todo item to support page size >1 and topk > 1 for other attention backends that really runs page size > 1.  https://github.com/sgl-project/sglang/blob/5f1ab327176297b984cfb957fc84aa7db1b4856f/python/sglang/srt/speculative/eagle_worker.py#L407-L414 

This PR re-applies https://github.com/sgl-project/sglang/pull/7163, which was reverted by https://github.com/sgl-project/sglang/pull/7210

### DIFF
--- a/python/sglang/srt/layers/attention/flashinfer_backend.py
+++ b/python/sglang/srt/layers/attention/flashinfer_backend.py
@@ -1049,14 +1049,13 @@ class FlashInferMultiStepDraftBackend:
             kv_indices_buffer,
             self.kv_indptr,
             forward_batch.positions,
-            num_seqs,
-            self.topk,
             self.pool_len,
             kv_indices_buffer.shape[1],
             self.kv_indptr.shape[1],
             next_power_of_2(num_seqs),
             next_power_of_2(self.speculative_num_steps),
             next_power_of_2(bs),
+            self.page_size,
         )
 
         assert forward_batch.spec_info is not None
--- a/python/sglang/srt/layers/attention/flashinfer_mla_backend.py
+++ b/python/sglang/srt/layers/attention/flashinfer_mla_backend.py
@@ -789,6 +789,7 @@ class FlashInferMLAMultiStepDraftBackend:
 
         # Cached variables for generate_draft_decode_kv_indices
         self.pool_len = model_runner.req_to_token_pool.req_to_token.shape[1]
+        self.page_size = model_runner.server_args.page_size
 
     def common_template(
         self,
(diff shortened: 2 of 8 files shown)


byjiang1996:
Hi @merrymercy , thanks for the PR!

I did some profiling on `get_last_loc_large_page_size_top_k_1` method. It seems `w/o torch.compile`,  the method actually runs faster: p50 93us -> 150us.

Do you have any idea? I can help remove those two torch.compile tags if you think they are safe to remove.

## W/o torch.compile: p50 `get_last_loc_large_page_size_top_k_1` latency = 93us
Query: 
image
Example: 
image

## With torch.compile: p50 `get_last_loc_large_page_size_top_k_1` latency = 150us
Query:
image
Example: 
image

## Command
```
Server: python3 -m sglang.launch_server --speculative-num-steps 3 --speculative-eagle-topk 1 --speculative-num-draft-tokens 4 --attention-backend fa3 --trust-remote-code --dtype float16 --speculative-algo EAGLE3 --model-path /shared/public/models/meta-llama/Meta-Llama-3.1-8B-Instruct --speculative-draft-model-path /shared/public/elr-models/jamesliu1/sglang-EAGLE3-Llama-3.1-Instruct-8B/e5ed08d66f528a95ce89f5d4fd136a28f6def714 --page-size 128
Client: python3 -m sglang.bench_serving --backend sglang --model /shared/public/models/meta-llama/Meta-Llama-3.1-8B-Instruct --dataset-name random --request-rate 16 --num-prompts 128 --random-range-ratio 1.0 --random-input-len 1024 --random-output-len 1024 --profile
```

PR #7216, opened 2025-06-15
Support radix cache for Lora feature

### LINKED ISSUE #2880: [Bug] Why can't I use multi-lora adapter and radix attention together?
### Checklist

- [X] 1. I have searched related issues but cannot get the expected help.
- [X] 2. The bug has not been fixed in the latest version.
- [X] 5. Please use English, otherwise it will be closed.

### Describe the bug

Why can't I use multi-lora adapter and radix attention together?
If I have multi-lora adapters, why not just insert the ID of the LoRA adapter before the first token?

When using a multi-lora adapter, it is extremely slow because radix attention cannot be used.

### Reproduction

https://github.com/sgl-project/sglang/blob/v0.4.1.post5/python/sglang/srt/server_args.py#L876-L881

### Environment

```
root@33e74a81f115:/sglang/python# python3 -m sglang.check_env                                                                                                                                                         

Python: 3.10.16 (main, Dec  4 2024, 08:53:37) [GCC 9.4.0]
CUDA available: True
GPU 0: NVIDIA A100-SXM4-80GB
GPU 0 Compute Capability: 8.0
CUDA_HOME: /usr/local/cuda
NVCC: Cuda compilation tools, release 12.4, V12.4.131
CUDA Driver Version: 550.127.05
PyTorch: 2.5.1+cu124
sglang: 0.4.0.post2
flashinfer: 0.1.6+cu124torch2.4
triton: 3.1.0
transformers: 4.47.0
torchao: 0.6.1
numpy: 1.26.4
aiohttp: 3.11.10
fastapi: 0.115.6
hf_transfer: 0.1.8
huggingface_hub: 0.26.3
interegular: 0.3.3
modelscope: 1.21.0
orjson: 3.10.12
packaging: 24.2
psutil: 6.1.0
pydantic: 2.10.3
multipart: 0.0.19
zmq: 26.2.0
uvicorn: 0.32.1
uvloop: 0.21.0
vllm: 0.6.4.post1
openai: 1.57.0
anthropic: 0.40.0
decord: 0.6.0
```

Fridge003:
## Motivation
Close #2880

## Modifications

To support radix cache for Lora, this PR does the following modifications:
- Add a new `LoRARadixCache` as a special version of `RadixCache`.  They are mostly the same, except that the node in `LoRARadixCache` uses `hash(lora_id + str(token_ids[0])` as its key for prefix matching, while the node in `RadixCache` uses `token_ids[0]` as key for prefix matching. Here I didn't modify the original `RadixCache` class since the different node keys will introduce a lot of modifications that will contaminate the code. 
- Add `test_lora_radix_cache` for testing the compatibility of lora and radix cache. In this test we launch a base model with two different adaptors, and the radix cache should be able to differentiate the same prompt tagged with different lora_ids.
- Enable radix cache by default in all the CI tests

## Next Steps (#9144)

Currently when enabling lora and radix cache together, some features are not supported:
- page_size > 1
- schedule policy other than fcfs
- hierarchical cache
They should be supported in the future

Also, the unified paging technique with Radix Cache is also to be supported.

## Benchmark

Changes in #10092 need to be applied to `bench_multiturn.py`
```bash
# Launching with radix cache
python3 -m sglang.launch_server --model-path meta-llama/Llama-3.1-8B-Instruct --lora-paths lora1=algoprog/fact-generation-llama-3.1-8b-instruct-lora --port 30001

# Launching without radix cache
python3 -m sglang.launch_server --model-path meta-llama/Llama-3.1-8B-Instruct --lora-paths lora1=algoprog/fact-generation-llama-3.1-8b-instruct-lora --port 30001 --disable-radix-cache

# Benchmark Commands
python3 benchmark/hicache/bench_multiturn.py --model-path meta-llama/Llama-3.1-8B-Instruct --port 30001 --lora-path lora1
```

Radix cache disabled:
```bash
Performance metrics summary:
  Total requests: 1280 at 16 requests per second
  Average TTFT: 0.53
  P90 TTFT: 1.33
  Median TTFT: 0.31
  Average latency: 10.28
  P90 latency: 15.33
  Median latency: 11.21
  Input token throughput: 16692.69 tokens per second
  Output token throughput: 641.85 tokens per second
  Request Throughput: 10.03 requests per second
  Cache Hit Rate: 0.000000
```

Radix cache enabled:
```bash
  Total requests: 1280 at 16 requests per second
  Average TTFT: 0.10
  P90 TTFT: 0.18
  Median TTFT: 0.07
  Average latency: 2.91
  P90 latency: 6.42
  Median latency: 2.13
  Input token throughput: 23926.39 tokens per second
  Output token throughput: 920.02 tokens per second
  Request Throughput: 14.38 requests per second
  Cache Hit Rate: 0.625577
```

### DIFF
--- a/docs/advanced_features/lora.ipynb
+++ b/docs/advanced_features/lora.ipynb
@@ -459,7 +459,7 @@
    "source": [
     "## Future Works\n",
     "\n",
-    "The development roadmap for LoRA-related features can be found in this [issue](https://github.com/sgl-project/sglang/issues/2929). Other features, including Unified Paging, Cutlass backend, and asynchronous loading/unloading are still under development."
+    "The development roadmap for LoRA-related features can be found in this [issue](https://github.com/sgl-project/sglang/issues/2929). Other features, including Embedding Layer,Unified Paging, Cutlass backend are still under development."
    ]
   }
  ],
--- a/docs/advanced_features/lora.ipynb
+++ b/docs/advanced_features/lora.ipynb
@@ -80,7 +80,6 @@
     "    --enable-lora \\\n",
     "    --lora-paths lora0=algoprog/fact-generation-llama-3.1-8b-instruct-lora \\\n",
     "    --max-loras-per-batch 1 --lora-backend triton \\\n",
-    "    --disable-radix-cache\n",
     "\"\"\"\n",
     ")\n",
     "\n",
@@ -140,7 +139,6 @@
     "    --lora-paths lora0=algoprog/fact-generation-llama-3.1-8b-instruct-lora \\\n",
     "    lora1=Nutanix/Meta-Llama-3.1-8B-Instruct_lora_4_alpha_16 \\\n",
     "    --max-loras-per-batch 2 --lora-backend triton \\\n",
-    "    --disable-radix-cache\n",
     "\"\"\"\n",
     ")\n",
     "\n",
@@ -215,7 +213,6 @@
     "    --enable-lora \\\n",
     "    --cuda-graph-max-bs 2 \\\n",
     "    --max-loras-per-batch 2 --lora-backend triton \\\n",
-    "    --disable-radix-cache\n",
     "    --max-lora-rank 256\n",
     "    --lora-target-modules all\n",
     "    \"\"\"\n",
@@ -462,7 +459,7 @@
    "source": [
     "## Future Works\n",
     "\n",
-    "The development roadmap for LoRA-related features can be found in this [issue](https://github.com/sgl-project/sglang/issues/2929). Currently radix attention is incompatible with LoRA and must be manually disabled. Other features, including Unified Paging, Cutlass backend, and dynamic loading/unloadingm, are still under development."
+    "The development roadmap for LoRA-related features can be found in this [issue](https://github.com/sgl-project/sglang/issues/2929). Other features, including Unified Paging, Cutlass backend, and asynchronous loading/unloading are still under development."
    ]
   }
  ],
--- a/docs/advanced_features/lora.ipynb
+++ b/docs/advanced_features/lora.ipynb
@@ -459,7 +459,7 @@
    "source": [
     "## Future Works\n",
     "\n",
-    "The development roadmap for LoRA-related features can be found in this [issue](https://github.com/sgl-project/sglang/issues/2929). Other features, including Embedding Layer,Unified Paging, Cutlass backend are still under development."
+    "The development roadmap for LoRA-related features can be found in this [issue](https://github.com/sgl-project/sglang/issues/2929). Other features, including Embedding Layer, Unified Paging, Cutlass backend are still under development."
    ]
   }
  ],
--- a/python/sglang/srt/managers/schedule_batch.py
+++ b/python/sglang/srt/managers/schedule_batch.py
@@ -58,6 +58,7 @@ from sglang.srt.mem_cache.allocator import (
 )
 from sglang.srt.mem_cache.base_prefix_cache import BasePrefixCache
 from sglang.srt.mem_cache.chunk_cache import ChunkCache, SWAChunkCache
+from sglang.srt.mem_cache.lora_radix_cache import LoRAKey, LoRARadixCache
 from sglang.srt.mem_cache.memory_pool import ReqToTokenPool
 from sglang.srt.mem_cache.swa_radix_cache import SWARadixCache
 from sglang.srt.metrics.collector import TimeStats
@@ -639,14 +640,26 @@ class Req:
     ):
         self.fill_ids = self.origin_input_ids + self.output_ids
         if tree_cache is not None:
-            (
-                self.prefix_indices,
-                self.last_node,
-                self.last_host_node,
-                self.host_hit_length,
-            ) = tree_cache.match_prefix(
-                key=self.adjust_max_prefix_ids(),
-            )
+            if isinstance(tree_cache, LoRARadixCache):
+                (
+                    self.prefix_indices,
+                    self.last_node,
+                    self.last_host_node,
+                    self.host_hit_length,
+                ) = tree_cache.match_prefix_with_lora_id(
+                    key=LoRAKey(
+                        lora_id=self.lora_id, token_ids=self.adjust_max_prefix_ids()
+                    ),
+                )
+            else:
+                (
+                    self.prefix_indices,
+                    self.last_node,
+                    self.last_host_node,
+                    self.host_hit_length,
+                ) = tree_cache.match_prefix(
+                    key=self.adjust_max_prefix_ids(),
+                )
         self.extend_input_len = len(self.fill_ids) - len(self.prefix_indices)
 
     def adjust_max_prefix_ids(self):
--- a/python/sglang/srt/managers/scheduler.py
+++ b/python/sglang/srt/managers/scheduler.py
@@ -130,6 +130,7 @@ from sglang.srt.managers.tp_worker_overlap_thread import TpModelWorkerClient
 from sglang.srt.managers.utils import DPBalanceMeta, validate_input_length
 from sglang.srt.mem_cache.chunk_cache import ChunkCache, SWAChunkCache
 from sglang.srt.mem_cache.hiradix_cache import HiRadixCache
+from sglang.srt.mem_cache.lora_radix_cache import LoRARadixCache
 from sglang.srt.mem_cache.radix_cache import RadixCache
 from sglang.srt.mem_cache.swa_radix_cache import SWARadixCache
 from sglang.srt.model_executor.forward_batch_info import ForwardMode, PPProxyTensors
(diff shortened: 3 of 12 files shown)


lifuhuang:
Nice! Is there any perf gains when radix cache is enabled? Also, why was is disabled at the first place?

lifuhuang:
nit: maybe we should remove this line as well 😄  

https://github.com/sgl-project/sglang/blob/96be97bfff197c96f88718cb15b2d198fe2391c3/test/srt/test_vision_openai_server_b.py#L201

lifuhuang (approved):
LGTM

zhyncs:
@Fridge003 any update?

Fridge003:
> @Fridge003 any update?

Still under working...Two of the tests cannot pass

Edenzzzz:
@Fridge003 Could you point to the failed tests? I see more than two failing.

Fridge003:
> @Fridge003 Could you point to the failed tests? I see more than two failing.

`test/srt/models/lora/test_lora.py` and `test/srt/models/lora/test_multi_lora_backend.py`. Other tests can actually pass.

test/srt/models/lora/test_lora.py
```diff
@@ -105,7 +105,6 @@ def _run_lora_multiple_batch_on_model_cases(self, model_cases: List[LoRAModelCas
                     lora_paths=[lora_adapter_paths[0], lora_adapter_paths[1]],
```
Edenzzzz: Should be `lora_adapter_paths`? Though there are other reasons for failure. I will probably do a layer-wise kv check with the prompt

Fridge003: thanks for your advice! I will check this later

zhyncs:
Baizhou is this pr ready

Fridge003:
> Baizhou is this pr ready

No, I found this feature is not as trivial as I thought. A new memory cache is needed
I might need another 1~2 weeks

Edenzzzz:
> No, I found this feature is not as trivial as I thought. A new memory cache is needed I might need another 1~2 weeks

Because of multi-lora?

Fridge003:
> > No, I found this feature is not as trivial as I thought. A new memory cache is needed I might need another 1~2 weeks
> 
> Because of multi-lora?

Yes, the same prompt might be run on different lora adapters

Edenzzzz:
How about storing the output before applying lora

lifuhuang:
> How about storing the output before applying lora

We do need to cache LoRA result as well for perf gains, otherwise we would still need to recompute LoRA each token, layer even if there is a prefix match.

Edenzzzz:
> We do need to cache LoRA result as well for perf gains, otherwise we would still need to recompute LoRA each token, layer even if there is a prefix match.

Don't think so, LoRA computation is not much, otherwise you need `num_lora` times KV cache. It's also the approach used by S-LoRA https://arxiv.org/pdf/2311.03285

Fridge003:
> > We do need to cache LoRA result as well for perf gains, otherwise we would still need to recompute LoRA each token, layer even if there is a prefix match.
> 
> Don't think so, LoRA computation is not much, otherwise you need `num_lora` times KV cache. It's also the approach used by S-LoRA https://arxiv.org/pdf/2311.03285

If we use LoRA recomputation, then all of the k-lora and q-lora needs to be recomputed for every decoding step. Even if extra time is low, this recomputation doesn't fit well in current SGLang framework.

Also, `num_lora` times KV cache  will not be a huge problem if we control the total number of served adaptors. If we unload any adaptor, its KV cache can be evicted accordingly.

Edenzzzz:
> If we use LoRA recomputation, then all of the k-lora and q-lora needs to be recomputed for every decoding step. Even if extra time is low, this recomputation doesn't fit well in current SGLang framework.

I think it just takes updating the LoRA backend?

> Also, num_lora times KV cache will not be a huge problem if we control the total number of served adaptors. If we unload any adaptor, its KV cache can be evicted accordingly.

It's hard to know how many users will use, and will need significant modifications to the kv pool.

Fridge003:
> > If we use LoRA recomputation, then all of the k-lora and q-lora needs to be recomputed for every decoding step. Even if extra time is low, this recomputation doesn't fit well in current SGLang framework.
> 
> I think it just takes updating the LoRA backend?
> 
> > Also, num_lora times KV cache will not be a huge problem if we control the total number of served adaptors. If we unload any adaptor, its KV cache can be evicted accordingly.
> 
> It's hard to know how many users will use, and will need significant modifications to the kv pool.

KV cache will be useless if we do recomputation. Since the kv vectors are the results after doing q_proj/k_proj. For applying Lora we need the activations before doing q_proj/k_proj

Edenzzzz:
I see, you are right

hnyls2002 (approved):

hnyls2002:
@Fridge003 LGTM, not a lora expert, but with two small questions:
1. Why choose to define the key `hash(lora_id + str(token_ids[0])` instead of the tuple, just interested in that.
2. In your summary and the modified code, we can see that most of the changes happen in the part of the KV matching. Is there a chance to define an abstraction about the key, not adding a new radix cache?

Fridge003:
> @Fridge003 LGTM, not a lora expert, but with two small questions:
> 
> 1. Why choose to define the key `hash(lora_id + str(token_ids[0])` instead of the tuple, just interested in that.
> 2. In your summary and the modified code, we can see that most of the changes happen in the part of the KV matching. Is there a chance to define an abstraction about the key, not adding a new radix cache?

@hnyls2002 Really want to thank you for your review~

For your questions:
1. Changing to tuple for matching is also OK. The key definition is valid as long as it contains lora information for prefix matching. It's just a matter of personal preference. 
2. Modifying the original radix cache is feasible, but not graceful. I will need to add a lot of extra logics to make it work, which will make future maintenance difficult.

yipengshenn:
@Fridge003  Hello! I'm doing some work with sglang and lora. I've read the code and here's 2 question.
1. Since #7446  have supported dynamic loading of lora, what will happened to the tree cache if I offload a lora adapter who has already hold some nodes?
2. Since multi-lora prefix matching is being solved, is there any potential difficulties for adapting this LoraRadixCache to HiLoraRadixCache? What's more, how about letting HLRCache's "Hi" support both kv nodes and lora? I'd like to contribute! 
thanks very much! orz orz orz

Fridge003:
> @Fridge003 Hello! I'm doing some work with sglang and lora. I've read the code and here's 2 question.
> 
> 1. Since [Support dynamic LoRA loading / unloading in engine/server API #7446](https://github.com/sgl-project/sglang/pull/7446)  have supported dynamic loading of lora, what will happened to the tree cache if I offload a lora adapter who has already hold some nodes?
> 2. Since multi-lora prefix matching is being solved, is there any potential difficulties for adapting this LoraRadixCache to HiLoraRadixCache? What's more, how about letting HLRCache's "Hi" support both kv nodes and lora? I'd like to contribute!
>    thanks very much! orz orz orz

@lazyyy77 Thank you for your great questions!
For question 1, I think it's really meaningful. As you say, when we try to offload a lora adaptor, all its nodes in the radix tree are supposed to be evicted (Or kept if the adaptor might be loaded again?). Current codes doesn't include such behavior, so this feature deserves further inspection.

For question 2, I'm not an expert in hierarchical cache, so I have no concept of potential difficulties. For any questions about hicache you can try contacting @xiezhq-hermann.

Edenzzzz:
For 1, I think they will be lazily evicted if the adaptor is not re-loaded (though it also depends on the depth of each lora tree, the longest tree will be evicted first).
For 2, lora weights are usually too small to require offloading? As long as key matching is correct it shouldn't be that hard to support HiCache

yipengshenn:
> > @Fridge003 Hello! I'm doing some work with sglang and lora. I've read the code and here's 2 question.
> > 
> > 1. Since [Support dynamic LoRA loading / unloading in engine/server API #7446](https://github.com/sgl-project/sglang/pull/7446)  have supported dynamic loading of lora, what will happened to the tree cache if I offload a lora adapter who has already hold some nodes?
> > 2. Since multi-lora prefix matching is being solved, is there any potential difficulties for adapting this LoraRadixCache to HiLoraRadixCache? What's more, how about letting HLRCache's "Hi" support both kv nodes and lora? I'd like to contribute!
> >    thanks very much! orz orz orz
> 
> @lazyyy77 Thank you for your great questions! For question 1, I think it's really meaningful. As you say, when we try to offload a lora adaptor, all its nodes in the radix tree are supposed to be evicted (Or kept if the adaptor might be loaded again?). Current codes doesn't include such behavior, so this feature deserves further inspection.
> 
> For question 2, I'm not an expert in hierarchical cache, so I have no concept of potential difficulties. For any questions about hicache you can try contacting @xiezhq-hermann.

Sure! Thanks a lot, and I'd like to give it a try when I finish current work~

yipengshenn:
> For 1, I think they will be lazily evicted if the adaptor is not re-loaded (though it also depends on the depth of each lora tree, the longest tree will be evicted first). For 2, lora weights are usually too small to require offloading? As long as key matching is correct it shouldn't be that hard to support HiCache

I agree with you! What's more, I know little about sglang's pcie bandwidth. How to evaluate which is better between lazyily offload and directly offload, while the latter one might merge the pieces operation but occupy the whole bandwidth?

PR #7225, opened 2025-06-16
feat: support compatibility between MTP and two-batch-overlap

Qiaolin-Yu:
## Motivation

Support MTP with two-batch-overlap

Thanks to @fzyzcjy @ch-wan @hebiao064 for their help 🚀!

## Checklist


test/srt/test_dp_attention.py
```diff
@@ -137,5 +137,86 @@ def test_gsm8k(self):
         self.assertGreater(avg_spec_accept_length, 2.5)
 
 
+class TestDPAttentionDP2TP2DeepseekV3MTPTBO(CustomTestCase):
+    @classmethod
+    def setUpClass(cls):
+        import os
+
+        # print debug log for tbo
+        os.environ["TBO_DEBUG"] = "1"
```
fzyzcjy: nit: looks like it is SGLANG_TBO_DEBUG

test/srt/test_dp_attention.py
```diff
@@ -137,5 +137,86 @@ def test_gsm8k(self):
...
+            "32",
+        ]
+        if not is_in_amd_ci():
+            other_args += ["--mem-frac", "0.7"]
+        cls.process = popen_launch_server(
+            cls.model,
+            cls.base_url,
+            timeout=DEFAULT_TIMEOUT_FOR_SERVER_LAUNCH,
+            other_args=other_args,
+            env={"SGL_ENABLE_JIT_DEEPGEMM": "1", **os.environ},
```
fzyzcjy: nit: no longer need to manually enable deepgemm - it is there by default (iirc)

python/sglang/srt/operations_strategy.py
```diff
@@ -73,6 +73,8 @@ def _compute_moe_deepseek_layer_operations_strategy_tbo(
         return _compute_moe_deepseek_blog_prefill(layer)
     elif forward_mode == ForwardMode.DECODE:
         return _compute_moe_deepseek_blog_decode(layer)
+    elif forward_mode == ForwardMode.TARGET_VERIFY:
+        return _compute_moe_deepseek_blog_decode(layer)
```
fzyzcjy: nit

```suggestion
    elif forward_mode == ForwardMode.DECODE or forward_mode == ForwardMode.TARGET_VERIFY:
        return _compute_moe_deepseek_blog_decode(layer)
```

python/sglang/srt/operations_strategy.py
```diff
@@ -148,6 +150,8 @@ def _compute_moe_qwen3_layer_operations_strategy_tbo(
         return _compute_moe_qwen3_prefill(layer)
     elif forward_mode == ForwardMode.DECODE:
         return _compute_moe_qwen3_decode(layer)
+    elif forward_mode == ForwardMode.TARGET_VERIFY:
```
fzyzcjy: (same)

python/sglang/srt/model_executor/forward_batch_info.py
```diff
@@ -352,7 +352,10 @@ def init_new(
 
         if ret.forward_mode.is_idle():
             ret.positions = torch.empty((0,), device=device)
-            TboForwardBatchPreparer.prepare(ret)
+            if not model_runner.is_draft_worker:
```
fzyzcjy: nit: shall we put this logic inside TboForwardBatchPreparer.prepare to avoid copy-paste the logic twice

python/sglang/srt/two_batch_overlap.py
```diff
@@ -317,6 +364,9 @@ def filter_batch(
             old_value = getattr(batch, key)
             if old_value is None:
                 continue
+            elif batch.forward_mode.is_target_verify() and key.startswith("extend_"):
```
fzyzcjy: nit: a bit worried whether this is too dynamic. I would personally suggest listing all extend_blahblah manually to the existing long list, probably in a separate section with a comment

python/sglang/srt/two_batch_overlap.py
```diff
@@ -416,18 +466,27 @@ def compute_tbo_children_num_token_non_padded_raw(
 
     @classmethod
     def _compute_split_token_index(cls, batch: ForwardBatch):
+        if batch.forward_mode.is_target_verify():
+            draft_token_num_per_batch = batch.spec_info.draft_token_num
+        else:
+            draft_token_num_per_batch = None
```
fzyzcjy: nit: looks like this logic is repeated several times, maybe we can extract to a short function

python/sglang/srt/two_batch_overlap.py
```diff
@@ -31,7 +32,11 @@ def compute_split_seq_index(
     forward_mode: "ForwardMode",
     num_tokens: int,
     extend_lens: Optional[Sequence[int]],
+    draft_token_num_per_batch: Optional[int],
 ) -> Optional[int]:
+    if forward_mode.is_target_verify():
```
fzyzcjy: nit: (not only for this line but for whole pr) it seems target verify behaves somehow like a generalization of normal decode, thus shall we somehow unify the mtp branch and the decode branch. For example, here we may unify w/ the is_decode branch as:

```
if is_target_verify or is_decode:
  return (num_tokens // token_num_per_batch) // 2
```

where token_num_per_batch is the draft token num, or exactly 1 in decode.

fzyzcjy (commented):
some optional nits

python/sglang/srt/layers/attention/tbo_backend.py
```diff
@@ -119,21 +119,40 @@ def _init_forward_metadata_cuda_graph_children(
...
+                    capture_num_tokens == bs
+                ), "Only support num_tokens==bs currently unless target-verify mode"
+            num_tokens = bs
+        else:
+            draft_token_num = spec_info.draft_token_num
+            if capture_num_tokens is not None:
+                assert (
+                    draft_token_num * bs == capture_num_tokens
+                ), f"For target-verify mode, num_tokens ({capture_num_tokens}) should be equal to draft_token_num ({draft_token_num}) * bs ({bs})"
+            num_tokens = draft_token_num * bs
```
fzyzcjy: e.g. for this seems we can unify as mentioned above

python/sglang/srt/layers/attention/tbo_backend.py
```diff
@@ -119,21 +119,40 @@ def _init_forward_metadata_cuda_graph_children(
...
+        bs_child_left = (
+            num_tokens_child_left
+            if not forward_mode.is_target_verify()
+            else tbo_split_seq_index
+        )
+        bs_child_right = (
+            num_tokens_child_right
+            if not forward_mode.is_target_verify()
+            else bs - tbo_split_seq_index
+        )
```
fzyzcjy: nit: I would personally feel the following is tiny slightly more readable

```
if forward_mode.is_target_verify()
    bs_child_left = tbo_split_seq_index
    bs_child_right = bs - tbo_split_seq_index
else
        bs_child_left = num_tokens_child_left
        bs_child_right = num_tokens_child_right
```

what's more, maybe we should use the former unconditionally, i.e.

```
    bs_child_left = tbo_split_seq_index
    bs_child_right = bs - tbo_split_seq_index
```

since it seems we want children's "bs", thus we use "seq index" instead of "token index"

python/sglang/srt/layers/attention/tbo_backend.py
```diff
@@ -208,16 +226,28 @@ def _init_forward_metadata_cuda_graph_split(
...
+                )
+            )
+        else:
+            assert (
+                capture_num_tokens == bs
+            ), "Only support num_tokens==bs currently unless target-verify mode"
+            ans.update(
+                dict(
+                    num_tokens=output_bs,
+                )
```
fzyzcjy: (same as above)

python/sglang/srt/model_executor/forward_batch_info.py
Qiaolin-Yu: If we put the logic inside `TboForwardBatchPreparer.prepare`, do you prefer to pass the `model_runner` or pass a bool like 'is_draft_worker'?

fzyzcjy: I am ok for both, maybe latter tiny slightly better

python/sglang/srt/two_batch_overlap.py
```diff
@@ -26,17 +27,33 @@
...
 # TODO: may smartly disable TBO when batch size is too small b/c it will slow down
 def compute_split_seq_index(
     forward_mode: "ForwardMode",
     num_tokens: int,
     extend_lens: Optional[Sequence[int]],
+    token_num_per_batch: Optional[int],
 ) -> Optional[int]:
+    if forward_mode.is_target_verify() or forward_mode.is_decode():
+        assert token_num_per_batch is not None
+        return (num_tokens // token_num_per_batch) // 2
```
fzyzcjy: tiny nit: shall we put it to the original location (after is_extend branch) to minimize code diff

python/sglang/srt/two_batch_overlap.py
```diff
@@ -67,12 +84,14 @@ def compute_split_token_index(
     split_seq_index: int,
     forward_mode: "ForwardMode",
     extend_seq_lens: Optional[Sequence[int]],
+    token_num_per_batch: Optional[int],
 ) -> int:
-    if forward_mode.is_extend():
+    if forward_mode.is_target_verify() or forward_mode.is_decode():
```
fzyzcjy: (same)

python/sglang/srt/two_batch_overlap.py
```diff
@@ -26,17 +27,33 @@
 # -------------------------------- Compute Basic Info ---------------------------------------
 
 
+def get_token_num_per_batch(
```
fzyzcjy: question: shall we call it "token_num_per_seq" (and change everywhere related), since I personally feel like:

* suppose we have batch size 256, and MTP gives one extra token
* "token_num_per_batch" means "256 x 2 = 512"
* token_num_per_seq means "2"

And here we want to mean the "2" instead of "512"

Please correct me if my understanding is wrong!

python/sglang/srt/two_batch_overlap.py
```diff
@@ -317,6 +373,16 @@ def filter_batch(
...
+            elif batch.forward_mode.is_target_verify() and (
+                key == "extend_seq_lens"
+                or key == "extend_prefix_lens"
+                or key == "extend_start_loc"
+                or key == "extend_prefix_lens_cpu"
+                or key == "extend_seq_lens_cpu"
+                or key == "extend_logprob_start_lens_cpu"
+            ):
+                output_dict[key] = None
+                continue
```
fzyzcjy, line 472: nit: oh I mean maybe we can put it to here

image

fzyzcjy (commented):
only optional tiny nits

python/sglang/srt/two_batch_overlap.py
Qiaolin-Yu: For `TARGET_VERIFY` mode, `is_extend` will return true. So I think we could not change the order?

python/sglang/srt/two_batch_overlap.py
Qiaolin-Yu: But we still need to write something like `if target-verify: ...`, right? So maybe it will be very similar.

python/sglang/srt/two_batch_overlap.py
Qiaolin-Yu: You are definitely correct. Updated.

python/sglang/srt/two_batch_overlap.py
fzyzcjy: Looks reasonable. I am wondering whether we should use is_extend or use == EXTEND though. When I wrote that code indeed I thought about latter.

python/sglang/srt/two_batch_overlap.py
fzyzcjy: Yes but I personally feel is a tiny bit unified (keep all "fill values such as none" logic in one place). Anyway this is optional and feel free to keep the current one.

EDIT: oops ignore me, now I realize you are handling the explicitly listed keys here (it gets folded in github gui...)

fzyzcjy (approved):

python/sglang/srt/two_batch_overlap.py
Qiaolin-Yu: Yes the latter is better fs, updated.

python/sglang/srt/layers/attention/tbo_backend.py
```diff
@@ -119,21 +119,27 @@ def _init_forward_metadata_cuda_graph_children(
...
+            assert (
+                capture_num_tokens == bs * token_num_per_seq
+            ), "For target-verify or decode mode, num_tokens should be equal to token_num_per_seq * bs"
+        num_tokens = bs * token_num_per_seq
 
         tbo_split_seq_index, tbo_split_token_index = (
             two_batch_overlap.compute_split_indices_for_cuda_graph_replay(
                 forward_mode=forward_mode,
                 cuda_graph_num_tokens=num_tokens,
+                spec_info=spec_info,
```
fzyzcjy, line 135: maybe split this spec_info into two microbatches

python/sglang/srt/layers/attention/tbo_backend.py
```diff
@@ -208,14 +213,19 @@ def _init_forward_metadata_cuda_graph_split(
         forward_mode=forward_mode,
         # ignore
         encoder_lens=None,
-        spec_info=None,
+        spec_info=spec_info,
```
fzyzcjy: maybe split spec_info into two microbatches

python/sglang/srt/two_batch_overlap.py
```diff
@@ -26,17 +28,35 @@
 # -------------------------------- Compute Basic Info ---------------------------------------
 
 
+def get_token_num_per_seq(
+    forward_mode: ForwardMode,
+    spec_info: Optional[Union[EagleDraftInput, EagleVerifyInput]] = None,
+):
+    if forward_mode.is_target_verify():
+        assert spec_info is not None
```
hebiao064: do we need this assertion?

python/sglang/srt/two_batch_overlap.py
```diff
@@ -26,17 +28,35 @@
...
+        return spec_info.draft_token_num
+    elif forward_mode.is_decode():
+        return 1
+    elif forward_mode.is_idle():
+        return 0
+    else:
+        # For extend, we should not use `token_num_per_seq`.
```
hebiao064, line 42: should we raise exception here?

python/sglang/srt/two_batch_overlap.py
```diff
@@ -26,17 +28,35 @@
...
 ) -> Optional[int]:
-    if forward_mode.is_extend():
+    if forward_mode == ForwardMode.EXTEND:
         assert extend_lens is not None
         return _split_array_by_half_sum(extend_lens)
-    elif forward_mode.is_decode():
-        return num_tokens // 2
+    elif forward_mode.is_target_verify() or forward_mode.is_decode():
+        assert token_num_per_seq is not None
+        return (num_tokens // token_num_per_seq) // 2
```
hebiao064, line 58: can we pass the batch size from forward_batch to this function, hence we can avoid a unnecessary div opeartion?

python/sglang/srt/two_batch_overlap.py
Qiaolin-Yu: updated

python/sglang/srt/two_batch_overlap.py
Qiaolin-Yu: updated

python/sglang/srt/two_batch_overlap.py
Qiaolin-Yu: I think it should be fine 🤔. Just trying to make it a more unified API for decoding and target-verify.

ch-wan (approved):

PR #7231, opened 2025-06-16
[Eagle] Fix kernel call after updating speculative sampling kernels

merrymercy:
Following https://github.com/sgl-project/sglang/pull/7207, we change the kernel calls.

- Use a different coin for the final sampling in speculative sampling
- Remove some uncessary `to(torch.int32)`

```
Co-authored-by: Sehoon Kim <sehoon@x.ai>
```

### DIFF
--- a/python/sglang/srt/speculative/build_eagle_tree.py
+++ b/python/sglang/srt/speculative/build_eagle_tree.py
@@ -92,7 +92,7 @@ def build_tree_kernel_efficient(
     sgl_build_tree_kernel_efficient(
         parent_list,
         top_scores_index,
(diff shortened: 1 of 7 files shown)


python/sglang/srt/speculative/build_eagle_tree.py
```diff
@@ -92,7 +92,7 @@ def build_tree_kernel_efficient(
     sgl_build_tree_kernel_efficient(
         parent_list,
         top_scores_index,
-        seq_lens.to(torch.int32),
+        seq_lens,
```
merrymercy, line 95: @stslxg-nv please note that we changed the type of this argument

PR #7236, opened 2025-06-16
[PD Disaggregation] replace transfer with batch transfer for better performance

ssssnow:
## Motivation
This PR aims to significantly improve data transfer performance in the mooncake connector by replacing the original simple transfer with a batch transfer interface.
## Modifications
Replaced the simple transfer implementation in the mooncake connector with a batch transfer interface.
## Performance Improvement
In local tests (3 prefill nodes + 9 decode nodes, DeepSeek V3 model, 10K context length):
Simple transfer (old implementation in deep_ep branch): max transfer speed ~1.5 GB/s;
Batch transfer (this PR): transfer speed increased to 10-15 GB/s, about 8-12x faster;
Transfer time reduced from 650ms to 60ms.
As shown below, batch transfer greatly improves bandwidth utilization and overall performance:
image

More speed test could be found in https://github.com/kvcache-ai/Mooncake/pull/499


ssssnow:
@binarycrayon @hnyls2002 Hi guys, please help to review this PR, thanks~

ShangmingCai:
will review and test this today， thx

doujiang24:
@ssssnow how many separated(non-contiguous page) chunks in a single layer in the real tests? seems the improvement based on the separated chunk numbers.

ssssnow:
> @ssssnow how many separated(non-contiguous page) chunks in a single layer in the real tests? seems the improvement based on the separated chunk numbers.

When it comes to single big chunk, batch transfer and non-batch transfer have comparable speed. But in the case of DeepSeek MLA, single chunk may be not large enough, so single iteration (61 alyers, 10K inputs) may result in many small chunks.

Typical speed test on various chunk size and chunk number could be found here: https://github.com/kvcache-ai/Mooncake/pull/499

ShangmingCai (commented):
LGTM. The change is clean and straightforward, but I need to run some tests when https://github.com/kvcache-ai/Mooncake/pull/499 gets merged. Also, we need to wait for a new release v0.3.4 of mooncake, and we can revisit this PR at that time. I will ping someone to speed up the review process.

doujiang24:
@ssssnow I see the performance testing: https://github.com/kvcache-ai/Mooncake/pull/499, and the number of chunks could be large.

I'm just curious, how many chunks in a single layer leads to `Transfer time reduced from 650ms to 60ms`, in the sglang testing, since the chunks in [performance testing](https://github.com/kvcache-ai/Mooncake/pull/499) start from 5000.

ShangmingCai:
@ssssnow Can you share the commands to help reproduce? I think this is a nice work and could be helpful when page size is set to 1. But I am not sure whether it will significantly improve the e2e throughput when the page size is large. Let me run some tests to verify.

ssssnow:
> @ssssnow Can you share the commands to help reproduce? I think this is a nice work and could be helpful when page size is set to 1. But I am not sure whether it will significantly improve the e2e throughput when the page size is large. Let me run some tests to verify.

I tested this in SGLang using 3 Prefill nodes + 9 Decode nodes + DeepSeek V3 model + 10K input length. I add some profiling codes at `send_kv_cache`. If non-batch transfer is used, it only reach 1.5 GB/s (on old deep_ep branch). Batch transfer could deliver 12 GB/s speed.

But I believe this could be tested in Mooncake's repo. Here's the code I used: https://gist.github.com/ssssnow/37f17c9815cd26c3337ceae2729397cb

test commands are listed in the file, you could try to reproduce that.

ssssnow:
> > @ssssnow Can you share the commands to help reproduce? I think this is a nice work and could be helpful when page size is set to 1. But I am not sure whether it will significantly improve the e2e throughput when the page size is large. Let me run some tests to verify.
> 
> I tested this in SGLang using 3 Prefill nodes + 9 Decode nodes + DeepSeek V3 model + 10K input length. I add some profiling codes at `send_kv_cache`. If non-batch transfer is used, it only reach 1.5 GB/s (on old deep_ep branch). Batch transfer could deliver 12 GB/s speed.
> 
> But I believe this could be tested in Mooncake's repo. Here's the code I used: https://gist.github.com/ssssnow/37f17c9815cd26c3337ceae2729397cb
> 
> test commands are listed in the file, you could try to reproduce that.

Not sure if these are enough to help. If I miss anything, just let me know.

python/sglang/srt/disaggregation/mooncake/transfer_engine.py
```diff
@@ -90,5 +90,29 @@ def transfer_sync(
...
         return ret
 
+    def batch_transfer_sync(
+        self,
+        session_id: str,
+        buffers: List[int],
+        peer_buffer_addresses: List[int],
+        lengths: List[int],
+    ) -> int:
+        """Synchronously transfer data to the specified address."""
```
ShangmingCai, line 100: nit: maybe optimize this comment? like "Synchronously batch transfer data to the specified addresses."

ShangmingCai (approved):
LGTM

ShangmingCai:
I think it is ok to get this merged first. If we get error reports from the users, I can add an env var to let the users decide whether to use this new batch API if its stability or performance improvement turns out to be not guaranteed. Currently, I think the current changes are great.

PR #7281, opened 2025-06-17
Bug Fix: Inconsistent Return Values in MiniCPMV Class

cjhgo:
**Bug Fix: Inconsistent Return Values in `MiniCPMV` Class**

**Issue:**
The current implementation of the `MiniCPMV` class has a bug where method calls like `model.eval()` and `model.train()` return `self.minicpmv` instead of `self`. This inconsistency breaks the chain call pattern, as the return value is not the `MiniCPMV` instance itself, leading to unexpected behavior when chaining methods.

**Impact:**

- **Chain Call Inconsistency:** Methods that should return `self` for chain calls instead return the internal `minicpmv` instance, disrupting the expected behavior of method chaining.
- **Unexpected Behavior:** Users may encounter errors or unexpected results when trying to chain methods, as the return value is not the `MiniCPMV` instance.
- **Method Availability:** When `model.eval()` is called, the type of `model` changes, causing subsequent calls to methods like `model.load_weights` to fail, as the method is no longer available. This affects subsequent calls like `update_weights_from_tensor`.

**Scope:**
This bug is specific to the `python/sglang/srt/models/minicpmv.py` file, as it is a proxy class design.

**Proposed Fix:**

- **Quick Fix:** Add a `def eval()` method that calls `forward` to ensure the return value is `self`.
- **Comprehensive Fix:** Modify the `__getattr__` implementation to ensure that methods like `eval()` and `train()` return `self`, maintaining the integrity of the `MiniCPMV` class and providing a more intuitive API for users.

**Minimal Reproducible Code for the Bug:**

```python
from sglang.srt.configs.model_config import ModelConfig
from sglang.srt.model_executor.model_runner import ModelRunner
from sglang.srt.server_args import PortArgs, ServerArgs
import logging
logging.basicConfig(level=logging.INFO)

def load_model(server_args, port_args, tp_rank):
    model_config = ModelConfig.from_server_args(server_args)
    model_runner = ModelRunner(
        model_config=model_config,
        mem_fraction_static=server_args.mem_fraction_static,
        gpu_id=tp_rank,
        tp_rank=tp_rank,
        tp_size=server_args.tp_size,
        pp_rank=0,
        pp_size=1,
        nccl_port=port_args.nccl_port,
        server_args=server_args
    )
    return model_runner

if __name__ == "__main__":
    server_args = ServerArgs(
        model_path="openbmb/MiniCPM-v-2_6",
        trust_remote_code=True,
    )

    print("开始加载模型...")
    model_runner = load_model(server_args, port_args, tp_rank=0)
    logging.info(f"after loading,model_runner.model.load_weights={hasattr(model_runner.model, 'load_weights')}")
    assert hasattr(model_runner.model, "load_weights"), "model_runner.model.load_weights got lost"
    print("模型加载完成！")
```

```shell
INFO:sglang.srt.model_executor.model_runner:Capture cuda graph end. Time elapsed: 6.49 s. mem usage=1.86 GB. avail mem=13.55 GB.
INFO:root:after loading,model_runner.model.load_weights=False
[rank0]: Traceback (most recent call last):
[rank0]:   File "/home/linaro/workspace/sglang/runs/units/fixcpmv.py", line 32, in <module>
[rank0]:     assert hasattr(model_runner.model, "load_weights"), "model_runner.model.load_weights got lost"
[rank0]: AssertionError: model_runner.model.load_weights got lost
```

### DIFF
--- a/python/sglang/srt/models/minicpmv.py
+++ b/python/sglang/srt/models/minicpmv.py
@@ -823,15 +823,23 @@ class MiniCPMV:
 
     def __getattr__(self, name):
         if name == "minicpmv":
-            return None
-        return getattr(self.minicpmv, name)
+            raise AttributeError(
+                f"'{type(self).__name__}' object has no attribute '{name}'"
+            )
+        attr = getattr(self.minicpmv, name)
+
+        if callable(attr):
+            def wrapper(*args, **kwargs):
+                result = attr(*args, **kwargs)
+                if result is self.minicpmv:
+                    return self
+                return result
+            return wrapper
+        else:
+            return attr
 
     def __call__(self, *args, **kwargs):
         return self.minicpmv(*args, **kwargs)
-
-    def eval(self):
-        self.minicpmv.eval()
-        return self
 
     def load_weights(self, weights: Iterable[Tuple[str, torch.Tensor]]):
         stacked_params_mapping = [
--- a/python/sglang/srt/models/minicpmv.py
+++ b/python/sglang/srt/models/minicpmv.py
@@ -829,6 +829,10 @@ class MiniCPMV:
     def __call__(self, *args, **kwargs):
         return self.minicpmv(*args, **kwargs)
 
+    def eval(self):
+        self.minicpmv.eval()
+        return self
+
     def load_weights(self, weights: Iterable[Tuple[str, torch.Tensor]]):
         stacked_params_mapping = [
             # (param_name, shard_name, shard_id)


cjhgo:
@mickqian 
hi  could  you please help review this tiny pr~

mickqian:
Hi, can you provide some context about it, are you using it with some framework? the instance method `eval` seems unnecessary

cjhgo:
> Hi, can you provide some context about it, are you using it with some framework? the instance method `eval` seems unnecessary


Hi @mickqian, thanks for your feedback!

Let me provide some context and examples to illustrate why this fix is necessary.

I'm indeed using **sglang** within a framework — specifically, in a **RLHF** setup, where **sglang serves as the rollout backend**. In this scenario, I rely on both the `generate` and `update_weights_from_tensor` interfaces provided by sglang.

The `update_weights_from_tensor` method internally calls `self.model.load_weights`, which is crucial for weight synchronization during training :

```python
sglang/python/sglang/srt/model_executor/model_runner.py:676
676:     def update_weights_from_tensor(
677:         self,
678:         named_tensors: List[Tuple[str, Union[torch.Tensor, "LocalSerializedTensor"]]],
679:         load_format: Optional[str] = None,
680:     ):
681:         named_tensors = [
682:             (name, _unwrap_tensor(tensor, tp_rank=self.tp_rank))
683:             for name, tensor in named_tensors
684:         ]
685:         if load_format == "direct":
686:             _model_load_weights_direct(self.model, named_tensors)
687:         elif load_format is None:
688:             self.model.load_weights(named_tensors)
689:         else:
690:             raise NotImplementedError(f"Unknown load_format={load_format}")
691:         return True, "Success"
```

The issue arises because calling `model.eval()` — which is invoked during model loading by `model_loader` — was previously returning `self.minicpmv`, not `self`. As a result, after invoking `model.eval()`, the returned object was no longer of type `MiniCPMV`, and thus lost access to methods like `load_weights`.

This breaks the workflow when trying to perform weight updates after evaluation — especially problematic in RLHF, where frequent weight synchronization is needed between the policy and rollout models.

To further validate this behavior, there is an existing test in sglang:
```bash
test/srt/test_update_weights_from_tensor.py
```
This test demonstrates how weight updates are expected to work, and how they fail if the model instance isn't preserved correctly after calling methods like `eval()`.



In short, due to the inconsistent return type from `eval()`, the `minicpmv` model loaded into the SGL engine loses the ability to perform weight updates, which is essential in our use case.

Thanks again for the review, and let me know if you'd like any additional clarification or testing!

mickqian:
I see. If I understand it correctly, you use `self.minicpmv` to make sure it is an instance of `nn.Module`?

You can try using `MiniCPMV` directly, since it already forwards all calls to `self.minicpmv`, without manually adding methods to it.

cjhgo:
> I see. If I understand it correctly, you use `self.minicpmv` to make sure it is an instance of `nn.Module`?
> 
> You can try using `MiniCPMV` directly, since it already forwards all calls to `self.minicpmv`, without manually adding methods to it.





Hi @mickqian, thanks for the suggestion!

Actually, I'm not directly using the `minicpmv` module from `python/sglang/srt/models/minicpmv.py`. Instead, I'm interacting with it through `sgl.engine`, which is the interface I rely on in my workflow.

The issue stems from how `sgl.engine` is currently implemented, combined with the existing structure of the `MiniCPMV` class. 
As it stands, calling methods like `.eval()` on the model returned by the engine leads to inconsistent return types — sometimes returning `self.minicpmv` instead of the `MiniCPMV` wrapper itself. This breaks access to key methods like `load_weights`, which are essential for weight synchronization during training.

To clarify further, here’s a simplified code snippet that illustrates the typical usage pattern (similar to what's used in `verl`):

```python
import logging
logging.basicConfig(level=logging.INFO)
import torch
from typing import Any, Dict, List, Tuple, TypedDict


@dataclass
class ExpArgs:
    model_path: str = ""
    chat_template: str = ""
    trust_remote_code: bool = True

exp_cpmv = ExpArgs(
    model_path="openbmb/MiniCPM-v-2_6",
    chat_template="minicpmv",
)

exp_janus = ExpArgs(
    model_path="deepseek-ai/Janus-Pro-1B",
    chat_template="janus-pro",
)


def load_hf_model(model_id):
    if "janus" in model_id.lower():
        # pip install git+https://github.com/deepseek-ai/Janus.git
        from janus.models import MultiModalityCausalLM
        # from transformers import AutoModelForCausalLM
        load_cls = MultiModalityCausalLM
    else:
        from transformers import AutoModelForCausalLM
        load_cls = AutoModelForCausalLM
    model = load_cls.from_pretrained(
        model_id,
        torch_dtype=torch.bfloat16,
        trust_remote_code=True,
    ).to(torch.cuda.current_device())

    return model

def gen_prompt(args: ExpArgs):
    from sglang.srt.conversation import chat_templates
    conv = chat_templates[args.chat_template].copy()
    conv.append_message(conv.roles[0], "介绍一下北京")
    return conv.get_prompt()

def exp_sgl_wsync(args: ExpArgs):
    from sglang.srt.server_args import ServerArgs
    import sglang as sgl
    server_args = ServerArgs(
        model_path=args.model_path,
        chat_template=args.chat_template,
        trust_remote_code=args.trust_remote_code,
        mem_fraction_static=0.4,
    )
    vlm = sgl.Engine(server_args=server_args)
    prompt = gen_prompt(args)
    print(vlm.generate(prompt))
    vlm.release_memory_occupation()
    model = load_hf_model(args.model_path)
    vlm.update_weights_from_tensor(list(model.named_parameters()) ,load_format=None)
    print(vlm.generate(prompt))


if __name__ == "__main__":
    logging.info("start")
    args = exp_cpmv
    exp_sgl_wsync(args)
```

In this flow, when using `minicpmv`, the call to `update_weights_from_tensor` fails because `model.eval()` (which is often called internally during model loading) returns a raw `minicpmv` instance rather than the `MiniCPMV` wrapper. As a result, `load_weights` becomes inaccessible.

This issue does **not** occur with other models like `janus`

So unfortunately, just relying on `MiniCPMV` to forward all calls to `self.minicpmv` isn't sufficient unless we ensure consistent return types across all method calls — especially those inherited from `nn.Module` such as `.eval()` .

mickqian:
I see, thanks for your detailed explanation!

Would you mind changing it to something like this, to make it more generalized? 
```python
    def __getattr__(self, name):
        if name == "minicpmv":
            raise AttributeError(
                f"'{type(self).__name__}' object has no attribute '{name}'"
            )
        attr = getattr(self.minicpmv, name)

        if callable(attr):
            def wrapper(*args, **kwargs):
                result = attr(*args, **kwargs)
                if result is self.minicpmv:
                    return self
                return result
            return wrapper
        else:
            return attr
```

cjhgo:
> I see, thanks for your detailed explanation!
> 
> Would you mind changing it to something like this, to make it more generalized?
> 
> ```python
>     def __getattr__(self, name):
>         if name == "minicpmv":
>             raise AttributeError(
>                 f"'{type(self).__name__}' object has no attribute '{name}'"
>             )
>         attr = getattr(self.minicpmv, name)
> 
>         if callable(attr):
>             def wrapper(*args, **kwargs):
>                 result = attr(*args, **kwargs)
>                 if result is self.minicpmv:
>                     return self
>                 return result
>             return wrapper
>         else:
>             return attr
> ```


Yes, absolutely! 
That’s indeed a more general fix.
I’ve already updated the code accordingly.

mickqian (approved):

ispobock (approved):

ispobock:
Could you fix the lint test with `pre-commit run --all-files`?

PR #7294, opened 2025-06-18
Support dp attention in distributed weights syncing, and gpu remapping

ClawSeven:
This PR addresses two issues:

1. The first issue: It enables SgLang to support DP (Data Parallelism) and DP Attention when weights are transmitted via NCCL. This is a necessary feature for running Bailing LLM internally at Ant Group. I believe that the functionality of enabling DPA (Data Parallel Attention) with NCCL weight transfer would also be in demand by other companies.
    
2. The second issue is: When creating an NCCL communication group, failing to specify the device ID for the newly created weight communication group can cause an NCCL error indicating a duplicate GPU.

### DIFF
--- a/python/sglang/srt/utils.py
+++ b/python/sglang/srt/utils.py
@@ -1496,6 +1496,7 @@ def init_custom_process_group(
     store=None,
     group_name=None,
     pg_options=None,
+    device_id=-1,
 ):
     from torch.distributed.distributed_c10d import (
         Backend,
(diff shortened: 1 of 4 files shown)


ClawSeven:
In addition, these commits have already been deployed in Ant Group's internal RL scenarios as well as in Areal (https://github.com/inclusionAI/AReaL), and have been validated through large-scale RL training on thousands of GPUs.

python/sglang/srt/utils.py
```diff
@@ -1543,6 +1549,7 @@ def init_custom_process_group(
         group_name=group_name,
         **{pg_options_param_name: pg_options},
         timeout=timeout,
+        device_id=device_idx,
```
zhuzilin, line 1552: I've just tried this PR with slime. It seems that if I remove the device_id relevant logic, I can run with dp attention on, while it will stuck with the `device_id`. Could you give some more details about the following reason?
> When creating an NCCL communication group, failing to specify the device ID for the newly created weight communication group can cause an NCCL error indicating a duplicate GPU.

ClawSeven:
@zhaochenyang20  This issue has been resolved by @zhuzilin https://github.com/sgl-project/sglang/pull/11669, which features a nearly identical implementation.

PR #7303, opened 2025-06-18
[Refactor] Clean up radix cache related API

DarkSharpness:
## Motivation

This PR is the first part of #7194. Its main goal is to unify the interface of `BasePrefixCache` in preparation for the upcoming C++ radix tree implementation.

## Modifications

1. **Unify the interface of `BasePrefixCache`:**  
   - Removes unnecessary `if self.enable_hierarchical_cache` checks.
   - Aligns the return value of the `match_prefix` function for consistency.
2. **Fix typing errors and minor clean-ups:**  
   - Addresses several typing issues and refines code in the scheduler for improved maintainability.

## Checklist


python/sglang/srt/managers/schedule_policy.py
```diff
@@ -442,46 +439,40 @@ def add_req_state(r, insert_sort=False):
...
             req.sampling_params.max_new_tokens, CLIP_MAX_NEW_TOKENS_ESTIMATION
         )
-        input_tokens = (
-            -(-req.extend_input_len // self.tree_cache.page_size)
-            * self.tree_cache.page_size
-        )
+
+        # we should consider the possible host indices length when scheduling
+        min_extend_tokens = req.extend_input_len - req.host_indices_length
+        min_input_tokens = -(-min_extend_tokens // self.page_size) * self.page_size
```
xiezhq-hermann: I am not sure I followed here, the host_indices_length also accounts for memory consumption of the request, as loading the data from host also requires allocating token slots.

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -519,13 +519,14 @@ def __init__(
...
+        self.prefix_indices: torch.Tensor = []
         # Number of tokens to run prefill.
         self.extend_input_len = 0
         # The relative logprob_start_len in an extend batch
         self.extend_logprob_start_len = 0
-        self.last_node = None
-        self.last_node_global = None
+        self.last_node: Any = None
+        self.last_node_global: Any = None
+        self.host_indices_length = 0
```
xiezhq-hermann: it actually means indices of data only resides on CPU, the name of `host_indices_length` can be a bit confusing

python/sglang/srt/managers/scheduler.py
```diff
@@ -1518,19 +1517,8 @@ def get_new_batch_prefill(self) -> Optional[ScheduleBatch]:
...
-                    if (prefix_computed and not self.enable_hierarchical_cache)
-                    else self.tree_cache
-                ),
-                self.enable_hierarchical_cache,
-            )
-
-            res = adder.add_one_req(
-                req, self.chunked_req, self.enable_hierarchical_cache
-            )
+            req.init_next_round_input(self.tree_cache)
```
xiezhq-hermann, line 1520: are we getting rid of `prefix_computed` completely even with normal radix tree?

python/sglang/srt/managers/schedule_policy.py
DarkSharpness: My point is that, if we don't have the memory for hicache device indices, then this request will not be eligible for a prefill (prefill need more device indices than hicache loading from host to device). Maybe we should write this into comment?

python/sglang/srt/managers/schedule_batch.py
DarkSharpness: What about `host_indices_to_load` or `extra_host_indices`?

python/sglang/srt/managers/scheduler.py
DarkSharpness: I think whether or not we use `prefix_computed` depends on whether there will be modifications to the tree cache in `add_one_req`. Currently only `HiRadixCache` will modify the tree when it allocates device indices for host to device load back. This can be fixed within 1 line, but I would still maintain that this depends on the implementation of `PrefillAdder` and only when the tree is not mutated will `prefix_computed` make sense. Maybe we can pass the old `prefix_indices` and `last_node` to the tree when performing a `match_prefix` and let the tree determine itself whether the node is still valid.

xiezhq-hermann (approved):

PR #7336, opened 2025-06-19
[NVIDIA] Jetson Compatibility

johnnynunez:
Nvidia-smi not work for cuda-tegra, it works in next generation Thor.


johnnynunez:
@zhyncs i close this. Orin will receive SBSA soon, doing compatible with pynvml


CUDA 13 also brings enhancements to developer tools for Jetson, including support for libraries like NVML and nv-smi. These tools provide developers with better insights and control over GPU resources, enabling more efficient and effective development.

PR #7424, opened 2025-06-21
Revert "[BugFix]fix qwen25 invoke function call streaming responses with curly braces as the starting indicator"

hnyls2002:
Reverts sgl-project/sglang#7394

This broke the CI


hnyls2002:
This is fixed by #7425

PR #7431, opened 2025-06-22
[env] Organize sglang environ variables.

hnyls2002:
### Usage example

https://github.com/sgl-project/sglang/blob/ff3f055fb6d262f2cd34648a5723193bc503dfa5/python/sglang/environ.py#L175-L195


zhyncs:
@hnyls2002 any update?

fzyzcjy:
maybe we can firstly merge a minimal version, where we do not remove existing env vars, but users can add new env vars using this cleaner new mechanism, then at least future env vars will be cleaner

hnyls2002:
Moved to #10479

PR #7508, opened 2025-06-25
[OAI] patch origin request_id logic

whybeyoung:
## Modifications

1.  Add back the logic where the user specifies the request ID originally.
2. When handling is_single=False, the method def _normalize_rid(self, num) generates a list of rids based on the specified rid.


CatherineSue (approved):
LGTM

JustinTong0323 (approved):
LGTM.

PR #6549, opened 2025-05-23
[CPU] Bind threads and numa node for each TP rank

chunyuan-w:
## Motivation

This PR binds threads and numa node for each TP rank when device is CPU, which will improvement the performance on CPU.
For example, when running `--tp 6` on a machine with 6 sub-numa nodes, with this PR, rank 0 will be bound to core 0-39 on node 0, rank 1 to core 40-79 on node 1, etc.
```
python3 -m sglang.bench_one_batch --batch-size 1 --input 1024 --output 8 --model  deepseek-ai/DeepSeek-R1  --trust-remote-code --device cpu --tp 6
```

```sh
NUMA:                    
  NUMA node(s):          6
  NUMA node0 CPU(s):     0-39,240-279
  NUMA node1 CPU(s):     40-79,280-319
  NUMA node2 CPU(s):     80-119,320-359
  NUMA node3 CPU(s):     120-159,360-399
  NUMA node4 CPU(s):     160-199,400-439
  NUMA node5 CPU(s):     200-239,440-479
```

https://github.com/sgl-project/sglang/pull/7524 adds the c++ kernel that this PR needs to use.

## Modifications

- In `python/sglang/srt/model_executor/model_runner.py`, we add the threads and numa node binding according to the `tp_size`. When `SGLANG_CPU_OMP_THREADS_BIND` is not set, we only support tp_size to be smaller than or equal to the number of sub-numa nodes on the current machine and we will bind each tp rank to one sub-numa node. We only bind to physical cores while logical cores are excluded. If this is not satisfied, an error message will be thrown.

PR #7524, opened 2025-06-25
[CPU] add c++ kernel to bind CPU cores and memory node

chunyuan-w:
## Motivation
Add c++ kernel to bind CPU cores and memory node.
This kernel will be used in https://github.com/sgl-project/sglang/pull/6549.

## Modifications
- A new OP `init_cpu_threads_env` is added in `sgl-kernel/csrc/cpu/numa_utils.cpp` to bind threads and numa node. UT for this OP is added in `test/srt/cpu/test_binding.py`.
- `sgl-kernel/csrc/cpu/CMakeLists.txt` is updated to link the numa lib. `conda install -y libnuma numactl` is needed as a pre-requisite, which is already added in the Dockerfile for Xeon.
- We don't need users to build vLLM CPU wheel anymore and updated the message in `python/pyproject.toml`.
-


sgl-kernel/csrc/cpu/numa_utils.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+  for (int offset = 0; offset < omp_cpu_mask->size; offset += group_size) {
+    unsigned long group_mask = omp_cpu_mask->maskp[offset / group_size];
+    int i = 0;
+    while (group_mask) {
+      if (group_mask & 1) {
+        omp_cpu_ids.emplace_back(offset + i);
+      }
+      ++i;
+      group_mask >>= 1;
```
mickqian, line 27: nit: would this work?
```cpp
while (group_mask) {
    auto index = __builtin_ctzl(group_mask);
    omp_cpu_ids.emplace_back(offset + index);
    group_mask &= (group_mask - 1);
}
```

zhyncs (approved):

PR #7622, opened 2025-06-28
Support mini max

divijdawar:
## Motivation

Added support for minimiax model 
ref: https://github.com/MiniMax-AI/MiniMax-01


acelyc111:
@divijdawar Hi, is this PR ongoing? I found this branch dosen't work to load the MiniMax-text-01.

qscqesze:
I had a chance to try out your PR and noticed that the code is missing `EntryClass = MiniMaxText01ForCausalLM,` as well as a few minor issues that prevent it from running properly. Would you be open to making these adjustments?

PR #7667, opened 2025-06-30
Add fp4 quantize before all-gather for Flashinfer cutlass MoE DP (max throughput)

trevor-m:
## Motivation

The goal of this PR is to optimize communications for DP with FlashInfer Cutlass MoE.

## Modifications

Improvements include:
1. Add Allgatherv collective. This is a pynccl implementation of TRT-LLM's allgather which supports varying sizes per rank and a list of tensors as inputs
2. Add reducescatterv collective. This is a pynccl implementation of TRT-LLM's reducescatter which supports varying sizes per rank
3. For Flashinfer MoE with DP, use allgatherv to dispatch tokens. We also move the fp4 quantize before the allgather so the communication is smaller. Finally, we use reducescatterv to combine the results instead of all_reduce.

## Usage

Enabled automatically when applicable: `--enable-flashinfer-cutlass-moe`, `--enable-dp-attention`, and `dp_size == ep_size` must all be true.
Can be disabled with `--disable-flashinfer-cutlass-moe-fp4-allgather`.

```
python3 -m sglang.launch_server --model-path nvidia/DeepSeek-R1-0528-FP4 --trust-remote-code --quantization modelopt_fp4 --tp 8 --enable-flashinfer-cutlass-moe --ep-size 8 --dp 8 --enable-dp-attention
```

### Accuracy

```
Accuracy: 0.958
Invalid: 0.000
Latency: 23.484 s
Output throughput: 6228.185 token/s
```

### Benchmark

End to end speedup: 9.38%

```
python3 -m sglang.bench_serving --backend sglang --dataset-name random --num-prompt 1024 --random-input 1024 --random-output 1024 --random-range-ratio 1 --max-concurrency 1024
```

BEFORE
```
Max request concurrency:                 1024
Benchmark duration (s):                  75.54
Total generated tokens:                  1048576
Total generated tokens (retokenized):    1045749
Request throughput (req/s):              13.56
Input token throughput (tok/s):          13881.54
Output token throughput (tok/s):         13881.54
Total token throughput (tok/s):          27763.09
Concurrency:                             1021.04
Mean E2E Latency (ms):                   75318.71
Median E2E Latency (ms):                 75293.46
Mean TTFT (ms):                          11651.64
Median TTFT (ms):                        11573.21
P99 TTFT (ms):                           20818.61
Mean ITL (ms):                           62.24
Median ITL (ms):                         53.41
P95 ITL (ms):                            66.44
P99 ITL (ms):                            72.01
Max ITL (ms):                            18181.14
```

AFTER
```
Max request concurrency:                 1024
Benchmark duration (s):                  69.06
Total generated tokens:                  1048576
Total generated tokens (retokenized):    1045519
Request throughput (req/s):              14.83
Input token throughput (tok/s):          15183.59
Output token throughput (tok/s):         15183.59
Total token throughput (tok/s):          30367.17
Concurrency:                             1021.50
Mean E2E Latency (ms):                   68891.28
Median E2E Latency (ms):                 68904.07
Mean TTFT (ms):                          10103.61
Median TTFT (ms):                        10065.93
P99 TTFT (ms):                           17769.58
Mean ITL (ms):                           57.47
Median ITL (ms):                         49.91
P95 ITL (ms):                            61.59
P99 ITL (ms):                            68.18
Max ITL (ms):                            15189.56
```

## Checklist

### DIFF
--- a/python/sglang/srt/distributed/device_communicators/pynccl.py
+++ b/python/sglang/srt/distributed/device_communicators/pynccl.py
@@ -148,7 +148,11 @@ class PyNcclCommunicator:
         )
 
     def all_gather(
-        self, output_tensor: torch.Tensor, input_tensor: torch.Tensor, stream=None
+        self,
+        output_tensor: torch.Tensor,
+        input_tensor: torch.Tensor,
+        stream=None,
+        sizes: Optional[list[int]] = None,
     ):
         if self.disabled:
             return
@@ -161,14 +165,33 @@ class PyNcclCommunicator:
         )
         if stream is None:
             stream = self.stream
-        self.nccl.ncclAllGather(
-            buffer_type(input_tensor.data_ptr()),
-            buffer_type(output_tensor.data_ptr()),
-            input_tensor.numel(),
-            ncclDataTypeEnum.from_torch(input_tensor.dtype),
-            self.comm,
-            cudaStream_t(stream.cuda_stream),
-        )
+
+        if sizes is not None:
+            split_offset = 0
+
+            self.nccl.ncclGroupStart()
+            for root, split_size in enumerate(sizes):
+                dst_slice = output_tensor[split_offset : split_offset + split_size]
+                self.nccl.ncclBroadcast(
+                    buffer_type(input_tensor.data_ptr()),
+                    buffer_type(dst_slice.data_ptr()),
+                    dst_slice.numel(),
+                    ncclDataTypeEnum.from_torch(input_tensor.dtype),
+                    root,
+                    self.comm,
+                    cudaStream_t(stream.cuda_stream),
+                )
+                split_offset += split_size
+            self.nccl.ncclGroupEnd()
+        else:
+            self.nccl.ncclAllGather(
+                buffer_type(input_tensor.data_ptr()),
+                buffer_type(output_tensor.data_ptr()),
+                input_tensor.numel(),
+                ncclDataTypeEnum.from_torch(input_tensor.dtype),
+                self.comm,
+                cudaStream_t(stream.cuda_stream),
+            )
 
     def reduce_scatter(
         self,
@@ -176,6 +199,7 @@ class PyNcclCommunicator:
         input_tensor: torch.Tensor,
         op: ReduceOp = ReduceOp.SUM,
         stream=None,
+        sizes: Optional[list[int]] = None,
     ):
         if self.disabled:
             return
@@ -188,15 +212,35 @@ class PyNcclCommunicator:
         )
         if stream is None:
             stream = self.stream
-        self.nccl.ncclReduceScatter(
-            buffer_type(input_tensor.data_ptr()),
-            buffer_type(output_tensor.data_ptr()),
-            output_tensor.numel(),
-            ncclDataTypeEnum.from_torch(input_tensor.dtype),
-            ncclRedOpTypeEnum.from_torch(op),
-            self.comm,
-            cudaStream_t(stream.cuda_stream),
-        )
+
+        if sizes is not None:
+            split_offset = 0
+            self.nccl.ncclGroupStart()
+            for root, split_size in enumerate(sizes):
+                chunk = input_tensor[split_offset : split_offset + split_size, ...]
+
+                self.nccl.ncclReduce(
+                    buffer_type(chunk.data_ptr()),
+                    buffer_type(output_tensor.data_ptr()),
+                    chunk.numel(),
+                    ncclDataTypeEnum.from_torch(input_tensor.dtype),
+                    ncclRedOpTypeEnum.from_torch(op),
+                    root,
+                    self.comm,
+                    cudaStream_t(stream.cuda_stream),
+                )
+                split_offset += split_size
+            self.nccl.ncclGroupEnd()
+        else:
+            self.nccl.ncclReduceScatter(
+                buffer_type(input_tensor.data_ptr()),
+                buffer_type(output_tensor.data_ptr()),
+                output_tensor.numel(),
+                ncclDataTypeEnum.from_torch(input_tensor.dtype),
+                ncclRedOpTypeEnum.from_torch(op),
+                self.comm,
+                cudaStream_t(stream.cuda_stream),
+            )
 
     def send(self, tensor: torch.Tensor, dst: int, stream=None):
         if self.disabled:
@@ -265,6 +309,12 @@ class PyNcclCommunicator:
 
     def deregister_comm_window(self, window):
         return self.nccl.ncclCommWindowDeregister(self.comm, window)
+
+    def group_start(self):
+        self.nccl.ncclGroupStart()
+
+    def group_end(self):
+        self.nccl.ncclGroupEnd()
 
     @contextmanager
     def change_state(
--- a/python/sglang/srt/layers/moe/fused_moe_triton/layer.py
+++ b/python/sglang/srt/layers/moe/fused_moe_triton/layer.py
@@ -28,6 +28,7 @@ from sglang.srt.layers.quantization.base_config import (
     QuantizationConfig,
     QuantizeMethodBase,
 )
+from sglang.srt.layers.quantization.modelopt_quant import ModelOptNvFp4FusedMoEMethod
 from sglang.srt.layers.quantization.unquant import UnquantizedFusedMoEMethod
 from sglang.srt.managers.schedule_batch import global_server_args_dict
 from sglang.srt.model_loader.weight_utils import narrow_padded_param_and_loaded_weight
@@ -621,9 +622,7 @@ class FusedMoE(torch.nn.Module):
 
         if "ModelOpt" in self.quant_method.__class__.__name__:
             # Determine per-tensor weight scale patterns based on variant
-            is_fp4_variant = (
-                "ModelOptNvFp4FusedMoEMethod" in self.quant_method.__class__.__name__
-            )
+            is_fp4_variant = isinstance(self.quant_method, ModelOptNvFp4FusedMoEMethod)
 
             # FP4 uses "weight_scale_2" for per-tensor, FP8 uses "weight_scale" for per-tensor
             per_tensor_conditions = (
--- a/python/sglang/srt/layers/moe/topk.py
+++ b/python/sglang/srt/layers/moe/topk.py
@@ -327,6 +327,13 @@ class TopK(CustomOp):
                 expert_location_dispatch_info=expert_location_dispatch_info,
             )
 
+    def empty_topk_output(self, device: torch.device) -> TopKOutput:
+        topk = self.topk_config.top_k - self.topk_config.num_fused_shared_experts
+        topk_weights = torch.empty((0, topk), dtype=torch.float32, device=device)
+        topk_idx = torch.full((0, topk), -1, dtype=torch.int32, device=device)
+        router_logits = torch.empty((0, topk), dtype=torch.float32, device=device)
+        return StandardTopKOutput(topk_weights, topk_idx, router_logits)
+
 
 # ------------------------------- TopK implementation -------------------------------------
 
--- a/python/sglang/srt/layers/moe/utils.py
+++ b/python/sglang/srt/layers/moe/utils.py
@@ -7,6 +7,11 @@ from typing import TYPE_CHECKING, Optional
 
 from packaging import version as pkg_version
 
+from sglang.srt.distributed.parallel_state import get_moe_expert_parallel_world_size
+from sglang.srt.layers.dp_attention import (
+    get_attention_dp_size,
+    is_dp_attention_enabled,
+)
 from sglang.srt.utils import logger
 
 if TYPE_CHECKING:
@@ -99,6 +104,7 @@ DEEPEP_MODE: Optional[DeepEPMode] = None
 IS_TBO_ENABLED: Optional[bool] = None
 TBO_TOKEN_DISTRIBUTION_THRESHOLD: Optional[float] = None
 DEEPEP_CONFIG: Optional[str] = None
+DISABLE_FLASHINFER_CUTLASS_MOE_FP4_ALLGATHER: Optional[bool] = None
 
 
 def initialize_moe_config(server_args: ServerArgs):
@@ -108,6 +114,7 @@ def initialize_moe_config(server_args: ServerArgs):
     global DEEPEP_CONFIG
     global IS_TBO_ENABLED
     global TBO_TOKEN_DISTRIBUTION_THRESHOLD
+    global DISABLE_FLASHINFER_CUTLASS_MOE_FP4_ALLGATHER
 
     MOE_A2A_BACKEND = MoeA2ABackend(server_args.moe_a2a_backend)
     MOE_RUNNER_BACKEND = MoeRunnerBackend(server_args.moe_runner_backend)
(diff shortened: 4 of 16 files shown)


python/sglang/srt/layers/quantization/modelopt_quant.py
```diff
@@ -777,38 +779,64 @@ def apply(
...
             assert (
                 not apply_router_weight_on_input
             ), "apply_router_weight_on_input is not supported for Flashinfer"
             # TRTLLM Cutlass moe takes in activations in BF16/Half/nvfp4 precision
             # and fp4 quantized weights loaded from the checkpoint
+            output_dtype = x.dtype
+            # Swizzle after communication for dp.
+            x_row = x.shape[0]
+            x_col = x.shape[1]
+            x, x_sf = fp4_quantize(
```
Alcanderian: We maight only do quantize outside when dp_size > 1. Passing bf16 values into flashinfer_cutlass_fused_moe directly will archieve higher performance because the quantization process will be fused.

Alcanderian:
Does varying sizes per rank allgather compatible with cuda graph?

trevor-m:
> Does varying sizes per rank allgather compatible with cuda graph?

@Alcanderian Yes it is working with cuda graph. Although I am not super familiar with cuda graph in torch, is it possible that certain operations are excluded from the graph if they aren't compatible?

python/sglang/srt/layers/quantization/modelopt_quant.py
trevor-m: Thanks, that makes sense

trevor-m:
@Alcanderian It does appear this PR is breaking the cuda graphs. Before I see `Registering 6273 cuda graph addresses ` but with my changes it's now `Registering 357 cuda graph addresses`. That would explain why decode is getting slower, while prefill is improved with this PR.

Alcanderian:
Hi, please resolve conflicts and provide some performance report. Thanks!

trevor-m:
> Hi, please resolve conflicts and provide some performance report. Thanks!

Hi @Alcanderian I updated the PR description with performance results - is there any more benchmarks I should run? Thanks!

Alcanderian:
> > Hi, please resolve conflicts and provide some performance report. Thanks!
> 
> Hi @Alcanderian I updated the PR description with performance results - is there any more benchmarks I should run? Thanks!

Thanks! IMO we should disable it in decode stage accroding to the benchmark result

trevor-m:
@Alcanderian I looked into disabling this for decode stage, but it would complicate the deepseek code. The problem is that the mlp_mode of LayerCommunicator would need to change depending on the stage, but currently it is determined statically at initialization. Instead, I think that with disaggregated PD this flag could simply be enabled only on prefill nodes.

Additionally, I'm working on improving the perf of NCCL by using [NCCL symmetric memory](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/bufferreg.html#window-registration). If that works, I think both prefill and decode would see benefits.

trevor-m:
@Alcanderian Can you please review?
Turns out all gather/reduce scatter with varying size per rank is not valid with cuda graphs because the sizes tensor must be on CPU. I've made sure this feature can only be enabled with cuda graph off. For prefill and high throughput case, this feature provides a 10% improvement to input throughput.

Further boost with nccl symmetric memory will come in a later PR (based on https://github.com/sgl-project/sglang/pull/8238).

Without this PR:
```
python3 -m sglang.launch_server --model-path nvidia/DeepSeek-R1-0528-FP4 --trust-remote-code --quantization modelopt_fp4 --tp 8 --enable-flashinfer-moe --disable-shared-experts-fusion --enable-ep-moe --ep-size 8 --dp 8 --enable-dp-attention --disable-cuda-graph --chunked-prefill-size 16384 --mem-fraction-static 0.85 --max-running-requests 4096 --stream-interval 5 --enable-dp-lm-head --attention-backend cutlass_mla
python3 -m sglang.bench_one_batch_server --model-path nvidia/DeepSeek-R1-0528-FP4 --base-url http://127.0.0.1:30000 --batch-size 32 --input-len 8196 --output-len 1024

#Input tokens: 262272
#Output tokens: 32768
batch size: 32
input_len: 8196
output_len: 1024
latency: 86.43 s
ttft: 13.00 s
last generation throughput: 56.40 tok/s
input throughput: 20169.52 tok/s
output throughput: 446.30 tok/s
```

With this PR:
```
python3 -m sglang.launch_server --model-path nvidia/DeepSeek-R1-0528-FP4 --trust-remote-code --quantization modelopt_fp4 --tp 8 --enable-flashinfer-moe --disable-shared-experts-fusion --enable-ep-moe --ep-size 8 --dp 8 --enable-dp-attention --disable-cuda-graph --chunked-prefill-size 16384 --mem-fraction-static 0.85 --max-running-requests 4096 --stream-interval 5 --enable-dp-lm-head --attention-backend cutlass_mla --enable-flashinfer-fp4-allgather
python3 -m sglang.bench_one_batch_server --model-path nvidia/DeepSeek-R1-0528-FP4 --base-url http://127.0.0.1:30000 --batch-size 32 --input-len 8196 --output-len 1024

#Input tokens: 262272
#Output tokens: 32768
batch size: 32
input_len: 8196
output_len: 1024
latency: 85.19 s
ttft: 11.70 s
last generation throughput: 55.95 tok/s
input throughput: 22411.31 tok/s
output throughput: 445.88 tok/s
```

trevor-m:
This PR has been updated. Issues with cuda graphs have been resolved and it's showing a 9.38% end to end speedup for max throughput scenario. Latest accuracy and benchmarks are in the PR description.

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -447,18 +448,26 @@ def forward(
             if (
                 self.alt_stream is not None
                 and self.num_fused_shared_experts == 0
+                and hidden_states.shape[0] > 0
```
nvcastet, line 441: Have you noticed the number of token be zero here?

python/sglang/srt/server_args.py
```diff
@@ -644,7 +645,13 @@ def print_deprecated_warning(message: str):
             self.disable_cuda_graph = True
             logger.warning("Cuda graph is disabled for prefill server")
 
-        # Propagate env vars
+        if self.enable_flashinfer_fp4_allgather:
+            assert (
+                self.enable_flashinfer_cutlass_moe
+                and self.ep_size == self.dp_size
+                and self.enable_dp_attention
+            ), "FP4 all-gather requires --enable-cutlass-flashinfer-moe, --enable-dp-attention, and dp_size == ep_size"
```
nvcastet: Any reason why we would not enable it automatically when those requirements are fulfilled instead of adding another flag?

python/sglang/srt/models/deepseek_v2.py
nvcastet: Is it in the case of an "idle" worker? What happens in that case before this PR?

python/sglang/srt/server_args.py
trevor-m: This feature is slower for low latency use case. Since flashinfer cutlass moe is the path for high throughput, I think it would make sense to enable it automatically - we should just document the use case for the different MoEs.

python/sglang/srt/models/deepseek_v2.py
trevor-m: Yes, it can happen often, for example during the server warmup.

Before this PR, the layercommunicator gathers the tokens from all gpus so the MoE runs on the global tokens. With this change https://github.com/sgl-project/sglang/pull/7667/files#diff-fd304e0c770e5a99efec686fa564d53303b587b790b9f2c518c52c2601e6abd1R111-R115 the gather is bypassed so the MoE only gets the local tokens which can now be 0.

nvcastet: That makes sense.

python/sglang/srt/server_args.py
trevor-m: Enabling automatically now.

python/sglang/srt/server_args.py
```diff
@@ -1432,18 +1437,23 @@ def add_cli_args(parser: argparse.ArgumentParser):
...
-            help="Enable FlashInfer TRTLLM MoE backend on Blackwell. Supports BlockScale FP8 MoE-EP",
+            help="Enable FlashInfer TRTLLM MoE backend on Blackwell. Used for low-latency scenarios. Supports BlockScale FP8 MoE-EP",
         )
         parser.add_argument(
             "--enable-flashinfer-allreduce-fusion",
             action="store_true",
             help="Enable FlashInfer allreduce fusion for Add_RMSNorm.",
         )
+        parser.add_argument(
+            "--enable-flashinfer-fp4-allgather",
```
nvcastet: Does this flag work outside of `--enable-cutlass-flashinfer-moe` and `--enable-dp-attention` and `dp_size == ep_size` condition?
If no, the flag won't be useful?

trevor-m: You're right, this flag only works with that combination. I guess instead of using a server arg we could have a helper function to check those flags.

On the only hand, having the server arg allows it to be disabled.

nvcastet: How do you disable it, `--enable-flashinfer-fp4-allgather=false` works?
Never tried :)

trevor-m: Nevermind, that doesnt work

nvcastet: you could make it an explicit bool arg, or switch it to `--disable-flashinfer-fp4-allgather`

python/sglang/srt/layers/moe/fused_moe_triton/layer.py
```diff
@@ -783,7 +784,12 @@ def weight_loader_fused(
                 f"Unsupported weight_name {weight_name} for FusedMoE weight_loader_fused. Nothing is loaded."
             )
 
-    def forward(self, hidden_states: torch.Tensor, topk_output: StandardTopKOutput):
+    def forward(
+        self,
+        hidden_states: torch.Tensor,
+        topk_output: StandardTopKOutput,
+        forward_batch: Optional[ForwardBatch] = None,
```
merrymercy: What are the needed fields? Passing the whole `forward_batch` makes this function kind of opaque.
We want the function to be more explicit about the argument.

merrymercy (changes requested):

python/sglang/srt/layers/moe/fused_moe_triton/layer.py
trevor-m: Hi @merrymercy thank you for reviewing. In modelopt_quant.py, we use `forward_batch.dp_padding_mode.is_max_len()`, `forward_batch.input_ids.shape[0]`, `forward_batch.gathered_buffer`, `forward_batch.global_num_tokens_cpu`.

Other backends like deepep which also integrate communication pass the whole forward_batch to MOE also: https://github.com/sgl-project/sglang/blob/cbbd685a46cd9345c0a87f243292c3ea59e5db32/python/sglang/srt/models/deepseek_v2.py#L611

python/sglang/srt/server_args.py
trevor-m: Added a server arg to disable. Otherwise it is automatically used when the required combination is set.

kushanam (approved):

kushanam:
@ch-wan could you help review/merge this PR plz?

python/sglang/srt/layers/quantization/modelopt_quant.py
```diff
@@ -1191,6 +1227,21 @@ def apply(
             )[0]
             if routed_scaling_factor is not None:
                 output *= routed_scaling_factor
+
+            if should_use_flashinfer_cutlass_moe_fp4_allgather():
+                output, global_hidden_states = (
+                    forward_batch.gathered_buffer[: forward_batch.input_ids.shape[0]],
```
ch-wan: The code logic is getting increasingly messy. It seems that `forward_batch` is only need for dp communication. We need to refactor `dp_attention.py` so that any part of the code can handle dp communication without using `forward_batch`.  Let me try this later.

trevor-m: Thanks @ch-wan - anything I can do for this PR to help?

ch-wan: I will work on this: https://github.com/sgl-project/sglang/pull/9136

python/sglang/srt/layers/quantization/modelopt_quant.py
```diff
@@ -1151,6 +1155,7 @@ def apply(
         ep_size: Optional[int] = None,
         tp_rank: Optional[int] = None,
         tp_size: Optional[int] = None,
+        forward_batch: Optional[ForwardBatch] = None,
```
merrymercy: we should explicitly list all the required arguments instead of passing a big `forward_batch`.
Passing a big `forward_batch` makes the input of this function very opaque.

merrymercy (changes requested):

python/sglang/srt/layers/quantization/modelopt_quant.py
trevor-m: Thanks, done.

ch-wan:
@trevor-m DP attn refactor is almost ready. I am doing extra local tests. Could you update this PR by using the util function developed in #9136? This should make the code much cleaner. Thanks.

**UPD:** I've just noticed that you did some refactor in this PR :) How about adding `get_dp_global_buffer_len` to DP utils? This can save args in your code.

trevor-m:
> @trevor-m DP attn refactor is almost ready. I am doing extra local tests. Could you update this PR by using the util function developed in #9136? This should make the code much cleaner. Thanks.
> 
> **UPD:** I've just noticed that you did some refactor in this PR :) How about adding `get_dp_global_buffer_len` to DP utils? This can save args in your code.



> `get_dp_global_buffer_len`

Thanks @ch-wan I added a helper so I don't need to pass `forward_batch.global_num_tokens_cpu` anymore. Please take a look

python/sglang/srt/layers/moe/topk.py
```diff
@@ -277,6 +277,12 @@ def forward_npu(
                 expert_location_dispatch_info=expert_location_dispatch_info,
             )
 
+    def empty_topk_output(self, device: torch.device) -> TopKOutput:
+        topk = self.top_k - self.num_fused_shared_experts
+        topk_weights = torch.empty((0, topk), dtype=torch.float32, device=device)
+        topk_idx = torch.full((0, topk), -1, dtype=torch.int32, device=device)
+        return StandardTopKOutput(topk_weights, topk_idx, None)
```
ch-wan: How about creating another empty tensor for the last element (`router_logits`) to make the engine more robust? Some quantization methods need it. Note that the shape of logits is (0, num_experts)

python/sglang/srt/layers/moe/utils.py
```diff
@@ -17,6 +22,18 @@ def should_use_flashinfer_trtllm_moe():
     return result
 
 
+def should_use_flashinfer_cutlass_moe_fp4_allgather():
```
ch-wan: use lru_cache

ch-wan (commented):
LGTM. I only have some minor comments. Could you fix conflicts? Thanks.

python/sglang/srt/layers/moe/topk.py
trevor-m: Thanks, done!

python/sglang/srt/layers/moe/utils.py
trevor-m: Thanks, done

ch-wan (approved):

python/sglang/srt/distributed/device_communicators/pynccl.py
```diff
@@ -161,21 +165,41 @@ def all_gather(
...
-            cudaStream_t(stream.cuda_stream),
-        )
+
+        if sizes is not None:
+            split_offset = 0
+
+            self.nccl.ncclGroupStart()
+            for root, split_size in enumerate(sizes):
+                dst_slice = output_tensor[split_offset : split_offset + split_size]
+                self.nccl.ncclBroadcast(
```
merrymercy, line 175: Is this broadcast or allgather?

trevor-m: It’s equivalent to all gather. Each rank does a broadcast but we group them to avoid overheads. This is done to allow each rank to have a different size

PR #7678, opened 2025-07-01
feat: add Hunyuan reasoning parser support

aldopareja:
Motivation

Reasoning parser support for the Hunyuan-A13B model to SGLang. The Hunyuan model uses a similar
  reasoning format to DeepSeek-R1 and Qwen3 with <think>...</think> tags, but additionally wraps its final
  responses in <answer>...</answer> tags. This PR enables SGLang to properly parse and separate reasoning
  content from final answers for Hunyuan models, allowing users to leverage the reasoning capabilities of these
  models through the standard SGLang API.

  Modifications

  This PR implements a minimal reasoning parser for Hunyuan by:

  - Adding a HunyuanDetector class to python/sglang/srt/reasoning_parser.py that extends
  BaseReasoningFormatDetector
  - The detector handles the dual-tag format: <think>reasoning</think><answer>response</answer>
  - Strips both <think> and <answer> tags from the output while preserving any <tool_calls> content that may
  appear within answer tags
  - Implements both streaming and non-streaming parsing modes with proper handling of partial tags
  - Registering the detector in ReasoningParser.DetectorMap so it's available via --reasoning-parser hunyuan
  - Adding comprehensive tests to test/srt/test_reasoning_parser.py including:
    - Basic detector tests in TestHunyuanDetector class (4 tests)
    - Integration with ReasoningParser in existing test methods
    - Complex scenarios in TestIntegrationScenarios for tool calls and streaming

  The implementation follows the exact patterns established by existing reasoning parsers (DeepSeek-R1, Qwen3,
  Kimi) to maintain consistency.

### DIFF
--- a/python/sglang/srt/reasoning_parser.py
+++ b/python/sglang/srt/reasoning_parser.py
@@ -175,6 +175,109 @@ class KimiDetector(BaseReasoningFormatDetector):
         )
 
 
+class HunyuanDetector(BaseReasoningFormatDetector):
+    """
+    Detector for Hunyuan model.
+
+    Handles format:
+      <think>reasoning</think><answer>response</answer>
+
+    Extracts reasoning from <think> tags and strips both <think> and <answer> tags,
+    leaving clean content for further processing.
+
+    Args:
+        stream_reasoning (bool): If False, accumulates reasoning content until the end tag.
+            If True, streams reasoning content as it arrives.
+    """
+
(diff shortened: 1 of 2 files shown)


hnyls2002:
Thanks @aldopareja! Closing as superseded - a Hunyuan reasoning parser is now native (`reasoning_parser.py`, registered as `hunyuan`).

PR #7723, opened 2025-07-02
[Bug] add flashinfer bool check for fusedmoe in Qwen moe models

yilian49:
## Motivation

The current Qwen MoE models (`qwen2_moe.py` and `qwen3_moe.py`) do not pass the `enable_flashinfer_moe` flag, causing `FuseMoE` to default to `enable_flashinfer_moe=False`.

## Modifications

Added checks to ensure the `enable_flashinfer_moe` flag is correctly passed to the `FuseMoE` initializer.

### DIFF
--- a/python/sglang/srt/models/qwen3_moe.py
+++ b/python/sglang/srt/models/qwen3_moe.py
@@ -100,7 +100,17 @@ class Qwen3MoeSparseMoeBlock(nn.Module):
                 f"the number of experts {config.num_experts}."
             )
 
-        self.experts = get_moe_impl_class()(
+        extra_args = {}
+
+        expert_type = get_moe_impl_class()
+        if expert_type is FusedMoE:
+            extra_args = {
+                "enable_flashinfer_moe": global_server_args_dict[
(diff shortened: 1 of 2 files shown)


python/sglang/srt/models/qwen3_moe.py
```diff
@@ -100,7 +100,17 @@ def __init__(
...
             )
 
-        self.experts = get_moe_impl_class()(
+        extra_args = {}
+
+        expert_type = get_moe_impl_class()
+        if expert_type is FusedMoE:
+            extra_args = {
+                "enable_flashinfer_moe": global_server_args_dict[
+                    "enable_flashinfer_moe"
```
Alcanderian: IMO we should pass extra_args like this
Ref: https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/models/deepseek_v2.py#L313

yilian49: Make sense. I have updated accordingly.

Alcanderian (approved):

PR #7818, opened 2025-07-07
[CPU]convert topk_weights to fp32 for INT8 and FP8 paths (for llama4) and fix LmHead weight pack

chunyuan-w:
## Motivation

1. Convert `topk_weights` to fp32 for INT8 and FP8 paths since the CPU kernel requires it to be fp32.
Previously we've fixed this issue in BF16 path for `llama4`. The same fix is needed for INT8 and FP8.
https://github.com/sgl-project/sglang/blob/3646f6bb3e42ef31b29e4bae3244a14333a2ba9b/python/sglang/srt/layers/moe/fused_moe_triton/layer.py#L320-L322

2. For the weight pack check in `ParallelLMHead`, we previously checked that `self.quant_config` should be None to ensure it's not quantized. But `self.quant_config` could be `Fp8Config` for FP8 model or `W8A8Int8Config` for INT8 model while the weight of `ParallelLMHead` is still `torch.bfloat16`. We updated the code to explicitly check the weight dtype to device whether weight pack is supported or not on CPU.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ Learn more in the release blogs: [v0.2 blog](https://lmsys.org/blog/2024-07-25-s
 [Development Roadmap (2025 H1)](https://github.com/sgl-project/sglang/issues/4042)
 
 ## Adoption and Sponsorship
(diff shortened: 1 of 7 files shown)


mingfeima:
@chunyuan-w open a jira tracking this issue internally and add bf16 / f16 support for topk_weights later on.

mingfeima (changes requested):
I'd prefer to do this inside C++ kernels and leave python level unchanged.

Say we don't want change the kernel level impls for now.

we can do something like:
```
// TODO: add bfloat16 and float16 support. Llama4MoE blahblah...
auto topk_weights_ = topk_weights.to(at::kFloat);
```

chunyuan-w:
> @chunyuan-w open a jira tracking this issue internally and add bf16 / f16 support for topk_weights later on.

Internal JIRA has been created to track this.

chunyuan-w:
> I'd prefer to do this inside C++ kernels and leave python level unchanged.
> 
> Say we don't want change the kernel level impls for now.
> 
> we can do something like:
> 
> ```
> // TODO: add bfloat16 and float16 support. Llama4MoE blahblah...
> auto topk_weights_ = topk_weights.to(at::kFloat);
> ```

Fixed in https://github.com/sgl-project/sglang/pull/7818/commits/4a9087f20063a25e7be5f029f184ef4147a20c35

mingfeima (approved):

PR #7820, opened 2025-07-07
[Quantization][w8a8_int8] Fix weight loading issue for w8a8_int8 path with "ignore" layer list in quantization config

jianan-gu:
This PR is referring to the way of [compressed_tensors](https://github.com/CatherineSue/sglang/blob/a41fcad7d9c53b3045bc8e8e9db3e6ad519b4f9d/python/sglang/srt/layers/quantization/compressed_tensors/compressed_tensors.py#L132) loading format from the quantization config, adding `ignore`  list for `w8a8_int8` path, which is necessary since models may not be fully quantized. 
For example, W8A8 models will contain the `ignore` list: https://huggingface.co/nytopop/Qwen3-30B-A3B.w8a8/blob/main/config.json#L62.


python/sglang/srt/layers/quantization/w8a8_int8.py
```diff
@@ -237,7 +241,7 @@ def get_quant_method(
         layer: torch.nn.Module,
         prefix: str,
     ) -> Optional["QuantizeMethodBase"]:
-        from sglang.srt.layers.linear import LinearBase
+        from sglang.srt.layers.linear import LinearBase, UnquantizedLinearMethod
         from sglang.srt.layers.moe.fused_moe_triton import FusedMoE
 
         if _is_npu:
```
AniZpZ, line 247: i think npu needs unquantized layer ignore as well

python/sglang/srt/layers/quantization/w8a8_int8.py
```diff
@@ -263,6 +267,10 @@ def get_quant_method(
                 return NPU_W8A8MoEMethod(self)
```
AniZpZ, line 265: i think npu needs unquantized layer ignore as well

jianan-gu: Hi, thanks for the review.
Refined to make this ignoring (fallback unquantized layer) a common process here.

AniZpZ (approved):
no more concerns from me

jianan-gu:
Checked the CI failures are not related to this PR.

jianan-gu:
Hi, @zhyncs 
Could you help merge if no concern. Thanks!
Checked CI failures are not related to this PR.

ch-wan (approved):

PR #7837, opened 2025-07-08
[CPU] fix load_qkv_weight loading when odd TP size

jianan-gu:


jianan-gu:
close this pr as covered in https://github.com/sgl-project/sglang/pull/7838

PR #7838, opened 2025-07-08
[CPU][Qwen3 MoE] Enable fused_topk CPU fusion and enhance FP8 TP padding

jianan-gu:
This PR contains the following fixes when run with fp8 MoE models on CPUs like https://huggingface.co/Qwen/Qwen3-30B-A3B-FP8  

1. Enable fused_topk CPU fusion
   ->  `fused_topk_cpu ` could cover the original `fused_topk `func on CPU path
2. fix `load_qkv_weight `loading when odd TP size
   -> When running with TP size that is not dividable, refering to other weight loader, adding `narrow_padded_param_and_loaded_weight `for `load_qkv_weight ` as well.


mingfeima (approved):

python/sglang/srt/layers/parameter.py
```diff
@@ -187,10 +187,25 @@ def load_qkv_weight(
         param_data = self.data
         shard_id = tp_rank if shard_id == "q" else tp_rank // num_heads
         param_data = param_data.narrow(self.output_dim, shard_offset, shard_size)
-        if not use_presharded_weights:
-            loaded_weight = loaded_weight.narrow(
-                self.output_dim, shard_id * shard_size, shard_size
+        from sglang.srt.model_loader.weight_utils import (
```
zhyncs: move to top

zhyncs (changes requested):

python/sglang/srt/layers/parameter.py
jianan-gu: Hi, @zhyncs, thanks for the comments.

Here we tried to import this from the top of this file, while we found that it will cause "circular import" (when benchmarking using bench_one_batch) if we import anything from `sglang.srt.model_loader.weight_utils` globally.

The "circular import" logic is:

bench_one_batch.py -> **model_config.py** -> quantization  -> linear.py -> parameter.py -> **weight_utils.py** -> **model_config.py**


Thus, we cannot introduce the import of weight_utils.py, but have to keep this import locally at runtime instead.

Raw back trace:
CMD: python3 -m sglang.bench_one_batch  --input 32 --output 32 --batch-size 1 --model {MODEL_PATH}  
```
  File "/sglang/python/sglang/bench_one_batch.py", line 59, in <module>
    from sglang.srt.configs.model_config import ModelConfig
  File "/sglang/python/sglang/srt/configs/model_config.py", line 30, in <module>
    from sglang.srt.layers.quantization import QUANTIZATION_METHODS
  File "/sglang/python/sglang/srt/layers/quantization/__init__.py", line 52, in <module>
    from sglang.srt.layers.linear import LinearBase, UnquantizedLinearMethod
  File "/sglang/python/sglang/srt/layers/linear.py", line 21, in <module>
    from sglang.srt.layers.parameter import (
  File "/sglang/python/sglang/srt/layers/parameter.py", line 28, in <module>
    from sglang.srt.model_loader.weight_utils import (
  File "/sglang/python/sglang/srt/model_loader/__init__.py", line 7, in <module>
    from sglang.srt.configs.model_config import ModelConfig
ImportError: cannot import name 'ModelConfig' from partially initialized module 'sglang.srt.configs.model_config' (most likely due to a circular import) (/sglang/python/sglang/srt/configs/model_config.py)
```

zhyncs: ok can we put this after `if _is_cpu`

jianan-gu: Sure, have made the import after `if _is_cpu`

zhyncs (approved):

PR #6543, opened 2025-05-23
Refactor vlm embedding routine to use precomputed feature

JustinTong0323:
## Motivation
- This PR is a Supplement for https://github.com/sgl-project/sglang/pull/6136

- Refactor vlm embedding routine to use precomputed feature

- Add tests for passing pixel values and precomputed features to VLMs(Split from test_vlm_acc)

## Checklist

- [x] Format your code according to the [Code Formatting with Pre-Commit](https://docs.sglang.ai/references/contribution_guide.html#code-formatting-with-pre-commit).
- [ ] Add unit tests as outlined in the [Running Unit Tests](https://docs.sglang.ai/references/contribution_guide.html#running-unit-tests-adding-to-ci).
- [ ] Update documentation / docstrings / example tutorials as needed, according to [Writing Documentation](https://docs.sglang.ai/references/contribution_guide.html#writing-documentation-running-docs-ci).
- [ ] Provide throughput / latency benchmark results and accuracy evaluation results as needed, according to [Benchmark and Profiling](https://docs.sglang.ai/references/benchmark_and_profiling.html) and [Accuracy Results](https://docs.sglang.ai/references/accuracy_evaluation.html).
- [ ] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co-author when merging the PR.
- [ ] Please feel free to join our Slack channel at https://slack.sglang.ai to discuss your PR.

Issue #6483, opened 2025-05-21
[Feature] Support precomputed multimodal features for remaining VLM


lifuhuang:
### Checklist

- [x] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 2. Please use English, otherwise it will be closed.

### Motivation

Currently SGL has supported precomputed multimodal features for Gemma3 and Qwen VL by @ysulsky, we should support all remaining VLM and add tests (Parent issue: https://github.com/sgl-project/sglang/pull/6136)

- [x] KimiVL (https://github.com/sgl-project/sglang/pull/6410)
- [ ] internvl
- [ ] mllama4
- [ ] pixtral
- [ ] deepseek_vl_v2
- [ ] minicpm
- [ ] janus_pro

JustinTong0323:
- #6481 
- See #6488 for refactored and more comprehensive tests.

JustinTong0323:
@ysulsky Hi Yury, may you have time to look at related PRs under this issue? Thanks a lot!

ysulsky:
Of course, happy to!

1fanwang:
Would love the pickup `internvl` support varied input formats if no one is already working on it.

b8zhong:
I don't think anyone is, feel free to work on it @1fanwang

1fanwang:
Pixtral is covered by https://github.com/sgl-project/sglang/pull/24412. H100 end-to-end validation passed all three input formats. cc @lifuhuang

1fanwang:
Opened the remaining model PRs: Phi-4 in https://github.com/sgl-project/sglang/pull/24469, DeepSeek-VL2 in https://github.com/sgl-project/sglang/pull/24474, and Llama 4 in https://github.com/sgl-project/sglang/pull/24489. The shared key-rename work is in https://github.com/sgl-project/sglang/pull/24486.

PR #7921, opened 2025-07-10
[feat]mmprocessor: support minicpm prep pixel values input

cjhgo:
## Motivation
follow the great work of 
https://github.com/sgl-project/sglang/issues/6483
https://github.com/sgl-project/sglang/pull/6543
in this pr, i try to adjust minicpm mmprocessor to support preprocessed pixels values input.

## Checklist

### DIFF
--- a/test/srt/test_vlm_input_format.py
+++ b/test/srt/test_vlm_input_format.py
@@ -215,6 +215,26 @@ class TestKimiVLImageUnderstandsImage(
             image_grid_hws=processor_output["image_grid_hws"],
         )
 
+class TestMiniCPMUnderstandsImage(VLMInputTestBase, unittest.IsolatedAsyncioTestCase):
+    #python test_vlm_input_format.py TestMiniCPMUnderstandsImage
+    model_path = "openbmb/MiniCPM-v-2_6"
+    chat_template = "minicpmv"
+
+    @classmethod
+    def _init_visual(cls):
+        pass
+
+    @unittest.skip("skip")
+    async def test_understands_precomputed_features(self):
+        pass
+
+    def _pixel_values_image_data(self, processor_output):
+        return dict(
+            modality="IMAGE",
+            pixel_values=processor_output["pixel_values"],
+            tgt_size=processor_output["tgt_sizes"]
+        )
+
 
 if __name__ == "__main__":
     unittest.main()
--- a/test/srt/test_vlm_input_format.py
+++ b/test/srt/test_vlm_input_format.py
@@ -229,11 +229,62 @@ class TestMiniCPMUnderstandsImage(VLMInputTestBase, unittest.IsolatedAsyncioTest
         pass
 
     def _pixel_values_image_data(self, processor_output):
-        return dict(
-            modality="IMAGE",
-            pixel_values=processor_output["pixel_values"],
-            tgt_size=processor_output["tgt_sizes"]
+        image_slice_patch_size = len(processor_output.pixel_values[0])
+        pre_dicts = []
+        for i in range(image_slice_patch_size):
+            pre_dicts.append(dict(
+                modality="IMAGE",
+                pixel_values=processor_output.pixel_values[0][i],
+                tgt_size=processor_output.tgt_sizes[0][i],
+            ))
+        return pre_dicts
+
+    async def test_understands_pixel_values(self):
+        req = self.get_completion_request()
+        processor_output = self.get_processor_output(req=req)
+        output = await self.engine.async_generate(
+            input_ids=processor_output["input_ids"][0].detach().cpu().tolist(),
+            image_data=self._pixel_values_image_data(processor_output),
+            sampling_params=dict(temperature=0.0),
         )
+        self.verify_response(output)
+
+    async def test_understands_pixel_values_with_imgs(self):
+        def local_get_completion_request() -> ChatCompletionRequest:
+            json_structure = {
+            "model": self.model_path,
+            "messages": [
+                {
+                    "role": "user",
+                    "content": [
+                        {"type": "image_url", "image_url": {"url": self.image_url}},
+                        {"type": "image_url", "image_url": {"url": self.image_url}},
+                        {"type": "text", "text": "What's the difference between these two pictures? focus angel"},
+                    ],
+                }
+                ],
+            }
+            json_str = json.dumps(json_structure)
+            return ChatCompletionRequest.model_validate_json(json_str)
+        def local_get_processor_output(req: ChatCompletionRequest):
+            image2 = self.main_image.rotate(90).crop((0,0, 300, 300))
+            conv = generate_chat_conv(req, template_name=self.chat_template)
+            text = conv.get_prompt()
+            inputs = self.processor(
+                text=[text],
+                images=[[self.main_image, image2]],
+                return_tensors="pt",
+            ).to(self.device)
+            return inputs
+        req = local_get_completion_request()
+        processor_output = local_get_processor_output(req=req)
+        output = await self.engine.async_generate(
+            input_ids=[processor_output["input_ids"][0].detach().cpu().tolist()],
+            image_data=[self._pixel_values_image_data(processor_output)],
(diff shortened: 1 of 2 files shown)


JustinTong0323:
Great work! Thanks for the contribution! May you first resolve the conflicts?

cjhgo:
> Great work! Thanks for the contribution! May you first resolve the conflicts?

sure, i will resolve the conflicts and test compatibility again.

cjhgo:
## Question: How to Handle Multi-Image Input for Preprocessed Data?

hi there! I've tested the latest code and verified that single-image input works correctly. 
This improvement is very helpful for improving throughput! 

Now I have a question about how to properly handle multi-image inputs with preprocessed data.

### Current Understanding

I'm looking at the `load_mm_data` method and trying to understand how it handles multi-image scenarios with preprocessed inputs. The method supports both raw and preprocessed inputs through the `_load_single_item` method at line ~198:

```python
def _load_single_item(data, is_video, is_audio, frame_count_limit=None, discard_alpha_channel=True):
    if isinstance(data, dict):
        return data  # Direct return for preprocessed data
```

### My Question

For multi-image VLM scenarios, I notice there seems to be a format difference:

**Raw multi-image input:**
- Format: `image_data: [img0, img1]`
- Each image is processed independently
- Result: `images = [processed_img0, processed_img1]`  
- Token generation: 2 tokens (one per image)

**Preprocessed multi-image input:**
- Format: `dict(pixel_values=...)` where `pixel_values` contains merged results of multiple image slices/patches
- For MiniCPM VLM preprocessing, **multiple images are combined into a single dict**
- Result: `images = [dict(pixel_values=...)]`
- Token generation: 1 token (but contains multiple images)

### What I'm Wondering


How should I properly handle multi-image inputs when using preprocessed data? Should the preprocessed format include information about the number of images, or is there a different approach you'd recommend?

I want to make sure I'm handling this correctly to maintain consistency between raw and preprocessed multi-image inputs.

Thanks for any guidance!

---

JustinTong0323:
Hey @cjhgo , this is a really relevant question. Thanks a lot for pointing it out to us. Right now, the instructions about preprocessed data are kind of unclear. To make it clear, we figure that each input data item should be a single thing. Like for images, the input image_data should be a list of individual ImageDataItem. Each of these could be a dictionary, holding all the necessary info for one image, like pixel values and such.

cjhgo:
## Update: Multi-Image Input Compatibility Implementation

hey @JustinTong0323 thanks for the golden suggestion!
I've implemented the multi-image input compatibility for preprocessed data as discussed earlier. 
Here are the key changes:

### Problem Solved

Previously, there was an inconsistency between raw and preprocessed multi-image inputs:
- **Raw multi-image**: `image_data: [img0, img1]` → generates 2 tokens  
- **Preprocessed multi-image**: `image_data: [dict(pixel_values=...)]` → generated only 1 token (but contained multiple images)

### Solution Implemented

**1. Modified `_pixel_values_image_data` method in MiniCPM test class:**
```python
def _pixel_values_image_data(self, processor_output):
    image_slice_patch_size = len(processor_output.pixel_values[0])
    pre_dicts = []
    for i in range(image_slice_patch_size):
        pre_dicts.append(dict(
            modality="IMAGE",
            pixel_values=processor_output.pixel_values[0][i],
            tgt_size=processor_output.tgt_sizes[0][i],
        ))
    return pre_dicts
```

Instead of returning a single merged dict, we now decompose preprocessed multi-image data into individual dict items, with each dict containing `pixel_values` and `tgt_size` for one image slice/patch.

**2. Enhanced `process_mm_data_async` method:**
- Added proper handling for both merged and individual preprocessed data formats
- Implemented data flattening logic for consistent processing
- Maintained backward compatibility with existing single-image preprocessed inputs

**3. Added comprehensive testing:**
- `test_understands_pixel_values_with_imgs`: Tests multi-image input with two different images
- Verifies that the model correctly processes multiple images in preprocessed format


### Verification

I've tested this implementation with:
- Single image preprocessed inputs ✅ (confirmed working, great throughput improvement!)
- Multi-image preprocessed inputs ✅ (now properly supported)


This solution ensures that `load_mm_data` correctly handles multi-image compatibility for preprocessed inputs while preserving all the performance benefits of the original implementation.

### Questions for Reviewers

**1. Implementation Feasibility:** 
Could you please help evaluate whether this current implementation approach is feasible and aligns with the overall architecture design? I want to make sure this solution is robust and follows best practices.

**2. Test Coverage:** 
I've added multi-image validation methods to the test file, including `test_understands_pixel_values_with_imgs` which tests scenarios with multiple different images. However, this has made the test file considerably longer. 

**Should I keep these extended test cases, or would you prefer them to be:**
- Moved to a separate test file?
- Simplified to cover only essential scenarios?
- Kept as is for comprehensive coverage?

I'd appreciate your feedback on both the technical approach and the testing strategy!

test/srt/test_vlm_input_format.py
```diff
@@ -215,6 +215,77 @@ def _pixel_values_image_data(self, processor_output):
...
+        req = self.get_completion_request()
+        processor_output = self.get_processor_output(req=req)
+        output = await self.engine.async_generate(
+            input_ids=processor_output["input_ids"][0].detach().cpu().tolist(),
+            image_data=self._pixel_values_image_data(processor_output),
+            sampling_params=dict(temperature=0.0),
+        )
+        self.verify_response(output)
+
+    async def test_understands_pixel_values_with_imgs(self):
```
JustinTong0323, line 252: This test is good! Maybe we could move to VLMInputTestBase.

test/srt/test_vlm_input_format.py
```diff
@@ -215,6 +215,77 @@ def _pixel_values_image_data(self, processor_output):
...
+
+    async def test_understands_pixel_values(self):
+        req = self.get_completion_request()
+        processor_output = self.get_processor_output(req=req)
+        output = await self.engine.async_generate(
+            input_ids=processor_output["input_ids"][0].detach().cpu().tolist(),
+            image_data=self._pixel_values_image_data(processor_output),
+            sampling_params=dict(temperature=0.0),
+        )
+        self.verify_response(output)
```
JustinTong0323, line 250: This test seems the same as that in base?

cjhgo: > This test seems the same as that in base?

some tiny difference.
with `input_ids=processor_output["input_ids"][0].detach().cpu().tolist()`
TokenizerManager.generate_request will construct a single GenerateReqInput.

when image_data is single, it is okay to write `image_data=[self._pixel_values_image_data(processor_output)]` 
which will be `image_data=[dict]`.
when image_data is multi image for one prompt, i'd like to  pass `image_data=[dict,dict,dict...]`.
then `image_data=self._pixel_values_image_data(processor_output)` at base will not fit,
it will resolve to `image_data=[[dict,dict,dict...]]`.
so i have to remove the outside `[]` in the custom sub method.


on the other hand，
if we provide `input_ids=[processor_output["input_ids"][0].detach().cpu().tolist()]` like `test_understands_pixel_values_with_imgs`,
then TokenizerManager.generate_request will construct a GenerateReqInput whose is_single = False,
it is okay to write `image_data=[self._pixel_values_image_data(processor_output)]` .

input type is so flexible that we have to handle the subtle `[]`.

JustinTong0323: I think for a single request with multiple images, `image_data=[[dict,dict,dict...]]` is acceptable.

JustinTong0323: > input type is so flexible that we have to handle the subtle [].

Yes, it confused me a lot at beginning. Maybe we could try to figure out a better way, or fix a pattern for the input format, as it's only for offline inference.

cjhgo: > I think for a single request with multiple images, `image_data=[[dict,dict,dict...]]` is acceptable.

Also reasonable.
let me figure out how to handle this case 
so that no need to duplicate most base test code ~

PR #7902, opened 2025-07-09
[Dockerfile] Multi-arch support for ROCm

mqhc2020:
## Motivation

To support multi-arch for ROCm

## Modifications

- Dockerfile
- pyproject.toml

## Checklist

- [ ] Format your code according to the [Code Formatting with Pre-Commit](https://docs.sglang.ai/references/contribution_guide.html#code-formatting-with-pre-commit).
- [ ] Add unit tests as outlined in the [Running Unit Tests](https://docs.sglang.ai/references/contribution_guide.html#running-unit-tests-adding-to-ci).
- [ ] Update documentation / docstrings / example tutorials as needed, according to [Writing Documentation](https://docs.sglang.ai/references/contribution_guide.html#writing-documentation-running-docs-ci).
- [ ] Provide throughput / latency benchmark results and accuracy evaluation results as needed, according to [Benchmark and Profiling](https://docs.sglang.ai/references/benchmark_and_profiling.html) and [Accuracy Results](https://docs.sglang.ai/references/accuracy_evaluation.html).
- [ ] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co-author when merging the PR.
- [ ] Please feel free to join our Slack channel at https://slack.sglang.ai to discuss your PR.

PR #7958, opened 2025-07-11
[AMD] remove vllm version dependency

hliuca:
The current vllm pin is too old, vllm==0.6.7.dev2.  If we remove the vllm dep, it will allow more and newer vllm versions.

If we build from source and use newer vllm, following https://docs.sglang.ai/start/install.html, the build will fail.

`cd sglang

pip install --upgrade pip
cd sgl-kernel
python setup_rocm.py install
cd ..
pip install -e "python[all_hip]"`

image


hubertlu-tw:
@HaiShaw

hubertlu-tw:
@hliuca please provide the error messages and repro before this PR.

hliuca:
> @hliuca please provide the error messages and repro before this PR.

done

hliuca:
@HaiShaw @merrymercy the installation instructions have broken for very long time if user doesn't use 0.6.7.dev2, which is 6 months old. The latest vllm is 0.9.x. Thank you.

hliuca:
fixed in https://github.com/sgl-project/sglang/pull/7902

PR #7968, opened 2025-07-12
Revert "[PD Disaggregation] replace transfer with batch transfer for better performance (#7236)"

fzyzcjy:
cc @ShangmingCai


zhyncs:
@fzyzcjy @ShangmingCai do we need to use the older version of transfer engine

fzyzcjy:
I test using latest mooncake and it works

ShangmingCai:
@zhyncs No need, we only need to fix batch_transfer_sync with nvlink transport, then I will release a new package and bring this change back. The older transfer_sync is functioning well.

PR #8027, opened 2025-07-14
[feature] Ascend NPU graph support

### LINKED ISSUE #8030: [Feature] support ACLGraph
### Checklist

### Motivation

# 1. Motivation

After some features had been merged, we could run SGLang on Ascend servers with eager mode. But if we want to get better performance, we need implement ACLGraph or NPUGraph now.

Goals
Goal 1: Define a `NPUGraphRunner` class for SGLang, which provides basic functions and supports llama or Qwen models.

Goal 2: Adapt to TP/DP , GraphTree and dynamic shape scenarios, including memory reuse.

Goal 3: Improve performance based on `torch.compile`, 

# 2. Technical Design

- Workflow Phases

Image

- Key messages

  we have [torch_npu.npu.NPUGraph](https://gitee.com/ascend/pytorch/blob/master/torch_npu/npu/graphs.py#L246), which has similar interfaces and functions to [torch.cuda.CUDAGraph](https://docs.pytorch.org/docs/stable/generated/torch.cuda.CUDAGraph.html)

  Concerning the level of RTS, we can refer to this [document](https://www.hiascend.com/document/detail/zh/canncommercial/81RC1/developmentguide/appdevg/aclcppdevg/aclcppdevg_000519.html). 

# 3. Roadmap

- Phase 1: Basic support for dense model
   
   Implement `NPUGraphRunner` refer to `CUDAGraphRunner`, but we should handle some special case:

   Because we use this [torch_npu.npu_fused_infer_attention_score](https://www.hiascend.com/doc_center/source/zh/Pytorch/60RC2/apiref/apilist/ptaoplist_000787.html) API, which has a host_list input, we have to update its value each time using `torch_npu.npu.NPUGraph.update`. For more details, please refer to [task update](https://www.hiascend.com/document/detail/zh/canncommercial/81RC1/developmentguide/appdevg/aclcppdevg/aclcppdevg_000519.html#ZH-CN_TOPIC_0000002284281029__section4544173013324). 

- Phase 2: Basic support for moe model

### Related resources

_No response_

VDV1985:
## Motivation

NPUGraph can give significant improvement of perfrormance by reducing overhead of operators calling
Closes #8030 

## Modifications

All the GraphRunner functionlity moved to base GraphRunner class, CudaGraphRunner completely reuses it.
Added NPU graph runner, that inherited from GraphRunner and redefined several APIs
Added attention operation that supports by NPUGraph.update()
NPU graph runner uses the same SGLang server options as CUDA graph with combination of 'device="npu"', i.e. "disable_cuda_graph" and enabled by default. 

## Checklist

- [x] Format your code according to the [Code Formatting with Pre-Commit]image
   test_ascend_attention was executed with Qwen2.5-7B-Instruct model and showed following results:
   - test_gsm8k: Accuracy 0.864
   - test_latency with graph: output_throughput=36.45
   - test_latency without graph: output_throughpu=33.91


python/sglang/srt/model_executor/npu_graph_runner.py
```diff
@@ -0,0 +1,757 @@
...
+    from sglang.srt.model_executor.model_runner import ModelRunner
+
+# Detect whether the current forward pass is in capture mode
+is_capture_mode = False
+torch.cuda.CUDAGraph = torch.npu.NPUGraph
+torch.cuda.synchronize = torch.npu.synchronize
+torch.cuda.graph = torch.npu.graph
+torch.cuda.stream = torch.npu.stream
+torch.cuda.Stream = torch.npu.Stream
+torch.cuda.current_stream = torch.npu.current_stream
```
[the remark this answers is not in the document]
VDV1985: added guard _is_npu to not break other platforms

python/sglang/srt/layers/attention/ascend_backend.py
```diff
@@ -75,6 +82,63 @@ def init_forward_metadata(self, forward_batch: ForwardBatch):
             )
         self.forward_metadata.seq_lens_cpu_int = forward_batch.seq_lens_cpu.int()
 
+        self.graph_mode = False
+
+    def init_cuda_graph_state(self, max_bs: int, max_num_tokens: int):
```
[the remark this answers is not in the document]
VDV1985: This API comes from base attention class

ping1jing2:
Now we have 17 commits in this PR, maybe you are supposed to squash these.

scripts/npu_ci_install_dependency.sh
```diff
@@ -27,10 +27,10 @@ git clone --depth 1 https://github.com/vllm-project/vllm.git --branch $VLLM_TAG
 ### Install PyTorch and PTA
 PYTORCH_VERSION=2.6.0
 TORCHVISION_VERSION=0.21.0
-PTA_VERSION=2.6.0rc1
 pip install torch==$PYTORCH_VERSION torchvision==$TORCHVISION_VERSION --index-url https://download.pytorch.org/whl/cpu
-pip install torch_npu==$PTA_VERSION
-
+wget https://pytorch-package.obs.cn-north-4.myhuaweicloud.com/pta/Daily/v2.6.0/20250710.1/pytorch_v2.6.0_py311.tar.gz --no-check-certificate
+tar -xzvf ./pytorch_v2.6.0_py310.tar.gz
+pip install ./torch_npu-2.6.0.dev20250710-cp311-cp311-manylinux_2_28_aarch64.whl
```
ping1jing2: please change it like this `PTA_VERSION=2.6.0`, it's not reasonable to use daily version

Alisehen:
We conducted a performance comparison between SGLang and vLLM-Ascend when running Qwen3-32B with graph on the 910B3 NPU. The results show that SGLang spends significantly more time on non-computation, which negatively impacts overall performance.

Config:
Input = 4000, Output = 64, TP = 4, Batch size = 4

Results:
  •vLLM takes 44.1ms per decode, with 7.6ms spent on non-computation tasks.
  •SGLang takes 67.1ms per decode, with 33ms spent on non-computation tasks.

The computation times are similar for both, but SGLang has a longer decode interval due to higher non-computation overhead.
c025a93e-ccd5-4626-a6c9-cb8b04b52c65
c53331c1-e2ff-4d55-a280-4b51775ff2e6

Alisehen:
We also observed that during the decode process, the earlier layers tend to take more time than the later layers. For example, in the case of the FusedInferAttentionScore operator, despite having the same input shape, the computation time varies greatly across layers.
f3d06e1a-2e55-4f11-b78f-7011dfccf607
bcf9b4a9-6e15-422c-bd68-a1aa5af09558

scripts/npu_ci_install_dependency.sh
VDV1985: changed to wget of release 2.6.0 version

ping1jing2:
> We also observed that during the decode process, the earlier layers tend to take more time than the later layers. For example, in the case of the FusedInferAttentionScore operator, despite having the same input shape, the computation time varies greatly across layers. f3d06e1a-2e55-4f11-b78f-7011dfccf607 bcf9b4a9-6e15-422c-bd68-a1aa5af09558

Thanks for your feedback. After verification, it has been confirmed that the problem is caused by the steam restrication. you are supposed to use [this torch_npu](https://gitee.com/ascend/pytorch/releases/download/v7.1.0-pytorch2.6.0). the corresponding PR is [PR23272](https://gitee.com/ascend/pytorch/pulls/23272/files)

python/sglang/srt/distributed/parallel_state.py
```diff
@@ -245,6 +245,8 @@ def __init__(
 
         if is_cuda_alike():
             self.device = torch.device(f"cuda:{local_rank}")
+        elif is_npu():
```
ping1jing2: with `_is_npu`

python/sglang/srt/layers/attention/ascend_backend.py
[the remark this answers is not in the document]
jgong5: I guess we'd better abstract a base class serving for both CUDA graph and NPU graph. Then we extend it for NPU related logic in the base class.

python/sglang/srt/model_executor/npu_graph_runner.py
```diff
@@ -0,0 +1,71 @@
...
+if _is_npu:
+    torch.cuda.CUDAGraph = torch.npu.NPUGraph
+    torch.cuda.synchronize = torch.npu.synchronize
+    torch.cuda.graph = torch.npu.graph
+    torch.cuda.stream = torch.npu.stream
+    torch.cuda.Stream = torch.npu.Stream
+    torch.cuda.current_stream = torch.npu.current_stream
+
+
+class NPUGraphRunner(CudaGraphRunner):
```
jgong5: Instead of directly inheriting CudaGraphRunner, shall we abstract a base GraphRunner class and extend it with CudaGraphRunner and NPUGraphRunner?

test/srt/test_ascend_attention_backend.py
```diff
@@ -24,6 +24,35 @@
 
 
 class TestAscendAttnBackend(CustomTestCase):
```
jgong5: Can we add more npu graph tests? Suggest to add a test plan in the PR description.

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -1504,7 +1508,9 @@ def init_cuda_graphs(self):
         logger.info(
             f"Capture cuda graph begin. This can take up to several minutes. avail mem={before_mem:.2f} GB"
         )
-        self.cuda_graph_runner = CudaGraphRunner(self)
+        self.cuda_graph_runner = (
```
PanXun2: suggest to use `graph_runner` instead of `cuda_graph_runner` to avoid misleading

python/sglang/srt/model_executor/npu_graph_runner.py
VDV1985: Scheme with base abstract class will lead to almost full duplication of CudaGraphRunner code

test/srt/test_ascend_attention_backend.py
VDV1985: duplicated tp1 & tp2 Ascend tests with graph

python/sglang/srt/distributed/parallel_state.py
VDV1985: done

ping1jing2:
lgtm

python/sglang/srt/layers/attention/ascend_backend.py
```diff
@@ -75,6 +82,63 @@ def init_forward_metadata(self, forward_batch: ForwardBatch):
...
+        req_pool_indices: torch.Tensor,
+        seq_lens: torch.Tensor,
+        encoder_lens: Optional[torch.Tensor],
+        forward_mode: ForwardMode,
+        spec_info: Optional[Union[EagleDraftInput, EagleVerifyInput]],
+    ):
+        metadata = ForwardMetadata()
+
+        metadata.block_tables = self.graph_metadata["block_tables"][:bs, :]
+        metadata.seq_lens_cpu_list = seq_lens.cpu().int().tolist()
```
Alcanderian, line 109: Replay seq_lens could be different from captrue seq_lens. And we shoud not keep it here

python/sglang/srt/model_executor/cuda_graph_runner.py
```diff
@@ -49,13 +49,16 @@
...
     get_device_memory_capacity,
+    is_npu,
     rank0_log,
     require_attn_tp_gather,
     require_gathered_buffer,
     require_mlp_sync,
     require_mlp_tp_gather,
 )
 
+_is_npu = is_npu()
```
Alcanderian: We should not modify cuda_graph_runner.py because we will never use it in npu_graph mode

python/sglang/srt/model_executor/npu_graph_runner.py
Alcanderian: > Instead of directly inheriting CudaGraphRunner, shall we abstract a base GraphRunner class and extend it with CudaGraphRunner and NPUGraphRunner?

Cannot agree more

python/sglang/srt/layers/attention/ascend_backend.py
VDV1985: This is the required argument (i.e. CPU seq_lens) for torch_npu attention API's and cannot be excluded from capturing. At the replay stage we use graph.update to pass current value of seq_lens_cpu_list to torch_npu attention

python/sglang/srt/model_executor/cuda_graph_runner.py
VDV1985: Removed

python/sglang/srt/model_executor/npu_graph_runner.py
VDV1985: All the utility stuff and CudaGraphRunner functionality moved to grap_runner.py and base GraphRunner class. CudaGraphRunner and NPUGraphRunner are inherited from GraphRunner

python/sglang/srt/model_executor/npu_graph_runner.py
```diff
@@ -0,0 +1,112 @@
...
+from sglang.srt.model_executor.forward_batch_info import ForwardBatch, PPProxyTensors
+
+if _is_npu:
+    torch.cuda.CUDAGraph = torch.npu.NPUGraph
+    torch.cuda.synchronize = torch.npu.synchronize
+    torch.cuda.graph = torch.npu.graph
+    torch.cuda.stream = torch.npu.stream
+    torch.cuda.Stream = torch.npu.Stream
+    torch.cuda.current_stream = torch.npu.current_stream
+    torch.cuda.graph_pool_handle = torch.npu.graph_pool_handle
```
ping1jing2: i think we don't need these codes now, right?

ping1jing2: we need these codes in GraphRunner class

python/sglang/srt/model_executor/cuda_graph_runner.py
jgong5, line 1: Instead of removing `cuda_graph_runner.py` and create a new `graph_runner.py`. Is it better to just do "git mv"? I guess this would keep the change history and ease the code review too.

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -341,8 +342,11 @@ def initialize(self, min_per_gpu_memory: float):
             self.init_cublas()
             self.init_attention_backend()
             self.init_cuda_graphs()
+        elif self.device == "npu":
+            self.init_attention_backend()
+            self.init_cuda_graphs()
```
jgong5: Weird that we "init_cuda_graphs" for npu. Perhaps we can use a more general name such as `init_device_graphs`?

test/srt/test_ascend_graph_tp1_bf16.py
```diff
@@ -0,0 +1,95 @@
...
+        cls.url = urlparse(DEFAULT_URL_FOR_TEST)
+        cls.common_args = [
+            "--trust-remote-code",
+            "--mem-fraction-static",
+            0.8,
+            "--attention-backend",
+            "ascend",
+        ]
+
+    def test_a_gsm8k(self):
```
jgong5, line 40: What does "a" mean here?

test/srt/test_ascend_graph_tp1_bf16.py
```diff
@@ -0,0 +1,95 @@
...
+
+                    metrics = run_eval_few_shot_gsm8k(args)
+                    self.assertGreaterEqual(
+                        metrics["accuracy"],
+                        TEST_MODEL_MATRIX[model]["accuracy"],
+                    )
+                finally:
+                    kill_process_tree(process.pid)
+
+    def test_b_throughput(self):
```
jgong5, line 73: ditto: what does "b" mean here?

python/sglang/srt/layers/attention/ascend_backend.py
jgong5, line 1: There are still "cuda_graph" names in the "ascend_backend" module. We'd better use "npu_graph" or rename them to a more general name if they are used by both backends.

python/sglang/srt/model_executor/graph_runner.py
```diff
@@ -0,0 +1,855 @@
...
+    return global_graph_memory_pool
+
+
+def set_global_graph_memory_pool(val):
+    global global_graph_memory_pool
+    global_graph_memory_pool = val
+
+
+class GraphRunner:
+    """A CudaGraphRunner runs the forward pass of a model with cuda graph and torch.compile."""
```
jgong5: We'd better revise the code doc here since It is supposed to support multiple backends.

python/sglang/srt/model_executor/graph_runner.py
```diff
@@ -0,0 +1,855 @@
...
+                )
+                + "\n\nSorted by CPU Time:\n"
+                + prof.key_averages(group_by_input_shape=True).table(
+                    sort_by="cpu_time_total", row_limit=10
+                )
+            )
+            logger.info(log_message)
+
+    def _create_graph(self):
+        return torch.cuda.CUDAGraph()
```
jgong5: Instead of hard-code `torch.cuda` here, we may consider to set the device tag variable (e.g. "cuda" or "npu") to the `GraphRunner` so that we can implement this in a more general way, e.g., `getattr(torch, self.device).CUDAGraph()`.

python/sglang/srt/model_executor/npu_graph_runner.py
jgong5: Monkey patch doesn't look good. Suggest to generalize the `GraphRunner` implementation further to avoid this.

jgong5 (changes requested):
The update is on the right track of abstracting the code to support multiple graph backends. But we'd better generalize the `GraphRunner` more to avoid hard-coded "cuda" in the code so that it can support multiple backends following the open/closed principle. We should avoid use "cuda" in the ascend backend too.

test/srt/test_ascend_graph_tp1_bf16.py
VDV1985: just a kind of numbering

test/srt/test_ascend_graph_tp1_bf16.py
VDV1985: just a kind of numbering

python/sglang/srt/layers/attention/ascend_backend.py
VDV1985: This is require change of base class API and all attentions that support graphs. Lets postpone this change till normal HAL introduction

python/sglang/srt/model_executor/graph_runner.py
VDV1985: it doesn't solve problem as torch.npu doesn't have CUDAGraph but NPUGraph instead

python/sglang/srt/model_executor/graph_runner.py
```diff
@@ -304,13 +309,13 @@ def __init__(self, model_runner: ModelRunner):
             self.model_runner.lora_manager.init_cuda_graph_batch_info(self.max_bs)
 
         # Graph inputs
-        with torch.device("cuda"):
+        with torch.device(model_runner.device):
```
jgong5: nit:
```suggestion
        with torch.device(self.device):
```

python/sglang/srt/model_executor/graph_runner.py
```diff
@@ -380,12 +385,12 @@ def __init__(self, model_runner: ModelRunner):
                     * self.num_tokens_per_bs
                 ),
                 dtype=torch.bool,
-                device="cuda",
+                device=model_runner.device,
```
jgong5: ditto: self.device

python/sglang/srt/model_executor/graph_runner.py
```diff
@@ -380,12 +385,12 @@ def __init__(self, model_runner: ModelRunner):
...
             )
             self.next_token_logits_buffer = torch.zeros(
                 (self.max_num_token, self.model_runner.model_config.vocab_size),
                 dtype=torch.float,
-                device="cuda",
+                device=model_runner.device,
```
jgong5: ditto

python/sglang/srt/model_executor/graph_runner.py
```diff
@@ -304,13 +309,13 @@ def __init__(self, model_runner: ModelRunner):
...
         # Graph inputs
-        with torch.device("cuda"):
+        with torch.device(model_runner.device):
             self.input_ids = torch.zeros((self.max_num_token,), dtype=torch.int64)
             self.req_pool_indices = torch.zeros((self.max_bs,), dtype=torch.int32)
             self.seq_lens = torch.full(
                 (self.max_bs,), self.seq_len_fill_value, dtype=torch.int32
             )
-            self.out_cache_loc = torch.zeros((self.max_num_token,), dtype=torch.int64)
+            self.out_cache_loc = self._zero_tensor_with_dtype((self.max_num_token,))
```
jgong5: This change looks weird. Why do we wrap this call? Also `_zero_tensor_with_dtype` is not given a dtype as an arg, quite confusing.

python/sglang/srt/model_executor/model_runner.py
jgong5: Can you address my question? @VDV1985

python/sglang/srt/model_executor/npu_graph_runner_impl.py
jgong5, line 1: Naming it "npu_graph_runner.py" is simpler.

python/sglang/srt/utils.py
jgong5, line 1: For those functions with a single line of code, shall we call the implementation directly?
get_device_current_stream(device) -> torch.get_device_module(device).current_stream()

python/sglang/srt/layers/attention/ascend_backend.py
```diff
@@ -75,6 +82,63 @@ def init_forward_metadata(self, forward_batch: ForwardBatch):
...
+
+        self.graph_metadata[bs] = metadata
+        self.forward_metadata = metadata
+
+        self.graph_mode = True
+
+    def init_forward_metadata_replay_cuda_graph(
```
jgong5, line 116: nit: no cuda name in ascend backend.

jgong5 (commented):
Looks much better now while there are still some nits to fix. Please check my comments.

python/sglang/srt/model_executor/graph_runner.py
```diff
@@ -516,8 +526,13 @@ def capture(self) -> None:
...
             logger.info(log_message)
 
+    def _capture_graph(self, graph, pool, stream, run_once_fn):
+        with device_graph_func(self.device)(graph, pool=pool, stream=stream):
+            out = run_once_fn()
+        return out
+
     def capture_one_batch_size(self, bs: int, forward: Callable):
-        graph = torch.cuda.CUDAGraph()
+        graph = create_device_graph()
```
Alcanderian: abstract
```
def _create_device_graph():
```

And do impl in cuda graph and npu graph

python/sglang/srt/utils.py
```diff
@@ -2966,3 +2966,43 @@ async def wait_for_zero(self):
 @lru_cache(maxsize=1)
 def is_triton_kernels_available() -> bool:
     return importlib.util.find_spec("triton_kernels") is not None
+
+
+""" Device Graph and Stream utils."""
+
+
+def create_device_graph():
```
Alcanderian: I am not suppose to put create_device_graph here

python/sglang/srt/utils.py
```diff
@@ -2966,3 +2966,43 @@ async def wait_for_zero(self):
...
+
+def device_graph_func(device):
+    return torch.get_device_module(device).graph
+
+
+def device_stream_func(device):
+    return torch.get_device_module(device).stream
+
+
+def create_device_stream(device):
```
Alcanderian: I am not suppose put `torch.get_device_module(device)` into util functions especially for those whos result could not be cached.

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -920,7 +925,7 @@ def update_weights_from_tensor(
             )
 
         # We need to get device after patch otherwise the device would be wrong
-        infered_device = torch.cuda.current_device()
+        infered_device = get_current_device(self.device)
```
Alcanderian: add `self.device_module = torch.get_device_module(self.device)`

python/sglang/srt/model_executor/graph_runner.py
```diff
@@ -233,12 +237,13 @@ def set_global_graph_memory_pool(val):
...
 
-class CudaGraphRunner:
-    """A CudaGraphRunner runs the forward pass of a model with cuda graph and torch.compile."""
+class GraphRunner:
+    """A GraphRunner is a base class to run the forward pass of a model with device graph and torch.compile."""
 
     def __init__(self, model_runner: ModelRunner):
         # Parse args
         self.model_runner = model_runner
+        self.device = model_runner.device
```
Alcanderian, line 243:
```suggestion
        self.device = model_runner.device
        self.device_module = torch.get_device_module(self.device)
```

And then directly call `self.device_module.synchronize()` and other device module function

python/sglang/srt/model_executor/cuda_graph_runner_impl.py
Alcanderian, line 1: just keep the filename

python/sglang/srt/model_executor/npu_graph_runner_impl.py
Alcanderian, line 1: just keep the filename

python/sglang/srt/model_executor/cuda_graph_runner_impl.py
ssshinigami: ok, but git cannot recognize history changes and will show cuda_graph_runner.py as a new file. git mv doesn't help

python/sglang/srt/model_executor/npu_graph_runner_impl.py
ssshinigami: ok, but git cannot recognize history changes and will show cuda_graph_runner.py as a new file. git mv doesn't help

python/sglang/srt/model_executor/model_runner.py
ssshinigami: it has been changed to "init_device_graphs"

python/sglang/srt/model_executor/graph_runner.py
ssshinigami: fixed

python/sglang/srt/utils.py
ssshinigami: removed

python/sglang/srt/utils.py
ssshinigami: removed

python/sglang/srt/model_executor/model_runner.py
ssshinigami: done

python/sglang/srt/model_executor/graph_runner.py
ssshinigami: done

python/sglang/srt/model_executor/cuda_graph_runner_impl.py
ssshinigami: done

python/sglang/srt/model_executor/npu_graph_runner_impl.py
ssshinigami: done

python/sglang/srt/model_executor/graph_runner.py
ssshinigami: done

python/sglang/srt/layers/attention/ascend_backend.py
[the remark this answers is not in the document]
ping1jing2: `init_cuda_graph_state` is called by many files. How about doing this in another new PR?

python/sglang/srt/layers/attention/ascend_backend.py
ping1jing2: these APIs are called by many files. How about doing this in another new PR?

jgong5 (approved):
LGTM

Alcanderian (approved):

PR #8142, opened 2025-07-18
fix compressed tensors WNA16 imports

qeternity:
## Motivation

The current compressed tensors WNA16 implementation is broken due to bad imports.

I understand the long term plan is to deprecate compressed tensors, but in the interim, we should support this.

## Modifications

Fix imports.


python/sglang/srt/layers/quantization/compressed_tensors/compressed_tensors.py
```diff
@@ -41,6 +41,10 @@
 
 try:
     import vllm
```
AniZpZ: import vllm seems useless for now. we can remove this line

qeternity: Good point!

AniZpZ (approved):

Issue #6328, opened 2025-05-15
[Feature] two-chunk overlap for DeepSeekV3/R1


House-West:
### Checklist

- [x] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 2. Please use English, otherwise it will be closed.

### Motivation

In actual  cases, the request sequence lengths are not uniform  in the prefill node (4 nodes, tp1, dp32) . Consider the following two cases:
* Case 1:
each dp has extend_seq_len = [100, 2000], extend_prefix_len = [0, 0].
If we split it into [100] and [2000], the computation and communication times of the two micro-batches will differ significantly,  impacting overlap efficiency.
﻿
* Case 2:
  * dp0:           batch_size = 1, extend_seq_len = [4000], extend_prefix_len = [0, 0].
  * dp1-dp31 : batch_size = 2, extend_seq_len = [2000, 2000], extend_prefix_len = [0, 0].
In this scenario: either overlap cannot be enabled, or dp0 must construct an idle batch to enable overlap.
In both, the full potential of overlap cannot be achieved.
﻿

To address these issues, we propose two-chunk overlap, which also belongs to the scope of two-batch overlap.
* Case 1:
we split it into: 
  * extend_seq_len0 = [100, 950], extend_prefix_len0 = [0, 0]
  * extend_seq_len1 = [1050], extend_prefix_len1 = [950]

* Case2:
we split it into:
  * extend_seq_len0=[2000], extend_prefix_len0=[0]
  * extend_seq_len1=[2000], extend_prefix_len1=[2000]

In two-chunk overlap, there is a latent dependency between two micro-batches in mla computation.

![Image](https://github.com/user-attachments/assets/576fd6e0-3e5c-4940-b650-41c7e75f7365)

We implemented two-chunk overlap in early version of sglang, which differs significantly from the current version. 

By the way, we support idle batch in two-chunk overlap, which works when some dps are idle. Such as the following case：
  * dp0：idle batch
  * dp1-dp31: extend_seq_len=[2000]

House-West:
I want to see what the community members think about this feature.

Ewan-Ding:
Have the same idea, and I am curious about how much the maximum throughput and low batch output speed will change after we implement this mechanism? 

In addition, will this overlap mechanism be effective for other models after implementation?

House-West:
[...]

PR #8144, opened 2025-07-18
[Feature] improve TBO: two chunk overlap

House-West:
## Motivation

mentioned in #6328

I run some benchmark on 2 x 8 x H800.
### 1. Special Case (one request with a length of 3072 per dp)
**Experiment setup**
```
SGLANG_TBO_DEBUG=1 SGL_CHUNKED_PREFIX_CACHE_THRESHOLD=1 python3 -m sglang.launch_server \
    --model-path /dev/shm/DeepSeek-V3-0324 --tp 16 --dp 16 --dist-init-addr $MASTER_IP:6676 \
    --chunked-prefill-size 65536 --max-prefill-tokens 170000 --nnodes 2 --node-rank $FED_POD_INDEX \
    --trust-remote-code --deepep-mode normal --enable-dp-attention --enable-deepep-moe \
    --enable-two-batch-overlap --disable-overlap-schedule --disable-radix-cache --disable-cuda-graph \
    --max-total-tokens 34000 --mem-fraction-static 0.90  --max-running-requests 16 \
    --tbo-token-distribution-threshold 0.48

        --backend sglang --host 127.0.0.1 --port 30000 \
        --dataset-name random \
        --num-prompt 1024 \
        --random-input 3072 --random-output 1 --random-range-ratio 1 \
        --max-concurrency 1024

```
For baseline and this PR,  change `tbo-token-distribution-threshold`, 0.0 indicates disable two-chunk-overlap, threshold > 0 indicates  enable two-chunk-overlap.

`two-chunk-overlap` will trigger chunk prefill, `SGL_CHUNKED_PREFIX_CACHE_THRESHOLD=1 ` can improve performance.

The bench-serving script is repeated 5 times, and throw away the 1st run (because it contains JIT compilation etc).

**Experiment result** 

Throughput

* baseline: 64820.55, 64901.24, 63900.35, 63931.12 
* ours: 72391.92, 72845.28, 71511.55, 73152.61

On average, it improves 12.56% throughput. 

### 2. General Case (variable length inputs (30-3072))
**Experiment setup**
```
SGLANG_TBO_DEBUG=1 SGL_CHUNKED_PREFIX_CACHE_THRESHOLD=1 python3 -m sglang.launch_server \
    --model-path /dev/shm/DeepSeek-V3-0324 --tp 16 --dp 16 --dist-init-addr $MASTER_IP:6676 \
    --chunked-prefill-size 65536 --max-prefill-tokens 170000 --nnodes 2 --node-rank $FED_POD_INDEX \
    --trust-remote-code --deepep-mode normal --enable-dp-attention --enable-deepep-moe \
    --enable-two-batch-overlap --disable-overlap-schedule --disable-radix-cache --disable-cuda-graph \
    --max-total-tokens 34000 --mem-fraction-static 0.90  --max-running-requests 512 \
    --tbo-token-distribution-threshold 0.48

        --backend sglang --host 127.0.0.1 --port 30000 \
        --dataset-name random \
        --num-prompt 2048 \
        --random-input 3072 --random-output 1 --random-range-ratio 0.01 \
        --max-concurrency 1024

```
**Experiment result** 

Throughput

* baseline: 71695.06, 68585.39, 75784.86, 77115.82
* ours: 77532.36, 77059.32, 75654.72, 78041.45

On average, it improves 5.15% throughput. The range of performance improvement depends on the distribution of input lengths

## Checklist


python/sglang/srt/managers/schedule_batch.py
```diff
@@ -84,6 +84,7 @@
     "disable_radix_cache",
     "enable_dp_attention",
     "enable_two_batch_overlap",
+    "two_batch_token_distribution_threshold",
```
fzyzcjy: nit: two batch overlap is called "tbo" in the code

python/sglang/srt/server_args.py
```diff
@@ -205,6 +205,7 @@ class ServerArgs:
     enable_dp_attention: bool = False
     enable_dp_lm_head: bool = False
     enable_two_batch_overlap: bool = False
+    two_batch_token_distribution_threshold: float = 0.48
```
fzyzcjy: nit: curious why it is not 0.50

python/sglang/srt/model_executor/forward_batch_info.py
```diff
@@ -398,16 +398,12 @@ def init_new(
...
-                positions, ret.extend_start_loc = compute_position_triton(
-                    ret.extend_prefix_lens,
-                    ret.extend_seq_lens,
-                    ret.extend_num_tokens,
-                )
-            else:
-                positions, ret.extend_start_loc = compute_position_torch(
-                    ret.extend_prefix_lens, ret.extend_seq_lens
-                )
+            positions, ret.extend_start_loc = compute_position(
```
fzyzcjy, line 423: nit: this refactor looks good, but maybe we can create a separate PR for this to reduce the code change of this PR

python/sglang/srt/two_batch_overlap.py
fzyzcjy, line 1: TODO(@fzyzcjy): review core

fzyzcjy:
core is not reviewed yet, firstly glanced at the non-core things

python/sglang/srt/managers/schedule_batch.py
House-West: Perhaps it would be more appropriate to name it `token_distribution_threshold_of_tbo`？

python/sglang/srt/server_args.py
House-West: When the input is [498, 502], it is more appropriate to split it into [498] and [502], rather than [498, 2] and [500], which would increase the latency of attention. The value of 0.48 has not been rigorously demonstrated, 0.485 or 0.49 could equally be considered.

python/sglang/srt/model_executor/forward_batch_info.py
House-West: The `compute_position` is also called in `two_batch_overlap.py`. To reduce code duplication, this part has been refactored. I think it is the part of this PR.

python/sglang/srt/managers/schedule_batch.py
fzyzcjy: I personally suggest to keep it aligned with existing ones, i.e. `tbo_token_distribution_threshold`

python/sglang/srt/server_args.py
fzyzcjy: That looks reasonable

python/sglang/srt/model_executor/forward_batch_info.py
fzyzcjy: I see

python/sglang/srt/two_batch_overlap.py
fzyzcjy, line 1: would be great to show some benchmark results to demonstrate it does speed up

python/sglang/srt/two_batch_overlap.py
```diff
@@ -63,7 +67,40 @@ def compute_split_seq_index(
         raise NotImplementedError()
 
 
+def _is_enabled_two_chunk_split(left_sum, overall_sum):
+    threshold = global_server_args_dict["two_batch_token_distribution_threshold"]
```
fzyzcjy: nit: what about adding `assert threshold <= 0.50` to avoid users accidentally give a wrong value

python/sglang/srt/two_batch_overlap.py
```diff
@@ -63,7 +67,40 @@ def compute_split_seq_index(
...
+    return left_sum < overall_sum * threshold or left_sum > overall_sum * (
+        1 - threshold
+    )
+
+
 def _split_array_by_half_sum(arr: Sequence[int]) -> int:
```
fzyzcjy: well the naming has some tiny issues. what about

```
# this is the one to be called in compute_split_seq_index
def _split_extend_seqs():
   split_seq_index = _split_array_by_half_sum(arr)
   ... other logic you have

def _split_array_by_half_sum(arr ...):
  the old code
  note that here we do not know the arr is seq or token or whatever. it is just general array.
```

python/sglang/srt/two_batch_overlap.py
```diff
@@ -63,7 +67,40 @@ def compute_split_seq_index(
...
+
+    left_sum = sum(arr[:split_seq_index])
+    overall_sum = sum(arr)
+    if _is_enabled_two_chunk_split(left_sum, overall_sum):
+        return _split_array_by_token(arr)
+
+    return split_seq_index
+
+
+def _split_array_by_token(arr):
```
fzyzcjy: would be great to make it a general function, i.e. be unaware of "the arr is some seq and some tokens" but mimic my old _split_array_by_half_sum, i.e. "it is just a list of numbers and I dnk what it repsents"

python/sglang/srt/two_batch_overlap.py
```diff
@@ -63,7 +67,40 @@ def compute_split_seq_index(
         raise NotImplementedError()
 
 
+def _is_enabled_two_chunk_split(left_sum, overall_sum):
```
fzyzcjy: tiny nit: shall we call it `is_two_chunk_split_enabled`

python/sglang/srt/two_batch_overlap.py
```diff
@@ -415,9 +468,72 @@ def prepare_raw(
...
+        *,
+        child_a: ForwardBatch,
+        child_b: ForwardBatch,
+        tbo_split_seq_index: int,
+    ):
+        extend_seq_lens_cpu = batch.extend_seq_lens_cpu
+        overall_sum = sum(extend_seq_lens_cpu)
+        half_sum = overall_sum // 2
+        left_token_num = half_sum - sum(extend_seq_lens_cpu[:tbo_split_seq_index])
+        right_token_num = extend_seq_lens_cpu[tbo_split_seq_index] - left_token_num
```
fzyzcjy: nit: maybe call it left_last_seq_token_num, right_first_seq_token_num

python/sglang/srt/two_batch_overlap.py
```diff
@@ -415,9 +468,72 @@ def prepare_raw(
...
+
+        child_a.extend_seq_lens = torch.tensor(
+            child_a.extend_seq_lens_cpu, dtype=batch.extend_seq_lens.dtype
+        ).to(device=global_server_args_dict["device"], non_blocking=True)
+        child_b.extend_seq_lens = torch.tensor(
+            child_b.extend_seq_lens_cpu, dtype=batch.extend_seq_lens.dtype
+        ).to(device=global_server_args_dict["device"], non_blocking=True)
+
+        child_a.extend_num_tokens = sum(child_a.extend_seq_lens_cpu)
+        child_b.extend_num_tokens = sum(child_b.extend_seq_lens_cpu)
```
fzyzcjy: nit: for such derived data, maybe extract to utils

e.g. most naive is

```
def some_postprocessing(x):
  x.extend_seq_lens = x.extend_seq_lens_cpu.to
  x.extend_num_tokens = ...
```

and call it

```
some_postprocessing(child_a)
some_postprocessing(child_b)
```

or even better, is it possible we extract these derived data computation from the main code to utils, and call it (again) here

python/sglang/srt/two_batch_overlap.py
```diff
@@ -415,9 +468,72 @@ def prepare_raw(
...
+        ), f"{child_a.extend_num_tokens=}, {half_sum=}"
+
+        seq_lens_cpu = (
+            child_a.seq_lens_cpu.clone()
+        )  # create a new seq_lens_cpu for child_a
+        seq_lens_cpu[-1] = (
+            child_a.extend_seq_lens_cpu[-1] + child_a.extend_prefix_lens_cpu[-1]
+        )
+        child_a.seq_lens_cpu = seq_lens_cpu
+        child_a.seq_lens = seq_lens_cpu.to(
```
fzyzcjy: (same for this, etc)

python/sglang/srt/two_batch_overlap.py
```diff
@@ -415,9 +468,72 @@ def prepare_raw(
...
+                batch,
+                child_a=child_a,
+                child_b=child_b,
+                tbo_split_seq_index=batch.tbo_split_seq_index,
+            )
+
         assert batch.tbo_children is None
         batch.tbo_children = [child_a, child_b]
 
+    @classmethod
```
fzyzcjy, line 519: seems this function can be short:

1. properly set child_a.extend_seq_lens_cpu, child_b.extend_seq_lens_cpu, etc
2. call that "derive all other fields" util

fzyzcjy, line 519: btw wondering whether we have some shorter way to implement it

python/sglang/srt/two_batch_overlap.py
fzyzcjy: ignore me, I see the speedup

python/sglang/srt/managers/schedule_batch.py
House-West: done.

python/sglang/srt/two_batch_overlap.py
House-West: done.

python/sglang/srt/two_batch_overlap.py
House-West: I have added parameter type.

python/sglang/srt/two_batch_overlap.py
House-West: I have changed it to `_split_extend_seqs`. Would it be more appropriate to call these two methods as `_split_array_by_sequence` and `_split_array_by_token`?

python/sglang/srt/two_batch_overlap.py
House-West: done.

python/sglang/srt/two_batch_overlap.py
House-West: done.

python/sglang/srt/two_batch_overlap.py
House-West: I moved this code to `_modify_some_fields`.

python/sglang/srt/two_batch_overlap.py
House-West: I moved this code to `_modify_some_fields`.

python/sglang/srt/two_batch_overlap.py
House-West: I changed the function name to `derive_some_fields_for_two_chunk`

House-West: I tried to modify it in `filter_batch` before, but due to `left_last_seq_token_num` and `right_first_seq_token_num`, it would result in many conditional branches, which would reduce the readability of the code.

fzyzcjy: I see

python/sglang/srt/two_batch_overlap.py
```diff
@@ -415,9 +497,78 @@ def prepare_raw(
...
+    def derive_some_fields_for_two_chunk(
```
fzyzcjy: oh I do not mean really call it "some" in the function name!

House-West: fixed

test/srt/test_two_batch_overlap.py
```diff
@@ -72,35 +75,46 @@ def test_mmlu(self):
 
 
 class TestTwoBatchOverlapUnitTest(unittest.TestCase):
-    # TODO change tests when having 6328
-    def test_compute_split_seq_index(self):
+    def test_compute_split_seq_and_token_index(self):
```
fzyzcjy, line 79: nit: this is not in CI thus would be great to check it locally and paste it here

python/sglang/srt/two_batch_overlap.py
```diff
@@ -63,7 +67,41 @@ def compute_split_seq_index(
...
+
+def _split_extend_seqs(arr: Sequence[int]) -> int:
+    split_seq_index = _split_array_by_sequence(arr)
+
+    left_sum = sum(arr[:split_seq_index])
+    overall_sum = sum(arr)
+    if _is_two_chunk_split_enabled(left_sum, overall_sum):
+        return _split_array_by_token(arr)
+
+    return split_seq_index
```
fzyzcjy: tiny nit: I still personally think it would be great to name like

* _split_array_by_sequence -> _split_array_by_balanced_sum
* _split_array_by_token -> _split_array_by_cum_less_than_half

to let name reflect the algorithm, then readers do not need to check the two func impl to understand what is going on

python/sglang/srt/two_batch_overlap.py
```diff
@@ -63,7 +67,41 @@ def compute_split_seq_index(
...
+
+
+def _split_array_by_token(arr: Sequence[int]) -> int:
+    left_sum = 0
+    overall_sum = sum(arr)
+    half_sum = overall_sum // 2
+    split_seq_index = 0
```
fzyzcjy: nit: maybe rename as chosen_index

python/sglang/srt/two_batch_overlap.py
```diff
@@ -82,6 +120,34 @@ def _split_array_by_half_sum(arr: Sequence[int]) -> int:
     return best_index
 
 
+def _modify_some_fields(
```
fzyzcjy: (same as before, "modify_some_fields" is a placeholder name instead of really doing this; same for other functions)

fzyzcjy (commented):
(I will check again later)

test/srt/test_two_batch_overlap.py
House-West: TestTwoBatchOverlapUnitTest：

extend_lens=[] expect=(0, 0) actual=(0, 0)
extend_lens=[42] expect=(0, 21) actual=(0, 21)
extend_lens=[42, 999] expect=(1, 520) actual=(1, 520)
extend_lens=[999, 42] expect=(0, 520) actual=(0, 520)
extend_lens=[498, 502] expect=(1, 498) actual=(1, 498)
extend_lens=[4096, 4096, 4096, 4096] expect=(2, 8192) actual=(2, 8192)
extend_lens=[4095, 4096, 4096, 4096, 1] expect=(2, 8191) actual=(2, 8191)
extend_lens=[1, 4095, 4096, 4096, 4096] expect=(3, 8192) actual=(3, 8192)
extend_lens=[4097, 4096, 4096, 4095, 1] expect=(2, 8193) actual=(2, 8193)
extend_lens=[1, 1, 1, 1, 99999] expect=(4, 50001) actual=(4, 50001)
extend_lens=[99999, 1, 1, 1, 1] expect=(0, 50001) actual=(0, 50001)
.

Ran 1 test in 0.000s

OK

python/sglang/srt/two_batch_overlap.py
House-West: done.

python/sglang/srt/two_batch_overlap.py
House-West: done

python/sglang/srt/two_batch_overlap.py
House-West: renamed as `_update_device_and_sum_field_from_cpu_field`

python/sglang/srt/two_batch_overlap.py
House-West: It has been fixed based on the following comment

test/srt/test_two_batch_overlap.py
House-West: TestQwen3TwoBatchOverlap and TestTwoBatchOverlap：

Writing report to /tmp/mmlu__ssd1_models_huggingface.co_Qwen_Qwen3-30B-A3B.html
{'other': np.float64(0.8125), 'other:std': np.float64(0.3903123748998999), 'score:std': np.float64(0.4236075534914362), 'stem': np.float64(0.8181818181818182), 'stem:std': np.float64(0.385694607919935), 'humanities': np.float64(0.6956521739130435), 'humanities:std': np.float64(0.4601306627938419), 'social_sciences': np.float64(0.7857142857142857), 'social_sciences:std': np.float64(0.41032590332414487), 'score': np.float64(0.765625)}
Writing results to /tmp/mmlu__ssd1_models_huggingface.co_Qwen_Qwen3-30B-A3B.json
Total latency: 512.358 s
Score: 0.766

Writing report to /tmp/mmlu__root_.cache_modelscope_hub_models_deepseek-ai_DeepSeek-Coder-V2-Lite-Instruct.html
{'other': np.float64(0.6875), 'other:std': np.float64(0.46351240544347894), 'score:std': np.float64(0.4938587696649721), 'stem': np.float64(0.6363636363636364), 'stem:std': np.float64(0.48104569292083466), 'humanities': np.float64(0.43478260869565216), 'humanities:std': np.float64(0.49572844569527735), 'social_sciences': np.float64(0.6428571428571429), 'social_sciences:std': np.float64(0.47915742374995496), 'score': np.float64(0.578125)}
Writing results to /tmp/mmlu__root_.cache_modelscope_hub_models_deepseek-ai_DeepSeek-Coder-V2-Lite-Instruct.json
Total latency: 104.254 s
Score: 0.578
.

python/sglang/srt/two_batch_overlap.py
```diff
@@ -63,7 +67,41 @@ def compute_split_seq_index(
...
+        return _split_array_by_cum_less_than_half(arr)
+
+    return split_seq_index
+
+
+def _split_array_by_cum_less_than_half(arr: Sequence[int]) -> int:
+    left_sum = 0
+    overall_sum = sum(arr)
+    half_sum = overall_sum // 2
+    chosen_seq_index = 0
```
fzyzcjy: nit: as mentioned before maybe chosen_index, i.e. make this function unaware of what the arr is

python/sglang/srt/two_batch_overlap.py
```diff
@@ -175,7 +241,11 @@ def compute_split_token_index(
 ) -> int:
     if forward_mode == ForwardMode.EXTEND:
         assert extend_seq_lens is not None
-        return sum(extend_seq_lens[:split_seq_index])
+        split_token_index = sum(extend_seq_lens[:split_seq_index])
```
fzyzcjy: nit: maybe name it vanilla_split_token_index, o/w readers may be a bit confused when we return sth other than the split_token_index variable

python/sglang/srt/two_batch_overlap.py
```diff
@@ -379,9 +449,17 @@ def prepare_raw(
 
         tbo_split_token_index = cls._compute_split_token_index(batch)
 
+        is_enable_two_chunk = (
+            batch.forward_mode == ForwardMode.EXTEND
+            and batch.extend_seq_lens_cpu is not None
```
fzyzcjy: nit: curious when will it be none when mode=EXTEND

python/sglang/srt/two_batch_overlap.py
```diff
@@ -379,9 +449,17 @@ def prepare_raw(
...
+            and tbo_split_token_index
+            != sum(batch.extend_seq_lens_cpu[: batch.tbo_split_seq_index])
```
fzyzcjy: hmm wondering whether this condition can be improved a bit, e.g. shall we have sth like

```
def _is_two_chunk_split_enabled(extend_lens):
    vanilla_split_seq_index = _split_array_by_balanced_sum(extend_lens)
    left_sum = sum(arr[:vanilla_split_seq_index])
    overall_sum = sum(arr)
    ... original logic ...
```

(note that we will waste some cpu recompute in this case, but that is too tiny to be visible)

the benefit of this function signature is that, we are clear whether two chunk overlap is enabled given a extend_lens, without the need to compute any others
and here, we can change `tbo_split_token_index            != sum(batch.extend_seq_lens_cpu[: **batch.tbo_split_seq_index])` to `_is_two_chunk_split_enabled(batch.extend_lens)` etc

python/sglang/srt/two_batch_overlap.py
```diff
@@ -415,9 +497,78 @@ def prepare_raw(
...
+        cls,
+        batch: ForwardBatch,
+        *,
+        child_a: ForwardBatch,
+        child_b: ForwardBatch,
+        tbo_split_seq_index: int,
+    ):
+        extend_seq_lens_cpu = batch.extend_seq_lens_cpu
+        overall_sum = sum(extend_seq_lens_cpu)
+        half_sum = overall_sum // 2
```
fzyzcjy: nit: overall_seq_lens_sum, half_seq_lens_sum, o/w "sum" may be tiny bit unclear

python/sglang/srt/two_batch_overlap.py
```diff
@@ -415,9 +497,78 @@ def prepare_raw(
...
+        assert (
+            child_a.extend_num_tokens == half_sum
+        ), f"{child_a.extend_num_tokens=}, {half_sum=}"
+
+        seq_lens_cpu = (
+            child_a.seq_lens_cpu.clone()
+        )  # create a new seq_lens_cpu for child_a
+        seq_lens_cpu[-1] = (
+            child_a.extend_seq_lens_cpu[-1] + child_a.extend_prefix_lens_cpu[-1]
+        )
```
fzyzcjy: nit: curious why we do a clone here, while on child_b we do not

fzyzcjy (commented):
LGTM, only some nits and ready to merge

python/sglang/srt/two_batch_overlap.py
House-West: done

python/sglang/srt/two_batch_overlap.py
House-West: The code has been simplified after modifying the implementation of `_is_two_chunk_split_enabled`.

python/sglang/srt/two_batch_overlap.py
House-West: It may not happen. I moved it to `_is_two_chunk_split_enabled`

python/sglang/srt/two_batch_overlap.py
House-West: done

python/sglang/srt/two_batch_overlap.py
House-West: done

python/sglang/srt/two_batch_overlap.py
House-West: seq_lens_cpu is a tensor, not a list. After slicing, the pointers of the seq_lens_cpu of origin forward_batch, child_a, and child_b are the same. If only the seq_lens_cpu of child_a is modified, it will modify both origin forward_batch and child_b. However, the seq_len of origin forward_batch and child_b does not need to be modified.

fzyzcjy: i see

fzyzcjy: tiny nit: then what about making child_b.extend_prefix_lens_cpu also deepcopy / cloned to be extra safe in case in the future somehow the upstream changes and makes it shared

fzyzcjy (approved):

fzyzcjy:
only that tiny nit and then ready to merge

python/sglang/srt/two_batch_overlap.py
House-West: fixed, and also deepcopy for extend_seq_lens_cpu

fzyzcjy:
LGTM, now only need to wait for CI green

House-West:
@zhyncs  Could you view this pr, waiting approval to merge

ch-wan:
@House-West Nice work! For the first case, I wonder to know where the performance gain comes from. Is two chunk overlap equivalent to two batch overlap under that case? Thanks.

House-West:
> @House-West Nice work! For the first case, I wonder to know where the performance gain comes from. Is two chunk overlap equivalent to two batch overlap under that case? Thanks.

@ch-wan as mentioned in #6328. When each dp has only one request, two-chunk-overlap and two- batch-overlap are not equivalent.
such as batch_size = 1, extend_seq_len = [3072], extend_prefix_len = [0]

In two- batch-overlap:
* micro batch0: extend_seq_len = [3072], extend_prefix_len = [0]
* micro batch1 (idle batch): extend_seq_len = [0], extend_prefix_len = [0]

In two- chunk-overlap:
* micro batch0: extend_seq_len = [1536], extend_prefix_len = [0]
* micro batch1: extend_seq_len = [1536], extend_prefix_len = [1536]

Compared to two-batch-overlap, the latency of group gemm, dispatch, and combine operations of the two micro batches is close in two-chunk-overlap, which is better for overlapping.

ch-wan (commented):
Can we add a small CI test to `test_dp_attention.py`?

House-West:
> Can we add a small CI test to `test_dp_attention.py`?

I saw the test case of TBO in `test_deepep_small`. If TBO is ebabled, the two-chunk-overlap will be enabled in most cases. I think the test case of two-chunk-overlap can reuse the TBO, without the need for additional additions. And I have updated the parameter descriptions in `server_arguments.md`

ch-wan (approved):

House-West:
> @House-West @fzyzcjy @ch-wan I'd like to ask if you achieved performance improvements using TBO on two machines (1P1D)? Or did you achieve results without separating the PD (processing and data) on different machines? On my H20 system, performance decreased both with and without PD separation (1P1D). I've seen some publicly available experimental results online where TBO showed improvements, but they used many machines, such as 4P9D and 4P16D. I also tried your parameters, but the performance decreased significantly. Did I do something wrong? Or does TBO require a significant number of machines to achieve performance improvements? My rough analysis suggests that at least three machines are needed for the overhead and communication costs of TBO to be offset by the computational overlap, and perhaps five machines are needed to see any improvement. Looking forward to your reply！

@programmer-lxj I had tested on h800 previously , and there was no performance improvement with TBO on 1P1D (one machine for prefill, one machine for decode).  Because intranode communication uses NVLink, which is relatively fast. In most cases, TBO can bring performance improvements when the communication time  more than 30% of end-to-end time, such as 2P9D or 4P9D. For 1P1D , you can try SBO(single batch overlap).

programmer-lxj:
@House-West I tried using SBO, but I found that this parameter can only be added to the P node; adding it to the D node results in an error. I checked the code and found that the `_combine_core` function in `sglang`'s `deeprp.py` passes the `overlap_args` dictionary parameter, but the `Buff.low_latency_combine` function in the `deepep` library doesn't have this parameter, which causes the error.  Does this mean SBO can only be added to the P node and not the D node? Also, when using 1P1D, adding SBO only to the P node (and not the D node) resulted in a slight performance decrease, not an improvement. I wonder if there are other techniques I should be aware of.
deepep
buffer

PR #8215, opened 2025-07-21
Support loading weights from remote instance

amysaq2023:
## Motivation

During engine initialization, the process of loading model weights is time-consuming, as the weights are currently fetched either from disk or from system memory, both of which are constrained by PCIe bandwidth. To optimize the startup time of the engine, we propose a method to load the weights directly from another running SGLang instance that is using the same model. In this approach, the data path for weight transfer is from the source instance's GPU memory → source instance's NIC → destination instance's NIC → destination instance's GPU memory.

## Modifications

This PR introduces a new load-format option, "remote_instance", which enables a new instance to load model weights from an already running remote instance during initialization. When using the "remote_instance" load format, the new instance will:
1. establish a communication group with the target instance; and
2. transfer the weights from the target instance via the established communication group.
sglang-upstream-2

To start a new instance using "remote_instance" load format:
`python -m sglang.launch_server`
`--model-path instance://[target_instance_ip]:[communication_group_port] `
`--tokenizer-path [local_tokenizer_path]`
`--seed-instance-url http://[target_instance_ip]:[target_instance_serving_port]`
`--load-format remote_instance`
`--client-instance-id [local_instance_id](optional)`

Performance tested with Deepseek-R1:
1. from disk: ～440s
2. from main memory: ~60s
3. from remote instance: <10s

## Checklist

### DIFF
--- a/python/sglang/srt/configs/device_config.py
+++ b/python/sglang/srt/configs/device_config.py
@@ -8,10 +8,12 @@ logger = logging.getLogger(__name__)
 
 class DeviceConfig:
     device: Optional[torch.device]
+    gpu_id: Optional[int]
 
-    def __init__(self, device: str = "cuda") -> None:
+    def __init__(self, device: str = "cuda", gpu_id: int = -1) -> None:
         if device in ["cuda", "xpu", "hpu", "cpu", "npu"]:
             self.device_type = device
         else:
             raise RuntimeError(f"Not supported device type: {device}")
         self.device = torch.device(self.device_type)
+        self.gpu_id = gpu_id
--- a/python/sglang/srt/configs/load_config.py
+++ b/python/sglang/srt/configs/load_config.py
@@ -23,6 +23,7 @@ class LoadFormat(str, enum.Enum):
     LAYERED = "layered"
     JAX = "jax"
     REMOTE = "remote"
+    REMOTE_INSTANCE = "remote_instance"
 
 
 @dataclass
--- a/python/sglang/srt/configs/model_config.py
+++ b/python/sglang/srt/configs/model_config.py
@@ -64,12 +64,28 @@ class ModelConfig:
         is_draft_model: bool = False,
         hybrid_kvcache_ratio: Optional[float] = None,
         model_impl: Union[str, ModelImpl] = ModelImpl.AUTO,
+        tp_rank: Optional[int] = None,
+        remote_instance_weight_loader_seed_instance_ip: Optional[str] = None,
+        remote_instance_weight_loader_seed_instance_service_port: Optional[int] = None,
+        remote_instance_weight_loader_send_weights_group_ports: Optional[
+            List[int]
+        ] = None,
     ) -> None:
         # Parse args
         self.model_path = model_path
         self.revision = revision
         self.quantization = quantization
         self.model_impl = model_impl
+        self.tp_rank = tp_rank
+        self.remote_instance_weight_loader_seed_instance_ip = (
+            remote_instance_weight_loader_seed_instance_ip
+        )
+        self.remote_instance_weight_loader_seed_instance_service_port = (
+            remote_instance_weight_loader_seed_instance_service_port
+        )
+        self.remote_instance_weight_loader_send_weights_group_ports = (
+            remote_instance_weight_loader_send_weights_group_ports
+        )
 
         self.maybe_pull_model_tokenizer_from_remote()
         self.model_override_args = json.loads(model_override_args)
@@ -320,6 +336,9 @@ class ModelConfig:
             quantization=server_args.quantization,
             hybrid_kvcache_ratio=server_args.hybrid_kvcache_ratio,
             model_impl=server_args.model_impl,
+            remote_instance_weight_loader_seed_instance_ip=server_args.remote_instance_weight_loader_seed_instance_ip,
+            remote_instance_weight_loader_seed_instance_service_port=server_args.remote_instance_weight_loader_seed_instance_service_port,
+            remote_instance_weight_loader_send_weights_group_ports=server_args.remote_instance_weight_loader_send_weights_group_ports,
             **kwargs,
         )
 
--- a/python/sglang/srt/connector/remote_instance.py
+++ b/python/sglang/srt/connector/remote_instance.py
@@ -1,0 +1,82 @@
+# SPDX-License-Identifier: Apache-2.0
+
+import logging
+from typing import Generator, List, Optional, Tuple
+from urllib.parse import urlparse
+
+import torch
+import torch.distributed as dist
+
+from sglang.srt.connector import BaseConnector
+from sglang.srt.utils import init_custom_process_group
+
+logger = logging.getLogger(__name__)
+
+
+class RemoteInstanceConnector(BaseConnector):
+
+    def __init__(self, url: str, device: torch.device = "cpu"):
+        assert (
+            device.type == "cuda"
+        ), "RemoteInstanceConnector only supports cuda device."
+        super().__init__(url)
+        self.url = url
+        self.device = device
+
+    def build_group(
+        self,
+        gpu_id: int = -1,
+        tp_rank: int = -1,
+        instance_ip: str = None,
+        group_rank: int = 1,
+        world_size: int = 2,
+    ):
+        assert (
+            self.device.type == "cuda"
+        ), "RemoteInstanceConnector only supports cuda device."
+        assert (
+            gpu_id != -1 and tp_rank != -1
+        ), "gpu_id and tp_rank must be specified for RemoteInstanceConnector. "
+
+        self.device_id = torch.device(self.device.type, gpu_id)
+
+        parsed_url = urlparse(self.url)
+        master_address = parsed_url.hostname
+        master_port = parsed_url.port
+        group_name = f"send_weights_{instance_ip}_{master_port}_{tp_rank}"
+        backend = "nccl"
+
+        logger.info(
+            f"init custom process group: master_address={master_address}, master_port={master_port}, "
+            f"rank_offset={group_rank}, world_size={world_size}, group_name={group_name}, backend={backend}"
+        )
+
+        try:
+            self._model_update_group = init_custom_process_group(
+                backend=backend,
+                init_method=f"tcp://{master_address}:{master_port}",
+                world_size=world_size,
+                rank=group_rank,
+                group_name=group_name,
+                device_id=self.device_id,
+            )
+            dist.barrier(group=self._model_update_group)
+            return True, "Succeeded to initialize custom process group."
+        except Exception as e:
+            message = f"Failed to initialize custom process group: {e}."
+            logger.error(message)
+            return False, message
+
+    # Implemented as a no-op to make BaseConnector interface consistent.
+    def pull_files(
+        self,
+        allow_pattern: Optional[list[str]] = None,
+        ignore_pattern: Optional[list[str]] = None,
+    ) -> None:
+        return
+
+    # Implemented as a no-op to make BaseConnector interface consistent.
+    def weight_iterator(
+        self, rank: int = 0
+    ) -> Generator[Tuple[str, torch.Tensor], None, None]:
+        return
--- a/python/sglang/srt/managers/io_struct.py
+++ b/python/sglang/srt/managers/io_struct.py
@@ -1021,6 +1021,44 @@ class UpdateWeightsFromTensorReqOutput:
 
 
 @dataclass
+class InitWeightsSendGroupForRemoteInstanceReqInput:
+    # The master address
+    master_address: str
+    # The ports for each rank's communication group
+    ports: str
+    # The rank in the communication group
+    group_rank: int
+    # The world size
+    world_size: int
+    # The group name
+    group_name: str = "weight_send_group"
+    # The backend
+    backend: str = "nccl"
+
+
+@dataclass
+class InitWeightsSendGroupForRemoteInstanceReqOutput:
+    success: bool
+    message: str
+
+
+@dataclass
+class SendWeightsToRemoteInstanceReqInput:
+    # The master address
+    master_address: str
+    # The ports for each rank's communication group
+    ports: str
+    # The group name
+    group_name: str = "weight_send_group"
+
+
+@dataclass
+class SendWeightsToRemoteInstanceReqOutput:
+    success: bool
+    message: str
+
+
+@dataclass
 class InitWeightsUpdateGroupReqInput:
     # The master address
     master_address: str
--- a/python/sglang/srt/model_executor/model_runner.py
+++ b/python/sglang/srt/model_executor/model_runner.py
@@ -19,18 +19,23 @@ import inspect
 import json
 import logging
 import os
+import socket
+import threading
 import time
 from collections import defaultdict
 from dataclasses import dataclass
 from typing import List, Optional, Tuple, Union
+from urllib.parse import urlparse
 
+import requests
 import torch
 import torch.distributed as dist
 
 from sglang.srt.configs.device_config import DeviceConfig
-from sglang.srt.configs.load_config import LoadConfig
+from sglang.srt.configs.load_config import LoadConfig, LoadFormat
 from sglang.srt.configs.model_config import AttentionArch, ModelConfig
 from sglang.srt.configs.update_config import adjust_config_with_unaligned_cpu_tp
+from sglang.srt.connector import ConnectorType
 from sglang.srt.constants import GPU_MEMORY_TYPE_WEIGHTS
 from sglang.srt.distributed import (
     get_pp_group,
@@ -104,6 +109,9 @@ from sglang.srt.offloader import (
     set_offloader,
 )
 from sglang.srt.patch_torch import monkey_patch_torch_reductions
+from sglang.srt.remote_instance_weight_loader_utils import (
+    trigger_init_weights_send_group_for_remote_instance_request,
+)
 from sglang.srt.sampling.sampling_batch_info import SamplingBatchInfo
 from sglang.srt.server_args import ServerArgs
 from sglang.srt.speculative.spec_info import SpeculativeAlgorithm
@@ -126,6 +134,7 @@ from sglang.srt.utils import (
     is_sm100_supported,
     monkey_patch_p2p_access_check,
     monkey_patch_vllm_gguf_config,
+    parse_connector_type,
     set_cuda_arch,
 )
 from sglang.srt.weight_sync.tensor_bucket import (
@@ -254,6 +263,7 @@ class ModelRunner:
 
         # For weight updates
         self._model_update_group = {}
+        self._weights_send_group = {}
 
     def initialize(self, min_per_gpu_memory: float):
         server_args = self.server_args
@@ -704,6 +714,20 @@ class ModelRunner:
         if self.server_args.load_format == "gguf":
             monkey_patch_vllm_gguf_config()
 
+        if self.server_args.load_format == LoadFormat.REMOTE_INSTANCE:
+            if self.tp_rank == 0:
+                instance_ip = socket.gethostbyname(socket.gethostname())
+                t = threading.Thread(
+                    target=trigger_init_weights_send_group_for_remote_instance_request,
+                    args=(
+                        self.server_args.remote_instance_weight_loader_seed_instance_ip,
+                        self.server_args.remote_instance_weight_loader_seed_instance_service_port,
+                        self.server_args.remote_instance_weight_loader_send_weights_group_ports,
+                        instance_ip,
+                    ),
+                )
+                t.start()
+
         # Load the model
         # Remove monkey_patch when linear.py quant remove dependencies with vllm
         monkey_patch_vllm_parallel_state()
@@ -713,7 +737,7 @@ class ModelRunner:
             self.model = get_model(
                 model_config=self.model_config,
                 load_config=self.load_config,
-                device_config=DeviceConfig(self.device),
+                device_config=DeviceConfig(self.device, self.gpu_id),
             )
         monkey_patch_vllm_parallel_state(reverse=True)
         monkey_patch_isinstance_for_vllm_base_layer(reverse=True)
@@ -844,6 +868,103 @@ class ModelRunner:
 
         logger.info("Update weights end.")
         return True, "Succeeded to update model weights."
+
+    def init_weights_send_group_for_remote_instance(
+        self,
+        master_address,
+        ports,
+        group_rank,
+        world_size,
+        group_name,
+        backend="nccl",
+    ):
+        assert (
+            torch.distributed.is_initialized()
+        ), "Default torch process group must be initialized"
+        assert group_name != "", "Group name cannot be empty"
+
+        ports_list = ports.split(",")
+        assert (
+            len(ports_list) == self.tp_size
+        ), f"Expected {self.tp_size} ports, but got {len(ports_list)} ports."
+        group_port = ports_list[self.tp_rank]
+        group_name = f"{group_name}_{group_port}_{self.tp_rank}"
+
+        logger.info(
+            f"init custom process group: tp_rank={self.tp_rank}, gpu_id={self.gpu_id}, master_address={master_address}, master_port={group_port}, "
+            f"group_rank={group_rank}, world_size={world_size}, group_name={group_name}, backend={backend}"
+        )
+
+        torch.cuda.empty_cache()
+        success = False
+        message = ""
+        try:
+            self._weights_send_group[group_name] = init_custom_process_group(
+                backend=backend,
+                init_method=f"tcp://{master_address}:{group_port}",
+                world_size=world_size,
+                rank=group_rank,
+                group_name=group_name,
+                device_id=torch.device("cuda", self.gpu_id),
+            )
+            dist.barrier(group=self._weights_send_group[group_name])
+            success = True
+            message = (
+                f"Succeeded to init group through {master_address}:{group_port} group."
+            )
+        except Exception as e:
+            message = f"Failed to init group: {e}."
+            logger.error(message)
+
+        torch.cuda.empty_cache()
+        return success, message
+
+    def send_weights_to_remote_instance(
+        self,
+        master_address,
+        ports,
+        group_name,
+    ):
+        assert (
+            torch.distributed.is_initialized()
+        ), "Default torch process group must be initialized"
+        assert group_name != "", "Group name cannot be empty"
+
+        ports_list = ports.split(",")
+        assert (
+            len(ports_list) == self.tp_size
+        ), f"Expected {self.tp_size} ports, but got {len(ports_list)} ports."
+        group_port = ports_list[self.tp_rank]
+        group_name = f"{group_name}_{group_port}_{self.tp_rank}"
+
+        if self._weights_send_group[group_name] is not None:
+            send_group = self._weights_send_group[group_name]
+        else:
+            message = f"Group {group_name} not in _weights_send_group list. Please call `init_weights_send_group_for_remote_instance` first."
+            logger.error(message)
+            return False, message
+
+        torch.cuda.empty_cache()
+        success = False
+        message = ""
+        try:
+            for _, weights in self.model.named_parameters():
+                torch.distributed.broadcast(
+                    weights,
+                    src=0,
+                    group=send_group,
+                )
+            success = True
+            message = f"Succeeded to send weights through {master_address}:{group_port} {group_name}."
+        except Exception as e:
+            message = f"Failed to send weights: {e}."
+            logger.error(message)
+
+        # destroy the process group after sending weights
+        del self._weights_send_group[group_name]
+        torch.distributed.distributed_c10d.destroy_process_group(send_group)
+        torch.cuda.empty_cache()
+        return success, message
 
     def init_weights_update_group(
         self,
--- a/python/sglang/srt/model_loader/loader.py
+++ b/python/sglang/srt/model_loader/loader.py
@@ -12,6 +12,9 @@ import json
 import logging
 import math
 import os
+import re
+import socket
+import threading
 import time
 from abc import ABC, abstractmethod
 from concurrent.futures import ThreadPoolExecutor
@@ -27,9 +30,11 @@ from typing import (
     Tuple,
     cast,
 )
+from urllib.parse import urlparse
 
 import huggingface_hub
 import numpy as np
+import requests
 import safetensors.torch
 import torch
 from huggingface_hub import HfApi, hf_hub_download
@@ -56,6 +61,7 @@ from sglang.srt.model_loader.utils import (
 )
 from sglang.srt.model_loader.weight_utils import (
     _BAR_FORMAT,
+    default_weight_loader,
     download_safetensors_index_file_from_hf,
     download_weights_from_hf,
     filter_duplicate_safetensors_files,
@@ -70,6 +76,9 @@ from sglang.srt.model_loader.weight_utils import (
     pt_weights_iterator,
     safetensors_weights_iterator,
     set_runai_streamer_env,
+)
+from sglang.srt.remote_instance_weight_loader_utils import (
+    trigger_transferring_weights_request,
 )
 from sglang.srt.utils import (
     get_bool_env_var,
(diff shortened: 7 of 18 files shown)


zhaochenyang20:
@amysaq2023 Shall find someone from the community to review this

stmatengss:
> @amysaq2023 Shall find someone from the community to review this

@zhaochenyang20 If no one else can get to it, I’m happy to review this PR

ryang-max:
wondering what's the use case of this new load format in practice?

zhaochenyang20:
> > @amysaq2023 Shall find someone from the community to review this
> 
> @zhaochenyang20 If no one else can get to it, I’m happy to review this PR

Please feel free to review

zhaochenyang20:
```bash
python -m sglang.launch_server
--model-path instance://[target_instance_ip]:[communication_group_port] 
--tokenizer-path [local_tokenizer_path]
--seed-instance-url http://[target_instance_ip]:[target_instance_serving_port]
--load-format remote_instance
--client-instance-id [local_instance_id](optional)
```

I think the launching command is quite strange. Why `--model-path instance://[target_instance_ip]:[communication_group_port] `? Is this a mistake? Maybe the `model_path` shall be:

`--model-path instance://[new instance ip]:[new instance port]`·

And, what's the `local_instance_id` here means?

python/sglang/srt/configs/load_config.py
```diff
@@ -50,6 +51,8 @@ class LoadConfig:
 
     load_format: Union[str, LoadFormat] = LoadFormat.AUTO
     download_dir: Optional[str] = None
+    seed_instance_url: Optional[str] = None
+    client_instance_id: Optional[str] = None
```
zhaochenyang20: what's the client here mean?

python/sglang/srt/connector/remote_instance.py
```diff
@@ -0,0 +1,117 @@
...
+        assert(
+            device.type == "cuda"
+        ), "RemoteInstanceConnector only supports cuda device."
+        super().__init__(url, device)
+        self.url = url
+
+    def build_group(self, gpu_id: int = -1, tp_rank: int = -1, client_instance_id: str = None, group_rank: int = 1, world_size: int = 2):
+        assert(
+            self.device.type == "cuda"
+        ), "RemoteInstanceConnector only supports cuda device."
```
zhaochenyang20: will double check wether this will affect AMD CI.

python/sglang/srt/connector/remote_instance.py
```diff
@@ -0,0 +1,117 @@
...
+                world_size=world_size,
+                rank=rank_offset,
+                group_name=group_name,
+                device_id=device_id,
+            )
+            return True, "Succeeded to initialize custom process group."
+        except Exception as e:
+            message = f"Failed to initialize custom process group: {e}."
+            logger.error(message)
+            return False, message
```
zhaochenyang20: shall we just use `init_custom_process_group` instead of wrap another layer of API?

zhaochenyang20: shall we just use `init_custom_process_group` instead of wrap another layer of API?

python/sglang/srt/connector/remote_instance.py
```diff
@@ -0,0 +1,117 @@
...
+        self,
+        allow_pattern: Optional[list[str]] = None,
+        ignore_pattern: Optional[list[str]] = None,
+    ) -> None:
+        return
+
+    def weight_iterator(
+        self, rank: int = 0
+    ) -> Generator[Tuple[str, torch.Tensor], None, None]:
+        return
```
zhaochenyang20: why there is empty functions here?

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -1028,6 +1046,32 @@ async def init_weights_update_group(
...
+    async def init_weights_send_group_for_remote_instance(
+        self,
+        obj: InitWeightsSendGroupForRemoteInstanceReqInput,
+        request: Optional[fastapi.Request] = None,
+    ) -> Tuple[bool, str]:
+        self.auto_create_handle_loop()
+        # TODO: support DP
+        assert (
+            self.server_args.dp_size == 1
+        ), "dp_size must be 1 for init_weights_send_group_for_remote_instance"
```
zhaochenyang20: so you can just create new instances in dp size 1 right? Specifically, create a new dp replica from a previous dp replica?

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -762,6 +793,77 @@ def model_load_weights(model, iter):
         logger.info("Update weights end.")
         return True, "Succeeded to update model weights."
 
+    def init_weights_send_group_for_remote_instance(
```
zhaochenyang20, line 872: I'd like to name it as `init_weights_send_group_for_remote_instance`, but feel free to go on your side.

python/sglang/srt/model_loader/loader.py
```diff
@@ -1342,6 +1347,109 @@ def load_model(
...
+                    self.load_model_from_remote_engine(
+                        model, client, model_config, device_config)
+                else:
+                    raise ValueError(
+                        f"Unsupported connector type {connector_type} for "
+                        f"remote tensor model loading."
+                    )
+        return model.eval()
+
+    def load_model_from_remote_engine(
```
zhaochenyang20: better to use `load_model_from_remote_instance`

python/sglang/srt/server_args.py
```diff
@@ -603,6 +621,24 @@ def add_cli_args(parser: argparse.ArgumentParser):
...
+            "--seed-instance-url",
+            type=str,
+            default=ServerArgs.seed_instance_url,
+            help="The url of the seed instance for loading weights from remote instance.",
+        )
+        parser.add_argument(
+            "--client-instance-id",
+            type=str,
+            default=ServerArgs.client_instance_id,
+            help="The id of the client instance for loading weights from remote instance.",
```
zhaochenyang20: Also, I do not understand why there is a client here.

zhaochenyang20 (changes requested):
Generally good PR. A 2-gpu unit test is needed.

python/sglang/srt/configs/load_config.py
amysaq2023: client here means the new instance who wants to load weights from the target(seed) instance. Feels like it may be confusing. How about changing to dst_instance?

zhaochenyang20: `dst_instance` is good to go!

python/sglang/srt/connector/remote_instance.py
```diff
@@ -0,0 +1,117 @@
...
+    def init_load_weights_group(
+        self,
+        master_address,
+        master_port,
+        rank_offset,
+        world_size,
+        group_name,
+        backend="nccl",
+        device_id=None,
+    ):
```
zhaochenyang20: refact this function.

python/sglang/srt/connector/remote_instance.py
zhaochenyang20: Add comments that just implement the ABC function.

python/sglang/srt/managers/io_struct.py
```diff
@@ -837,6 +837,39 @@ class UpdateWeightsFromTensorReqOutput:
     success: bool
     message: str
 
+@dataclass
+class InitWeightsSendGroupForRemoteInstanceReqInput:
```
zhaochenyang20, line 1024: destination send `InitWeightsSendGroupForRemoteInstanceReq` to seed, and seed send `SendWeightsToRemoteInstance` to destination.

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -623,6 +630,28 @@ def load_model(self):
         if self.server_args.load_format == "gguf":
             monkey_patch_vllm_gguf_config()
 
+        if self.server_args.load_format == LoadFormat.REMOTE_INSTANCE:
+            def trigger_seed():
```
zhaochenyang20: maybe change the name of `trigger_seed`, but the design is clear.

python/sglang/srt/model_loader/loader.py
```diff
@@ -1342,6 +1347,109 @@ def load_model(
...
+        model_weights = model_config.model_path
+
+        if hasattr(model_config, "model_weights"):
+            model_weights = model_config.model_weights
+
+        with set_default_torch_dtype(model_config.dtype):
+            with torch.device(device_config.device):
+                model = _initialize_model(model_config, self.load_config)
+
+            def update_connector_port(url: str, rank: int) -> str:
```
zhaochenyang20: shall we make the names clearer?

python/sglang/srt/model_loader/loader.py
```diff
@@ -1342,6 +1347,109 @@ def load_model(
...
+    ) -> nn.Module:
+        load_config = self.load_config
+
+        client.build_group(gpu_id=device_config.gpu_id, tp_rank=device_config.tp_rank, client_instance_id=load_config.client_instance_id)
+
+        model_weights = model_config.model_path
+        if hasattr(model_config, "model_weights"):
+            model_weights = model_config.model_weights
+
+        def trigger_seed():
```
zhaochenyang20: also change the name

python/sglang/srt/server_args.py
zhaochenyang20: change client to dst

zhaochenyang20 (changes requested):
Adding unit tests:

https://github.com/sgl-project/sglang/blob/main/test/srt/test_update_weights_from_distributed.py

zhaochenyang20:
`if is_in_ci():` use 2-gpu to test, tp1 server init a tp1 server.
if testing locally, shall test use a tp2 server to init another tp2 server.

```python
        if is_in_ci():
            mode = random.choice(["Engine", "Server"])
            test_suits = [
                (1, 1, DEFAULT_SMALL_MODEL_NAME_FOR_TEST, mode),
            ]
        else:
            test_suits = [
                (1, 1, DEFAULT_SMALL_MODEL_NAME_FOR_TEST, "Engine"),
                (1, 1, DEFAULT_MODEL_NAME_FOR_TEST, "Sever"),
            ]

            if torch.cuda.device_count() >= 4:
                test_suits.extend(
                    [
                        (2, 1, DEFAULT_SMALL_MODEL_NAME_FOR_TEST, "Engine"),
                        (1, 2, DEFAULT_MODEL_NAME_FOR_TEST, "Server"),
                    ]
                )

            if torch.cuda.device_count() >= 5:
                test_suits.extend(
                    [
                        (2, 2, DEFAULT_SMALL_MODEL_NAME_FOR_TEST, "Engine"),
                        (2, 2, DEFAULT_MODEL_NAME_FOR_TEST, "Server"),
                    ]
                )

```

zhaochenyang20:
Also, add the unit tests file's name into `run_suite`.

amysaq2023:
> `if is_in_ci():` use 2-gpu to test, tp1 server init a tp1 server. if testing locally, shall test use a tp2 server to init another tp2 server.
> 
> ```python
>         if is_in_ci():
>             mode = random.choice(["Engine", "Server"])
>             test_suits = [
>                 (1, 1, DEFAULT_SMALL_MODEL_NAME_FOR_TEST, mode),
>             ]
>         else:
>             test_suits = [
>                 (1, 1, DEFAULT_SMALL_MODEL_NAME_FOR_TEST, "Engine"),
>                 (1, 1, DEFAULT_MODEL_NAME_FOR_TEST, "Sever"),
>             ]
> 
>             if torch.cuda.device_count() >= 4:
>                 test_suits.extend(
>                     [
>                         (2, 1, DEFAULT_SMALL_MODEL_NAME_FOR_TEST, "Engine"),
>                         (1, 2, DEFAULT_MODEL_NAME_FOR_TEST, "Server"),
>                     ]
>                 )
> 
>             if torch.cuda.device_count() >= 5:
>                 test_suits.extend(
>                     [
>                         (2, 2, DEFAULT_SMALL_MODEL_NAME_FOR_TEST, "Engine"),
>                         (2, 2, DEFAULT_MODEL_NAME_FOR_TEST, "Server"),
>                     ]
>                 )
> ```

Thanks for advice! We have added a unit test for loading weights from remote instance.

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -356,6 +360,12 @@ def __init__(
         self.init_weights_update_group_communicator = _Communicator(
```
zhaochenyang20: could you help me to change this into init_update_weights_group_communicator

The name should follow the update_weights_from_distributed_communicator

Also, put it as:


        self.init_weights_update_group_communicator = _Communicator(
Comment


could you help me to change this into init_update_weights_group_communicator

The name should follow the update_weights_from_distributed_communicator

Also, put it as:

        self.init_update_weights_group_communicator = _Communicator(
            self.send_to_scheduler, server_args.dp_size
        )
       self.update_weights_from_distributed_communicator = _Communicator(
            self.send_to_scheduler, server_args.dp_size
        )
        self.init_send_weights_group_for_remote_instance_communicator = _Communicator(
            self.send_to_scheduler, server_args.dp_size
        )
        self.send_weights_to_remote_instance_communicator = _Communicator(
            self.send_to_scheduler, server_args.dp_size
        )

zhaochenyang20: rename   init_weights_send_group_for_remote_instance_communicator to init_send_weights_group_for_remote_instance_communicator

send before weights

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -1050,6 +1068,32 @@ async def init_weights_update_group(
         result = (await self.init_weights_update_group_communicator(obj))[0]
         return result.success, result.message
 
+    async def init_weights_send_group_for_remote_instance(
```
zhaochenyang20: send weights group

test/srt/test_load_weights_from_remote_instance.py
```diff
@@ -0,0 +1,359 @@
...
+    event_seed_ready,
+    event_dst_ready,
+):
+    torch.cuda.set_device(rank)
+    torch.cuda.synchronize()
+    base_gpu_id = 1 if rank == 1 else 1 + tp_size
+
+    event_seed_ready.wait()
+
+    if backend == "Engine":
```
zhaochenyang20, line 178: I only see you changed the codes of https_server. Does this feature also support running on the engine? If not, please delete this and demonstrate that this feature only supports the server at the beginning of this file.

zhaochenyang20: Oh. Sorry, seed should be a server, but destination can be an engine or server?

zhaochenyang20: Nice done!

zhaochenyang20: Could you demonstrate this at the beginning? " seed should be a server, but destination can be an engine or server"

Also, just pondering that if this can also be made on the engine side, why isn't the engine code changed?

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -662,6 +669,28 @@ def load_model(self):
...
+                url = self.server_args.seed_instance_url
+                parsed_group_url = urlparse(self.server_args.model_path)
+                group_port = int(parsed_group_url.port)
+                requests.post(
+                    f"{url}/init_weights_send_group_for_remote_instance",
+                    json={
+                        "master_address": parsed_group_url.hostname,
+                        "master_port": group_port,
+                        "group_rank": 0,
+                        "world_size": 2,
```
zhaochenyang20: if the tp size is larger than 1, is the world_size of 2 still valid?

test/srt/test_load_weights_from_remote_instance.py
```diff
@@ -0,0 +1,359 @@
...
+        "seed": results.get("seed_params"),
+        "sgl_dp_1_dest": results.get("sgl_dp_1_dst_params"),
+    }
+
+    if dp_size == 2:
+        dp2_params = {
+            "sgl_dp_2_dest": results.get("sgl_dp_2_dst_params"),
+        }
+        assert all(v is not None for v in dp2_params.values())
+        params.update(dp2_params)
```
zhaochenyang20, line 310: In this test, you initialize a server as the seed at rank 0, then for 1/2 new instances, you want to create it from rank 0.

On rank 1 or rank 2, they run exactly the same code with dp 1 in their process? That is to say, we can not init a SGLang server with dp 2 on this. But we create 2 server with dp 1 and get the weights from the seed?

        assert (
            self.server_args.dp_size == 1
        ), "dp_size must be 1 for init_weights_send_group_for_remote_instance"

test/srt/test_load_weights_from_remote_instance.py
```diff
@@ -0,0 +1,359 @@
...
+        if is_in_ci():
+            mode = random.choice(["Engine", "Server"])
+            test_suits = [
+                (1, 1, DEFAULT_SMALL_MODEL_NAME_FOR_TEST, mode, "ci_test"),
+            ]
+        else:
+            test_suits = [
+                (1, 1, DEFAULT_SMALL_MODEL_NAME_FOR_TEST, "Engine", "engine_test"),
+                (1, 1, DEFAULT_SMALL_MODEL_NAME_FOR_TEST, "Sever", "server_test"),
+            ]
```
zhaochenyang20: shall add more tests:

### On the CI

Rank 0 hosts the seed server and sends it to rank 1 for a small model engine/server. (randomly choosing server or engine for the test on CI). This is only a 2 gpu test.

### Off the CI

1. Rank 0 hosts the seed server and sends it to rank 1 for a small model engine and server. (Test both server and engine offline).
2. Use TP 2 to host the seed server, and send it to ranks 2 and 4 for a small model engine and server. (seed TP2, and having two destinations, both with TP 2, and the destination should be server and engine). This is a 6-GPU test.
3. Use TP 1 to host the seed server. For the destinations, one is an engine and the other is a server; initialize them at the same time. Just want to make sure that the engine and server do not conflict. This is a 3-GPU test. I am pondering wether this is not reasonable. If not, please let me know.

zhaochenyang20 (changes requested):

zhaochenyang20:
Also, adds the tests to test/srt/run_suite.py @amysaq2023

zhaochenyang20:
Fix lint as well.

python/sglang/srt/managers/tokenizer_manager.py
amysaq2023: Sure thing! Besides these communicators, do we also need to refactor all the function names including init_weights_update_group prefix? Such as those in tokernizer_manager, http_server etc.
```python
@app.post("/init_weights_update_group")
async def init_weights_update_group(
    obj: InitWeightsUpdateGroupReqInput, request: Request
):
```

amysaq2023: Besides, do we need to refactor these request input/output classes as well?
```python
@dataclass
class InitWeightsUpdateGroupReqInput:

@dataclass
class InitWeightsUpdateGroupReqOutput:

```

python/sglang/srt/model_executor/model_runner.py
amysaq2023: Since we build up communication group per tp rank, which means, the seed instance tp rank 0 and the dst instance tp rank 0 are in one communication group, while the seed instance tp rank 1 and the dst instance tp rank 1 are in another communication group, the world size here for sending weights is always 2.
(We currently do not support concurrently sending weights to multiple dst instances by one single seed.)

amysaq2023:
> Also, adds the tests to test/srt/run_suite.py @amysaq2023

Thanks for pointing out. Already added the new unit test to run_suite.

test/srt/test_load_weights_from_remote_instance.py
amysaq2023: For "Off the CI" tests, I got the first two cases, but the third one is a little confusing. Currently, we do not support concurrently sending weights to multiple destination instances. The third case will end up with one dst instance start earlier than the other, so I don't think they have the chance to disturb each other. Or probably I did not get the purpose of the third case correctly? Could you please explain more about this case? Thanks a lot.

python/sglang/srt/managers/tokenizer_manager.py
zhaochenyang20: Quite embarrassing. I also think this is a hard problem. In this case, I think no need to change the naming of it. qvq

zhaochenyang20: At least, we shall change the order of these four lines.

Let the weights update group 2 line go first, then for the send weights group 2 line go

python/sglang/srt/model_executor/model_runner.py
zhaochenyang20: Cool. Could you add your feedback as a comment to the `world_size`.

test/srt/test_load_weights_from_remote_instance.py
zhaochenyang20: Let's have this offline. thanks!

python/sglang/srt/managers/tokenizer_manager.py
zhaochenyang20: change the order

python/sglang/srt/model_executor/model_runner.py
zhaochenyang20: Add a comment to say that only support model with same parallelism strategy. TP 8 to TP8. And will have 8 process groups, "tp 1 -> tp 1", "tp2 -> tp2"

python/sglang/srt/model_loader/loader.py
```diff
@@ -1354,6 +1359,109 @@ def load_model(
...
+    def __init__(self, load_config: LoadConfig):
+        super().__init__(load_config)
+        if load_config.model_loader_extra_config:
+            raise ValueError(
+                f"Model loader extra config is not supported for "
+                f"load format {load_config.load_format}"
+            )
+
+    def download_model(self, model_config: ModelConfig) -> None:
+        pass  # Nothing to download
```
zhaochenyang20: raise something. if someone go into this, raise an error, this is not expected.

python/sglang/srt/server_args.py
```diff
@@ -689,6 +705,24 @@ def add_cli_args(parser: argparse.ArgumentParser):
             help="The path of the model weights. This can be a local folder or a Hugging Face repo ID.",
             required=True,
         )
+        parser.add_argument(
+            "--model-config-path",
+            type=str,
+            default=ServerArgs.model_config_path,
```
zhaochenyang20: I think that we should keep the model name always in `model_path`, and add a new parameter for the remote instance url.

zhaochenyang20: better to have a check between these two instance, require them to be the same model.

zhaochenyang20: I guess that we can remove `"--model-config-path"`, only have ` "--seed-instance-url",`

python/sglang/srt/server_args.py
```diff
@@ -689,6 +705,24 @@ def add_cli_args(parser: argparse.ArgumentParser):
...
+            "--seed-instance-url",
+            type=str,
+            default=ServerArgs.seed_instance_url,
+            help="The url of the seed instance for loading weights from remote instance.",
+        )
+        parser.add_argument(
+            "--dst-instance-id",
+            type=str,
+            default=ServerArgs.dst_instance_id,
+            help="The id of the client instance for loading weights from remote instance.",
```
zhaochenyang20: we can remove this and distinguish the destination model by port. The port is unique.

test/srt/run_suite.py
```diff
@@ -118,6 +118,7 @@ class TestFile:
         TestFile("rl/test_update_weights_from_distributed.py", 103),
         TestFile("test_data_parallelism.py", 73),
         TestFile("test_dp_attention.py", 277),
+        TestFile("test_load_weights_from_remote_instance.py", 120),
```
zhaochenyang20: niche the constraint.

zhaochenyang20 (changes requested):

python/sglang/srt/connector/remote_instance.py
```diff
@@ -0,0 +1,104 @@
...
+            gpu_id != -1 and tp_rank != -1
+        ), "gpu_id and tp_rank must be specified for RemoteInstanceConnector. "
+
+        self.device_id = torch.device(self.device.type, gpu_id)
+
+        parsed_url = urlparse(self.url)
+        master_address = parsed_url.hostname
+        master_port = parsed_url.port
+        group_name = f"send_weights_{instance_ip}_{master_port}_{tp_rank}"
+        backend = "nccl"
```
stmatengss, line 47: Should the nccl backend be configurable via a new parameter?

python/sglang/srt/server_args.py
```diff
@@ -821,6 +838,24 @@ def add_cli_args(parser: argparse.ArgumentParser):
...
+            type=str,
+            default=ServerArgs.seed_instance_ip,
+            help="The ip of the seed instance for loading weights from remote instance.",
+        )
+        parser.add_argument(
+            "--seed-instance-service-port",
+            type=int,
+            default=ServerArgs.seed_instance_service_port,
+            help="The service port of the seed instance for loading weights from remote instance.",
+        )
```
stmatengss: To distinguish these arguments, could you add a remote-instance prefix?

python/sglang/srt/connector/remote_instance.py
zhaochenyang20: reasonable suggestion!

zhaochenyang20 (changes requested):

zhaochenyang20 (approved):

stmatengss:
We evaluate this PR with Qwen3-32B, and get the following performance results:
```
python -m sglang.launch_server  --model /root/models/Qwen3-32B --port 30000 --host 0.0.0.0 --tensor-parallel-size 8 --dtype auto
python -m sglang.launch_server --model-path instance://192.168.xx.xx:29400 --tokenizer-path /root/models/Qwen3-32B --seed-instance-url http://192.168.xx.xx:30000 --load-format remote_instance --tensor-parallel-size 8 --port 30001 --host 0.0.0.0
```
The bandwidth is 200Gb with 2 x 100Gb erdma NIC.
Loading weights from files on the server takes 5 minutes and 46 seconds, while loading weights from the server to the client takes only 4 seconds.

stmatengss:
Great work; this looks promising for large-scale deployment. I have two questions:

1.  Will client-side weight loading impact the performance of the remote instance (e.g., high TTFT)?
2.  What is the multi-client support like? Can multiple clients load weights from the same remote instance, which would be useful for bulk starts?

zxpdemonio:
For clustered deployments, multi‑node weight update efficiency needs attention. Concurrent transfer mechanisms and P2P weight distribution strategies may help.

python/sglang/srt/server_args.py
```diff
@@ -126,6 +131,9 @@ def add_disagg_transfer_backend_choices(choices):
 class ServerArgs:
     # Model and tokenizer
     model_path: str
```
hnyls2002: Suggestion: I think a common prefix like `--remote-instance-weight-loader-xxx` should be applied here, as all the newly added arguments are remote weight loader specific.

python/sglang/srt/server_args.py
hnyls2002: Do not worry about a long name, `--remote-instance-weight-loader` is clearer

python/sglang/srt/model_loader/loader.py
hnyls2002: @amysaq2023

python/sglang/srt/configs/model_config.py
```diff
@@ -64,12 +64,18 @@ def __init__(
...
+        seed_instance_ip: Optional[str] = None,
+        seed_instance_service_port: Optional[int] = None,
+        send_weights_group_ports: Optional[List[int]] = None,
     ) -> None:
         # Parse args
         self.model_path = model_path
         self.revision = revision
         self.quantization = quantization
         self.model_impl = model_impl
+        self.seed_instance_ip = seed_instance_ip
```
hnyls2002: Maybe better to introduce a new config class here. (You can probably add it in later PRs)

python/sglang/srt/configs/device_config.py
```diff
@@ -8,10 +8,16 @@
 
 class DeviceConfig:
     device: Optional[torch.device]
+    tp_rank: Optional[int]
+    gpu_id: Optional[int]
 
-    def __init__(self, device: str = "cuda") -> None:
+    def __init__(
+        self, device: str = "cuda", tp_rank: int = -1, gpu_id: int = -1
```
hnyls2002: I there any method to remove `tp_rank` from device config? Device's information should be isolated from model's parallelism infos.

python/sglang/srt/model_loader/loader.py
```diff
@@ -1366,6 +1372,116 @@ def load_model(
...
+        end_build_group_tic = time.time()
+        logger.debug(
+            f"finish building group for remote instance, time used: {(end_build_group_tic - start_build_group_tic):.4f}s"
+        )
+
+        def trigger_send_weights_to_remote_instance_request():
+            seed_instance_service_url = f"http://{model_config.seed_instance_ip}:{model_config.seed_instance_service_port}"
+            try:
+                requests.post(
+                    f"{seed_instance_service_url}/send_weights_to_remote_instance",
```
hnyls2002: Confusing, this is the new instance that wants to **get weights** from the remote instance. You can rename it to `trigger_sending_weights_xxx` or `get_weights_xxx`

hnyls2002 (changes requested):

hnyls2002:
@amysaq2023  By the way, just curious about the current implementation of two calls, one is initializing a sending group, and one is triggering sending action. Why are two calls needed here?

python/sglang/srt/configs/device_config.py
merrymercy: +1， remove these

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -681,6 +688,48 @@ def load_model(self):
         if self.server_args.load_format == "gguf":
             monkey_patch_vllm_gguf_config()
 
+        if self.server_args.load_format == LoadFormat.REMOTE_INSTANCE:
+
+            def trigger_init_weights_send_group_for_remote_instance_request():
```
merrymercy: Do not write long functions inside the function. Move this into a separate function.

python/sglang/srt/server_args.py
```diff
@@ -126,6 +131,9 @@ def add_disagg_transfer_backend_choices(choices):
...
+    seed_instance_ip: Optional[str] = None
+    seed_instance_service_port: Optional[int] = None
+    send_weights_group_ports: Optional[List[int]] = None
```
merrymercy: move this to the bottom (after custom_weight_loader)

merrymercy (changes requested):

amysaq2023:
> @amysaq2023 By the way, just curious about the current implementation of two calls, one is initializing a sending group, and one is triggering sending action. Why are two calls needed here?

Getting the idea from how update_weights_from_distributed implemented, we also separate loading weights from remote instance into two steps. Beside, we are going to add P2P port negotiation between seed and destination instances during initializing group (just in the coming PR ), this action needs to be done before loading weights since the destination instance will need to know from the seed instance what ports to use.

python/sglang/srt/configs/model_config.py
amysaq2023: Thanks for advice. We will refactor this in the next PR.

python/sglang/srt/remote_instance_weight_loader_mixin.py
```diff
@@ -0,0 +1,69 @@
...
+
+import logging
+from typing import List
+
+import requests
+
+logger = logging.getLogger(__name__)
+
+
+def trigger_init_weights_send_group_for_remote_instance_request(
```
hnyls2002: This is not a mixin class; it is a util

hnyls2002 (approved):

stmatengss:
This PR adds new server arguments and FastAPI endpoints. In my opinion, the documentation should also be updated.

stmatengss (approved):

merrymercy:
Please address the following comments
- move `remote_instance_weight_loader_utils.py` under `model_loader`
- move `tp_rank`, `remote_instance_weight_loader_seed_instance_ip`, ... out of `ModelConfig` and to `LoadConfig`
- move `parser.add_argument("--remote-instance-weight-loader-seed-instance-ip",` after `parser.add_argument("--weight-loader-disable-mmap",`. The order of parser.add_argument should be the same as the order of how they are defined in `class ServerArgs:`

amysaq2023:
> Please address the following comments
> 
> * move `remote_instance_weight_loader_utils.py` under `model_loader`
> * move `tp_rank`, `remote_instance_weight_loader_seed_instance_ip`, ... out of `ModelConfig` and to `LoadConfig`

Thanks for suggestion :) We have updated the format in new PR: https://github.com/sgl-project/sglang/pull/10941

> * move `parser.add_argument("--remote-instance-weight-loader-seed-instance-ip",` after `parser.add_argument("--weight-loader-disable-mmap",`. The order of parser.add_argument should be the same as the order of how they are defined in `class ServerArgs:`

I think in the original version, the order has already been followed. Please let me know if I understood this comment wrong. Thanks

merrymercy:
thanks!

PR #8243, opened 2025-07-22
[CPU] Fix TP padding case with weight block size

jianan-gu:
### Motivation
Fixes the [Kimi-K2-Instruct](https://huggingface.co/moonshotai/Kimi-K2-Instruct) (FP8) TP=6 failure on CPU.

ValueError: Weight output_partition_size = 2112 is not divisible by weight quantization block_n = 128.

### Modifications
Considering `weight_block_size `when padding TP, to make self.num_heads * self.qk_head_dim / tp_size divisible by `weight_block_size `in the below ColumnParallelLinear.
```
  self.q_b_proj = ColumnParallelLinear(
      q_lora_rank,
      self.num_heads * self.qk_head_dim,
      bias=False,
      quant_config=quant_config,
      prefix=add_prefix("q_b_proj", prefix),
      tp_rank=attn_tp_rank,
      tp_size=attn_tp_size,
  )
```
In our case, self.num_heads = 64, self.qk_head_dim =192, tp_size = 6, block_size = 128

Before this PR, self.num_heads is padded to 66, self.num_heads * self.qk_head_dim / tp_size = 2112, which is not divisible by 128.
After this PR, self.num_heads is padded to 72, self.num_heads * self.qk_head_dim / tp_size = 2304, which is divisible by 128.

**MISC.**
This PR also covers a minor fix for unquant MoE module with `apply_router_weight_on_input` config, sine CPU AMX path supports now.


python/sglang/srt/configs/update_config.py
```diff
@@ -41,10 +41,16 @@ def get_moe_padding_size(weight_block_size):
...
+def get_num_heads_padding_size(tp_size, weight_block_size, head_dim):
+    pad_size = tp_size
+
+    if weight_block_size is not None and head_dim % weight_block_size[0] != 0:
+        import math
+
+        pad_size = tp_size * (
+            math.lcm(head_dim, weight_block_size[0]) / weight_block_size[0]
+        )
+
```
mingfeima, line 53: what is the input size here and how much padding do we need to add ?

mingfeima:
I reckon that we still need to move fast for DP-Attn and EPMoE to skip padding for head dimension... not very decent to have a TP=6 for any scenario

python/sglang/srt/configs/update_config.py
jianan-gu: With this PR:
For KIMI k2, we will make num_head pad to 72 for TP=6 (original num_head is 64), total dim will be 72 * 192 (qk_head_dim) for TP 6 
For DeepSeek r1 as reference, we have not changed padding input here, it is still making num_head pad to 132 for TP=6 (original num_head is 128), total dim will be 132 * 192 (qk_head_dim) for TP 6

jianan-gu:
> I reckon that we still need to move fast for DP-Attn and EPMoE to skip padding for head dimension... not very decent to have a TP=6 for any scenario

Yes, agree and noted.

Alcanderian (approved):

Alcanderian (approved):

PR #6092, opened 2025-05-07
[Performance] Use the max num_tokens per DP rank as the CUDA graph batch size

xpmemeda:
## Motivation

With --enable-dp-attention flag, cuda_graph_runner uses the batch size of each DP rank to capture CUDA graphs, but uses the sum of tokens from all DP ranks to select and replay the CUDA graph, which results in selecting an excessively large batch size.

See this [issue 5527](https://github.com/sgl-project/sglang/issues/5527)

## Modifications

Change the CUDA graph selection logic from sum to max (tokens).

## Benchmarks

H20 * 16，python -u -m sglang.launch_server --model-path /sgl-workspace/DeepSeek-R1 --nnodes 2 --trust-remote-code  --served-model-name DeepSeek-R1 --dist-init-addr 29.226.64.238:5000 --node-rank 0 --host 0.0.0.0 --port 8000 --tp 16 --disable-radix-cache   --schedule-policy fcfs --chunked-prefill-size 32768 --disable-overlap-schedule --mem-fraction-static 0.79 --attention-backend flashinfer --enable-dp-attention --dp-size 16

python -m sglang.bench_one_batch_server --model None --base-url http://127.0.0.1:8000 --batch-size 128 --input-len 1000 --output-len 1000

Before
企业微信截图_17466305911505

After
image

## Checklist

- [x] Format your code according to the [Code Formatting with Pre-Commit](https://docs.sglang.ai/references/contribution_guide.html#code-formatting-with-pre-commit).
- [ ] Add unit tests as outlined in the [Running Unit Tests](https://docs.sglang.ai/references/contribution_guide.html#running-unit-tests-adding-to-ci).
- [ ] Update documentation / docstrings / example tutorials as needed, according to [Writing Documentation](https://docs.sglang.ai/references/contribution_guide.html#writing-documentation-running-docs-ci).
- [ ] Provide throughput / latency benchmark results and accuracy evaluation results as needed, according to [Benchmark and Profiling](https://docs.sglang.ai/references/benchmark_and_profiling.html) and [Accuracy Results](https://docs.sglang.ai/references/accuracy_evaluation.html).
- [ ] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co-author when merging the PR.
- [ ] Please feel free to join our Slack channel at https://slack.sglang.ai to discuss your PR.

PR #5557, opened 2025-04-19
[DeepEP] Eliminate unnecessary DP cudagraph padding

yuleil:
## Motivation

In DP attention, the batch size of the CUDA graph is expanded to the sum of batches across all DP ranks to enable `all_gather` global tokens before MLP fwd. Since decoding is memory-bound, this padding does not introduce significant performance overhead. Under DeepEP, the padded tokens from each DP rank participate in dispatch/combine operations, resulting in a DP times increase in communication costs.  

DeepEP's MoE computation adopts a fixed shape with masking instead of gathering tokens from all DP ranks, so padding batch size  to the global tokens is actually unnecessary.

Before this fix:
image

After this fix:
image

With H20, EP16,120 batch decoding with cudagraph, the combine time has been reduced by 10 times, and TPOT has decreased from 300ms to ~100ms.

## Checklist

- [x] Format your code according to the [Code Formatting with Pre-Commit](https://docs.sglang.ai/references/contribution_guide.html#code-formatting-with-pre-commit).
- [ ] Add unit tests as outlined in the [Running Unit Tests](https://docs.sglang.ai/references/contribution_guide.html#running-unit-tests-adding-to-ci).
- [ ] Update documentation / docstrings / example tutorials as needed, according to [Writing Documentation](https://docs.sglang.ai/references/contribution_guide.html#writing-documentation-running-docs-ci).
- [x] Provide throughput / latency benchmark results and accuracy evaluation results as needed, according to [Benchmark and Profiling](https://docs.sglang.ai/references/benchmark_and_profiling.html) and [Accuracy Results](https://docs.sglang.ai/references/accuracy_evaluation.html).
- [ ] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co-author when merging the PR.
- [ ] Please feel free to join our Slack channel at https://slack.sglang.ai to discuss your PR.

PR #4390, opened 2025-03-13
Improve DP attention

merrymercy:
- Use a better padding strategy for cuda graph. If TP=8, DP=8, when batch size = 1, the previous implementation will pad it to global batch size 8. The new implementation will allow running global batch size 1, so it is faster at low bs range.  It is 1.15x faster then the old implementation for `deepseek-ai/DeepSeek-Coder-V2-Lite-Instruct` at TP=8 and bs=1.
- Support TP != DP. Now you need to explicitly specify `--dp` and `--tp`. The constraint is that `--dp` should be smaller than `--tp`. You can first set `--tp` as the number of total GPUs you have, then tune `--dp` to trade-off between latency and KV cache capacity (or throughput).  For example, to achieve better latency for small bs, you can do `--tp 8 --dp 2`. To allow more KV cache capacity for larger bs, you can do `--tp 8 --dp 8`. An example command:
```
python3 -m sglang.launch_server --model-path deepseek-ai/DeepSeek-Coder-V2-Lite-Instruct --trust-remote-code --tp 8 --enable-dp-attention --dp 2
```

99% of the code is done by @dhou-xai .

```
Co-authored-by: dhou-xai <dhou@x.ai>
Co-authored-by: SangBin Cho <rkooo567@gmail.com>
```

PR #8278, opened 2025-07-23
[3/n] DP Enhancement: Padding tokens to max length when workload is balanced

### LINKED ISSUE #5527: Cuda graph supported bs in DP attention
I have been testing and recording the output throughput of SGLang on 2*8 H100 GPUs, and I've observed a significant regression in output throughput for long outputs in the `enable-dp-attention` scenarios following this [PR](https://github.com/sgl-project/sglang/pull/4390). Through debugging and profiling with Nsight Systems, I confirmed that the performance degradation is caused by the CUDA graph not being properly launched.

See the [code](https://github.com/sgl-project/sglang/blob/00391f58a3c2b13b00dc0a0a983a5c7fab399883/python/sglang/srt/model_executor/cuda_graph_runner.py#L280)
```
if self.enable_dp_attention:
    total_global_tokens = sum(forward_batch.global_num_tokens_cpu)

    is_bs_supported = forward_batch.can_run_dp_cuda_graph and (
        total_global_tokens in self.graphs
        if self.disable_padding
        else total_global_tokens <= self.max_bs
    )
```
After `enable-dp-attention`, `total_global_tokens` equals the sum of tokens across all DP ranks. For example, during the decode phase with DP = TP = 16 and a per-rank batch size of 32, the `total_global_tokens` would be 32 * 16 = 512. However, the maximum batch size allowed for CUDA graph capture defaults to 160. As a result, the `can_run` function during the decode phase returns `False`, and CUDA graph execution is consequently skipped.

In fact, according to the design logic of the [PR](https://github.com/sgl-project/sglang/pull/4390), I don't consider this a bug—it consistently uses the `total_global_tokens` across all ranks as the batch size to be captured by the CUDA graph. A straightforward solution would be to set a sufficiently large `cuda-graph-max-bs` when launching the server, though this might consume a significant amount of additional memory.

I believe using the `num_tokens` per DP rank as the CUDA graph batch size might be a more reasonable approach, similar to the code prior to this [PR](https://github.com/sgl-project/sglang/pull/4390). It would only require reserving adequate space for the `gathered_buffer`.

Below are my test output throughput before and after this [PR](https://github.com/sgl-project/sglang/pull/4390).
2*8 H100, input_len=output_len=1000, DP=TP=16
| Concurrency| before PR | after PR | after PR-fix |
|:---:|:---:|:---:|:---:|
| 1024  | 5115.13  | 2581.07  | 5469.53  |
| 512  | 3897.78  | 1527.95  | 4509.96  |

### LINKED ISSUE #7951: [Bug] Tensor shape is wrong when cudagraph+enable_dp_attention
### Checklist

- [ ] 1. I have searched related issues but cannot get the expected help.
- [ ] 2. The bug has not been fixed in the latest version.
- [ ] 5. Please use English, otherwise it will be closed.

### Describe the bug

I tried to run DSR1 fp4 model on 8xB200, but found that some issue when I opened cudagraph and attndp, the input tensor dimension for each MoE layer is padded to global bs. For example, I take global bs 4096 and attention dp 8, which each rank should have 512 reqs for decode and the input tensor M dimension should be 512 for local rank. 
But I tried to do some profiling, I found that when cudagraph is on, each rank has input M dim 4096, not 512. When cudagraph is off, each rank has input M dim 512 which looks good.
Is this known or a bug?
Without cudagraph

Image

With cudagraph

Image

### Reproduction

**Server:**
--model-path nvidia/DeepSeek-R1-0528-FP4 \
--trust-remote-code \
--quantization modelopt_fp4 \
--dp-size 8 --enable-dp-attention --enable-dp-lm-head\
--tp-size 8 \
--attention-backend cutlass_mla \
--enable-ep-moe \
--enable-flashinfer-moe \
--cuda-graph-bs 1 2 4 8 16 32 64 128 256 512 1024 2048 4096 \
--mem-fraction-static 0.85 \
--max-running-requests 4096 \
--stream-interval 5 
**Client:**
benchmark_serving.py with isl/osl 1024/1024, concurrency 4096.

### Environment

latest main.

ch-wan:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #8279
* __->__ #8278
* #8277
* #8276

## Background

Integrating CUDA Graph with DP attention is tricky because different DP ranks have different workload. Simply gathering tensors from all DP ranks would incur dynamic shape. There are two design choices to make DP attention compatible with CUDA Graph:

- **Padding tokens to `max(len)` and use *all-gather* to gather tensors.** This design pads the input tensor of FFN to `max_len * dp_size` and incurs redundant computation (see examples in [#4390](https://github.com/sgl-project/sglang/pull/4390) and response [here](https://github.com/sgl-project/sglang/pull/6092#issuecomment-2860473971)).
- **Padding tokens to `sum(len)` and use *all-reduce* to gather tensors.** This design padded tokens to `sum(len)` which increases computation in the attention layers ([#6092](https://github.com/sgl-project/sglang/pull/6092) [#7951](https://github.com/sgl-project/sglang/issues/7951)). In addition, the cuda graph batch size is `sum(len)` rather than the actual running batch size on each DP worker. The users need to increase cuda graph batch size when DP size in creased, which causes a lot of confusion or misuse ([#5527](https://github.com/sgl-project/sglang/issues/5527) [#5557](https://github.com/sgl-project/sglang/issues/5557) [#6092](https://github.com/sgl-project/sglang/issues/6092)).

This PR offers two gather modes. When CUDA graph is disabled, we the select the option that minimizes communication. When CUDA graph is in use, we use the first option.

## Future Work

Under `max_len` padding mode, the input hidden_state of FFN is padded to `max_len * dp_size`, which may incur redundant computation. In the future, we can use masked (gouped-)GeMM kernels (e.g., DeepGEMM) to optimize it.

### DIFF
--- a/python/sglang/srt/layers/dp_attention.py
+++ b/python/sglang/srt/layers/dp_attention.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import functools
 import logging
 from contextlib import contextmanager
+from enum import IntEnum, auto
 from typing import TYPE_CHECKING, List, Tuple
 
 import torch
@@ -28,6 +29,33 @@ _ATTN_DP_RANK = None
 _ATTN_DP_SIZE = None
 _LOCAL_ATTN_DP_SIZE = None
 _LOCAL_ATTN_DP_RANK = None
+
+
+class DPGatherMode(IntEnum):
+
+    # Gather tokens using `all_gather_into_tensor`
+    ALL_GATHER = auto()
+    # Gather tokens using `all_reduce`
+    ALL_REDUCE = auto()
+
+    def is_all_gather(self):
+        return self == DPGatherMode.ALL_GATHER
+
+    def is_all_reduce(self):
+        return self == DPGatherMode.ALL_REDUCE
+
+    @classmethod
+    def get_dp_gather_mode(cls, global_num_tokens: List[int]) -> DPGatherMode:
+        max_len = max(global_num_tokens)
+        sum_len = sum(global_num_tokens)
+        if sum_len * 2 > max_len * get_attention_dp_size():
+            return cls.ALL_GATHER
+        else:
+            return cls.ALL_REDUCE
+
+    @classmethod
+    def get_default_mode_in_cuda_graph(cls) -> DPGatherMode:
+        return cls.ALL_GATHER
 
 
 def compute_dp_attention_world_info(enable_dp_attention, tp_rank, tp_size, dp_size):
@@ -221,7 +249,7 @@ def memcpy_triton(dst, src, dim, offset, sz, offset_src):
     memcpy_triton_kernel[grid](dst, src, offset, sz, offset_src, chunk_size, BLOCK_SIZE)
 
 
-def _dp_gather(
+def _dp_gather_via_all_reduce(
     global_tokens: torch.Tensor,
     local_tokens: torch.Tensor,
     forward_batch: ForwardBatch,
@@ -254,6 +282,38 @@ def _dp_gather(
 
     else:
         global_tokens[:] = tensor_model_parallel_all_reduce(global_tokens)
+
+
+def _dp_gather_via_all_gather(
+    global_tokens: torch.Tensor,
+    local_tokens: torch.Tensor,
+    forward_batch: ForwardBatch,
+    is_partial: bool,
+):
+    if not is_partial:
+        if get_attention_tp_rank() != 0:
+            local_tokens.fill_(0)
+    scattered_local_tokens = local_tokens.tensor_split(get_attention_tp_size())[
+        get_attention_tp_rank()
+    ]
+    get_attention_tp_group().reduce_scatter_tensor(scattered_local_tokens, local_tokens)
+    get_tp_group().all_gather_into_tensor(global_tokens, scattered_local_tokens)
+
+
+def _dp_gather(
+    global_tokens: torch.Tensor,
+    local_tokens: torch.Tensor,
+    forward_batch: ForwardBatch,
+    is_partial: bool,
+):
+    if forward_batch.dp_gather_mode.is_all_gather():
+        _dp_gather_via_all_gather(
+            global_tokens, local_tokens, forward_batch, is_partial
+        )
+    else:
+        _dp_gather_via_all_reduce(
+            global_tokens, local_tokens, forward_batch, is_partial
+        )
 
 
 def dp_gather_partial(
--- a/python/sglang/srt/layers/dp_attention.py
+++ b/python/sglang/srt/layers/dp_attention.py
@@ -31,31 +31,32 @@ _LOCAL_ATTN_DP_SIZE = None
 _LOCAL_ATTN_DP_RANK = None
 
 
-class DPGatherMode(IntEnum):
+class DPPaddingMode(IntEnum):
 
-    # Gather tokens using `all_gather_into_tensor`
-    ALL_GATHER = auto()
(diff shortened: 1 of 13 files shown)


python/sglang/srt/layers/dp_attention.py
```diff
@@ -30,6 +31,33 @@
...
+        return self == DPGatherMode.ALL_GATHER
+
+    def is_all_reduce(self):
+        return self == DPGatherMode.ALL_REDUCE
+
+    @classmethod
+    def get_dp_gather_mode(cls, global_num_tokens: List[int]) -> DPGatherMode:
+        max_len = max(global_num_tokens)
+        sum_len = sum(global_num_tokens)
+        if sum_len * 2 > max_len * get_attention_dp_size():
```
ispobock, line 52: Just curious, how did we get this formula?

python/sglang/srt/speculative/eagle_draft_extend_cuda_graph_runner.py
```diff
@@ -306,12 +308,13 @@ def replay(self, forward_batch: ForwardBatch):
         raw_bs = forward_batch.batch_size
         num_tokens = forward_batch.input_ids.shape[0]
         if self.require_mlp_tp_gather:
-            total_batch_size = (
-                sum(forward_batch.global_num_tokens_cpu) // self.num_tokens_per_bs
+            max_num_tokens = max(forward_batch.global_num_tokens_cpu)
```
ispobock, line 311: nit: use `forward_batch.global_num_tokens_cpu.max().item()` may be more efficient?

python/sglang/srt/model_executor/forward_batch_info.py
```diff
@@ -585,34 +592,158 @@ def prepare_chunked_kv_indices(self, device: torch.device):
             )
             self.prefix_chunk_kv_indices.append(chunk_kv_indices)
 
-    def _pad_tensor_to_size(self, tensor: torch.Tensor, size: int):
-        return torch.cat(
-            [tensor, tensor.new_zeros(size - tensor.shape[0], *tensor.shape[1:])], dim=0
-        )
+    def _pad_tensor_to_size(self, tensor: torch.Tensor, size: int, *, value: int = 0):
+        if value == 0:
+            return torch.cat(
```
ispobock, line 597: nit: why don't use torch pad function?

ispobock (approved):
LGTM

python/sglang/srt/layers/dp_attention.py
ch-wan: The communication volume of each worker for AllGather mode and AllReduce mode are `max_len * dp_size` and `sum_len * 2`, respectively.

python/sglang/srt/model_executor/forward_batch_info.py
ch-wan: Are you asking this one (https://docs.pytorch.org/docs/stable/generated/torch.nn.functional.pad.html)? It is for image padding. Unfortunately, I'm not sure how to use them.

python/sglang/srt/speculative/eagle_draft_extend_cuda_graph_runner.py
ch-wan: Good catch. I'm going to revise this.

ch-wan: lol, `global_num_tokens_cpu` is a list of int

PR #8224, opened 2025-07-21
GLM-4.5 Model Support

zRzRzRzRzRzRzR:
The SGLang version of the complete implementation of the GLM-4.5 model, which includes:
1. Model implementation (with MTP and without MTP)
2. Tool call and Reasoning parser
3. transformers lib should use 4.54.0 or higher

Everything is ready.

PR #7708, opened 2025-07-02
Fix the issue of incorrect finish reason in final stream response chunk returned during tool call

xianzhiT:
## Motivation
Current release v0.4.8 in order to comply with the OpenAI protocol, class OpenAIServingChat adds an additional chunk at the end of the streaming output to describe the reason for the end of the streaming output. 
Clipboard_Screenshot_1751429477

When the streaming output ends with tool_calls, the finish_reason in the reply chunk should be tool_calls. The current code did not perform a tool call check, which resulted in the final chunk always having a 'stop' instead of the expected 'tool_calls' value for the' finish.rason '. This can lead to unpredictable behavior in some open source software (eg. strand agent).
Clipboard_Screenshot_1751433449

## Modifications
Corrected the logic for obtaining finish_reason, returning and setting finish_reason to tool_calls in the final chunk after determining it as a tool call.

Clipboard_Screenshot_1751433921

## Checklist

- [x] Format your code according to the [Code Formatting with Pre-Commit](https://docs.sglang.ai/references/contribution_guide.html#code-formatting-with-pre-commit).
- [ ] Add unit tests as outlined in the [Running Unit Tests](https://docs.sglang.ai/references/contribution_guide.html#running-unit-tests-adding-to-ci).
- [ ] Update documentation / docstrings / example tutorials as needed, according to [Writing Documentation](https://docs.sglang.ai/references/contribution_guide.html#writing-documentation-running-docs-ci).
- [ ] Provide throughput / latency benchmark results and accuracy evaluation results as needed, according to [Benchmark and Profiling](https://docs.sglang.ai/references/benchmark_and_profiling.html) and [Accuracy Results](https://docs.sglang.ai/references/accuracy_evaluation.html).
- [ ] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co-author when merging the PR.
- [ ] Please feel free to join our Slack channel at https://slack.sglang.ai to discuss your PR.

PR #8316, opened 2025-07-24
Fix: Improve test_openai_function_calling unit test and fix reasoning_parser.py think_start_token logic

byjiang1996:
## Motivation & Modifications

As part of the collaboration with Zhipu team on https://github.com/sgl-project/sglang/pull/8224. we noticed:
* `reasoning_parser.py's detect_and_parse think_start_token logic` is inconsistent with `reasoning_parser.py's detect_and_parse parse_streaming_increment logic`. In the latter case, it uses `self.think_start_token in current_text` to tell whether the current model output is in reasoning mode but the former case uses `current_text.startswith(self.think_start_token)`. It leads to something like `\n${self.think_start_token}` text are not treated as reasoning mode due to one extra \n in the begining. In vllm, it uses `start_token in text` check too in both cases: https://github.com/vllm-project/vllm/blob/6d8d0a24c02bfd84d46b3016b865a44f048ae84b/vllm/reasoning/qwen3_reasoning_parser.py#L99
* Add one more unit test for https://github.com/sgl-project/sglang/pull/7708 to check whether `test_function_calling_streaming_args_parsing` finishes with `finish_reason=tool_call` instead of `finish_reason=stop` to catch any regression. The regression will break Zhipu team's key feature integration.
* Add `_test_function_calling_multiturn` to  `test_openai_function_calling` unit test to catch future regressions b/c this test is critical for Zhipu team's business integration
* Add `max_tokens=2048` to  all unit tests in `test_openai_function_calling`. Sometimes i noticed the LLM never stops which takes forever to run the unit tests.... Adding max token should help fail faster.
* Change ` messages = [{"role": "user", "content": "What is the temperature in Paris?"}]` to `{"role": "user", "content": "What is the temperature in Paris in celsius?"}`. The provided tool function has both celsius and fahrenheit option. If we ask LLM `What is the temperature in Paris`, it may reply with `what temperature unit you need` instead of calling tool function directly. It will fail the unit test in a flaky way. Asking `What is the temperature in Paris in celsius?` significantly improves the unit test success rate
image
* Fix finish_reason != tool_call bug when tool_call and reasoning are both open. This PR https://github.com/sgl-project/sglang/pull/7708 only fixes the bug when only `tool-call-parser` is defined. If both `tool-call-parser` and `reasoning-parser` are defined, it will skip (due to `for-loop's continue`) _process_tool_call_stream's `yield None, tool_calls` logic: https://github.com/sgl-project/sglang/pull/7708/files#diff-20a5bbe903ddc72a708cae54cfc9bd1ba42f3693a90dab09fc10904329bb6377R931
image
   - Before this change in GLM4: this `TestOpenAIServerFunctionCalling's test_function_calling_streaming_simple` unit test will fail with `finish_reason != 'tool_calls'`
image
   - After the change, it works. Also added one more unit test `test_function_calling_streaming_no_tool_call`
image

## Checklist

- [x] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co-author when merging the PR.


python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -493,9 +493,6 @@ async def _generate_chat_stream(
                         )
                         yield f"data: {chunk.model_dump_json()}\n\n"
 
-                    if not delta:
```
CatherineSue, line 496: why do we remove this block?

test/srt/openai_server/function_call/test_openai_function_calling.py
```diff
@@ -76,6 +76,7 @@ def test_function_calling_format(self):
         messages = [{"role": "user", "content": "Compute (3+5)"}]
         response = client.chat.completions.create(
             model=self.model,
+            max_tokens=2048,
```
CatherineSue, line 79: What is the reason to add max_tokens here?

byjiang1996: Sometimes i noticed the LLM never stops which takes forever to run the unit tests (for example, it is EBNF issue and sometimes it is model's own issue).... Adding max token should help UT to fail faster instead of getting timed out after 10~30 min. (I also noticed some other existing sglang UT timed out occassionally after 30min; if every UT can have a max_tokens limit, it should help UT fail sooner and reduce CI pressure)

python/sglang/srt/entrypoints/openai/serving_chat.py
byjiang1996: https://github.com/sgl-project/sglang/pull/7708 partially addresses the issue where finish_reason = stop even when tool_call exists in the response.

However, that logic won't necessarily work when both reasoning parser and tool call parser are enabled. `If not delta` logic in reasoning_parser will skip the for-loop via `continue` and thus skip the `finish_reason_type = tool_call_finish_reason_type` logic. `finish_reason_type` is still `stop` when tool_call exists in the response.
image

test/srt/openai_server/function_call/test_openai_function_calling.py
```diff
@@ -166,6 +248,82 @@ def test_function_calling_streaming_simple(self):
...
+        self.assertTrue(len(chunks) > 0, "Streaming should return at least one chunk")
+
+        found_function_name = False
+        for chunk in chunks:
+            choice = chunk.choices[0]
+            # Check whether the current chunk contains tool_calls
+            if choice.delta.tool_calls:
+                tool_call = choice.delta.tool_calls[0]
+                if tool_call.function.name:
+                    self.assertEqual(
```
CatherineSue: This assestion seems to be unnecessary. The test should not have tool call. We can simply set found_function_name to True after L304.

python/sglang/srt/entrypoints/openai/serving_chat.py
CatherineSue: Is this for the case when the last delta is empty but we still need to parse finish_type?

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -942,5 +939,7 @@ async def _process_tool_call_stream(
             )
             yield f"data: {chunk.model_dump_json()}\n\n", finish_reason_type
 
-        if finish_reason_type == "stop":
+        if finish_reason_type == "stop" and any(
```
CatherineSue: Could you explain more what this situation is for? Is it to make sure there are tools before?

python/sglang/srt/entrypoints/openai/serving_chat.py
byjiang1996: > Is this for the case when the last delta is empty but we still need to parse finish_type?

Correct. This is the streaming sequence that will cause finish_reason != 'tool_calls' before my fix:
- ....
- delta="something", finish_reason = None, tool_calls = "something"
- delta = "", finish_reason = "stop", tool_calls=None

The bug is b/c the implementation before my change simply skip the `if finish_reason== stop: set finish_reason = tool_calls` logic in `self. _process_tool_call_stream` due to `if not delta: continue`

python/sglang/srt/entrypoints/openai/serving_chat.py
byjiang1996: > Is it to make sure there are tools before?

Yes, correct. Zhipu reported a bug where there is no tool_call in the streamed response at all but finish_reason is still tool_call - it will cause claude code and other downstream to fail. This change fixes the behavior by checking whether there is any tool call before in the same response: if yes, set finsh_reason = tool_calls; otherwise, do nothing.

It is the same logic as in VLLM https://github.com/vllm-project/vllm/blob/a9b2a1d7048c7129ee6f55af68624256da864fed/vllm/entrypoints/openai/serving_chat.py#L881
image

image

test/srt/openai_server/function_call/test_openai_function_calling.py
byjiang1996: oh! good catch. Fixed the check condition

python/sglang/srt/entrypoints/openai/serving_chat.py
CatherineSue: Thanks for the info. I am actually not sure if their impl is bug free. In rare case, if request.n > 1, and index=0 doesn't have tool call, index=1 has it. Their implementation might simple put tool_calls for all indices.

CatherineSue: Is it possible to revert this line of change, so we can focus it on another PR? I will introduce the fix

byjiang1996: Sounds good. Done. 

I removed this line and commented out a UT that covers this scenario.

CatherineSue (approved):

PR #8320, opened 2025-07-24
Molly/cpu ci refinement

DiweiSun:
## Summary

This PR extends the CI coverage by integrating the following configurations into the test pipeline:

- **Qwen3-1.7B FP8 models**
- **LLama3.2-3B W8A8 (INT8 weight-only quantization) models**
- **Qwen3-30B MoE FP8 models**
- **Tensor Parallel execution for the above model types**

These additions significantly expand the range of supported model formats and execution modes in our CI system.

## Test Plan

- Integrated new test jobs into the existing CI workflow.
- Verified successful execution of all added configurations on tensor parallel.
- Confirmed correct integration with existing test infrastructure and reporting tools.

## Motivation

As we continue to support more advanced LLM inference features, it's essential to ensure they are validated continuously through our CI system. This change enables early detection of regressions and improves overall test coverage for upcoming  workloads.

## Additional Information

- These configurations are now part of the standard CI matrix.
- No functional changes to the core inference engine were made in this PR.


test/srt/test_intel_amx_attention_backend.py
```diff
@@ -19,8 +28,29 @@
 )
 
 
+def with_cpu_omp_threads_bind(value="0-15|16-31|32-47|48-63|64-79|80-95"):
```
mingfeima: we can ignore `with_cpu_omp_threads_bind`, hard code this is not a good option.

if we don't set `SGLANG_CPU_OMP_THREADS_BIND`, it will use all available numa nodes on the machine by default.

mingfeima (changes requested):

test/srt/test_intel_amx_attention_backend.py
DiweiSun: Automatically binding does not work if the number of numas is less than tp rank. And the ci host is 2 numa, but we'd like to test tp=6 in this case.

PR #8383, opened 2025-07-26
Add assertion for abnormal input id range

fzyzcjy:
## Motivation

o/w give hard-to-understand errors for users

## Checklist


ispobock (approved):

ispobock:
https://github.com/sgl-project/sglang/actions/runs/16539732751/job/46780472339?pr=8383#step:4:12449
ci failed

fzyzcjy:
fixed (and wait again)

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -592,6 +592,15 @@ def _validate_one_request(
                     "Please set `--enable-custom-logits-processor` to enable this feature."
                 )
 
+            if (self.tokenizer is not None) and ((max_input_id := max(input_ids)) >= len(self.tokenizer)):
+                raise ValueError(
+                    f"The maximum input id ({max_input_id}) should be less than num tokens ({len(self.tokenizer)})"
+                )
+            if (min_input_id := min(input_ids)) < 0:
+                raise ValueError(
+                    f"The minimum input id ({min_input_id}) should be non-negative"
```
JustinTong0323, line 693: I heard that some inhouse VLM use negative id as image token id @nanjiangwill

fzyzcjy: btw, for "inhouse" do you mean it is not open sourced and not compatible with sglang etc?

if it is only vlm, then maybe we can gate the `if` with is-it-llm, not sure whether that is enough

nanjiangwill: for general case, all tokens(including image token) should be positive

i think it is ok to write in this way.

nanjiangwill: maybe adding a gate is enough

fzyzcjy: > for general case, all tokens(including image token) should be positive
i think it is ok to write in this way.

I see, then maybe keep it unchanged and if anyone has issues we can fix it.

Sunt-ing:
Confirming this is still a live issue and worth landing. On current `main` a single request with an out-of-range prompt `input_ids` (negative, or `>= vocab_size`) reaches the GPU embedding gather and triggers a CUDA device-side assert that takes down the whole engine, so it is a request-level DoS rather than a per-request error.

<details>
<summary>Independent repro (A800-80GB, TinyLlama vocab 32000, current main)</summary>

Reproduces on any llama-family model; below is the exact command I ran (a w4a16-quantized TinyLlama, vocab 32000).

```bash
python -m sglang.launch_server --model-path tinyllama-w4a16-channel-v2 \
  --host 127.0.0.1 --port 31021 --attention-backend triton \
  --sampling-backend pytorch --disable-cuda-graph --disable-piecewise-cuda-graph \
  --mem-fraction-static 0.5 --dtype float16 --max-running-requests 8

curl -s 127.0.0.1:31021/generate -d '{"input_ids":[-1],"sampling_params":{"max_new_tokens":4}}'
```

Before this PR:

```
attack input_ids=[-1]    -> RemoteDisconnected (engine died)
/health_generate after   -> Connection refused
normal /generate after   -> Connection refused
server process           -> exit -9
server log               -> indexSelectSmallIndex: Assertion 'srcIndex < srcSelectDimSize' failed (repeated), SIGQUIT
```

The OpenAI-compatible `/v1/completions` with an integer-list `prompt` (e.g. `prompt: [-1]`) hits the exact same crash, and that is the path most real clients use.

</details>

A few gaps that may be worth covering before this lands (the branch also conflicts with `main` now):

1. The upper-bound check is gated on `self.tokenizer is not None`. With `--skip-tokenizer-init` (exactly the mode where clients pass raw `input_ids`), `self.tokenizer` is `None`, so the `>= num_tokens` check is skipped and a huge OOV id such as `2000000000` still crashes. Only the `min < 0` check is unconditional in that mode.
2. `len(self.tokenizer)` can differ from the model's real embedding size; `self.model_config.vocab_size` matches the embedding table that the gather actually indexes into, which is what the device-side assert checks against.
3. No regression test. A small case under `test/registered/openai_server/validation/` that drives `/v1/completions` with `prompt=[-1]` and `[2_000_000_000]` and asserts a 400 would lock this in.
4. The check lives in the generate-request branch only. `EmbeddingReqInput` with an integer-list input has the same unbounded path into the embedding gather.

<details>
<summary>After the fix: out-of-range ids return 400 and the server stays healthy</summary>

```
/generate input_ids=[-1]          -> 400 (out-of-range input id rejected)
/generate input_ids=[2000000000]  -> 400 (out-of-range input id rejected)
/v1/completions prompt=[-1]       -> 400 (BadRequestError)
/generate input_ids=[1,2,3]       -> 200 (normal completion)
health + normal request after each-> ok, server running throughout
```

</details>

<details>
<summary>Regression test (drop into test/registered/openai_server/validation/test_request_length_validation.py)</summary>

```python
def test_out_of_vocabulary_prompt_input_ids(self):
    client = openai.Client(api_key=self.api_key, base_url=f"{self.base_url}/v1")

    for bad_prompt in ([-1], [2_000_000_000]):
        with self.assertRaises(openai.BadRequestError) as cm:
            client.completions.create(
                model=DEFAULT_SMALL_MODEL_NAME_FOR_TEST,
                prompt=bad_prompt,
                max_tokens=4,
                temperature=0,
            )
        self.assertIn("input id", str(cm.exception))
```

</details>

PR #8444, opened 2025-07-28
[Improvements] Merge health check route

whybeyoung:
Although PR #8115 (which introduced a refined health check mechanism) has been reverted, the core issues it aimed to address—inadequate health monitoring in production and cloud-native environments—remain unresolved. 

Merge the routing logic of the original /health and /health_generate endpoints. To maintain compatibility, both routes will be retained but will follow the same logic:

Step 1: When sglang starts, initialize the state as Starting. Once the warmup request succeeds, set the state to Healthy.

Step 2: Only when the state is Healthy will the system intelligently determine whether it can successfully generate a token based on the load situation to judge if the current server is healthy.

This PR, combined with Kubernetes probes, ensures that traffic is only routed to the engine after it has fully started, and the health_generate logic will only be executed when the engine is running normally.

CC @ByronHsu   @merrymercy

### DIFF
--- a/python/sglang/srt/entrypoints/http_server.py
+++ b/python/sglang/srt/entrypoints/http_server.py
@@ -234,8 +234,13 @@ async def validate_json_request(raw_request: Request):
 @app.get("/health")
 @app.get("/health_generate")
 async def health_generate(request: Request) -> Response:
-    """Two Step Check the health of the server."""
-    """Step1: Check the health of the server."""
+    """
+    Check the health of the inference server by sending a special request to generate one token.
+
+    If the server is running something, this request will be ignored, so it creates zero overhead.
+    If the server is not running anything, this request will be run, so we know whether the server is healthy.
+    """
+
     if _global_state.tokenizer_manager.gracefully_exit:
         logger.info("Health check request received during shutdown. Returning 503.")
         return Response(status_code=503)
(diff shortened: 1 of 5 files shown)


slin1237 (approved):

python/sglang/srt/entrypoints/http_server.py
```diff
@@ -218,15 +219,14 @@ async def validate_json_request(raw_request: Request):
 
 
 @app.get("/health")
-async def health() -> Response:
```
zhyncs, line 233: In order to be compatible, don't delete this

whybeyoung: ok , i add it back

whybeyoung:
Prefill: image
Decode:
image


Tested OK

python/sglang/srt/entrypoints/http_server.py
```diff
@@ -230,30 +233,39 @@ async def validate_json_request(raw_request: Request):
...
             input_ids=[0],
             sampling_params=sampling_params,
             log_metrics=False,
         )
+        if (
+            _global_state.tokenizer_manager.server_args.disaggregation_mode
+            != DisaggregationMode.NULL
+        ):
+            gri.bootstrap_host = FAKE_BOOTSTRAP_HOST
+            gri.bootstrap_room = random.randint(
```
ByronHsu: QQ: if fake, can we always use 0?

merrymercy (approved):

PR #7884, opened 2025-07-09
[kernel] opt moe align block kernel by block/warp scan algorithm

yuan-luo:
<!-- Thank you for your contribution! We appreciate it. The following guidelines will help improve your pull request and facilitate feedback. If anything is unclear, don't hesitate to submit your pull request and ask the maintainers for assistance. -->

## Motivation

<!-- Explain the purpose of this PR and the goals it aims to achieve. -->
This PR is to introduce block / warp scan algorithm in fused MoE path **moe_align_block_size_kernel**  which gains approximately 10% speedup.

Here is the benchmark result.
Note: num_experts >= 128 appliable to this PR. num_experts < 128 appliable to **moe_align_block_size_small_batch_expert_kernel**  kernel.

This PR:
```
$python ./sgl-kernel/benchmark/bench_moe_align_block_size.py
INFO 07-09 13:15:28 [__init__.py:244] Automatically detected platform cuda.
✅ VLLM implementation works with 64 experts!
✅ SGL and Triton implementations match
✅ SGL and VLLM implementations match

📊 Running performance benchmark for 64 experts...
moe-align-block-size-performance:
     num_tokens  num_experts  topk        SGL  SGL Fusion       Triton
0           1.0          8.0   1.0  16.672000   14.912000    44.032000
1           1.0          8.0   2.0  16.896000   15.072000    44.256002
2           1.0          8.0   4.0  16.672000   15.008000    43.264002
3           1.0          8.0   8.0  16.432000   15.104000    44.767998
4           1.0         32.0   1.0  19.200001   17.424000    44.512000
5           1.0         32.0   2.0  19.231999   17.503999    44.464000
6           1.0         32.0   4.0  19.200001   17.535999    43.728001
7           1.0         32.0   8.0  19.231999   17.568000    44.767998
8           1.0         64.0   1.0  22.655999   20.864001    44.160001
9           1.0         64.0   2.0  22.608001   20.927999    44.799998
10          1.0         64.0   4.0  22.624001   20.927999    44.863999
11          1.0         64.0   8.0  22.496000   20.959999    45.343999
12          1.0        128.0   1.0  19.296000   17.696001    47.295999
13          1.0        128.0   2.0  19.264000   17.472001    44.927999
14          1.0        128.0   4.0  19.328000   17.728001    43.968000
15          1.0        128.0   8.0  19.296000   17.503999    44.560000
16          1.0        256.0   1.0  19.231999   17.632000    45.311999
[...]

PR #8457, opened 2025-07-28
Revert "[kernel] opt moe align block kernel by block/warp scan algorithm"

BBuf:
Reverts sgl-project/sglang#7884

PR has bug, and it caused a lot of test failture in ci.


yuan-luo:
7884 has issue when running in AMD environment, fixing in progress.

Issue #8402, opened 2025-07-27
[Bug] DeepSeek-V3 model gets bad accuracy result on gsm8k benchmark when EP is enabled


zminglei:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Describe the bug

**DeepSeek-V3** model gets **bad** **accuracy** result on gsm8k benchmark when **EP** is enabled.
TP 8 result: 0.945 (Accuracy)
EP result: 0.845 (Accuracy)

### Reproduction

**EP** result reproduce:
```
# Terminal 1
python -m sglang.launch_server --model /shared/public/elr-models/deepseek-ai/DeepSeek-V3-0324/e9b33add76883f293d6bf61f6bd89b497e80e335/ --tp 8 --trust-remote-code --enable-ep-moe --mem-fraction-static 0.8


# Terminal 2
python benchmark/gsm8k/bench_sglang.py
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 200/200 [00:28<00:00,  7.11it/s]
Accuracy: 0.845
Invalid: 0.000
Latency: 29.142 s
Output throughput: 595.128 token/s
```
**TP** result reproduce:
```
# Terminal 1
python -m sglang.launch_server --model /shared/public/elr-models/deepseek-ai/DeepSeek-V3-0324/e9b33add76883f293d6bf61f6bd89b497e80e335/ --tp 8 --trust-remote-code --mem-fraction-static 0.8


# Terminal 2
python benchmark/gsm8k/bench_sglang.py --data-path /shared/public/data/gsm8k/test.jsonl
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 200/200 [00:22<00:00,  8.92it/s]
Accuracy: 0.945
Invalid: 0.000
Latency: 22.641 s
Output throughput: 886.078 token/s
```

### Environment

[...]

PR #8364, opened 2025-07-25
[1/2] sgl-kernel: Fuse routed scaling factor into select_experts

trevor-m:
## Motivation

Follow up to https://github.com/sgl-project/sglang/pull/8333
Fuse the multiply by routed_scaling_factor into select_experts, following example of TRT-LLM:
https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/_torch/models/modeling_deepseekv3.py#L323
https://github.com/NVIDIA/TensorRT-LLM/blob/738ab615930fd08dccb94fa388bd74dc91c5f235/cpp/tensorrt_llm/kernels/noAuxTcKernels.cu#L651

For the non-FP4 paths, the routed_scaling_factor is fused into moe_sum_reduce. However, we could move it into select_experts for those paths too if we wanted to simplify the code.

## Modifications

Add boolean argument to fused_moe_gate()

## Results

Prefill:
10.46% speedup at BS 1
1.86% speedup at BS 128
Decode:
1.22% speedup at BS1
0.26% speedup at BS128

Server command
```
python3 -m sglang.launch_server --model-path nvidia/DeepSeek-R1-0528-FP4 --trust-remote-code --quantization modelopt_fp4 --tp 8 --enable-flashinfer-cutlass-moe --enable-ep-moe --ep-size 8
```

BEFORE results
```
python3 -m sglang.bench_one_batch_server --model-path nvidia/DeepSeek-R1-0528-FP4 --base-url http://127.0.0.1:30000/ --batch-size 1 --input-len 1024 --output-len 1024

#Input tokens: 1024
#Output tokens: 1024
batch size: 1
input_len: 1024
output_len: 1024
latency: 10.60 s
ttft: 0.11 s
last generation throughput: 96.61 tok/s
input throughput: 9392.46 tok/s
output throughput: 97.62 tok/s

python3 -m sglang.bench_one_batch_server --model-path nvidia/DeepSeek-R1-0528-FP4 --base-url http://127.0.0.1:30000/ --batch-size 128 --input-len 1024 --output-len 1024

#Input tokens: 131072
#Output tokens: 131072
batch size: 128
input_len: 1024
output_len: 1024
latency: 39.79 s
ttft: 7.56 s
last generation throughput: 3693.98 tok/s
input throughput: 17344.15 tok/s
output throughput: 4066.24 tok/s
```

AFTER results
```
[...]

PR #8511, opened 2025-07-29
Bug: apply final_hidden_states*=self.routed_scaling_factor at MoE lay…

byjiang1996:
…er if epmoe is enabled

## Motivation

Issue: https://github.com/sgl-project/sglang/issues/8402

Also noticed significant regression when running epmoe during recent GLM4.5 support work: GSM8K accuracy drops from 0.965 to 0.745 when EPMOE is enabled. Accuracy is good for TP & DeepEP.

```
python3 -m sglang.launch_server --model  /shared/public/elr-models/zai-org/GLM-4.5 --tp-size 8 --trust-remote-code

python3 benchmark/gsm8k/bench_sglang.py
```

## Modifications

There is a bug in DeepSeek-V2 and GLM-4.5 related to how routed_scaling_factor is applied in MoE (Mixture-of-Experts) layers.

Currently, the routed_scaling_factor is applied in three different places, leading to ambiguity:
- self.topk.forward: The scaling is applied only in the n out of m condition, but not in the m - n out of m condition.
- self.experts.forward: Same as above — the factor is only applied in n out of m.
- Model-level logic: The model itself may apply routed_scaling_factor, but it cannot know whether self.topk or self.experts have already done so.

This results in uncertain and inconsistent scaling, as the model layer has no visibility into whether routed_scaling_factor has already been applied upstream.

🎪 TL;DR: The model can't know if it should apply * routed_scaling_factor or not, because topk and experts may or may not have already done it, depending on the code path.

My PR forces model layer to apply * routed_scaling_factor when EPMOE is enabled because from the current codebase, epmoe won't apply *routed_scaling_factor by itself. 

Need to follow up and possibly refactor sglang moe codebase to make it clear which layer should apply * routed_scaling_factor 

## Accuracy Test
- GLM4.5 GSM8K accuracy jumps from 0.745 to 0.965 under EPMOE which is the same as TP & DeepEP results
- Need to run gsm8k for deepseek-v3 too - WIP
## Checklist

- [x] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co-author when merging the PR.

### DIFF
--- a/python/sglang/srt/models/deepseek_v2.py
+++ b/python/sglang/srt/models/deepseek_v2.py
@@ -506,7 +506,7 @@ class DeepseekV2MoE(nn.Module):
         else:
             kwargs["router_logits"] = router_logits
         final_hidden_states = self.experts(**kwargs)
-        if not _is_cuda and not _use_aiter or global_server_args_dict["enable_ep_moe"]:
+        if not _is_cuda and not _use_aiter:
             # fused in biased_grouped_topk so we can skip here
             # TODO: please fix me
             # However, biased_grouped_topk is not guaranteed to run *=self.routed_scaling_factor when not _is_cuda and not _use_aiter
--- a/python/sglang/srt/models/deepseek_v2.py
+++ b/python/sglang/srt/models/deepseek_v2.py
@@ -508,9 +508,6 @@ class DeepseekV2MoE(nn.Module):
         final_hidden_states = self.experts(**kwargs)
         if not _is_cuda and not _use_aiter:
             # fused in biased_grouped_topk so we can skip here
-            # TODO: please fix me
-            # However, biased_grouped_topk is not guaranteed to run *=self.routed_scaling_factor when not _is_cuda and not _use_aiter
-            # In addition, `self.experts` sometimes run *=self.routed_scaling_factor again during TP & deepep
             final_hidden_states *= self.routed_scaling_factor
         if shared_output is not None:
             final_hidden_states = final_hidden_states + shared_output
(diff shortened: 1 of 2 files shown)


python/sglang/srt/models/deepseek_v2.py
```diff
@@ -506,8 +506,11 @@ def forward_normal(
         else:
             kwargs["router_logits"] = router_logits
         final_hidden_states = self.experts(**kwargs)
-        if not _is_cuda and not _use_aiter:
+        if not _is_cuda and not _use_aiter or global_server_args_dict["enable_ep_moe"]:
```
trevor-m: Maybe we can have a property in `self.experts` whether to apply routing scaling factor in the model?
Or can we fuse it for EpMoe too?

These are the scenarios I know about for cuda:
* FusedMoE triton backend: multiply is fused into `moe_sum_reduce`, but also need to divide from shared experts in `biased_grouped_topk` to cancel it out
* FusedMoE model_opt FP4 (with or without enable_ep_moe): applied in `ModelOptNvFp4FusedMoEMethod` but https://github.com/sgl-project/sglang/pull/8364 will fuse multiply into `biased_grouped_topk`. I'm worried this change will cause it to be applied twice for this path
* EpMoE - was missing but this PR will fix and bring to model forward
* DeepEpMoE - in model forward

ch-wan: Let me fix this after 8515 is merged.

trevor-m (approved):

PR #8519, opened 2025-07-29
update cudnn to support cuda graph capture with FP4 quantization

lingjiew:
upgrade cudnn to fix the error when using fp4 quantization with cudagraph.
```
  File "/sgl-workspace/sglang/python/sglang/srt/model_executor/cuda_graph_runner.py", line 362, in __init__
    raise Exception(
Exception: Capture cuda graph failed: cuDNN FP4 requires backend version >= 91002, found 90701. Please upgrade cuDNN backend.
```


lingjiew:
close this because there is fixed in new cuda base image

Issue #8569, opened 2025-07-30
[Bug] Reasoning models may lost some characters when streaming inference.


xjpang:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Describe the bug

During streaming inference, if a reasoning model is in a non-reasoning stage and the output contains part of the "begin thinking" or "end thinking" label, that fragment may be lost.

I discovered an issue in the parse_streaming_increment function of BaseReasoningFormatDetector. For instance, when _in_reasoning is False and the new_text input is <\/, which is part of the closing thinking tag  </think>, this fragment is stored in the buffer and an empty string is returned. In the next step, if new_text is "answer", since _in_reasoning remains False, the buffer is cleared and "answer" in new_text is returned directly—causing the </ from the previous step to be lost.

### Reproduction

During streaming inference, if a reasoning model is in a non-reasoning stage and the output includes part of the "begin thinking" or "end thinking" label, that output may be lost.

### Environment

I can reproduce in model Qwen/Qwen3-235B-A22B-Thinking-2507

PR #8572, opened 2025-07-30
Fix characters lost in reasoning model stream inference

xjpang:
## Motivation
I discovered an issue in the parse_streaming_increment function of BaseReasoningFormatDetector. For instance, when _in_reasoning is False and the new_text input is </, which is part of the closing thinking tag , this fragment is stored in the buffer and an empty string is returned. In the next step, if new_text is "answer", since _in_reasoning remains False, the buffer is cleared and "answer" in new_text is returned directly—causing the </ from the previous step to be lost.

Try to fix this issue https://github.com/sgl-project/sglang/issues/8569
## Checklist


CatherineSue (approved):

CatherineSue (commented):
Let's wait for CI to pass before merging this.

CatherineSue (changes requested):
Hi @xjpang sorry for revoking the approval here. I have misunderstood the logic. It seems the PR is trying to fix the case when the model is not in reasoning, but it tries to generate a partial `</think>` token? I am a bit lost because why would the code got executed into this block if it already has generated a `<think>` token and `self._in_reasoning` would be `True`.

CatherineSue:
> if a reasoning model is in a non-reasoning stage and the output contains part of the "begin thinking" or "end thinking" label, that fragment may be lost

If the output contains part of the </think> label, it will be entering this if block and there is no lost fragment

https://github.com/sgl-project/sglang/blob/a79a5d70121c3a4ad2add48ffdd6c7c5f7b1bc8b/python/sglang/srt/reasoning_parser.py#L69-L74

Can you explain a bit further regarding the model and the reasoning parser you are using? Perhaps a request which hits the bug? Thanks!

xjpang:
> not in reasoning

For example, in non-reasoning part, the output is string like “I am normal token </I am normal token>” . The </ token will lost. I can reproduce in Qwen/Qwen3-235B-A22B-Thinking-2507.

xjpang:
> Hi @xjpang sorry for revoking the approval here. I have misunderstood the logic. It seems the PR is trying to fix the case when the model is not in reasoning, but it tries to generate a partial `</think>` token? I am a bit lost because why would the code got executed into this block if it already has generated a `<think>` token and `self._in_reasoning` would be `True`.
In a reasoning model after reasoning part. in non-reasoning part, the output is string like “I am normal token </I am normal token>” . The </ token will lost. I can reproduce in Qwen/Qwen3-235B-A22B-Thinking-2507.

xjpang:
> > if a reasoning model is in a non-reasoning stage and the output contains part of the "begin thinking" or "end thinking" label, that fragment may be lost
> 
> If the output contains part of the label, it will be entering this if block and there is no lost fragment
> 
> https://github.com/sgl-project/sglang/blob/a79a5d70121c3a4ad2add48ffdd6c7c5f7b1bc8b/python/sglang/srt/reasoning_parser.py#L69-L74
> 
> Can you explain a bit further regarding the model and the reasoning parser you are using? Perhaps a request which hits the bug? Thanks!

After the reasoning stage of a reasoning model, during the non-reasoning stage, output string like “I am normal token </I am normal token>” may lose the closing </ token. This issue is reproducible with Qwen/Qwen3-235B-A22B-Thinking-2507.

CatherineSue (approved):

CatherineSue:
@xjpang Thanks for the explanation!

JustinTong0323:
I cherry-picked this PR into #8583 to resolve the bug that arose during model testing, and I've added you as a co-author, thanks for your contribution!

PR #8577, opened 2025-07-30
[Optimization][Perf] Disable the GC during CUDA graph capture to speed up by up to 3x

b8zhong:
## Motivation

Disable the Python GC during the CUDA graph capture, to speed up the capture process and startup time. **Disabling Python GC during CUDA graph capture resulted in a 2.3x–3.7x speedup in capture time.**

Speedup occurs because `gc.freeze()` prevents Python's garbage collector from scanning long-lived objects during CUDA graph capture, eliminating `gc.collect()` overhead.

From https://github.com/vllm-project/vllm/pull/21146, thanks to @mgoin for the original idea!

## Modifications

Freeze the GC.

## Accuracy Test
`lm-eval` is the same.

## Benchmark & Profiling

For Llama4 and Qwen3, fresh container environment on both. I believe the CUDA graph capture is disabled on server shutdown so it does not matter, but I did so anyway. I used the default CUDA graph capture sizes without any overrides.

 ###  Before & After: for `meta-llama/Llama-4-Maverick-17B-128E-Instruct-FP8`
```
Capturing batches (bs=1 avail_mem=8.89 GB): 100%|█████████████████████████████| 23/23 [00:25<00:00,  1.09s/it]
```
After:
```
Capturing batches (bs=1 avail_mem=8.89 GB): 100%|█████████████████████████████| 23/23 [00:10<00:00,  2.25it/s]
```

###  Before & After: for `Qwen/Qwen3-0.6B` (small model)
```
Capturing batches (bs=1 avail_mem=7.61 GB): 100%|█████████████████████████████| 23/23 [00:06<00:00,  3.55it/s]
```

```
Capturing batches (bs=1 avail_mem=7.61 GB): 100%|█████████████████████████████| 23/23 [00:01<00:00, 13.31it/s]
```

### DIFF
--- a/python/sglang/srt/model_executor/cuda_graph_runner.py
+++ b/python/sglang/srt/model_executor/cuda_graph_runner.py
@@ -16,6 +16,7 @@
 from __future__ import annotations
 
 import bisect
+import gc
 import inspect
 import logging
 import os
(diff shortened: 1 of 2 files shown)


ch-wan (approved):
This PR accelerates cuda graph capture by 2x in my local test. Thank you for your contribution.

PR #8583, opened 2025-07-30
model: support Step3V

CatherineSue:
## Motivation

This PR adds the support for Step3VModel.

Co-authored-by: Qiaolin-Yu <liin1211@outlook.com>
Co-authored-by: ispobock <ispobaoke@gmail.com>
Co-authored-by: nnnobody-code <nnnobody@foxmail.com>
Co-authored-by: jimpang <jimpang@tencent.com>

## Modifications

- Add Stepv3 model
- Add tool_call_parser `Step3VDetector`
- Add reasoning parser registry for `stepv3`

## Checklist

- [ ] Format your code according to the [Code Formatting with Pre-Commit](https://docs.sglang.ai/references/contribution_guide.html#code-formatting-with-pre-commit).
- [ ] Add unit tests as outlined in the [Running Unit Tests](https://docs.sglang.ai/references/contribution_guide.html#running-unit-tests-adding-to-ci).
- [ ] Update documentation / docstrings / example tutorials as needed, according to [Writing Documentation](https://docs.sglang.ai/references/contribution_guide.html#writing-documentation-running-docs-ci).
- [ ] Provide throughput / latency benchmark results and accuracy evaluation results as needed, according to [Benchmark and Profiling](https://docs.sglang.ai/references/benchmark_and_profiling.html) and [Accuracy Results](https://docs.sglang.ai/references/accuracy_evaluation.html).
- [ ] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co-author when merging the PR.
- [ ] Please feel free to join our Slack channel at https://slack.sglang.ai to discuss your PR.

PR #8606, opened 2025-07-31
[Bugfix] Reasoning parser must not drop parts of xml tags

sfbemerk:
## Motivation

In streaming mode, the sglang reasoning parser buffers chunks that can be completed to an opening <think> token. Unfortunately, when the reasoning parser finds out that the buffered text was NOT part of a think token, only the new text is returned, but not the full buffer. As a result, xml-like model output that comes in chunks like `"<", "my-tag", ">"` will be scrambled to just `"my-tag", ">"`. 

## Modifications

The reasoning parser now flushes the entire buffer (not just the new text) once it knows that the buffer can not be completed to a think token. 

## Unit Tests

Two tests were added:
- test_parse_streaming_increment_mixed_content_chunks
- test_parse_streaming_increment_mixed_content_chunks_with_xml_tags (this one failed without the suggested modification)

### DIFF
--- a/python/sglang/srt/reasoning_parser.py
+++ b/python/sglang/srt/reasoning_parser.py
@@ -105,7 +105,7 @@ class BaseReasoningFormatDetector:
         # If we're not in a reasoning block return as normal text
         if not self._in_reasoning:
             self._buffer = ""
-            return StreamingParseResult(normal_text=new_text)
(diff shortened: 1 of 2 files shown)


JustinTong0323:
Thanks for finding the bug! However, it's already recognized and would be solved after this pr merge: #8583

PR #8623, opened 2025-07-31
Fix hf3fs_fuse import error

ispobock:
## Motivation
Fix error log:
```
hf3fs_fuse.io is not available: No module named 'hf3fs_fuse'
```


ispobock:
cc: @xiezhq-hermann

PR #8666, opened 2025-08-01
[CPU] Fix issues when running llama3.2-11B vision model with image tasks

blzheng:
## Motivation

This PR fixes several issues encountered when running the LLaMA 3.2 11B vision model with image tasks on CPU.

## Modifications

- Fixed kv cache access logic for cross attention
  - Affected both backends: 
      - python/sglang/srt/layers/attention/torch_native_backend.py 
      - python/sglang/srt/layers/attention/intel_amx_backend.py
  - Updated the CPU kernel to support such case:
      -  sgl-kernel/csrc/cpu/extend.cpp
      -  sgl-kernel/csrc/cpu/decode.cpp
      -  sgl-kernel/csrc/cpu/torch_extension_cpu.cpp
- Removed hard-coded CUDA device in python/sglang/srt/models/mllama.py
- Fixed incorrect attribute access of mm_inputs in python/sglang/srt/models/mllama.py
- Fixed unrecognized chat template parsing in python/sglang/bench_serving.py
- Supported >2D inputs for weight_packed_linear in python/sglang/srt/layers/quantization/unquant.py

## Checklist

---
### CI States

Latest PR Test (Base): :x: [Run #26202393853](https://github.com/sgl-project/sglang/actions/runs/26202393853)
Latest PR Test (Extra): :x: [Run #26202393811](https://github.com/sgl-project/sglang/actions/runs/26202393811)


python/sglang/srt/layers/attention/intel_amx_backend.py
```diff
@@ -49,6 +50,183 @@ def init_forward_metadata(self, forward_batch: ForwardBatch):
...
+                    enable_gqa=enable_gqa,
+                    scale=scaling,
+                    is_causal=causal,
+                )
+                .squeeze(0)
+                .movedim(query.dim() - 2, 0)
+            )
+            output[start_q:end_q, :, :] = per_req_out_redudant[prefill_seq_len_q:, :, :]
+            start_q, start_kv = end_q, end_kv
+        return output
```
mingfeima: My understanding is that this PR implements torch implementation for both `extend` and  `decode` under the backend of `intel_amx_backend`?

This is not very good practice. I recommend:
* upgrade intel_amx_backend to support vl models natively
* upgrade torch native backend to support vl models as well (non-amx path)

mingfeima (changes requested):

mingfeima:
@blzheng you can summarize the topic a little bit and we can discuss internally, what we need to know:

* difference than existing model, e.g. K is none or whatever.
* input shapes and dtypes for each tensor in the targeted model or model(s).

python/sglang/srt/layers/attention/intel_amx_backend.py
blzheng: I plan to first use SDPA (via PyTorch's Flash Attention) for the cross-attention part in `intel_amx_backend`, while keeping the rest on the existing `extend` and `decode` kernels. A follow-up PR will add native cross-attention support to those kernels. Does this approach work for you?

blzheng: I have updated both extend and decode to support cross attention in this model. Please help review again. Thanks~

python/sglang/bench_serving.py
```diff
@@ -879,22 +880,37 @@ def sample_mmmu_requests(
...
+                                                "type": "image_url",
+                                                "image_url": {"url": image_data},
+                                            },
+                                            {"type": "text", "text": prompt},
+                                        ],
+                                    }
+                                ],
+                                add_generation_prompt=True,
+                                tokenize=False,
+                            )
```
mingfeima:
```suggestion
                        if "llama" in tokenizer.name_or_path.lower():
                            image_field = {"type": "image", "image": image_data}
                        else:
                            image_field = {"type": "image_url", "image_url": {"url": image_data}}

                       prompt = tokenizer.apply_chat_template(
                            [
                                 {
                                     "role": "user",
                                      "content": [
                                          image_field,
                                          {"type": "text", "text": prompt},
                                       ],
                                   }
                               ],
                              add_generation_prompt=True,
                              tokenize=False,
                          )
```

try remove the duplicated code.

python/sglang/srt/layers/attention/intel_amx_backend.py
```diff
@@ -62,14 +62,17 @@ def forward_extend(
...
+            if not layer.is_cross_attention
+            else forward_batch.encoder_out_cache_loc
+        )
         if save_kv_cache:
-            forward_batch.token_to_kv_pool.set_kv_buffer(
-                layer, forward_batch.out_cache_loc, k, v
-            )
+            if k is not None:
+                assert v is not None
+                forward_batch.token_to_kv_pool.set_kv_buffer(layer, cache_loc, k, v)
```
mingfeima:
```
if save_kv_cache and k is not None and v is not None:
    forward_batch.token_to_kv_pool.set_kv_buffer(layer, cache_loc, k, v)
```

python/sglang/srt/layers/attention/intel_amx_backend.py
```diff
@@ -62,14 +62,17 @@ def forward_extend(
             o = q.new_empty((q.shape[0], layer.tp_q_head_num * layer.v_head_dim))
         else:
             o = torch.empty_like(q)
-
+        cache_loc = (
+            forward_batch.out_cache_loc
+            if not layer.is_cross_attention
+            else forward_batch.encoder_out_cache_loc
+        )
```
mingfeima, line 106: is this a runtime constant, if yet we can move `cache_loc` to `self.forward_metadata`, which means it will be only executed once, not for every iteration.

python/sglang/srt/layers/attention/intel_amx_backend.py
```diff
@@ -105,23 +110,28 @@ def forward_decode(
...
```
mingfeima, line 152: same as above.

python/sglang/srt/layers/attention/torch_native_backend.py
```diff
@@ -35,9 +35,11 @@ def _run_sdpa_forward_extend(
         seq_lens: torch.Tensor,
         extend_prefix_lens: torch.Tensor,
         extend_seq_lens: torch.Tensor,
+        encoder_lens=None,
```
mingfeima:
```suggestion
        encoder_lens: Optional[torch.Tensor] = None,
```

Use Optional.

python/sglang/srt/layers/attention/torch_native_backend.py
```diff
@@ -75,8 +79,14 @@ def _run_sdpa_forward_extend(
...
-            end_kv = start_kv + seq_len_kv
-
+            if encoder_lens is not None:
+                start_kv = 0 if is_cross_attn else encoder_lens[seq_idx]
+                end_kv = (
+                    encoder_lens[seq_idx] if is_cross_attn else start_kv + seq_len_kv
+                )
+            else:
+                start_kv = 0
+                end_kv = start_kv + seq_len_kv
```
mingfeima, line 91:
```suggestion
if encoder_lens is not None:
    if is_cross_attn:
        start_kv = 0
        end_kv = encoder_lens[seq_idx]
    else:
        start_kv = encoder_lens[seq_idx]
        end_kv = start_kv + seq_len_kv
else:
    start_kv = 0
    end_kv = seq_len_kv
```

python/sglang/srt/layers/attention/torch_native_backend.py
```diff
@@ -151,14 +165,22 @@ def _run_sdpa_forward_decode(
...
             end_q = start_q + seq_len_q
-            end_kv = start_kv + seq_len_kv
+            if encoder_lens is not None:
+                start_kv = 0 if is_cross_attn else encoder_lens[seq_idx]
+                end_kv = (
+                    encoder_lens[seq_idx] if is_cross_attn else start_kv + seq_len_kv
+                )
+            else:
+                start_kv = 0
+                end_kv = start_kv + seq_len_kv
```
mingfeima, line 184: same as above.

python/sglang/srt/layers/attention/torch_native_backend.py
```diff
@@ -192,11 +214,15 @@ def forward_extend(
             o = q.new_empty((q.shape[0], layer.tp_q_head_num * layer.v_head_dim))
         else:
             o = torch.empty_like(q)
-
+        cache_loc = (
+            forward_batch.out_cache_loc
+            if not layer.is_cross_attention
+            else forward_batch.encoder_out_cache_loc
+        )
```
mingfeima: same as above

python/sglang/srt/layers/attention/torch_native_backend.py
```diff
@@ -192,11 +214,15 @@ def forward_extend(
...
         if save_kv_cache:
-            forward_batch.token_to_kv_pool.set_kv_buffer(
-                layer, forward_batch.out_cache_loc, k, v
-            )
+            if k is not None:
+                assert v is not None
+                forward_batch.token_to_kv_pool.set_kv_buffer(layer, cache_loc, k, v)
```
mingfeima: same as above.

python/sglang/srt/layers/attention/torch_native_backend.py
```diff
@@ -217,9 +243,11 @@ def forward_extend(
             forward_batch.seq_lens,
             forward_batch.extend_prefix_lens,
             forward_batch.extend_seq_lens,
+            encoder_lens=forward_batch.encoder_lens,
```
mingfeima:
```suggestion
            forward_batch.encoder_lens,
```

I remember intel python team suggests not use '=' if possible, it is slower a little bit.

python/sglang/srt/layers/attention/torch_native_backend.py
```diff
@@ -240,11 +268,16 @@ def forward_decode(
             o = q.new_empty((q.shape[0], layer.tp_q_head_num * layer.v_head_dim))
         else:
             o = torch.empty_like(q)
+        cache_loc = (
+            forward_batch.out_cache_loc
+            if not layer.is_cross_attention
+            else forward_batch.encoder_out_cache_loc
+        )
```
mingfeima, line 288: same as above.

python/sglang/srt/layers/attention/torch_native_backend.py
```diff
@@ -259,9 +292,11 @@ def forward_decode(
             forward_batch.req_to_token_pool.req_to_token,
             forward_batch.req_pool_indices,
             forward_batch.seq_lens,
+            encoder_lens=forward_batch.encoder_lens,
```
mingfeima: same as above

python/sglang/srt/layers/linear.py
```diff
@@ -196,7 +196,7 @@ def __init__(
 
         if bias:
             self.bias = Parameter(
-                torch.empty(self.output_size, dtype=self.params_dtype)
+                torch.zeros(self.output_size, dtype=self.params_dtype)
```
mingfeima: why we need this?

`empty` is lazy initialization, it is much faster.

python/sglang/srt/layers/quantization/unquant.py
```diff
@@ -116,9 +116,18 @@ def apply(
...
+                x = x.view(-1, layer.weight.shape[-1])
+            output = torch.ops.sgl_kernel.weight_packed_linear(
                 x, layer.weight, bias, True  # is_vnni
             )
+            if len(x_shapes) >= 3:
+                if x_shapes[-1] == layer.weight.shape[-1]:
+                    output = output.view(*x_shapes[:-1], -1)
+                else:
+                    output = output.view(*x_shapes[:-2], layer.weight.shape[0])
+            return output
```
mingfeima, line 156: fix this is python is going to create a lot of python/torch overhead.

i prefer fix it in C++:
* make input: [*, K]
* make output: [*, K]

where * can be 1d to nd.

just do something in C++:

```
const int64_t ndim = input.ndimensions();
const auto input_sizes = input.sizes();
int64_t K = input_sizes[ndim - 1];
int64_t M = input.numel() / K;
int64_t N = weight.size(0);

// get output sizes
input_sizes[ndim - 1] = N;


```

mingfeima (changes requested):
generally OK. just some minor changes needed.

python/sglang/srt/layers/attention/intel_amx_backend.py
blzheng: It is not a runtime constant because it depends on information from the `layer`, which is provided at runtime.

python/sglang/srt/layers/linear.py
blzheng: Some biases are missing in the model weights. They should default to 0, but when initialized with `empty`, the bias may contain unexpected values, leading to accuracy issue.

sgl-kernel/csrc/cpu/extend.cpp
```diff
@@ -433,119 +437,122 @@ void extend_attention_kernel_impl(
...
-
-        // apply causal mask
-        if (num_keys - n <= BLOCK_N) {
-          for (int row = 0; row < m_size; ++row) {
-            int last_col = m + row - n;
-            // fill [last_col + 1, n_size) to -inf
-            float* row_ptr = s_i + row * BLOCK_N;
-            fill_stub(row_ptr + last_col + 1, -std::numeric_limits<float>::infinity(), n_size - last_col - 1);
+      }
+      if (!is_cross_attn) {
```
mingfeima, line 180: so for `is_cross_attn`, we don't have stage2?

mingfeima (approved):

sgl-kernel/csrc/cpu/extend.cpp
blzheng: Yes

mingfeima: shall we do 
```
if (is_cross_attn) {
  return;
}
```

save the indenting diff +- ...

blzheng: We probably cannot return here, because the following post-processing is necessary.
https://github.com/sgl-project/sglang/blob/d08663eec1786424d33f2a7055bd03f51370e867/sgl-kernel/csrc/cpu/extend.cpp#L549-L553

zhyncs:
@blzheng please fix the lint issue, thanks

blzheng:
> @blzheng please fix the lint issue, thanks

Done.

python/sglang/srt/layers/linear.py
zhyncs, line 1: why do we change to zeros here

blzheng: Some biases are missing in the model weights. They should default to 0, but when initialized with empty, the bias may contain unexpected values, leading to accuracy issue.

python/sglang/srt/layers/attention/torch_native_backend.py
```diff
@@ -129,12 +143,14 @@ def _run_sdpa_forward_decode(
             output: [num_tokens, num_heads, head_size]
             k_cache: [max_total_num_tokens, num_heads, head_size]
             v_cache: [max_total_num_tokens, num_heads, head_size]
-            req_to_token: [max_num_reqs, max_context_len]
-            req_pool_indices: [num_seqs]
+            req_to_token: [max_num_reqs, max_context_len],
+            req_pool_indices: [num_seqs],
```
Alcanderian: typo?

python/sglang/srt/layers/linear.py
Alcanderian: > Some biases are missing in the model weights. They should default to 0, but when initialized with empty, the bias may contain unexpected values, leading to accuracy issue.

Can we tell which model?

blzheng: https://huggingface.co/meta-llama/Llama-3.2-11B-Vision-Instruct

python/sglang/srt/layers/attention/torch_native_backend.py
blzheng: Thanks. Fixed.

python/sglang/srt/layers/linear.py
blzheng: I changed the bias initialization to use torch.zeros only on CPU. Does that make sense to you?

test/srt/cpu/test_decode.py
```diff
@@ -18,9 +18,11 @@ def _run_sdpa_forward_decode(
         req_to_token: torch.Tensor,
         req_pool_indices: torch.Tensor,
         seq_lens: torch.Tensor,
+        encoder_lens=None,
```
mingfeima, line 21: No need to change `srt/cpu`, this folder will be removed later.

mingfeima (changes requested):

mingfeima (approved):

python/sglang/srt/layers/linear.py
mingfeima: i am still confused why cpu need this. No worries let me check the code again.

PR #8709, opened 2025-08-02
Fix fused MoE when `routed_scaling_factor is None`

hnyls2002:
Reproduction:

```bash
python3 -m sglang.compile_deep_gemm --host=0.0.0.0 --port=8000 --model=Qwen/Qwen3-235B-A22B-Instruct-2507-FP8 --tp=8 --trust-remote-code --enable-ep-moe --tool-call-parser=qwen25 --reasoning-parser=qwen3 --context-length=131072
```

Issue:

```bash
  File "/root/sglang/python/sglang/srt/layers/moe/ep_moe/layer.py", line 283, in forward_deepgemm
    return output * self.routed_scaling_factor
TypeError: unsupported operand type(s) for *: 'Tensor' and 'NoneType'
```

### DIFF
--- a/python/sglang/srt/layers/moe/fused_moe_triton/layer.py
+++ b/python/sglang/srt/layers/moe/fused_moe_triton/layer.py
@@ -134,9 +134,7 @@ class FusedMoE(torch.nn.Module):
             if not self.enable_flashinfer_cutlass_moe:
                 self.expert_map_gpu = self.expert_map_cpu.to(device="cuda")
 
-        self.routed_scaling_factor = (
-            routed_scaling_factor if routed_scaling_factor is not None else 1.0
-        )
+        self.routed_scaling_factor = routed_scaling_factor
         assert intermediate_size % self.moe_tp_size == 0
         self.intermediate_size_per_partition = intermediate_size // self.moe_tp_size
         self.reduce_results = reduce_results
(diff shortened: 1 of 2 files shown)


python/sglang/srt/layers/moe/fused_moe_triton/layer.py
```diff
@@ -134,7 +134,9 @@ def __init__(
             if not self.enable_flashinfer_cutlass_moe:
                 self.expert_map_gpu = self.expert_map_cpu.to(device="cuda")
 
-        self.routed_scaling_factor = routed_scaling_factor
+        self.routed_scaling_factor = (
+            routed_scaling_factor if routed_scaling_factor is not None else 1.0
```
ch-wan: Would this change incur any performance degradation? See this:

https://github.com/sgl-project/sglang/blob/f9f0138f80a32ecba8a4da619cb51dce2bb3381c/python/sglang/srt/layers/quantization/modelopt_quant.py#L1002-L1003

hebiao064: +1

we can do:


modified: python/sglang/srt/layers/moe/ep_moe/layer.py
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
@ python/sglang/srt/layers/moe/ep_moe/layer.py:283 @ class EPMoE(FusedMoE):
            m_max * self.start_expert_id,
            BLOCK_SIZE=512,
        )
       -- return output * self.routed_scaling_factor
        ++ return output * self.routed_scaling_factor if self.routed_scaling_factor is not None else output

PR #8732, opened 2025-08-03
Fix triton kernels topk with keyword arguments

ispobock:
## Motivation

For the triton kernels topk routing function, the meaning of `sm_first` is not exactly the same as `renormalize`. Use keyword arguments to avoid confusion and unknown bugs.
- If `sm_first` = True, the softmax will before the topk, so the result can be unnormalized before sort. `renormalize` is not supported in triton kernels rounting funciton.
- If `sm_first` = False, the softmax will after the topk, so the result is already normalized, so we don't need to renormalize.

### DIFF
--- a/python/sglang/srt/layers/moe/topk.py
+++ b/python/sglang/srt/layers/moe/topk.py
@@ -649,11 +649,12 @@ def biased_grouped_topk_cpu(
 
 def triton_kernels_topk(
     router_logits: torch.Tensor,
-    topk: torch.Tensor,
+    topk: int,
     renormalize: bool = False,
     sm_first: bool = False,
-):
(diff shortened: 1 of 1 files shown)


zhyncs (approved):

ch-wan:
@zhyncs @ispobock  This PR should be reverted. The original implementation was correct. `sm_first=True` is handled internally in `routing` (https://github.com/triton-lang/triton/blob/8f9f695ea8fde23a0c7c88e4ab256634ca27789f/python/triton_kernels/triton_kernels/routing.py#L294-L297)

ispobock:
As discussed with @ch-wan, if `topk-softmax` <=> `softmax-topk-renormalize` , then `not sm_first` = `renormalize` (keep output sum to one). We can just set `sm_first` = `not renormalize`.

Issue #8433, opened 2025-07-28
[Feature] Add cache for Multimoal  input data


kousakawang:
### Checklist

- [x] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 2. Please use English, otherwise it will be closed.

### Motivation

When processing  Multimodal model 's Input data(image, vedio) , an additional time cost was obeserved.

And after profiling, we found the following two steps cost about 100ms for one 1280x768 image.

[two_steps](https://github.com/mickqian/sglang/blob/54a8f315196200d7e81c76ade0a745d4de543249/python/sglang/srt/managers/tokenizer_manager.py#L801C17-L801C78)

I noticed that this PR https://github.com/sgl-project/sglang/pull/6003 try to solve the data send efficency problem by using cuda IPC.   This step can  reduce  the data send time while the data processing time is still existing. 

so I want to add a cache Feature for multimidal data ,  which is especially needed when we use multimodal model in a AI agent scence. In this case , every step more than one images will be  used as model's input and among these datas ususally only one data is new and the others are the history data.

PR #8742, opened 2025-08-04
[Feature] : add mm input item cache support

kousakawang:
## Modifications
related to issue: https://github.com/sgl-project/sglang/issues/8433

### Image Input Data Cache & CUDA IPC Support

This feature introduces **image input caching** and **CUDA IPC-based transmission** for the `qwen2-image-processor` and `glm4v-image-processor`.  
It is designed to reduce redundant computation and minimize inter-process communication overhead.

#### 🔄 Image Cache
- If an image has already been processed, it will be **retrieved from cache** instead of being processed again.
- Avoids repeated preprocessing of the same image.
- Stored tensors remain in the `image_processor`’s internal cache table.

#### ⚡ CUDA IPC Transmission
- Replaces the default **raw tensor transmission via ZMQ** with a **CUDA IPC handle dictionary**.
- Reduces transmission time for large `pixel_values` tensors.
- **Important:** CUDA IPC is only used when caching is enabled, ensuring that cached tensors are not recycled by PyTorch’s memory manager.

### Use Cases

- **AI Agents**  
  Multiple images are often passed as input, but usually only one is new. Caching prevents redundant preprocessing of repeated images.

- **Robotics / Navigation**  
  Frequent re-use of similar frames can benefit from fast cache retrieval and IPC-based transmission.

### How to Enable

```bash
export SGL_CACHE_MM_IMAGE=1
# set cache max tokenizer cache size,  or will use default value or max-available memorysize)
export SGL_TOKENIZER_CACHED_IMAGE_SIZE_MB=1024 #(max cache 1024MB image data)

```

### [TODO]
~~mange cache by using MB_SIZE instead of  image nums~~ (done)

### Notes

* Works with both qwen2-image-processor and glm4v-image-processor.

* CUDA IPC is automatically applied only in cache mode.

* Designed for scenarios where image reuse is common, improving both speed and efficiency.

### Ut for this Feature
python3 -m unittest test_vision_openai_server_a.TestQwen2_5_AI_AGENT.test_ai_agent_opt

### test_result: 

**Accuray: (no change with Feature On/Off)**

Action_Summary: yes, I understand. firstly, I need to open the browser
Action: click(start_box="<bbox>367 7   

Action_Summary: yes, I understand. firstly, I need to open the browser
Action: click(start_box="<bbox>367 7

**Performance:**
opt : 709.1031 ms  (with  cache)
normal:  2222.0058  ms (wo cache)

## Checklist

## Summary by CodeRabbit

- New Features
  - Added multimodal image caching for GLM-4v and Qwen VL processors with async support and env-based controls.
  - Enabled handling of dict-based image features during batching.
- Performance
  - Faster repeated image requests via GPU-backed cache and lightweight image hashing, reducing latency and memory overhead.
- Bug Fixes
  - Improved stability by safely handling optional model attributes to prevent runtime errors.
- Tests
  - Introduced end-to-end tests comparing cached vs. non-cached flows, validating identical outputs and measurable latency improvements.


python/sglang/srt/managers/schedule_batch.py
```diff
@@ -1270,6 +1270,40 @@ def prepare_for_extend(self):
...
+
+                    real_pixel_values = []
+                    take_start_idx = 0
+                    take_end_idx = 0
+                    take_idx = 0
+                    for send_idx in range(len(use_cache_marks)):
+                        if use_cache_marks[send_idx]:
+                            real_pixel_values.append(
+                                self.mminput_cache[hash_keys[send_idx]]
+                            )
```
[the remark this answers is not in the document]
kousakawang: done

python/sglang/srt/multimodal/processors/base_processor.py
```diff
@@ -223,18 +325,179 @@ def process_mm_data(
...
+                            device=input_ids.device,
+                        )
+                        new_tensor[0][:pos] = input_ids[0][:pos]
+                        new_tensor[0][pos : pos + len(insert_ids)] = torch.tensor(
+                            insert_ids, dtype=input_ids.dtype, device=input_ids.device
+                        )
+                        new_tensor[0][pos + len(insert_ids) :] = input_ids[0][pos + 1 :]
+                        return new_tensor
+
+                return input_ids
```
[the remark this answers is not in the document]
kousakawang: Done

python/sglang/srt/multimodal/processors/base_processor.py
```diff
@@ -9,11 +10,110 @@
...
+
+
+def image_to_int(img: Image.Image) -> int:
+    img_bytes = img.tobytes()
+    hash_obj = hashlib.md5(img_bytes)
+    hash_bytes = hash_obj.digest()
+    return int.from_bytes(hash_bytes, byteorder="big")
+
+
+class FIFOHashTable:
```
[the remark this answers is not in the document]
kousakawang: Done

python/sglang/srt/multimodal/processors/base_processor.py
```diff
@@ -223,18 +325,179 @@ def process_mm_data(
...
+                            insert_ids, dtype=input_ids.dtype, device=input_ids.device
+                        )
+                        new_tensor[0][pos + len(insert_ids) :] = input_ids[0][pos + 1 :]
+                        return new_tensor
+
+                return input_ids
+
+            to_replace_str = "<|vision_start|><|image_pad|><|vision_end|>"
+            repalce_str = "<|vision_end|>"
+            v_start_token, img_pad_token, v_end_token = 151652, 151655, 151653
```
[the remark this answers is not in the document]
kousakawang: done

python/sglang/srt/multimodal/processors/base_processor.py
```diff
@@ -233,18 +362,152 @@ def process_mm_data(
...
+                        ]
+                    )
+                    use_cache_mark.append(False)
+                # add input ids and insert tensor
+                else:
+                    cached_tensor = self.hash_table.get(img_hash_keys[img_idx])
+                    assert isinstance(
+                        cached_tensor, torch.Tensor
+                    ), "invalid cached_tensor"
+                    # tensor_lists.append(cached_tensor)
```
byjiang1996: Is there any reason why `self.hash_table's` value is not used but eventually use the value from `mminput_cache`?

It seems `self.hash_table` alone would suffice

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -1290,6 +1292,40 @@ def prepare_for_extend(self):
                 pixel_values = getattr(mm_item, "feature", None)
                 if isinstance(pixel_values, torch.Tensor):
                     mm_item.feature = pixel_values.to(self.device, non_blocking=True)
+                elif isinstance(pixel_values, dict):
```
byjiang1996, line 1236: Is there any specific reason why this logic cannot be moved over into base_processor.py?

python/sglang/srt/multimodal/processors/base_processor.py
```diff
@@ -233,18 +362,152 @@ def process_mm_data(
...
+                img_token_num = int(img_height / QWEN2_PER_TOKEN_PATCH_NUM)
+
+                if img_token_num < 1:
+                    raise ValueError("invalid img token num")
+
+                if self.hash_table.get(hash_key) is None:
+                    new_processed_img_idxes.append(img_idx)
+                    new_processed_imgs.append(images[img_idx])
+                else:
+                    remove_image_idx.append(img_idx)
```
byjiang1996: Is there any specific reason why cache is used only once? (not sure if i understood correctly)

python/sglang/srt/multimodal/processors/base_processor.py
```diff
@@ -233,18 +362,152 @@ def process_mm_data(
...
+            # result["pixel_values"] = torch.concat(tensor_lists, dim = 0)
+            send_pixel_values = (
+                torch.concat(tensor_lists, dim=0) if len(tensor_lists) > 0 else None
+            )
+            proxy_pixel_values = {}
+            proxy_pixel_values["send_data"] = send_pixel_values
+            proxy_pixel_values["use_cache_mark"] = use_cache_mark
+            proxy_pixel_values["hash_keys"] = img_hash_keys
+            proxy_pixel_values["img_heights"] = processed_img_heights
+            proxy_pixel_values["delete_keys"] = delete_keys
```
byjiang1996: Similar to this [comment](https://github.com/sgl-project/sglang/pull/8742/files#r2284122037), I'm wondering why we need `proxy_pixel_values` but can't set `pixel_values` correctly within base_processor.py

python/sglang/srt/multimodal/processors/base_processor.py
```diff
@@ -233,18 +362,152 @@ def process_mm_data(
...
+                # add input ids and insert tensor
+                else:
+                    cached_tensor = self.hash_table.get(img_hash_keys[img_idx])
+                    assert isinstance(
+                        cached_tensor, torch.Tensor
+                    ), "invalid cached_tensor"
+                    # tensor_lists.append(cached_tensor)
+                    insert_cached_ids = (
+                        [v_start_token]
+                        + img_token_nums[img_idx] * [img_pad_token]
```
byjiang1996: I'm wondering if there is any `result['xx']` field that can be used directly to obtain img_token_num instead of calculating `img_token_num` manually

python/sglang/srt/multimodal/processors/base_processor.py
```diff
@@ -233,18 +362,152 @@ def process_mm_data(
...
+            delete_keys = []
+            for img_idx in range(len(images)):
+                # cache Tensor
+                if img_idx in new_processed_img_idxes:
+                    img_height = img_heights[img_idx]
+                    processed_img_heights.append(img_height)
+             
+                    start_height = end_height
+                    end_height = start_height + img_height
+                    to_cache_tensor = result["pixel_values"][start_height:end_height]
```
byjiang1996: Similar comments as https://github.com/sgl-project/sglang/pull/8742/files#r2284154152

Is there any result['xx'] field to use instead of calculating `img_height/start_height/end_height` manually?

python/sglang/srt/multimodal/processors/base_processor.py
```diff
@@ -233,18 +362,152 @@ def process_mm_data(
...
+            transformers.models.qwen2_vl.image_processing_qwen2_vl_fast.Qwen2VLImageProcessorFast,
         )
-        # move feature tensors to cpu
-        for feature_name in self.FEATURE_NAMES:
-            if feature_name in result and isinstance(
-                result[feature_name], torch.Tensor
-            ):
-                result[feature_name] = result[feature_name].to("cpu")
+
+        if cache_mm_image_items and is_qwen2_processor:
```
byjiang1996: Is this logic universal to most VLMs? (seems not from my limited understanding)

If no, could you please move this logic into qwen vl processor's processor adapter similar to what we did for [phi4-mm](https://github.com/sgl-project/sglang/blob/7e8187e00428d292b98526ba7b78f95a30f05fba/python/sglang/srt/multimodal/processors/phi4mm.py#L18)? That way this model specfic code can be moved over to model.py instead of in common code places

python/sglang/srt/multimodal/processors/base_processor.py
```diff
@@ -9,11 +10,137 @@
...
+
+
+def image_to_int(img: Image.Image) -> int:
+    img_bytes = img.tobytes()
+    hash_obj = hashlib.md5(img_bytes)
+    hash_bytes = hash_obj.digest()
+    return int.from_bytes(hash_bytes, byteorder="big")
+
+
+class FIFOHashTable:
```
byjiang1996: Similar comments as https://github.com/sgl-project/sglang/pull/8742/files#r2284122880

If we can use cache more than once, we should consider `@lru_cache` i think

python/sglang/srt/managers/schedule_batch.py
kousakawang: I think in base_processor.py, we just load the image (raw or base64) and convert it to tensor(which will be accepted by model as input) , this action happen in tokenizer process.   After this  tokenizer process will send data to scheduler process by using ZMQ. As a result, if we only cache data in tokenizer process, the scheduler can not get the correct pixel values.

python/sglang/srt/multimodal/processors/base_processor.py
kousakawang: yes, you are right, this  logic just for Qwen2VLImageProcessorFast,  I will try to move code to qwen2.

python/sglang/srt/multimodal/processors/base_processor.py
kousakawang: Not sure I undrerstand your question correctly.  If you mean why do “remove_image_idx.append(img_idx)” here， this is for rocording which image was alredy cached, and then the coresponding image token  will be removed from input text to avoiding processsing image data err(ths input_txt should hold  same image nums which is equal to len(new_processed_imgs)

python/sglang/srt/multimodal/processors/base_processor.py
kousakawang: yes, I think we can store the height of each image in to a struct and next time we can use it directly. But for an new image , still we need to calculate it. just because each image hold different size and we can not use a  fixed value. (The more imporant reason is that , we can not modify the logic in  "processor.__call__" which was  realized in transformers not sglang...)

python/sglang/srt/multimodal/processors/base_processor.py
kousakawang: As this comment  https://github.com/sgl-project/sglang/pull/8742/files#r2284939106 said: 
mminput_cache was manged by scheduler subprocess,
and self.hash_table was managed by tokenizer subprocess.
Inter-process communication is done through ICP (each process uses a different port) via the ZMQ library.
so we cache the image data in tokenizer process  aims at :  

1. avoiding processing the same image again, 
2. and also saving the data-send cost between tokenizer process and scheduler process.

To avoid sending a pixel-values(torch.tensor) again, we need to established mm_cache in scheduler process, also. (and this can be a gpu tensor cache)
By doing so , for a normal image, the normal processing route is GPU-> cpu ->send via ZMQ -> cpu ->GPU
for a cached image, we can process it  via raw_image or bas64 -> send hash -> getGPU tensor to avoid send any real pixel values or do any D2H H2D.

Actually, I think the IPC-handle is a better way to avoid sending data but share gpu memory between differnt process, ( in this case, cache is not necessary)
But since get ipc-handle will not increase the ref count of torch's tensor, so without keeping all data always active (by a global container or someting) , access tensor via ipc-handle in scheduler process will have some risk.  (also, ipchandle can only avoiding the send cost, but not the image processing cost)

python/sglang/srt/multimodal/processors/base_processor.py
kousakawang: This is because we can not modify the result["pixel_value"] which is processed via transformer's processor.

kousakawang:
close since cuda-ipc feature  was merged in #11917

PR #8690, opened 2025-08-02
[2/2] Fuse routed scaling factor into select_experts

trevor-m:
## Motivation

See https://github.com/sgl-project/sglang/pull/8364

Requires https://github.com/sgl-project/sglang/pull/8770 [1/2][resubmit] sgl-kernel: Fuse routed scaling factor into moe_fused_gate (select_experts)

## Modifications

See https://github.com/sgl-project/sglang/pull/8364

## Accuracy Test

See https://github.com/sgl-project/sglang/pull/8364

## Benchmark & Profiling

See https://github.com/sgl-project/sglang/pull/8364

## Checklist

- [ ] Format your code according to the [Code Formatting with Pre-Commit](https://docs.sglang.ai/references/contribution_guide.html#code-formatting-with-pre-commit).
- [ ] Add unit tests as outlined in the [Running Unit Tests](https://docs.sglang.ai/references/contribution_guide.html#running-unit-tests-adding-to-ci).
- [ ] Update documentation / docstrings / example tutorials as needed, according to [Writing Documentation](https://docs.sglang.ai/references/contribution_guide.html#writing-documentation-running-docs-ci).
- [ ] Provide throughput / latency benchmark results and accuracy evaluation results as needed, according to [Benchmark and Profiling](https://docs.sglang.ai/references/benchmark_and_profiling.html) and [Accuracy Results](https://docs.sglang.ai/references/accuracy_evaluation.html).
- [ ] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co-author when merging the PR.
- [ ] Please feel free to join our Slack channel at https://slack.sglang.ai to discuss your PR.

PR #8770, opened 2025-08-04
[1/2][resubmit] sgl-kernel: Fuse routed scaling factor into moe_fused_gate (select_experts)

trevor-m:
## Motivation

Resubmit of https://github.com/sgl-project/sglang/pull/8364 which was missing changes required for the unit test.

https://github.com/sgl-project/sglang/pull/8690 will enable using this fusion for deepseek.

Prefill:
10.46% speedup at BS 1
1.86% speedup at BS 128
Decode:
1.22% speedup at BS1
0.26% speedup at BS128

### DIFF
--- a/python/sglang/srt/layers/moe/topk.py
+++ b/python/sglang/srt/layers/moe/topk.py
@@ -132,6 +132,7 @@ class TopK(CustomOp):
         scoring_func: str = "softmax",
         correction_bias: Optional[torch.Tensor] = None,
         routed_scaling_factor: Optional[float] = None,
+        apply_routed_scaling_factor_on_output: Optional[bool] = False,
     ):
         # NOTE: scoring_func is not used for now, but we keep it for future use
         # see https://github.com/sgl-project/sglang/pull/4505 for more details
@@ -147,6 +148,7 @@ class TopK(CustomOp):
         self.custom_routing_function = custom_routing_function
         self.correction_bias = correction_bias
         self.routed_scaling_factor = routed_scaling_factor
+        self.apply_routed_scaling_factor_on_output = apply_routed_scaling_factor_on_output
 
         self.use_triton_kernels = global_server_args_dict["enable_triton_kernel_moe"]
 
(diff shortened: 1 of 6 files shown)


trevor-m:
@hnyls2002 I have fixed the unit test

python/sglang/srt/layers/moe/topk.py
```diff
@@ -422,6 +425,8 @@ def grouped_topk_gpu(
             else topk_weights[:, :-1].sum(dim=-1, keepdim=True)
         )
         topk_weights = topk_weights / topk_weights_sum
+        if apply_routed_scaling_factor_on_output:
+            topk_weights *= routed_scaling_factor
```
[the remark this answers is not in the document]
wenscarl: Maybe skip if routed_scaling_factor == 1?

python/sglang/srt/layers/moe/topk.py
```diff
@@ -519,6 +525,8 @@ def biased_grouped_topk_impl(
...
```
wenscarl, line 532: ditto

python/sglang/srt/layers/moe/topk.py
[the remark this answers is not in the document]
trevor-m: I don't think any model would have a routed scaling factor of 1 since it would be a no-op, I think it just wouldn't have it in which case apply_routed_scaling_factor_on_output would be false anyway

python/sglang/srt/layers/moe/topk.py
```diff
@@ -183,6 +183,7 @@ def forward_cuda(
         *,
         num_token_non_padded: Optional[torch.Tensor] = None,
         expert_location_dispatch_info: Optional[ExpertLocationDispatchInfo] = None,
+        apply_routed_scaling_factor_on_output: Optional[bool] = False,
```
ch-wan: this should be topk config rather than topk input

trevor-m: Thanks @ch-wan, fixed.

kushanam (approved):

trevor-m:
Hi @zhyncs, may I know why this was reverted?

zhyncs:
It broke the uts

trevor-m:
What is uts - is it unit tests? Can you point me to the error?

zhyncs:
@trevor-m we have synced at slack channel before

PR #8133, opened 2025-07-18
[Performance][PD Disaggregation] optimize TokenToKVPoolAllocator by sorting free pages

YiXR:
## Motivation

In large-scale distributed inference, especially with PD Disaggregation, fast and efficient management of KV cache token indices is crucial. Fragmentation of available indices can disrupt peer-to-peer (p2p) data transfers by making it harder to allocate large contiguous regions, resulting in inefficient communication patterns. This fragmentation directly increases end-to-end latency, especially the time to first token (TTFT).
This PR introduces release cache mechanisms to optimize TokenToKVPoolAllocator page management. By deferring the actual release of freed pages and supporting batch reclamation, this design significantly reduces allocation contention and fragmentation during high-concurrency operations, thereby boosting p2p transfer efficiency and reducing TTFT.

### Introduced release_pages
Freed token indices are now buffered in a temporary cache rather than immediately added to the free pool.

### On allocation, merging and sorting
When an allocation request cannot be satisfied by the current free pool alone, the allocator will merge the buffered indices from release cache with the free pool, sort them, and then retry the allocation. This sorting and merging process helps to coalesce adjacent free regions, mitigating fragmentation.

### Test on 2 * A10 * 8（RDMA），qwen 0.6b，1P1D

Case 1（200 warmup；400 nr；16 concurrency; 1024 input len; 512 output len）
without optimization:  TTFT 81ms
with optimization: TTFT 69ms（improve 14.8%）

Case 2（200 warmup；400 nr；16 concurrency; 2048 input len; 1024 output len）
without optimization:  TTFT 144ms
with optimization: TTFT 108ms（improve 25%）

A simple comparison of the number of kv_blocks that need to be transferred by send_kvcache() (1024 in; 512 out)
sglang

#### 1024 input len; 512 output len without optimization
Backend:                                 sglang
Traffic request rate:                    inf
Max request concurrency:                 16
Successful requests:                     400
Benchmark duration (s):                  188.20
Total input tokens:                      409600
Total generated tokens:                  204800
Total generated tokens (retokenized):    204796
Request throughput (req/s):              2.13
Input token throughput (tok/s):          2176.35
Output token throughput (tok/s):         1088.18
Total token throughput (tok/s):          3264.53
Concurrency:                             15.98
----------------End-to-End Latency----------------
Mean E2E Latency (ms):                   7517.90
Median E2E Latency (ms):                 7501.12
---------------Time to First Token----------------
Mean TTFT (ms):                          81.29
[...]

PR #8794, opened 2025-08-05
[Optimization] Update estimated_num_new_pages logic in TokenToKVPoolAllocator

YiXR:
## Motivation

This PR update estimated_num_new_pages logic and format code which is introduced by the previous PR: https://github.com/sgl-project/sglang/pull/8133

## Checklist


python/sglang/srt/mem_cache/allocator.py
```diff
@@ -79,6 +79,9 @@ def free_group_end(self):
         if self.free_group:
             self.free(torch.cat(self.free_group))
 
+    def estimated_num_new_pages(self, bs, extend_num_tokens):
+        return bs * ((extend_num_tokens + self.page_size - 1) // self.page_size)
```
ShangmingCai, line 85: Maybe add a note to clarify the logic here.

ShangmingCai (approved):
LGTM.

python/sglang/srt/mem_cache/allocator.py
YiXR: Here I use upper bound `bs * ((extend_num_tokens + self.page_size - 1) // self.page_size)`
instead of accurate value `((seq_lens + self.page_size - 1) // self.page_size - (prefix_lens + self.page_size - 1) // self.page_size).sum().item()` 
to  reduce the computational overhead caused by each estimate. The increase in the frequency of reordering caused by using the upper bound is very small and can be ignored.

merrymercy (commented):
Looks good to me.
It seems this sort is only useful for PD case. Should we add a flag so that it only turns on during PD?

YiXR:
> Looks good to me. It seems this sort is only useful for PD case. Should we add a flag so that it only turns on during PD?

done

xiezhq-hermann (approved):

python/sglang/srt/mem_cache/allocator.py
```diff
@@ -433,8 +453,9 @@ def alloc(self, need_size: int):
             ), "The allocation size should be page-aligned"
 
         num_pages = need_size // self.page_size
-        if num_pages > len(self.free_pages):
-            self.merge_and_sort_free()
+        if self.need_sort:
+            if num_pages > len(self.free_pages):
+                self.merge_and_sort_free()
```
ShangmingCai: nit: should we use `and` instead of another `if`?

YiXR: fixed

PR #8820, opened 2025-08-05
[Do Not Merge] Test capacity.

deedongala:
This PR is merely for only testing new capacity that we are migrating the mi325 runners to.


saienduri:
Testing and migration completed, so closing PR

PR #8866, opened 2025-08-06
[PD]decode: add CLIP_MAX_NEW_TOKEN for pop_preallocated

jimmy-evo:
add a env CLIP_MAX_NEW_TOKEN to clip params.max_new_tokens. Too many req's max_token is large but finish with a short generation.

## Motivation

Numerous requests generate merely minimal tokens (<1k), yet the `sampling_param.max_new_token` is configured at an excessively large value (120k). This configuration results in substantial GPU memory wastage and renders the `unbootstrap_queue` severely congested.

## Modifications

Users can specify a clipped value for param.max_new_token based on actual requirements, setting it to an appropriate threshold to mitigate wastage.

## Benchmark & Profiling

As for deepseek V3(Avg output length = 190), I clip max_new_token to 256. then no more unbootstraped request in prefill. and decode can process more requests.

## Checklist


cctry:
In my opinion, while this is a valid concern, the proposed solution is not fully effective.

1. Setting this env var is akin to overriding max_new_tokens: both rely on estimating output lengths, but the former is risky to hangs. It at least needs to warn the user if set. 
2. For running requests that generate far fewer tokens than their limit, they will complete quickly anyway, freeing up space and unblocking prefill bootstrapping

ByronHsu (approved):

jimmy-evo:
> 2. For running requests that generate far fewer tokens than their limit, they will complete quickly anyway, freeing up space and unblocking prefill bootstrapping

@cctry 
In the PD disag scenario, the `decode` must ensure sufficient memory is available (which called pre_allocate) before initiating `bootstrap` `prefill`. Even if requests are generated quickly, they may get stuck at preallocation and fail to begin prefill entirely.

python/sglang/srt/disaggregation/decode.py
```diff
@@ -374,7 +378,6 @@ def pop_preallocated(self) -> List[DecodeRequest]:
                 break
 
             # Memory estimation: don't add if the projected memory cannot be met
-            # TODO: add new_token ratio
```
hnyls2002: There is still a TODO task here

hnyls2002 (changes requested):

hnyls2002:
I think you'd better reuse the env var from the non-pd-disaggregation mode `SGLANG_CLIP_MAX_NEW_TOKENS_ESTIMATION`. A server cannot be pd-disagg and non-pd-diagg at the same time.

jimmy-evo:
> I think you'd better reuse the env var from the non-pd-disaggregation mode `SGLANG_CLIP_MAX_NEW_TOKENS_ESTIMATION`. A server cannot be pd-disagg and non-pd-diagg at the same time.

@hnyls2002  I agree. I do use `SGLANG_CLIP_MAX_NEW_TOKENS_ESTIMATION` in my dev env.

PR #8539, opened 2025-07-29
Use reduce scatter for DP

trevor-m:
## Motivation

Similar to how https://github.com/sgl-project/sglang/pull/8280 enables all-gather for DP when using padding, this PR enables reducescatter instead of all-reduce following MOE/MLP layers in Deepseek.

## Modifications

If DP with padding is used, all-reduce is skipped after MoE and MLP. The layer communicator will perform reduce scatter on hidden states instead of just scatter.

Currently this is implemented for deepseek, but we can extend this easily to other models that use `LayerCommunicator` later.

## Accuracy Test

```
python3 -m sglang.launch_server --model-path nvidia/DeepSeek-R1-0528-FP4 --trust-remote-code --quantization modelopt_fp4 --tp 8 --enable-flashinfer-cutlass-moe --enable-ep-moe --ep-size 8 --dp 8 --enable-dp-attention
python3 benchmark/gsm8k/bench_sglang.py --num-shots 8 --num-questions 1319 --parallel 1319 --port=30000
Accuracy: 0.959
Invalid: 0.000
Latency: 20.852 s
Output throughput: 6857.902 token/s
```

## Benchmark & Profiling

```
python3 -m sglang.launch_server --model-path nvidia/DeepSeek-R1-0528-FP4 --trust-remote-code --quantization modelopt_fp4 --tp 8 --enable-flashinfer-cutlass-moe --enable-ep-moe --ep-size 8 --dp 8 --enable-dp-attention
python3 -m sglang.bench_serving --backend sglang --dataset-name random --num-prompt 1024 --random-input 1024 --random-output 1024 --random-range-ratio 1 --max-concurrency 1024
```

with PR:
```
============ Serving Benchmark Result ============
Backend:                                 sglang
Traffic request rate:                    inf
Max request concurrency:                 1024
Successful requests:                     1024
Benchmark duration (s):                  70.14
Total input tokens:                      1048576
Total generated tokens:                  1048576
Total generated tokens (retokenized):    1046022
Request throughput (req/s):              14.60
Input token throughput (tok/s):          14949.45
Output token throughput (tok/s):         14949.45
Total token throughput (tok/s):          29898.91
Concurrency:                             1020.88
----------------End-to-End Latency----------------
Mean E2E Latency (ms):                   69927.88
Median E2E Latency (ms):                 69911.65
---------------Time to First Token----------------
Mean TTFT (ms):                          9383.06
Median TTFT (ms):                        9239.48
P99 TTFT (ms):                           17455.72
[...]

PR #8875, opened 2025-08-06
[bugfix] Fix storage of hidden states with "layernorm before all gather"

trevor-m:
## Motivation

Alternative fix to https://github.com/sgl-project/sglang/pull/8785

https://github.com/sgl-project/sglang/pull/8631 added an optimization to perform layernorm before all-gather for DP.
Layernorm allocates a new tensor for it's output. Now that layernorm is before AG, the hidden states stay in gathered_buffer which can causes the following error later when global and local tokens point to the same buffer. This only applies to certain backends where the MLP is in-place.

```
  File "/opt/mycode/python/sglang/srt/layers/communicator.py", line 237, in postprocess_layer
    return self._communicate_summable_tensor_pair_fn(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/mycode/python/sglang/srt/layers/communicator.py", line 544, in _scatter_hidden_states
    dp_scatter(hidden_states, global_hidden_states, forward_batch)
  File "/opt/mycode/python/sglang/srt/layers/dp_attention.py", line 350, in dp_scatter
    local_tokens.untyped_storage() is not global_tokens.untyped_storage()
AssertionError: aliasing between local_tokens and global_tokens not allowed
```

## Modifications

1. Small optimization to use fused_add_rmsnorm for layernorm before AG case.
2. Clone hidden states for layernorm before AG case so that they won't alias to gathered_buffer.

## Accuracy Test

```
SGL_ENABLE_JIT_DEEPGEMM=0 python3 -m sglang.launch_server  --model-path=deepseek-ai/DeepSeek-R1-0528 --trust-remote-code --attention-backend trtllm_mla --page-size 64 --tp-size 8 --max-running-requests 64 --cuda-graph-max-bs 64 --enable-dp-attention --dp-size 8 --mem-fraction-static 0.90 --dist-timeout 1800
```

```
Accuracy: 0.964
Invalid: 0.000
Latency: 169.495 s
Output throughput: 832.945 token/s
```

## Checklist

### DIFF
--- a/python/sglang/srt/layers/communicator.py
+++ b/python/sglang/srt/layers/communicator.py
@@ -402,15 +402,13 @@ class CommunicateWithAllReduceAndLayerNormFn:
             )
             attn_tp_all_gather_into_tensor(residual, local_residual)
         if context.attn_dp_size != 1:
-            if context.attn_tp_rank == 0:
-                hidden_states += residual
-
             # Perform layernorm on smaller data before comm. Only valid when attn_tp_size is 1 (tp_size == dp_size)
             use_layer_norm_before_gather = context.attn_tp_size == 1
             if use_layer_norm_before_gather:
-                residual.copy_(hidden_states)
                 if hidden_states.shape[0] != 0:
-                    hidden_states = layernorm(hidden_states)
+                    hidden_states, residual = layernorm(hidden_states, residual)
+            elif context.attn_tp_rank == 0:
+                hidden_states += residual
 
             hidden_states, local_hidden_states = (
                 forward_batch.gathered_buffer,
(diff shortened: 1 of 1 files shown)


python/sglang/srt/layers/communicator.py
```diff
@@ -422,6 +420,12 @@ def _gather_hidden_states_and_residual(
                 dp_scatter(residual, hidden_states, forward_batch)
                 if hidden_states.shape[0] != 0:
                     hidden_states = layernorm(hidden_states)
+            else:
+                # CommunicateSummableTensorPairFn._scatter_hidden_states will
+                # copy from global_hidden_states to hidden_states. If we don't
+                # have this clone, they will both point to gathered_buffer.
+                # Normally, layernorm would perform this clone.
+                hidden_states = hidden_states.clone()
```
ch-wan: Can we use `torch.empty_like(forward_batch.gathered_buffer)` in line 414 to avoid this clone? You may refer to #7705

trevor-m: Thanks, let me try that

trevor-m: I found that https://github.com/sgl-project/sglang/pull/8539 fixes this issue by avoiding dp_scatter, so this fix won't be necessary.
Edit: reduce scatter only used with padding. error still occurs for non-padded mode

trevor-m: Pushed a fix with `empty_like()`. I prefer to do this in `_scatter_hidden_states` because the new buffer we create is smaller than if we were to do in line 414. We also only need to create the new buffer if the storage is the same which is only true for a few cases

trevor-m:
Closing because https://github.com/sgl-project/sglang/pull/8785 addresses the issue

PR #8238, opened 2025-07-21
Add support for NCCL symmetric memory for TP allreduces

nvcastet:
## Motivation

Add support for [NCCL symmetric memory](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/bufferreg.html#window-registration).
This PR only enables TP allreduce. Follow up will enable other collectives.

NCCL symmetric memory uses a separate memory pool to allocate tensors that will be used for communications, therefore it can lead to more memory fragmentation. This new feature is off by default and can be turned on with `--enable-symm-mem` server argument.

## FP8
Baseline:
```
SGLANG_CUTLASS_MOE=1 SGLANG_ENABLE_FLASHINFER_GEMM=1 python3 -m sglang.launch_server --tp=8 --trust-remote-code --disable-radix-cache --model-path=deepseek-ai/DeepSeek-R1-0528 --attention-backend=cutlass_mla
python3 -m sglang.bench_serving --backend sglang --dataset-name random --num-prompt 512 --random-input 1024 --random-output 1024 --random-range-ratio 1 --max-concurrency 512 --warmup-requests 512
```
`6671.94 tok/s`

with this PR:
```
SGLANG_CUTLASS_MOE=1 SGLANG_ENABLE_FLASHINFER_GEMM=1 python3 -m sglang.launch_server --tp=8 --trust-remote-code --disable-radix-cache --model-path=deepseek-ai/DeepSeek-R1-0528 --attention-backend=cutlass_mla --enable-symm-mem
python3 -m sglang.bench_serving --backend sglang --dataset-name random --num-prompt 512 --random-input 1024 --random-output 1024 --random-range-ratio 1 --max-concurrency 512 --warmup-requests 512
```
`7129.24 tok/s`

### Performance with concurrency of `256`:
Baseline:
```
python3 -m sglang.launch_server --model-path nvidia/DeepSeek-R1-FP4 --trust-remote-code --quantization modelopt_fp4 --tp 8 --enable-flashinfer-cutlass-moe  --enable-ep-moe --ep-size 8 --max-running-requests 256 --disable-radix-cache --mem-fraction-static 0.9 --stream-output --cuda-graph-bs 256  --cuda-graph-max-bs 256
python3 -m sglang.bench_serving --backend sglang --dataset-name random --num-prompt 256 --random-input 1024 --random-output 1024 --random-range-ratio 1 --max-concurrency 256
```
`5286.31 tok/s`

with this PR:
```
[...]

PR #8934, opened 2025-08-07
Register allgather/reducescatter buffers with symm memory

nvcastet:
## Motivation

Speedup AllGather and ReduceScatter for max-throughput configs (DP-attention)

## Benchmark & Profiling
E2E Speedup: 4.7%
### Baseline

Server:
```
python3 -m sglang.launch_server --model-path nvidia/DeepSeek-R1-FP4 --trust-remote-code --quantization modelopt_fp4 --tp 8 --enable-flashinfer-cutlass-moe --enable-ep-moe --ep-size 8 --dp 8 --enable-dp-attention --chunked-prefill-size 16384 --mem-fraction-static 0.85 --max-running-requests 4096 --stream-interval 5 --enable-dp-lm-head --attention-backend trtllm_mla --cuda-graph-bs 1 2 4 8 16 32 64 128 256 512 1024 2048 4096 8192 --disable-radix-cache
```

Client:
```
python3 -m sglang.bench_serving --backend sglang --dataset-name random --num-prompt 1024 --random-input 1024 --random-output 2048 --random-range-ratio 1 --warmup-request 1024
```

```
Max request concurrency:                 not set
Benchmark duration (s):                  115.62
Total generated tokens (retokenized):    2090152
Request throughput (req/s):              8.86
Input token throughput (tok/s):          9068.80
Output token throughput (tok/s):         18137.61
Total token throughput (tok/s):          27206.41
Concurrency:                             1022.52
Mean E2E Latency (ms):                   115457.41
Median E2E Latency (ms):                 115462.20
Mean TTFT (ms):                          10490.22
Median TTFT (ms):                        10378.01
P99 TTFT (ms):                           18484.57
Mean ITL (ms):                           51.28
Median ITL (ms):                         47.05
P95 ITL (ms):                            52.15
P99 ITL (ms):                            54.71
Max ITL (ms):                            3283.82
```

### With this PR

```
python3 -m sglang.launch_server --model-path nvidia/DeepSeek-R1-FP4 --trust-remote-code --quantization modelopt_fp4 --tp 8 --enable-flashinfer-cutlass-moe --enable-ep-moe --ep-size 8 --dp 8 --enable-dp-attention --chunked-prefill-size 16384 --mem-fraction-static 0.85 --max-running-requests 4096 --stream-interval 5 --enable-dp-lm-head --attention-backend trtllm_mla --cuda-graph-bs 1 2 4 8 16 32 64 128 256 512 1024 2048 4096 8192 --disable-radix-cache  --enable-symm-mem
```

```
Max request concurrency:                 not set
Benchmark duration (s):                  110.42
Total generated tokens (retokenized):    2090080
Request throughput (req/s):              9.27
Input token throughput (tok/s):          9496.07
Output token throughput (tok/s):         18992.14
Total token throughput (tok/s):          28488.20
Concurrency:                             1022.55
Mean E2E Latency (ms):                   110265.53
Median E2E Latency (ms):                 110274.49
Mean TTFT (ms):                          10447.13
Median TTFT (ms):                        10271.72
P99 TTFT (ms):                           18478.40
Mean ITL (ms):                           48.76
Median ITL (ms):                         44.57
P95 ITL (ms):                            49.59
P99 ITL (ms):                            52.35
Max ITL (ms):                            3298.69
```

## Checklist

### DIFF
--- a/python/sglang/srt/distributed/device_communicators/pynccl_allocator.py
+++ b/python/sglang/srt/distributed/device_communicators/pynccl_allocator.py
@@ -5,7 +5,6 @@ from packaging import version
 from torch.cuda.memory import CUDAPluggableAllocator
 
 from sglang.srt.distributed.parallel_state import GroupCoordinator
-from sglang.srt.managers.schedule_batch import global_server_args_dict
 
 nccl_allocator_source = """
 #include <nccl.h>
@@ -29,10 +28,24 @@ _allocator = None
 _mem_pool = None
 _registered_base_addrs = set()
 _graph_pool_id = None
+_cached_pool_snapshot = None
 
 
 def is_symmetric_memory_enabled():
+    # Import here to avoid circular import
+    from sglang.srt.managers.schedule_batch import global_server_args_dict
+
     return global_server_args_dict["enable_symm_mem"]
+
+
+def is_symmetric_memory_tensor(tensor: torch.Tensor):
+    if not is_symmetric_memory_enabled() or _cached_pool_snapshot is None:
+        return False
+    for segment in _cached_pool_snapshot:
+        for block in segment["blocks"]:
+            if block["address"] == tensor.untyped_storage().data_ptr():
+                return True
+    return False
 
 
 def set_graph_pool_id(graph_pool_id):
@@ -64,8 +77,17 @@ def get_nccl_mem_pool():
 
 
 class use_symmetric_memory:
-    def __init__(self, group_coordinator: GroupCoordinator):
-        if not is_symmetric_memory_enabled():
+    def __init__(
+        self,
+        group_coordinator: GroupCoordinator,
+        disabled: bool = False,
+    ):
+        self.disabled = (
+            disabled
+            or not is_symmetric_memory_enabled()
+            or group_coordinator.world_size == 1
+        )
+        if self.disabled:
             self.group_coordinator = None
             self._mem_pool_ctx = None
             self.is_graph_capture = None
@@ -79,7 +101,7 @@ class use_symmetric_memory:
             self.pre_2_8_0 = version.parse(torch.__version__) < version.parse("2.8.0")
 
     def __enter__(self):
-        if not is_symmetric_memory_enabled():
+        if self.disabled:
             return self
         assert (
             self.group_coordinator.pynccl_comm is not None
@@ -101,17 +123,14 @@ class use_symmetric_memory:
         self._mem_pool_ctx.__enter__()
         return self
 
-    def tag(self, tensor: torch.Tensor):
-        if not is_symmetric_memory_enabled():
-            return
-        tensor.symmetric_memory = True
-
     def __exit__(self, exc_type, exc_val, exc_tb):
-        if not is_symmetric_memory_enabled():
+        if self.disabled:
             return
+        global _cached_pool_snapshot
         global _registered_base_addrs
         self._mem_pool_ctx.__exit__(exc_type, exc_val, exc_tb)
-        for segment in get_nccl_mem_pool().snapshot():
+        _cached_pool_snapshot = get_nccl_mem_pool().snapshot()
+        for segment in _cached_pool_snapshot:
             if segment["address"] not in _registered_base_addrs:
                 if segment["stream"] == 0 and self.pre_2_8_0:
                     # PyTorch version < 2.8.0 has a multi-thread MemPool bug
--- a/python/sglang/srt/distributed/parallel_state.py
+++ b/python/sglang/srt/distributed/parallel_state.py
@@ -284,7 +284,13 @@ class GroupCoordinator:
         from sglang.srt.distributed.device_communicators.pynccl import (
             PyNcclCommunicator,
         )
+        from sglang.srt.distributed.device_communicators.pynccl_allocator import (
+            is_symmetric_memory_tensor,
+            use_symmetric_memory,
+        )
 
+        self.is_symmetric_memory_tensor = is_symmetric_memory_tensor
+        self.use_symmetric_memory = use_symmetric_memory
         if is_hip():
             from sglang.srt.distributed.device_communicators.quick_all_reduce import (
                 QuickAllReduce,
@@ -509,11 +515,7 @@ class GroupCoordinator:
         if self.npu_communicator is not None and not self.npu_communicator.disabled:
             return self.npu_communicator.all_reduce(input_)
 
-        if (
-            self.pynccl_comm is not None
-            and hasattr(input_, "symmetric_memory")
-            and input_.symmetric_memory
-        ):
+        if self.pynccl_comm is not None and self.is_symmetric_memory_tensor(input_):
             with self.pynccl_comm.change_state(
                 enable=True, stream=torch.cuda.current_stream()
             ):
@@ -579,9 +581,23 @@ class GroupCoordinator:
         self,
         output: torch.Tensor,
         input: torch.Tensor,
-    ) -> None:
-        # TODO(ch-wan): support other backends
-        torch.distributed.reduce_scatter_tensor(output, input, group=self.device_group)
+    ) -> torch.Tensor:
+        pynccl_comm = self.pynccl_comm
+        if pynccl_comm is not None and (
+            not pynccl_comm.disabled
+            or (
+                self.is_symmetric_memory_tensor(output)
+                and self.is_symmetric_memory_tensor(input)
+            )
+        ):
+            with pynccl_comm.change_state(
+                enable=True, stream=torch.cuda.current_stream()
+            ):
+                pynccl_comm.reduce_scatter(output, input)
+        else:
+            torch.distributed.reduce_scatter_tensor(
+                output, input, group=self.device_group
+            )
         return output
 
     def reduce_scatter(
@@ -628,8 +644,17 @@ class GroupCoordinator:
 
     def _all_gather_into_tensor(self, output: torch.Tensor, input: torch.Tensor):
         pynccl_comm = self.pynccl_comm
-        if pynccl_comm is not None and not pynccl_comm.disabled:
-            pynccl_comm.all_gather(output, input)
+        if pynccl_comm is not None and (
+            not pynccl_comm.disabled
+            or (
+                self.is_symmetric_memory_tensor(output)
+                and self.is_symmetric_memory_tensor(input)
+            )
+        ):
+            with pynccl_comm.change_state(
+                enable=True, stream=torch.cuda.current_stream()
+            ):
+                pynccl_comm.all_gather(output, input)
         else:
             torch.distributed.all_gather_into_tensor(
                 output, input, group=self.device_group
@@ -691,9 +716,10 @@ class GroupCoordinator:
         # torch.compile . see https://github.com/pytorch/pytorch/issues/138795
         output_size = (input_size[0] * world_size,) + input_size[1:]
         # Allocate output tensor.
-        output_tensor = torch.empty(
-            output_size, dtype=input_.dtype, device=input_.device
-        )
+        with self.use_symmetric_memory(self):
+            output_tensor = torch.empty(
+                output_size, dtype=input_.dtype, device=input_.device
+            )
 
         # All-gather.
         if input_.is_cpu and is_shm_available(
--- a/python/sglang/srt/layers/dp_attention.py
+++ b/python/sglang/srt/layers/dp_attention.py
@@ -17,6 +17,9 @@ from sglang.srt.distributed import (
     get_tp_group,
     tensor_model_parallel_all_reduce,
 )
+from sglang.srt.distributed.device_communicators.pynccl_allocator import (
+    use_symmetric_memory,
+)
 
 if TYPE_CHECKING:
     from sglang.srt.configs.model_config import ModelConfig
@@ -72,6 +75,7 @@ class _DpGatheredBufferWrapper:
     _device: torch.device
     _global_dp_buffer_len: int
     _local_dp_buffer_len: int
+    _dp_max_padding: bool = True
     _global_num_tokens: Optional[List[int]]
 
     @classmethod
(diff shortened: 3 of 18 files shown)


python/sglang/srt/distributed/parallel_state.py
```diff
@@ -567,9 +569,20 @@ def reduce_scatter_tensor(
         self,
         output: torch.Tensor,
         input: torch.Tensor,
-    ) -> None:
-        # TODO(ch-wan): support other backends
-        torch.distributed.reduce_scatter_tensor(output, input, group=self.device_group)
+    ) -> torch.Tensor:
+        pynccl_comm = self.pynccl_comm
+        out_symmetric_memory = self.is_symmetric_memory_tensor(output)
```
trevor-m: Does the input need to be in symmetric memory too?

python/sglang/srt/distributed/parallel_state.py
```diff
@@ -268,7 +268,13 @@ def __init__(
         from sglang.srt.distributed.device_communicators.pynccl import (
             PyNcclCommunicator,
         )
+        from sglang.srt.distributed.device_communicators.pynccl_allocator import (
+            is_symmetric_memory_tensor,
+            use_symmetric_memory,
+        )
 
+        self.is_symmetric_memory_tensor = is_symmetric_memory_tensor
```
trevor-m, line 292: I see the comment `# lazy import to avoid documentation build error` but I wonder if this is still necessary? pynccl_allocator looks safe to import. Not a big deal, but `self.use_symmetric_memory(self)` just looks a little weird.

python/sglang/srt/layers/dp_attention.py
```diff
@@ -364,7 +367,12 @@ def dp_reduce_scatter_tensor(output: torch.Tensor, input: torch.Tensor):
             get_tensor_model_parallel_world_size()
         )[get_tensor_model_parallel_rank()]
         get_tp_group().reduce_scatter_tensor(scattered_local_tokens, input)
-        get_attention_tp_group().all_gather_into_tensor(output, scattered_local_tokens)
+        if get_attention_tp_size() > 1:
```
trevor-m: I think line 363 `if get_tensor_model_parallel_world_size() == get_attention_dp_size():` is equivalent to `get_attention_tp_size() == 1` since `attn_tp_size = tp_size // dp_size`. So we shouldn't need this change. 

Maybe change line 363 to `get_attention_tp_size() == 1` so its more clear?

python/sglang/srt/model_executor/cuda_graph_runner.py
```diff
@@ -342,26 +343,30 @@ def __init__(self, model_runner: ModelRunner):
...
                         (self.dp_size,), dtype=torch.int32
                     )
-                    self.gathered_buffer = torch.zeros(
-                        (
-                            self.max_num_token * self.dp_size,
-                            self.model_runner.model_config.hidden_size,
-                        ),
-                        dtype=self.model_runner.dtype,
-                    )
+                    with use_symmetric_memory(get_tp_group(), disable_war=True) as sm:
```
trevor-m: Is disable_war used here because this is outside of the cuda graph?

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -459,25 +463,36 @@ def forward(
...
-                    hidden_states, can_fuse_mlp_allreduce, use_reduce_scatter
+                    hidden_states,
+                    can_fuse_mlp_allreduce,
+                    use_reduce_scatter,
+                    disable_symmetric_memory,
                 )
         else:
-            return self.forward_deepep(hidden_states, forward_batch)
+            return self.forward_deepep(
+                hidden_states, forward_batch, disable_symmetric_memory
```
trevor-m: Not sure if we need to touch deepep path

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -1188,10 +1222,14 @@ def forward_normal_prepare(
...
 
-    def forward_normal_core(self, q, k, v, forward_batch):
+    def forward_normal_core(
+        self, q, k, v, forward_batch, disable_symmetric_memory: bool = False
+    ):
         attn_output = self.attn_mha(q, k, v, forward_batch, save_kv_cache=False)
         attn_output = attn_output.reshape(-1, self.num_local_heads * self.v_head_dim)
-        output, _ = self.o_proj(attn_output)
+        output, _ = self.o_proj(
+            attn_output, disable_symmetric_memory=disable_symmetric_memory
```
trevor-m: It would be nice if we could get this from a helper function to avoid needing to pass it everywhere. We have forward_batch here but we would also need the layer_communicator in order to check should_use_reduce_scatter...

trevor-m (changes requested):
Nice work @nvcastet!

python/sglang/srt/distributed/parallel_state.py
nvcastet: Yes Good point I should probably check for that too.

python/sglang/srt/distributed/parallel_state.py
nvcastet: Yes I agree, I could try to put pynccl import at the top and see what I get.

python/sglang/srt/model_executor/cuda_graph_runner.py
nvcastet: Those allocations associated with stream zero which conflicts with the WAR for pre 2.8 pytorch.
For 2.8 and beyond, it is a noop.

python/sglang/srt/models/deepseek_v2.py
nvcastet: yes probably not since i did not test it.

python/sglang/srt/models/deepseek_v2.py
nvcastet: I agree maybe adding an attribute to forward batch? but in another PR review, it was mentioned avoiding passing big object like forward_batch to know explicitly what is accessed by the function.

python/sglang/srt/distributed/parallel_state.py
nvcastet: Does not work:
```
ImportError: cannot import name 'GroupCoordinator' from partially initialized module 'sglang.srt.distributed.parallel_state' (most likely due to a circular import) (/sgl-workspace/sglang/python/sglang/srt/distributed/parallel_state.py)
```

trevor-m: Got it, thanks for trying!

python/sglang/srt/layers/dp_attention.py
```diff
@@ -357,7 +360,7 @@ def dp_scatter(
 
 
 def dp_reduce_scatter_tensor(output: torch.Tensor, input: torch.Tensor):
-    if get_tensor_model_parallel_world_size() == get_attention_dp_size():
+    if get_attention_dp_size() == 1:
```
trevor-m: This should be attention TP size

python/sglang/srt/layers/vocab_parallel_embedding.py
```diff
@@ -473,7 +473,9 @@ def forward(self, input_):
         else:
             masked_input = input_
         # Get the embeddings.
-        with use_symmetric_memory(parallel_state.get_tp_group()) as sm:
+        with use_symmetric_memory(
+            parallel_state.get_tp_group(), disabled=(self.tp_size == 1)
```
trevor-m: In use_symmetric_memory should we automatically disable if the group's world size is 1?

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -495,7 +510,9 @@ def forward_normal_dual_stream(
             if not _is_cuda:
                 final_hidden_states *= self.routed_scaling_factor
         current_stream.wait_stream(self.alt_stream)
-        with use_symmetric_memory(parallel_state.get_tp_group()) as sm:
+        with use_symmetric_memory(
+            parallel_state.get_tp_group(), disabled=disable_symmetric_memory
+        ) as sm:
             final_hidden_states_out = torch.empty_like(final_hidden_states)
         torch.add(final_hidden_states, shared_output, out=final_hidden_states_out)
         final_hidden_states = final_hidden_states_out
```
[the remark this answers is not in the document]
trevor-m: Do we need to tag it?

trevor-m (changes requested):

python/sglang/srt/layers/dp_attention.py
nvcastet: Yes good catch.

python/sglang/srt/models/deepseek_v2.py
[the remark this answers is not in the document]
nvcastet: Yes it is tag a few lines below

python/sglang/srt/layers/vocab_parallel_embedding.py
nvcastet: yes that would make sense.

trevor-m (approved):
Thanks, looks good!

python/sglang/srt/models/deepseek_v2.py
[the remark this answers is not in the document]
merrymercy: Why do you tag the tensor outside of the `with` scope?

merrymercy (changes requested):
This PR changed the model forward code too much. Some changes are not intuitive (e.g., which tensor to tag, which region to use `with use_symmetric_memory`.

Is it possible to make it more transparent so that we get this feature without change any model forward code? (even `logits_processor.py`)

python/sglang/srt/models/deepseek_v2.py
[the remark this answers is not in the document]
nvcastet: The `with` scope make all pytorch allocations under it come from the symmetric memory pool.
The `.tag(tensor)` is just book-keeping to flag a tensor that has been allocated via symmetric memory so that when we call a collective on it we select NCCL to get best perf instead of alternative custom kernels.

But I can move it there if it is clearer for people and AIs. :)

nvcastet:
> This PR changed the model forward code too much. 
> 
> Is it possible to make it more transparent so that we get this feature without change any model forward code? (even `logits_processor.py`)

Thanks for taking the time to review @merrymercy !
The takeaway of this PR and previous PR (#8238) is to leverage NCCL new symmetric kernels requiring memory used for communication to be allocated via a separate allocator&registration.
We could be less intrusive by making a buffer copy before the comm op instead of finding the source allocation that is used for communication but we would leave perf on the table especially for larger buffer sizes.

An other option is to ignore code sections where perf gain is small. I believe i could ignore comms inside `logits_processor.py` for example.

@merrymercy Could you provide some guidance on the direction to take?

As a reference, here is the MNNVL NCCL allreduce standalone perf gain using this feature:
image
Source: https://developer.nvidia.com/blog/enabling-fast-inference-and-resilient-training-with-nccl-2-27/#low-latency_kernels_with_symmetric_memory

python/sglang/srt/layers/logits_processor.py
```diff
@@ -186,18 +190,20 @@ def compute_dp_attention_metadata(self):
...
-                (
-                    sum(self.global_num_tokens_for_logprob_cpu),
-                    self.gathered_buffer.shape[1],
-                ),
-                dtype=self.gathered_buffer.dtype,
-                device=self.gathered_buffer.device,
-            )
-        else:
-            self.gathered_buffer = torch.empty_like(self.gathered_buffer)
+        with use_symmetric_memory(get_tp_group()) as sm:
```
ch-wan: When is symmetric memory efficient? I'm a little bit confused here. Why we do not check padding mode here?

python/sglang/srt/models/deepseek_v2.py
ch-wan: You may refer to our recent refactor in `dp_attention.py`. We prepared some util functions to avoid tedious coding efforts. A similar util function can be prepared for symmetric_memory.

nvcastet: Thanks let me have a look.

python/sglang/srt/layers/logits_processor.py
nvcastet: It should be always more efficient. The problem is that we don't want to allocate buffers that are not symmetric across GPU (otherwise we get undefined behavior) which can happen when we use the symmetric context manager across a big code section with DP-attention.
Here specifically the only allocation under the context manager is `self.gathered_buffer` which always has the same size across GPUs.

python/sglang/srt/layers/linear.py
```diff
@@ -1301,7 +1301,7 @@ def weight_loader_v2(self, param: BasevLLMParameter, loaded_weight: torch.Tensor
             # It does not support additional parameters.
             param.load_row_parallel_weight(loaded_weight)
 
-    def forward(self, input_, skip_all_reduce=False):
+    def forward(self, input_, skip_all_reduce=False, disable_symmetric_memory=True):
```
trevor-m: Can we just check `not is_max_padding()` here instead of adding an arg?

python/sglang/srt/layers/dp_attention.py
```diff
@@ -119,14 +133,19 @@ def get_local_dp_buffer_len(cls) -> int:
     def get_dp_global_num_tokens(cls) -> List[int]:
         return cls._global_num_tokens
 
+    @classmethod
+    def is_max_padding(cls) -> bool:
```
trevor-m: Maybe we should call this `is_dp_max_padding` to match the other functions

trevor-m (commented):
Thanks, overall looks good just some minor questions.

python/sglang/srt/layers/linear.py
nvcastet: Good catch. Let me check.

python/sglang/srt/layers/dp_attention.py
nvcastet: Yes

trevor-m (approved):
Looks good, thanks!

python/sglang/srt/layers/quantization/modelopt_quant.py
```diff
@@ -1263,7 +1270,12 @@ def apply(
             topk_weights, topk_ids = topk_output.topk_weights, topk_output.topk_ids
 
             output_dtype = x.dtype
+            output_col = x.shape[1]
             x_sf = None
+
+            with use_symmetric_memory(get_tp_group(), disabled=not is_dp_max_padding()):
+                symm_output = torch.empty_like(x)
```
trevor-m: No longer needed

python/sglang/srt/distributed/device_communicators/pynccl_allocator.py
```diff
@@ -32,9 +31,22 @@
...
+    # Import here to avoid circular import
+    from sglang.srt.managers.schedule_batch import global_server_args_dict
+
     return global_server_args_dict["enable_symm_mem"]
 
 
+def is_symmetric_memory_tensor(tensor: torch.Tensor):
+    if not is_symmetric_memory_enabled():
+        return False
+    for segment in get_nccl_mem_pool().snapshot():
```
trevor-m: It's great that we no longer need to tag the tensors, but I'm concerned about the efficiency of this check. [Collecting the snapshot](https://github.com/pytorch/pytorch/blob/c5cb255625deb4cdbc5780e6911b73498e17ed5a/torch/csrc/cuda/Module.cpp#L724) seems to do a lot and then we also have to iterate over all segments/blocks.

1. Do you know if basic caching/memoization in this method would work using `tensor.untyped_storage.data_ptr()` as the key? Will the storage change from iteration to iteration or remain consistent?
2. If that doesn't work, could you measure how long this check takes?

trevor-m (changes requested):

python/sglang/srt/distributed/device_communicators/pynccl_allocator.py
nvcastet: It would change, depending on what gets allocated in the pool But we could at least cache the snapshot at the exit of the context manager.

2. Yes I could time the call.

nvcastet: 133 us (not cached) vs 5us (cached)
So I made the changes to cache it at context manager exit.

trevor-m (approved):

kushanam (approved):

nvcastet:
@merrymercy Do you mind having another look?

nvcastet:
Closing in favor of https://github.com/sgl-project/sglang/pull/9358

PR #6555, opened 2025-05-23
Support Multi Process Tokenizer Manager

whybeyoung:
## Target

Support  https://github.com/sgl-project/sglang/issues/6553

Improve the processing speed of the tokenizer, increase throughput, and reduce latency by using a multi-process approach with  multi Tokenizer manager.

## Proposal Arch

image

## Initialization Startup Process:
1.  Add the worker_num parameter as an input argument during startup.
2. The main process launch_server initializes and starts including the scheduler subprocess and detokenizer manager subprocess. It stores information required for subprocess initialization (e.g., server_args, port_args) via shared memory. Set the workers=worker_num for fastapi unicorn' startup.
3. Worker(with it's own tokenizer manager) load initialization information from shared memory, initialize their tokenizer manager , and write their tokenizer_rpc_name and worker ID into shared memory for the detokenizer  manager to access.
4. The detokenizer manager runs  an event loop, continuously polling until all workers are ready (all tokenizer_rpc_name entries are available in shared memory). It then establishes a mapping between worker ids and their corresponding ZMQ communication endpoints. Then it start cousume data from scheduler.

## Request Process:
1. Upon receiving a request at the FastAPI frontend, it is routed to a Unicorn worker subprocess, each of which maintains an independent Tokenizer Manager. The subprocess appends its worker ID to the request's rid and forwards it to the scheduler. The scheduler proceeds with the inference pipeline as usual and returns the result to the Detokenizer Manager.
2. Upon receiving the data, the Detokenizer Manager extracts the worker_id from the rid, then uses the registered mapping information to route and send the response back to the corresponding frontend worker (Tokenizer Manager).
3. The frontend subprocess worker (Tokenizer Manager) then proceeds with the normal response handling logic.

CC  @lw9527 @ByronHsu @ShangmingCai @zhyncs @hnyls2002 @fzyzcjy

Issue #6553, opened 2025-05-23
[PD] Support Multi-Process for TokenizerManager


whybeyoung:
### Checklist

- [x] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 2. Please use English, otherwise it will be closed.

### Motivation

The TokenizerManager is responsible for three main tasks:

Receiving requests

Tokenizing

Streaming results

Through testing, we’ve identified that TokenizerManager may encounter single-core bottlenecks under high-concurrency scenarios. Therefore, I propose decoupling the tokenization step from the manager process as a first step.

PR #8964, opened 2025-08-08
Support Multi Process Tokenizer Manager(#6555)

whybeyoung:
Due to PR #6555 being reverted and the need for a more in-depth review, a new PR has been submitted and is pending Lianmin's review. The historical review records can be traced back to PR #6555.

## Target

Support  https://github.com/sgl-project/sglang/issues/6553

Improve the processing speed of the tokenizer, increase throughput, and reduce latency by using a multi-process approach with  multi Tokenizer manager.

## Proposal Arch

image

## Initialization Startup Process:
1.  Add the worker_num parameter as an input argument during startup.
2. The main process launch_server initializes and starts including the scheduler subprocess and detokenizer manager subprocess. It stores information required for subprocess initialization (e.g., server_args, port_args) via shared memory. Set the workers=worker_num for fastapi unicorn' startup.
3. Worker(with it's own tokenizer manager) load initialization information from shared memory, initialize their tokenizer manager , and write their tokenizer_rpc_name and worker ID into shared memory for the detokenizer  manager to access.
4. The detokenizer manager runs  an event loop, continuously polling until all workers are ready (all tokenizer_rpc_name entries are available in shared memory). It then establishes a mapping between worker ids and their corresponding ZMQ communication endpoints. Then it start cousume data from scheduler.

## Request Process:
1. Upon receiving a request at the FastAPI frontend, it is routed to a Unicorn worker subprocess, each of which maintains an independent Tokenizer Manager. The subprocess appends its worker ID to the request's rid and forwards it to the scheduler. The scheduler proceeds with the inference pipeline as usual and returns the result to the Detokenizer Manager.
2. Upon receiving the data, the Detokenizer Manager extracts the worker_id from the rid, then uses the registered mapping information to route and send the response back to the corresponding frontend worker (Tokenizer Manager).
3. The frontend subprocess worker (Tokenizer Manager) then proceeds with the normal response handling logic.

CC  @lw9527 @ByronHsu @ShangmingCai @zhyncs @hnyls2002 @fzyzcjy @LLLL114 @merrymercy


Kevin-XiongC:
Hi, Thanks for your great contribution. Are there any benchmarks showing performance gain? We've also found that the single tokenizer could be a bottleneck for the system.

ShangmingCai:
> Hi, Thanks for your great contribution. Are there any benchmarks showing performance gain? We've also found that the single tokenizer could be a bottleneck for the system.

@Kevin-XiongC There are many results in the original PR #6555, you can click in and check the history.

kyleliang-nv:
I have been testing PD-disagg on GB200 NVL72, and noticed a perf bottleneck on the de-tokenzier on the decode server (single decode server with EP48). I tried this PR which enabled multi-process de-tokenzier, and was able to see the bottleneck resolved with 4 workers.

ShangmingCai:
Is it possible to accelerate the review process of this PR?

zhyncs:
@ShangmingCai lianmin and I are busy these days, once @hnyls2002 @ShangmingCai @whybeyoung approve, we can merge this first

python/sglang/srt/entrypoints/http_server.py
```diff
@@ -208,6 +277,30 @@ async def lifespan(fast_api_app: FastAPI):
...
+    # Check if metrics are enabled via environment variable
+    enable_metrics = get_bool_env_var("SGLANG_ENABLE_METRICS", "false")
+    if enable_metrics:
+        add_prometheus_middleware(app)
+        enable_func_timer()
+        logger.info(f"Worker {worker_pid} added prometheus middleware")
+
+
+# Call setup function at module level for multi-process compatibility
+setup_middlewares()
```
ShangmingCai: nit: should we put and await this line in the
```python
async def lifespan(fast_api_app: FastAPI):
```

LLLL114: resolved

python/sglang/srt/entrypoints/http_server.py
```diff
@@ -191,7 +268,15 @@ async def lifespan(fast_api_app: FastAPI):
...
         warmup_thread.start()
-    yield
+
+    try:
+        yield
+    finally:
+        if server_args.tokenizer_worker_num > 1:
+            pid = os.getpid()
+            logger.info(f"uvicorn worker {pid} ending...")
+            warmup_thread.join()
```
ShangmingCai, line 280: Is there any reason why multi tokenizer warmup join here instead of:
https://github.com/sgl-project/sglang/blob/961f673a14cda38ea70dd873a2d5a499435cf993/python/sglang/srt/entrypoints/http_server.py#L1238-L1244

python/sglang/srt/managers/detokenizer_manager.py
```diff
@@ -32,11 +32,14 @@
...
     BatchTokenIDOut,
     FreezeGCReq,
+    MultiTokenizerRegisterReq,
 )
+from sglang.srt.managers.multi_tokenizer_mixin import MultiTokenizerMixin
 from sglang.srt.server_args import PortArgs, ServerArgs
 from sglang.srt.utils import (
     configure_logger,
     freeze_gc,
+    get_workerids_from_rids,
```
ShangmingCai: nit: How about changing `get_workerids_from_rids` to `get_worker_ids_from_req_rids`?

python/sglang/srt/managers/detokenizer_manager.py
```diff
@@ -116,6 +120,39 @@ def event_loop(self):
             if output is not None:
                 self.send_to_tokenizer.send_pyobj(output)
 
+    def mtm_event_loop(self):
```
ShangmingCai: nit: I think it is ok to use `multi_tokenizer_manager_event_loop`, even though it is a bit long, I think better readability is way important.

python/sglang/srt/managers/detokenizer_manager.py
```diff
@@ -285,7 +322,10 @@ def run_detokenizer_process(
 
     try:
         manager = DetokenizerManager(server_args, port_args)
-        manager.event_loop()
+        if server_args.tokenizer_worker_num == 1:
+            manager.event_loop()
+        else:
+            manager.mtm_event_loop()
```
ShangmingCai: ditto

python/sglang/srt/entrypoints/engine.py
```diff
@@ -814,18 +815,24 @@ def _launch_subprocesses(
         ),
     )
     detoken_proc.start()
+    if server_args.tokenizer_worker_num == 1:
```
ShangmingCai: I see sometimes we use ```if server_args.tokenizer_worker_num == 1```, but most of times, ```if server_args.tokenizer_worker_num > 1``` is used, can we sync all branch logics to ```if server_args.tokenizer_worker_num > 1``` to improve maintainability?

python/sglang/srt/managers/detokenizer_manager.py
```diff
@@ -285,7 +322,10 @@ def run_detokenizer_process(
 
     try:
         manager = DetokenizerManager(server_args, port_args)
-        manager.event_loop()
+        if server_args.tokenizer_worker_num == 1:
```
ShangmingCai: ditto

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -264,9 +266,15 @@ def __init__(
         self.recv_from_detokenizer = get_zmq_socket(
             context, zmq.PULL, port_args.tokenizer_ipc_name, True
         )
-        self.send_to_scheduler = get_zmq_socket(
-            context, zmq.PUSH, port_args.scheduler_input_ipc_name, True
-        )
+        if self.server_args.tokenizer_worker_num == 1:
```

python/sglang/srt/entrypoints/http_server.py
LLLL114: for multi tokenizer mode, `warmup_thread` will be init in lifespan `if server_args is None:`, so warmup_thread join in lifespan function

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -1596,9 +1619,12 @@ def _handle_batch_output(
                 )
                 continue
 
+            originRid = rid
+            if self.server_args.tokenizer_worker_num > 1:
+                originRid = get_origin_rid(rid)
             # Build meta_info and return value
             meta_info = {
-                "id": rid,
+                "id": originRid,
```
ShangmingCai: LowerCamelCase looks really weird in Python, is there any reason why we don't use `origin_rid`?

python/sglang/srt/managers/detokenizer_manager.py
LLLL114: resolved

python/sglang/srt/managers/detokenizer_manager.py
LLLL114: resolved

python/sglang/srt/managers/detokenizer_manager.py
LLLL114: resolved

python/sglang/srt/entrypoints/engine.py
LLLL114: resolved

python/sglang/srt/managers/detokenizer_manager.py
LLLL114: resolved

python/sglang/srt/managers/tokenizer_manager.py
LLLL114: resolved

python/sglang/srt/managers/tokenizer_manager.py
LLLL114: resolved

python/sglang/srt/managers/multi_tokenizer_mixin.py
```diff
@@ -0,0 +1,582 @@
...
+            if isinstance(recv_obj, MultiTokenizerRegisterReq):
+                return self.init_tokenizer_mapping(recv_obj, worker_id)
+            else:
+                logger.error(
+                    f"Worker {worker_id} not registered and not found in tokenizer mapping . "
+                    "Please ensure the worker is registered correctly."
+                )
+        return False
+
+    def _handle_output_by_index(self, output, i):
```
ShangmingCai, line 87: I think there are so many similar logics in this util func, can we merge similar logics based on field name? Just like something we did in the TBO: https://github.com/sgl-project/sglang/blob/e5b29bf14e4008f6a9bb08dc6e7fcac1a22351fe/python/sglang/srt/two_batch_overlap.py#L729-L733

python/sglang/srt/managers/multi_tokenizer_mixin.py
```diff
@@ -0,0 +1,582 @@
...
+
+    def init_tokenizer_mapping(
+        self, recv_obj: MultiTokenizerRegisterReq, worker_id: str
+    ):
+        """init tokenizer mapping from register request"""
+        ipc_name = recv_obj.ipc_name
+        worker_id_int = int(worker_id)
+
+        if worker_id_int not in self.tokenizer_mapping:
+            socket = get_zmq_socket(self._zmq_context, zmq.PUSH, ipc_name, False)
```
ShangmingCai, line 68: might need a close func to clean up the socket resource.

ShangmingCai (commented):
I think that's all my comments. Will retrigger the CI when they are addressed. @hnyls2002 please find some time to review this PR as well, I think it has been delayed for too long, LOL.

python/sglang/srt/managers/multi_tokenizer_mixin.py
LLLL114: resolved

python/sglang/srt/managers/multi_tokenizer_mixin.py
LLLL114: I tried `field`. It increases TTFT about 25% becasue of `getattr` and the for loop. I push this code. if TTFT is more important, I can reset it.

ShangmingCai: Yeah. Performance is way more important, you can reset this first, maybe we can add a comment here and revisit it later.

justadogistaken:
Hi, I wonder if the detokenize manager is multi-process?

justadogistaken:
> I have been testing PD-disagg on GB200 NVL72, and noticed a perf bottleneck on the de-tokenzier on the decode server (single decode server with EP48). I tried this PR which enabled multi-process de-tokenzier, and was able to see the bottleneck resolved with 4 workers.

Hi, I met the same problem. But I don't see "multi-process de-tokenizer" in this PR, only "multi-process tokenizer" are presented. I might miss something. I would be appreciate it if you can point "multi-process de-tokenizer" out. Thank you.

LLLL114:
> > I have been testing PD-disagg on GB200 NVL72, and noticed a perf bottleneck on the de-tokenzier on the decode server (single decode server with EP48). I tried this PR which enabled multi-process de-tokenzier, and was able to see the bottleneck resolved with 4 workers.
> 
> Hi, I met the same problem. But I don't see "multi-process de-tokenizer" in this PR, only "multi-process tokenizer" are presented. I might miss something. I would be appreciate it if you can point "multi-process de-tokenizer" out. Thank you.

there are no multi detokenizer manager in this PR, as the detokenizer manager is not the bottle neck. We find some process in tokenizer manager is too slow which block the queue, so use multi tokenizer manager to solve it.

python/sglang/srt/managers/multi_tokenizer_mixin.py
LLLL114: reset.

python/sglang/srt/entrypoints/engine.py
```diff
@@ -814,18 +815,24 @@ def _launch_subprocesses(
...
     )
     detoken_proc.start()
+    if server_args.tokenizer_worker_num > 1:
+        # Launch tokenizer process
+        tokenizer_manager = MultiTokenizerRouter(server_args, port_args)
 
-    # Launch tokenizer process
-    tokenizer_manager = TokenizerManager(server_args, port_args)
+        # Initialize templates
+        template_manager = None
```
ShangmingCai, line 823: Should these lines be removed? I think multi-tokenizer's `template_manager` doesn't really initialize here?

LLLL114: `template_manager` doesn't initialize here ,but `_launch_subprocesses` function need to return a `template_manager`, so I set `None` here.

hnyls2002 (approved):

python/sglang/srt/managers/detokenizer_manager.py
```diff
@@ -116,6 +120,39 @@ def event_loop(self):
             if output is not None:
                 self.send_to_tokenizer.send_pyobj(output)
 
+    def multi_tokenizer_manager_event_loop(self):
```
merrymercy, line 123: This should go to the `MultiTokenizerMixin`

python/sglang/srt/utils.py
```diff
@@ -2787,6 +2787,20 @@ def wrapper(*args, **kwargs):
...
+        worker_ids = [int(rid.split("_")[0]) for rid in rids]
+    elif isinstance(rids, str):
+        worker_ids = [int(rids.split("_")[0])]
+    else:
+        worker_ids = []
+    return worker_ids
+
+
+def get_origin_rid(rid):
+    return rid.split("_", 1)[1] if "_" in rid else rid
```
merrymercy, line 2801: move to `multi_tokenizer_mixin.py`

python/sglang/srt/managers/detokenizer_manager.py
ShangmingCai: Fixed in https://github.com/sgl-project/sglang/pull/9902

YingshengGeng:
Hi, I'm getting an OSError: [Errno 9] Bad file descriptor when launching the sglang server on Ray with multi-process tokenization enabled. I'm wondering if this is a known issue or if anyone has found a workaround. image

YingshengGeng:
I've temporarily resolved the issue by skipping this line, as suggested in:  [https
://github.com/Kludex/uvicorn/issues/877#issue-759008860](https://github.com/Kludex/uvicorn/issues/877#issue-759008860) and https://github.com/Kludex/uvicorn/discussions/1602#discussioncomment-3412588.

PR #8968, opened 2025-08-08
Fix kimi k2 function call format

merrymercy:
Remove the space after `functions.{func_name}:{index}` in the template.
Otherwise, xgrammar will mask all tokens as invalid, then the probability becomes NaN, and the engine will sample negative token ids.

### DIFF
--- a/python/sglang/srt/function_call/kimik2_detector.py
+++ b/python/sglang/srt/function_call/kimik2_detector.py
@@ -24,7 +24,7 @@ class KimiK2Detector(BaseFormatDetector):
     Format Structure:
     ```
     <|tool_calls_section_begin|>
-    <|tool_call_begin|>functions.{func_name}:{index} <|tool_call_argument_begin|>{json_args}<|tool_call_end|>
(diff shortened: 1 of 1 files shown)


CatherineSue (approved):

python/sglang/srt/function_call/kimik2_detector.py
```diff
@@ -240,6 +240,6 @@ def build_ebnf(self, tools: List[Tool]) -> str:
             sequence_start_token=self.bot_token,
             sequence_end_token=self.eot_token,
             tool_call_separator="",
-            call_rule_fmt='"<|tool_call_begin|>functions.{name}:" [0-9]+ " <|tool_call_argument_begin|>" {arguments_rule} "<|tool_call_end|>"',
+            call_rule_fmt='"<|tool_call_begin|>functions.{name}:" [0-9]+ "<|tool_call_argument_begin|>" {arguments_rule} "<|tool_call_end|>"',
```
merrymercy:
```suggestion
            call_rule_fmt='"<|tool_call_begin|>functions.{name}:"[0-9]+"<|tool_call_argument_begin|>"{arguments_rule}"<|tool_call_end|>"',
```

PR #8969, opened 2025-08-08
bugfix: fix health_generate in pd disaggregation + dp

Bruce-x-1997:
## Motivation

when I test health_generate interface with mini_lb, I find it does not work, and always return 200, so we try to fix it

## Modifications

1. I change mini_lb /health_generate interface and make it catch the wrong status_code
2. I fix /health_generate in http_server and make it support pd disaggregation case. I set GenerateReqInput as warmup case, so the health req could work in pd+dp case.And it meets my requirement that one req could check the status of all dp workers.


Bruce-x-1997:
just a bugfix, and I see some code has been merged like the bug of time.perfcount

Bruce-x-1997:
please help trigger the ci  @zhaochenyang20

Bruce-x-1997:
please trigger ci @zhaochenyang20 thanks

python/sglang/srt/entrypoints/http_server.py
```diff
@@ -260,6 +260,14 @@ async def validate_json_request(raw_request: Request):
         )
 
 
+def _get_disagg_bootstrap_rooms(server_args: ServerArgs) -> List[int]:
```
ShangmingCai, line 263: I think this is only used for warm-up and health checks; I think maybe the current function name doesn't reveal the usage?

ShangmingCai (commented):
Let me trigger the ci for u.

Issue #8924, opened 2025-08-07
[Bug] aime25 issue


zhyncs:
### Checklist

- [ ] 1. I have searched related issues but cannot get the expected help.
- [ ] 2. The bug has not been fixed in the latest version.
- [ ] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [ ] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [ ] 5. Please use English, otherwise it will be closed.

### Describe the bug

```
# latest main
python3 -m sglang.launch_server --model openai/gpt-oss-120b --tp 8
```

```
git clone https://github.com/openai/gpt-oss
cd gpt-oss
pip3 install -e .
```

```
OPENAI_API_KEY="" python3 -m gpt_oss.evals --base-url http://127.0.0.1:30000/v1 --eval aime25 --model gpt-oss-120b --reasoning-effort high --sampler chat_completions
```

```
[2025-08-07 20:16:58] The server is fired up and ready to roll!
[2025-08-07 20:17:13] INFO:     127.0.0.1:37734 - "POST /v1/chat/completions HTTP/1.1" 400 Bad Request
[2025-08-07 20:17:13] INFO:     127.0.0.1:37740 - "POST /v1/chat/completions HTTP/1.1" 400 Bad Request
```

```
OPENAI_API_KEY="" python3 -m gpt_oss.evals --base-url http://127.0.0.1:30000/v1 --eval aime25 --model gpt-oss-120b --reasoning-effort high
```

```
Rate limit exception so wait and retry 0 after 1 sec 'Response' object has no attribute 'output_text'
Rate limit exception so wait and retry 0 after 1 sec 'Response' object has no attribute 'output_text'
Rate limit exception so wait and retry 0 after 1 sec 'Response' object has no attribute 'output_text'
```

### Reproduction

N/A

### Environment

N/A

zhyncs:
We need to use 1.99.1.
```
pip3 install openai==1.99.1
```

zhyncs:
ref https://github.com/sgl-project/sglang/pull/8927

zhyncs:
```
[...]

PR #8983, opened 2025-08-08
Clean up server_args.py to have a dedicated function for model specific adjustments

merrymercy:


.github/workflows/vllm-dependency-test.yml
```diff
@@ -32,7 +32,8 @@ jobs:
           bash scripts/ci_install_dependency.sh
           pip install "vllm==0.10.0"
           pip install "bitsandbytes>=0.44.0"
-          pip3 install torch==2.8.0 torchvision torchaudio --index-url https://download.pytorch.org/whl/test/cu126
+          pip install torch==2.8.0 torchvision torchaudio --index-url https://download.pytorch.org/whl/test/cu126
+          pip install "openai>=1.99.1"
```
zhyncs: Don't do this, 1.99.2 has issues

python/pyproject.toml
```diff
@@ -29,6 +30,7 @@ runtime_common = [
     "modelscope",
     "msgspec",
     "ninja",
+    "openai>=1.99.1",
```
zhyncs: same here

zhyncs:
ref https://github.com/sgl-project/sglang/issues/8924

PR #8985, opened 2025-08-08
Create cancel-all-pr-test-runs

merrymercy:
it might not work. merge it so we can trigger it to see whether it can work


merrymercy:
move to https://github.com/sgl-project/sglang/pull/8986

PR #8989, opened 2025-08-09
[Fix] Fix wrong backend chosen in hybrid backend

DarkSharpness:
## Motivation

When both `--prefill-attention` and `--decode-attention` are set to the same backend, the current logic may incorrectly fall back to a different backend, which may be misleading.

**Reproduction:**
```bash
    --model-path meta-llama/Llama-3.1-8B-Instruct \
    --prefill-attention fa3 \
    --decode-attention fa3 \
    --enable-hierarchical-cache
````

This code will choose `flashinfer` as the attention backend. However, when both `--prefill-attention` and `--decode-attention` are explicitly specified as `fa3`, the configuration should be overridden accordingly. This is caused by the logic [here](https://github.com/sgl-project/sglang/blob/f352b793be65a555194db31280a69b53dd40c4d4/python/sglang/srt/model_executor/model_runner.py#L1336), which fails to handle the case where `--prefill-attention` and `--decode-attention` are the same and not `None`.

While this is admittedly a rare edge case, the behavior should still be correct.

## Modifications

Fix the fallback logic when `--prefill-attention` and `--decode-attention` are identical. It will override the `attention-backend` in this case.

## Checklist

### DIFF
--- a/python/sglang/srt/model_executor/model_runner.py
+++ b/python/sglang/srt/model_executor/model_runner.py
@@ -391,7 +391,12 @@ class ModelRunner:
                 2.3 Otherwise, we will use triton backend.
             """
 
-            if not self.use_mla_backend:
+            if server_args.prefill_attention_backend is not None and (
+                server_args.prefill_attention_backend
+                == server_args.decode_attention_backend
+            ):  # override the default attention backend
+                server_args.attention_backend = server_args.prefill_attention_backend
+            elif not self.use_mla_backend:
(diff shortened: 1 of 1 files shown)


Fridge003 (approved):
Thanks for fixing this bug

Fridge003:
cc @Qiaolin-Yu @zhyncs

Qiaolin-Yu (approved):

PR #9014, opened 2025-08-09
Fuse writing KV buffer into rope kernel (part 2: srt)

JeremieMelo:
## Motivation

Fuse  set_kv_buffer to sgl-kernel rope function, only for trtllm_mha attention

-----

(below is from @fzyzcjy)

speed may be suboptimal (I have not done any ncu profile or thorough optimization), but anyway it is faster than non-fused

```
tests/test_rotary_embedding.py ..................                                                                  [100%]

=================================================== 18 passed in 1.90s ===================================================
bench_rotary_embedding:
    batch_size  seq_len  save_kv_cache  SGL Kernel
0            1        1          False       2.788
1            1        1           True       3.119
2           32        1          False       3.052
3           32        1           True       3.299
4          128        1          False       3.167
5          128        1           True       3.463
6          512        1          False       3.925
7          512        1           True       4.091
8            2      512          False       5.643
9            2      512           True       6.371
10           4     4096          False      46.858
11           4     4096           True      57.892
```

acc looks good: 20B TP4, reasoning low, gpt_oss.evals gpqa: 55.1%

acc from @BBuf (he checked this agree with main):

```
| model_name                                              |   (‘metric’, ‘mmlu’) |
|:--------------------------------------------------------|---------------------:|
| o4-mini-with-chat-completion-and-4k-gen_20250811_143850 |                0.866 |
```

speedup: @BBuf test to be 345->355


python/sglang/srt/layers/attention/trtllm_mha_backend.py
```diff
@@ -225,11 +225,11 @@ def forward_decode(
         **kwargs,
     ) -> torch.Tensor:
         """Run forward for decode using TRTLLM MHA kernel."""
-        cache_loc = forward_batch.out_cache_loc
-        if save_kv_cache and k is not None:
-            forward_batch.token_to_kv_pool.set_kv_buffer(
-                layer, cache_loc, k, v, layer.k_scale, layer.v_scale
-            )
+        # cache_loc = forward_batch.out_cache_loc
```
fzyzcjy: (don't forget to re-enable these code and do branching at correct places)

python/sglang/srt/layers/rotary_embedding.py
```diff
@@ -222,17 +222,36 @@ def forward_cuda(
...
+            # elementwise.apply_rope_with_cos_sin_cache_inplace(
             apply_rope_with_cos_sin_cache_inplace(
                 positions=positions,
                 query=query,
                 key=key,
                 head_size=self.head_size,
                 cos_sin_cache=self.cos_sin_cache,
                 is_neox=self.is_neox_style,
+                layer=layer,
+                forward_batch=forward_batch,
```
fzyzcjy: maybe we should not pass such objects to this API

what about e.g.

```
def apply_rope_with_cos_sin_cache_inplace(
  ...,
  # in non-fused version we do `k_buffer[loc] = data` etc
  k_buffer: Tensor, v_buffer: Tensor, loc: Tensor,
)
```

and if none, it means we do not save kv cache; if non-none then we need to save

sgl-kernel/csrc/elementwise/pos_enc.cuh
fzyzcjy, line 1: confused why do we have a new file instead of minor modifications

sgl-kernel/csrc/elementwise/pos_enc.cuh
```diff
@@ -0,0 +1,502 @@
...
+      DType* v_buffer_ptr =
+          v_buffer + get_elem_offset_impl(cache_offset, kv_head_idx, 0, v_buffer_stride_n, v_buffer_stride_h);
+      vec_t<float, vec_size> k_vec;
+      if constexpr (interleave) {
+        k_vec = vec_apply_llama_rope_cos_sin_interleave_reuse_half<vec_size, bdx>(k_ptr, cos, sin, rotary_dim);
+      } else {
+        k_vec = vec_apply_llama_rope_cos_sin<vec_size, bdx>(k_ptr, cos, sin, rotary_dim);
+      }
+      k_vec.cast_store(k_rope_ptr + tx * vec_size);
+      k_vec.cast_store(k_buffer_ptr + tx * vec_size);
```
fzyzcjy: wondering whether we need that `loc` conversion

sgl-kernel/csrc/elementwise/pos_enc.cuh
```diff
@@ -0,0 +1,502 @@
...
+      v.cast_store(v_buffer_ptr + tx * vec_size);
```
fzyzcjy: wondering whether this is inefficient, since workload of q and k will be imbalanced (k has more work)

sgl-kernel/csrc/elementwise/pos_enc.cuh
JeremieMelo: for the loc conversion, you mean "const IdType cache_offset = cache_loc[idx];" right?
this one is for the LHS indexing for the k_buffer, in pytorch it is k_buffer[loc] = data.
so we need to use cache_offset = cache_loc[idx] to find the first dimension's index.
Correct me if I am wrong. Thanks Tom.

sgl-kernel/csrc/elementwise/pos_enc.cuh
JeremieMelo: for gpt-oss, which uses GQA, q_rope copy is for tensor shape of [seq, 64, 64], k_rope copy is [seq, 8, 64].
we add k_buffer copy [seq, 8, 64] and v_buffer copy [seq, 8, 64] to the k branch, I guess it is more balanced? Correct me if I am wrong. Thanks Tom.
You are right, for normal MHA, it is imbalanced.

sgl-kernel/csrc/elementwise/pos_enc.cuh
JeremieMelo: I am unsure about the best practice for such modifications, e.g., ensuring compatibility. I currently use this style to avoid touching any old files/functions/logic, and introduce fully standalone files to decouple. Any suggestion to make it more compatible /extensible for future updates is appreciated.

fzyzcjy: I personally think the fuse is only a dozen lines thus can be inlined. note that you can use c++ template and constexpr if there are overheads

sgl-kernel/csrc/elementwise/pos_enc.cuh
fzyzcjy: yes I mean `k_buffer[loc] = data`, i.e. do a indexing using that loc tensor. `cache_offset = cache_loc[idx]` roughly lgtm (though not think in detail, but the general direction is right)

sgl-kernel/csrc/elementwise/pos_enc.cuh
fzyzcjy: good point for gpt-oss!

python/sglang/srt/layers/rotary_embedding.py
```diff
@@ -222,17 +222,52 @@ def forward_cuda(
         query: torch.Tensor,
         key: torch.Tensor,
         offsets: Optional[torch.Tensor] = None,
+        layer: Any = None,  # RadixAttention
+        forward_batch=None,
```
fzyzcjy: maybe we should not put layer,forward_batch etc into such a low-level. I personally suggest to put the k_buffer_ptr, v_buffer_ptr, etc, down to here. If too verbose, maybe make a simple @dataclass to pack them

python/sglang/srt/models/gpt_oss.py
```diff
@@ -293,15 +293,39 @@ def forward_prepare(
             return hidden_states, forward_batch, None
         qkv, _ = self.qkv_proj(hidden_states)
         q, k, v = qkv.split([self.q_size, self.kv_size, self.kv_size], dim=-1)
-        q, k = self.rotary_emb(positions, q, k)
+
+        if forward_batch.attn_backend.__class__.__name__ == "TRTLLMHAAttnBackend":
```
fzyzcjy: this if may be improved

sgl-kernel/csrc/elementwise/pos_enc.cuh
```diff
@@ -0,0 +1,344 @@
...
+    size_t v_stride_h,
+    size_t q_rope_stride_n,
+    size_t q_rope_stride_h,
+    size_t k_rope_stride_n,
+    size_t k_rope_stride_h,
+    size_t k_buffer_stride_n,
+    size_t k_buffer_stride_h,
+    size_t v_buffer_stride_n,
+    size_t v_buffer_stride_h,
+    float k_scale,
```
fzyzcjy: this seems pass in but unused, i.e. may be buggy (but we can just rm all these since iirc we do not use it)

sgl-kernel/csrc/elementwise/pos_enc.cuh
```diff
@@ -0,0 +1,344 @@
...
+        k_vec = vec_apply_llama_rope_cos_sin_interleave_reuse_half<vec_size, bdx>(k_ptr, cos, sin, rotary_dim);
+      } else {
+        k_vec = vec_apply_llama_rope_cos_sin<vec_size, bdx>(k_ptr, cos, sin, rotary_dim);
+      }
+      k_vec.cast_store(k_rope_ptr + tx * vec_size);
+      k_vec.cast_store(k_buffer_ptr + tx * vec_size);  // store the k_rope to the k_buffer
+      // v.cast_store(v_buffer_ptr + tx * vec_size);      // store the v to the v_buffer
+      vec_t<float, vec_size> v_vec;
+      v_vec.cast_load(v_ptr + tx * vec_size);
+      v_vec.cast_store(v_buffer_ptr + tx * vec_size);
```
fzyzcjy: hmm wondering whether a bit inefficient

fzyzcjy:
I have meetings in 40min... let me briefly write it partially now

BBuf (approved):

zhyncs:
@JeremieMelo please fix the conflicts

fzyzcjy:
it's me

zhyncs (approved):

PR #9045, opened 2025-08-11
feat(hicache): Support passing prefix keys for l3 store.

hzh0425:
## Motivation

Effective lifecycle management of the ​L3 store​ is critically important in production environments.

In the previous implementation, ​hicache​ read/wrote data to the L3 store using a Key-Value (KV) format, ​but don't transparently transmit prefix information of KV blocks​ to the L3 store. 

This would cause:

1. When ​L3 TTL is enabled: High probability of the L3 store ​deleting prefix blocks, resulting in failure to match subsequent blocks.

This Pull Request aims to:

- ​Transmit complete request prefix information​ to the L3 Store;
- L3 backend connector can rebuild the missing pages by the prefix information.

help cc @xiezhq-hermann 

CoAuthor: @pansicheng 

## Checklist

- [ ] Format your code according to the [Code formatting with pre-commit](https://docs.sglang.ai/references/contribution_guide.html#code-formatting-with-pre-commit).
- [ ] Add unit tests according to the [Running and adding unit tests](https://docs.sglang.ai/references/contribution_guide.html#running-unit-tests-adding-to-ci).
- [ ] Update documentation according to [Writing documentations](https://docs.sglang.ai/references/contribution_guide.html#writing-documentation-running-docs-ci).
- [ ] Provide accuracy and speed benchmark results according to [Testing the accuracy](https://docs.sglang.ai/references/contribution_guide.html#testing-the-accuracy) and [Benchmark and profiling]()


xiezhq-hermann:
Thank you @hzh0425 for the PR, I think the implementation might be a little bit overkill. An easier alternatives I can  think of, is to just pass all prior host indices and hash values into `write_storage` function, so the storage backend get the whole information and fill potential holes. But it will come with overhead of excessive existence checking, which this PR would encounter as well, which seems to be the critical problem we want to solve here.

hzh0425:
> Thank you @hzh0425 for the PR, I think the implementation might be a little bit overkill. An easier alternatives I can think of, is to just pass all prior host indices and hash values into `write_storage` function, so the storage backend get the whole information and fill potential holes. But it will come with overhead of excessive existence checking, which this PR would encounter as well, which seems to be the critical problem we want to solve here.

 Thanks for the reply, let me make some adjustments.

xiezhq-hermann:
Can you try this PR out see does it solve this dependency reconstruct problem? We can always add something off the critical paths in the future. https://github.com/sgl-project/sglang/pull/9797

python/sglang/srt/managers/cache_controller.py
```diff
@@ -191,12 +191,14 @@ def __init__(
...
         last_hash: Optional[str] = None,
         hash_value: Optional[List[str]] = None,
+        previous_keys: Optional[List[str]] = None,
     ):
         self.host_indices = host_indices
         self.token_ids = token_ids
         self.last_hash = last_hash
         self.completed_tokens = 0
         self.hash_value = hash_value if hash_value is not None else []
+        self.previous_keys = previous_keys
```
pansicheng: Can we directly pass the TreeNode itself? For example, StorageOperation could maintain the current TreeNode awaiting backup, while PrefetchOperation could track the last matched TreeNode in the radix tree.

Later, we could consider passing the TreeNode directly to the storage_backend's batch_set and batch_get methods. This would allow the storage backend to determine whether a recursive prefix lookup is necessary, rather than requiring the cache controller to preemptively resolve all prefixes. As a result, backends that don’t require prefix lookups wouldn’t be forced to perform them unnecessarily.

@hzh0425 @xiezhq-hermann

hzh0425: good idea

xiezhq-hermann:
it looks good to me, can you help do a rebase and we will get this merged

python/sglang/srt/server_args.py
```diff
@@ -2366,6 +2367,11 @@ def add_cli_args(parser: argparse.ArgumentParser):
             default=ServerArgs.hicache_storage_backend_extra_config,
             help="A dictionary in JSON string format containing extra configuration for the storage backend.",
         )
+        parser.add_argument(
+            "--hicache-storage-pass-prefix-keys",
```
xiezhq-hermann: can we make it part of the "hicache-storage-backend-extra-config"?

hzh0425: Sure, updated  @xiezhq-hermann

xiezhq-hermann (approved):

PR #9076, opened 2025-08-11
supported configurable eagle3 aux layers for cuda graph

FrankLeeeee:
## Motivation

Eagle3 has a list of default aux layer IDs, however, sometimes this is not enough as we may use some self-defined layer ids for hidden states. For example, GPT-OSS has a hybrid structure and we better set the aux layers to full attention layers. To enable this, we need to make aux layer IDs configurable.

## Modifications

Read the aux layer configurations before we set the aux layers in CUDA graph runner. It has been done so in the model runner class, just not in the cuda graph runner.

## Checklist

### DIFF
--- a/python/sglang/srt/model_executor/cuda_graph_runner.py
+++ b/python/sglang/srt/model_executor/cuda_graph_runner.py
@@ -27,6 +27,7 @@ import torch
 import tqdm
 from torch.profiler import ProfilerActivity, profile
 
+from sglang.srt.configs.model_config import ModelConfig
 from sglang.srt.custom_op import CustomOp
 from sglang.srt.distributed import get_tensor_model_parallel_rank
 from sglang.srt.distributed.device_communicators.pynccl_allocator import (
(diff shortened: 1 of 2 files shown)


python/sglang/srt/model_executor/cuda_graph_runner.py
```diff
@@ -331,7 +332,29 @@ def __init__(self, model_runner: ModelRunner):
...
+                eagle_config = getattr(
+                    draft_model_config.hf_config, "eagle_config", None
+                )
+                if eagle_config is None:
+                    eagle_aux_hidden_state_layer_ids = None
+                else:
+                    eagle_aux_hidden_state_layer_ids = eagle_config.get(
+                        "eagle_aux_hidden_state_layer_ids", None
+                    )
+                print(eagle_aux_hidden_state_layer_ids)
```
fzyzcjy, line 354: nit:

* maybe we can extract it to a function
* maybe we can simplify the code a bit, e.g.

```
eagle_config = getattr(draft_model_config.hf_config, "eagle_config", {})
eagle_aux_hidden_state_layer_ids = eagle_config.get("eagle_aux_hidden_state_layer_ids", None)
```

* btw the print may be removed or changed to a logger.info w/ more text

fzyzcjy (commented):
LGTM, only a nit

PR #9090, opened 2025-08-12
[CPU] fix OOM when mem-fraction is not set

ZailiWang:
## Motivation

In `server_args.py`, we set [the default `mem-fraction-static` as 0.88](https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/server_args.py#L354) when it's not set at the command line. This is well applied for accelerator devices, yet need to be adjusted for CPU servers in which the memory is shared among all the TP ranks.

On CPU, the absence of `--mem-fraction-static` setting or a straightforward setting (e.g. 0.8, 0.9) without `--max-total-tokens` specified may lead to OOM.

## Modifications

For CPU, inquire the per-numa available memory amounts with the APIs provided in `py-libnuma`.
Since the per-rank memory status is unavailable if user specify `SGLANG_CPU_OMP_THREADS_BIND`, warning message are added in both doc and code to inform that the user should take care of the memory status and avoid OOM.

Update: Due to the compatibility issue of `py-libnuma` (it does not work with py311/py313), The per-numa free memory inquiry approach has changed to invoking the APIs in `libnuma.so` directly.

## Accuracy Tests

N/A

## Benchmarking and Profiling

N/A

## Checklist


python/sglang/srt/server_args.py
```diff
@@ -391,8 +392,13 @@ def print_deprecated_warning(message: str):
         # Model-specific adjustments
         self.model_specific_adjustments()
 
-        # Set kernel backends
+        # Adjust memory fraction and set kernel backends
         if self.device == "cpu":
+            # On CPU the memory is shared by the numa nodes
+            n_numa_node: int = len(get_cpu_ids_by_node())
```
mingfeima: any chance that `get_cpu_ids_by_node()` returns a empty list?

ZailiWang: I don't think it's possible. The origination of the list is from [`lscpu` command](https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/utils.py#L2640), and it uses a try-catch with RuntimeError raised so that the failure of `lscpu` execution will directly terminates the server process. If `lscpu` is executed as expected, it will return at least 1 (the minimum possible numa node number of a server)

mingfeima (approved):

ZailiWang:
`py-libnuma` has compatibility issues with py311 & py313 (only works with py312) so the lib should not be used.
Need to try another way to retrieve per-numa free memory info.

Alcanderian (approved):

ZailiWang:
Please don't merge the PR now. Since the compatibility issue (see my comment above), I'm validating another approach to get per-numa memory info (directly invoke APIs in libnuma.so). Will update the PR if validation is completed with no issue.

Alcanderian:
> Please don't merge the PR now. Since the compatibility issue (see my comment above), I'm validating another approach to get per-numa memory info (directly invoke APIs in libnuma.so). Will update the PR if validation is completed with no issue.

Got it

mingfeima:
re-trigger CI.

ZailiWang:
Hi @Alcanderian,  plz help remove the `wip` label since it's completed now. Is the content of the PR is all right, or would you share your review comments? thanks.
Is re-trigger of the CI required after rebasing?

python/sglang/srt/utils.py
```diff
@@ -412,7 +412,30 @@ def get_available_gpu_memory(
...
+                use_errno=True,
+            )
+            libnuma.numa_max_node.restype = ctypes.c_int
+            libnuma.numa_node_size64.argtypes = [
+                ctypes.c_int,
+                ctypes.POINTER(ctypes.c_longlong),
+            ]
+            libnuma.numa_node_size64.restype = ctypes.c_longlong
+            if gpu_id < 0 or gpu_id > libnuma.numa_max_node():
+                free_gpu_memory = psutil.virtual_memory().available
```
Alcanderian: Shall we put a warning here? The 'free_gpu_memory' will differ between numa nodes that are in range and numa nodes that are out of range.

ZailiWang: Thanks for the reminder. I checked the src codes and found this path should be invalid.
This erroneous setting should have been detected at https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/model_executor/model_runner.py#L1627  at the `ModelRunner` initialization. I added a `raise ValueError` here to double confirm the validity of the TP rank index (but it is not expected to be triggered).

Alcanderian (approved):

ZailiWang:
The previous Xeon CI failure is due that the `mem-fraction-static` setting in the test case should not just be scaled linearly (0.05 -> 0.3) since some system level overheads. Per my local test, setting it as 0.4 is good.

For the other AMD/Ascend CI failures, the error were some strange "key error", and in principle it is very unlikely that this CPU specific PR would affect other platforms.

@Alcanderian would you help trigger the CI again? much appreciated.

ZailiWang:
Fixed the lint error but from source code change perspective, it's unlikely it would affect devices other than CPU (everywhere wrapped by `if device == 'cpu': `
I'll be absent next week and will further check if anything neglected when I'm back.

Alcanderian:
> Fixed the lint error but from source code change perspective, it's unlikely it would affect devices other than CPU (everywhere wrapped by `if device == 'cpu': ` I'll be absent next week and will further check if anything neglected when I'm back.

Fine, lets wait for CI

ZailiWang:
Reason for recent solution changes: the `libnuma.so` API retrieves the per-numa memory info from `/sys/devices/system/node/nodeX/meminfo`. However the "free memory" number read from this file does not include CPU memory buffers/caches, hence it is not the "available memory" amount that we need. For some large models (e.g. DeepSeek-R1 671B), the free memory is almost 0 after model loading, which unreasonably blocks the KV cache memory reservation step.

After some  study and discussion, it looks like by simply using the `available_memory()` API provided by `psutils`, which retrieves the total available memory amount, then divide it with numa node number of the machine, we can get more accurate per-rank memory amount.

Since the core distribution is unpredicted when user set `SGLANG_CPU_OMP_THREADS_BIND`, a warning message is added to notify that the user should take care of the memory usage by setting proper `--max-total-tokens`.

ZailiWang:
Xeon CI [passed](https://github.com/sgl-project/sglang/actions/runs/17613142121/job/50098523585?pr=9090), along with CI for other devices except AMD. CI failure for AMD test cases but from source code change perspective the PR is unlikely to affect AMD device.

PR #9092, opened 2025-08-12
Enable TP all-reduce overlap via dedicated NCCL stream + event handoff (opt-in)

vyom1611:
## Motivation

All-reduce on the default stream serializes compute and comm, inflating TTFT/E2E. Offloading all-reduce to a dedicated CUDA stream and delaying the consumer wait reduces exposed comm time.

### What this does 
Moves tensor-parallel all-reduce off the default CUDA stream onto a separate “comm” stream.
Immediately after enqueueing the NCCL op, records a CUDA event and attaches it to the output tensor.
Downstream code doesn’t block on the all-reduce until it actually needs the result: consumers wait on that event right before use.
Net effect: we hide a chunk of the collective’s latency under ongoing compute, reducing exposed stall time. This typically lowers time-to-first-token (TTFT) and end-to-end (E2E) latency and modestly increases throughput.
### Why it optimizes
Without overlap, compute and all-reduce serialize on the same stream.
With a dedicated comm stream, kernels keep running while NCCL progresses in parallel. The only synchronization is at the exact point where the reduced tensor is required.
When symmetric memory + pynccl is available, NCCL uses the fast path and overlap yields bigger wins.

## Modifications

Key changes

- python/sglang/srt/distributed/parallel_state.py

  1. Adds opt-in env flag SGLANG_ENABLE_TP_ALLREDUCE_OVERLAP (default: “false”).
  2. For pynccl all-reduce on symmetric memory tensors, uses a dedicated CUDA stream.
  3. After enqueue, records an event and tags the tensor with _sglang_pending_allreduce_event.
  4. Skips overlap while CUDA graphs are being captured.

- Event waits at consumption points

  1. python/sglang/srt/layers/communicator.py: wait on pending event before LayerNorm/use.
  2. python/sglang/srt/layers/linear.py: wait on pending event before returning row-parallel reduced outputs.
  3. High-traffic model call sites that consume directly already covered via the above central points.

- Server CLI
python/sglang/srt/server_args.py: --disable-tp-allreduce-overlap for hard-off switch (internally sets env “false”).

## NOTE
```
Model coverage: Overlap correctness is enforced centrally in communicator.py and row-parallel linear.py, which cover the majority of TP all-reduce consumers. I also added targeted waits to high-traffic model paths (e.g., GPT/DeepSeek/Mixtral/LLaMA variants) that immediately consume reduced tensors. Other models not touched explicitly still run correctly (default OFF or no event present), but won’t benefit from overlap unless their consumption paths pass through the guarded sites.
```

## Accuracy Tests

- Functional correctness: Event-based handoff guards all overlapped TP all-reduces. Producers tag tensors with a CUDA event; consumers (communicator.py, row-parallel linear.py, and representative model callsites) wait-and-clear before use; overlap is auto-disabled during CUDA graph capture; non-fast paths fall back to original collectives.

- Performance validation: CSV-only A/B harness spins up baseline vs overlap with small memory settings, probes readiness, runs a short bench_serving, and prints per-mode latency/throughput rows (for sanity, not as a feature). A minimal TP microbenchmark also confirms reduced exposed all-reduce time.

## Benchmarking and Profiling

model used : meta-llama/Llama-3.2-1B (4xH100s)
```
python benchmark/overlap/run_overlap_bench_csv.py \
  --models  meta-llama/Llama-3.2-1B \
  --tp 4 --base-port 30000 --max-total-tokens 16384 \
  --csv result.csv
```

### Highlights (OPTIMIZED vs BASELINE):

Requests/s: +9.3% (24.96 → 22.83)
Total tokens/s: +9.3% (20,628 → 18,869)
Mean E2E: −10.5% (1507 ms vs 1683 ms)
Mean TTFT: −11.8% (928 ms vs 1052 ms)
P99 TTFT: −9.8% (1722 ms vs 1910 ms)
Mean ITL: −8.3% (12.49 ms vs 13.62 ms)
P95 ITL: −7.8% (23.80 ms vs 25.80 ms)
P99 ITL: −34.2% (28.71 ms vs 43.63 ms)
Max ITL: −39.7% (182 ms vs 302 ms)

output-7

output-6

## Checklist


yuan-luo:
I don't think this approach can reduce the TTFT. AR is running on a separate cuda stream, but the following layer has data dependency on this AR result. Only one batch can't overlap. We need to have two batch to do overlap. Moreover, it needs to fuse the AR with rmsnorm and add so as to literally reduce TTFT for each single request. 
image
Another point is this PR only handles in-place AR, sglang also supports out-place AR which is the more common case. But thumbs up to this PR as well.

BTW, I'm working on an optimization to cover both overlap and fuse.

PR #9106, opened 2025-08-12
Add PDL support for quant kernel and rope kernel

fzyzcjy:
## Motivation

speed: 340->350
acc: 64.5 for 120B TP8 low, which is normal

## Checklist


Alcanderian:
👀 Amazing!

fzyzcjy:
@Alcanderian could you please have a review about this :)

Alcanderian (approved):

sgl-kernel/python/sgl_kernel/elementwise.py
```diff
@@ -307,6 +308,10 @@ def apply_rope_with_cos_sin_cache_inplace(
     if cos_sin_cache.dtype != torch.float32:
         raise ValueError("cos_sin_cache should be float32")
 
+    if enable_pdl is None:
+        # the non-fused branch does not yet support PDL, but after we switch to our impl for that branch it will
+        enable_pdl = is_arch_support_pdl() and (fused_set_kv_buffer_arg is not None)
```
zhyncs, line 313: do we need to invoke `is_arch_support_pdl` every time

fzyzcjy: I follow the existing code pattern, and it seems that func has a lru_cache so may be not very expensive

image

PR #8279, opened 2025-07-23
[4/n] DP Enhancement: Optimize communication when `dp < tp` by using `all_gather_into_tensor` and `reduce_scatter_tensor`

ch-wan:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #8279
* #8278
* #8277
* #8276

## Motivation and Modifications

#8278 padded the token size to a multiple of `attn_tp_size`. As a result, each DP rank's hidden states can be evenly scattered across its TP group. This enables the use of `reduce_scatter_tensor` and `all_gather_into_tensor` to optimize communication efficiency.

### TP MoE

Benchmark command:

```bash
# launch
python3 -m sglang.launch_server --model-path /dev/shm/DeepSeek-V3-0324 --tp-size 8 --enable-dp-attention --dp-size 4 --enable-deepep-moe --trust-remote-code  --chunked-prefill-size 4096 --cuda-graph-max-bs 256 --max-running-requests 1024 --enable-dp-lm-head --moe-dense-tp-size 1 --mem-fraction-static 0.75 --disable-radix-cache
# benchmark
python3 -m sglang.bench_one_batch_server --model /dev/shm/DeepSeek-V3-0324 --base-url http://localhost:30000/ --batch-size 4096 --input-len 128 --output-len 32
```

Output throughput: 

| Main | #8278 | #8279 |
|---|---|---|
| 4154.71 | 4163.29 | 4261.29

### DeepEP MoE

Benchmark command:

```bash
# launch
python3 -m sglang.launch_server --model-path /dev/shm/DeepSeek-V3-0324 --tp-size 8 --enable-dp-attention --dp-size 4 --enable-deepep-moe --trust-remote-code  --chunked-prefill-size 4096 --cuda-graph-max-bs 128 --max-running-requests 512 --enable-deepep-moe --enable-dp-lm-head --moe-dense-tp-size 1 --mem-fraction-static 0.75 --disable-radix-cache 
# benchmark
python3 -m sglang.bench_one_batch_server --model /dev/shm/DeepSeek-V3-0324 --base-url http://localhost:30000 --batch-size 4096 --input-len 128 --output-len 32
```

Output throughput: 

| Main | #8278 | #8279 |
|---|---|---|
| 2405.84 | 2397.82 | 2443.36 |

PR #9107, opened 2025-08-12
[5/n] DP Enhancement: Correct `num_token_non_padded`

ch-wan:
## Motivation

`num_token_non_padded` is used for masking padded tokens in MoE compute. When `dp<tp`, the current implementation overestimates non padded tokens because its value is token size in the current DP rank but tokens are sharded across DP group under DeepEP. This PR fixes this issue.

## Accuracy Tests

gsm8k's accuracy is correct under different concurrency.

## Benchmarking and Profiling

```
python3 -m sglang.launch_server --model-path deepseek-ai/DeepSeek-V3-0324 --trust-remote-code --tp 8 --enable-dp-attention --dp 2 --moe-dense-tp-size 1 --moe-a2a-backend deepep --cuda-graph-max-bs 32 --max-running-requests 32 --speculative-algo EAGLE --speculative-draft lmsys/DeepSeek-V3-0324-NextN --speculative-num-steps 2 --speculative-eagle-topk 4 --speculative-num-draft-tokens 4 --disable-radix-cache --stream-output
python3 -m sglang.bench_serving --backend sglang --dataset-name random --num-prompt 32 --random-input 512 --random-output 32 --random-range-ratio 1 --host 127.0.0.1 --port 30000 --max-concurrency 1
```

Before this PR, the output throughput is 53.05 tok/s. With this PR, the output throughput is 57.07 tok/s.

## Checklist

### DIFF
--- a/python/sglang/srt/model_executor/cuda_graph_runner.py
+++ b/python/sglang/srt/model_executor/cuda_graph_runner.py
@@ -33,7 +33,11 @@ from sglang.srt.distributed.device_communicators.pynccl_allocator import (
     set_graph_pool_id,
 )
 from sglang.srt.distributed.parallel_state import GroupCoordinator, graph_capture
-from sglang.srt.layers.dp_attention import DPPaddingMode, get_attention_tp_size
+from sglang.srt.layers.dp_attention import (
+    DPPaddingMode,
+    get_attention_tp_rank,
+    get_attention_tp_size,
+)
 from sglang.srt.layers.logits_processor import LogitsProcessorOutput
 from sglang.srt.layers.torchao_utils import save_gemlite_cache
 from sglang.srt.model_executor.forward_batch_info import (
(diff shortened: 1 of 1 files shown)


miter6:
thanks a lot.
I got this issue yesterday.

miter6:
Capture cuda graph bs [1, 2, 4, 8, 16, 24, 32, 40, 48, 56, 64] for dp-attention=16. 
Capture cuda graph bs [16, 32, 48, 64] for tp=16.     
If running batchsize is below 16, it must padded to 16. Is this a bug??

ch-wan:
@MiterV1 This is expected when attn tp size is 16. We have this constraint to keep reduce-scatter and all-gather efficient (see #8279). In addition, when the running request is small, EP + DP attention is not an optimal approach because its workload is imbalanced.

PR #9127, opened 2025-08-12
[MTP] Force greedy sampling on AMD

datdo-msft:
## Motivation

In EAGLE utils, CUDA kernels are used as part of sampling tokens, specifically getting target probs here: https://github.com/valarLip/sglang/blob/8d463fe351c4b10ed8b65d7fe5ed0c68eea6d731/python/sglang/srt/speculative/eagle_utils.py#L370. There are already [existing checks](https://github.com/valarLip/sglang/blob/8d463fe351c4b10ed8b65d7fe5ed0c68eea6d731/python/sglang/srt/speculative/eagle_utils.py#L24) for `is_cuda()` and `is_hip()` which dictates which kernels get imported. As a result, we should treat all requests as greedy sampling on AMD.

## Checklist


datdo-msft:
Hi @Ying1123 and @merrymercy , please help review and let me know if this makes sense. Thanks!

python/sglang/srt/speculative/eagle_utils.py
```diff
@@ -410,8 +410,8 @@ def verify(
                 logits=logits_output.next_token_logits, vocab_mask=vocab_mask
             )
 
-        # Sample tokens
-        if batch.sampling_info.is_all_greedy:
+        # Sample tokens. Force greedy sampling on AMD
+        if batch.sampling_info.is_all_greedy or is_hip():
```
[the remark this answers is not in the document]
HaiShaw: @datdo-msft please take this suggest change, so that when forward improvement is made, we will have it more clearly.

HaiShaw (changes requested):
It is w.i.p. Still better to take code-assist's

datdo-msft:
@HaiShaw , updated please help take a look again. Thanks!

HaiShaw (approved):
LGTM

hubertlu-tw:
@datdo-msft could you please run the following commands to fix lint errors:
```
pip3 install pre-commit
pre-commit run --all-files
```

python/sglang/srt/speculative/eagle_utils.py
```diff
@@ -410,8 +410,11 @@ def verify(
                 logits=logits_output.next_token_logits, vocab_mask=vocab_mask
             )
 
-        # Sample tokens
-        if batch.sampling_info.is_all_greedy:
+        # Sample tokens. Force greedy sampling on AMD
```
merrymercy, line 428: We should print a warning here

merrymercy (changes requested):

python/sglang/srt/speculative/eagle_utils.py
datdo-msft: Done

python/sglang/srt/speculative/eagle_utils.py
```diff
@@ -423,8 +423,15 @@ def verify(
...
+        if (
+            batch.sampling_info.is_all_greedy
+            or "tree_speculative_sampling_target_only" not in globals()
+        ):
+            logger.warning(
```
merrymercy, line 431: Only print this warning for amd!

merrymercy (changes requested):

python/sglang/srt/speculative/eagle_utils.py
```diff
@@ -423,8 +425,15 @@ def verify(
                 logits=logits_output.next_token_logits, vocab_mask=vocab_mask
             )
 
-        # Sample tokens
-        if batch.sampling_info.is_all_greedy:
+        # Sample tokens. Force greedy sampling on AMD
+        is_all_greedy = sampling_info.is_all_greedy
+        if (not is_all_greedy) and (not TREE_SPEC_KERNEL_AVAILABLE) and is_hip():
```
merrymercy: drop `is_hip`

merrymercy (changes requested):

python/sglang/srt/speculative/eagle_utils.py
datdo-msft: Done

merrymercy (approved):

PR #9131, opened 2025-08-13
[chore] Clean up redundant lora_weight_names concept to simplify code 

lifuhuang:
## Motivation

Previously, we had multiple different entities to represent a very similar concepts:
- `weight_name`: used in LoRAAdapter to represent the actual full weight name, e.g., `base_model.model.model.layers.0.self_attn.o_proj.lora_A.weight`
- `lora_weight_name`: represent the normalized "module type name" used in SGLang, e.g., `qkv_proj`, `kv_proj`
- `target_module`: represent the "module type name", e.g., `qkv_proj`

This differentiation between `target_module` and `lora_weight_name` was necessary in the past because flashinfer backend used different names for weight A (`qkv_proj`) and weight B (`kv_proj`, `q_proj`). However, now that it has been deprecated, these two concepts are now essentially storing the same thing twice. 

This PR unify them by removing the `lora_weight_names`, because:
1. `target_modules` is already part of our user-facing interface (serverargs) and it is also a standard field in PEFT config.
2. `lora_weight_name` arguably could be confused with the actual full weight name, as is used correctly in LoRAAdapter. _(to be fair, techncially `target_modules` could similarly be confused with the full target module name. Personally, I believe a more accurate name would be something like `target_module_types`, but I kept it as-is as it has been unfortunately made standard thanks to PEFT.)

## Checklist


Fridge003 (approved):
LGTM

PR #9149, opened 2025-08-13
[PD] optimize kv cache transfer directly using batch transfer

ssssnow:
## Motivation
To accelerate the KV Cache transfer via batch transfer interface.

## Modifications
Pack all layers' transfer parameters to one single batch. Call batch transfer interface directly instead of using multiple executors to fully unleash the potential of batch transfer.

## Accuracy Tests
This PR does not involve change of computation or communication process. It's just pack transfer task into one single batch task, so there shouldn't be any effect on accuracy.

## Benchmarking and Profiling
We tested the transfer speed on DeepSeek V3, using Prefill PP2 + Decode TP16 + input 10240
Baseline:
use multiple executor to trigger batch transfer per layer
transfer speed : 0.6 - 1.1 GB/s
transfer time per rank : 180+ ms
image

Optimize:
directly call batch transfer, do not use executor
transfer speed : 3.8 - 3.9 GB/s
trasnsfer time per rank : 80+ ms
image

## Checklist

### DIFF
--- a/python/sglang/srt/disaggregation/mooncake/conn.py
+++ b/python/sglang/srt/disaggregation/mooncake/conn.py
@@ -356,7 +356,7 @@ class MooncakeKVManager(BaseKVManager):
             ]
         assert layers_params is not None
 
-        def set_transfer_blocks(src_ptr: int, dst_ptr: int, item_len: int) -> int:
+        def set_transfer_blocks(src_ptr: int, dst_ptr: int, item_len: int) -> List[Tuple[int, int, int]]:
             transfer_blocks = []
             for prefill_index, decode_index in zip(prefill_kv_blocks, dst_kv_blocks):
                 src_addr = src_ptr + int(prefill_index[0]) * item_len
@@ -371,7 +371,7 @@ class MooncakeKVManager(BaseKVManager):
             return self._transfer_data(mooncake_session_id, transfer_blocks)
         
         # Worker function for processing all layers in a batch
-        def process_layers(layers_params):
+        def process_layers(layers_params: List[Tuple[int, int, int]]) -> int:
             transfer_blocks = []
             for src_ptr, dst_ptr, item_len in layers_params:
                 transfer_blocks.extend(set_transfer_blocks(src_ptr, dst_ptr, item_len))
--- a/python/sglang/srt/disaggregation/mooncake/conn.py
+++ b/python/sglang/srt/disaggregation/mooncake/conn.py
@@ -356,33 +356,47 @@ class MooncakeKVManager(BaseKVManager):
             ]
         assert layers_params is not None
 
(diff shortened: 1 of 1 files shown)


python/sglang/srt/disaggregation/mooncake/conn.py
```diff
@@ -356,33 +356,47 @@ def send_kvcache(
...
-        def process_layer(src_ptr: int, dst_ptr: int, item_len: int) -> int:
+        def set_transfer_blocks(src_ptr: int, dst_ptr: int, item_len: int) -> int:
             transfer_blocks = []
             for prefill_index, decode_index in zip(prefill_kv_blocks, dst_kv_blocks):
                 src_addr = src_ptr + int(prefill_index[0]) * item_len
                 dst_addr = dst_ptr + int(decode_index[0]) * item_len
                 length = item_len * len(prefill_index)
                 transfer_blocks.append((src_addr, dst_addr, length))
-
+            return transfer_blocks
```
ShangmingCai, line 368: Seems like Gemini is right, the return type should be fixed.

python/sglang/srt/disaggregation/mooncake/conn.py
```diff
@@ -356,33 +356,47 @@ def send_kvcache(
...
-                for f in futures:
-                    f.cancel()
-                return status
+        
+        # Worker function for processing all layers in a batch
+        def process_layers(layers_params):
+            transfer_blocks = []
+            for src_ptr, dst_ptr, item_len in layers_params:
+                transfer_blocks.extend(set_transfer_blocks(src_ptr, dst_ptr, item_len))
+            return self._transfer_data(mooncake_session_id, transfer_blocks)
```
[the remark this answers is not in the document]
ShangmingCai: ditto, nice suggestion from Gemini.

python/sglang/srt/disaggregation/mooncake/conn.py
ssssnow: all fixed

python/sglang/srt/disaggregation/mooncake/conn.py
```diff
@@ -356,33 +356,47 @@ def send_kvcache(
...
+            ]
+            for future in concurrent.futures.as_completed(futures):
+                status = future.result()
+                if status != 0:
+                    for f in futures:
+                        f.cancel()
+                    return status
+        else:
+            # use batch transfer, so we pass all layers params to one thread
+            # compared to use multiple executors, this is more efficient
```
ShangmingCai:
```suggestion
            # Combining all layers' params in one batch transfer is more efficient 
            # compared to using multiple threads
```

ShangmingCai (approved):
LGTM!

PR #9153, opened 2025-08-13
Support manually set dp_ranks for slow_down

ollybbmonster:
## Motivation

Support specifying dp_ranks when sending a /slow_down request. This enables targeted slowdown for specific dp_ranks, providing finer-grained control in test.

## Modifications

Added support for passing a dp_ranks_for_slowdown parameter in the /slow_down API endpoint.
Updated the slowdown handling logic in the scheduler to apply the slowdown only to the specified data parallel ranks.
Set {"dp_ranks_for_slowdown": 1, "forward_sleep_time": 0.08} or {"dp_ranks_for_slowdown": [0, 1, 2, 3], "forward_sleep_time": 0.08} for multi dp ranks.

## Checklist

### DIFF
--- a/python/sglang/srt/managers/io_struct.py
+++ b/python/sglang/srt/managers/io_struct.py
@@ -910,6 +910,7 @@ class ResumeMemoryOccupationReqOutput:
 @dataclass
 class SlowDownReqInput:
     forward_sleep_time: Optional[float]
+    dp_ranks: Optional[Union[int, List[int]]] = None
 
 
 @dataclass
--- a/python/sglang/srt/managers/scheduler.py
+++ b/python/sglang/srt/managers/scheduler.py
@@ -418,6 +418,7 @@ class Scheduler(
         if self.device == "cpu":
             self.current_stream.synchronize = lambda: None  # No-op for CPU
         self.forward_sleep_time = None
+        self.dp_ranks_for_slowdown = None
 
         # Init chunked prefill
         self.chunked_prefill_size = server_args.chunked_prefill_size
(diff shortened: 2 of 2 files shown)


python/sglang/srt/managers/io_struct.py
```diff
@@ -910,6 +910,7 @@ class ResumeMemoryOccupationReqOutput:
 @dataclass
 class SlowDownReqInput:
     forward_sleep_time: Optional[float]
+    dp_ranks_for_slowdown: Optional[Union[int, List[int]]] = None
```
acelyc111: Now that the request is named **SlowDown**ReqInput, the 'for_slowdown' in this field maybe redundant?

python/sglang/srt/managers/scheduler.py
```diff
@@ -1713,9 +1714,10 @@ def run_batch(
 
         # Whether to run the profiler
         self._profile_batch_predicate(batch)
-        if self.forward_sleep_time is not None:
-            logger.info(f"Scheduler.run_batch sleep {self.forward_sleep_time}s")
-            time.sleep(self.forward_sleep_time)
+        if self.dp_ranks_for_slowdown is not None and self.dp_rank in self.dp_ranks_for_slowdown:
+            if self.forward_sleep_time is not None:
+                logger.info(f"DP{self.dp_rank}: Scheduler.run_batch sleep {self.forward_sleep_time}s")
```
acelyc111: `DP{self.dp_rank}: ` is not necessary because the logger is prefixed with it's DP/EP/TP ranks already.

acelyc111 (approved):

PR #9160, opened 2025-08-13
fix(quantization): Add missing w8a8_int8 quantization scheme

SethMess:
This commit adds the missing `compressed_tensors_w8a8_int8.py` file to the SGLang quantization schemes and updates the necessary import statements.

The change addresses a `NameError` that occurred when trying to load models using the `w8a8` quantization format, such as `Qwen2-0.5B-Instruct-quantized.w8a8`. The SGLang codebase contained the logic to handle this quantization scheme but lacked the file defining the `CompressedTensorsW8A8Int8` class, causing the model to fail during initialization.

The implementation file was sourced from the vLLM repository, which SGLang's `compressed_tensors` module is adapted from. This change ensures that SGLang can correctly process models that rely on this quantization method.

## Motivation
This PR addresses an issue where SGLang fails to load models using the w8a8 quantization format, such as Qwen2-0.5B-Instruct-quantized.w8a8. The core problem was a NameError that occurred because the compressed_tensors module, which is adapted from vLLM, was missing the implementation file for this specific quantization scheme. This fix is necessary to enable support for a wider range of quantized models.

## Modifications
This commit adds the missing compressed_tensors_w8a8_int8.py file to the SGLang codebase at sglang/srt/layers/quantization/compressed_tensors/schemes/. The implementation file was sourced directly from the vLLM repository to maintain consistency with the adapted module.

In addition, the __init__.py file within the same directory was updated to correctly import the new CompressedTensorsW8A8Int8 class. This ensures the class is available to the rest of the SGLang library, resolving the NameError that previously prevented models using this scheme from initializing.

## Accuracy Tests
No accuracy tests are required for this change as it is a bug fix that enables model loading rather than altering the model's output or performance. The change simply provides the missing code for a quantization scheme that SGLang was already attempting to use.

## Benchmarking and Profiling
No performance benchmarking or profiling is needed. This PR addresses a model loading failure and does not impact inference speed. The model, once loaded, will use its standard kernels.

## Checklist


weixlu:
@SethMess Hi Seth. I'm not a reviewer, but I have faced the same error as yours in this pr. I used your code, and it works well. I would like to express my thanks to you : )

hnyls2002:
Thanks @SethMess! The w8a8_int8 compressed-tensors scheme is supported on main now, so I'm closing this
  as superseded. Please reopen if I've missed a gap.

PR #9171, opened 2025-08-14
Fix docker container DeepEP error on Blackwell

fzyzcjy:
## Motivation

test: not run the full image, but only run that single command on an existing container and see the error disappears

## Checklist

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -102,7 +102,7 @@ RUN cd /sgl-workspace/nvshmem && \
 
 # Install DeepEP
 RUN cd /sgl-workspace/DeepEP && \
-    NVSHMEM_DIR=${NVSHMEM_DIR} pip install .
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -102,7 +102,7 @@ RUN cd /sgl-workspace/nvshmem && \
 
 # Install DeepEP
 RUN cd /sgl-workspace/DeepEP && \
-    NVSHMEM_DIR=${NVSHMEM_DIR} pip install .
+    TORCH_CUDA_ARCH_LIST='9.0;10.0' NVSHMEM_DIR=${NVSHMEM_DIR} pip install .
```
zhyncs, line 105: have we tested on gb200

zhyncs: https://github.com/sgl-project/sglang/blob/main/docker/Dockerfile.gb200

nvm

Issue #9155, opened 2025-08-13
[Bug] start failed with --host 0.0.0.0 option


wadeLouis:
### Checklist

- [ ] 1. I have searched related issues but cannot get the expected help.
- [ ] 2. The bug has not been fixed in the latest version.
- [ ] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [ ] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [ ] 5. Please use English, otherwise it will be closed.

### Describe the bug

When I start with the --host 0.0.0.0 option, the following issue occurs. I don't know why this is happening.
[2025-08-13 11:01:02] INFO:     Started server process [132351]
[2025-08-13 11:01:02] INFO:     Waiting for application startup.
[2025-08-13 11:01:02] Ignoring mcp import error
[2025-08-13 11:01:02] Ignoring mcp import error
[2025-08-13 11:01:02] INFO:     Application startup complete.
[2025-08-13 11:01:02] INFO:     Uvicorn running on http://0.0.0.0:25600 (Press CTRL+C to quit)
[2025-08-13 11:01:03] Starting new HTTP connection (1): 172.16.37.200:3138
[2025-08-13 11:01:03] http://172.16.37.200:3138 "GET http://0.0.0.0:25600/get_model_info HTTP/1.1" 403 3582
[2025-08-13 11:01:04] Starting new HTTP connection (1): 172.16.37.200:3138
[2025-08-13 11:01:04] http://172.16.37.200:3138 "GET http://0.0.0.0:25600/get_model_info HTTP/1.1" 403 3582
[2025-08-13 11:01:05] Starting new HTTP connection (1): 172.16.37.200:3138
[2025-08-13 11:01:05] http://172.16.37.200:3138 "GET http://0.0.0.0:25600/get_model_info HTTP/1.1" 403 3582
[2025-08-13 11:01:06] Starting new HTTP connection (1): 172.16.37.200:3138
....
[2025-08-13 11:03:03] Initialization failed. warmup error: Traceback (most recent call last):
[...]

PR #9175, opened 2025-08-14
fix(server_args): handle special IP addresses in url generation

csbo98:
## Modifications

According to RFC 1122 §3.2.1.3(https://datatracker.ietf.org/doc/rfc1122/), addresses such as 0.0.0.0 and :: are only allowed to be used as source addresses when a server is initializing itself, and cannot be used as destination addresses for client requests.

Directly accessing 0.0.0.0 and :: from a client may also lead to compatibility issues, as many clients and HTTP proxies will outright reject or specially handle targets like 0.0.0.0 and ::. For example, see #9155.

Therefore, when the host is specified as 0.0.0.0 or ::, the local machine should explicitly use 127.0.0.1 or ::1 to access the sglang HTTP service.

## Checklist


csbo98:
@CatherineSue @slin1237 @ispobock Could you please review this code?

hnyls2002:
Thanks @csbo98! ServerArgs.url() now maps special addresses to loopback on main, so I believe this is
  handled - closing. Please reopen if I've missed a case.

PR #9192, opened 2025-08-14
bugfix for kimi reasoning_parser

RayKolWe:
think symbol garbled text repair

## Motivation

bugfix for kimi reasoning_parser

## Modifications

think symbol garbled text repair

## Checklist


JustinTong0323:
Which model are you testing? For `moonshotai/Kimi-VL-A3B-Thinking` I think this think token is correct.

PR #9198, opened 2025-08-14
[router] allow more health check configuration

slin1237:
## Motivation

The current health check implementation in sgl-router immediately marks workers as unhealthy on the first failed health check, which can lead to false positives when workers are temporarily busy processing requests. This PR implements more robust health check configuration with consecutive success/failure thresholds to make health checking more robust and configurable.

Additionally, the router was using the same interval configuration (`worker_startup_check_interval`) for both startup checks and runtime health monitoring, and the health check endpoint was hardcoded to `/health`. This PR separates these concerns and makes them independently configurable.

### 1. **Health Check Thresholds**
- Added `consecutive_failures` and `consecutive_successes` counters to track health check results
- Workers now require multiple consecutive failures before being marked unhealthy (default: 3)
- Workers need multiple consecutive successes to recover and be marked healthy (default: 2)
- This prevents transient issues from causing unnecessary worker removals

### 2. **Separate Runtime Health Check Configuration**
- Added `health_check_interval_secs` (default: 60s) for runtime health monitoring, separate from `worker_startup_check_interval`
- Added `health_check_endpoint` (default: "/health") to allow custom health check endpoints
- Added `health_check_timeout_secs` (default: 5s) for health check request timeouts

### 3. **Configuration Updates**
- **Python Launcher**: Added new CLI arguments:
  - `--health-failure-threshold`: Number of consecutive failures before marking unhealthy
  - `--health-success-threshold`: Number of consecutive successes before marking healthy
  - `--health-check-timeout-secs`: Timeout for health check requests
  - `--health-check-interval-secs`: Interval between runtime health checks
  - `--health-check-endpoint`: Health check endpoint path

- **Rust CLI**: Added corresponding CLI arguments with the same names and defaults

- **Configuration Types**: Added `HealthCheckConfig` struct to encapsulate all health check settings

### 4. **Default Values Made More Permissive**
As requested, default values across the router have been made more loose/permissive:
- Worker startup timeout: 300s → 600s
- Worker startup check interval: 10s → 30s
- Cache threshold: 0.5 → 0.3
- Balance thresholds: More tolerant of imbalance
- Request timeout: 600s → 1800s (30 minutes)
- Max concurrent requests: 64 → 256
- Max payload size: 256MB → 512MB
- Retry configuration: More retries with gentler backoff
- Circuit breaker: Higher failure tolerance

### 5. **Worker Implementation Updates**
- Updated `BasicWorker` and `DPAwareWorker` to use the new health check logic
- Health check configuration is properly propagated to all worker types (Regular, Prefill, Decode)
- Workers maintain their health status through the threshold mechanism

### Example Usage

```bash
# Using custom health check configuration
sglang-router \
  --worker-urls http://worker1:8000 http://worker2:8000 \
  --health-failure-threshold 5 \
  --health-success-threshold 3 \
  --health-check-interval-secs 30 \
  --health-check-endpoint /healthz \
  --health-check-timeout-secs 10
```

## Checklist


tonyluj:
> which can lead to false positives when workers are temporarily busy processing requests

+1 — this will really help us. We ran into this during benchmarking: the default one-shot 5s timeout check can be missed under heavy load.

PR #9224, opened 2025-08-15
[Feature][PD] Support parallel_sample_num n > 1 for PD

ZhengWG:
## Motivation

Support parallel_sample_num > 1 with PD disggregation mode.

## Modifications

The proposed solution is designed as follows: 

Retain the preliminary run logic for Prefill without kv-cache transmission. Subsequent n Prefill requests will follow normal inference/kv_cache transmission while maximizing prefix-cache reuse. 
Decode requests will be treated as n independent requests, executing the original kv_cache reception and inference process, with request aggregation performed upon response return. The following additions are required:

1. Each of the n sub-requests must carry its own bootstrap room, passed through minilb.

2. During the Prefill preliminary run, dummy transmission must be performed based on the fake_connector."

TODO:

- [ ] support generate request
- [ ] add uni-test

## Checklist


ShangmingCai (commented):
mini_lb is no longer maintained. If this parallel sample case is a popular request, please reopen and rebase this PR. (But I think maybe this req can be separated into multiple reqs at the frontend)

PR #9241, opened 2025-08-15
Support GC Freezing to improve latency & throughput

chanh:
## Motivation

We are seeing the server stall for 100ms - 300ms at a time every 1.5secs or so. This is very very bad for example when we are trying to maintain a 500ms P99 latency.

The culprit was the garbage collector doing a gen2 cycle detection through a large number of objects.

In general, it is usually helpful to start the server and warm it up with real requests to initialize many of the long-lived objects that do not need to be garbage collected. After sufficient warmup, the user can call this freeze_gc API so that all objects created before this point are considered out of scope for garbage collection. This greatly reduces both the need to do gen2 GC and the time it takes when it triggers.

## Checklist


fzyzcjy:
cc @hnyls2002 for this

python/sglang/srt/server_args.py
```diff
@@ -1157,6 +1158,12 @@ def add_cli_args(parser: argparse.ArgumentParser):
             default=ServerArgs.collect_tokens_histogram,
             help="Collect prompt/generation tokens histogram.",
         )
+        parser.add_argument(
+            "--gc-warning-threshold-secs",
+            type=float,
+            default=ServerArgs.gc_warning_threshold_secs,
+            help="The threshold for long GC warning. If a GC takes longer than this, a warning will be logged. Set to 0 to disable.",
+        )
```
hnyls2002, line 1181: I suggest using a negative number to disable it.

hnyls2002 (approved):

hnyls2002:
@chanh Please fix the lint use `pre-commit run --all-files`

PR #9249, opened 2025-08-16
Support pinning adapter via server args.

lifuhuang:
Follow-up PR to add server args for loading pinned adapter.


python/sglang/srt/server_args.py
```diff
@@ -151,7 +151,7 @@ class ServerArgs:
     enable_lora: Optional[bool] = None
     max_lora_rank: Optional[int] = None
     lora_target_modules: Optional[Union[set[str], List[str]]] = None
-    lora_paths: Optional[Union[dict[str, str], dict[str, LoRARef], List[str]]] = None
+    lora_paths: Union[None, dict[str, str], List[str], List[LoRARef]] = None
```
Fridge003: I think we can keep Optional[.., .., ..] as before, just to show that this argument is optional

python/sglang/srt/server_args.py
```diff
@@ -2368,13 +2380,22 @@ def init_new(server_args, dp_rank: Optional[int] = None) -> "PortArgs":
 
 class LoRAPathAction(argparse.Action):
```
Fridge003, line 2373: Can you update the help message for `--lora-paths` argument, when calling `parser.add_argument('--lora-paths')`?

python/sglang/test/runners.py
```diff
@@ -488,7 +488,7 @@ def __init__(
         tp_size: int = 1,
         model_impl: str = "auto",
         port: int = DEFAULT_PORT_FOR_SRT_TEST_RUNNER,
-        lora_paths: List[str] = None,
+        lora_paths: Union[None, list[str], list[dict]] = None,
```
Fridge003: Also, we may use Optional instead

Fridge003 (approved):

lifuhuang:
@zhyncs can I get a merge? Thanks!

PR #9255, opened 2025-08-16
Fix scale in trtllm_mla attention backend

fzyzcjy:
## Motivation

not useful for today's dpsk r1 0528, since the scale of it is exactly 1.0 (and thus not verifiable using this model as well), but put it here in case we need it in some future models.

## Checklist


hnyls2002:
Thanks @fzyzcjy! The trtllm_mla scale handling was fixed on main, so I'm closing this. Please reopen if
  I've missed something.

PR #9256, opened 2025-08-16
[HiCache] feat: add multi tenant with prefix tag

stmatengss:
## Motivation

This PR introduces support for multi-tenancy scenarios when using HiCache with shared storage backends like Mooncake or 3fs. In global service deployments where multiple HiCache (sglang) instances share a common storage backend, we need tenant isolation to prevent key collisions. The ~~`--hicache-storage-backend-tag`~~ --hicache-storage-backend-extra-config parameter allows for logical separation of cached data by adding a tag as a prefix namespace at the storage layer.
CC. @ykwd @xiezhq-hermann 

## Modifications

1. Add a new joson attribute `storage-backend-tag` in `--hicache-storage-backend-extra-config` 
2. Add a get_hash_str_with_prefix function, and the return value is a hashing function.
3. Support prefix tag for storage backends.

Usage: --hicache-storage-backend-extra-config '{"storage_backend_tag": "xxx"}'

## Benchmarking and Profiling

```
# Start Mooncake service
python -m mooncake.http_metadata_server
mooncake_master

# Start SGLang Instance1
MOONCAKE_TE_META_DATA_SERVER="http://127.0.0.1:8080/metadata" \
MOONCAKE_GLOBAL_SEGMENT_SIZE=4294967296 \
MOONCAKE_LOCAL_BUFFER_SIZE=134217728 \
MOONCAKE_PROTOCOL="rdma" \
MOONCAKE_DEVICE="erdma_0" \
MOONCAKE_MASTER=127.0.0.1:50051 \
    --enable-hierarchical-cache \
    --hicache-storage-backend mooncake\
    --hicache-storage-backend-extra-config '{"storage_backend_tag": "deploy1"}' \
    --model-path Qwen2.5-2B-Instruct

# Start SGLang Instance2
MOONCAKE_TE_META_DATA_SERVER="http://127.0.0.1:8080/metadata" \
MOONCAKE_GLOBAL_SEGMENT_SIZE=4294967296 \
MOONCAKE_LOCAL_BUFFER_SIZE=134217728 \
MOONCAKE_PROTOCOL="rdma" \
MOONCAKE_DEVICE="erdma_0" \
MOONCAKE_MASTER=127.0.0.1:50051 \
    --enable-hierarchical-cache \
    --hicache-storage-backend mooncake\
    --hicache-storage-backend-extra-config '{"storage_backend_tag": "deploy2"}' \
    --model-path Qwen2.5-2B-Instruct

```

## Checklist


xiezhq-hermann:
Thank you for the PR, I think there is indeed a need for something to distinguish KV caches of different models and configs. We had a discussion offline, instead of adding a tag, I think it would be a better design to have a more comprehernsive config file. Let's follow-up on this : )

stmatengss:
Great point! I agree that a more comprehensive config file sounds like a much more robust and scalable solution than a simple tag. I will work on this and discuss offline. Thx!

stmatengss:
@hzh0425 Refactored code. PTAL. Thanks.

hzh0425:
> @hzh0425 Refactored code. PTAL. Thanks.

Thanks 

Can we directly using HiCacheStorageConfig to concatenate the suffix of the key, like this:
https://github.com/sgl-project/sglang/pull/9869/files#diff-771d3d16b3e77375489f7e01b78afefb0b37f763862dff210ccc32bbb0149063

stmatengss:
> > @hzh0425 Refactored code. PTAL. Thanks.
> 
> Thanks
> 
> Can we directly using HiCacheStorageConfig to concatenate the suffix of the key, like this: https://github.com/sgl-project/sglang/pull/9869/files#diff-771d3d16b3e77375489f7e01b78afefb0b37f763862dff210ccc32bbb0149063

For two sglang instances with the same model and TP size, they might share the same suffix. To ensure security and multi-tenancy, we should provide a configurable, user-defined tag. It is necessary for production deployment. I think we can discuss this PR after #9869 is merged. CC. @xiezhq-hermann

stmatengss:
PTAL, thx, @xiezhq-hermann

stmatengss:
Issue #10317 suggests a similar extra key for different LoRA adapters, but it's a pre-request and can't be applied directly to HiCache. Instead, we need an extra key for storage backend.

JustinTong0323:
> Issue #10317 suggests a similar extra key for different LoRA adapters, but it's a pre-request and can't be applied directly to HiCache. Instead, we need an extra key for storage backend.

Yes, two levels of isolation.

xiezhq-hermann (approved):

xiezhq-hermann:
I think directly changing hash_str is actually not appropriate because the hash str has very strict semantics that will carry on when creating next hash_str. It's probably better to have this tag prefix implemented in backends mapping different tags instead.

xiezhq-hermann (changes requested):
inserting tags into the hash_str is inappropriate

stmatengss:
> inserting tags into the hash_str is inappropriate

> I think directly changing hash_str is actually not appropriate because the hash str has very strict semantics that will carry on when creating next hash_str. It's probably better to have this tag prefix implemented in backends mapping different tags instead.

OK, I will add an extra tag for different storage backends, without modification in hash_str method.

stmatengss:
@xiezhq-hermann PTAL, thx

stmatengss:
This feature is implemented for the mooncake store but can be adapted for other storage backends. Omitting `storage_backend_tag` will not affect the current data path.

stmatengss:
@xiezhq-hermann PTAL, thx. The new implementations require no modifications to the cache controller.  Other storage backends can implement their own tags for future use.

xiezhq-hermann (approved):

PR #9261, opened 2025-08-17
[PD] Fix nvlink transport accuracy through transferring metadata with tcp

ShangmingCai:
## Motivation
cc: @fzyzcjy 

## Checklist


fzyzcjy:
This seems not fixed :(

test code: feat/dev_20250817, 896b56ab

example 1: non PD, no empty output

image

example 2: enable PD, enable this fix, then has 150 empty output

image

double check the code is synced:

image

fzyzcjy:
I can confirm there is no longer empty metadata using this commit

python/sglang/srt/disaggregation/mooncake/conn.py
```diff
@@ -784,9 +826,31 @@ def start_decode_thread(self):
 
         def decode_thread():
             while True:
-                (bootstrap_room, status, prefill_rank) = (
-                    self.server_socket.recv_multipart()
-                )
+                msg = self.server_socket.recv_multipart()
+                if msg[0] == b"AUX_DATA":
```
fzyzcjy: nit: maybe we can extract a const for this magic str

python/sglang/srt/disaggregation/mooncake/conn.py
```diff
@@ -784,9 +826,31 @@ def start_decode_thread(self):
...
+                    room = int(msg[1].decode("ascii"))
+                    buffer_index = int(msg[2].decode("ascii"))
+                    aux_index = int(msg[3].decode("ascii"))
+                    data_length = struct.unpack(">I", msg[4])[0]
+                    data = msg[5]
```
fzyzcjy: nit: what about extracting serialize and deserialize logic , e.g. I may do

```
class AuxDataCodec:
  @staticmethod def serialize(): ...
  @staticmethod def deserialize(): ...
```

I may also extract other things but anyway this code may be temporary so it may be ok

fzyzcjy (approved):
only some optional nits

python/sglang/srt/disaggregation/mooncake/conn.py
```diff
@@ -784,9 +845,30 @@ def start_decode_thread(self):
 
         def decode_thread():
             while True:
-                (bootstrap_room, status, prefill_rank) = (
-                    self.server_socket.recv_multipart()
-                )
+                msg = self.server_socket.recv_multipart()
+                if msg[0] == b"AUX_DATA":
```
fzyzcjy: I personally think we may make the code a bit clearer, e.g. extract such logics to func to avoid a single long func that contains many unrelated logic, but since this is temporary solution it may not be big issues

ShangmingCai: Nice suggestion. Fixed.

PR #8846, opened 2025-08-06
[PD] Support PD disaggregation with Prefill PP

ShangmingCai:
## Motivation
To reduce TTFT, we want to support Prefill PP with PD disaggregation through this PR.
 - Init commit to support mooncake with PP #8571
 - Include the deepseek PP support from #6434
 - Support Prefill PP with PD different TP size.
 - Comaptible with DP attention and pure dp

Decode PP will be supported in the next PR. @ssssnow 

## Accuracy Test
```
Qwen3-8B(Prefill TP 2 PP2, Decode TP 1):
python sglang/benchmark/gsm8k/bench_sglang.py --port 8000 --parallel 300 --num-questions 300
100%|█| 300/300 [02:29<00:00,  2.0
Accuracy: 0.943
Invalid: 0.000
Latency: 149.493 s
Output throughput: 249.315 token/s

Qwen3-8B(Prefill TP 2 PP2, Decode TP 2):
python sglang/benchmark/gsm8k/bench_sglang.py --port 8000 --parallel 300 --num-questions 300
100%|█████| 300/300 [00:24<00:00, 12.01it/s]
Accuracy: 0.947
Invalid: 0.000
Latency: 25.186 s
Output throughput: 1469.696 token/s

Qwen3-8B(Prefill TP 1 PP2, Decode TP 2):
python sglang/benchmark/gsm8k/bench_sglang.py --port 8000 --parallel 300 --num-questions 300
100%|█| 300/300 [00:26<00:00, 11.15i
Accuracy: 0.950
Invalid: 0.000
Latency: 27.047 s
Output throughput: 1378.700 token/s

Qwen3-8B(Prefill DP 2 TP 1 PP2, Decode TP 4):
python sglang/benchmark/gsm8k/bench_sglang.py --port 8000 --parallel 300 --num-questions 300
100%|█| 300/300 [00:20<00:00, 14.44i
Accuracy: 0.950
Invalid: 0.000
Latency: 21.158 s
Output throughput: 1755.561 token/s

DeepSeek(Prefill TP 8 PP2, Decode TP 16): verified by @ssssnow 
python benchmark/gsm8k/bench_sglang.py --port 9091 --parallel 300 --num-questions 300
Accuracy: 0.947
Invalid: 0.000
Latency: 32.184 s
Output throughput: 927.104 token/s
```

## Checklist

[...]

PR #9262, opened 2025-08-17
Fix GLM 4.5 failure by supporting PP for GLM4.5 MOE

hebiao064:
## Motivation

PP not fully supported yet, just fixed some bug which broke TP.

- [This PR](https://github.com/sgl-project/sglang/pull/8846) modified Deepseek forward logic about PP
- GLM4.5 is a subclass (inherited model) from Deepseek MOE
- GLM4.5 doesn't support PP yet, hence forward will fail

```bash
100%|███████████████████████████████████████████████| 250/250 [00:23<00:00, 10.59it/s]
Accuracy: 0.928
Invalid: 0.000
Latency: 24.050 s
```

FYI, this is not related to the gsm score lower than threshold issue in nightly test. I just need to fix this first then I can investigate whether the threshold should be lower down or whether we should fix the model logic.

## Checklist

### DIFF
--- a/python/sglang/srt/models/glm4_moe.py
+++ b/python/sglang/srt/models/glm4_moe.py
@@ -24,6 +24,7 @@ from transformers import PretrainedConfig
 
 from sglang.srt.distributed import (
     get_moe_expert_parallel_world_size,
+    get_pp_group,
     get_tensor_model_parallel_rank,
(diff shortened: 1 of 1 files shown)


ShangmingCai:
Thanks for the report and the fix.

PR #9268, opened 2025-08-17
`from python.sglang.srt` -> `from sglang.srt`

netanel-haber:
This is a slight fix to `nemotron_nas` imports so import after sglang packaging will be guaranteed to work, and also consistency.


netanel-haber:
Hey, @zhyncs, someone on my team noticed a possibly broken import that would probably only break when sglang is packaged.

Issue #8715, opened 2025-08-02
[Roadmap] MoE Refactor


ch-wan:
# Background
We aim to optimize the code structure of MoE modules in SGLang to enhance extensibility. Currently, there are three main MoE modules: `FusedMoE`, `EPMoE`, and `DeepEPMoE` (along with recent additions like `FlashInferEPMoE` and `FlashInferFusedMoE` as of this document's preparation). Their implementations suffer from several issues:

- **Inconsistent logic flow.** Computation logics vary across modules. For instance, `FusedMoE` computes `select_experts` within its forward function, while `DeepEPMoE` handles it externally. Similarly, some forward functions manage `routed_scaling_factor` internally, but others do not.
- **Poor extensibility.** We plan to support multiple all-to-all communication backends under EP (e.g., DeepEP, PPLX, etc) and grouped-GEMM backends (e.g., Triton, DeepGEMM, Triton Kernels, FlashInfer MoE, etc). The current design requires a dedicated forward function for each backend combination, leading to redundancy.
- **Lengthy and duplicated code.** Common variable combinations are repeated across functions. For example, over 10 MoE quantization methods each handle about 15 nearly identical inputs in their `apply` functions. DeepEP dispatch outputs (8 in total) are duplicated in multiple model files.

# Design
To streamline the code structure, we will deprecate all MoE modules except `FusedMoE` and gradually merge existing functionalities into it. Below is an overview of the target code structure:

```
[input_hidden_states]
          |
          v
     TopK.forward -> `select_experts` / `triton_kernels.routing` / bypass
          |
          V
     [TopKOutput]
          |
          v
   FusedMoE.forward -> Dispatcher.dispatch -> DeepEP / PPLX / bypass
          |                     |
          |                     v
          |              [DispatchOutput]
          |                     |
          |                     v
          |             quant_mothod.apply -> MoeRunner.forward -
          |                     |                                |
          |                     |                                v
          |                     |            pre-permute + grouped_gemm + post-permute
          |                     |                                |
          |                     |--------------------------------
          |                     v
          |               [CombineInput]
          |                     |
          |                     v
          |            Dispatcher.combine -> DeepEP / PPLX / bypass
          |                     |
          |---------------------                  
          v
[final_hidden_states]
```

In addition to existing arguments like `--quantization`, we will introduce `--moe-a2a-backend` and `--moe-grouped-gemm-backend` to allow users to select the optimal dispatching and grouped-GEMM backends for their use cases.

[...]

PR #9269, opened 2025-08-17
[7/N] MoE Refactor: the implementation of new framework

ch-wan:
## Motivation

This PR implements a new MoE framework introduced in #8715 to streamline the integration of new all-to-all backends or grouped-gemm backends. 

## Modifications

One MoE forward function can be decoupled into three parts: dispatch -> pre-permute -> core runner -> post-permute -> combine. With this PR, a developer can implement dispatch/combine in the FusedMoE module and define their customized moe runner. The detailed workflow is illustrated in #8715.

We offer two modes for implementing layer permutation:

- Mode 1: `register_fused_func` for static fused_op for each pair of a2a x runner backend. This mode is compatible with torch.compile.
- Mode 2: `register_pre_permute` and `register_post_permute` for static runner + dynamic permutation functions. This mode ensures better extensibility.

## Checklist

### DIFF
--- a/python/sglang/srt/layers/quantization/w8a8_fp8.py
+++ b/python/sglang/srt/layers/quantization/w8a8_fp8.py
@@ -5,6 +5,7 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional
 import torch
 from torch.nn.parameter import Parameter
 
+from sglang.srt.layers.moe.moe_runner.triton import TritonMoeQuantInfo
 from sglang.srt.layers.parameter import ChannelQuantScaleParameter, ModelWeightParameter
 from sglang.srt.layers.quantization.base_config import (
     FusedMoEMethodBase,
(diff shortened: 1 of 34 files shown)


python/sglang/srt/layers/quantization/w8a8_fp8.py
```diff
@@ -266,25 +280,26 @@ def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
             layer.w2_weight_scale.data, requires_grad=False
         )
 
+    def create_moe_runner(
+        self, layer: torch.nn.Module, moe_runner_config: MoeRunnerConfig
+    ):
+        self.moe_runner_config = moe_runner_config
+        self.runner = MoeRunner(MoeRunnerBackend.TRITON, moe_runner_config)
```
BowenBao, line 287: It seems `MoeRunner` and `MoeRunnerBackend` and used without import.

ch-wan: Thank you for sharing it. It appears that the code path is not covered in our CI test. We will fix it.

PR #5624, opened 2025-04-22
Fix FA3 DeepSeek prefill performance regression

Alcanderian:
## Motivation

Fix FA3 DeepSeek prefill performance regression

Significantly boost 20% at context len 1024, the longer context, the more benifitions

cmd: `python3 -m sglang.bench_one_batch --model lmsys/sglang-ci-dsv3-test --batch-size 8 --input-len 1024 --output-len 128 --trust-remote-code`

- before
```
Prefill. latency: 0.32884 s, throughput:  24911.50 token/s
Decode. Batch size: 8, latency: 0.01430 s, throughput:    559.60 token/s
Decode. Batch size: 8, latency: 0.01438 s, throughput:    556.49 token/s
Decode. Batch size: 8, latency: 0.01415 s, throughput:    565.55 token/s
Decode. Batch size: 8, latency: 0.01433 s, throughput:    558.10 token/s
Decode. Batch size: 8, latency: 0.01465 s, throughput:    546.01 token/s
Decode.  median latency: 0.01490 s, median throughput:    536.79 token/s
Total. latency:  2.217 s, throughput:   4156.30 token/s
```

- after

```
Benchmark ...
Prefill. latency: 0.27612 s, throughput:  29668.45 token/s
Decode. Batch size: 8, latency: 0.01418 s, throughput:    564.37 token/s
Decode. Batch size: 8, latency: 0.01412 s, throughput:    566.39 token/s
Decode. Batch size: 8, latency: 0.01405 s, throughput:    569.45 token/s
Decode. Batch size: 8, latency: 0.01408 s, throughput:    568.10 token/s
Decode. Batch size: 8, latency: 0.01437 s, throughput:    556.84 token/s
Decode.  median latency: 0.01411 s, median throughput:    566.95 token/s
Total. latency:  2.071 s, throughput:   4449.77 token/s
```

===================================================

MORE advice: we should consider both context len and prefix len to select MLA or MHA CHUNKED KV. Since they both affect the FLOPS of the core attention

ATTN FLOPS = bh * sq * sk * (dq + dv) * 2
MHA:  128b * context * (context + prefix) * (192 + 128) * 2
MLA: 128b * context * (context + prefix) * (576 + 512) * 2

Extract KV FLOPS: b * prefix * kvlora * (h * dqn) * 2 = bh * prefix * kvlora * dqn * 2 = 128b * prefix * 512 * 128 * 2

Goal: MHA + Extact KV < MLA
[...]

PR #9295, opened 2025-08-18
feat: Prefix optimization using MHA

Starick-L:
## Motivation
Hi，Team！

This PR for optimizing MLA performance in prefix scenarios by dynamically switching the prefill phase from MLA to MHA for long sequences.

The choice between MHA and MQA mainly depends on the attention computation profile:

- **Long sequences :**
   - MHA is more efficient since the attention cost of MLA is ~3.4× that of MHA.
   - As prefix_len increases, MQA’s computation grows rapidly, making MHA increasingly favorable.
   
- **Short sequences :**
  - MQA generally has lower memory traffic.
  - As prefix_len increases, MHA’s memory usage grows faster, making MQA more advantageous.
  
- The threshold depends on the GPU’s compute-to-memory bandwidth ratio — higher ratios shift the threshold upward.

Based on benchmarks on H20 GPUs:

- When average_len >= 512: MHA outperforms MQA.
- When average_len < threshold:
  - Longer prefix_len favors MLA.
- When average_len > threshold:
  - Longer prefix_len favors MQA.

**In prefix scenarios:**
- Use MHA when average_len >= 512.
- Use MLA when average_len < 512

## Benchmarking and Profiling

Command:
`python3 -m sglang.bench_one_batch --model-path xxxx/DeepSeek-V3-0324 --batch 1 --input-len 1024 --output-len 1 --tp 8 --dp 8`

with this PR:
```
Prefill. latency: 3.59127 s, throughput:    285.14 token/s
Total. latency:  3.591 s, throughput:    285.41 token/s
Benchmark ...
Prefill. latency: 0.23663 s, throughput:   4327.44 token/s
Total. latency:  0.237 s, throughput:   4331.66 token/s
```
without this PR:
```
Prefill. latency: 3.60378 s, throughput:    284.15 token/s
Total. latency:  3.604 s, throughput:    284.42 token/s
Benchmark ...
Prefill. latency: 0.27267 s, throughput:   3755.46 token/s
Total. latency:  0.273 s, throughput:   3759.12 token/s
```

**Additional tests (long prefix, no perf regression)**
```
bs = 1 seq_len = 1024 prefix_len = 1024
with this PR:  
Prefill. latency: 0.99795 s, throughput:   2028.17 token/s
without this PR:
Prefill. latency: 1.79457 s, throughput:   1127.84 token/s

bs = 1 seq_len = 1024 prefix_len = 2048
with this PR:  
Prefill. latency: 0.88650 s, throughput:   3465.32 token/s
without this PR:
Prefill. latency: 2.12460 s, throughput:   1445.92 token/s
```

## Checklist

### DIFF
--- a/python/sglang/srt/models/deepseek_v2.py
+++ b/python/sglang/srt/models/deepseek_v2.py
@@ -923,6 +923,11 @@ class DeepseekV2AttentionMLA(nn.Module):
             "SGL_CHUNKED_PREFIX_CACHE_THRESHOLD", 8192
         )
 
+        # TODO: Design a finer way to determine the threshold for prefix MHA
+        self.mha_prefix_threshold = get_int_env_var(
+            "MHA_PREFIX_THRESHOLD", 1 * 512 * 512
+        )
+
         # If we have self.fused_qkv_a_proj_with_mqa and we're running on CPU, we will choose the torch.ops.sgl_kernel.qkv_proj_with_rope_fused_weight kernel
         # which requires self.w_kc and self.w_vc to be packed.
         # If not, we will use torch.bmm and weight shouldn't be packed in this case
@@ -1019,6 +1024,19 @@ class DeepseekV2AttentionMLA(nn.Module):
             else:
                 return _dispatch_mla_subtype()
         elif attention_backend == "fa3":
+            # When the sequence length exceeds the threshold and the GPU is in compute-bound state, 
+            # MHA performs better than MQA. Attention complexity: O(bs·seq_len^2)
+            # Where: 
+            # - average_seq_len approximately represents sequence length (average_seq_len ≈ sum_seq_len / batch_size)
+            # - Current computational workload estimation: current_workload ≈ batch_size * average_seq_len^2
+            current_workload = 0
+            if(
+                forward_batch.forward_mode.is_extend() 
+                and forward_batch.extend_seq_lens_cpu is not None
+            ):
+                sum_seq_lens = sum(forward_batch.extend_seq_lens_cpu) 
+                assert forward_batch.batch_size != 0, "Invalid batch size: zero"
+                current_workload = sum_seq_lens // forward_batch.batch_size * sum_seq_lens
             # Flash Attention: Use MHA with chunked KV cache when prefilling on long sequences.
             if forward_batch.extend_prefix_lens_cpu is not None:
                 sum_extend_prefix_lens = sum(forward_batch.extend_prefix_lens_cpu)
@@ -1030,6 +1048,7 @@ class DeepseekV2AttentionMLA(nn.Module):
                 and (
                     sum_extend_prefix_lens >= self.chunked_prefix_cache_threshold
                     or sum_extend_prefix_lens == 0
+                    or current_workload >= self.mha_prefix_threshold
                 )
             ):
                 return AttnForwardMethod.MHA_CHUNKED_KV
--- a/python/sglang/srt/models/deepseek_v2.py
+++ b/python/sglang/srt/models/deepseek_v2.py
@@ -1036,7 +1036,7 @@ class DeepseekV2AttentionMLA(nn.Module):
             ):
                 sum_seq_lens = sum(forward_batch.extend_seq_lens_cpu) 
                 assert forward_batch.batch_size != 0, "Invalid batch size: zero"
-                current_workload = sum_seq_lens // forward_batch.batch_size * sum_seq_lens
+                current_workload = sum_seq_lens / forward_batch.batch_size * sum_seq_lens
             # Flash Attention: Use MHA with chunked KV cache when prefilling on long sequences.
             if forward_batch.extend_prefix_lens_cpu is not None:
                 sum_extend_prefix_lens = sum(forward_batch.extend_prefix_lens_cpu)


xu-yfei:
There are other factors to consider when we decide whether to choose MHA, such as the prefix length. In the case of multiple batches, the distribution of different extension lengths in each batch can also have a significant impact on performance.

This is our performance data for fa3 on H20 under different configurations.
 **batch size=1**: 
```
prefix 0, extend 1280, mla: 0.50, mha: 0.48
prefix 1, extend 2048, mla: 0.91, mha: 0.79
prefix 256, extend 1792, mla: 0.88, mha: 0.79
prefix 512, extend 1536, mla: 0.83, mha: 0.82
prefix 768, extend 1536, mla: 0.93, mha: 0.86
prefix 1024, extend 1280, mla: 0.93, mha: 0.90
prefix 1280, extend 1280, mla: 1.03, mha: 0.94
prefix 1536, extend 1024, mla: 0.92, mha: 0.88
prefix 1792, extend 1024, mla: 1.00, mha: 0.90
prefix 2048, extend 1024, mla: 1.08, mha: 0.92
prefix 2304, extend 1024, mla: 1.17, mha: 0.93
prefix 2560, extend 1024, mla: 1.25, mha: 0.96
prefix 2816, extend 768, mla: 1.02, mha: 1.00
prefix 3072, extend 768, mla: 1.08, mha: 1.03
```

 In scenarios where MLA has the same sum of prefix length and extend length, the performance varies significantly with different distributions of extend length.
**batch size=2**.
```
prefix 0, extend 1280, mla: 0.49, min: 0.49, max: 0.50, mha: 0.47, min: 0.46, max: 0.48
prefix 2, extend 2048, mla: 0.90, min: 0.90, max: 0.91, mha: 0.77, min: 0.75, max: 0.79
prefix 256, extend 1792, mla: 0.81, min: 0.75, max: 0.95, mha: 0.80, min: 0.75, max: 1.23
prefix 512, extend 1792, mla: 0.88, min: 0.76, max: 1.00, mha: 0.80, min: 0.76, max: 0.83
prefix 768, extend 1792, mla: 0.94, min: 0.76, max: 1.13, mha: 0.82, min: 0.77, max: 0.87
prefix 1024, extend 1536, mla: 0.84, min: 0.62, max: 1.04, mha: 0.83, min: 0.76, max: 0.89
prefix 1280, extend 1536, mla: 0.84, min: 0.62, max: 1.12, mha: 0.83, min: 0.78, max: 0.92
prefix 1536, extend 1536, mla: 0.93, min: 0.63, max: 1.23, mha: 0.86, min: 0.79, max: 0.96
prefix 1792, extend 1536, mla: 0.99, min: 0.62, max: 1.34, mha: 0.88, min: 0.80, max: 1.00
prefix 2048, extend 1280, mla: 0.92, min: 0.52, max: 1.34, mha: 0.90, min: 0.80, max: 1.03
prefix 2304, extend 1280, mla: 0.94, min: 0.52, max: 1.43, mha: 0.90, min: 0.81, max: 1.06
```

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -1019,6 +1024,19 @@ def _dispatch_mla_subtype():
...
+            # - average_seq_len approximately represents sequence length (average_seq_len ≈ sum_seq_len / batch_size)
+            # - Current computational workload estimation: current_workload ≈ batch_size * average_seq_len^2
+            current_workload = 0
+            if(
+                forward_batch.forward_mode.is_extend() 
+                and forward_batch.extend_seq_lens_cpu is not None
+            ):
+                sum_seq_lens = sum(forward_batch.extend_seq_lens_cpu) 
+                assert forward_batch.batch_size != 0, "Invalid batch size: zero"
+                current_workload = sum_seq_lens // forward_batch.batch_size * sum_seq_lens
```
Fridge003: Maybe here we should change `//` to `/`? Since mean length can be a floating point number.

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -923,6 +923,11 @@ def __init__(
             "SGL_CHUNKED_PREFIX_CACHE_THRESHOLD", 8192
         )
 
+        # TODO: Design a finer way to determine the threshold for prefix MHA
+        self.mha_prefix_threshold = get_int_env_var(
+            "MHA_PREFIX_THRESHOLD", 1 * 512 * 512
```
Fridge003, line 928: According to my understand, 1*512*512 is related to the hardware for use (in this case H20).
So maybe we should assign different default values for different hardwares? Then it will be better to convert `self.mha_prefix_threshold` into a method.

Fridge003:
@xu-yfei  Thank you for your examples~
Actually in #5624 me and @Alcanderian tried to model this problem. According to our calculation the curve of prefix length vs extend length is more like a conic section. It's complex so currently we only apply some simple heuristics

python/sglang/srt/models/deepseek_v2.py
Starick-L: Thanks for your suggestion,`/` would be more appropriate.

python/sglang/srt/models/deepseek_v2.py
Starick-L: Yes，this variable should account for hardware characteristics, like the GPU’s compute/memory bandwidth ratio.However, I currently do not have the resources to test it across different GPU models.

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -1019,6 +1024,19 @@ def _dispatch_mla_subtype():
...
+            # MHA performs better than MQA. Attention complexity: O(bs·seq_len^2)
+            # Where: 
+            # - average_seq_len approximately represents sequence length (average_seq_len ≈ sum_seq_len / batch_size)
+            # - Current computational workload estimation: current_workload ≈ batch_size * average_seq_len^2
+            current_workload = 0
+            if(
+                forward_batch.forward_mode.is_extend() 
+                and forward_batch.extend_seq_lens_cpu is not None
+            ):
+                sum_seq_lens = sum(forward_batch.extend_seq_lens_cpu)
```
Alcanderian, line 1037: I prefer square sum here extend_seq_lens_cpu.pow(2).sum().item()

hnyls2002:
Thanks @Starick-L! Prefix MLA->MHA auto-switching is now native (dsa_backend.py), so I'm closing this as
  superseded. Please reopen if your path isn't covered.

PR #9303, opened 2025-08-18
minor: determine mm attn backend based on platforms

mickqian:
## Motivation
1. fa3 is unavailable on Blackwell

## Checklist


mickqian:
cc @JustinTong0323

python/sglang/srt/layers/attention/vision.py
```diff
@@ -461,6 +456,28 @@ def __init__(
             prefix=add_prefix("proj", prefix),
         )
 
+    def _determine_attention_backend(self, passed_backend: Optional[str]) -> str:
```
zhyncs, line 464: shall we use `fa3` by default on sm90

PR #9312, opened 2025-08-18
[bugfix] fix 3 online stability in pd_router - filter 4xx response in failure record/support SGL_ROUTER_SAFE_DISPATCH/fix error resp when streaming

Bruce-x-1997:
## Motivation

I met stability issue at online case when using sgl-router, and when we use minilb, it works well. After reading the code of minilb and sgl-router, I fix 3 bugs.

## Modifications

1. I just provide a SGL_ROUTER_SAFE_DISPATCH env var, and when it's true, just run safe branch, which use tokio.join to  wait prefill & decode response(reuse the return_logprobs code), the same as minilb.
2. in pd_router, 4xx response(means client error) would be recorded as failure and trigger circuit breaker to be open,  and the decode pod would be recorded as unavaliable.Now I just filter client_error.
3. when a streaming request meets error , it should return a streaming response(using create_streaming_response), not a non-stream response

## Checklist


Bruce-x-1997:
@slin1237 please help check it) thanks

PR #9336, opened 2025-08-19
Fix incorrect logic in chat template handling.

lifuhuang:
This PR addresses two minor issues:

1. Incorrect message about chat templates:
image

2. _force_reasoning not set when hf_template is chosen.


python/sglang/srt/managers/template_manager.py
```diff
@@ -129,10 +129,12 @@ def load_chat_template(
...
-            logger.info("No chat template found, defaulting to 'string' content format")
+                else:
+                    # Default to string content format if no template was found
+                    self._jinja_template_content_format = "string"
+                    logger.info(
+                        "No chat template found, defaulting to 'string' content format"
+                    )
 
         # Detect reasoning pattern from chat template
         if tokenizer_manager.tokenizer:
```
lifuhuang, line 140: Hi @JustinTong0323 , I am not familiar with the logic below. Can you help confirm that it is expected that we should set `_force_reasoning` in all cases unless we can resolve a hf_template? Thanks!

JustinTong0323: If we have a hf_template, we should detect if it shown as force_reasoning pattern.

lifuhuang: Discussed offline.

JustinTong0323 (approved):
Thanks Lifu!

lifuhuang:
@zhyncs , can I get a merge for this minor fix? Thanks! (co-author @JustinTong0323 )

PR #9343, opened 2025-08-19
fix: fix max_new_tokens uninitialized error

### LINKED ISSUE #9328: [Bug] online serving crash because of not supported operation
### Checklist

### Describe the bug

online serving crash because of not supported operation

Image

### Reproduction

Qwen3 coder 480B fp8 with tp8, dp-attention dp8

Image

### Environment

sglang 0.5.0.rc2 docker image

mickqian:
fix #9328
## Motivation

If an error is found when building a req in `handle_generate_request`, the `max_new_tokens` could get uninitialized from early-return behavior.

## Modifications

Make sure the field of req is initialized before returning.

## Checklist

### DIFF
--- a/python/sglang/srt/managers/scheduler.py
+++ b/python/sglang/srt/managers/scheduler.py
@@ -1155,6 +1155,16 @@ class Scheduler(
                 else:
                     self.send_to_tokenizer.send_pyobj(output)
 
+    def update_req_max_new_tokens(self, req):
+        req.sampling_params.max_new_tokens = min(
+            (
+                req.sampling_params.max_new_tokens
+                if req.sampling_params.max_new_tokens is not None
+                else 1 << 30
+            ),
+            self.max_req_len - len(req.origin_input_ids) - 1,
+            )
+
     def handle_generate_request(
         self,
         recv_req: TokenizedGenerateReqInput,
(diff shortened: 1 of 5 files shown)


mickqian:
#9328

mickqian:
ref #9328

python/sglang/srt/sampling/sampling_params.py
```diff
@@ -54,7 +54,7 @@ def __init__(
         stream_interval: Optional[int] = None,
         logit_bias: Optional[Dict[str, float]] = None,
     ) -> None:
-        self.max_new_tokens = max_new_tokens
+        self.max_new_tokens = max_new_tokens or 128
```
slin1237: why hardcode to 128 in sampler

slin1237: should this be default to model ctx length? if not specified?

mickqian: previously 128 is the default parameter for `max_new_tokens`. I've now updated to `context_len`.

slin1237 (approved):

python/sglang/srt/sampling/sampling_params.py
hnyls2002: This should be hardcored to `128` or even smaller numbers

python/sglang/srt/managers/scheduler.py
```diff
@@ -1155,6 +1155,16 @@ def process_input_requests(self, recv_reqs: List):
                 else:
                     self.send_to_tokenizer.send_pyobj(output)
 
+    def update_req_max_new_tokens(self, req):
```
hnyls2002:
```suggestion
    def init_req_max_new_tokens(self, req):
```

hnyls2002 (approved):

Issue #8814, opened 2025-08-05
[Bug] zai-org/GLM-4.5-Air starts emitting only `!` (token id 0) under both single- and multi-concurrency


Atream:
### Checklist

- [ ] 1. I have searched related issues but cannot get the expected help.
- [ ] 2. The bug has not been fixed in the latest version.
- [ ] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [ ] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [ ] 5. Please use English, otherwise it will be closed.

### Describe the bug

When testing the model GLM-4.5-Air-FP4 on GSM8K, we observed that many requests returned outputs consisting solely of “!”, whose token index is 0. This happens under both high concurrency and single concurrency.

### Server launch command

```bash
python -m sglang.launch_server \
  --model GLM-4.5-Air-FP4 \
  --tp-size 2 \
  --trust-remote-code \
  --served-model-name GLM-4.5-Air-FP4 \
  --quantization modelopt_fp4
```

### Benchmark command

```bash
python benchmark/gsm8k/bench_sglang.py \
  --num-questions 2000 \
  --parallel 2000
```

```bash
python benchmark/gsm8k/bench_sglang.py \
  --num-questions 2000 \
  --parallel 1
```

### Environment

On commit a4b0d5c9e5cb2b36eacdc30bc9259a213cd17a16
Quantize zai-org/GLM-4.5-Air using TensorRT-Model-Optimizer
Change code:
```
diff --git a/python/sglang/srt/layers/quantization/modelopt_quant.py b/python/sglang/srt/layers/quantization/modelopt_quant.py
index 36ffd1275..0b90a6ba1 100755
--- a/python/sglang/srt/layers/quantization/modelopt_quant.py
+++ b/python/sglang/srt/layers/quantization/modelopt_quant.py
@@ -667,9 +667,9 @@ class ModelOptFp4LinearMethod(LinearMethodBase):
         padded_scales = padded_scales.permute((0, 1, 4, 3, 2, 5))
         padded_scales = padded_scales.contiguous().cuda()
         padded_scales = (
-            padded_scales.reshape(M, K)
+            padded_scales.reshape(M_padded, K_padded)
[...]

PR #9346, opened 2025-08-19
Fix FP4 inference corruption issue in glm4.5-air model

Azure-Tang:
# Background
When using FP4 quantization with the glm4.5-air model, the inference outputs may contain corrupted/garbled text. #8814 
# Root Cause
The quantization operator in sglang used `torch.empty` to create tensors;
* During padding, `empty` could introduce random values;
* These random values were later converted into NaN when mapped to nvfp4 format;
* In glm4.5’s first MLP layer, the intermediate size is 10944, which under tp=2 triggers padding. This may caused NaN values to be introduced into the matrix and propagated into subsequent GEMM computations;
* Therefore, the issue is always triggered in glm4.5’s first layer. In other layers or other models, the same issue may also occur whenever irregular shapes cause padding during quantization.
# Fix
* Replace torch.empty with explicitly initialized tensors (e.g., torch.zeros);
* Ensure padding regions are zeroed out, preventing random values from being converted to NaN in nvfp4 format and propagated downstream.
# Validation
* After the fix, glm4.5-air FP4 inference no longer produces corrupted output;
* Multiple benchmarks ran successfully with correct results.

image

By @Azure-Tang and @qiuqiu10


sgl-kernel/python/sgl_kernel/gemm.py
```diff
@@ -73,7 +73,7 @@ def bmm_fp8(
     out: Optional[torch.Tensor] = None,
 ) -> torch.Tensor:
     if out is None:
-        out = torch.empty(
+        out = torch.zeros(
```
Alcanderian: I am not appreciate that introducing extra zeros kernel in model forward phase. Can we double check whether these zeros is necessary？

Azure-Tang: I see, I have updated the pr and minimized modifications.

Azure-Tang:
Hi, I have double checked pr and minimized modifications. @Alcanderian

Alcanderian (approved):

PR #8328, opened 2025-07-24
[new feat] ascend backend support fia fusion kernel

ZhengdQin:
## Motivation
In this MR, we implemented the NPU fusion kernels npu_fused_infer_attention_score in the Qwen2.5-7b, deepseek-v2-lite and deepseek-v3 models, this fusion kernel is  suitable for the graph mode. One needs to export **ASCEND_USE_FIA=ture** to activate this  fusion kernel.

## Modifications
Ascend Backend: support npu_fused_infer_attention_score kernel
Add unittest: test_ascend_tp_fia_bf16.py and test_ascend_mla_fia_w8a8int8.py

> Exclusive support for paged attention, currently ONLY support page size 128.
> You can activate this feature by setting export ASCEND_USE_FIA=ture and --attention-backend ascend.

## Memory Management Advancement
We modify the AscendMLAPagedTokenToKVPool class, split the kvbuffer to the k_buffer and v_buffer, in order to remove the split op in MLA attention.

## Testing Framework
Unit tests for the Ascend attention backend have been added and can be found at: /test/srt/ascend/test_ascend_tp_fia_bf16.py and /test/srt/ascend/test_ascend_mla_fia_w8a8int8.py

## Checklist

- [x] Format your code according to the [Code Formatting with Pre-Commit](https://docs.sglang.ai/references/contribution_guide.html#code-formatting-with-pre-commit).
- [x] Add unit tests as outlined in the [Running Unit Tests](https://docs.sglang.ai/references/contribution_guide.html#running-unit-tests-adding-to-ci).
- [ ] Update documentation / docstrings / example tutorials as needed, according to [Writing Documentation](https://docs.sglang.ai/references/contribution_guide.html#writing-documentation-running-docs-ci).
- [x] Provide throughput / latency benchmark results and accuracy evaluation results as needed, according to [Benchmark and Profiling](https://docs.sglang.ai/references/benchmark_and_profiling.html) and [Accuracy Results](https://docs.sglang.ai/references/accuracy_evaluation.html).
- [ ] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co-author when merging the PR.
- [ ] Please feel free to join our Slack channel at https://slack.sglang.ai to discuss your PR.

## Accuracy and performance result
`python -m unittest test_npu_mla_backend.TestNpuMlaBackend.test_gsm8k`
1314ci2

## Pre-commit check
precom

PR #9355, opened 2025-08-19
[Feature] Support NPUGraph for DeepSeek on Ascend NPU

chenxu140:
## Motivation

this pr improves deepseek model (mla to be exact) performance with npugraph support
check initial npugraph support on mha model here #9399 and #8030

## Modifications

- support concurrent kvcache transfer following mooncake transfer engine
- add mla graph support in ascend attention backend
- some performance and accuracy improvements

## Accuracy Tests

87DF5544-5E77-4E8C-F895-71FDFDFEEEEB

## Benchmarking and Profiling

Reusing performance result [here](https://github.com/sgl-project/sglang/pull/8328#issuecomment-3222527169)
FIA result: sglang/test/srt/ascend/test_ascend_mla_fia_w8a8int8.py (tp == 2)
image

After this pr, with cuda graph enabled
```bash
python3 -m sglang.launch_server --model DeepSeek-V2-Lite-W8A8 --tp 2 --trust-remote-code --quantization w8a8_int8  --device npu --attention-backend ascend --cuda-graph-bs 1 2 4 6 8 16 24 32 40 48 56 64 72 80 88 96 104 112 120 128 136 --disable-radix-cache --chunked-prefill-size 32768 --mem-fraction-static 0.8
```
D0A19C18-A5A9-4ED0-B5B6-8B5F1BCB3246

## Checklist


python/sglang/srt/managers/data_parallel_controller.py
```diff
@@ -281,7 +284,21 @@ def round_robin_scheduler(self, req: Req):
                 logger.debug(f"Direct routing to DP rank {req.data_parallel_rank}")
                 self.workers[req.data_parallel_rank].send_pyobj(req)
             else:
-                self.workers[req.bootstrap_room % len(self.workers)].send_pyobj(req)
+                DP_LOAD_BALANCE = os.getenv("DP_LOAD_BALANCE", "0") == "1"
```
iforgetmyname: Revert this block

python/sglang/srt/managers/data_parallel_controller.py
```diff
@@ -101,6 +102,8 @@ def __init__(
         # Launch data parallel workers
         self.scheduler_procs = []
         self.workers = [None] * server_args.dp_size
+        self.work_req_num = [0] * server_args.dp_size
```
iforgetmyname: why we need to do dp load balance here

python/sglang/srt/mem_cache/memory_pool.py
```diff
@@ -894,7 +894,7 @@ def __init__(
         start_layer: Optional[int] = None,
         end_layer: Optional[int] = None,
     ):
-        super(MLATokenToKVPool, self).__init__(
+        super(MHATokenToKVPool, self).__init__(
```
iforgetmyname: use super().__init__ instead as this is a direct inheritance

python/sglang/srt/disaggregation/ascend/conn.py
```diff
@@ -29,6 +35,77 @@ def register_buffer_to_engine(self):
...
+        layers_params = [
+            (
+                self.kv_args.kv_data_ptrs[layer_id],
+                dst_kv_ptrs[layer_id],
+                self.kv_args.kv_item_lens[layer_id],
+            )
+            for layer_id in range(num_layers)
+        ]
+
+        assert layers_params is not None
```
iforgetmyname: remove this line

python/sglang/srt/managers/data_parallel_controller.py
chenxu140: done

python/sglang/srt/managers/data_parallel_controller.py
chenxu140: done

python/sglang/srt/disaggregation/ascend/conn.py
chenxu140: done

python/sglang/srt/mem_cache/memory_pool.py
chenxu140: done

iforgetmyname:
image

this is the comment i deleted here #8328, and it should be fixed in this pr

python/sglang/srt/disaggregation/ascend/conn.py
```diff
@@ -29,6 +35,75 @@ def register_buffer_to_engine(self):
             self.kv_args.aux_data_ptrs, self.kv_args.aux_data_lens
         )
 
+    def send_kvcache(
```
ErvinXie, line 38: Most of the function is identical to mooncacke/conn.py, could we find a way to avoid this?

python/sglang/srt/layers/attention/ascend_backend.py
```diff
@@ -54,13 +96,12 @@ def __init__(self, model_runner: ModelRunner):
         super().__init__()
         self.forward_metadata = None
         self.device = model_runner.device
-        self.gen_attention_mask(128, model_runner.dtype)
+        self.attn_mask_builder = AttentionMaskBuilder(8192, model_runner.dtype)
```
ErvinXie: can we use compressed mask as described in  https://www.hiascend.com/document/detail/zh/canncommercial/81RC1/apiref/ascendtbapi/ascendtb_01_0259.html

python/sglang/srt/model_executor/forward_batch_info.py
```diff
@@ -297,6 +297,7 @@ class ForwardBatch:
     # For padding
     padded_static_len: int = -1  # -1 if not padded
     num_token_non_padded: Optional[torch.Tensor] = None  # scalar tensor
+    num_token_non_padded_cpu: int = None
```
Alcanderian: Unused variable

chenxu140: fixed

Alcanderian:
CI passed
image

Alcanderian (approved):

Alisehen:
We profiled the decode time of a single layer, and the performance improvement is significant.
截屏2025-08-28 16 01 22

ErvinXie (approved):
After thorough testing, we confirmed that enabling NPUGraph delivers a 6x improvement in decode performance. In addition, it shows a clear performance advantage compared to vllm ascend. Based on these results, we believe this PR is ready to be merged.

python/sglang/srt/layers/attention/ascend_backend.py
iforgetmyname: atb will stop maintaining by the end of this year and also it does't work well with aclnn ops

python/sglang/srt/disaggregation/ascend/conn.py
iforgetmyname: directly inherit from mooncake and add logics to parent method could also be a pain in the butt, we will try to find if there's a way out

python/sglang/srt/layers/attention/ascend_backend.py
```diff
@@ -158,7 +158,7 @@ def init_forward_metadata_replay_cuda_graph(
         self.graph_mode = True
 
     def get_cuda_graph_seq_len_fill_value(self):
-        return 1
+        return 0
```
wangtiance, line 161: Hello, anyone care to explain why the value is changed to 0 here? Thanks.

PR #9358, opened 2025-08-19
Use NCCL symmetric memory for DP (includes allgather, fp4 allgatherv, and reducescatter)

nvcastet:
## Motivation

7.4% e2e perf gain
(Best perf with NCCL 2.28 that just released)
After this PR:

```
python3 -m sglang.launch_server --model-path nvidia/DeepSeek-R1-0528-FP4 --trust-remote-code --quantization modelopt_fp4 --tp 8 --enable-flashinfer-cutlass-moe --enable-ep-moe --ep-size 8 --dp 8 --enable-dp-attention --chunked-prefill-size 16384 --mem-fraction-static 0.85 --max-running-requests 4096 --stream-interval 5 --enable-dp-lm-head --attention-backend trtllm_mla --cuda-graph-bs 1 2 4 8 16 32 64 128 256 512 1024 2048 4096 8192 --disable-radix-cache  --enable-symm-mem

python3 -m sglang.bench_serving --backend sglang --dataset-name random --num-prompt 1024 --random-input 1024 --random-output 2048 --random-range-ratio 1 --warmup-request 1024 --max-concurrency 1024
```

```
Max request concurrency:                 1024
Benchmark duration (s):                  97.55
Total generated tokens (retokenized):    2083551
Request throughput (req/s):              10.50
Input token throughput (tok/s):          10749.27
Output token throughput (tok/s):         21498.54
Total token throughput (tok/s):          32247.81
Concurrency:                             1022.43
Mean E2E Latency (ms):                   97398.95
Median E2E Latency (ms):                 97407.47
Mean TTFT (ms):                          8711.54
Median TTFT (ms):                        8592.25
P99 TTFT (ms):                           15501.95
Mean ITL (ms):                           43.33
Median ITL (ms):                         39.74
P95 ITL (ms):                            43.96
P99 ITL (ms):                            46.83
Max ITL (ms):                            2868.54
```

Before:

```
Max request concurrency:                 1024
Benchmark duration (s):                  104.79
Total generated tokens (retokenized):    2089689
Request throughput (req/s):              9.77
Input token throughput (tok/s):          10006.11
Output token throughput (tok/s):         20012.21
Total token throughput (tok/s):          30018.32
Concurrency:                             1022.57
Mean E2E Latency (ms):                   104647.36
Median E2E Latency (ms):                 104659.58
Mean TTFT (ms):                          9131.11
Median TTFT (ms):                        9061.90
P99 TTFT (ms):                           16866.24
Mean ITL (ms):                           46.66
Median ITL (ms):                         42.67
P95 ITL (ms):                            46.43
P99 ITL (ms):                            48.57
Max ITL (ms):                            3148.65
```

## Checklist


python/sglang/srt/layers/quantization/modelopt_quant.py
```diff
@@ -1306,7 +1320,15 @@ def apply(
                 tune_max_num_tokens=next_power_of_2(x.shape[0]),
             )[0]
             if moe_runner_config.routed_scaling_factor is not None:
-                output *= moe_runner_config.routed_scaling_factor
+                with use_symmetric_memory(
```
trevor-m: Can we allocate a symm buffer and pass it as the [output ](https://github.com/flashinfer-ai/flashinfer/blob/54d0d970e431fc969ec7b73c5428982cbf33f51e/flashinfer/fused_moe/core.py#L707) argument in `flashinfer_cutlass_fused_moe()` instead?
This multiply will be removed in https://github.com/sgl-project/sglang/pull/8690
Also `*=` is done in-place so the symmetric buffer will still be used until then

trevor-m (changes requested):

python/sglang/srt/layers/quantization/modelopt_quant.py
nvcastet: Yes thanks for catching that, let me do that.

python/sglang/srt/layers/quantization/modelopt_quant.py
```diff
@@ -1268,22 +1275,35 @@ def apply(
...
+                    sm.tag(x_sf)
                 topk_weights, topk_ids, x, x_sf = get_tp_group().all_gatherv(
                     [topk_weights, topk_ids, x, x_sf], sizes=get_dp_global_num_tokens()
                 )
                 x_sf = nvfp4_block_scale_interleave(x_sf)
 
+            with use_symmetric_memory(
+                get_tp_group(), disabled=not is_max_padding()
+            ) as sm:
+                symm_output = torch.empty_like(x)
```
trevor-m: When x is quantized this will have the wrong dtype and shape (since hidden size will be half). We will want to use output_dtype and maybe store original x_col

trevor-m (changes requested):

python/sglang/srt/layers/quantization/modelopt_quant.py
nvcastet: yes I fixed that in #8934 . Thanks.

kushanam:
This PR includes the changes (and is a replacement ) from https://github.com/sgl-project/sglang/pull/8934.

nvcastet:
> This PR includes the changes (and is a replacement ) from #8934.

This PR builds on #8934 (first commit of the PR).
We can either merge #8934 and rebase this one or just merge this one in one shot.

zhyncs:
@nvcastet @kushanam please fix the conflicts. thanks

nvcastet:
> @nvcastet @kushanam please fix the conflicts. thanks

@zhyncs Done

python/sglang/srt/layers/dp_attention.py
```diff
@@ -72,6 +75,7 @@ class _DpGatheredBufferWrapper:
     _device: torch.device
     _global_dp_buffer_len: int
     _local_dp_buffer_len: int
+    _dp_max_padding: bool = True
```
merrymercy: do not set the default value here, set it in the method?

python/sglang/srt/distributed/device_communicators/pynccl_allocator.py
```diff
@@ -64,8 +77,17 @@ def get_nccl_mem_pool():
...
+        group_coordinator: GroupCoordinator,
+        disabled: bool = False,
+    ):
+        self.disabled = (
+            disabled
+            or not is_symmetric_memory_enabled()
+            or group_coordinator.world_size == 1
+        )
+        if self.disabled:
             self.group_coordinator = None
```
merrymercy: if it is disable, should it return immediately?

python/sglang/srt/layers/linear.py
```diff
@@ -1316,9 +1317,8 @@ def forward(self, input_, skip_all_reduce=False):
         # Only fuse bias add into GEMM for rank 0 (this ensures that
         # bias will not get added more than once in TP>1 case)
         bias_ = None if (self.tp_rank > 0 or self.skip_bias_add) else self.bias
-        with use_symmetric_memory(parallel_state.get_tp_group()) as sm:
+        with use_symmetric_memory(get_tp_group(), disabled=not is_dp_max_padding()):
```
merrymercy: you are writing super-critical runtime code, so please try to reduce python overhead as much as possible.

Can you achieve the following effect?
When symmetric_memory flag is disabled, the overhead should just be < 2 python if/else or function call per layer.

I feel the current way of inserting many `with use_symmetric_memory(get_tp_group(), disabled=not is_dp_max_padding()):` adds python overhead

python/sglang/srt/layers/moe/fused_moe_triton/layer.py
```diff
@@ -1037,6 +1038,8 @@ def forward(self, hidden_states: torch.Tensor, topk_output: TopKOutput):
 
         router_logits = router_logits.to(torch.float32)
 
+        with use_symmetric_memory(get_tp_group(), disabled=not is_dp_max_padding()):
+            symm_output = torch.empty_like(hidden_states)
```
merrymercy, line 1047: when use_symmetric_memory is not enabled, this created a useless tensor.
The principal is: when symmetric_memory is not enabled, the added overhead in the model forward pass should just be < 2 if/else and function calls.

merrymercy (changes requested):

python/sglang/srt/layers/dp_attention.py
nvcastet: The issue is when we are not in "DP" mode, the "setter" is not called, so when the "getter" is called you need to have a default otherwise we are accessing an unset bool.
It is called in non-DP mode to check symmetry. It is always symmetric in non-DP cases and in DP cases it is set on each batch.

python/sglang/srt/distributed/device_communicators/pynccl_allocator.py
nvcastet: It is good practice in a python initializer "__init__" to set  data members to None even if not used.
When symmetric memory is enabled, the context manager is a no-op.

python/sglang/srt/layers/linear.py
nvcastet: > with use_symmetric_memory(get_tp_group(), disabled=not is_dp_max_padding())
I have not noticed any cpu overhead here. 

>> I feel the current way of inserting many with use_symmetric_memory(get_tp_group(), disabled=not is_dp_max_padding()): adds python overhead

The memory pool stack of pytorch has already a lot of conditionals and checks. Symmetric memory context manager are always around memory allocations. So I don't believe adding an empty context manager with an extra conditional add overhead. I would be glad to be proven wrong.

python/sglang/srt/layers/moe/fused_moe_triton/layer.py
nvcastet: There is no overhead here.
When you pass `output=None` or non `output` to `trtllm_fp4_block_scale_moe`, it will call internally `torch.empty_like(hidden_states)`.
So we end up with the exact same behavior.

nvcastet: Allocation is at https://github.com/flashinfer-ai/flashinfer/blob/f7defbf9905b0d897c59afe84763fb6752ca8c22/flashinfer/fused_moe/core.py#L1456-L1462 when output is None

python/sglang/srt/layers/linear.py
nvcastet: I changed it to a **single** conditional using `contextlib.nullcontext` if disabled.

python/sglang/srt/distributed/device_communicators/pynccl_allocator.py
nvcastet: I made the change to return right away.

python/sglang/srt/layers/dp_attention.py
nvcastet: I found a way to remove it by creating a new method "is_allocation_symmetric" that will also check if DP is enabled

fzyzcjy:
hi could you please rebase it thanks

python/sglang/srt/layers/communicator.py
```diff
@@ -513,7 +517,12 @@ def _gather_hidden_states_and_residual(
             use_layer_norm_before_gather = context.attn_tp_size == 1
             if use_layer_norm_before_gather and hidden_states.shape[0] != 0:
                 residual = hidden_states
-                hidden_states = layernorm(hidden_states)
+                with use_symmetric_memory(
+                    get_tp_group(),
+                    disabled=not forward_batch.dp_padding_mode.is_max_len(),
+                ):
```
merrymercy, line 550: Can you try to cache as much variables as possible?

python/sglang/srt/layers/dp_attention.py
```diff
@@ -99,27 +103,33 @@ def set_dp_buffer_len(
...
         cls._global_num_tokens = global_num_tokens
 
     @classmethod
     def get_global_dp_buffer(cls) -> torch.Tensor:
-        return torch.empty(
-            (cls._global_dp_buffer_len, cls._hidden_size),
-            dtype=cls._dtype,
-            device=cls._device,
-        )
+        with use_symmetric_memory(get_tp_group()):
```
merrymercy, line 117: so it always use symmetric memory?

python/sglang/srt/layers/moe/topk.py
```diff
@@ -364,8 +373,11 @@ def forward_npu(
 
     def empty_topk_output(self, device: torch.device) -> TopKOutput:
         topk = self.topk_config.top_k - self.topk_config.num_fused_shared_experts
-        topk_weights = torch.empty((0, topk), dtype=torch.float32, device=device)
-        topk_idx = torch.full((0, topk), -1, dtype=torch.int32, device=device)
+        with use_symmetric_memory(
+            get_tp_group(), disabled=not is_allocation_symmetric()
+        ):
```
merrymercy, line 403: Can we further simplify the code?

Our principal: When symmetric_memory is not enabled, the added overhead in the model forward pass should just be < 2 if/else and function calls.

For example, you can
- and hopefully cache the result of `is_allocation_symmetric`
- move the function `get_tp_group` into `use_symmetric_memory`, when it is disabled, even do not call `get_tp_group`.
- In the best case, even do not create a context when it is disabled.

python/sglang/srt/layers/quantization/modelopt_quant.py
```diff
@@ -1393,27 +1400,42 @@ def apply(
...
+                            0, x_col // 16, dtype=torch.uint8, device=x.device
+                        )
                 topk_weights, topk_ids, x, x_sf = get_tp_group().all_gatherv(
                     [topk_weights, topk_ids, x, x_sf], sizes=get_dp_global_num_tokens()
                 )
                 x_sf = nvfp4_block_scale_interleave(x_sf)
 
+            with use_symmetric_memory(
+                get_tp_group(), disabled=not is_allocation_symmetric()
+            ):
```
merrymercy, line 1600: It seems you call `with use_symmetric_memory(` twice here. There are overheads. Can you try to cache and simplify things so that the overhead is minimal and only some if/else when the symmetric memory is not used?

Ideally, when the symmetric memory is not used, it should just be some simple if/else, even without context creation or redundant function calls (e.g., `get_tp_group`).
In the worst case, we can create empty context to avoid code duplication.

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -728,11 +722,7 @@ def forward_normal(
             # fused in biased_grouped_topk so we can skip here
             final_hidden_states *= self.routed_scaling_factor
         if shared_output is not None:
-            with use_symmetric_memory(parallel_state.get_tp_group()) as sm:
-                final_hidden_states_out = torch.empty_like(final_hidden_states)
-            torch.add(final_hidden_states, shared_output, out=final_hidden_states_out)
-            final_hidden_states = final_hidden_states_out
-            sm.tag(final_hidden_states)
+            final_hidden_states += shared_output
```
merrymercy, line 820: This is good!

merrymercy (changes requested):

python/sglang/srt/distributed/parallel_state.py
```diff
@@ -584,15 +607,37 @@ def _all_reduce_in_place(self, input_: torch.Tensor) -> None:
...
+        ):
+            with pynccl_comm.change_state(
+                enable=True, stream=torch.cuda.current_stream()
+            ):
+                pynccl_comm.reduce_scatter(output, input)
+        else:
+            torch.distributed.reduce_scatter_tensor(
+                output, input, group=self.device_group
+            )
         return output
```
yuan-luo, line 675: return output is unnecessary now.

python/sglang/srt/layers/quantization/modelopt_quant.py
yuan-luo: Agree. The pervasive use of use_symmetric_memory compromises the readability of the code.

python/sglang/srt/layers/communicator.py
nvcastet: Those 2 variables are already cached: tp group and is_max_len for this current batch.

python/sglang/srt/layers/dp_attention.py
nvcastet: Yes when the server flag `--enable-symm-mem` is used. Tensors here are always symmetric, so there is not a case where it needs to be disabled.

python/sglang/srt/layers/moe/topk.py
nvcastet: `is_allocation_symmetric` is batch-specific so it is cached in `dp_attention.py` for the scope of a batch.
`get_tp_group` gets the communicator spanning the symmetric group, depending on the cases it does not have to be the generic TP communicator, it could the TP attention communicator, etc...
Besides those 2 read-accessor calls and use_symmetric_memory, a nullcontext will be created if symmetric memory is disabled.
During cuda graph replays, this host code will not be re-executed for the life-time of the application.

python/sglang/srt/layers/quantization/modelopt_quant.py
nvcastet: First use  is to annotate the input used for the "quantize-before-comm" allgatherv.
Second use is to annotate the output buffer of the fused-MoE since it is not an in-place op. This buffer will then be used for reducescatter or allreduce comm.

> In the worst case, we can create empty context to avoid code duplication.

@merrymercy `use_symmetric_memory` uses a nullcontext when it is disabled.

There is no alternative to achieve "zero-copy" between compute and communication buffers than to annotate tensors that will be used for communication.
PyTorch put in-place this context-base mechanism and PyTorch team is moving forward to expand its support to `torch.compile` internals.

python/sglang/srt/distributed/parallel_state.py
nvcastet: I kept the code "as-is" to avoid unrelated changes. But yes we could clean-up those function signatures in another PR if needed.

nvcastet:
@merrymercy I will wait to rebase and solve merge-conflicts until you tell me it is ok to move forward with this PR.

To be honest, it has been 2 months since I opened the PR, we have made a lot of efforts to improve it and make it as clean as possible while still delivering significant e2e perf gain. I don't want all of the us including you to spend more time here if there is not a clear end to the back and forth in sight.

I hear that you have concerned about python-side perf when this feature is disabled. I tried to address them as best as possible while keeping the code modular. I don't believe there is currently overhead in critical code paths that is visible in a e2e run. I also believe the host code will only be executed once per batch-size when cuda-graph is used.

Please, let us know your thoughts.

python/sglang/srt/layers/moe/topk.py
nvcastet: For both the `communicator` arg and the `disabled` arg we could pass the function itself or a lambda where the function calls will be deferred after the "disabled" check. So those functions/lambda will not be called when symmetric memory is globally disabled.
@merrymercy Would that address your concerns?

nvcastet:
@merrymercy Let me know if implemented the logic in comment https://github.com/sgl-project/sglang/pull/9358#discussion_r2466299081 would address your python perf concerns.

python/sglang/srt/distributed/device_communicators/pynccl_allocator.py
```diff
@@ -29,12 +30,23 @@
...
 
 
 def is_symmetric_memory_enabled():
     return get_global_server_args().enable_symm_mem
 
 
+def is_symmetric_memory_tensor(tensor: torch.Tensor):
+    if not is_symmetric_memory_enabled() or _cached_pool_snapshot is None:
+        return False
+    for segment in _cached_pool_snapshot:
```
merrymercy, line 43: This runs a for loop over a long list, will it be slow?
I suspect it is even slower than the old appraoch `sm.tag`

python/sglang/srt/layers/dp_attention.py
```diff
@@ -298,6 +317,10 @@ def is_dp_attention_enabled() -> bool:
     return _ENABLE_DP_ATTENTION_FLAG
 
 
+def is_allocation_symmetric() -> bool:
+    return not is_dp_attention_enabled() or is_dp_max_padding()
```
merrymercy, line 321: why cannot we use symmetric memory for sum padding?

python/sglang/srt/distributed/device_communicators/pynccl_allocator.py
nvcastet: So I measured the perf it was around 2-3us, the `_cached_pool_snapshot` is just a dictionary with only symmetric memory segments, not the full memory used by the app.

python/sglang/srt/layers/dp_attention.py
nvcastet: Because in "sum padding", the input tensors are not allocated with the same size on each GPU (batch size) which breaks the requirement for symmetric allocation.

PR #9379, opened 2025-08-20
Fix model loading error when doing weights sync in RL training

rich-junwang:
## Motivation

In RL training, we have to sync training model weights to inference engine, e.g. sgl backend. When we are doing the weight sync, sometimes, we pass in the whole model weights, but we only need to load the model parameters that are only needed on the current rank. This PR is to fix the gpt-oss weight sync issue.

## Modifications
When there are extra model parameters that are not loaded by the model, it's okay. We don't want to raise the error.

## Accuracy Tests

Test training run works. 

## Checklist

### DIFF
--- a/python/sglang/srt/models/gpt_oss.py
+++ b/python/sglang/srt/models/gpt_oss.py
@@ -1139,7 +1139,7 @@ class GptOssForCausalLM(nn.Module):
         not_loaded_params = [k for k, v in params_checker.items() if not v]
         if tp_rank == 0:
             if len(not_loaded_params) > 0:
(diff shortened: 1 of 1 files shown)


rich-junwang:
Hi @Ying1123 , @zhaochenyang20 , can either of you take a look at this PR? This is a small fix for the weight sync for RL training. Let me know if you have any questions. Thanks.

zhaochenyang20:
> Hi @Ying1123 , @zhaochenyang20 , can either of you take a look at this PR? This is a small fix for the weight sync for RL training. Let me know if you have any questions. Thanks.

What kind of frameworks are you using for RL of gpt-oss?

PR #9238, opened 2025-08-15
Enables speculative decoding for the trtllm_mla attention backend

pranavm-nvidia:
## Motivation

Enables speculative decoding for the trtllm_mla attention backend

## Modifications

Adds a new `TRTLLMMLAMultiStepDraftBackend` and allows trtllm_mla backend to be used when speculative decoding is enabled.

## Accuracy Tests

<details>

<summary>TRTLLM-MLA FP16 deepseek/Deepseek-R1 FP8 model GSM8K and MMLU </summary>

Server Command:
```bash
python -m sglang.launch_server --model-path "deepseek-ai/DeepSeek-R1" --trust-remote-code \
    --attention-backend trtllm_mla --page-size 64 --tp-size 8 --max-running-requests 512 --cuda-graph-max-bs 512 \
    --host "0.0.0.0" --port 30000 --mem-fraction-static 0.70 --dist-timeout 1800 --speculative-algorithm EAGLE \
    --speculative-eagle-topk=1 --speculative-num-draft-tokens 4 --speculative-num-steps 3
```

GSM8K Command:
```bash
$ python3 benchmark/gsm8k/bench_sglang.py --num-shots 5 --num-questions 1319 --parallel 512 --port 30000
...
Accuracy: 0.957
Invalid: 0.000
Latency: 36.817 s
Output throughput: 3414.123 token/s
```

MMLU Command:
```bash
$ python3 benchmark/mmlu/bench_sglang.py --ntrain 5 --data_dir benchmark/mmlu/data --save_dir benchmark/mmlu/results  --parallel 512 --port 30000
...
subject: abstract_algebra, #q:100, acc: 0.760
subject: anatomy, #q:135, acc: 0.830
subject: astronomy, #q:152, acc: 0.941
subject: business_ethics, #q:100, acc: 0.880
subject: clinical_knowledge, #q:265, acc: 0.928
subject: college_biology, #q:144, acc: 0.965
subject: college_chemistry, #q:100, acc: 0.630
subject: college_computer_science, #q:100, acc: 0.830
subject: college_mathematics, #q:100, acc: 0.770
subject: college_medicine, #q:173, acc: 0.879
subject: college_physics, #q:102, acc: 0.833
subject: computer_security, #q:100, acc: 0.910
subject: conceptual_physics, #q:235, acc: 0.932
subject: econometrics, #q:114, acc: 0.746
subject: electrical_engineering, #q:145, acc: 0.862
subject: elementary_mathematics, #q:378, acc: 0.939
[...]

PR #9384, opened 2025-08-20
[NVIDIA] Fix trtllm fp4 moe backend when used in MTP

kaixih:
This PR addresses the issue mentioned [here](https://github.com/sgl-project/sglang/pull/9238#issuecomment-3199279509).

The root cause is that some MoE layers may use an unquantized method. For example:
```
model.layers.3.mlp.experts → quantized  
model.decoder.mlp.experts  → unquantized
```

This may be triggered by the MTP settings.

`FusedMoE` works with both quantized and unquantized methods, which explains why the `flashinfer_cutlass` fp4 backend works fine.

`FlashInferFP4MoE` (i.e. the `flashinfer_trtllm` backend), however, only supports the modelopt_fp4 quantized method and fails with unquantized ones.

This PR fixes the issue by checking the quantization config. If no quantization config is found, we fall back to `FusedMoE`.

cc. @pranavm-nvidia @zhyncs @kushanam


pranavm-nvidia (approved):

zhyncs:
Hi @kaixih can you help fix the conflicts

kaixih:
@zhyncs Done. PTAL.

PR #9388, opened 2025-08-20
[bug] fix errors related to context length in SD

hnyls2002:
According to the document, `SGLANG_ALLOW_OVERWRITE_LONGER_CONTEXT_LEN` should be set to `False` by default.

When the draft model's context length is smaller than target model, there are tons of illegal memory access.

<details>
<summary>KV cache location IMA reproduce</summary>

```bash
MODEL=meta-llama/Meta-Llama-3.1-8B-Instruct
SPEC_MODEL=lmsys/sglang-EAGLE-LLaMA3-Instruct-8B
ATTN=${ATTN:-fa3}
NUM=${NUM:-1}
python3 -m sglang.bench_offline_throughput \
    --model-path $MODEL \
    --dataset-name random \
    --random-input-len 256 \
    --random-output-len 1750 \
    --random-range-ratio 1.0 \
    --num-prompts $NUM \
    --speculative-algorithm EAGLE \
    --speculative-draft-model-path $SPEC_MODEL \
    --speculative-num-steps 5 \
    --speculative-eagle-topk 8 \
    --speculative-num-draft-tokens 64 \
    --mem-fraction-static 0.7 \
    --cuda-graph-max-bs 2 \
    --dtype float16 \
    --disable-radix \
    --context-length 2048 \
    --allow-auto-truncate \
    --attention-backend $ATTN \
    --disable-cuda-graph
```
</details>

<details>
<summary>Error</summary>

```bash
/pytorch/aten/src/ATen/native/cuda/ScatterGatherKernel.cu:163: operator(): block: [0,0,0], thread: [62,0,0] Assertion `idx_dim >= 0 && idx_dim < index_size && "scatter gather kernel index out of bounds"` failed.
/pytorch/aten/src/ATen/native/cuda/ScatterGatherKernel.cu:163: operator(): block: [0,0,0], thread: [63,0,0] Assertion `idx_dim >= 0 && idx_dim < index_size && "scatter gather kernel index out of bounds"` failed.
[2025-08-20 01:17:22] Scheduler hit an exception: Traceback (most recent call last):
  File "/root/sglang/python/sglang/srt/managers/scheduler.py", line 2577, in run_scheduler_process
    scheduler.event_loop_normal()
  File "/usr/local/lib/python3.10/dist-packages/torch/utils/_contextlib.py", line 120, in decorate_context
    return func(*args, **kwargs)
  File "/root/sglang/python/sglang/srt/managers/scheduler.py", line 783, in event_loop_normal
[...]

PR #9391, opened 2025-08-20
Fix the `--allow-auto-truncate` argument in tokenizer manager.

hnyls2002:
For some unknown reason, when enabling the `--allow-auto-truncate` argument, the scheduler will perform the truncation, but the tokenizer manager will directly abort this request. This is not expected.


hnyls2002:
Forgot to push the in-place slice in this PR, will add it in #9388 
```
-               input_ids = input_ids[:_max_req_len]
+               del input_ids[_max_req_len:]
```

PR #9397, opened 2025-08-20
fix: HiRadixCache: fix prefetch completion race

pabloiyu:
## Motivation

SGLang allows for hierarchical KV caching. Under the default storage prefetch policy `best_effort`, the scheduler doesn’t have to wait for prefetch to finish: it can proceed to stage the request, which terminates the prefetch early and inserts no pages from storage. To wait for prefetch, we can use `wait_complete`.

 However, there is a race under `wait_complete`: the scheduler can call `self.tree_cache.check_prefetch_progress()` before the prefetch thread populates `operation.hash_value`. The completion check then treats 0 completed == 0 planned as done.

Fix: In `HiRadixCache.can_terminate_prefetch()`, only consider prefetch complete when `operation.hash_value` is non-empty and `operation.completed_tokens == len(operation.hash_value) * page_size`. This ensures the scheduler waits until the page list is prepared (and with `wait_complete`, until IO finishes).

## Checklist


xiezhq-hermann:
Thank you for the catch, would checking things like `token_id` be a better solution? It does not introduce the extra complexity.

pabloiyu:
Yes I think there might be a better way to do it.

This is my understanding of the handling of a `PrefetchOperation` before my change
- Step 1: When request is added to the scheduler queue, the scheduler calls `HiRadixCache.prefetch_from_storage()` which in turn calls `HiCacheController.prefetch()`. This creates the `operation: PrefetchOperation` object which is now referenced in the following two places
	- `HiCacheController.prefetch_queue`
	- `HiRadixCache.ongoing_prefetch`
	
- Step 2: Two things proceed concurrently:
	- `Scheduler.get_new_batch_prefill()` calls 
		- `HiRadixCache.check_prefetch_progress()`
			- Depending on the given policy, if the prefetch is considered complete, it terminates it and marks operation as done. CHANGE 1 needed here.
			- Deletes it from `HiRadixCache.ongoing_prefetch`
		- `HiRadixCache.check_hicache_events()`
			- Takes out `operation.req_id` from `HiCacheController.prefetch_revoke_queue` and deletes it from `HiRadixCache.ongoing_prefetch`
	- The prefetch thread of `HiCacheController` takes the op out of its `prefetch_queue` and then does one of two things
		- If `storage_hit_count < self.prefetch_threshold`, enqueue to `prefetch_revoke_queue`
		- Else updates `operation.hash_values` and `operation.host_indices`. Also puts operation into `HiCacheController.prefetch_buffer` where the `prefetch_aux` thread does the actual page transfer. CHANGE 2 needed here.

Change 1:
- I think the cleanest way might be to not have the `HiCacheController.terminate_prefetch` called at all. Instead we simply look at `operation.is_done()` if we want to check if prefetch has finished for the `wait_complete` policy, with `HiCacheController` as the single source of truth updating that flag.

Change 2:
- It would make sense to mark the operation as done when the `prefetch_aux` thread has finished the page transfer. This ensures `is_done()` reflects the true completion of data movement.

ykwd:
The condition for marking an operation as *completed* is:

```
operation.completed_tokens == len(operation.hash_value) * self.page_size
```

However, when the operation is first initialized, `hash_value` is empty. The `hash_value` is computed during the prefetch process by invoking the `storage_hit_query` interface, which determines how many pages were hit. Therefore, when `can_terminate_prefetch` is checked for the first time, there is a high chance that `hash_value` has not yet been computed. At this point, both `len(hash_value)` and `completed_tokens` are `0`, which makes the condition appear satisfied, incorrectly signaling that the prefetch is already complete.

It seems the current code in this PR has some issues. The `mark_done` method is intended to be invoked **after termination**, to notify the prefetch thread that the operation has ended and that `completed_token` should no longer be modified by the prefetch thread.

However, in the current implementation, `mark_done` is called by the prefetch thread itself after prefetching finishes, which may introduce two problems:

1. The prefetch thread cannot detect when the operation actually terminates (e.g., due to timeout or best-effort conditions).
2. The original purpose of `mark_done` was to address memory release issues based on `completed_token` (previously, there was a risk of memory leakage). With the new change, it seems this problem may reappear.

So, how about the following simple solution:

```python
        if len(operation.hash_value) == 0:
            completed = False
        else:
            completed = (
                operation.completed_tokens == len(operation.hash_value) * self.page_size
            )
```

pabloiyu:
Ok makes sense, thanks for the explanation. Will go back to the original proposal of checking that `operation.hash_value` is not empty.

xiezhq-hermann:
Would it be easier to just check for 
```
completed = (
                operation.completed_tokens == len(operation.token_ids)
            )
```

pabloiyu:
I don’t think that would work. `operation.token_ids` is only the upper bound of what we’d like to prefetch. In `prefetch_thread_func` we probe the store with `self.storage_backend.exists(...)` and shrink the plan to the pages that actually exist; that finalized plan is captured in `operation.hash_value`. The number of tokens we can really fetch is therefore `len(operation.hash_value) * self.page_size`, and prefetch should be considered complete when `operation.completed_tokens` reaches that value.

xiezhq-hermann (approved):

ykwd:
I find during my benchmark that `if len(operation.hash_value) == 0:` may not mean the prefetch operation has not started. If the `storage_hit_count < self.prefetch_threshold`, then the hash_value is still empty. But in this case, we should terminate prefetch immediately.

ykwd:
By the way, it would be better if we could terminate the prefetch earlier when a data fetch failure or partial failure (i.e., some data are retrieved while others fail) occurs in `_page_transfer`, instead of waiting for timeout or waiting for complete.

xiezhq-hermann:
> I find during my benchmark that `if len(operation.hash_value) == 0:` may not mean the prefetch operation has not started. If the `storage_hit_count < self.prefetch_threshold`, then the hash_value is still empty. But in this case, we should terminate prefetch immediately.

In this case the operation will be revoked already, might not be necessary to be terminated here.

ykwd:
> > I find during my benchmark that `if len(operation.hash_value) == 0:` may not mean the prefetch operation has not started. If the `storage_hit_count < self.prefetch_threshold`, then the hash_value is still empty. But in this case, we should terminate prefetch immediately.
> 
> In this case the operation will be revoked already, might not be necessary to be terminated here.

Yes, I just verified it — the operation is revoked very quickly, so there’s no issue in this case.

PR #9404, opened 2025-08-20
Fixed the issue where eagle3  TPOT was not as good as without eagle3.

jiapingW:
## Motivation
eagle3 has a accept length over 2 but don't get a speedup TPOT than without eagle3. We found that the top_p_renorm operation is time-consuming in Eagle3 verify stage. Disabling this operation when the top_p parameter is not set can significantly improve the default Eagle3 decoding speed when temperature is not 0 and  top_p is not set.

## Checklist

### DIFF
--- a/python/sglang/srt/speculative/eagle_utils.py
+++ b/python/sglang/srt/speculative/eagle_utils.py
@@ -453,12 +453,13 @@ class EagleVerifyInput:
                     sampling_info.top_ks, self.draft_token_num, dim=0
                 ),
             )  # (bs * draft_token_num, vocab_size)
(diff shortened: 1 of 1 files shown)


jiapingW:
This can benchmark use qwq-32b and eagle3 model from https://huggingface.co/w497273/qwq-32b-eagle3/tree/main. The server argument are speculative-eagle-topk=1, speculative-num-steps=3 and speculative-num-draft-tokens=4. Request's arguments are max-concurrency=1 and temperature=1.0.

zhyncs (approved):

PR #9405, opened 2025-08-20
Use dual stream for DS MoE whenever cuda graph is used (instead of with token threshold)

trevor-m:
## Motivation

The token threshold was hardcoded to 1024, add an env var to make it more flexible.

## Checklist


python/sglang/srt/models/deepseek_v2.py
```diff
@@ -436,12 +439,11 @@ def forward(
         use_reduce_scatter: bool = False,
     ) -> torch.Tensor:
         if not self._enable_deepep_moe:
-            DUAL_STREAM_TOKEN_THRESHOLD = 1024
             if (
                 self.alt_stream is not None
                 and self.num_fused_shared_experts == 0
                 and hidden_states.shape[0] > 0
-                and hidden_states.shape[0] <= DUAL_STREAM_TOKEN_THRESHOLD
+                and hidden_states.shape[0] <= self.dual_stream_token_threshold
```
trevor-m: TRT-LLM doesn't have a threshold and instead enables dual stream whenever cuda graph is used. https://github.com/NVIDIA/TensorRT-LLM/blob/e5e417019b3b06d8f4b5e870937e18b77424211d/tensorrt_llm/_torch/modules/multi_stream_utils.py#L60

Should we adopt the same? If the number of tokens is high, the driver should automatically serialize the streams anyway. @Alcanderian

merrymercy: makes sense. We can use TRT-LLM method

merrymercy (changes requested):

Fridge003 (approved):

Fridge003:
@trevor-m Can you please add a profile screenshot on how it looks like when token > 1024

PR #9429, opened 2025-08-21
Add target module validation for init adapters

Beichen-Ma:
## Motivation

Currently, we assume that users specify --lora-target-modules is correct target module names. However, if this assumption is violated—specifically when:
	•	`--lora-path` is provided, and
	•	`--lora-target-modules` does not include all the actual adapter’s target_modules,
then it would raise ValueError during inference, such as: 
`ValueError: Cannot find target module name for base_model.model.model.layers.0.mlp.down_proj.lora_A.weight in {'embed_tokens'}`

## Modifications

We add checks that, during initialization, each LoRA adapter's required modules are included in the user-specified `--lora-target-modules` list, and provides clear error messages if not.

## Accuracy Tests

Tested with specifying limited `lora_target_modules` and expected ValueError raised, e.g. `ValueError: LoRA adapter 'ce371f2bff874589bf7bda027ccf6300' contains target modules ['down_proj', 'gate_up_proj', 'o_proj', 'qkv_proj'] that are not included in the specified --lora-target-modules ['embed_tokens']. Please update --lora-target-modules to include all required modules: ['down_proj', 'embed_tokens', 'gate_up_proj', 'o_proj', 'qkv_proj'], or use 'all' to enable all supported modules.`

## Checklist


python/sglang/srt/lora/lora_manager.py
```diff
@@ -422,6 +422,28 @@ def init_lora_shapes(
 
         if target_modules is not None:
             self.target_modules = set(target_modules)
+            user_normalized_modules = get_normalized_target_modules(self.target_modules)
+            for lora_id, config in self.configs.items():
+                if not isinstance(config.target_modules, list):
+                    raise ValueError(
+                        f"SGLang currently only supports inferring LoRA target modules when a list of "
+                        "suffixes is provided in `target_modules` field of PEFT config. Please explicitly "
+                        "specify `--lora-target-modules` during server startup. You can specify `all` to "
```
lifuhuang: nit: Is there any chance we can re-organize the code a bit to reduce repetition?

e.g., would it be more succinct if we reuse the same loop for both cases and check subset only if target_modules is not None? That way you don't need to copy paste the error message twice.

python/sglang/srt/lora/lora_manager.py
```diff
@@ -422,6 +422,28 @@ def init_lora_shapes(
...
+                config_normalized_modules = get_normalized_target_modules(
+                    config.target_modules
+                )
+                if not config_normalized_modules.issubset(user_normalized_modules):
+                    unsupported_modules = (
+                        config_normalized_modules - user_normalized_modules
+                    )
+                    raise ValueError(
+                        f"LoRA adapter '{lora_id}' contains target modules {sorted(unsupported_modules)} "
+                        f"that are not included in the specified --lora-target-modules {sorted(user_normalized_modules)}. "
```
lifuhuang: You can get user-friendly lora_name in `self.lora_refs[id].lora_name`. lora_id is an auto-generated random UUID.

lifuhuang (commented):
Thank you for the contribution :)

python/sglang/srt/lora/lora_manager.py
Beichen-Ma: Thanks for the information! I'll make sure to use the user-friendly name.

python/sglang/srt/lora/lora_manager.py
Beichen-Ma: Sure, I already fixed it to reduce the redundant loop.

python/sglang/srt/lora/lora_manager.py
```diff
@@ -422,6 +422,28 @@ def init_lora_shapes(
 
         if target_modules is not None:
             self.target_modules = set(target_modules)
+            user_normalized_modules = get_normalized_target_modules(self.target_modules)
+            for lora_id, config in self.configs.items():
```
lifuhuang: nit: with your change, now L458 seems a bit redundant. In theory, we only need to call get_normalized_target_modules once for each target_modules (e.g., once for the --lora-target-modules and once for each adapter config).

lifuhuang: It's a bit hard to describe in text so I pushed a commit directly to your branch, please let me know if it looks good to you. Feel free to make adjustment / revert as needed. Thanks!

Beichen-Ma: Thanks for making the changes! I agree with your point and it makes sense to me. Appreciate your help!

lifuhuang (approved):
LGTM

Fridge003:
@Beichen-Ma Please fix lint

PR #9434, opened 2025-08-21
enable llama3.1-8B on xpu

huaiyuzh:
This PR is to enable meta-llama/Llama-3.1-8B-Instruct with datatype BF16 on XPU.


python/sglang/srt/layers/activation.py
```diff
@@ -81,17 +81,20 @@ def forward_npu(self, x: torch.Tensor) -> torch.Tensor:
         out = torch_npu.npu_swiglu(x)
         return out
 
+    def forward_xpu(self, x: torch.Tensor) -> torch.Tensor:
+        d = x.shape[-1] // 2
+        output_shape = x.shape[:-1] + (d,)
+        out = torch.empty(output_shape, dtype=x.dtype, device=x.device)
+        silu_and_mul(x, out)
+        return out
```
mingfeima, line 89: call `forward_native` from forward_xpu util we have C++ kernels ready.

we don't have to rewrite it again here.

mingfeima (changes requested):

python/sglang/srt/layers/activation.py
mingfeima: sorry, my mistake. I misunderstood.

mingfeima (approved):

Alcanderian (approved):

python/sglang/srt/layers/activation.py
```diff
@@ -70,8 +72,6 @@ def forward_cuda(self, x: torch.Tensor) -> torch.Tensor:
 
     def forward_cpu(self, x: torch.Tensor) -> torch.Tensor:
         if _is_cpu_amx_available:
-            d = x.shape[-1] // 2
```
zhyncs, line 73: why did you remove this?

huaiyuzh: This is useless code.

PR #9450, opened 2025-08-21
Fix: /flush_cache API hangs when using --skip-server-warmup.

solrex:
## Motivation

The `/flush_cache` API is currently no-return when the server is launched with the --skip-server-warmup argument, causing the client to hang.

```
python3 -m sglang.launch_server --model-path ANY_MODEL --disable-cuda-graph --skip-server-warmup
curl http://localhost:30000/flush_cache 
```

## Modifications

* Add `self.auto_create_handle_loop()` when `/flush_cache` is called.

## Checklist


hnyls2002:
Thanks @solrex! flush_cache now starts the handle loop, so the --skip-server-warmup hang is fixed on
  main - closing. Please reopen if it still hangs.

PR #9456, opened 2025-08-21
Add Qwen3-30B-A3B-Thinking-2507 support on AMD GPUs.

sogalin:
## Motivation
Add Qwen3-30B-A3B-Thinking model support on AMD GPUs. 

## Modifications
Enabling triton backend on AMD GPU.

## Accuracy Tests
SGLANG_USE_AITER=0 python3 -m sglang.launch_server --model-path Qwen3-30B-A3B-Thinking-2507/ --tp 8 --trust-remote-code --chunked-prefill-size 130172 --max-running-requests 128 --mem-fraction-static 0.85 --attention-backend aiter --enable-torch-compile

python3 benchmark/gsm8k/bench_sglang.py --num-questions 2000 --parallel 2000
Accuracy: 0.858
Invalid: 0.000
Latency: 103.769 s
Output throughput: 2743.774 token/s

## Checklist

### DIFF
--- a/python/sglang/srt/layers/moe/fused_moe_triton/fused_moe.py
+++ b/python/sglang/srt/layers/moe/fused_moe_triton/fused_moe.py
@@ -44,11 +44,12 @@ _is_cpu_amx_available = cpu_has_amx_support()
 _is_cpu = is_cpu()
 _use_aiter = get_bool_env_var("SGLANG_USE_AITER") and _is_hip
 
-if _is_cuda or _is_hip:
+if _is_cuda:
     from sgl_kernel import gelu_and_mul, silu_and_mul
 elif _is_cpu and _is_cpu_amx_available:
     pass
 elif _is_hip:
+    from sgl_kernel import gelu_and_mul, silu_and_mul
     from vllm import _custom_ops as vllm_ops  # gelu_and_mul, silu_and_mul
 
     if _use_aiter:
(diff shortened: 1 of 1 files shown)


python/sglang/srt/layers/moe/fused_moe_triton/fused_moe.py
```diff
@@ -49,6 +49,7 @@
 elif _is_cpu and _is_cpu_amx_available:
     pass
 elif _is_hip:
+    from sgl_kernel import gelu_and_mul, silu_and_mul
     from vllm import _custom_ops as vllm_ops  # gelu_and_mul, silu_and_mul
```
hubertlu-tw: Shall we remove `from vllm import _custom_ops as vllm_ops` for `moe_sum` or only add this when `_is_hip==True and _is_aiter==False`? Since MI250X does not have full aiter supports I think.

python/sglang/srt/layers/moe/fused_moe_triton/fused_moe.py
```diff
@@ -1537,7 +1538,7 @@ def fused_experts_impl(
                     gemm1_alpha,
                     gemm1_limit,
                 )
-            elif _is_cuda:
+            elif _is_cuda or _is_hip:
```
hubertlu-tw, line 1542: There is another `gelu_and_mul` can be imported from `sgl_kernel `when `_is_hip ==True`.

HaiShaw (approved):
LGTM

PR #9433, opened 2025-08-21
[fix] Fix mxfp4 weight loading bug with TP sharding in GPT-OSS

hlu1:
## Motivation

gpqa and mmlu with tp=4 were incorrect. The tp=1 and tep=4 scores are correct though. The bug is due to incorrect TP sharding of mxfp4 MoE weights, which missed loading the last couple of rows of data in the last rank. 

## Modifications

Fixed the bug in TP sharding of mxfp4 weights.

## Accuracy Tests
The following results are all on B200. I'll run more on H200 later.
```
mmlu
correct:
120b high + temperature: 1.0 + tp1
[{'eval_name': 'mmlu', 'model_name': 'dummy-high_temp1.0_20250821_001009', 'metric': 0.8871955561885772}]

120b high + temperature: 1.0 + tep4
[{'eval_name': 'mmlu', 'model_name': 'dummy-high_temp1.0_20250821_010817', 'metric': 0.887409200968523}]

120b high + temperature: 1.0 + tp4
bad, before fix:
[{'eval_name': 'mmlu', 'model_name': 'dummy-high_temp1.0_20250821_004638', 'metric': 0.8835635949294972}]

good, after fix
[{'eval_name': 'mmlu', 'model_name': 'dummy-high_temp1.0_20250821_055350', 'metric': 0.8876228457484688}]

gpqa

120b high + temperature: 1.0 + tp4
bad, before fix:
[{'eval_name': 'gpqa', 'model_name': 'dummy-high_temp1.0_20250821_063952', 'metric': 0.7651515151515151}]

good, after fix
[{'eval_name': 'gpqa', 'model_name': 'dummy-high_temp1.0_20250821_061341', 'metric': 0.8017676767676768}]
```

## Benchmarking and Profiling

N/A

## Checklist

- [x] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
[...]

PR #9473, opened 2025-08-22
[fix] Fix mxfp4 triton MoE tp bug

hlu1:
## Motivation

https://github.com/sgl-project/sglang/pull/9433 broke GPT-OSS on Hopper because the MoE intermediate size was not padded after sharding on Hopper. Adding padding would fix the issue. However, it's only a workaround. The correct fix is to change the way of the weight sharding for block scaling data formats (mxfp4, nvfp4 etc) to the following before creating the weights and then add padding on top of that:

```
        intermediate_size_block = intermediate_size // mxfp4_block
        per_rank_intermediate_size_block = math.ceil(
            intermediate_size_block / moe_tp_size
        )
        per_rank_intermediate_size = per_rank_intermediate_size_block * mxfp4_block
```
This is a bigger change and will require more testing. I will do this in a followup PR.

## Modifications

Added intermediate size padding to triton MoE

## Accuracy Tests

8 x H100 tp8
```
before
[{'eval_name': 'gpqa', 'model_name': 'gpt-oss-120b-high_temp1.0_20250822_210411', 'metric': 0.7556818181818182}]

After
[{'eval_name': 'gpqa', 'model_name': 'gpt-oss-120b-high_temp1.0_20250823_015440', 'metric': 0.7986111111111112},
{'eval_name': 'mmlu', 'model_name': 'gpt-oss-120b-high_temp1.0_20250823_024653', 'metric': 0.8889759293547927}, 
{'eval_name': 'aime25', 'model_name': 'gpt-oss-120b-high_temp1.0_20250823_024653', 'metric': 0.9041666666666667}]
```

Perf on Hopper is going to regress a little bit because of padding, which means more computation than necessary. But the padding is necessary. Previously the weights were not even fully loaded so the computation is less that what it should have been.

## Checklist


zhyncs:
@hlu1 please also update this https://github.com/sgl-project/sglang/pull/9469

hlu1:
> @hlu1 please also update this #9469

Working on it.

hlu1:
Updated PR and summary with accuracy test results.

zhyncs (approved):

PR #9485, opened 2025-08-22
Partially unify triton per token group quant kernels

fzyzcjy:
## Motivation

reasons
* prereq of the +6% e2e quant kernel optimization
* enable future unifying the almost-identical fp8 and int8 code

## Checklist

### DIFF
--- a/python/sglang/srt/layers/quantization/fp8_kernel.py
+++ b/python/sglang/srt/layers/quantization/fp8_kernel.py
@@ -45,7 +45,7 @@ _use_aiter = get_bool_env_var("SGLANG_USE_AITER") and _is_hip
 if _is_cuda:
     from sgl_kernel import (
         sgl_per_tensor_quant_fp8,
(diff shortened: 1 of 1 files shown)


python/sglang/srt/layers/quantization/fp8_kernel.py
```diff
@@ -297,6 +313,117 @@ def per_token_group_quant_fp8(
     return x_q, x_s
 
 
+# backward compatibility
+per_token_group_quant_fp8 = _per_token_group_quant_8bit_raw
+
+
+def _per_token_group_quant_8bit_fuse_silu_and_mul(
```
Alcanderian, line 320: Is this only for deepgemm?

Alcanderian (approved):

python/sglang/srt/layers/quantization/fp8_kernel.py
fzyzcjy: (discussed offline)

PR #9498, opened 2025-08-22
[router] add right rustls dependency in sgl-router cargo.toml

Bruce-x-1997:
## Motivation

When I enable service-discovery in sgl-router (0.5.0rc1) , I meet a traceback when launch the router pod:
`Could not automatically determine the process-level CryptoProvider from Rustls crate features.
Call CryptoProvider::install_default() before this point to select a provider manually
`
and I could not find any thing about CryptoProvider and rustls.
then I rebuild image in 0.5.0rc2, there is no use.But I could launch router pod without bug in 0.4.10.post2.
The traceback is from the code
`let client = Client::try_default().await?;
`
and the Client is a class in kube.So I revert kube(from 1.1.0 to 0.88.1)and kubeapi(from 0.25.0 to 0.21.0)
it works well.
 
## Modifications

just fix kube & kubeapi version in cargo.toml
update:
if we add 0.23 rustls dependency, it works as well, but 0.21 version could meet a crypto-not-found bug
so just add a 0.23 rustls dependency.And add related code in service_discovery.rs

## Checklist


slin1237:
instead of doing this
can you try
adding
```
let _ = rustls::crypto::ring::default_provider().install_default();
```
around here
https://github.com/sgl-project/sglang/blob/main/sgl-router/src/service_discovery.rs#L189

and add
```
rustls = { version = "0.23", default-features = false, features = ["ring", "std"] }
```
to `Cargo.toml`

Bruce-x-1997:
@slin1237 please help review, thanks

LukasBluebaum:
`metrics-exporter-prometheus` pulls in `aws-lc-rs`. So it also seems to be enough to restrict its features:

`metrics-exporter-prometheus = { version = "0.17", default-features = false, features = ["http-listener"] }`

Bruce-x-1997:
> instead of doing this can you try adding
> 
> ```
> let _ = rustls::crypto::ring::default_provider().install_default();
> ```
> 
> around here https://github.com/sgl-project/sglang/blob/main/sgl-router/src/service_discovery.rs#L189
> 
> and add
> 
> ```
> rustls = { version = "0.23", default-features = false, features = ["ring", "std"] }
> ```
> 
> to `Cargo.toml`

ok, i will do it, thanks, but before I ever tried to add rustls and call rustls::crypto::CryptoProvider but crypto could not be found, when rustls version is 0.21.

Bruce-x-1997:
> `metrics-exporter-prometheus` pulls in `aws-lc-rs`. So it also seems to be enough to restrict its features:
> 
> `metrics-exporter-prometheus = { version = "0.17", default-features = false, features = ["http-listener"] }`

so should I add this one in cargo as well?

Bruce-x-1997:
0.23  rustls could work, 0.21 could not.
so we might need a service-disacovery ci to cover such case.
I push commit immediately

Bruce-x-1997:
@slin1237 has add 0.23 dependency, and it could work.please review again, thanks

slin1237 (approved):

PR #9503, opened 2025-08-22
[router] ignore client error when record failure in pd_router

Bruce-x-1997:
## Motivation

we met a problem like that:  many abnormal client requests(sglang return 400 because of context-length etc), and pd_router find a decode pod return too many 400 response and the circuit breaker records the errors and mark the decode pod is unhealth.But client err req could not be recorded as a server error.So I just filter client error when recoding errors.

## Modifications

in pd_router, 4xx response(means client error) would be recorded as failure and trigger circuit breaker to be open, and the decode pod would be recorded as unavaliable.Now I just filter client_error.

## Checklist


Bruce-x-1997:
@slin1237 please help review, thanks

slin1237 (approved):

PR #9540, opened 2025-08-23
Fix occasional OOM inside LogitsProcessor

pjin-nvidia:
## Motivation

Occasionally an OOM can occur inside `LogitsProcessor` where the logits are casted to fp32. This tends to occur after a few steps of RL, depending on the initial `mem_fraction_static`, due to an increase in the overall GPU memory usage during RL training.

## Modifications

This PR simply defers the fp32 cast to after the logits are sliced.

## Checklist

### DIFF
--- a/python/sglang/srt/layers/logits_processor.py
+++ b/python/sglang/srt/layers/logits_processor.py
@@ -517,6 +517,7 @@ class LogitsProcessor(nn.Module):
             logits = logits[:, : self.config.vocab_size]
 
         if self.final_logit_softcapping:
+            logits = logits.to(dtype=torch.float32)
             fused_softcap(logits, self.final_logit_softcapping)
 
         return logits
--- a/python/sglang/srt/layers/logits_processor.py
+++ b/python/sglang/srt/layers/logits_processor.py
@@ -326,7 +326,7 @@ class LogitsProcessor(nn.Module):
         logits = self._get_logits(pruned_states, lm_head, logits_metadata)
         sampled_logits = (
             logits[sample_indices] if sample_indices is not None else logits
-        )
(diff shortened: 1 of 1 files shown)


python/sglang/srt/layers/logits_processor.py
```diff
@@ -514,7 +514,7 @@ def _get_logits(
             logits_buffer.copy_(logits[:, : self.config.vocab_size])
             logits = logits_buffer
         else:
-            logits = logits[:, : self.config.vocab_size].float()
+            logits = logits[:, : self.config.vocab_size]
```
ch-wan, line 517: we should defer the above branch too to keep code logic consistent

pjin-nvidia: in the branch above, the logits point to `logits_metadata.next_token_logits_buffer` which should already be pre-allocated in float32, so there is no fp32 cast to defer

see this assert: https://github.com/sgl-project/sglang/pull/9540/files/56c19973d3a50f4cb2319a982e996e5b7438bccd#diff-e7734e99e6b8d0c2279a1599685fef4b9fe4907e1ba5e0d9f240072670ae5879R513

pjin-nvidia: also I added a fallback cast back to fp32 before the final softcap in `_get_logits`

ch-wan: You are correct. But the code logic would be pretty messy. We'd better to keep output dtype consistent.

hnyls2002:
@pjin-nvidia I agree with Chen's opinion. Could you please keep the output dtype consistent?

PR #9541, opened 2025-08-23
Require users to disable chunked prefill when sending input_embeds

yhyang201:
## Motivation

At https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/managers/schedule_batch.py#L1198-L1202, there is no handling of `input_embeds` for requests with `chunked_prefill`. Therefore, users need to be informed that `chunked_prefill` should be disabled when sending `input_embeds`.

## Checklist


python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -530,6 +530,12 @@ async def _tokenize_one_request(
                     "Please add `--disable-radix-cache` when you launch the server "
                     "if you want to use input_embeds as inputs."
                 )
+            if self.server_args.chunked_prefill_size != -1:
+                raise ValueError(
+                    "input_embeds is provided while chunked prefill is enabled. "
```
JustinTong0323, line 607: maybe directly modify the chunked_prefill_size to -1 and log it?

hnyls2002: @yhyang201 Agree, please do it

PR #9549, opened 2025-08-24
fix(grok): remove duplicate replicate_lm_head configuration

vincentzed:
## Motivation

It's a duplication of line 837

## Checklist


zhyncs (approved):

python/sglang/srt/models/grok.py
```diff
@@ -842,10 +842,6 @@ def __init__(
         if self.is_weights_presharded:
             setattr(DefaultModelLoader, "_prepare_weights", _prepare_presharded_weights)
 
-        default_replicate_lm_head = False
```
vincentzed, line 845: duplicated on `837-840`, so it isn't necessary below

PR #9555, opened 2025-08-24
refactor(hicache): Introduce generic HiCacheStorageConfig for improved configuration management

hzh0425:
## Motivation

Introduce generic HiCacheStorageConfig for improved configuration management and simplify backend initialization

cc @xiezhq-hermann @stmatengss 

## Checklist


stmatengss:
Great work! Once the generic HiCacheStorageConfig is prepared, I can rebase #9256 accordingly.

xiezhq-hermann:
I think it would make more sense to combine with the proposal here to have common config like model name, and a customizable entry as well.
https://github.com/sgl-project/sglang/pull/9256

python/sglang/srt/mem_cache/hicache_storage.py
```diff
@@ -32,6 +33,13 @@ def get_hash_str(token_ids: List[int], prior_hash: str = None) -> str:
     return hasher.hexdigest()
 
 
+@dataclass
+class HiCacheStorageConfig:
+    tp_rank: int
+    tp_size: int
+    is_mla_model: bool
+
```
xiezhq-hermann, line 32: we probably need a model name as well.

ykwd:
Good work! One thing I don’t quite understand: why do we need to specify tp_rank, tp_size, and is_mla_model in the config file? These three parameters can already be dynamically obtained at runtime.

hzh0425:
> Good work! One thing I don’t quite understand: why do we need to specify tp_rank, tp_size, and is_mla_model in the config file? These three parameters can already be dynamically obtained at runtime.

This is to encapsulate these common configurations so that the backend can access them more conveniently

ykwd:
> > Good work! One thing I don’t quite understand: why do we need to specify tp_rank, tp_size, and is_mla_model in the config file? These three parameters can already be dynamically obtained at runtime.
> 
> This is to encapsulate these common configurations so that the backend can access them more conveniently

I see. Thanks for the explanation.

xiezhq-hermann (approved):

PR #9564, opened 2025-08-24
[RL] fix register the same ops multiple times

hebiao064:
## Motivation
Fix duplicate operator registration error in multi-engine scenarios

### Problem
When import SGLang multiple times within the same Python process (e.g., in the VERL framework), users encounter a `RuntimeError` due to duplicate PyTorch operator registration:
```
RuntimeError: Tried to register an operator (sglang::deep_gemm_fp8_fp8_bf16_nt(...)) with the same name and overload name multiple times. Each overload's schema should only be registered with a single call to def().
```

This occurs because PyTorch doesn't allow the same custom operator to be registered multiple times, but SGLang was attempting to register operators every time a module was imported.

### Solution
Enhanced the `direct_register_custom_op` function in `utils.py` to gracefully handle duplicate registrations by:

1. **Pre-registration check**: Before attempting to register an operator, check if it already exists in the target library namespace
2. **Silent duplicate handling**: Catch and silently ignore `RuntimeError` exceptions specifically related to duplicate operator registration
3. **Backward compatibility**: Maintain existing functionality for single-engine scenarios

#### 🔧 `python/sglang/srt/utils.py`
- Added pre-registration existence check using `hasattr(torch.ops, lib_name)`
- Enhanced error handling to catch and ignore duplicate registration `RuntimeError`
- Added comprehensive documentation explaining the multi-engine support
- Improved code formatting for better readability

#### 🔧 `python/sglang/srt/layers/quantization/mxfp4.py`
- Removed redundant hasattr checks (now handled in `direct_register_custom_op`)
- Simplified exception handling to focus on `AttributeError` only
- Cleaned up registration flow for both `dequant_mxfp4` and `quant_dequant_mxfp4` operators

### Testing
This fix has been designed to:
- ✅ Allow multiple SGLang engines to run in the same Python process without errors
- ✅ Maintain backward compatibility with existing single-engine deployments
- ✅ Preserve all existing operator functionality
- ✅ Support frameworks like VERL that spawn multiple engines

### Impact
- **Users**: Eliminates crashes when using SGLang in multi-engine frameworks like VERL
- **Developers**: Provides a robust foundation for operator registration that handles edge cases
- **Performance**: No performance impact; checks are minimal and only occur during initialization

### Related Issues
Fixes the duplicate operator registration error reported when using SGLang with VERL framework (2-4 engines in the same process).

## Checklist


hebiao064:
@nanjiangwill

nanjiangwill:
this feature is great! it is possible for people to import sglangrollout multiple times for different purpose(such as validation)

BBuf (approved):
LGTM

merrymercy:
Why do we need to register custom ops?
They are just python functions? Can we just import them and call them and get rid of all custom op registration?

hebiao064:
> Why do we need to register custom ops? They are just python functions? Can we just import them and call them and get rid of all custom op registration?

This PR intends to fix the multiple import of deep gemm: https://github.com/sgl-project/sglang/blob/db0831e0193d38df348b0b3ffc694b479cafb6a4/python/sglang/srt/layers/quantization/fp8_kernel.py#L107


and I just updated the mxfp4 as well since it also utilized `direct_register_custom_op`, I think you are right, mxfp4 doesn't need it, I'll try file a PR to simplify that.

The direct registration seems to be faster but the benefit of larger model is minimal: https://gist.github.com/youkaichao/ecbea9ec9fc79a45d2adce1784d7a9a5, maybe we should remove it entirely from our codebase

PR #9573, opened 2025-08-25
Support speculative decoding in hybrid attention backend

### LINKED ISSUE #9330: [Bug] AssertionError: Currently HybridAttentionBackend does not support speculative decoding.
### Checklist

- [ ] 1. I have searched related issues but cannot get the expected help.
- [ ] 2. The bug has not been fixed in the latest version.
- [ ] 5. Please use English, otherwise it will be closed.

### Describe the bug

```
python3 -m sglang.launch_server --model deepseek-ai/DeepSeek-V3-0324 --tp 8 --speculative-algorithm EAGLE --attention-backend trtllm_mla --prefill-attention-backend flashinfer
```

### Reproduction

N/A

### Environment

N/A

Qiaolin-Yu:
## Motivation

Close #9330
### Launch the server
```
SGLANG_ALLOW_OVERWRITE_LONGER_CONTEXT_LEN=1 python -m sglang.launch_server   --model-path meta-llama/Llama-3.1-8B-Instruct   --port 8005   --tensor-parallel-size 1   --speculative-algo EAGLE3   --speculative-draft jamesliu1/sglang-EAGLE3-Llama-3.1-Instruct-8B   --speculative-num-steps 3   --speculative-eagle-topk 1   --speculative-num-draft-tokens 4   --mem-fraction 0.6     --dtype bfloat16 --prefill-attention-backend flashinfer --decode-attention-backend fa3
```

### Test accuracy
```
python3 benchmark/gsm8k/bench_sglang.py --num-shots 8 --num-questions 1319 --parallel 1319 --port 8005
```

Accuracy: 0.794

## Checklist


python/sglang/srt/layers/attention/hybrid_attn_backend.py
```diff
@@ -36,15 +45,26 @@ def init_forward_metadata_capture_cuda_graph(
...
+                bs,
+                num_tokens,
+                req_pool_indices,
+                seq_lens,
+                encoder_lens,
+                forward_mode,
+                spec_info,
+            )
+        else:
+            self.prefill_backend.init_forward_metadata_capture_cuda_graph(
```
Fridge003, line 59: Do we need to check whether speculative is enabled here? 
When we don't use spe, init_forward_metadata_capture_cuda_graph is not needed for prefill backend

python/sglang/srt/layers/attention/hybrid_attn_backend.py
```diff
@@ -57,16 +77,28 @@ def init_forward_metadata_replay_cuda_graph(
...
+                req_pool_indices,
+                seq_lens,
+                seq_lens_sum,
+                encoder_lens,
+                forward_mode,
+                spec_info,
+                seq_lens_cpu,
+            )
+        else:
+            self.prefill_backend.init_forward_metadata_replay_cuda_graph(
```
Fridge003, line 92: Same as above

test/srt/test_hybrid_attn_backend.py
```diff
@@ -105,5 +111,26 @@ def get_server_args(cls):
...
+    def get_server_args(cls):
+        return DEFAULT_SERVER_ARGS + [
+            "--speculative-algorithm",
+            "EAGLE",
+            "--speculative-draft",
+            DEFAULT_EAGLE_DRAFT_MODEL_FOR_TEST,
+            "--speculative-num-steps",
+            "3",
+            "--speculative-eagle-topk",
+            "1",
```
Fridge003: Do we need test for topk>1?

python/sglang/srt/layers/attention/hybrid_attn_backend.py
Qiaolin-Yu: But if speculative is not enabled, it will not enter this path?

test/srt/test_hybrid_attn_backend.py
Qiaolin-Yu: done

python/sglang/srt/layers/attention/hybrid_attn_backend.py
Fridge003: Oh, do you mean when sepc is disabled, init_forward_metadata_capture_cuda_graph will not be called for prefill backend?

I didn't check this

Qiaolin-Yu: yes. since the forward_mode in `capture_cuda_graph` will always be `decode` if spec is disabled.

Fridge003 (approved):
LGTM

PR #9601, opened 2025-08-25
fix host memory leaks in HiCacheStorage

huangtingwei9988:
## Motivation

[L261](https://github.com/sgl-project/sglang/blob/3aec3d4f8b9f4c2cc84df13b71c9880a4ec2a6aa/python/sglang/srt/mem_cache/hiradix_cache.py#L261)  [L230](https://github.com/sgl-project/sglang/blob/3aec3d4f8b9f4c2cc84df13b71c9880a4ec2a6aa/python/sglang/srt/mem_cache/hiradix_cache.py#L230) in the evict method, the node that is not backed up will be deleted from the radix tree. The last_host_node in prefetch kvcache may be this kind of node,  which will cause host memory leaks.
## Checklist

### DIFF
--- a/python/sglang/srt/mem_cache/hiradix_cache.py
+++ b/python/sglang/srt/mem_cache/hiradix_cache.py
@@ -473,6 +473,15 @@ class HiRadixCache(RadixCache):
         if not self.can_terminate_prefetch(operation):
             return False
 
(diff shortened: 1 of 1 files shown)


xiezhq-hermann:
replaced by https://github.com/sgl-project/sglang/pull/9648

PR #9634, opened 2025-08-26
metrics: support customer buckets for prompt/generation_tokens_histogram

acelyc111:
## Motivation

The histogram buckets for metrics prompt_tokens_histogram and generation_tokens_histogram are fixed, they may not match the real usage. For example, the generare tokens are averaged to 20k which is larger than the max fixed bucket, i.e. 10k, this cause the P99 always to 10k, which can't describe the real situation.

## Modifications

Keep the default fixed buckets, but also introduce customer rules, the behaviors are specified by newly added arguments `--prompt-tokens-buckets` and `--generation-tokens-buckets`. There are 3 behaviors:
- `default`: use the former fixed buckets
- `customer <buckets list>`: e.g. `--generation-tokens-buckets customer 10 20 30 50 100 1000 10000`
- `tse <middle>, <base>, <count>`: e.g. `--prompt-tokens-buckets tse 1000 3 15`, it generate bucket list as: `[0, 271, 757, 919, 973, 991, 997, 1000, 1003, 1009, 1027, 1081, 1243, 1729, 3187, 7561]`

## Checklist


whybeyoung:
LGTM, I have a similar use case.
In addition, I’d like to ask whether it’s possible to support analyzing input/output lengths by different users. For example, we could add a custom request-level label in the request header, such as:

user_id: 10001
user_id: 10002
user_id: 10003


and so on.
This way, I can perform data analysis based on inputs from different users.

acelyc111:
> For example, we could add a custom request-level label in the request header, such as:
> 
> user_id: 10001 user_id: 10002 user_id: 10003
> 
> and so on. This way, I can perform data analysis based on inputs from different users.

@whybeyoung What a coincidence, we have the same requirement. I think we can extract the labels in the header to the request object, then apply the labels when update metrics. I can take some time to implement it this week.

whybeyoung (approved):

python/sglang/srt/server_args.py
```diff
@@ -2185,6 +2201,12 @@ def check_server_args(self):
 
         # Check multi tokenizer
         assert self.tokenizer_worker_num > 0, "Tokenizer worker num must >= 1"
+        self.validate_buckets_rule(
+            "--prompt-tokens-buckets", self.prompt_tokens_buckets
+        )
+        self.validate_buckets_rule(
+            "--generation-tokens-buckets", self.generation_tokens_buckets
+        )
```
whybeyoung, line 2215: The configuration options for these two parameters are slightly complex, so perhaps we should add a separate documentation page to explain them.

acelyc111: @whybeyoung It make sense. I've updated the help message and docs, please take a look again, thanks!

whybeyoung (approved):

hnyls2002 (approved):

merrymercy:
what is tse?

acelyc111:
> what is tse?

it is short for "two sides exponential"

acelyc111:
It is described in the command help description.

https://github.com/sgl-project/sglang/blob/fce170480af716482736e80217e56dd1f86a8352/python/sglang/srt/server_args.py#L1577-L1579

PR #9643, opened 2025-08-26
fix(hicache): fix prefetch under heavy workload

xiaguan:
## Motivation

When the system encounters high hit ratio scenarios (such as dozens of long requests with nearly 100% hit ratios during my stress testing), it causes two critical issues:

1. **Memory allocation failures**

My current fix involves disabling prefetch operations.

2. **Rate limiting causes undefined operation states, leading to deadlocks**

The solution is to implement a brief requeue mechanism.

Here are some potential solutions worth discussing:

1. Defer memory allocation until just before batch_get operations to prevent excessive memory pinning in short timeframes, which would simultaneously address both issues mentioned above.

2. Implement backpressure using queue size limitations.

## Checklist

### DIFF
--- a/python/sglang/srt/managers/cache_controller.py
+++ b/python/sglang/srt/managers/cache_controller.py
@@ -772,12 +772,31 @@ class HiCacheController:
                 if operation is None:
                     continue
 
-                if (
-                    operation.host_indices is not None
-                ) and self.prefetch_rate_limit_check():
-                    hash_value, storage_hit_count = self._generic_storage_hit_query(
-                        operation
-                    )
+                # Check if this operation has already been queried for storage hits
+                if hasattr(operation, '_hash_value_cached'):
+                    # Restore cached values from previous query
+                    hash_value = operation._hash_value_cached
+                    storage_hit_count = operation._storage_hit_count_cached
+                else:
+                    # First time processing this operation
+                    hash_value = []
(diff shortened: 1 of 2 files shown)


xiaguan:
I think the prefetch limit is unnecessary. I think the root cause is that prefetch operations in the queue shouldn't consume memory - allocation should only occur when actually needed.

xiezhq-hermann:
> I think the prefetch limit is unnecessary. I think the root cause is that prefetch operations in the queue shouldn't consume memory - allocation should only occur when actually needed.

Thank you for the comment and I do think it is a very valid one. The reason to alloc memory beforehand is to ensure the consistency across TP workers. In the current design, any change made to the radix tree has to happen in the main controller thread to avoid potential state divergence.

xiaguan:
> > I think the prefetch limit is unnecessary. I think the root cause is that prefetch operations in the queue shouldn't consume memory - allocation should only occur when actually needed.
> 
> Thank you for the comment and I do think it is a very valid one. The reason to alloc memory beforehand is to ensure the consistency across TP workers. In the current design, any change made to the radix tree has to happen in the main controller thread to avoid potential state divergence.

Thank you for the explanation. In that case, I'll refrain from modifying the allocation logic in this PR and simply ensure the current code functions properly under extreme scenarios.

PR #9651, opened 2025-08-26
[model] Support POINTSV15Chat model

josephydu:
This PR support a new vision model, see https://github.com/Tencent/POINTS-Reader
1. **Simplicity**: POINTS-Reader is a very streamlined model that fully follows the structure of POINTS1.5, except that we have replaced Qwen2.5-7B-Instruct with Qwen2.5-3B-Instruct. Moreover, the input and output of POINTS-Reader are extremely straightforward. The input consists of a fixed prompt and a document image, and the output contains only a string (text extracted from the document image). The model's output is the final result delivered to the user without any post-processing.

2. **Performance**: Currently, POINTS-Reader supports extraction from both Chinese and English documents, achieving impressive results, with scores of 0.133 for English and 0.212 for Chinese on OmniDocBench.

3. **High Throughput**: With current mainstream inference frameworks, such as SGLang and vLLM, optimization is predominantly focused on LLMs. Thus, a large ViT would significantly impact the model’s throughput, which is why we selected a ViT with a moderate number of parameters (600M NaViT used in POINTS1.5). Combined with our support for SGLang, we currently achieve a very satisfactory throughput.

4. **Open-source Technical Approach**: In the POINTS-Reader paper, we propose a two-stage data augmentation strategy. The first stage leverages automated data to endow the model with basic document extraction capabilities. In the subsequent stage, continuous self-evolution improves the quality of data generated by the model. The self-evolution approach in the second stage is highly extensible and can be applied to virtually any model.


python/sglang/srt/multimodal/processors/points_v15_chat.py
JustinTong0323, line 1: Seems mostly copy from qwen_vl processor, Why not inherited from Qwen2_5VLImageProcessor?

python/sglang/srt/conversation.py
```diff
@@ -960,6 +960,25 @@ def generate_chat_conv(
...
+        stop_str=["<|im_end|>"],
+        image_token="<|vision_start|><|image_pad|><|vision_end|>",
+        video_token="<|vision_start|><|video_pad|><|vision_end|>",
+    )
+)
+
+@register_conv_template_matching_function
+def match_points_v15_chat(model_path: str):
+    if re.search(r"points", model_path, re.IGNORECASE):
+        return "points-v15-chat"
```
JustinTong0323: If chat-template is given by tokenizer/processor, we don't need to manually add one.

josephydu: fixed in new commits.

python/sglang/srt/multimodal/processors/points_v15_chat.py
josephydu: fixed in new commits.

josephydu:
Hello @JustinTong0323 
We didn't use any templates during training, so I think we should not add any template in inference, too. 
Does not adding any template here have the same effect as add a empty chat template?

JustinTong0323:
> Hello @JustinTong0323 We didn't use any templates during training, so I think we should not add any template in inference, too. Does not adding any template here have the same effect as add a empty chat template?

I think during training you just use the template from qwen vl by using the same tokenizer, it should be the same during inference. I am not familiar with training so point out if I am wrong, thx)

Also, update main on your side would lead to rerun all CIs, so please just ping me on slack to rerun the failed CI only.

josephydu:
When I try to fix lint, I met the following error when running `pre-commit run --all-files --show-diff`:
image
It wants to add an empty line into `python/sglang/srt/mem_cache/cpp_radix_tree/.clang-format`, which is not same as the main branch in sgl.

After adding this empty line, I rerun the precommit, and got the following error, which is same as CI lint error:
image

I copied the clang-format in sgl-kernel to `cpp_radix_tree`, then the lint error fix.
So, I push this commit. Do you have other methods to help solve the lint error?

python/sglang/srt/mem_cache/cpp_radix_tree/.clang-format
```diff
@@ -1 +1,15 @@
-../../../../../sgl-kernel/.clang-format
\ No newline at end of file
+BasedOnStyle: Google
```
mickqian: please remove this change

python/sglang/srt/multimodal/processors/points_v15_chat.py
```diff
@@ -0,0 +1,52 @@
...
+        *args,
+        **kwargs,
+    ):
+        base_output = self.load_mm_data(
+            prompt=input_text,
+            image_data=image_data,
+            multimodal_tokens=self.mm_tokens,
+        )
+
+        if base_output.images and isinstance(base_output.images[0], Image.Image):
```
mickqian, line 40: can we directly set Qwen2_5VLImageProcessor as the image processor for this one?

mickqian (changes requested):

python/sglang/srt/mem_cache/cpp_radix_tree/.clang-format
josephydu: When I try to fix lint error, I met some trouble. You can check my comment in this pr. Do you have any other good solutions to this problem?

python/sglang/srt/multimodal/processors/points_v15_chat.py
josephydu: Our processor is not same as Qwen2_5VLImageProcessor, cause we do not have rope index but qwen has.

python/sglang/srt/mem_cache/cpp_radix_tree/.clang-format
mickqian: No, I've never seen this problem before. Can you make sure your `pre-commit` is up-to-date?

josephydu: Hi, I have tried many times but still can't pass the pre-commit if I remove this change.
Can you give me som advice?

JustinTong0323 (approved):

python/sglang/srt/multimodal/processors/points_v15_chat.py
```diff
@@ -0,0 +1,52 @@
...
+from PIL import Image
+
+from sglang.srt.models.points_v15_chat import POINTSV15ChatModel
+from sglang.srt.multimodal.processors.qwen_vl import (
+    Qwen2_5VLImageProcessor,
+    resize_image_async,
+)
+
+
+class POINTSV15ChatProcessor(Qwen2_5VLImageProcessor):
```
mickqian, line 15: since they are basically the same, could you add your model as an arch to qwenimageprocessor?

yukavio: > since they are basically the same, could you add your model as an arch to qwenimageprocessor?

Currently, only the rope part is different from qwen, but the structure of the model may have some major changes in the future, so we hope to set up a separate file for easy maintenance.

mickqian (approved):

PR #9659, opened 2025-08-26
Fix None-Type comparison causing TypeError in schedule_batch.py

georg-wind:
## Motivation

Uncaught TypeError in schedule_batch.py if self.sampling_params.max_new_tokens = None.

Goal: prevent this from raising an Exception.
TypeError ("> between instances of Type None and int not allowed").

## Modifications

Condition now checks that self.sampling_params.max_new_tokens is not None (comes before 

Note: the statement executed if the condition resolves true, currently doesnt have any sensible effect given the statement in line 666 before (not addressed in this MR).

## Checklist

### DIFF
--- a/python/sglang/srt/managers/schedule_batch.py
+++ b/python/sglang/srt/managers/schedule_batch.py
@@ -665,7 +665,7 @@ class Req:
         # request has at least one token. Later, we can relax this requirement and use `input_len`.
         max_prefix_len = input_len - 1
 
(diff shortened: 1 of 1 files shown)


hnyls2002:
Thanks @georg-wind! That code path was rewritten and no longer does the None comparison, so the
  TypeError can't occur on main - closing. Please reopen if I've misread it.

PR #9663, opened 2025-08-26
feat: Add flexible validation for partial weight updates

GeLee-Q:
Co-authored-by: RichardW <rich-junwang@users.noreply.github.com> 
We will discard https://github.com/sgl-project/sglang/pull/9379 after merge this PR

## Motivation

The previous weight loading validation was overly strict, treating every load as a full model initialization. This caused legitimate partial updates (e.g., from RL training) to fail if they didn't include every model parameter.

This commit refactors the validation logic to be context-aware, intelligently distinguishing between two scenarios:

1.  **Partial Update (e.g., RL)**: The validation is lenient, only raising an error on complete failure (i.e., when 0 weights are successfully loaded despite being requested). This prevents silent failures.

2.  **Full Model Load**: The default behavior for larger weight sets. It maintains the original strict policy, ensuring no parameters are missing.
## Checklist


python/sglang/srt/models/gpt_oss.py
```diff
@@ -1115,11 +1115,23 @@ def _load_normal_weights(
...
+            if is_rl_update:
+                # RL update: Partial success is allowed, but complete failure (0 loaded) is an error.
+                num_params_updated = len(params_dict) - len(not_loaded_params)
+                if len(weights) > 0 and num_params_updated == 0:
+                    raise Exception("RL weight update failed completely. 0 weights were successfully loaded.")
+                logging.info(f"RL weight update complete. Updated {num_params_updated} parameters.")
             else:
-                logging.info("All parameters loaded successfully.")
+                # Full load: All parameters must be loaded, none can be missing.
+                if not_loaded_params:
```
hebiao064: I think we can just remove this check instead of adding Heuristic to determine whether it's RL or Serving workload, since other model implementations don’t include it.
What do you think, @Ying1123 ?

python/sglang/srt/models/gpt_oss.py
```diff
@@ -1115,11 +1115,23 @@ def _load_normal_weights(
                         logger.warning(f"Parameter {name} not found in params_dict")
 
         not_loaded_params = [k for k, v in params_checker.items() if not v]
+
         if tp_rank == 0:
-            if len(not_loaded_params) > 0:
-                raise Exception(f"Not all parameters loaded: {not_loaded_params}")
+            # Heuristic: If the number of requested weights is much smaller than the total
+            # model parameters (< 10%), we assume it's a dynamic RL update.
+            is_rl_update = (len(weights) < len(params_dict) * 0.1)
```
zhaochenyang20: Looks too hacky for this

zhaochenyang20 (changes requested):

python/sglang/srt/models/gpt_oss.py
GeLee-Q: I agree, removing this check seems like a good path forward. Differentiating between weights updates for RL and full updates for standard inference would likely add unnecessary complexity to the code and could degrade the performance of the weight update process.

zhaochenyang20:
@GeLee-Q I think that all the loading checks have been removed. What will happen if all the parameter updates fail? I do not see any check for this case, but your description said you have.

@PrinsYin

python/sglang/srt/models/gpt_oss.py
rich-junwang: I have a pr to remove this check: https://github.com/sgl-project/sglang/pull/9379

ispobock (approved):
LGTM. pls make sure gpt-oss related ci passed

python/sglang/srt/models/gpt_oss.py
GeLee-Q: > I have a pr to remove this check: #9379

Thanks a lot for your contribution! To ensure overall consistency with the codebase, we've made some further modifications and have included you as a co-author in the initial commit.

zhaochenyang20 (approved):
LGTM

PR #9692, opened 2025-08-27
feat(hicache-3fs): 3FS-Store Backup Optimizations For MLA Model.

hzh0425:
## Motivation

3FS-Store Backup Optimizations For MLA Model.

## Checklist


xiezhq-hermann (approved):

xiezhq-hermann:
I now think the change in the storage backend might not be necessary. We can just skip the put operation in the cache controller like this as in my latest PR https://github.com/sgl-project/sglang/pull/9797.
```
    def backup_thread_func(self):
        """
        Manage backup operations from host memory to storage backend.
        """
        while not self.stop_event.is_set():
            try:
                operation = self.backup_queue.get(block=True, timeout=1)
                if operation is None:
                    continue

                if not self.backup_skip:
                    self._page_backup(operation)
                self.ack_backup_queue.put(operation.id)

            except Empty:
                continue
```

PR #9712, opened 2025-08-27
[ModelOpt] Fix Weight Loading for DSR1-FP4 Quantization

pavanimajety:
## Motivation
A simple fix to load new style of DeepSeek R1 FP4 Models with a new hf_quant_config.json specification where all layers are explicitly specified in exclude_modules. 

## Modifications

Changes the exclude_modes logic slightly to accomodate loading of `*qkv_a_proj*` weights

## Accuracy Tests

N/A Doesn't modify the existing path. 

## Benchmarking and Profiling
N/A

## Checklist

- [N/A] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [N/A] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
- [N/A] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.ai/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.ai/developer_guide/contribution_guide.html#benchmark-the-speed).


trevor-m (approved):
LGTM

Edwardf0t1 (approved):
Thanks @pavanimajety for the fix. We didn't expect our new config could break weight loading of fp4 ckpts in SGLang.

yiakwy-xpu-ml-framework-team:
The previous ModelOpt quant method to mxfp4 from bf16 python script is extremely slow (no kernel, just pytorch codes to select closest value from 15 fp4 candidates). Which version of code are we using ?

pavanimajety:
@yiakwy-xpu-ml-framework-team This is for nvfp4 models. Could you provide more code pointer into what is extremely slow? This PR changes weight loading logic for quantized linear layers.

python/sglang/srt/layers/quantization/modelopt_quant.py
```diff
@@ -599,6 +599,13 @@ def is_layer_excluded(self, prefix: str, exclude_modules: list):
             regex_str = pattern.replace(".", r"\.").replace("*", r".*")
             if re.fullmatch(regex_str, prefix):
                 return True
+
+            # Check if the last part of the excluded pattern is contained in the last part of the prefix
+            # This handles fused modules like fused_qkv_a_proj_with_mqa that contain q_a_proj and kv_a_proj_with_mqa
+            pattern_last_part = pattern.split(".")[-1]
+            prefix_last_part = prefix.split(".")[-1]
+            if pattern_last_part in prefix_last_part:
+                return True
```
donglinz, line 608: When ```pattern_last_part``` is like ```model.layers.34.mlp.gate``` and ```prefix_last_part``` is like ```model.layers.1.mlp.gate_up_proj``` this could false positively into the if condition for nvidia/DeepSeek-R1-FP4 and result in not quantizing up gate in the first three layer.

This will get you something like:

```
File "/sgl-workspace/sglang/python/sglang/srt/model_loader/loader.py", line 476, in load_weights_and_postprocess
    model.load_weights(weights)
  File "/sgl-workspace/sglang/python/sglang/srt/models/deepseek_v2.py", line 2598, in load_weights
    param = params_dict[name]

KeyError: 'model.layers.0.mlp.gate_up_proj.input_scale'
```

pavanimajety: We only check the patterns in excluded_modules which are loaded from the checkpoint: https://huggingface.co/nvidia/DeepSeek-R1-FP4/blob/main/hf_quant_config.json
 
 I don't see a match for `model.layers.0.mlp` here. I also tested the model on main and don't see any issues. Could you confirm the reproduction steps for this? Thanks!

donglinz: yeah model.layers.*.mlp.gate is in the excluded_modules.

I am testing with ```lmsysorg/sglang:v0.5.2rc0-cu128-b200```

```
python3 -m sglang.launch_server --model-path nvidia/DeepSeek-R1-FP4  --port 30001     --dist-init-addr 127.0.0.1:40001     --tp-size 4
```

pavanimajety: Thanks for the repro, fix here: https://github.com/sgl-project/sglang/pull/9940

PR #9741, opened 2025-08-28
[Feature] LMCache Connector Integration

Oasis-Git:
## Motivation

LMCache is an LLM serving engine extension to reduce TTFT and increase throughput, especially under long-context scenarios. By storing the KV caches of reusable texts across various locations, including (GPU, CPU DRAM, Local Disk), Thus, LMCache saves precious GPU cycles and reduces user response delay. For more details about LMCache, please refer to: https://lmcache.ai.

In this PR we provide LMCache connector for SGLang engine. Table below list some key differences compared with existing Hierarchical Cache design.

| Framework       | Hierarchical Cache | LMCache   |
| :-------------- | :----------------- | :-------- |
| **Layout** | Tree Base          | Hash Base |
| **G/CPU exclude** | True               | False     |
| **Load/Store Thread** | True               | False     |

Moreover, LMCache offers flexible KV cache management with features like compression, PD disaggregation, and support for multiple storage backends, including Mooncake and Redis.

## Modifications

Key modifications were made to `python/sglang/srt/mem_cache/lmc_radix_cache.py` to optimize KV cache management.

The `match_prefix` function now handles loading the KV cache from the LMCache backend when part of prefix is missed in the GPU radix cache. Conversely, the `cache_finished_req` function stores the KV cache upon request completion.

To improve performance, these I/O operations are overlapped with computation. Specifically, the KV cache for layer `I+1` is loaded concurrently with the computation for layer `i`. Similarly, the cache storing operation for a completed request is overlapped with the computation of the next batch.

## Accuracy Tests

The accuracy test is ensured in `python/sglang/srt/mem_cache/storage/lmcache/unit_test.py`

## Benchmarking and Profiling

We provide the benchmark result below. The benchmark is done under settings:
```
H100 TP=2
Qwen3-32B model
benchmark script: sglang/hicache/bench_multiturn.py
```

image

The `WoLW` label refers to the LMCache connector version that does not include the layer-wise mechanism. The slight difference under high qps is mainly caused by GPU-CPU kvcache exclusion mechanism.

## Checklist


python/sglang/srt/mem_cache/memory_pool.py
```diff
@@ -352,7 +356,8 @@ def get_key_buffer(self, layer_id: int):
         # same applies to get_value_buffer and get_kv_buffer
         if self.layer_transfer_counter is not None:
             self.layer_transfer_counter.wait_until(layer_id - self.start_layer)
-
+        elif self.layer_transfer_executor is not None:
```
xiezhq-hermann: can we combine the newly introduced `layer_transfer_executor` with `layer_transfer_counter`? while the implementation differs, the interface is actually almost the same

python/sglang/srt/mem_cache/lmc_radix_cache.py
```diff
@@ -0,0 +1,738 @@
...
+
+    def wait_until(self, layer_id):
+        # if self.printable:
+        #     print(f"Waiting for layer {layer_id}")
+        self.load_stream.synchronize()
+        with self.load_stream:
+            self.lmc_connector.load_kv_layerwise(layer_id)
+
+
+class LMCRadixCache(BasePrefixCache):
```
xiezhq-hermann: I see the value of having the whole radix tree instance managed separately, but this indeed duplicated the code and might lead to divergent behaviors, it would be better to reuse the code in radix tree and only implement the different functions like match_prefix.

xiezhq-hermann:
Thank you @Oasis-Git for finally made the PR and the performance looks good. I think the overall design makes sense, but it would benefit from removing duplicated codes, as managing the radix tree whole by yourself could actually lead to divergent behaviors and future maintenance problem. You actually only need to have a few functions implemented by inherenting radix tree, and I think would be better to move the lmcache_radix into the lmcache repo as well, so you can easily make editions into it later.

xiezhq-hermann:
One thing I am not very certain about is how does it handle TP especially with storage layers enabled. For example, what if rank 0 matched and fetched the result on disk, but rank 1 failed to fetch it? There does not seem to have a synchronization to protect it.

python/sglang/srt/mem_cache/memory_pool.py
Oasis-Git: The code has been factored.

python/sglang/srt/mem_cache/lmc_radix_cache.py
Oasis-Git: The code has been factored.

Oasis-Git:
To improve integration with the base radix cache design, the code has been refactored. The logic within the sglang component was simplified as much as possible with most responsibilities now centralized in the LMCache. This approach is intended to minimize the need for frequent changes in sglang.

python/sglang/srt/server_args.py
```diff
@@ -2109,6 +2113,14 @@ def add_cli_args(parser: argparse.ArgumentParser):
             help="(Deprecated) Enable FlashInfer MXFP4 MoE backend for modelopt_fp4 quant on Blackwell.",
         )
 
+        # LMCache
+        logger.info("Adding LMCache arguments")
```
xiezhq-hermann: no need for this, you can have a logger.info for LMCache Radix creation though.

python/sglang/srt/mem_cache/lmc_radix_cache.py
```diff
@@ -0,0 +1,280 @@
...
+            num_retrieved = self.lmcache_connector.start_load_kv(
+                LoadMetadata(
+                    token_ids=key,  # full page-aligned key
+                    slot_mapping=slot_mapping,
+                    offset=value.numel() - prefix_pad,  # LMCache offset convention
+                )
+            )
+        logger.debug("num_retrieved_tokens: %s", num_retrieved)
+
+        if num_retrieved > 0:
```
xiezhq-hermann: I think you need sync across TP workers to ensure consistent change to the radix tree.

python/sglang/srt/mem_cache/lmc_radix_cache.py
```diff
@@ -0,0 +1,280 @@
...
+
+    def evict(self, num_tokens: int) -> None:  # type: ignore[override]
+        """Before base eviction, wait for any outstanding stores and release locks."""
+        if self.disable:
+            return
+
+        self.store_stream.synchronize()
+        with self._node_lock:
+            for node in self._in_flight_nodes:
+                self.dec_lock_ref(node)
```
xiezhq-hermann: there would be a risk of inconsistence across workers of different TP ranks

python/sglang/srt/server_args.py
Oasis-Git: resolved

python/sglang/srt/mem_cache/lmc_radix_cache.py
Oasis-Git: To ensure consistency, LMCache synchronizes token retrieval across all Tensor Parallel (TP) engines. During the retrieve operation, LMCache first looks up tokens on each engine individually. It then uses an all-reduce operation to find the minimum number of tokens retrieved across all engines, guaranteeing that every process agrees on and uses the same token count.

For more details, see the code in https://github.com/LMCache/LMCache/pull/1510

python/sglang/srt/mem_cache/lmc_radix_cache.py
Oasis-Git: After discussion we think that there is no risk for radix tree inconsistence in this part.

xiezhq-hermann (approved):

python/sglang/srt/mem_cache/lmc_radix_cache.py
```diff
@@ -0,0 +1,280 @@
+from __future__ import annotations
```
xiezhq-hermann: consider move this file to `storage/lmcache` as well

jianyingzhu:
I ran the command:
`python -m sglang.launch_server --model /workspace/Qwen3-8B --trust-remote-code  --host 0.0.0.0 --port 30000 --mem-fraction-static 0.7 --chunked-prefill-size 16384 --max-running-requests 256 --enable-metrics --enable-lmcache`

And reported an error:

  File "/sgl-workspace/sglang/python/sglang/srt/managers/scheduler.py", line 2787, in run_scheduler_process
    scheduler = Scheduler(
                ^^^^^^^^^^
  File "/sgl-workspace/sglang/python/sglang/srt/managers/scheduler.py", line 444, in __init__
    self.init_memory_pool_and_cache()
  File "/sgl-workspace/sglang/python/sglang/srt/managers/scheduler.py", line 713, in init_memory_pool_and_cache
    self.tree_cache = LMCRadixCache(
                      ^^^^^^^^^^^^^^
  File "/sgl-workspace/sglang/python/sglang/srt/mem_cache/storage/lmcache/lmc_radix_cache.py", line 91, in __init__
    self.lmcache_connector = LMCacheLayerwiseConnector(
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: LMCacheLayerwiseConnector.__init__() got an unexpected keyword argument 'tp_group'

Is there any LMCache version that contains 'tp_group’?

PR #8840, opened 2025-08-06
[Generative Score API] Optimization to Remove Decode.

sundar24295s:
## 🚀 Motivation

- This PR is a follow-up to the Decoder-only Scoring API introduced in [PR #6460](https://github.com/sgl-project/sglang/pull/6460), which was initially proposed and discussed in [Issue #5973](https://github.com/sgl-project/sglang/issues/5973).
- Achieved a **40% reduction in single-request latency** for a 300-token input (120 query tokens + 180 item tokens), decreasing from 33 ms to 20 ms.

## 🔧 Modifications

- **Remote Decode Optimization**:  
  Scoring workloads require only log probabilities at last token position and do not involve token generation. This PR removes the decode phase entirely for such workloads, resulting in significantly lower latency and improved throughput.
- This optimization applies to all prefill-only workloads.

## 📈 Future Optimizations

This refactor lays the foundation for further improvements, including:

- **Memory Transfer Optimization**:  
  Move all CPU-GPU synchronization to the post-processing loop to better overlap `run_batch` with post-processing.

- **Multi-Item Scoring**:  
  Support scoring multiple items within a single prompt using custom attention masks via FlashInfer.  
  → See [flashinfer-ai/flashinfer#1015](https://github.com/flashinfer-ai/flashinfer/pull/1015)

## Accuracy
- Scores before this change
```
jobuser [ /shared/user/repos3/sglang ]$ curl -X POST "http://localhost:30000/v1/score"   -H "Content-Type: application/json"   -d '{
    "query": "What is the capital of California? Answer Yes or No for each of the following options:",
    "items": ["Scaramento", "San Jose", "San Francisco"],
    "label_token_ids": [9454, 2753],
    "model": "/shared/public/sharing/job-rank/kbehdin/f389cde308efd4dbb8d9-2025-06-06-18-31-30/best_model/epoch=0-step=498-HF"
  }'
{"scores":[[4.234364670752685e-06,1.2348638303110892e-05],[7.162677269222586e-05,0.0003160321422383422],[0.0001203001937164321,0.00030480667807191645]],"model":"/shared/public/sharing/job-rank/kbehdin/f389cde308efd4dbb8d9-2025-06-06-18-31-30/best_model/epoch=0-step=498-HF","usage":null,"object":"scoring"}
```

- Scores after this change
```
(sglang) jobuser [ /shared/user/repos3/sglang ]$ curl -X POST "http://localhost:30000/v1/score"   -H "Content-Type: application/json"   -d '{
[...]

PR #9748, opened 2025-08-28
[Generative Score API] Scoring(Prefill-only) optimizations.

sundar24295s:
##  🚀 Motivation
- Follow-up to [PR #8840](https://github.com/sgl-project/sglang/pull/8840) to reduce latency and increase throughput for the generative score API.
- Scoring only needs the next-token distribution after the full prompt, not per-token logprobs inside the prompt or any sampling.

**Performance Impact:**
- On Qwen3-0.6B with 300 input tokens, at QPS 100 and 10 items per request, P99 latency improved from **6220 ms to 454 ms** (~**13.7× faster**, ~**92.7% reduction**) with this PR.
- With a P99 latency threshold of 500 ms, throughput increased from **800 to 1000 items/s per H100 GPU** (~**25% increase**).

image

### ⚡ Optimization 1: Skip Input Token Logprobs Computation
For scoring requests like:
```python
# Item 1:
full_prompt = "What is the capital of California? Answer Yes or No for each of the following options: Sacramento"
# Item 2: 
full_prompt = "What is the capital of California? Answer Yes or No for each of the following options: San Jose"
```
- We do not need input token logprobs like P(California | What is the capital of), rather only need the next-token distribution after the full prompt namely P(Yes | full_prompt) and P(No | full_prompt).

### ⚡ Optimization 2: Skip Sampling Step  
- Added `Sampler.compute_logprobs_only()` to compute logprobs without sampling for prefill-only scoring:

### ⚡ Optimization 3: Delayed GPU→CPU Copy with Overlap
- Added `get_token_ids_logprobs_batch_optimized(logprobs, token_ids, delay_cpu_copy=True)`, Single vectorized gather on GPU for the entire batch. Optionally defers .tolist() until result processing, improving overlap with the next batch’s compute.

## Accuracy Tests
- Scores before this PR
```
$ curl -X POST "http://localhost:30000/v1/score"   -H "Content-Type: application/json"   -d '{
    "query": "What is the capital of California? Answer Yes or No for each of the following options:",
    "items": ["Scaramento", "San Jose", "San Francisco"],
    "label_token_ids": [9454, 2753],
    "model": "/shared/public/elr-models/Qwen/Qwen3-0.6B/c1899de289a04d12100db370d81485cdf75e47ca"           
  }'
{"scores":[[4.234364670752685e-06,1.2348638303110892e-05],[7.162677269222586e-05,0.0003160321422383422],[0.0001203001937164321,0.00030480667807191645]],"model":"/shared/public/elr-models/Qwen/Qwen3-0.6B/c1899de289a04d12100db370d81485cdf75e47ca","usage":null,"object":"scoring"}
```

- Scores after this PR
```
$ curl -X POST "http://localhost:30000/v1/score"   -H "Content-Type: application/json"   -d '{
    "query": "What is the capital of California? Answer Yes or No for each of the following options:",
    "items": ["Scaramento", "San Jose", "San Francisco"],
    "label_token_ids": [9454, 2753],
    "model": "/shared/public/elr-models/Qwen/Qwen3-0.6B/c1899de289a04d12100db370d81485cdf75e47ca"           
  }'
{"scores":[[4.234364670752685e-06,1.2348638303110892e-05],[7.162677269222586e-05,0.0003160321422383422],[0.0001203001937164321,0.00030480667807191645]],"model":"/shared/public/elr-models/Qwen/Qwen3-0.6B/c1899de289a04d12100db370d81485cdf75e47ca","usage":null,"object":"scoring"}
```

## 🧪 Benchmark Comparison: Qwen3-0.6B on H100 (CUDA 12.8)

**Setup**:  
- Model: [Qwen3-0.6B](https://huggingface.co/Qwen/Qwen3-0.6B)
- Prompt length: 300 tokens  
- Hardware: H100 GPU  
- Duration: 120s  
- Target RPS: 70 , 80, 90, 100
- Item Count: 10  per request
- Distribution: Poisson  

**Server Start**:
```
(sglang) jobuser [ /shared/user/repos3/sglang ]$ python -m sglang.launch_server --model-path Qwen/Qwen3-0.6B --port 30000 --host 0.0.0.0 --chunked-prefill-size -1 --enable-torch-compile --dtype float16 --max-prefill-tokens 31000 --mem-fraction-static 0.5 --enable-tokenizer-batch-encode --disable-radix-cache --disable-cuda-graph
```

**Benchmark Script**:
```
python3.10 sglang/benchmark/score/bench_score.py
```

### 🔍 Summary of Improvement

| Items Per Second | Baseline P99 Latency (ms) | This PR P99 Latency (ms) |
|------------------|---------------------------|--------------------------|
| 600              | 226.00                    | 139.16                   |
| 700              | 282.21                    | 193.78                   |
| 800              | 413.14                    | 227.20                   |
| 900              | 1200.72                   | 302.39                   |
| 1000             | 6220.00                   | 454.20                   |
| 1100             | 8694.97                   | 1459.81                  |
| 1200             | 11606.46                  | 6406.18                  |

---

## Profiling
- Current baseline profile showing sampling, logits extraction causing memsyncs and delay in next batch scheduling:
Screenshot 2025-08-28 at 1 47 07 AM

---
- Profile after optimizations showing almost zero gap in GPU kernel launches between batches under high load.
Screenshot 2025-08-28 at 1 50 35 AM

## Checklist


fortunecookiee (approved):
LGTM!

python/sglang/srt/layers/sampler.py
```diff
@@ -164,6 +164,65 @@ def forward(
...
         return batch_next_token_ids
 
+    def compute_logprobs_only(
+        self,
+        logits_output: LogitsProcessorOutput,
+        sampling_info: SamplingBatchInfo,
+        return_logprob: bool,
+        top_logprobs_nums: List[int],
+        token_ids_logprobs: List[List[int]],
+    ) -> None:
```
chanh, line 184: this method is duplicating a lot of logic that i presume also lives in another method. can we write in a way that doesn't introduce that duplication?

python/sglang/srt/layers/sampler.py
```diff
@@ -233,6 +292,75 @@ def get_top_logprobs(
...
+            end_pos = start_pos + num_tokens
+            request_logprobs = batch_logprobs[start_pos:end_pos]
+
+            # Handle delayed vs immediate CPU copy
+            if delay_cpu_copy:
+                # Keep GPU tensor for later conversion during output processing
+                output_token_ids_logprobs_val[original_idx] = request_logprobs
+            else:
+                # Immediate CPU copy
+                output_token_ids_logprobs_val[original_idx] = request_logprobs.tolist()
```
chanh: do we really need delay vs immediate? can simplify and just make it always be delayed.

python/sglang/srt/layers/sampler.py
```diff
@@ -233,6 +292,75 @@ def get_top_logprobs(
...
+            batch_col_indices, device=logprobs.device, dtype=torch.long
+        )
+
+        # Extract all required logprobs in one kernel call
+        batch_logprobs = logprobs[batch_row_tensor, batch_col_tensor]
+
+        # Split results back into per-request format
+        for original_idx, start_pos, num_tokens in request_info:
+            end_pos = start_pos + num_tokens
+            request_logprobs = batch_logprobs[start_pos:end_pos]
```
chanh: This could be made cleaner with torch operations. E.g. this is a draft that chatgppt gave me

```
    lens_t = torch.tensor([len(ids or []) for ids in token_ids_logprobs], device=dev)
    total = int(lens_t.sum().item())
    if total == 0:
        return [logprobs.new_empty(0) for _ in token_ids_logprobs]

    # Build flattened row/col indices on the SAME device as logprobs
    rows = torch.repeat_interleave(torch.arange(B, device=dev), lens_t)
    cols = torch.tensor([j for ids in token_ids_logprobs for j in (ids or [])],
                        device=dev, dtype=torch.long)

    vals = logprobs[rows, cols]  # shape: [total]

    # Split back per row (handles zeros safely)
    parts = torch.split_with_sizes(vals, lens_t.tolist(), dim=0)
    ```

File: python/sglang/srt/managers/schedule_batch.py [hunk 4]
```diff
@@ -1208,11 +1211,19 @@ def prepare_for_extend(self):
...
+                    self.is_prefill_only
+                    and req.logprob_start_len == len(req.origin_input_ids) - 1
+                ):
+                    req.extend_logprob_start_len = req.extend_input_len
+                else:
+                    req.extend_logprob_start_len = min(
+                        req.logprob_start_len - pre_len,
+                        req.extend_input_len,
+                        req.seqlen - 1,
+                    )
```
chanh: can you explain what this has to do with prefill only? 

logic here is a little not obvious -- maybe we can discuss in person

python/sglang/srt/layers/sampler.py
sundar24295s: Refactored the `logits` validation and pre-processing to common helper function, other than that no code duplication.

python/sglang/srt/layers/sampler.py
sundar24295s: - Yeah we don't need this flag right now as in scores computation we wanted to delay the copy to cpu always.
- We can add this back if we later decide to eeplace get_token_ids_logprobs() with the batch optimized version in forward().

python/sglang/srt/managers/schedule_batch.py
sundar24295s: I have added some detailed comments in code. 
```
            # - logprob_start_len: Absolute position in full sequence where logprob computation begins
            # - extend_logprob_start_len: Relative position within current extend batch where logprob computation begins
            # - extend_input_len: Number of tokens that need to be processed in this extend batch
            #   (= len(fill_ids) - len(prefix_indices), where fill_ids = origin_input_ids + output_ids
            #    and prefix_indices are the cached/shared prefix tokens)
```

Basically.

- We intentionally avoid introducing “scoring” terminology here. Instead, we treat scoring requests as prefill-only and rely on the default logprob_start_len = -1, which means only output-token logprobs are needed.

- In the scheduler, there’s existing logic that resets `req.logprob_start_len = len(req.origin_input_ids) - 1` whenever the value is -1. For prefill-only requests, we override this to `req.logprob_start_len = len(req.origin_input_ids)` so that no extra input token logprob is computed.

- The existing logic for `req.extend_logprob_start_len` doesn’twork for prefill-only scoring requests, since the `req.seqlen - 1` constraint always forces computation of the last input token, defeating our optimization.
```
req.extend_logprob_start_len = min(
    req.logprob_start_len - pre_len,  # Convert to relative position
    req.extend_input_len,             # Cap by extend batch size
    req.seqlen - 1,                   # Cap by sequence length - 1
```

python/sglang/srt/layers/sampler.py
sundar24295s: Nice, thanks replaced applicable computations with corresponding torch operations.

python/sglang/srt/managers/schedule_batch.py
sundar24295s: - We’re essentially treating Scoring requests as prefill-only with return_logprob=True and [logprob_start_len=-1](https://github.com/sgl-project/sglang/blob/19d64f2b725889cfbdb000937a2d57c07db5cfa8/python/sglang/srt/managers/io_struct.py#L91)(default, meaning only output token logprobs are returned).


```
        batch_request = GenerateReqInput(
            token_ids_logprob=label_token_ids,
            return_logprob=True,
            logprob_start_len=-1, #default value is -1
            stream=False,
            sampling_params={"max_new_tokens": 0},
        )
```
- To skip input logprob computation, we have two options: (1) Continue using the existing logprob_start_len param, (2) Introduce an explicit skip_input_token_logprobs flag.
```
        batch_request = GenerateReqInput(
            token_ids_logprob=label_token_ids,
            return_logprob=True,
            skip_input_token_logprobs=True,
            stream=False,
            sampling_params={"max_new_tokens": 0},
        )
```
Personally, I think the flag is redundant since logprob_start_len already covers this. But happy to add a new flag as well. @hnyls2002 / @zhyncs thoughts?
Callout: All these changes are limited to the score/prefill-only path and won’t affect the generate path.

hnyls2002: Yes, I think the current design is good.

hnyls2002 (approved):

sundar24295s:
Moved the benchmarking scripts to a separate [PR](https://github.com/sgl-project/sglang/pull/10240) will rebase on this.

PR #9755, opened 2025-08-28
[feat] Add P/D attention select for draft model

Ximingwang-09:
## Motivation
https://github.com/sgl-project/sglang/pull/9573 has already enabled compatibility of Hybrid Attention in Speculative Decoding. However, for the draft model, the functionality to select the attention backend separately for the extend mode and decode mode has not been fully implemented—this PR supplements and completes the implementation of this feature.

### Launch the server
```
python3 -m sglang.launch_server --model /mnt/Qwen3-8B --trust-remote-code --tp-size 8 --enable-cache-report --dtype bfloat16 --log-level info  --max-running-requests 16 --mem-fraction-static 0.85 --host 0.0.0.0 --port 9122 --speculative-algorithm EAGLE3 --speculative-num-steps 5 --speculative-eagle-topk 2 --speculative-num-draft-tokens 8 --speculative-draft-model-path /mnt/qwen3_8b_eagle3 --prefill-attention-backend fa3 --decode-attention-backend flashinfer
```

### Test accuracy
```
python3 benchmark/gsm8k/bench_sglang.py  --num-shots 8 --num-questions 1319 --parallel 1319 --port 9122
```
Accuracy: 0.901
## Checklist


Qiaolin-Yu (commented):
For the multiple  `_create_xxx_decode_backend` functions, would writing it following the [code](https://github.com/sgl-project/sglang/blob/c112bcc46168f9e69b4dd0ff3ff053fee4539605/python/sglang/srt/model_executor/model_runner.py#L1433-L1586) here makes the implementation simpler?

Ximingwang-09:
> For the multiple `_create_xxx_decode_backend` functions, would writing it following the [code](https://github.com/sgl-project/sglang/blob/c112bcc46168f9e69b4dd0ff3ff053fee4539605/python/sglang/srt/model_executor/model_runner.py#L1433-L1586) here makes the implementation simpler?

Thanks for the suggestion! I think multiple `_create_xxx_decode_backend` functions avoid large if/else that mixes multiple backend details together. And keeps the main method short and focused. I currently feel the separate methods give us cleaner structure as the number of backends grows.That said, do you think the potential clarity of a single if/else still outweighs these benefits in our case?

Qiaolin-Yu (approved):
lgtm. wait for ci.

Qiaolin-Yu:
> > For the multiple `_create_xxx_decode_backend` functions, would writing it following the [code](https://github.com/sgl-project/sglang/blob/c112bcc46168f9e69b4dd0ff3ff053fee4539605/python/sglang/srt/model_executor/model_runner.py#L1433-L1586) here makes the implementation simpler?
> 
> Thanks for the suggestion! I think multiple `_create_xxx_decode_backend` functions avoid large if/else that mixes multiple backend details together. And keeps the main method short and focused. I currently feel the separate methods give us cleaner structure as the number of backends grows.That said, do you think the potential clarity of a single if/else still outweighs these benefits in our case?

hmm. I think maybe using if/else here could make it easier to read, but just a nit.

PR #9764, opened 2025-08-28
[NVIDIA] Remove unused `get_fused_moe_impl_class` function

kaixih:
While checking the MTP MoE, I noticed this function is never used in the repo. We should remove `get_fused_moe_impl_class` to avoid confusion with the correct function: `get_moe_impl_class`.

### DIFF
--- a/python/sglang/srt/layers/quantization/fp8.py
+++ b/python/sglang/srt/layers/quantization/fp8.py
@@ -619,11 +619,7 @@ class Fp8MoEMethod(FusedMoEMethodBase):
             layer.register_parameter("w13_weight_scale_inv", w13_weight_scale)
             layer.register_parameter("w2_weight_scale_inv", w2_weight_scale)
             assert self.quant_config.activation_scheme == "dynamic"
-            if (
-                get_bool_env_var("SGLANG_CUTLASS_MOE")
-                and self.cutlass_fp8_supported
-                and (is_sm100_supported() or is_sm90_supported())
-            ):
+            if self.use_cutlass_fused_experts_fp8:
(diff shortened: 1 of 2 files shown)


kaixih:
@elfiegg can you take a look? since I also make a minor change to fp8 path.

elfiegg (approved):
LGTM

python/sglang/srt/layers/quantization/fp8.py
```diff
@@ -619,11 +619,7 @@ def create_weights(
             layer.register_parameter("w13_weight_scale_inv", w13_weight_scale)
             layer.register_parameter("w2_weight_scale_inv", w2_weight_scale)
             assert self.quant_config.activation_scheme == "dynamic"
-            if (
-                get_bool_env_var("SGLANG_CUTLASS_MOE")
-                and self.cutlass_fp8_supported
-                and (is_sm100_supported() or is_sm90_supported())
-            ):
+            if self.use_cutlass_fused_experts_fp8:
```
zhyncs, line 622: do we need block_quant here?

kaixih: I think this is all for the deepseek styled fp8, so the block_quant is a must. So, ideally, we should check it. @elfiegg right?

elfiegg: we do, block_quant checker is already in the file
https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/layers/quantization/fp8.py#L521

zhyncs (approved):

PR #9800, opened 2025-08-29
[Interface] Add an alias for `sglang launch server`

vincentzed:
## Motivation

@zhyncs for review

## Modifications
Do it minimal as possible

After installing from the `python/` directory (e.g., pip install -e python), you can launch the server via: `sglang --model-path dir --port 30000 --host 0.0.0.0
Nothing else changes.

## Checklist

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -167,4 +167,4 @@ skip = "*.json,*.jsonl,*.patch,*.txt"
 
 
 [project.scripts]
-sglang = "sglang.launch_server:main"
+sglang = "sglang.sglang:main"
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -164,3 +164,7 @@ exclude = [
 [tool.codespell]
 ignore-words-list = "ans, als, hel, boostrap, childs, te, vas, hsa, ment"
 skip = "*.json,*.jsonl,*.patch,*.txt"
+
+
+[project.scripts]
+sglang = "sglang.launch_server:main"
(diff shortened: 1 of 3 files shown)


zhyncs:
cool

zhyncs:
how about
`sglang serve`
`sgl serve`
`sglang launch server`

JustinTong0323:
Thanks Vincent! It's actually a really good minimum CLI entrypoint, no doubt about that. But when we think about the future extensibility we could have, maybe we could try some modern CLI packages. Take https://click.palletsprojects.com/en/stable/ for example.

vincentzed:
I agree with the click take, I thought of wrapping custom argparse utils, but it's a lot of code we have to maintain + test, seems like a simple solution. If we can come to consensus happy to use click or something similar. Also if you haven't seen it this is pretty cool (kinda terminal gui inspired, but too complex to integrate. https://github.com/Chen-zexi/vllm-cli)

JustinTong0323:
> I agree with the click take, I thought of wrapping custom argparse utils, but it's a lot of code we have to maintain + test, seems like a simple solution. If we can come to consensus happy to use click or something similar. Also if you haven't seen it this is pretty cool (kinda terminal gui inspired, but too complex to integrate. https://github.com/Chen-zexi/vllm-cli)

Yes, I think we could potentially create an RFC to utilize click for refactoring the CLI. This could attract contributions from community members, allowing us to collaborate on the implementation. Could you help draft one? I can also write one later if you're busy.

And yes, I've seen the vLLM CLI, and I think it's really cool. We're happy to see similar projects from the community and could help add a reference to [awesome-sglang](https://github.com/sgl-project/awesome-sglang), but it might be somewhat heavy to include directly in the main sglang repository (I believe that is mainly for developer testing and is not suitable for production use).

vincentzed:
@JustinTong0323 Do you still think we should implement the full CLI or not? This is pretty small and minimal, but if you want to save it to have the full design with click then that works too

zhyncs:
We need to discuss the CLI interface. I will close it for now. Thank you for your contribution.

PR #9803, opened 2025-08-29
[router] Fix short timeout for the prefill client

LukasBluebaum:
## Motivation

Currently, the timeout for the prefill client in the `return_logprobs = False` path is set to only 2 secs. This leads to many premature request aborts as soon as some load is on the server: #9266 . Here I also set the `request_timeout_secs` on the prefill client in that path. It might also make sense to decouple this and have an extra timeout variable for the prefill client.

I think in a future PR further improvements could be to cancel the decode in this path immediately if the prefill fails for some reason and also to maybe set a timeout on the background drains too.

## Modifications

- Propagate `request_timeout_secs` to the prefill client
- Some renaming for timeout and interval variables for clarity

## Checklist


finetunej:
Great find, this fixes the largest part of the issue!

The additional optimizations would certainly be good to have, but this makes it usable.

LukasBluebaum:
So the reason the test breaks now, is that after the timeout change the test processes ~200 requests in the 120sec while it only processed ~100 requests before. So some metrics are worse. 

However, I did not have time yet to look into why exactly this happens. Will do that tomorrow or on Monday.

LukasBluebaum:
So I think the timeout changes are working fine. The problem is that aborted requests in the prefill do not get immediately propagated to the decode instance. What was happening in this test with the 2 seconds timeout before, was that the router correctly times out the connection to the prefill and the prefill then aborts the request. However, due to them using streaming the prefill and decode both already send a 200 to the router (and the router to genai bench). So both the router and genai-bench were waiting on the decode to return something.

  In genai-bench this resulted in many of the 64 users being stuck and not doing anything. Others went through and could keep running, because of the lower concurrency now the timeout of 2 seconds was fine and no more requests were aborted. With the higher timeout now, we always have the full concurrency resulting in more requests finishing in the 120 second test window and in worse metrics per request as expected. I adjusted the thresholds in the test accordingly.
  
    I also added some commits here where I made a suggestion to solve this issue so that aborted requests are propagated from the prefill to the decode. Then the decode will also abort and write an SSE error chunk to the stream. Running the test then with the 2 sec timeout correctly results in many errors in genai-bench, where barely any requests get through.
    
 If you want this change I am also happy to move it to another PR and also add the other direction that aborted decodes inform the prefill. Also need to think a bit more about the additional check in the `decode_thread`, which is needed because there seems to be a race condition now in the case where we have a small timeout (like the 2 seconds) and the client abruptly disconnects while many requests are still inflight.
 
 Other changes like an early fail on prefill failures in the router would still make sense I think.

sgl-router/src/routers/http/router.rs
```diff
@@ -52,8 +52,8 @@ impl Router {
         worker_urls: Vec<String>,
         policy: Arc<dyn LoadBalancingPolicy>,
         client: Client,
-        timeout_secs: u64,
-        interval_secs: u64,
+        worker_startup_timeout_secs: u64,
```
slin1237, line 55: nice

python/sglang/srt/disaggregation/mooncake/conn.py
```diff
@@ -1197,6 +1203,22 @@ def abort(self):
...
+                self.bootstrap_room
+            ].values():
+                self.kv_mgr.sync_status_to_decode_endpoint(
+                    transfer_info.endpoint,
+                    transfer_info.dst_port,
+                    transfer_info.room,
+                    KVPoll.Failed,
+                    local_rank,
+                )
+
```
ShangmingCai: I think we should not sync failed status to the decode endpoint here, it will couple the liveness of P and D, which might cause potential problems when actively killing some nodes for scaling. A better solution is when a req failed at the prefill node, the router gets notified and should send an abort req to the decode node.

LukasBluebaum: Makes sense, I reverted the changes on the branch. The one downside I thought about is that this relies on the router to implement this behavior, also when a different router than the SGLang one is used. But because it does not affect correctness and mainly leads to a slower timeout, if not implemented, it should be fine.

slin1237 (approved):

ShangmingCai (approved):
LGTM

PR #8390, opened 2025-07-26
fix fp8 update_weights for block_quant

GuoweiWangU:
## Motivation
After sglang successfully deploys the fp8 block quant model, an error will be reported when calling the update_weights related interface.
```
File "/workspace/bin/sglang/python/sglang/srt/model_loader/weight_utils.py", line 535, in default_weight_loader
      assert param.size() == loaded_weight.size()
```
## Modifications
Delete re-register-parameters for fp8 block quant linear.

## Checklist

- [x] Format your code according to the [Code Formatting with Pre-Commit](https://docs.sglang.ai/references/contribution_guide.html#code-formatting-with-pre-commit).
- [ ] Add unit tests as outlined in the [Running Unit Tests](https://docs.sglang.ai/references/contribution_guide.html#running-unit-tests-adding-to-ci).
- [ ] Update documentation / docstrings / example tutorials as needed, according to [Writing Documentation](https://docs.sglang.ai/references/contribution_guide.html#writing-documentation-running-docs-ci).
- [ ] Provide throughput / latency benchmark results and accuracy evaluation results as needed, according to [Benchmark and Profiling](https://docs.sglang.ai/references/benchmark_and_profiling.html) and [Accuracy Results](https://docs.sglang.ai/references/accuracy_evaluation.html).
- [ ] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co-author when merging the PR.
- [ ] Please feel free to join our Slack channel at https://slack.sglang.ai to discuss your PR.

PR #9810, opened 2025-08-30
fix weights update failure on block_quant model

tianyuzhou95:
## Motivation

Fix error during updating weights on the fp8 block quant model like deepseek-r1.

## Modifications

Only change the data in weights.

## Checklist


tianyuzhou95:
Same as https://github.com/sgl-project/sglang/pull/8390 , but in simple way. cc @BBuf

tianyuzhou95:
> Same as #8390 , but in simple way. cc @BBuf

Origin PR is better, seems things would be different in ROCm case.

PR #9818, opened 2025-08-30
[Feature][1/n] Custom pin memory allocation (numa supported)

DarkSharpness:
## Motivation

`torch`'s pin_memory allocator aligns large allocations (e.g., 65 GiB) to the next power of 2 (e.g., 128 GiB), introducing substantial memory overhead. This is particularly problematic when hierarchical KVCache is enabled, as we allocate large pinned host buffers. Furthermore, on NUMA systems, accessing pinned memory from a mismatched NUMA node incurs significant latency, exacerbating performance penalties when host memory is not properly aligned with the GPU's locality.

Reference:  [Where does torch align the allocation](https://github.com/pytorch/pytorch/blob/b994f6e3b331faeac693970bd1e14972f3fc9d4a/aten/src/ATen/core/CachingHostAllocator.h#L203)

## Modifications

Add a pin memory allocation function in `sgl_kernel`, supporting optional NUMA allocation.

TODO: Integrate this into `HiRadixCache`.

## Checklist

### DIFF
--- a/sgl-kernel/CMakeLists.txt
+++ b/sgl-kernel/CMakeLists.txt
@@ -22,6 +22,24 @@ enable_language(CUDA)
 find_package(CUDAToolkit REQUIRED)
 set_property(GLOBAL PROPERTY CUDA_SEPARABLE_COMPILATION ON)
 
+# NUMA
+if(DEFINED ENV{CONDA_PREFIX})
+    set(CONDA_LIB_DIR "$ENV{CONDA_PREFIX}/lib")
+    message(STATUS "Using Conda lib dir: ${CONDA_LIB_DIR}")
+    link_directories(${CONDA_LIB_DIR})
+    set(CONDA_INCLUDE_DIR "$ENV{CONDA_PREFIX}/include")
+    include_directories(${CONDA_INCLUDE_DIR})
+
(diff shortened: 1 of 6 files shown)


xiaguan:
Nice work! Really looking forward to this.

stmatengss:
@DarkSharpness Should we implement this as a CUDA program, given that we don't utilize GPU device memory? Would a C++ interface be sufficient?

FlamingoPg:
One more question, how should we use this feature with sglang engine?

DarkSharpness:
> @DarkSharpness Should we implement this as a CUDA program, given that we don't utilize GPU device memory? Would a C++ interface be sufficient?

Hi. I put this code in sgl-kernel since it requires NUMA lib, and I just want to reuse the NUMA lib config in CMakeList. I agree that a C++ interface might be sufficient, and I'm not sure about that.

DarkSharpness:
> One more question, how should we use this feature with sglang engine?

The Python integration will be handled in a separate PR. On GPU nodes with NUMA affinity, each GPU's CPU pin memory pool will be allocated on the NUMA node to which the GPU is affiliated. This can significantly improve the performance of data transfer between CPU and GPU, potentially accelerating hierarchical caching.

PR #9819, opened 2025-08-30
[WIP] External shared experts 

qmkakaxi:
**Co-authored-by:** raindaywhu raindaywhu@163.com
releated issue #9438 
## Motivation

In MoE inference under ​**Expert Parallelism (EP)**​, each device executes both the shared experts and the routed experts **sequentially** on the same device, which increases per-device load and end-to-end latency. To address this, we **externalize shared experts** by placing them on dedicated ranks/devices and aggregating the outputs of shared and routed experts via ​dispatch/combine, thereby eliminating on-device serialization and reducing load and latency.

## Overview

Empirically, shared-expert demand is much lower; with **top-8** gating (DeepSeek-style), we observe roughly an **8:1** ratio of routed-expert compute to shared-expert usage. By separating shared experts and routed experts onto different ranks and using the **`dispatch`** operator to route tokens (then **`combine`** to aggregate results), the shared-expert cost is largely overlapped/hidden behind routed-expert computation. This reduces end-to-end latency and eases per-rank load.

external_moe_shared_experts

#### 1) New parameter

* `moe_shared_expert_rank_num`: the number of ranks dedicated to shared experts (externalized).

#### 2) Inference adaptation

We leverage Ascend NPU **`dispatch`/`combine`** operators. By configuring the number of externalized shared experts, forwarding and aggregation are handled automatically, so the **top-k** logic remains unchanged. The only requirement during inference is to divide the shared-expert outputs by the coefficient `routed_scaling_factor`.

#### 3) Weight-loading adaptation

The existing `num_fused_shared_experts` path already supports loading shared experts as routed experts. Here we only adjust the expert map so that **shared-expert ranks** load only shared experts, and **routed-expert ranks** load only routed experts.

## Experimental results

On ​**Ascend NPU**​, with ​**EP = 64**​, we observe a performance improvement of ​**over 5 ms**​.## Advantages and Limitations
We will present additional experimental results in subsequent updates.

## Advantages and Limitations

* **Advantages.** With minimal code changes, we enable ​**externalized shared experts**​, improving inference performance.
* **Limitations.** This relies on  token dispatch/combine with external shared experts. Currently  supported by Ascend NPU  on A3.

**Update pending:** The code adapted for EPLB is done and will be supplemented later.

### DIFF
--- a/python/sglang/srt/layers/moe/ep_moe/layer.py
+++ b/python/sglang/srt/layers/moe/ep_moe/layer.py
@@ -364,7 +364,8 @@ class DeepEPMoE(EPMoE):
             routed_scaling_factor=routed_scaling_factor,
         )
         self.deepep_mode = get_deepep_mode()
-
+        self.rank = torch.distributed.get_rank()
+        
         # TODO: move to the beginning of the file
         from sglang.srt.distributed.parallel_state import get_tp_group
         from sglang.srt.two_batch_overlap import MaybeTboDeepEPDispatcher
@@ -741,7 +742,9 @@ class DeepEPMoE(EPMoE):
         seg_indptr = seg_indptr.to(torch.int64)
 
         import torch_npu
-
+        if self.moe_ep_rank < 0:
+            seg_indptr = None
+            
         # gmm1: gate_up_proj
         hidden_states = torch_npu.npu_grouped_matmul(
             x=[hidden_states],
@@ -778,6 +781,9 @@ class DeepEPMoE(EPMoE):
             group_list=seg_indptr,
             output_dtype=output_dtype,
         )[0]
+
+        if self.moe_ep_rank < 0:
+            hidden_states /= self.routed_scaling_factor
 
         return hidden_states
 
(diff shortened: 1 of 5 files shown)


python/sglang/srt/layers/moe/ep_moe/layer.py
```diff
@@ -364,7 +364,8 @@ def __init__(
             routed_scaling_factor=routed_scaling_factor,
         )
         self.deepep_mode = get_deepep_mode()
-
+        self.rank = torch.distributed.get_rank()
```
Alcanderian, line 367: unused

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -352,6 +352,18 @@ def __init__(
         self.shared_experts_is_int8 = False
         self.shared_experts_is_fp8 = False
         self.shared_experts_weight_block_size = None
+        self.rank = torch.distributed.get_rank()
```
Alcanderian, line 355: unused

Issue #9151, opened 2025-08-13
[Bug] CUDA error: uncorrectable ECC error encountered when using HiCache with xPyD disaggregation.


ericschreiber:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Describe the bug

When running PD-disaggregation in setup 2P2D with HiCache active, running some larger batch size (in my case 512 concurrent requests) leads to `CUDA error: uncorrectable ECC error encountered` which blocks the GPU and I have to restart the node.
Running the same setup in 1P1D works fine. Therefore, I suspect some illegal memory access of the HiCache from the GPU to the CPU memory. However, this is just a guess.

<details>

<summary>Stack trace of P 0</summary>
[2025-08-13 02:01:54 TP0] Prefill batch. #new-seq: 3, #new-token: 8192, #cached-token: 32, token usage: 0.03, #unbootstrapped-req: 155, #queue-req: 105, #transferring-req: 2, input throughput (token/s): 9175.92, 
[2025-08-13 02:01:56 TP0] Scheduler hit an exception: Traceback (most recent call last):
  File "/sgl-workspace/sglang/python/sglang/srt/managers/scheduler.py", line 2584, in run_scheduler_process
    scheduler.event_loop_overlap_disagg_prefill()
  File "/usr/local/lib/python3.10/dist-packages/torch/utils/_contextlib.py", line 120, in decorate_context
    return func(*args, **kwargs)
  File "/sgl-workspace/sglang/python/sglang/srt/disaggregation/prefill.py", line 334, in event_loop_overlap_disagg_prefill
    self.process_batch_result_disagg_prefill(tmp_batch, tmp_result)
  File "/sgl-workspace/sglang/python/sglang/srt/disaggregation/prefill.py", line 373, in process_batch_result_disagg_prefill
    logits_output, next_token_ids, _ = self.tp_worker.resolve_last_batch_result(
[...]

PR #9821, opened 2025-08-30
[Bugfix]add hicache event check before mooncake transfer in pd case

Bruce-x-1997:
## Motivation

https://github.com/sgl-project/sglang/issues/9151
as this issue say, I find  it could trigger err error in prefill pod when using pd & hicache at the same time, and transferengine is mooncake.
so I try to fix it, I suspect it is a bug about sync between data movement of hicache and transferEngine.

## Modifications

just add self.tree_cache.check_hicache_events() before send_kvcache in MooncakeKVManager

## Checklist

### DIFF
--- a/python/sglang/srt/disaggregation/mooncake/conn.py
+++ b/python/sglang/srt/disaggregation/mooncake/conn.py
@@ -167,12 +167,14 @@ class MooncakeKVManager(BaseKVManager):
         args: KVArgs,
         disaggregation_mode: DisaggregationMode,
         server_args: ServerArgs,
+        tree_cache: RadixCache,
         is_mla_backend: Optional[bool] = False,
     ):
         self.kv_args = args
         self.local_ip = get_local_ip_auto()
         self.is_mla_backend = is_mla_backend
         self.disaggregation_mode = disaggregation_mode
+        self.tree_cache = tree_cache
         self.init_engine()
         # for p/d multi node infer
         self.bootstrap_port = server_args.disaggregation_bootstrap_port
@@ -697,6 +699,7 @@ class MooncakeKVManager(BaseKVManager):
                                 )
                                 break
 
+                        self.tree_cache.check_hicache_events()
                         chunked_dst_kv_indice = req.dst_kv_indices[kv_chunk.index_slice]
 
                         # NOTE: This is temporarily a workaround to deal with the case where the prefill_kv_indices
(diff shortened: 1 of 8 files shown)


python/sglang/srt/disaggregation/mooncake/conn.py
```diff
@@ -697,6 +699,7 @@ def transfer_worker(
                                 )
                                 break
 
+                        self.tree_cache.check_hicache_events()
```
xiezhq-hermann: I think it probably work too even if you put the check outside the for loop

ShangmingCai: @Bruce-x-1997 Is it possible we do this in here: 
https://github.com/sgl-project/sglang/blob/9970e3bf328a1f59d9a7f79613c7d620ca522068/python/sglang/srt/disaggregation/prefill.py#L583C9-L591

Bruce-x-1997: oh, I can try it, you might be right)

Bruce-x-1997: yeah...but I worry about it could lead to more time cost
as I see send_* only is used in this thread, so I add it here)

ShangmingCai: I understand. But I wonder if this is a mooncake-specific change, so it doesn't look clean to modify all the interfaces. I think we need @ByronHsu to check on this.

xiezhq-hermann: I think we should not merge this PR till we find out the root cause of the problem. Can you also help take a look why this problem might occur? @hnyls2002

Bruce-x-1997: yeah, but I think adding sync between hicache & transfer engine might be a right strategy?
or add a lower sync  @xiezhq-hermann 
and I change all interfaces just because I think transfer engine might should be aware of cache, as both two are for data movement about hbm in gpu @ShangmingCai

xiezhq-hermann:
btw can you try reproduce this with `--hicache-io-backend direct` and a larger page size like `--page-size-64`, if the problem occurs at the interference this might solve the issue. that way we can also scope the problem.

Bruce-x-1997:
948b01a0 [Refactor] Remove Hicache Load & Write threads (#10127)
have fix ecc bug, close the pr

wonderisland:
> [948b01a](https://github.com/sgl-project/sglang/commit/948b01a04c2d5be16ba325a8ac849359ba99e295) [Refactor] Remove Hicache Load & Write threads (#10127) have fix ecc bug, close the pr

hi,  did u fix this ecc bug?  I also ecountered it.....

PR #9340, opened 2025-08-19
[sgl-kernel] misc: update deepgemm version for sgl-kernel

FlamingoPg:
## Motivation

DeepGEMM updated for unify cuda version.
So we need upd for TORCH LIBRARY.

It depends on: https://github.com/sgl-project/sglang/pull/9167

## Checklist

- [x] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [x] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [x] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.ai/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.ai/developer_guide/contribution_guide.html#benchmark-the-speed).

PR #9844, opened 2025-08-31
Change tensor alignment method to mn major

mmangkad:
## Modifications

Change tensor alignment method to mn major (from changes on #9340).


zhaotyer:
> ## Modifications
> Change tensor alignment method to mn major (from changes on #9340).

It's work on H100

zhyncs:
Thanks for this quick fix!

PR #9849, opened 2025-08-31
EPLB: prefer to use physical experts in the same node

acelyc111:
## Motivation

Because of NVLink, intranode communication is faster than that of internode, when it's possible to use physical experts on the same node, we'd prefer to route to these GPU ranks.

## Modifications

When setting `--ep-dispatch-algorithm static`, select the experts on the same GPU at first if possible, then select experts on the same node if possible, at last, select the experts randomly as a fall back method.

## Checklist


acelyc111:
@fzyzcjy Please take a look, thanks!

ch-wan (approved):

Issue #9806, opened 2025-08-29
[Bug] Accuracy drop for nvidia/DeepSeek-R1-0528-FP4 with dp attetnion


trevor-m:
### Checklist

- [ ] 1. I have searched related issues but cannot get the expected help.
- [ ] 2. The bug has not been fixed in the latest version.
- [ ] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [ ] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [ ] 5. Please use English, otherwise it will be closed.

### Describe the bug

Using ToT main, Deepseek R1 FP4 is giving bad accuracy when dp-attention is enabled. I did a git bisect and found that https://github.com/sgl-project/sglang/pull/8616 introduced the regression

```
Accuracy: 0.446
Invalid: 0.540
Latency: 68.608 s
Output throughput: 6281.439 token/s
```

Previously the accuracy is around 0.954.

### Reproduction

```
python3 -m sglang.launch_server --model-path nvidia/DeepSeek-R1-0528-FP4 --trust-remote-code --quantization modelopt_fp4 --tp 4  --dp 4 --enable-dp-attention
python3 benchmark/gsm8k/bench_sglang.py --num-shots 8 --num-questions 1319 --parallel 1319 --port=30000
```

### Environment

```
Python: 3.10.12 (main, May 27 2025, 17:12:29) [GCC 11.4.0]
CUDA available: True
GPU 0,1,2,3: NVIDIA GB200
GPU 0,1,2,3 Compute Capability: 10.0
CUDA_HOME: /usr/local/cuda
NVCC: Cuda compilation tools, release 12.9, V12.9.86
CUDA Driver Version: 580.65.06
PyTorch: 2.8.0+cu129
sglang: 0.5.0rc2
sgl_kernel: 0.3.6
flashinfer_python: 0.2.14.post1
triton: 3.4.0
transformers: 4.55.2
torchao: 0.9.0
numpy: 2.2.6
aiohttp: 3.12.15
fastapi: 0.116.1
hf_transfer: 0.1.9
huggingface_hub: 0.34.4
interegular: 0.3.3
modelscope: 1.29.0
orjson: 3.11.2
outlines: 0.1.11
[...]

PR #9861, opened 2025-09-01
[NVIDIA] disable chunked prefix cache when dp is used

kaixih:
This PR is to work around an accuracy issue found in https://github.com/sgl-project/sglang/issues/9806.

It seems the chunked prefix cache doesn't work well with the DP. So, we disable it for now to recover the accuracy.

cc @trevor-m @kushanam @zhyncs

### DIFF
--- a/python/sglang/srt/model_executor/model_runner.py
+++ b/python/sglang/srt/model_executor/model_runner.py
@@ -522,10 +522,11 @@ class ModelRunner:
 
         if not self.use_mla_backend:
             server_args.disable_chunked_prefix_cache = True
-
-        if self.dp_size > 1:
-            logger.info("Disable chunked prefix cache when dp size > 1.")
-            server_args.disable_chunked_prefix_cache = True
+        # TODO(kaixih@nvidia): remove this once we have a better solution for DP attention.
+        #  For more details, see: https://github.com/sgl-project/sglang/issues/8616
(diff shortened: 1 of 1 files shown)


Fridge003:
@kaixih The accuracy issue is only reported on flashinfer backend, so maybe we should disable chunked prefix cache for flashinfer backend + dp. On fa3 backend it should work well as before so no need to disable for dp.

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -523,6 +523,10 @@ def model_specific_adjustment(self):
         if not self.use_mla_backend:
             server_args.disable_chunked_prefix_cache = True
```
Fridge003: Please add a TODO, since this should be fixed in the future

kaixih: Done. PTAL. Thx.

kaixih:
> @kaixih The accuracy issue is only reported on flashinfer backend, so maybe we should disable chunked prefix cache for flashinfer backend + dp. On fa3 backend it should work well as before so no need to disable for dp.

Sure. I tested it and found that the accuracy issue also occurs with `trtllm_mla` and `cutlass_mla`, in addition to the default `flashinfer` backend. I’ve updated the check accordingly.

Fridge003 (approved):
LGTM

kaixih:
@zhyncs Can we merge?

PR #9871, opened 2025-09-01
[Bug fix] Fix Gemma 2 and fix Gemma 3 multimodal with bs > 1 on NPU

ssshinigami:
## Motivation

Fix inference of Gemma 2 and Gemma 3 multimodal scenario with batch > 1 on Ascend devices and Phi3 model problem

## Modifications

To fix Gemma 3 mulimodal with bs > 1: make page allignment for chunked batch.
To fix Gemma 2 - modified forward_npu method for GemmaRMSNorm operation.
Fix phi-3 model accuracy from with tp_size: tp1: 0.85  tp2: 0.75  tp4: 0.56

## Accuracy Tests

GSM8K test for Gemma2:

image

## Benchmarking and Profiling

No performance changes

## Checklist

- [  x ] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- 
image


JustinTong0323 (approved):

python/sglang/srt/managers/schedule_policy.py
```diff
@@ -549,7 +549,7 @@ def add_one_req(self, req: Req, has_chunked_req: bool):
                 )
             else:
                 # Make sure at least one page is available
-                trunc_len = self.rem_chunk_tokens - self.page_size + 1
+                trunc_len = self.rem_chunk_tokens // self.page_size * self.page_size
```
xiezhq-hermann, line 552: can you justify this change? the point of trunc_len is at lease to reserve one page for prefilling not only to align with page size. e.g., if rem_chunk_tokens == 2 * page_size, the trunc_len would leave no pages left

Makcum888e: trunc_len cheks that at least one page is available. `if trunc_len <= 0:` this if will work the same as before changes. 
in this case `rem_chunk_tokens == 2 * page_size` 
    before: `trunc_len == page_size + 1 > 0`, 
    after: `trunc_len == 2 * page_size > 0`.

after that check tranc_len tokens will be add to the batch for run prefill
in this case
    before: `page_size + 1` new tokens
    after: `2 * page_size` new tokens
and according to the prefill budget we can serve such number of tokens

image

in this case `rem_chunk_tokens` will be decreased on `2 * page_size` and before and after changes, because number of occupied slots aligns in `_update_prefill_budget`

image
image


so we can process more tokens in one batch and it should be available

xiezhq-hermann (approved):

PR #9900, opened 2025-09-02
[CPU] Fix phi4-mm prompt issue in bench_serving

blzheng:
## Motivation

This PR fixes the following issue when running the phi-4-multimodal model with `bench_serving`.
image

## Checklist

### DIFF
--- a/python/sglang/bench_serving.py
+++ b/python/sglang/bench_serving.py
@@ -896,7 +896,11 @@ def sample_mmmu_requests(
                 prompt = f"Question: {question}\n\nAnswer: "
                 if apply_chat_template:
                     try:
-                        if "phi-4-multimodal" in tokenizer.name_or_path.lower():
+                        is_phi4_multimodal = (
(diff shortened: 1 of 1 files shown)


JustinTong0323 (approved):

JustinTong0323:
No side effects for other models, but it seems not related with the error given above

blzheng:
> No side effects for other models, but it seems not related with the error given above

The error indicates that it went through a text-only path, which is unexpected since we are running a multimodal case.
https://github.com/sgl-project/sglang/blob/369b143366a4c58cc37f81bd4141c8794f26e56f/python/sglang/srt/multimodal/processors/base_processor.py#L589-L598

lifuhuang:
> No side effects for other models, but it seems not related with the error given above

It seems that the current bench_serving does not support phi4mm (and likely other VLM with different token style). We might need to think about how to generalize this in the future. But this PR seems to be a reasonable temp workaround.

PR #9909, opened 2025-09-02
[Hicache] Generic page get bugfix

ykwd:
## Motivation

The call to `self.mem_pool_host.set_from_flat_data_page` must be executed *before* `operation.increment(self.page_size)`.
If the order is reversed and `operation.increment(self.page_size)` runs first, other threads might read incorrect data before `self.mem_pool_host.set_from_flat_data_page` finishes. 

## Checklist

### DIFF
--- a/python/sglang/srt/managers/cache_controller.py
+++ b/python/sglang/srt/managers/cache_controller.py
@@ -659,13 +659,13 @@ class HiCacheController:
                     f"Prefetch operation {operation.request_id} failed to retrieve page {hash_values[i]}."
                 )
                 break
-            if operation.increment(self.page_size):
-                self.mem_pool_host.set_from_flat_data_page(
-                    host_indices[i * self.page_size],
-                    page_data[i],
-                )
-            else:
-                break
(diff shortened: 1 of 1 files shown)


python/sglang/srt/managers/cache_controller.py
```diff
@@ -659,13 +659,14 @@ def _generic_page_get(self, operation, hash_values, host_indices):
...
                 )
                 break
-            if operation.increment(self.page_size):
-                self.mem_pool_host.set_from_flat_data_page(
-                    host_indices[i * self.page_size],
-                    page_data[i],
-                )
-            else:
-                break
+            # Must set the data before increasing the completed tokens.
```
hzh0425, line 662: image

If `operation.increment(self.page_size)` is called before `mark_done()`, and after it succeeds, another thread calls `mark_done()`, while `set_from_flat_data_page` has not yet completed, there could indeed be a risk of reading dirty data.

hzh0425 (approved):
LGTM ， cc @xiezhq-hermann

xiezhq-hermann (approved):

PR #9910, opened 2025-09-02
add prefetched_tokens to prompt_tokens_details

pansicheng:
## Modifications

When deploying HiCache Storage, customers require visibility into whether requests are served from the storage layer. This PR enhances the cached_details field in the request response by adding prefetched_tokens information. This metric explicitly indicates how many tokens were prefetched from storage (in addition to existing cached_tokens tracking), providing users with granular insights into storage utilization and improving overall observability for performance monitoring and debugging.

## Checklist

### DIFF
--- a/python/sglang/srt/entrypoints/openai/serving_chat.py
+++ b/python/sglang/srt/entrypoints/openai/serving_chat.py
@@ -462,6 +462,7 @@ class OpenAIServingChat(OpenAIServingBase):
         prompt_tokens = {}
         completion_tokens = {}
         cached_tokens = {}
+        prefetched_tokens = {}
         hidden_states = {}
 
         try:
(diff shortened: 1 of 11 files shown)


xiezhq-hermann:
I think it is a bit hard to justify adding an entry to the response specifically for historage metrics, can you consider moving the functionality under the --enable-metrics umbrella and using standard metrics collector as I suggested in https://github.com/sgl-project/sglang/pull/9965? It is quite straightforward to query aggregated metrics, and I don't think we really need per-request logging.

pansicheng:
> I think it is a bit hard to justify adding an entry to the response specifically for historage metrics, can you consider moving the functionality under the --enable-metrics umbrella and using standard metrics collector as I suggested in #9965? It is quite straightforward to query aggregated metrics, and I don't think we really need per-request logging.

Sure, replaced by #9965

PR #9927, opened 2025-09-02
fix(cache): move ongoing_prefetch pop after validation to prevent leak

xiaguan:
There is a high priority bug.

It will cause all prefetch operations to leak, as prefetch operations cannot be released, which subsequently causes hicache to malfunction (due to memory not being freed), and prefetch operations cannot proceed because the prefetch limit is constantly being triggered.

@xiezhq-hermann 

## Checklist


huangtingwei9988 (approved):
LGTM

hzh0425 (approved):
I've encountered this issue too

xiezhq-hermann (approved):

PR #9931, opened 2025-09-02
[Vulnerability]feat(conn): set bootstrap server host

jimmy-evo:
set kv bootstrapserver's host with server_args.host, do not use 0.0.0.0 which is dangerous

## Motivation

Using 0.0.0.0 on a public network is risky. In my experience, I've observed that some junk data sent to this server from the public network caused JSON parsing to fail, which ultimately led to the bootstrap server crashing.

## Modifications

set arg: `host` with server_args.host

When deploying on a public network, set the host to the internal IP, while the upper-layer gateway or load balancer listens on the public IP.

## Checklist


ShangmingCai (commented):
Motivation looks good to me. @hnyls2002 will check the implementation.

jinmingyi1998:
The code in this section is not frequently modified. I will rebase it after the final review. @hnyls2002

hnyls2002 (approved):

ShangmingCai:
@jinmingyi1998 It seems like this PR will break the sgl-router CI: https://github.com/sgl-project/sglang/actions/runs/17428441201/job/49482688578?pr=9931

jimmy-evo:
@ShangmingCai  
I think it did not break the sgl-router, but there are some conflict with CI network env?
the only difference is bootstrap server listen on server's host instead of 0.0.0.0
can you help analyse what happened in CI?

key4ng:
Hi @slin1237 , I'm looking at the pd-router ci failure issue: https://github.com/sgl-project/sglang/actions/runs/17428441201/job/49482688578?pr=9931

Before this PR,  the CI is working because all the kv bootstrap servers bind to 0.0.0.0. But after this PR, the the prefill bootstrap will bound to its own loopback. a quick fix in CI will be to change all the service to host in one ip. Not sure if it's the right way to resolve the issue, would like to hear from your advice

slin1237:
router (minilb or rust router) all use original prefill server host, port, and generated rand as injected value to request payload
changing this shouldnt impact anything, because it shouldnt use 0.0.0.0
but it looks like we are missing something?
otherwise the ci shouldnt fail

key4ng:
Seems it still use get_ip, the bootstrap server won't bound to the host
https://github.com/sgl-project/sglang/blob/de9217334b28f9a1d9e3ab7c9a249abd4f71730b/python/sglang/srt/disaggregation/mooncake/conn.py#L1031

https://github.com/sgl-project/sglang/blob/de9217334b28f9a1d9e3ab7c9a249abd4f71730b/python/sglang/srt/utils.py#L1941

jimmy-evo:
@ShangmingCai @slin1237 
```
I0903 09:28:43.566442 2269537 transfer_engine.cpp:422] Metrics reporting is disabled (set MC_TE_METRIC=1 to enable)
I0903 09:28:43.566462 2269537 transfer_engine.cpp:44] Transfer Engine starting. Server: 10.0.0.35, Metadata: P2PHANDSHAKE, ip_or_host_name: , rpc_port: 0
I0903 09:28:43.566466 2269537 transfer_engine.cpp:63] Transfer Engine parseHostNameWithPort. server_name: 10.0.0.35 port: 12001
I0903 09:28:43.566493 2269537 transfer_engine.cpp:114] Transfer Engine RPC using P2P handshake, listening on 10.0.0.35:15623
I0903 09:28:43.566565 2269537 transfer_engine.cpp:138] Auto-discovering topology...
I0903 09:28:43.567268 2269537 transfer_engine.cpp:153] Topology discovery complete. Found 1 HCAs.
```
In CI, transfer engine should also listen 127.0.0.1~8. So, is this the problem here?
`self.local_ip = get_local_ip_auto()`

https://github.com/sgl-project/sglang/blob/b648d8621663bbc2356964f8a5660da6ea58876e/python/sglang/srt/disaggregation/mooncake/conn.py#L162-L173

jimmy-evo:
> ```
> I0903 09:28:43.566442 2269537 transfer_engine.cpp:422] Metrics reporting is disabled (set MC_TE_METRIC=1 to enable)
> I0903 09:28:43.566462 2269537 transfer_engine.cpp:44] Transfer Engine starting. Server: 10.0.0.35, Metadata: P2PHANDSHAKE, ip_or_host_name: , rpc_port: 0
> I0903 09:28:43.566466 2269537 transfer_engine.cpp:63] Transfer Engine parseHostNameWithPort. server_name: 10.0.0.35 port: 12001
> I0903 09:28:43.566493 2269537 transfer_engine.cpp:114] Transfer Engine RPC using P2P handshake, listening on 10.0.0.35:15623
> I0903 09:28:43.566565 2269537 transfer_engine.cpp:138] Auto-discovering topology...
> I0903 09:28:43.567268 2269537 transfer_engine.cpp:153] Topology discovery complete. Found 1 HCAs.
> ```
> 
> In CI, transfer engine should also listen 127.0.0.1~8. So, is this the problem here? `self.local_ip = get_local_ip_auto()`
> 
> https://github.com/sgl-project/sglang/blob/b648d8621663bbc2356964f8a5660da6ea58876e/python/sglang/srt/disaggregation/mooncake/conn.py#L162-L173

emmmm.... CI error is not caused by here

ShangmingCai:
Maybe check whether `bootstrap_addr` is correctly passed in `kv_receiver` in decode.py?

jimmy-evo:
I think the real problem is SGLANG_HOST_IP not set in CI. it should be set when listen 127.0.0.1~8 which is different to get_ip()

This is an error caused by incorrect usage.



https://github.com/sgl-project/sglang/blob/27e8ffed37ef4497b55c4d1745edd301744b72ee/python/sglang/srt/utils.py#L1943C26-L1943C40

@key4ng  Can you help review, just set env `SGLANG_HOST_IP`

ShangmingCai:
The 8 gpu tests have failed. I think this PR somehow makes SGLANG_HOST_IP a must-config-env-var, which gonna cause a lot of complain if we merge this. I will discuss this with @ByronHsu when he is available.

jimmy-evo:
@ShangmingCai SGLANG_HOST_IP is not a must-config-env-var any more. i just fix it.
the root cause is kv manager dont know bootstrap server host. Now kv_mgr can get bootstrap server host because it is `server_args.host`. So it doesn't need `dist_init_dir` or `get_ip` any more！

python/sglang/srt/disaggregation/common/conn.py
```diff
@@ -71,17 +72,8 @@ def __init__(
...
-                    host = self.dist_init_addr
-                else:
-                    host, _ = self.dist_init_addr.rsplit(":", 1)
-            else:
-                host = socket.gethostbyname(self.dist_init_addr.rsplit(":", 1)[0])
-        else:
-            host = get_ip()
-            host = maybe_wrap_ipv6_address(host)
+        host = self.bootstrap_host
+        host = maybe_wrap_ipv6_address(host)
```
ShangmingCai: Since you removed the branch when `dist_init_addr` is not None, can you run tests to verify that if we run one prefill instance with multiple nodes, will `self.bootstrap_host = server_args.host` also works for this case?

jimmy-evo:
@ShangmingCai you're right, when using multi node, `server_args.host` can't set to `dist_init_addr`

ShangmingCai (approved):
LGTM now.

python/sglang/srt/disaggregation/common/conn.py
```diff
@@ -80,7 +82,8 @@ def _register_to_bootstrap(self):
             else:
                 host = socket.gethostbyname(self.dist_init_addr.rsplit(":", 1)[0])
         else:
-            host = get_ip()
+            # single node: bootstrap server's host is same as http server's host
+            host = self.bootstrap_host
```
ShangmingCai, line 86: cc: @ishandhanani , maybe check whether this change is sufficient for NIXL backend as well.

jimmy-evo:
Summary of changes related to bootstrap server host:
1. `sgl router`‘s bootstrap server host == `server_args.host`
2. bootstrap server listen host == `server_args.host`
3. register_to_bootstrap server host == `server_args.host` / `dist_init_addr`

PR #9846, opened 2025-08-31
feat(hicache): Add generic hicache ci e2e test and benchmark test

hzh0425:
## Checklist

- [ ] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.ai/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.ai/developer_guide/contribution_guide.html#benchmark-the-speed).

PR #9938, opened 2025-09-02
fix 3fs zerocopy

pansicheng:
## Motivation

In the page_first zero-copy scenario, the attention key and value caches are stored separately in 3FS on a per-page basis. This layout necessitates special handling for prefix hashing.

## Checklist


xiezhq-hermann:
do you think it would make more sense to separate K and V on host memory pool as well?

pansicheng:
> do you think it would make more sense to separate K and V on host memory pool as well?

@xiezhq-hermann The current modification is mainly driven by the fact that, under the MHA (Multi-Head Attention) architecture, the dimensions of the kv_buffer are defined as (2, self.size, self.layer_num, self.head_num, self.head_dim). In this layout, the K and V of a single page are not stored contiguously in memory. Would it be better to change the layout so that K and V of the same page are stored contiguously? For example, changing the dimensions to (self.size, 2, self.layer_num, self.head_num, self.head_dim).

python/sglang/srt/mem_cache/storage/hf3fs/storage_hf3fs.py
```diff
@@ -415,22 +415,12 @@ def exists(self, key: str) -> bool:
         return result[0] if result else False
 
     def batch_exists(self, keys: List[str]) -> int:
-        if self.is_page_first_layout and not self.is_mla_model:
-            query_keys = []
-            # Compatible with page_first layout's key format, Refer to memory_pool_host.py#get_buffer_with_hash
-            for key in keys:
-                query_keys.append(f"{key}-k")
-                query_keys.append(f"{key}-v")
```
pansicheng, line 423: @hzh0425 Considering that the keys for kcache and vcache are constructed by the get_buffer_with_hash function when zero-copy is involved, what do you think about moving the key construction logic from batch_exists to the outside and converging it into get_buffer_with_hash to improve maintainability?

hzh0425: Sure

hzh0425 (approved):
LGTM，I‘ve tested. 

Pls Merge after this pr: https://github.com/sgl-project/sglang/pull/9846

python/sglang/srt/managers/cache_controller.py
```diff
@@ -746,7 +750,15 @@ def _storage_hit_query(self, operation) -> tuple[list[str], int]:
                     batch_tokens[i : i + self.page_size], last_hash
                 )
                 batch_hashes.append(last_hash)
-            hit_page_num = self.storage_backend.batch_exists(batch_hashes)
+            if self.is_3fs_zerocopy:
```
xiezhq-hermann: can we try to get rid of this kind of condition?

pansicheng: Fixed! PTAL

xiezhq-hermann (approved):

hzh0425:
> LGTM，I‘ve tested.
> 
> Pls Merge after this pr: #9846

It can be merged now.

PR #9947, opened 2025-09-03
feat: Add FlexAttention Backend for Efficient Sparse Attention

yukiy927:
## Motivation

This PR introduces a new attention backend leveraging PyTorch's flex_attention API. While existing backends like FlashAttention are highly optimized for dense attention, they offer limited flexibility for custom sparse patterns.

The FlexAttention backend addresses this gap by enabling efficient computation with user-defined sparse attention masks. This is particularly beneficial for:

1. Long-Context Models: Applying sparse patterns can significantly reduce the computational and memory overhead for very long sequences.
2. Research and Development: It provides an easy-to-use, high-performance foundation for developers and researchers experimenting with novel attention mechanisms.

The implementation is verified to maintain numerical parity with the torch native attention backend.

We can support flexible variant sparse attention pattern based on this feature.

## Modifications

- **Added TorchFlexAttnBackend**: Implemented a new backend class that wraps torch.nn.functional.flex_attention.
- **Backend Integration**: Integrated the new backend into the model's attention factory, allowing it to be selected via a configuration option.
- **Unit Test**: Added a unit test to verify the correctness and accuracy of the TorchFlexAttnBackend against the baseline torch native implementation

## Requirements

PyTorch Version: torch>=2.5.0 is required to use the flex_attention API.

## How to Use
To enable the FlexAttention backend, set the `--attention-backend` server argument to `flex_attention`.

**Example:**
```bash
    --model-path <path/to/your/model> \
    --attention-backend flex_attention
```
## Checklist


yukiy927:
Can you take a review of this @merrymercy

python/sglang/srt/server_args.py
```diff
@@ -528,6 +529,12 @@ def __post_init__(self):
             )
             self.disable_cuda_graph = True
 
+        if self.attention_backend == "flex_attention":
+            logger.warning(
+                "Cuda graph is disabled because of using torch Flex Attention backend"
+            )
+            self.disable_cuda_graph = True
```
Fridge003, line 599: Is it possible to support cuda graph for this backend in the future? Since cuda graph is crucial for inference performance.

python/sglang/srt/layers/attention/torch_flex_backend.py
```diff
@@ -0,0 +1,304 @@
...
+
+    def init_forward_metadata(self, forward_batch: ForwardBatch):
+        """Init the metadata for a forward pass."""
+        pass
+
+    def _causal_mask(self, b, h, q_idx, kv_idx):
+        return q_idx >= kv_idx
+
+    def _decode_mask(self, b, h, q_idx, kv_idx):
+        return q_idx == q_idx
```
Fridge003: Maybe it should be q_idx == kv_idx ?

python/sglang/srt/server_args.py
Fridge003: Seems speculative decoding is also not supported yet. We should add the assertion:
`assert self.speculative-algorithm is None`

python/sglang/srt/layers/attention/torch_flex_backend.py
```diff
@@ -0,0 +1,304 @@
...
+            per_req_query_redudant[:, prefill_seq_len_q:, :] = per_req_query
+
+            # get key and value from cache. per_req_tokens contains the kv cache
+            # index for each token in the sequence.
+            req_pool_idx = req_pool_indices[seq_idx]
+            per_req_tokens = req_to_token[req_pool_idx, :seq_len_kv]
+            per_req_key = k_cache[per_req_tokens].movedim(0, query.dim() - 2)
+            per_req_value = v_cache[per_req_tokens].movedim(0, query.dim() - 2)
+
+            if causal:
```
Fridge003: Do we need to create block mask and recompile the kernel for each sequence?
I feel this is not really efficient. 

In https://pytorch.org/blog/flexattention/, there is a method called prefix_mask. Maybe we can use this once and directly compute the output.

python/sglang/srt/layers/attention/torch_flex_backend.py
```diff
@@ -0,0 +1,304 @@
...
+
+        Returns:
+            output: [num_tokens, num_heads, head_size]
+        """
+
+        # [num_tokens, num_heads, head_size] -> [num_heads, num_tokens, head_size]
+        query = query.movedim(0, query.dim() - 2)
+
+        start_q, start_kv = 0, 0
+        for seq_idx in range(seq_lens.shape[0]):
```
Fridge003, line 200: The same problem as above. We should find a way to create mask and  run attention once.

python/sglang/srt/layers/attention/torch_flex_backend.py
yukiy927: Yes, we can create the mask only once during each forward pass, which is much more efficient. This has been updated in [ed373d7](https://github.com/sgl-project/sglang/pull/9947/commits/ed373d79668247aaaaf09b51b72f6455f14af977).
Previously, we put the mask construction inside each layer because many mainstream sparse attention methods apply different masks per layer. This allows for dynamic configurations based on the varying observations from each layer’s mask heatmap. 
However, for now this approach is indeed inefficient. We plan to implement more per-layer sparse attention in the future.

python/sglang/srt/layers/attention/torch_flex_backend.py
yukiy927: Yess, same update as above. We also put mask generation outside per layer during decode pass.

python/sglang/srt/layers/attention/torch_flex_backend.py
yukiy927: We combined the prefill and decode mask generation into a single nested function `_causal_mask`. This function now handles the query offset and avoids redundant query computations in [7e43ae8](https://github.com/sgl-project/sglang/pull/9947/commits/7e43ae843e41234a95d91f7a613500aba10699c7).

python/sglang/srt/server_args.py
yukiy927: cuda graph is currently not compatible with this backend, and there are still upstream fixes pending in PyTorch (see [pytorch/pytorch#145059](https://github.com/pytorch/pytorch/pull/145059)).

We’ll keep tracking this and introduce a multi-pass decode backend to support speculative decoding once the path is cudagraphable. This will be an ongoing effort, we can land it now and continue refining CUDA Graph and speculative decoding support incrementally.

yukiy927: > Seems speculative decoding is also not supported yet. We should add the assertion: `assert self.speculative-algorithm is None`

For now we added the assertion in [54e663e](https://github.com/sgl-project/sglang/pull/9947/commits/54e663e0d53cf5a37a300f1f4758c8898cb8fddc)

drisspg: what version of PT are you using, flexattention has cuda graph support

yukiy927: @drisspg I am using PT version as follows
```
torch                     2.8.0
torch_memory_saver        0.0.8
torchao                   0.9.0
torchaudio                2.8.0
torchvision               0.23.0
```

and I get notimplementederror when I run without set --disable-cuda-graph
```
  File "/sglang/python/sglang/srt/managers/scheduler.py", line 2586, in run_scheduler_process
    scheduler = Scheduler(
                ^^^^^^^^^^
  File "/sglang/python/sglang/srt/managers/scheduler.py", line 329, in __init__
    self.tp_worker = TpWorkerClass(
                     ^^^^^^^^^^^^^^
  File "/sglang/python/sglang/srt/managers/tp_worker_overlap_thread.py", line 67, in __init__
    self.worker = TpModelWorker(
                  ^^^^^^^^^^^^^^
  File "/sglang/python/sglang/srt/managers/tp_worker.py", line 89, in __init__
    self.model_runner = ModelRunner(
                        ^^^^^^^^^^^^
  File "/sglang/python/sglang/srt/model_executor/model_runner.py", line 248, in __init__
    self.initialize(min_per_gpu_memory)
  File "/sglang/python/sglang/srt/model_executor/model_runner.py", line 364, in initialize
    self.init_device_graphs()
  File "/sglang/python/sglang/srt/model_executor/model_runner.py", line 1661, in init_device_graphs
    self.graph_runner = graph_runners[self.device](self)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/sglang/python/sglang/srt/model_executor/cuda_graph_runner.py", line 293, in __init__
    self.model_runner.attn_backend.init_cuda_graph_state(
  File "/sglang/python/sglang/srt/layers/attention/base_attn_backend.py", line 24, in init_cuda_graph_state
    raise NotImplementedError()
NotImplementedError
```

drisspg: that is coming from here though right: https://github.com/sgl-project/sglang/blob/cda7e47ce7ce868fc06ccfbe2e88af8acfe9b263/python/sglang/srt/layers/attention/base_attn_backend.py#L26-L55

e.g. there methods this PR has not implemented enabling cuda-graphs

Fridge003:
@yukiy927 I just ran your branch, and found flex-attention backend is easy to cause OOM when there are more than 1 prompts. Do you have any clues?
截屏2025-09-10 21 24 31

yukiy927:
> @yukiy927 I just ran your branch, and found flex-attention backend is easy to cause OOM when there are more than 1 prompts. Do you have any clues? 截屏2025-09-10 21 24 31

Can you provide what device you are using? I can run longbench v2 which have 500-ish samples with tasks lengths as below on single H20.
image

yukiy927:
> @yukiy927 I just ran your branch, and found flex-attention backend is easy to cause OOM when there are more than 1 prompts. Do you have any clues? 截屏2025-09-10 21 24 31

here "more than one prompts" means batch_size > 1?

Fridge003:
> > @yukiy927 I just ran your branch, and found flex-attention backend is easy to cause OOM when there are more than 1 prompts. Do you have any clues? 截屏2025-09-10 21 24 31
> 
> here "more than one prompts" means batch_size > 1?

Yes. And even I set batch size to 1,  OOM will happen after several batches

I'm running on H100 with llama3-8b, these commands can reproduce this bug
```bash
python3 -m sglang.launch_server --model-path meta-llama/Llama-3.1-8B-Instruct --attention-backend flex_attention

```

Cuda version: 12.6
Pytorch version: 2.8.0+cu126


Are you using H20 with 144GB global memory?

yukiy927:
> > > @yukiy927 I just ran your branch, and found flex-attention backend is easy to cause OOM when there are more than 1 prompts. Do you have any clues? 截屏2025-09-10 21 24 31
> > 
> > 
> > here "more than one prompts" means batch_size > 1?
> 
> Yes. And even I set batch size to 1, OOM will happen after several batches
> 
> I'm running on H100 with llama3-8b, these commands can reproduce this bug
> 
> ```shell
> python3 -m sglang.launch_server --model-path meta-llama/Llama-3.1-8B-Instruct --attention-backend flex_attention
> 
> python3 benchmark/gsm8k/bench_sglang.py --num-shots 8 --num-questions 1319 --parallel 1319
> ```
> 
> Cuda version: 12.6 Pytorch version: 2.8.0+cu126
> 
> Are you using H20 with 144GB global memory?

Thx for your script. I am using H20 with 96GB global mem and I have reproduced it on single card. It runs successfully and takes about 92GB of memory. I guess H100 you are using has 80GB capacity, 16GB less. So the OOM is expected and not an implementation bug.

You may try setting `--mem-fraction-static 0.50`. I tested this on H20 and it reduced the usage to around 56GB, which should fit well on the H100.
```
python3 -m sglang.launch_server --model-path meta-llama/Llama-3.1-8B-Instruct --attention-backend flex_attention --mem-fraction-static 0.50

```
The issue is confirmed to be caused by device memory capacity differences, we can safely move forward.

Fridge003:
> python3 -m sglang.launch_server --model-path meta-llama/Llama-3.1-8B-Instruct --attention-backend flex_attention --mem-fraction-static 0.50

Is it possible to save the memory usage? Since usually we don't set mem-fraction-static to such a small value as 0.50

Fridge003:
I just ran a few test cases in GSM8K with llama3-8b, the accuracy also seems lower than expected
```bash
 python3 benchmark/gsm8k/bench_sglang.py --num-shots 8 --num-questions 100 --parallel 10
100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 100/100 [11:42<00:00,  7.03s/it]
Accuracy: 0.340
Invalid: 0.000
Latency: 702.946 s
Output throughput: 51.037 token/s

```

yukiy927:
> > python3 -m sglang.launch_server --model-path meta-llama/Llama-3.1-8B-Instruct --attention-backend flex_attention --mem-fraction-static 0.50
> 
> Is it possible to save the memory usage? Since usually we don't set mem-fraction-static to such a small value as 0.50

The --mem-fraction-static setting is more like a workaround for the device difference I mentioned above. It can help you run the benchmark smoothly on your H100 for further testing. 

Is there any globally acceptable memory coverage, e.g. one H100? I’ll look into optimizing the memory usage more systematically later on.

btw, have you ever tested gsm8k benchmark on torch native backend? It takes 86GB or so on H20 which I don't think it can fit in one h100 using the following scripts:
```
python3 -m sglang.launch_server --model-path meta-llama/Llama-3.1-8B-Instruct --attention-backend torch_native

python3 benchmark/gsm8k/bench_sglang.py --num-shots 8 --num-questions 100 --parallel 10
```

xwpaul3:
It seems this implement does not support cuda graph and bs>1, results in a extremely low throughput  in PD disaggration.

yukiy927:
> I just ran a few test cases in GSM8K with llama3-8b, the accuracy also seems lower than expected
> 
> ```shell
>  python3 benchmark/gsm8k/bench_sglang.py --num-shots 8 --num-questions 100 --parallel 10
> 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 100/100 [11:42<00:00,  7.03s/it]
> Accuracy: 0.340
> Invalid: 0.000
> Latency: 702.946 s
> Output throughput: 51.037 token/s
> ```

just fix the bug in the latest commit [244e4fc](https://github.com/sgl-project/sglang/pull/9947/commits/244e4fc2f03db19e730605c32af970806fa5e391). fyi, I created prefill and decode block mask lists for each forward pass. And we can reuse it in each layer.
I reran the benchmark as below and the acc issue was resolved:
```
100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 100/100 [03:42<00:00,  2.22s/it]
Accuracy: 0.810
```

yukiy927:
> > I just ran a few test cases in GSM8K with llama3-8b, the accuracy also seems lower than expected
> > ```shell
> >  python3 benchmark/gsm8k/bench_sglang.py --num-shots 8 --num-questions 100 --parallel 10
> > 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 100/100 [11:42<00:00,  7.03s/it]
> > Accuracy: 0.340
> > Invalid: 0.000
> > Latency: 702.946 s
> > Output throughput: 51.037 token/s
> > ```
> 
> just fix the bug in the latest commit [244e4fc](https://github.com/sgl-project/sglang/pull/9947/commits/244e4fc2f03db19e730605c32af970806fa5e391). fyi, I created prefill and decode block mask lists for each forward pass. And we can reuse it in each layer. I reran the benchmark as below and the acc issue was resolved:
> 
> ```
> 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 100/100 [03:42<00:00,  2.22s/it]
> Accuracy: 0.810
> ```

also worth to note, memory also is lowered to 85GB, maintaning the same memory and accuracy as torch natvie backend.

python/sglang/srt/layers/attention/torch_flex_backend.py
```diff
@@ -25,33 +25,48 @@ def __init__(self, model_runner: ModelRunner):
...
+                    device=self.device,
+                    _compile=False,
+                ))
 
     def _causal_mask(self, b, h, q_idx, kv_idx):
-        return self.q_offset + q_idx >= kv_idx
+        return q_idx >= kv_idx
+
+    def _decode_mask(self, b, h, q_idx, kv_idx):
+        return q_idx <= kv_idx
```
Fridge003, line 73: Is this always True since q_idx is always 0?

python/sglang/srt/layers/attention/torch_flex_backend.py
```diff
@@ -114,18 +137,22 @@ def _run_flex_forward_extend(
             per_req_key = k_cache[per_req_tokens].movedim(0, query.dim() - 2)
             per_req_value = v_cache[per_req_tokens].movedim(0, query.dim() - 2)
 
-            output[start_q:end_q, :, :] = (
+            if not causal:
+                raise NotImplementedError("Non-causal mode is not yet implemented.")
+            
+            per_req_out_redudant = (
```
Fridge003: Should be redundant?

test/srt/test_torch_flex_attention_backend.py
```diff
@@ -0,0 +1,47 @@
...
+        base_url = DEFAULT_URL_FOR_TEST
+        process = popen_launch_server(
+            model,
+            base_url,
+            timeout=DEFAULT_TIMEOUT_FOR_SERVER_LAUNCH,
+            other_args=["--attention-backend", "flex_attention"],
+        )
+
+        try:
+            args = SimpleNamespace(
```
Fridge003, line 32: I feel this test is not strong enough. Maybe you can use gsm8k instead?
You can refer to `test_full_deepseek_v3.py` on how to write gsm8k test

Fridge003:
This PR seems working after fixing accuracy bug. However it's still not efficient since the attention is computed on each sequence separately. In the end we need to compute attention for all the sequences within a single kernel. Maybe we can leave that for future

python/sglang/srt/layers/attention/torch_flex_backend.py
yukiy927: yes, seq_len_q here is always equal to 1, so actually every current q can attend to all kv_idx.

test/srt/test_torch_flex_attention_backend.py
yukiy927: ok, let me work on this.

python/sglang/srt/layers/attention/torch_flex_backend.py
yukiy927: Thanks for pointing out. It's a typo, fixed now.

test/srt/test_torch_flex_attention_backend.py
yukiy927: > I feel this test is not strong enough. Maybe you can use gsm8k instead? You can refer to `test_full_deepseek_v3.py` on how to write gsm8k test

Hi, before diving into it, I am curious about two simple questions: Is the recommendation to switch from MMLU to GSM8K because:
- MMLU has limitations for testing this particular feature, or 
- GSM8K generally provides more robust verification capabilities?

It would be great if you can share why and how to choose the appropriate benchmark going forward.

Fridge003: Mostly because this test didn't detect the accuracy bug I found past days ago, so I guess GSM8K can provide more robust verification in our case

yukiy927: ok, I have pushed a new gsm8k test replacing the original mmlu in [b969f2b](https://github.com/sgl-project/sglang/pull/9947/commits/b969f2be173285486d93fc8c39196bc23a3c3768). And I set the test threshold as 0.62, referenced from `test_mla_tp.py` and `test_mla_flashinfer.py` fyi.

yukiy927:
I'm not sure what brings this PR test failure here, do you have any clue? @Fridge003

Fridge003:
> I'm not sure what brings this PR test failure here, do you have any clue? @Fridge003

Some of them are flaky tests. I will rerun it.

Fridge003 (approved):
LGTM

zongy17:
Hi @yukiy927 @Fridge003 , this feature is exciting! From which version of SGLang it could be enabled by setting --attention-bakckend flex_attention? I tried v0.5.3rc0, but it failed and showed the valid choices not including flex_attention.

Fridge003:
> Hi @yukiy927 @Fridge003 , this feature is exciting! From which version of SGLang it could be enabled by setting --attention-bakckend flex_attention? I tried v0.5.3rc0, but it failed and showed the valid choices not including flex_attention.

Please try pulling the latest main branch, or wait for the next released version.

zongy17:
> > Hi @yukiy927 @Fridge003 , this feature is exciting! From which version of SGLang it could be enabled by setting --attention-bakckend flex_attention? I tried v0.5.3rc0, but it failed and showed the valid choices not including flex_attention.
> 
> Please try pulling the latest main branch, or wait for the next released version.

Hi @Fridge003 , thanks for your reply! I tried pip install the latest main branch, but when I specified --attention-backend flex_attention, it crashed down after showing "Application startup complete". Does it have any constraints on usage like models, TP, chunked-prefill size, ...? My running command is:

`
CUDA_VISIBLE_DEVICES=0,1,2,3 python -m sglang.launch_server --served-model-name GLM-Air --model-path .../GLM-4.5-Air-FP8/ --context-length 65536 --tensor-parallel-size 4 --chunked-prefill-size 8192 --port 8011 --mem-fraction-static 0.7 --disable-radix-cache --attention-bakckend flex_attention
`

When I deleted the `--attention-bakckend flex_attention`, it could work normally. Could you tell any possible reasons?

Fridge003:
> ll the latest main branch, but when I specified --attention-backend flex_attention, it crashed down after showing "Application startup complete". Does it have any constraints on usage like models, TP, chunked-prefill size, ...? My running command is:

Can you please paste error log?

alex1720-web:
> > ll the latest main branch, but when I specified --attention-backend flex_attention, it crashed down after showing "Application startup complete". Does it have any constraints on usage like models, TP, chunked-prefill size, ...? My running command is:
> 
> Can you please paste error log?

The error log is attached. Thanks!
[flex_error.txt](https://github.com/user-attachments/files/22458567/flex_error.txt)

Fridge003:
> > > ll the latest main branch, but when I specified --attention-backend flex_attention, it crashed down after showing "Application startup complete". Does it have any constraints on usage like models, TP, chunked-prefill size, ...? My running command is:
> > 
> > 
> > Can you please paste error log?
> 
> The error log is attached. Thanks! [flex_error.txt](https://github.com/user-attachments/files/22458567/flex_error.txt)

It seems to trigger assertion on num_q_head / num_kv_head. I just looked up the head num of GLM-air and found, and found num_q_head / num_kv_head is 96 / 8 = 12. This doesn't seem to satisfy restriction of flex attention (it should be power of 2).

Maybe we can just change for another backend

PR #9960, opened 2025-09-03
add variable TP Decode > Prefill size support

shaharmor98:
## Checklist

### DIFF
--- a/python/sglang/srt/disaggregation/nixl/conn.py
+++ b/python/sglang/srt/disaggregation/nixl/conn.py
@@ -437,7 +437,7 @@ class NixlKVManager(CommonKVManager):
 
             notif = "_".join([str(req.room), "kv", str(chunk_id), str(int(is_last))])
             decode_tp_size = self.decode_kv_args_table[req.agent_name].decode_tp_size
-            
+
             if decode_tp_size == self.tp_size:
                 kv_xfer_handle = self.send_kvcache(
                     req.agent_name,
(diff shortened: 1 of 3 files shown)


ShangmingCai (commented):
Is there any problem with the current mooncake slice implementation? I remember I verified the benchmark accuracy, and the result looks correct.

I think there is no need to calculate any global index, it has already been done at the `__init__` process, all we need to do is to calculate the relative index in consideration of prefill's `attn_tp_size` and decode's `attn_tp_size`.

ShangmingCai (commented):
Thx for the fix. I think it can be fixed through `src_head_start_offset = (
                dst_tp_rank_in_group * dst_heads_per_rank
            ) % src_heads_per_rank`, which is shorter and clearer.

python/sglang/srt/disaggregation/nixl/conn.py
```diff
@@ -49,6 +49,9 @@ class TransferInfo:
     dst_kv_indices: npt.NDArray[np.int32]
     dst_aux_index: int
     required_dst_info_num: int
+    decode_tp_size: int = 1
+    decode_tp_rank: int = 0
+    dst_kv_item_len: int = 0
```
ShangmingCai: Maybe this can be moved to `KVArgsRegisterInfo` since it is identical for all reqs?

shaharmor98: done

ShangmingCai (approved):
LGTM.

hnyls2002 (approved):

PR #9962, opened 2025-09-03
[Feature] Sglang Tracing: Fine-Grained Tracking for Request Latency - Part 1

sufeng-buaa:
## Motivation

The PR is response to https://github.com/sgl-project/sglang/issues/8965 . For details on the motivation and visual output, please refer to the issue.

## Modifications
To avoid overwhelming reviewers with a large amount of code, we have split the patch into two parts.

This is the first part, which includes:
* A request tracing package, which provides a set of static instrumentation APIs.
* Implementation of request tracing for normal requests, covering scenarios with MultiTokenizer and TP>1.

The second part will be submitted after this part has been reviewed or merged, include:
* Request tracing support for PD disaggregation and DP attention scenarios
* A script for converting opentelemetry data to perfetto data

To help reviewers better understand the design, we would like to clarify a few key points:

1. To trace multiple concurrently executing requests and observe intra-request parallelism (e.g., TP > 1), a global variable is used to maintain the trace context of the currently active request.
```
SglangTraceReqContext (rid="req-123")
├── SglangTraceThreadContext(thread_label="scheduler", tp_rank=0)
│ └── SglangTraceSliceContext (name="prefill") # cur slice
|
└── SglangTraceThreadContext(thread_label="scheduler", tp_rank=1)
  └── SglangTraceSliceContext (name="prefill") # cur slice
```

2. To capture the execution flow of a request, adjacent spans are linked using the opentelemetry.sdk.trace.span.Span.add_link() API. As a result, the SglangTraceSliceContext keeps track of the prev_span_context to enable proper linking.

3. When a request crosses thread boundaries, the trace context is explicitly propagated using a dedicated structure, `SglangTracePropagateContext`, which includes both the root span context and the previous span context to ensure continuity across threads.

## How to enable request tracing?
Please refer to `docs/references/production_request_trace.md`

## How to use request tracing APIs?
Please refer to `docs/references/production_request_trace.md` and `test/srt/test_tracing.py`

## Instrumentation Overhead Evaluation
Our testing platform is based on an Intel® Xeon® Platinum 8469C processor, configured with 192 CPU cores, 1 TB of RAM, and 8 NVIDIA H20 GPUs.

1. enable tracing
* The overhead of tracing a single slice is approximately 40 μs, including the combined cost of trace_slice_start() and trace_slice_end().
    In the scheduler, multiple requests are processed concurrently, so the actual overhead = single slice overhead * batch_size.<br>
    However, due to the overlap scheduling mechanism, the tracing overhead can be largely hidden by the GPU-side forward computation time.

* The overhead of a single cross-thread trace context propagation is approximately 90 μs.

2. disable tracing
* The overhead of an empty instrumentation point is approximately 300–400 ns, which is negligible under the overlap scheduling mechanism.

## TODO
1. Dynamically enable or disable via Http Post interface.
2. Request tracing for EP and PP scenarios is currently under implementation.

## Checklist

### DIFF
--- a/python/sglang/srt/entrypoints/http_server.py
+++ b/python/sglang/srt/entrypoints/http_server.py
@@ -31,6 +31,8 @@ from typing import Any, AsyncIterator, Callable, Dict, List, Optional
 
 import setproctitle
 
+from sglang.srt.tracing.trace import process_tracing_init, trace_set_thread_info
+
 # Fix a bug of Python threading
 setattr(threading, "_register_atexit", lambda *args, **kwargs: None)
 
@@ -177,6 +179,13 @@ async def init_multi_tokenizer() -> ServerArgs:
             scheduler_info=scheduler_info,
         )
     )
+
+    if server_args.enable_trace:
+        process_tracing_init(server_args.otel_endpoint, "sglang")
+        if server_args.disaggregation_mode == "null":
+            thread_label = f"MultiTokenizer-{tokenizer_manager.worker_id}"
+            trace_set_thread_info(thread_label)
+
     return server_args
 
 
@@ -1170,6 +1179,12 @@ def launch_server(
         tokenizer_manager, template_manager, scheduler_info = _launch_subprocesses(
             server_args=server_args,
         )
+
+        if server_args.enable_trace:
+            process_tracing_init(server_args.otel_endpoint, "sglang")
+            if server_args.disaggregation_mode == "null":
+                thread_label = "Tokenizer"
+                trace_set_thread_info(thread_label)
 
     set_global_state(
         _GlobalState(
--- a/python/sglang/srt/entrypoints/http_server.py
+++ b/python/sglang/srt/entrypoints/http_server.py
@@ -181,7 +181,7 @@ async def init_multi_tokenizer() -> ServerArgs:
     )
 
     if server_args.enable_trace:
-        process_tracing_init(server_args.otel_endpoint, "sglang")
+        process_tracing_init(server_args.oltp_traces_endpoint, "sglang")
         if server_args.disaggregation_mode == "null":
             thread_label = f"MultiTokenizer-{tokenizer_manager.worker_id}"
             trace_set_thread_info(thread_label)
@@ -1181,7 +1181,7 @@ def launch_server(
         )
 
         if server_args.enable_trace:
-            process_tracing_init(server_args.otel_endpoint, "sglang")
+            process_tracing_init(server_args.oltp_traces_endpoint, "sglang")
             if server_args.disaggregation_mode == "null":
                 thread_label = "Tokenizer"
                 trace_set_thread_info(thread_label)
--- a/python/sglang/srt/managers/scheduler.py
+++ b/python/sglang/srt/managers/scheduler.py
@@ -145,6 +145,15 @@ from sglang.srt.parser.reasoning_parser import ReasoningParser
 from sglang.srt.server_args import PortArgs, ServerArgs
 from sglang.srt.speculative.spec_info import SpeculativeAlgorithm
 from sglang.srt.torch_memory_saver_adapter import TorchMemorySaverAdapter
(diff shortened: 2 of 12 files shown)


python/sglang/srt/entrypoints/http_server.py
```diff
@@ -192,6 +194,13 @@ async def init_multi_tokenizer() -> ServerArgs:
             scheduler_info=scheduler_info,
         )
     )
+
+    if server_args.enable_trace:
```
ishandhanani, line 183: Seems like we can enable this when using the http server entrypoint. What do you think about enabling this for the `sgl.Engine` API as well?

sufeng-buaa: Yes, that makes sense. I'll add it, run some tests, and push an update shortly.

sufeng-buaa: I've enabled this for the sgl.Engine API.

ishandhanani: Neat! Thank you. 

I've kicked off the PR checks

ishandhanani:
Seems like there are some conflicts. Can you resolve?

sufeng-buaa:
> Seems like there are some conflicts. Can you resolve?

sufeng-buaa:
> Seems like there are some conflicts. Can you resolve?

ok, I have rebased my branch to the latest main.

ishandhanani:
@zhyncs - can you take a look?

ishandhanani (approved):

hnyls2002 (approved):

python/sglang/srt/managers/scheduler.py
```diff
@@ -814,6 +823,10 @@ def event_loop_normal(self):
             batch = self.get_next_batch_to_run()
             self.cur_batch = batch
 
+            if batch:
+                for req in batch.reqs:
+                    trace_event("schedule", req.rid)
```
merrymercy, line 828: wrap this into a function and only call it if tracing is enabled.
The principal: if tracing is not enabled, the overhead should just be a single if/else, not a for loop

python/sglang/srt/managers/scheduler.py
```diff
@@ -1880,8 +1904,23 @@ def process_batch_result(
...
+                trace_slice(
+                    "decode loop",
+                    req.rid,
+                    auto_next_anon=not req.finished(),
+                    thread_finish_flag=req.finished(),
+                )
+
         elif batch.forward_mode.is_extend():
             self.process_batch_result_prefill(batch, result, launch_done)
+            for req in batch.reqs:
```
merrymercy, line 1917: If tracing is not enabled, the overhead should just be a single if/else, not a for loop

python/sglang/srt/managers/scheduler.py
```diff
@@ -835,6 +848,10 @@ def event_loop_overlap(self):
             batch = self.get_next_batch_to_run()
             self.cur_batch = batch
 
+            if batch:
```
merrymercy, line 851: There is a `if batch` condition right after this, should you put this block under L855?

python/sglang/srt/managers/scheduler.py
```diff
@@ -835,6 +848,10 @@ def event_loop_overlap(self):
...
+                for req in batch.reqs:
+                    trace_event("schedule", req.rid)
```
merrymercy, line 853: Wrap this into a function called `trace_event_batch`.

The code in `event_loop_XXXX` should be very concise, only expose core logics

stmatengss:
@sufeng-buaa could you resolve these issues from @merrymercy even after this PR is merged?

sufeng-buaa:
> @sufeng-buaa could you resolve these issues from @merrymercy even after this PR is merged?

I have fixed these issue in https://github.com/sgl-project/sglang/pull/10804

PR #9965, opened 2025-09-03
Add storage read/write bandwidth logs to monitor kvcache performance

pansicheng:
## Motivation

Add storage read/write bandwidth logs to monitor kvcache performance, enhancing online deployment observability.

## Checklist

### DIFF
--- a/python/sglang/srt/mem_cache/hiradix_cache.py
+++ b/python/sglang/srt/mem_cache/hiradix_cache.py
@@ -32,9 +32,6 @@ class HiRadixCache(RadixCache):
         req_to_token_pool: ReqToTokenPool,
         token_to_kv_pool_allocator: BaseTokenToKVPoolAllocator,
         tp_cache_group: torch.distributed.ProcessGroup,
-        tp_rank: int,
-        pp_rank: int,
-        dp_rank: Optional[int],
         page_size: int,
         hicache_ratio: float,
         hicache_size: int,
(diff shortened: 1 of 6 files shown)


xiezhq-hermann:
Thank you for the PR, how about we capsulate everything about observability into --enable-metrics? I think that would simplify some implementation and make it more coherent to the collection of other metrics.

pansicheng:
> Thank you for the PR, how about we capsulate everything about observability into --enable-metrics? I think that would simplify some implementation and make it more coherent to the collection of other metrics.

@xiezhq-hermann Sure! PTAL

python/sglang/srt/mem_cache/hiradix_cache.py
```diff
@@ -31,12 +32,16 @@ def __init__(
         req_to_token_pool: ReqToTokenPool,
         token_to_kv_pool_allocator: BaseTokenToKVPoolAllocator,
         tp_cache_group: torch.distributed.ProcessGroup,
+        tp_rank: int,
```
xiezhq-hermann: if we choose to pass these parameters, we should not be querying them again in `_generate_storage_config`, which one would be better?

pansicheng: Upon review, the tp_rank and dp_rank being passed down in the metrics labels are global ranks, whereas cache_controller retrieves ranks only within the dp group. To ensure storage reusability across dp groups, it's preferable to rely on _generate_storage_config. Since the incoming ranks are global, they require conversion. The latest changes have standardized the use of _generate_storage_config to obtain the correct ranks.

python/sglang/srt/metrics/collector.py
```diff
@@ -551,3 +551,100 @@ def observe_inter_token_latency(self, internval: float, num_new_tokens: int):
...
+            self.prefetched_tokens_total.labels(**self.labels).inc(prefetched_tokens)
+
+    def log_backuped_tokens(self, backuped_tokens: int):
+        if backuped_tokens > 0:
+            self.backuped_tokens_total.labels(**self.labels).inc(backuped_tokens)
+
+    def _log_histogram(self, histogram, data: Union[int, float]):
+        histogram.labels(**self.labels).observe(data)
+
+    def log_storage_metrics(self, storage_metrics: StorageMetrics):
```
xiezhq-hermann: right now this is very 3fs specific, it would fail if other backends does not implement `get_stats` or implemented it slightly differently.

pansicheng: @xiezhq-hermann Add handling logic for cases where the get_stats interface is not implemented and return None. Assert if isinstance(storage_metrics, StorageMetrics) is false

xiezhq-hermann (approved):

PR #9968, opened 2025-09-03
fix:incorrect throughput calculation caused by a TTFT metric computation error in bench_one_batch_server

JoyFuture:
## Motivation

When KV cache capacity is small, long prompts cannot be fully prefetched in one shot, causing prefill–decode–prefill interleaving. The existing TTFT logic in bench_one_batch_server measured start → last first-token wall time and thus included decode time, inflating TTFT and skewing throughput metrics. This PR revises TTFT calculation so it only accumulates elapsed time segments that end at a first-token event, effectively excluding decode windows.
[#9967](https://github.com/sgl-project/sglang/issues/9967)

## Modifications

Scope: client-side benchmark script only (python/sglang/bench_one_batch_server.py). No runtime/server changes.

Logic change: instead of setting ttft = now - tic at the last “first-token” event, we accumulate the elapsed time only when a first-token event is observed, and reset the segment start.

## Accuracy Tests

N/A — This change does not affect model forward/evaluation logic or outputs.

## Benchmarking and Profiling

Model: Qwen3-8B

Batch size: 32

Input/Output: 2048 / 1024

## Repro Steps

Small KV pool (interleaving prefill/decode)
Server (reduced KV cache capacity):

```
  --model-path /llm/HFModels/Qwen3-8B/ \
  --disable-radix-cache --trust-remote-code \
  --host 127.0.0.1 --port 30001 \
  --mem-fraction-static 0.3
```

Same benchmark command.

Before fix (incorrect TTFT):

```
latency: 35.14 s
ttft: 20.52 s
last generation throughput: 876.83 tok/s
input throughput: 3193.53 tok/s
output throughput: 2242.14 tok/s
```

After fix (decode excluded from TTFT):

```
latency: 35.08 s
ttft: 4.23 s
last generation throughput: 876.98 tok/s
input throughput: 15499.39 tok/s
output throughput: 1062.05 tok/s
```

## Checklist

### DIFF
--- a/python/sglang/bench_one_batch_server.py
+++ b/python/sglang/bench_one_batch_server.py
@@ -190,6 +190,7 @@ def run_one_case(
 
     # The TTFT of the last request in the batch
     ttft = 0.0
+    last_time = tic
     for chunk in response.iter_lines(decode_unicode=False):
         chunk = chunk.decode("utf-8")
         if chunk and chunk.startswith("data:"):
@@ -204,7 +205,8 @@ def run_one_case(
                 or data["meta_info"]["finish_reason"]["type"] == "length"
             )
             if data["meta_info"]["completion_tokens"] == 1:
-                ttft = time.perf_counter() - tic
+                ttft += time.perf_counter() - last_time
+            last_time = time.perf_counter()
 
     latency = time.perf_counter() - tic
     input_throughput = batch_size * input_len / ttft


JoyFuture:
@hnyls2002 Hi , could you please take a look at my PR when you have time? Thanks!

zheng1:
This fix causes another problem: TTFT is computed too low.
See my test below for more details.

```
(llm-test) (base) root@h20-02:/data01/llm-test# python3 -m sglang.bench_one_batch_server --model-path /data01/models/DeepSeek-V3.1/ --base-url http://10.0.18.1:30000 --tokenizer-path /data01/models/DeepSeek-V3.1/ --batch-size 128 --input-len 2048 --output-len 2048 --skip-warmup
`torch_dtype` is deprecated! Use `dtype` instead!
#Input tokens: 262144
#Output tokens: 262144
batch size: 128
input_len: 2048
output_len: 2048
latency: 138.98 s
ttft: 1.18 s
last generation throughput: 10.33 tok/s
input throughput: 221857.86 tok/s
output throughput: 1902.34 tok/s
```
In fact, when PD is separated, TTFT is not 1.18s. The sglang-router buffers the request in memory after receiving it, then forwards it to the prefill. The client receives a response first, which causes the TTFT anomaly to appear smaller.

hnyls2002:
Thanks @JoyFuture! bench_one_batch_server was rewritten with new TTFT semantics, so this fix no longer
  applies - closing. Please reopen if the throughput number is still off.

PR #9976, opened 2025-09-03
[BugFix] Fix incorrect hidden_states_tensor in pd disaggregation + eagle

ZeldaHuang:
## Motivation
When use pd disaggregation with eagle , we should transfer the hidden states in `spec_info`, But now we transfer the  hidden states in target model's `logits_output`.

## Checklist


python/sglang/srt/disaggregation/prefill.py
```diff
@@ -410,9 +410,14 @@ def process_batch_result_disagg_prefill(
                     last_hidden_index = (
                         hidden_state_offset + extend_input_len_per_req[i] - 1
                     )
-                    req.hidden_states_tensor = (
-                        logits_output.hidden_states[last_hidden_index].cpu().clone()
-                    )
+                    if self.spec_algorithm.is_eagle():
```
ShangmingCai: I think MTP will make this condition True as well, at least it is True for Deepseek V3/R1. Is there a way to make it work for both worlds? When using MTP, does `batch.spec_info.hidden_states[i]` equal None in this case?

ZeldaHuang: I think in MTP, we also need to transfer the `batch.spec_info.hidden_states[i]`, cause we get the get 1 token from draft prefill in P node, ` batch.spec_info.hidden_states` will be updated to draft model's hidden states in [https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/speculative/eagle_worker.py#L940](https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/speculative/eagle_worker.py#L940)

ShangmingCai: If that's so, do we still need this `else` logic? Just wondering if the current changes fit both MTP and vanilla EAGLE's requirements? I will trigger the CI, but it would be better if you could verify whether it works for both cases.

BTW, these lines look like they might be the same thing:
https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/speculative/eagle_worker.py#L1043-L1048

ShangmingCai:
@JustinTong0323 Can you help reviewing this PR？

ShangmingCai (commented):
@JustinTong0323 BTW, I am OK with this change if the goal is only to avoid raising an error. More metadata that needs to be transmitted in the PD module will be addressed in other PRs.

python/sglang/srt/disaggregation/prefill.py
```diff
@@ -412,9 +412,14 @@ def process_batch_result_disagg_prefill(
                     last_hidden_index = (
                         hidden_state_offset + extend_input_len_per_req[i] - 1
                     )
-                    req.hidden_states_tensor = (
-                        logits_output.hidden_states[last_hidden_index].cpu().clone()
-                    )
+                    if self.spec_algorithm.is_eagle3():
```
JustinTong0323, line 424: why not `spec_algorithm.is_eagle()`?
ref:
https://github.com/sgl-project/sglang/blob/99757cc3e68418f48713e69e5543a5783abcd84d/python/sglang/srt/speculative/spec_info.py#L13-L17

ZeldaHuang: As mentioned in https://github.com/sgl-project/sglang/issues/10059, the error only happened in eagle3, which has three aux hidden_states in logits_output, this pr avoid raising it. As for the accept rate of first draft decode, will be fixed in other PRs.

JustinTong0323 (approved):

PR #8882, opened 2025-08-07
Add draft attention backend selection for draft model running 

yilian49:
## Motivation

The current implementation of the draft model automatically chooses the same attention backend as the target model backend. This PR allows an optional argument for a separate draft model backend selection. 

## Modifications
A new server argument is added for draft model backend selection. If not provided, it will default to the target model attention backend.

## Checklist

- [x] Format your code according to the [Code Formatting with Pre-Commit](https://docs.sglang.ai/references/contribution_guide.html#code-formatting-with-pre-commit).
- [x] Add unit tests as outlined in the [Running Unit Tests](https://docs.sglang.ai/references/contribution_guide.html#running-unit-tests-adding-to-ci).
- [ ] Update documentation / docstrings / example tutorials as needed, according to [Writing Documentation](https://docs.sglang.ai/references/contribution_guide.html#writing-documentation-running-docs-ci).
- [ ] Provide throughput / latency benchmark results and accuracy evaluation results as needed, according to [Benchmark and Profiling](https://docs.sglang.ai/references/benchmark_and_profiling.html) and [Accuracy Results](https://docs.sglang.ai/references/accuracy_evaluation.html).
- [ ] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co-author when merging the PR.
- [ ] Please feel free to join our Slack channel at https://slack.sglang.ai to discuss your PR.

PR #9981, opened 2025-09-03
Add speculator attention backend switch

cicirori:
- Add --speculative-verify-backend parameter to control which backend (prefill or decode) is used for speculative decoding operations
- Support both TARGET_VERIFY and DRAFT_EXTEND forward modes
- Default to prefill backend for backward compatibility
- Enable users to choose decode backend for potential performance gains

🤖 Generated with [Claude Code](https://claude.ai/code)

## Checklist


python/sglang/srt/models/deepseek_v2.py
```diff
@@ -992,6 +992,15 @@ def _dispatch_mla_subtype():
...
+            forward_batch.forward_mode.is_target_verify()
+            or forward_batch.forward_mode.is_draft_extend()
+        ):
+            # Use the specified backend for speculative operations (both verify and draft extend)
+            if global_server_args_dict["speculative_attention_backend"] == "decode":
+                attention_backend = global_server_args_dict["decode_attention_backend"]
+            else:  # default to prefill
+                attention_backend = global_server_args_dict["prefill_attention_backend"]
         else:
             attention_backend = global_server_args_dict["prefill_attention_backend"]
```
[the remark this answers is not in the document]
cicirori: let's keep here easy to read.

python/sglang/srt/server_args.py
```diff
@@ -1512,6 +1513,13 @@ def add_cli_args(parser: argparse.ArgumentParser):
             help="The path of the draft model's small vocab table.",
             default=ServerArgs.speculative_token_map,
         )
+        parser.add_argument(
```
hnyls2002, line 1516: "backend" is confusing here, as we usually refer flashinfer/fa3/triton as the backend. Rename it to `--speculative-attention-mode` or `--speculative-attention-imp`

hnyls2002 (changes requested):

Fridge003:
@cicirori @hnyls2002 
Do you think it's possible to make draft attention backend different from both prefill backend and decode backend? 
So that we can use any attention kernel we want for draft attention 
Ref: #8882

test/srt/test_hybrid_attn_backend.py
```diff
@@ -132,5 +132,51 @@ def get_server_args(cls):
         ]
 
 
+class TestHybridAttnBackendSpeculativeDecodingPrefillBackend(TestHybridAttnBackendBase):
```
Fridge003, line 135: Can we remove the test `TestHybridAttnBackendSpeculativeDecoding` above, since both prefill backend and decode backend will be tried by draft model?

zhyncs: @Fridge003 @cicirori Can we do this in follow up? Thanks!

Fridge003: @zhyncs  #10149

PR #7149, opened 2025-06-13
Enable native ModelOpt quantization support (1/3) 

Edwardf0t1:
## Motivation

- We would like to add native `modelopt` quantization support in SGLang, by allowing modelopt quantization configurations as an argument in SGLang's high level API, e.g.,
`llm = sgl.Engine(model_path="TinyLlama/TinyLlama-1.1B-Chat-v1.0", modelopt_quant="fp8")`  

- Users would have the option to choose ModelOpt to quantize their model by calling `modelopt` quantization API and deploy with SGLang runtime.

- This is part of the efforts to promote modelopt as the default quantization library in SGLang. See https://github.com/sgl-project/sglang/issues/5251 

## Modifications

This is the first of the PR series to add native `modelopt` quantization support. The major changes are the following in this PR:

- Add an argument `modelopt_quant` in `ServerArgs` and corresponding functions.
- Create a model loader class `ModelOptModelLoader` to load model with HF API for `modelopt` to quantize, and modelopt quantization APIs.  
- Add a `modelopt_utils.py` to host modelopt related constants for now. 
- Add a unit test for `ModelOptModelLoader`.

The 2nd PR and 3rd PR are also ready for review: 

- https://github.com/sgl-project/sglang/pull/9991 
- https://github.com/sgl-project/sglang/pull/10154

## Checklist

- [x] Format your code according to the [Code Formatting with Pre-Commit](https://docs.sglang.ai/references/contribution_guide.html#code-formatting-with-pre-commit).
- [x] Add unit tests as outlined in the [Running Unit Tests](https://docs.sglang.ai/references/contribution_guide.html#running-unit-tests-adding-to-ci).
- [ ] Update documentation / docstrings / example tutorials as needed, according to [Writing Documentation](https://docs.sglang.ai/references/contribution_guide.html#writing-documentation-running-docs-ci).
- [ ] Provide throughput / latency benchmark results and accuracy evaluation results as needed, according to [Benchmark and Profiling](https://docs.sglang.ai/references/benchmark_and_profiling.html) and [Accuracy Results](https://docs.sglang.ai/references/accuracy_evaluation.html).
- [ ] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co-author when merging the PR.
- [ ] Please feel free to join our Slack channel at https://slack.sglang.ai to discuss your PR.

PR #9991, opened 2025-09-04
Enable native ModelOpt quantization support (2/3)

Edwardf0t1:
This is the second PR in a three-part series to enable native ModelOpt quantization in SGLang. It includes changes from the first PR (https://github.com/sgl-project/sglang/pull/7149) and will be rebased once the first PR is merged.

## Motivation

We aim to enhance SGLang's quantization capabilities, making ModelOpt integration more robust and user-friendly while providing checkpoint persistence for better performance in production environments.

## Modifications

- Created `_setup_modelopt_quantization()` and added calibration functionalities.
- Added `modelopt_checkpoint_restore_path` and `modelopt_checkpoint_save_path` parameters to both `ModelConfig` and `ServerArgs`. These allow users to save and restore quantized checkpoints, avoiding re-quantization on subsequent runs
- Improved error handling during the ModelOpt quantization process.
- Added mode unit tests in `test_modelopt_loader.py` to verify the ModelOpt functionality.

The 3rd PR are also ready for review: https://github.com/sgl-project/sglang/pull/10154 

## Checklist


zhyncs:
hi @Edwardf0t1 can you help fix the conflicts? thanks

Edwardf0t1:
> hi @Edwardf0t1 can you help fix the conflicts? thanks

@zhyncs Just rebased and resolved the conflicts. Could you or @Qiaolin-Yu help review the PR? Thanks.

python/sglang/srt/configs/model_config.py
```diff
@@ -443,36 +452,52 @@ def _parse_quant_hf_config(self):
...
                 )
                 with open(quant_config_file) as f:
                     quant_config_dict = json.load(f)
-                json_quant_configs = quant_config_dict["quantization"]
-                quant_algo = json_quant_configs.get("quant_algo", None)
-                if quant_algo == "MIXED_PRECISION":
-                    quant_cfg = {"quant_method": "w4afp8"}
-                else:
-                    quant_cfg = modelopt_quant_config
+                quant_cfg = self._parse_modelopt_quant_config(quant_config_dict)
```
jingyu-ml, line 516: Is W4AFP8 a quantization format from MO?, if not, i think we can just rename this function to `_parse_quant_config `

python/sglang/srt/entrypoints/engine.py
```diff
@@ -123,6 +123,8 @@ def __init__(self, **kwargs):
         self.port_args = PortArgs.init_new(server_args)
         logger.info(f"{server_args=}")
 
+        print(f"Launching subprocesses with {server_args=}")
```
jingyu-ml: I don’t think using print is a good approach in distributed inference code.

jingyu-ml:
I think we should add example code in this PR to demonstrate how to use `modelopt_checkpoint_restore_path` or other new functions, so users can understand without needing deep context.

python/sglang/srt/model_loader/loader.py
```diff
@@ -1664,9 +1729,181 @@ def load_model_with_cpu_quantization(
...
+                tokenizer.padding_side = "left"
+
+            from modelopt.torch.utils.dataset_utils import (
+                create_forward_loop,
+                get_dataset_dataloader,
+            )
+
+            # Create calibration dataloader
+            calib_dataloader = get_dataset_dataloader(
+                dataset_name="cnn_dailymail",
```
jingyu-ml: It’s better to make this configurable as TODOs below.

python/sglang/srt/model_loader/loader.py
```diff
@@ -1664,9 +1729,181 @@ def load_model_with_cpu_quantization(
...
+                num_samples=512,  # TODO: Consider making this configurable
+                device=model.device,
+                include_labels=False,
+            )
+
+            calibrate_loop = create_forward_loop(dataloader=calib_dataloader)
+
+            # Apply quantization
+            mtq.quantize(model, quant_cfg, forward_loop=calibrate_loop)
+            mtq.print_quant_summary(model)
```
jingyu-ml: Same as above, only run print_quant_summary on a single rank (print once in rank 0).

python/sglang/srt/model_loader/loader.py
```diff
@@ -474,12 +478,73 @@ def download_model(self, model_config: ModelConfig) -> None:
...
     def load_model(
         self,
         *,
         model_config: ModelConfig,
         device_config: DeviceConfig,
     ) -> nn.Module:
+
+        if hasattr(model_config, "modelopt_quant") and model_config.modelopt_quant:
+            # Load base model using shared method
+            model = self._load_modelopt_base_model(model_config)
```
jingyu-ml, line 568: I’m a bit confused about this function, what’s the detailed purpose of designing this function? Is the goal to merge all MO-related model loading functions into one, instead of modifying each model’s loading definition separately?

python/sglang/srt/configs/model_config.py
Edwardf0t1: Actually we have w4afp8 recipe in modelopt, e.g., `w4a8_nvfp4_fp8` and `w4a8_mxfp4_fp8`.

python/sglang/srt/model_loader/loader.py
Edwardf0t1: Good point, updated.

python/sglang/srt/model_loader/loader.py
Edwardf0t1: done

python/sglang/srt/entrypoints/engine.py
Edwardf0t1: Good point, updated with `rank0_log()`

python/sglang/srt/model_loader/loader.py
Edwardf0t1: The primary purpose is to eliminate code duplication between two different loaders that both need to load models for ModelOpt processing.

The function handles ModelOpt's unique loading requirements such as `accelerate` dependency (for device mapping); empty weights initialization (memory efficiency during quantization); auto device mapping (for distributed/multi-GPU setups); memory management (CPU fallback handling).

There are two usages: 1) `DefaultModelLoader` usage for loading pre-quantized models, and 2) `ModelOptModelLoader` usage for full quantization + export flow.

Edwardf0t1:
> I think we should add example code in this PR to demonstrate how to use `modelopt_checkpoint_restore_path` or other new functions, so users can understand without needing deep context.

The usage is covered in unit tests: `test/srt/model_loader/test_modelopt_loader.py`

python/sglang/srt/server_args.py
```diff
@@ -151,6 +151,13 @@ class ServerArgs:
     load_format: str = "auto"
     model_loader_extra_config: str = "{}"
     trust_remote_code: bool = False
+    dtype: str = "auto"
```
Qiaolin-Yu: dtype, kv_cache_dtype, quantization already exist

test/srt/model_loader/test_modelopt_loader.py
```diff
@@ -0,0 +1,810 @@
...
+
+import os
+import sys
+import unittest
+from unittest.mock import MagicMock, patch
+
+import torch.nn as nn
+
+# Add the sglang path for testing
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), "../../../python"))
```
Qiaolin-Yu: Check if this is really needed

test/srt/model_loader/test_modelopt_loader.py
```diff
@@ -0,0 +1,810 @@
...
+            )
+
+            # Verify print_quant_summary was called
+            mock_mtq.print_quant_summary.assert_called_once_with(self.mock_base_model)
+
+            # Verify eval() was called on the returned model
+            self.mock_base_model.eval.assert_called()
+
+            # Verify we get back the expected model
+            self.assertEqual(result_model, self.mock_base_model)
```
Qiaolin-Yu: When does `mock_base_model` become `tinyllama`

python/sglang/srt/server_args.py
Edwardf0t1: Removed.

test/srt/model_loader/test_modelopt_loader.py
Edwardf0t1: It's needed to make sure the tests run with the current development branch rather than the default installed sglang library.  
We can also remove it but need to set  `PYTHONPATH=python` when run the tests.

test/srt/model_loader/test_modelopt_loader.py
Edwardf0t1: Actually `mock_base_model` never actually becomes `TinyLlama` - it remains a mock object throughout the test. What happens is: 

1. `self.model_path = "TinyLlama/TinyLlama-1.1B-Chat-v1.0"` (just a string path)
2. `self.mock_base_model = MagicMock(spec=nn.Module)` (a mock object that pretends to be a PyTorch module)
3. The TinyLlama model loading is **mocked** with: `with patch.object(loader, "_load_modelopt_base_model", return_value=self.mock_base_model):`

 `self.mock_base_model.eval.return_value = self.mock_base_model` this was configured earlier so `result_model == self.mock_base_model` is `True`.

jingyu-ml (approved):

python/sglang/srt/entrypoints/engine.py
```diff
@@ -126,6 +126,8 @@ def __init__(self, **kwargs):
         self.port_args = PortArgs.init_new(server_args)
         logger.info(f"{server_args=}")
 
+        logger.info("Launching subprocesses for distributed inference...")
```
Qiaolin-Yu: Why do we need to add this?

python/sglang/srt/model_loader/loader.py
```diff
@@ -1764,6 +1764,96 @@ def __init__(self, load_config: LoadConfig):
...
+            except Exception as e:
+                logger.warning(
+                    f"Failed to restore from {quantized_ckpt_restore_path}: {e}"
+                )
+                rank0_log("Proceeding with calibration-based quantization...")
+
+        # Set up calibration-based quantization
+        try:
+            # Left padding tends to work better for batched generation with decoder-only LMs
+            with suppress(Exception):
```
Qiaolin-Yu, line 1818: Why does it need to suppress the exception?

python/sglang/srt/model_loader/loader.py
```diff
@@ -1808,33 +1897,26 @@ def load_model(
...
+                tokenizer,
+                quant_cfg,
+                quantized_ckpt_restore_path=quantized_ckpt_restore_path,
+                quantized_ckpt_save_path=quantized_ckpt_save_path,
+            )
         except Exception as e:
-            logger.error(f"Error during ModelOpt mtq.quantize call: {e}")
-            raise
-        mtq.print_quant_summary(model)
+            logger.warning(f"ModelOpt quantization failed: {e}")
```
Qiaolin-Yu, line 1918: Why not raise exception here?

python/sglang/srt/entrypoints/engine.py
Edwardf0t1: This was added for debugging. Removed.

python/sglang/srt/model_loader/loader.py
Edwardf0t1: Not all tokenizers support the `padding_side` attribute: Some tokenizer implementations may not have this attribute or may not allow it to be modified. If this modification fails, the quantization process should still continue, since it's not essential.

python/sglang/srt/model_loader/loader.py
Edwardf0t1: This is in the `ModelOptModelLoader.load_model()` method, which is designed to load a model with optional quantization. If quantization fails, the system falls back to loading the unquantized model rather than completely failing.

Qiaolin-Yu (approved):
LGTM

Edwardf0t1:
> LGTM

Thanks @Qiaolin-Yu for the review and approval.  
The failed CI tests seem unrelated to the PR. Could we unblock to merge? @zhyncs @Ying1123 @merrymercy

PR #10000, opened 2025-09-04
[router] clean up dependency injector to use ctx

slin1237:
## Motivation
both http and grpc routers (pd or regular) takes considerable amount of input and configurations one by one, while those are all in `ctx` from `App`, we should just pass in `ctx` instead

## Modifications
1. modify grpc regular and PD router to accept ctx and use it to initialize router
2. modify http regular and PD router to accept ctx and use it to initialize router

## Checklist


zhyncs:
10k PRs! HUGE :)

PR #10007, opened 2025-09-04
Support copying tensor from cpu to gpu without using copy engines

fzyzcjy:
## Checklist


fzyzcjy:
The sgl-kernel CI fails, but it is also failing on main: https://github.com/sgl-project/sglang/commit/f78b7fd16dbfe32c2ee73c1f3fef49fc1257b27f, https://github.com/sgl-project/sglang/actions/runs/17430655251/job/49517819550. since the new code does not touch existing code at all (and is tested e2e), maybe it is safe


image

Alcanderian (approved):

sgl-kernel/python/sgl_kernel/elementwise.py
```diff
@@ -367,3 +367,7 @@ def downcast_fp8(
     torch.ops.sgl_kernel.downcast_fp8(
         k, v, k_out, v_out, k_scale, v_scale, loc, mult, offset, get_cuda_stream()
     )
+
+
+def copy_to_gpu_no_ce(input: List[int], output: torch.Tensor):
```
ispobock, line 372: could you add a unit test and also include it to https://github.com/sgl-project/sglang/blob/main/sgl-kernel/python/sgl_kernel/__init__.py ?

fzyzcjy: good point, but I am rushing for the ddl, maybe I will add it later after finishing the ddl :/

fzyzcjy: done add to init

PR #10020, opened 2025-09-04
fix is_sm*_supported functions check for non CUDA devices

polisettyvarma:
## Motivation

to support XPU.

## Modifications

check cuda device available or not before checking capabilities.

## Accuracy Tests

Not Applicable

## Benchmarking and Profiling

Not Applicable

## Checklist

### DIFF
--- a/python/sglang/srt/utils.py
+++ b/python/sglang/srt/utils.py
@@ -174,6 +174,8 @@ def is_blackwell():
 
 @lru_cache(maxsize=1)
 def is_sm100_supported(device=None) -> bool:
+    if not is_cuda():
+        return False
     return (torch.cuda.get_device_capability(device)[0] == 10) and (
         torch.version.cuda >= "12.8"
     )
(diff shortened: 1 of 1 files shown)


polisettyvarma:
@Edwardf0t1 @kushanam @BBuf @ch-wan @HaiShaw @ispobock @zhyncs @Ying1123 @merrymercy @Alcanderian please review

deepvars (approved):

polisettyvarma:
@mingfeima please review

mingfeima:
we have aligned the strategy that we won't need this API for xpu, aka. we don't not fix `is_sm_*_supported` (which does not make too much sense for non-CUDA devices), we fix where it is called.

polisettyvarma:
> we have aligned the strategy that we won't need this API for xpu, aka. we don't not fix `is_sm_*_supported` (which does not make too much sense for non-CUDA devices), we fix where it is called.

another way to look at it is instead of fixing in many places, fix the function and not worry about future usages

mingfeima:
don't need this one, as we won't call `is_sm*_suported` on non-CUDA devices. on intel gpus, we should avoiding calling these utils, instead of fixing them directly.

polisettyvarma:
@mingfeima if we don't fix this way, need to fix in multiple places and files https://github.com/search?q=repo%3Asgl-project%2Fsglang%20is_sm&type=code

polisettyvarma:
fixed in #10130

PR #10032, opened 2025-09-04
Fix RMSNorm API CALL mismatch issue.

sogalin:
## Motivation
With updating of the vllm, we see API mismatch in SGLang.

## Modifications
Two new arguments were introduced into the vllm::fused_add_rms_norm(). Need to revise the usage in SGLang.

## Accuracy Tests
SGLANG_USE_AITER=0 python3 -m sglang.launch_server --model-path /data2/Qwen3-30B-A3B-Thinking-2507/ --tp 8 --trust-remote-code --chunked-prefill-size 130172 --max-running-requests 128 --mem-fraction-static 0.85 --attention-backend aiter

python3 benchmark/gsm8k/bench_sglang.py --num-questions 2000 --parallel 2000
Accuracy: 0.849
Invalid: 0.000
Latency: 54.786 s
Output throughput: 5336.137 token/s

## Checklist


HaiShaw (changes requested):
Let move away from explicit vllm builds.
cc @kkHuang-amd

HaiShaw (approved):

PR #10036, opened 2025-09-04
[FIX] CUDA_VERSION detection automatically

johnnynunez:
Before:
image
after:
image

If not must be forced, example:

```bash
export CUDA_VERSION=13.0


pip wheel . --no-build-isolation --wheel-dir "${PIP_WHEEL_DIR}" \
  --config-settings=cmake.args="-G;Ninja" \
  --config-settings=cmake.define.CUDA_VERSION="${CUDA_VERSION}" \
```


johnnynunez:
image

https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html

Issue #8250, opened 2025-07-22
[Feature] Support disabling of `any_whitespace` for XGrammar JSON structured outputs


lulor:
### Checklist

- [x] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 2. Please use English, otherwise it will be closed.

### Motivation

When using JSON structured outputs, the generation can get stuck on endlessly repeating whitespace characters (#2216, #2414, #1438, dottxt-ai/outlines#691). For the `outlines` and `llguidance` backends, it is possible to avoid this issue by leveraging the `constrained_json_whitespace_pattern` server argument, but for `xgrammar` such parameter is [ignored](https://github.com/sgl-project/sglang/blob/e50109f2edfec7cc48a56c41b05fcaef3190087f/python/sglang/srt/constrained/base_grammar_backend.py#L183).

XGrammar supports a similar `any_whitespace` argument that, although less flexible, would at least allow to avoid the infinite loop issue in generation (mlc-ai/xgrammar#345). It would be useful for SGLang to expose such parameter as a server argument (something like `disable-any-whitespace`), especially considering that XGrammar is the default and recommended backend for constrained generation. Furthermore, I think it would be useful if the documentation mentioned that the `constrained-json-whitespace-pattern` is not supported for XGrammar.

I'm happy to open a small PR with these changes (which should be quite minimal, unless I'm missing something).

### Related resources

vllm-project/vllm#15252

aflah02:
+1 This is something I'm facing as well

justadogistaken:
Hi, I encountered the problem of generating endlessly. And I tried your solution. But it's not working.
https://github.com/sgl-project/sglang/issues/11550

PR #10037, opened 2025-09-04
[draft] Use jsonschema to constrain required or specific tool choice

TJ5:
## Checklist

### DIFF
--- a/python/sglang/srt/entrypoints/openai/serving_chat.py
+++ b/python/sglang/srt/entrypoints/openai/serving_chat.py
@@ -746,10 +746,9 @@ class OpenAIServingChat(OpenAIServingBase):
             tool_calls = None
             if request.tool_choice != "none" and request.tools:
                 tool_call_parser = self.tokenizer_manager.server_args.tool_call_parser
-                result = self._process_tool_calls(
+                tool_calls, text, finish_reason = self._process_tool_calls(
                     text, request.tools, tool_call_parser, finish_reason, request.tool_choice
                 )
-                tool_calls, text, finish_reason = result.tool_calls, result.remaining_text, result.finish_reason
 
             choice_data = ChatCompletionResponseChoice(
                 index=idx,
--- a/python/sglang/srt/entrypoints/openai/serving_chat.py
+++ b/python/sglang/srt/entrypoints/openai/serving_chat.py
@@ -34,6 +34,8 @@ from sglang.srt.entrypoints.openai.utils import (
     to_openai_style_logprobs,
 )
 from sglang.srt.function_call.function_call_parser import FunctionCallParser
+from sglang.srt.function_call.utils import _get_json_schema_constraint
+from sglang.srt.function_call.json_array_detector import JsonArrayDetector
 from sglang.srt.managers.io_struct import GenerateReqInput
 from sglang.srt.managers.template_manager import TemplateManager
 from sglang.srt.managers.tokenizer_manager import TokenizerManager
@@ -186,14 +188,14 @@ class OpenAIServingChat(OpenAIServingBase):
             else:
                 tools = [item.function.model_dump() for item in request.tools]
 
-            # Use JSON parser for required or named tool choice
+            # Handle JSON schema constraint directly for required or named tool choice
             if request.tool_choice == "required" or isinstance(request.tool_choice, ToolChoice) or (isinstance(request.tool_choice, dict) and "function" in request.tool_choice):
-                tool_call_parser = "json"
+                json_schema = _get_json_schema_constraint(request.tools, request.tool_choice)
+                tool_call_constraint = ("json_schema", json_schema)
             else:
                 tool_call_parser = self.tokenizer_manager.server_args.tool_call_parser
-            
-            parser = FunctionCallParser(request.tools, tool_call_parser)
-            tool_call_constraint = parser.get_structure_constraint(request.tool_choice)
+                parser = FunctionCallParser(request.tools, tool_call_parser)
+                tool_call_constraint = parser.get_structure_constraint(request.tool_choice)
 
         # Use chat template
         if self.template_manager.chat_template_name is None:
@@ -989,19 +991,25 @@ class OpenAIServingChat(OpenAIServingBase):
     ):
         """Process tool calls in streaming response"""
         if index not in parser_dict:
-            # Use JSON parser for required or named tool choice
+            # Use JSON detector directly for required or named tool choice
             if request.tool_choice == "required" or isinstance(request.tool_choice, ToolChoice) or (isinstance(request.tool_choice, dict) and "function" in request.tool_choice):
-                tool_call_parser = "json"
+                # Store JsonArrayDetector directly for JSON schema constraints
+                parser_dict[index] = JsonArrayDetector()
             else:
                 tool_call_parser = self.tokenizer_manager.server_args.tool_call_parser
-            
-            parser_dict[index] = FunctionCallParser(
-                tools=request.tools,
-                tool_call_parser=tool_call_parser,
-            )
+                parser_dict[index] = FunctionCallParser(
+                    tools=request.tools,
+                    tool_call_parser=tool_call_parser,
+                )
+        
         parser = parser_dict[index]
 
-        normal_text, calls = parser.parse_stream_chunk(delta)
+        # Handle both FunctionCallParser and JsonArrayDetector
+        if isinstance(parser, JsonArrayDetector):
+            result = parser.parse_streaming_increment(delta, request.tools)
+            normal_text, calls = result.normal_text, result.calls
+        else:
+            normal_text, calls = parser.parse_stream_chunk(delta)
 
         # Yield normal text
         if normal_text:
@@ -1061,7 +1069,7 @@ class OpenAIServingChat(OpenAIServingBase):
 
     def _check_for_unstreamed_tool_args(
         self,
-        parser: FunctionCallParser,
+        parser: Union[FunctionCallParser, JsonArrayDetector],
         content: Dict[str, Any],
         request: ChatCompletionRequest,
         index: int,
@@ -1071,30 +1079,33 @@ class OpenAIServingChat(OpenAIServingBase):
         when generation finishes. This ensures tool calls are properly completed
         even if the model generates the final arguments in the last chunk.
         """
-        # Only check if we have tool calls and the parser has tracked data
+        # Get the detector - either from FunctionCallParser or directly if json detector
+        detector = parser.detector if hasattr(parser, 'detector') else parser
+        
+        # Only check if we have tool calls and the detector has tracked data
         if (
-            not hasattr(parser.detector, "prev_tool_call_arr")
-            or not parser.detector.prev_tool_call_arr
+            not hasattr(detector, "prev_tool_call_arr")
+            or not detector.prev_tool_call_arr
         ):
             return None
 
         if (
-            not hasattr(parser.detector, "streamed_args_for_tool")
-            or not parser.detector.streamed_args_for_tool
+            not hasattr(detector, "streamed_args_for_tool")
+            or not detector.streamed_args_for_tool
         ):
             return None
 
         # Get the last tool call that was being processed
-        tool_index = len(parser.detector.prev_tool_call_arr) - 1
-        if tool_index < 0 or tool_index >= len(parser.detector.streamed_args_for_tool):
+        tool_index = len(detector.prev_tool_call_arr) - 1
+        if tool_index < 0 or tool_index >= len(detector.streamed_args_for_tool):
             return None
 
         # Get expected vs actual arguments
-        expected_args = parser.detector.prev_tool_call_arr[tool_index].get(
+        expected_args = detector.prev_tool_call_arr[tool_index].get(
             "arguments", {}
         )
         expected_call = json.dumps(expected_args, ensure_ascii=False)
-        actual_call = parser.detector.streamed_args_for_tool[tool_index]
+        actual_call = detector.streamed_args_for_tool[tool_index]
 
         # Check if there are remaining arguments to send
         remaining_call = (
--- a/python/sglang/srt/entrypoints/openai/serving_chat.py
+++ b/python/sglang/srt/entrypoints/openai/serving_chat.py
@@ -69,6 +69,16 @@ class OpenAIServingChat(OpenAIServingBase):
         ):
             return "Tools cannot be empty if tool choice is set to required."
 
+        if (
+            not isinstance(request.tool_choice, str)
+            and request.tool_choice is not None
+            and request.tools
+        ):
+            tool_name = request.tool_choice.function.name
+            tool_exists = any(tool.function.name == tool_name for tool in request.tools)
+            if not tool_exists:
+                return f"Tool '{tool_name}' not found in tools list."
+
         max_output_tokens = request.max_completion_tokens or request.max_tokens
         server_context_length = self.tokenizer_manager.server_args.context_length
         if (
--- a/python/sglang/srt/entrypoints/openai/serving_chat.py
+++ b/python/sglang/srt/entrypoints/openai/serving_chat.py
@@ -434,7 +434,6 @@ class OpenAIServingChat(OpenAIServingBase):
                     constraint_value.model_dump(by_alias=True)
                 )
             elif constraint_type == "json_schema":
-                # Handle vLLM-style modified JSON schema constraint
                 sampling_params[constraint_type] = convert_json_schema_to_str(
                     constraint_value
                 )
--- a/python/sglang/srt/entrypoints/openai/serving_chat.py
+++ b/python/sglang/srt/entrypoints/openai/serving_chat.py
@@ -846,22 +846,26 @@ class OpenAIServingChat(OpenAIServingBase):
     ) -> tuple[Optional[List[ToolCall]], str, Dict[str, Any]]:
         """Process tool calls in the response"""
 
-        # Handle required tool choice
-        if tool_choice == "required" or (isinstance(tool_choice, ToolChoice) and tool_choice.type == "function"):
+        def get_tool_id(name: str, index: int) -> str:
+            if tool_call_parser == "kimi_k2" and name is not None:
+                return f"functions.{name}:{index}"
+            else:
+                return f"call_{uuid.uuid4().hex[:24]}"
+
+        def set_finish_reason_tool_calls():
             if finish_reason["type"] == "stop":
                 finish_reason["type"] = "tool_calls"
                 finish_reason["matched"] = None
+
+        # Handle required or named tool choice
+        if tool_choice == "required" or (isinstance(tool_choice, ToolChoice) and tool_choice.type == "function"):
+            set_finish_reason_tool_calls()
             try:
                 # For required tool choice, we expect a JSON array of tool calls
                 tool_call_data = json.loads(text)
                 tool_calls = []
                 for i, tool in enumerate(tool_call_data):
-                    # For Kimi-K2, align tool_call_id with the model format: functions.{name}:{index}
-                    if tool_call_parser == "kimi_k2" and tool['name'] is not None:
-                        tool_id = f"functions.{tool['name']}:0"
-                    else:
-                        tool_id = f"call_{uuid.uuid4().hex[:24]}"
-
+                    tool_id = get_tool_id(tool.get("name"), i)
                     tool_calls.append(
                         ToolCall(
                             id=tool_id,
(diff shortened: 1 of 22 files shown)


python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -69,6 +69,16 @@ def _validate_request(self, request: ChatCompletionRequest) -> Optional[str]:
         ):
             return "Tools cannot be empty if tool choice is set to required."
 
+        if (
+            not isinstance(request.tool_choice, str)
+            and request.tool_choice is not None
```
CatherineSue:
```
 and request.tool_choice is not None
```

i think this should be the first check?

python/sglang/srt/function_call/function_call_parser.py
```diff
@@ -172,6 +172,9 @@ def get_structure_constraint(
             strict_tag = self.get_structure_tag()
             return ("structural_tag", strict_tag)
         elif tool_choice == "required" or isinstance(tool_choice, ToolChoice):
+            json_schema = self.get_json_schema_constraint(tool_choice)
+            if json_schema is not None:
+                return ("json_schema", json_schema)
             ebnf = self.get_ebnf(tool_choice)
```
CatherineSue: remove L178 to L179?

python/sglang/srt/function_call/function_call_parser.py
```diff
@@ -172,6 +172,9 @@ def get_structure_constraint(
             strict_tag = self.get_structure_tag()
             return ("structural_tag", strict_tag)
         elif tool_choice == "required" or isinstance(tool_choice, ToolChoice):
+            json_schema = self.get_json_schema_constraint(tool_choice)
+            if json_schema is not None:
```
CatherineSue: what is json_schema is None? there is no handling of that

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -826,8 +841,47 @@ def _process_tool_calls(
...
+                        function=FunctionResponse(
+                            name=tool_choice["function"]["name"],
+                            arguments=text,
+                        ),
+                    )
+                ]
+                return tool_calls, "", finish_reason
+            except Exception as e:
+                logger.error(f"Tool call parsing error: {e}")
+                return None, text, finish_reason
```
CatherineSue: the finish_reason should be updated, see L887-889.

When there is a tool call in the final response, OpenAI-compatible server should return finish_reason as `tool_calls` instead of `stop`.

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -826,8 +841,47 @@ def _process_tool_calls(
...
         finish_reason: Dict[str, Any],
+        tool_choice: Optional[Union[str, Any]] = None,
     ) -> tuple[Optional[List[ToolCall]], str, Dict[str, Any]]:
         """Process tool calls in the response"""
+        if tool_choice == "required":
+            try:
+                tool_call_data = json.loads(text)
+                tool_calls = []
+                for i, call_info in enumerate(tool_call_data):
+                    tool_id = f"call_{uuid.uuid4().hex[:24]}"
```
CatherineSue: Please see L895-896:
```
for call_info in call_info_list:
                    # For Kimi-K2, align tool_call_id with the model format: functions.{name}:{index}
                    if tool_call_parser == "kimi_k2" and call_info.name is not None:
                        tool_id = f"functions.{call_info.name}:{call_info.tool_index}"
```

Kimi-K2 requires a specific format of the tool_call_id.

Maybe having this as a function so we don't have to do this block everywhere

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -826,8 +841,47 @@ def _process_tool_calls(
...
+                        index=0,
+                        function=FunctionResponse(
+                            name=tool_choice["function"]["name"],
+                            arguments=text,
+                        ),
+                    )
+                ]
+                return tool_calls, "", finish_reason
+            except Exception as e:
+                logger.error(f"Tool call parsing error: {e}")
```
CatherineSue: I am not sure what exceptions could raise here?

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -826,8 +841,47 @@ def _process_tool_calls(
...
+                            id=tool_id,
+                            index=i,
+                            function=FunctionResponse(
+                                name=call_info['name'], 
+                                arguments=json.dumps(call_info['parameters'], ensure_ascii=False)
+                            ),
+                        )
+                    )
+                return tool_calls, "", finish_reason
+            except Exception as e:
```
CatherineSue: it seems the only exception here is Json error? Please specific the exceptions here.

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -826,8 +841,47 @@ def _process_tool_calls(
         tools: List[Any],
         tool_call_parser: Optional[str],
         finish_reason: Dict[str, Any],
+        tool_choice: Optional[Union[str, Any]] = None,
     ) -> tuple[Optional[List[ToolCall]], str, Dict[str, Any]]:
```
CatherineSue: maybe have this as a NamedTuple? It became hard to tell what the this three fields mean.

python/sglang/srt/function_call/json_detector.py
```diff
@@ -0,0 +1,78 @@
...
+    
+    This detector is used when tool_choice="required" or a specific tool is named,
+    bypassing model-specific parsers in favor of direct JSON array parsing.
+    """
+
+    def __init__(self):
+        super().__init__()
+        # Configure for JSON array parsing
+        self.bot_token = "["
+        self.eot_token = ""
```
CatherineSue: eot_token should be `]`?

python/sglang/srt/entrypoints/openai/serving_chat.py
CatherineSue: Friendly ping.

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -826,9 +842,64 @@ def _process_tool_calls(
...
         tool_call_parser: Optional[str],
         finish_reason: Dict[str, Any],
+        tool_choice: Optional[Union[str, ToolChoice]] = None,
     ) -> tuple[Optional[List[ToolCall]], str, Dict[str, Any]]:
         """Process tool calls in the response"""
-        parser = FunctionCallParser(tools, tool_call_parser)
+
+        # Handle required tool choice
+        if tool_choice == "required":
+            if finish_reason["type"] == "stop":
```
CatherineSue, line 880: the correct logic should be detect if there is a tool call, then change the finish reason.

I feel it is better to handle this at last? Otherwise there are quite some repeated code blocks for this

python/sglang/srt/entrypoints/openai/serving_chat.py
CatherineSue: Friendly ping.

python/sglang/srt/function_call/function_call_parser.py
```diff
@@ -47,13 +48,16 @@ class FunctionCallParser:
...
     }
 
-    def __init__(self, tools: List[Tool], tool_call_parser: str):
-        detector: Type[BaseFormatDetector] = None
-        detector_class = self.ToolCallParserEnum.get(tool_call_parser)
-        if detector_class:
-            detector = detector_class()
+    def __init__(self, tools: List[Tool], tool_call_parser: str, tool_choice: Optional[Union[str, ToolChoice]] = None):
+        # Use JSON parsing for tool_choice="required" or named tool
+        if tool_choice == "required" or isinstance(tool_choice, ToolChoice) or (isinstance(tool_choice, dict) and "function" in tool_choice):
```
CatherineSue: why do we have a detector here?

python/sglang/srt/function_call/json_detector.py
```diff
@@ -0,0 +1,78 @@
...
+        Check if the given text contains a JSON tool call (array or single object).
+        
+        Args:
+            text: The text to check for tool calls
+            
+        Returns:
+            True if the text starts with a JSON array or object, False otherwise
+        """
+        stripped = text.strip()
+        return stripped.startswith("[") or stripped.startswith("{")
```
CatherineSue: Why is this check inconsistent with the `self.bot_token`?

python/sglang/srt/function_call/json_detector.py
```diff
@@ -0,0 +1,78 @@
...
+        Args:
+            new_text: The new chunk of text to parse
+            tools: List of available tools
+            
+        Returns:
+            StreamingParseResult with parsed tool calls
+        """
+        return super().parse_streaming_increment(new_text, tools)
+
+    def structure_info(self) -> callable:
```
CatherineSue: Please remove function calls that will not be used

python/sglang/srt/function_call/json_detector.py
```diff
@@ -0,0 +1,78 @@
...
+        """
+        return super().detect_and_parse(text, tools)
+
+    def build_ebnf(self, tools: List[Tool]) -> str:
+        """
+        Build an EBNF grammar for constrained generation.
+        This is not used for JSON schema constraints as they are handled
+        by the constraint backends directly.
+        """
+        return ""
```
CatherineSue: We should return a NotImplementedError instead of empty here.

python/sglang/srt/function_call/json_detector.py
```diff
@@ -0,0 +1,78 @@
...
+        Parse JSON tool calls using the base class implementation.
+        
+        Args:
+            text: The text to parse
+            tools: List of available tools
+            
+        Returns:
+            StreamingParseResult with parsed tool calls
+        """
+        return super().detect_and_parse(text, tools)
```
CatherineSue: Do we need to implement this function as it is already been implemented in the base class?

TJ5: Yes because it's abstract in the base class

python/sglang/srt/function_call/function_call_parser.py
TJ5: We are only using the jsondetector if the request specifies a required tool choice or a specific tool. In these two cases we override the usual model-specific detector with the jsondetector.

python/sglang/srt/function_call/json_detector.py
TJ5: detect_and_parse is abstract in base_format_detector so we have to implement it, even if there's no use for it.

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -846,22 +846,26 @@ def _process_tool_calls(
...
 
-        # Handle required tool choice
-        if tool_choice == "required" or (isinstance(tool_choice, ToolChoice) and tool_choice.type == "function"):
+        def get_tool_id(name: str, index: int) -> str:
+            if tool_call_parser == "kimi_k2" and name is not None:
+                return f"functions.{name}:{index}"
+            else:
+                return f"call_{uuid.uuid4().hex[:24]}"
+
+        def set_finish_reason_tool_calls():
```
CatherineSue: nit:
1. It is not a good practice to have nested functions defined. Please make them private and define as a separate function.

2. Do we need this function? I suppose we can simply convert this finish reason at the top of `_process_tool_calls`.

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -826,23 +844,54 @@ def _process_tool_calls(
...
+            set_finish_reason_tool_calls()
+            try:
+                # For required tool choice, we expect a JSON array of tool calls
+                tool_call_data = json.loads(text)
+                tool_calls = []
+                for i, tool in enumerate(tool_call_data):
+                    tool_id = get_tool_id(tool.get("name"), i)
+                    tool_calls.append(
+                        ToolCall(
+                            id=tool_id,
```
CatherineSue: nit: if tool_id only used here, we can do

`id = get_tool_id(tool.get("name"), i)`

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -23,10 +23,12 @@
     LogProbs,
     MessageProcessingResult,
     ToolCall,
+    ToolChoice,
     TopLogprob,
 )
 from sglang.srt.entrypoints.openai.serving_base import OpenAIServingBase
 from sglang.srt.entrypoints.openai.usage_processor import UsageProcessor
+from sglang.srt.function_call.core_types import ToolCallProcessingResult
```
CatherineSue: Let's not import from a downstream dependency? 1) `ToolCallProcessingResult` is not used anywhere in function_call but in entrypoints.openai. 2) Import from a different module is prone to cause circular dependency, not in this case though.

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -729,9 +746,10 @@ def _build_chat_response(
             tool_calls = None
             if request.tool_choice != "none" and request.tools:
                 tool_call_parser = self.tokenizer_manager.server_args.tool_call_parser
-                tool_calls, text, finish_reason = self._process_tool_calls(
```
CatherineSue, line 732: I believe we don't need to change this. Python unpacks it automatically

python/sglang/srt/function_call/json_detector.py
```diff
@@ -0,0 +1,76 @@
+from typing import List
+
+from sglang.srt.entrypoints.openai.protocol import Tool
+from sglang.srt.function_call.base_format_detector import BaseFormatDetector
+from sglang.srt.function_call.core_types import StreamingParseResult
+
+
+class JsonDetector(BaseFormatDetector):
```
CatherineSue: Is `JsonArrayDetector` a better name?

python/sglang/srt/function_call/json_detector.py
```diff
@@ -0,0 +1,76 @@
...
+        # Configure for JSON array parsing
+        self.bot_token = "["
+        self.eot_token = "]"
+        self.tool_call_separator = ", "
+
+    def has_tool_call(self, text: str) -> bool:
+        """
+        Check if the given text contains a JSON tool call (array or single object).
+        
+        Args:
```
CatherineSue: nit: we can remove Args and Returns in this file as the detailed info is already included in BaseFormatDetector

Args and Returns are useful when the logic is complex, here it is pretty straightforward

python/sglang/srt/function_call/function_call_parser.py
CatherineSue: Essentially this is changing the design of FunctionCallParser. It should not be changed unless necessary. Is it possible to separate the design here?

For instance, `JsonDetector` is only used during streaming + tool_choice specific or required. Why do we have to make the general API interface for this specific change?

python/sglang/srt/entrypoints/openai/serving_chat.py
TJ5: We can't just put it at the top because we only want to set the finish_reason if it has a tool call (returned from a parser) or if it's from a request that uses a jsonschema constraint. In this case I think since it's just 3 lines it's fine if we just copy it in the two places it's being used. When we last reviewed this it was being used in 3 places, but I merged the logic for handling required and named tool choices in serving_chat which makes it much more concise.

python/sglang/srt/function_call/function_call_parser.py
TJ5: Makes sense. The desired behavior is that the jsondetector is used if a tool choice is supplied that is 'required' or a named tool. In these cases we just switch the tool-call-parser to 'json' when calling the function call parser, so we don't need to change the API interface.

python/sglang/srt/function_call/core_types.py
```diff
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from typing import Callable, List, Optional
+from typing import Any, Callable, Dict, List, NamedTuple, Optional
```
CatherineSue: Please remove unused imports

python/sglang/srt/entrypoints/openai/serving_chat.py
CatherineSue: Friendly ping.

python/sglang/srt/function_call/function_call_parser.py
```diff
@@ -171,9 +173,9 @@ def get_structure_constraint(
         ):
             strict_tag = self.get_structure_tag()
             return ("structural_tag", strict_tag)
-        elif tool_choice == "required" or isinstance(tool_choice, ToolChoice):
-            ebnf = self.get_ebnf(tool_choice)
-            return ("ebnf", ebnf) if ebnf is not None else None
+        elif tool_choice == "required" or isinstance(tool_choice, ToolChoice) or (isinstance(tool_choice, dict) and "function" in tool_choice):
```
CatherineSue: Please remove the dict check. That is not one of the valid type of tool_choice

python/sglang/srt/function_call/json_array_detector.py
```diff
@@ -0,0 +1,58 @@
...
+        This is not used for JSON schema constraints as they are handled
+        by the constraint backends directly.
+        """
+        raise NotImplementedError("EBNF generation is not supported for JSON schema constraints.")
+
+    def parse_streaming_increment(self, new_text: str, tools: List[Tool]) -> StreamingParseResult:
+        """
+        Parse streaming JSON tool calls using the base class implementation.
+        """
+
```
CatherineSue: Please remove extra empty line.

python/sglang/srt/function_call/function_call_parser.py
```diff
@@ -45,6 +46,7 @@ class FunctionCallParser:
         "glm45": Glm4MoeDetector,
         "step3": Step3Detector,
         "gpt-oss": GptOssDetector,
+        "json": JsonArrayDetector,
```
CatherineSue: Adding a field here is misleading. From a user perspective, it makes me feel this should be used with a model, but it is not.

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -173,7 +186,12 @@ def _process_messages(
             else:
                 tools = [item.function.model_dump() for item in request.tools]
 
-            tool_call_parser = self.tokenizer_manager.server_args.tool_call_parser
+            # Use JSON parser for required or named tool choice
+            if request.tool_choice == "required" or isinstance(request.tool_choice, ToolChoice) or (isinstance(request.tool_choice, dict) and "function" in request.tool_choice):
+                tool_call_parser = "json"
```
CatherineSue: I suppose we don't need a specific tool_call_parser here for the tool call constraint, correct?

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -820,45 +842,81 @@ def _process_response_logprobs(self, ret_item: Dict[str, Any]) -> ChoiceLogprobs
...
+                        )
+                    )
+                return ToolCallProcessingResult(tool_calls, "", finish_reason)
+            except json.JSONDecodeError as e:
+                logger.error(f"Tool call parsing error: {e}")
+                return ToolCallProcessingResult(None, text, finish_reason)
+        
+        # Use JSON parser for required or named tool choice
+        if tool_choice == "required" or isinstance(tool_choice, ToolChoice) or (isinstance(tool_choice, dict) and "function" in tool_choice):
+            tool_call_parser = "json"
```
CatherineSue, line 912: Same here, why do we need to pass a json tool call parser here? The situation has already been dealt with at L862?

python/sglang/srt/function_call/function_call_parser.py
CatherineSue: `get_json_schema_constraint` is essentially static. It doesn't use any detector-specific logic.

I feel right now it is mixed with ebnf and get_structural_tag as they are both stateful. 

Can we move this to utils.py as a utility function so we don't need to init a FunctionCallParser in _process_messages for tool_choice is required or ToolChoice?

In serving_chat.py we can directly call get_json_schema_constraint

CatherineSue (changes requested):
Action Items:
- [ ] Rebase with the latest main.
- [ ] Resolve all the design comments. Please reply to all of them one by one regarding the changes.
 (The most important one: if JsonArrayDetector is only for our cases for tool_choice + streaming, then we don't need it as a Enum)

- [ ] Update all the tests in the PR description. Tests should include: 
  - [ ] 1) tool_choice required (test single and multiple tool call) 
  - [ ] 2) tool_choice: ToolChoice 
    - Regarding tool_choice: specific function, stating it returns only one tool call. And compare it with ebnf
  - [ ] What happens when user gives an invalid json schema in the request?

- [ ] Update the PR description to include:
  - [ ] Why we need this change, 
  - [ ] How json_schema is better than ebnf, 
  - [ ] Put a note related to #8250, how we patched the PR (disallowing white space?)

test/srt/test_function_call_parser.py
```diff
@@ -2190,5 +2191,68 @@ def test_partial_tool_call(self):
         self.assertEqual(self.detector._buffer, "")
 
 
+class TestJsonArrayDetector(unittest.TestCase):
+    def setUp(self):
```
CatherineSue, line 2195: Can we update the ut to include more thorough cases?

1. More complicated json schema: strict: True, minItems, etc.
2. Invalid JsonSchema?

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -34,6 +34,8 @@
     to_openai_style_logprobs,
 )
 from sglang.srt.function_call.function_call_parser import FunctionCallParser
+from sglang.srt.function_call.utils import _get_json_schema_constraint
```
CatherineSue: _ means private function. Private function should not be imported in general

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -1061,7 +1069,7 @@ async def _process_tool_call_stream(
 
     def _check_for_unstreamed_tool_args(
         self,
-        parser: FunctionCallParser,
+        parser: Union[FunctionCallParser, JsonArrayDetector],
```
CatherineSue: Please rename JsonArrayDetector to a parser to avoid confusions here.

python/sglang/srt/entrypoints/openai/serving_chat.py
TJ5: I replied elsewhere, but I modified this function a bit since it was first reviewed. Initially we had this finish_reason block copied three times, which was very repetitive. In the current state we're just repeating it twice, and since it's three lines putting it in a function to avoid repetitiveness doesn't even save any lines.

TJ5:
Moved here: https://github.com/sgl-project/sglang/pull/10550

PR #10041, opened 2025-09-04
Remove non-accelerated targets(100 and up) from cmake

hlu1:
## Motivation

Non-accelerated targets(100 and up) don't need to be included in cmake. Building with sm_100a/103a/110a/120a/121a is sufficient. For forward compatibility **without accelerated features like fp8 or fp4**, we can include the highest sm, which is sm121, but it won't be of much use so I left it out. The sgl-kernel wheel is smaller as a result.

```
before:
-rw-r--r-- 1 root root 339M Sep  4 20:35 ./dist/sgl_kernel-0.3.8-cp310-abi3-linux_x86_64.whl

after:
-rw-r--r-- 1 root root 292M Sep  4 19:26 ./dist/sgl_kernel-0.3.8-cp310-abi3-linux_x86_64.whl
```

## Accuracy Tests

CI

## Checklist


rainj-me (approved):

rainj-me:
LGTM

PR #10056, opened 2025-09-05
Fix cuda graph mode in flashinfer attn backend

benbarsdell:
cc @kushanam

## Motivation

Cuda graph mode causes an error when running Llama models on Blackwell systems.

## Modifications

Changes the scale factors passed to the backend kernel to use the float versions instead of the Parameter versions. This ensures they don't end up triggering device-to-host copies that synchronize the device and break cuda graph capture.

## Accuracy Tests

N/A

## Benchmarking and Profiling

N/A

## Checklist

### DIFF
--- a/python/sglang/srt/layers/attention/flashinfer_backend.py
+++ b/python/sglang/srt/layers/attention/flashinfer_backend.py
@@ -501,8 +501,9 @@ class FlashInferAttnBackend(AttentionBackend):
                 sm_scale=layer.scaling,
                 window_left=layer.sliding_window_size,
                 logits_soft_cap=logits_soft_cap,
-                k_scale=layer.k_scale,
-                v_scale=layer.v_scale,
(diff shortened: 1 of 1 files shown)


python/sglang/srt/layers/attention/flashinfer_backend.py
```diff
@@ -501,8 +501,9 @@ def forward_extend(
                 sm_scale=layer.scaling,
                 window_left=layer.sliding_window_size,
                 logits_soft_cap=logits_soft_cap,
-                k_scale=layer.k_scale,
-                v_scale=layer.v_scale,
+                # Must use _float to avoid device-to-host copy that breaks cuda graph capture.
```
zhyncs, line 504: can you add a unit test

hnyls2002 (approved):

PR #10071, opened 2025-09-05
[feat] support in-flight  weight update

ShawnY112358:
## Motivation

In agent RL training, we need to synchronize model parameters without pausing the SGLang engine.
Existing pause()-based approaches break the inference service: the agent continues to send requests while the engine is paused, resulting in runtime errors.
This PR enables parameter updates without interrupting live inference, keeping the agent unaware of the sync process.

## Modifications
1. pause_generation API with Three Operation Modes
The pause_generation API now supports three distinct modes to provide fine-grained control over the scheduler’s behavior during runtime:
- abort: Immediately aborts all currently processing requests and returns them.
- in_place: Pauses the scheduler’s event loop from performing inference; only non-inference requests (e.g., control commands) are handled. ⚠️ In this mode, flush_cache will fail if there are any requests in the running_batch.
- retract: Pauses inference and retracts all currently running requests back to the waiting_queue, allowing them to be rescheduled later. Only non-inference requests are processed while paused.

2. Non-blocking update_weights
Building on the pause_generation mechanism, the update_weights API can now deliver new model weights directly to the scheduler without acquiring the global model_update_lock.

## Checklist


ShawnY112358:
@ocss884

zhaochenyang20:
> This PR enables parameter updates without interrupting live inference, keeping the agent unaware of the sync process.

It is quite strange. Which version of the parameter do you want to use? I do not think that this need is valid (correct me if I'm wrong).

Also, are you sure that the token can be correctly sampled during the refit? For example, given a 50 - layer model, 1~24 layers are the old version, the others are the new version. Could this give you a correct sample token?

ShawnY112358:
> > This PR enables parameter updates without interrupting live inference, keeping the agent unaware of the sync process.
> 
> It is quite strange. Which version of the parameter do you want to use? I do not think that this need is valid (correct me if I'm wrong).
> 
> Also, are you sure that the token can be correctly sampled during the refit? For example, given a 50 - layer model, 1~24 layers are the old version, the others are the new version. Could this give you a correct sample token?

In RL scenarios, our algorithm supports partial rollout, where different segments of response can be inferred using different checkpoints. We aim to update the inference engine parameters promptly. Using writer_lock would require waiting for ongoing inference requests to complete, creating idle periods. In contrast, online parameter updates can maximize inference resource utilization. 
To support this, we've added an `online` parameter to the update weights request, enabling users to choose whether to flush pending requests from the inference engine or proceed with online parameter updates based on their requirements.

python/sglang/srt/managers/io_struct.py
```diff
@@ -1076,6 +1093,8 @@ class UpdateWeightsFromTensorReqInput(BaseReq):
     abort_all_requests: bool = False
     # Optional: Update weight version along with weights
     weight_version: Optional[str] = None
+    # Whether to update weights online
```
zhaochenyang20: Leave more explanation of online, and I think use words like `non_blocking`?

python/sglang/srt/managers/io_struct.py
```diff
@@ -1051,6 +1064,10 @@ class UpdateWeightsFromDistributedReqInput(BaseReq):
...
+    online: bool = False
+    # Optional format specification for loading
+    flattened_bucket_meta: Optional[dict] = None
```
zhaochenyang20: please give more explanation to this parameter, so as the api of `update_bucketed_weights_from_distributed`.

zhaochenyang20 (changes requested):

python/sglang/srt/managers/io_struct.py
ShawnY112358: This parameter was used to accelerate parameter synchronization with FlattenedTensorBucket. I've removed this code in the new commit. It would be clearer to open a separate PR for this.

python/sglang/srt/managers/io_struct.py
ShawnY112358: done

python/sglang/srt/managers/io_struct.py
```diff
@@ -1039,8 +1039,9 @@ class UpdateWeightFromDiskReqInput(BaseReq):
     recapture_cuda_graph: bool = False
     # The trainer step id. Used to know which step's weights are used for sampling.
     token_step: int = 0
-    # Whether to update weights online
-    online: bool = False
+    # Whether to wait for all ongoing inference requests to complete before updating
+    # parameters, or interrupt them and update parameters directly.
```
zhaochenyang20: or abort and return all the unfinished requests, then update parameters.

python/sglang/srt/managers/io_struct.py
```diff
@@ -1064,10 +1065,9 @@ class UpdateWeightsFromDistributedReqInput(BaseReq):
     abort_all_requests: bool = False
     # Optional: Update weight version along with weights
     weight_version: Optional[str] = None
-    # Whether to update weights online
-    online: bool = False
-    # Optional format specification for loading
-    flattened_bucket_meta: Optional[dict] = None
+    # Whether to wait for all ongoing inference requests to complete before updating
+    # parameters, or interrupt them and update parameters directly.
```
zhaochenyang20: as above, change it

python/sglang/srt/managers/io_struct.py
```diff
@@ -1093,8 +1093,9 @@ class UpdateWeightsFromTensorReqInput(BaseReq):
     abort_all_requests: bool = False
     # Optional: Update weight version along with weights
     weight_version: Optional[str] = None
-    # Whether to update weights online
-    online: bool = False
+    # Whether to wait for all ongoing inference requests to complete before updating
+    # parameters, or interrupt them and update parameters directly.
```
zhaochenyang20: change it as above

test/srt/rl/test_update_weights_from_tensor.py
```diff
@@ -167,6 +177,106 @@ def test_update_weights_from_tensor_load_format_flattened_bucket(self):
...
+        ret = response.json()
+        return ret
+
+    def continue_generation(self):
+        response = requests.post(
+            self.base_url + "/continue_generation",
+            json={},
+        )
+        ret = response.json()
+        return ret
```
zhaochenyang20: Sorry for not catching the development of these APIs for a long time. Just pondering what these two APIs, "pause" and "continue" generation mean in SGLang. What's the difference of `abort_all`? In my understanding, the pause and continue don't return the unfinished requests to the user. It just stops generation, then maybe we update the weights but not flush KV cache, then continue generation?

test/srt/rl/test_update_weights_from_disk.py
```diff
@@ -231,6 +231,99 @@ def test_update_weights_abort_all_requests(self):
...
+            ]
+
+            # ensure the decode has been started
+            time.sleep(2)
+
+            new_model_path = DEFAULT_SMALL_MODEL_NAME_FOR_TEST.replace("-Instruct", "")
+            ret = self.pause_generation()
+            ret = self.run_update_weights(new_model_path, non_blocking=True)
+            self.assertTrue(ret["success"])
+            ret = self.continue_generation()
```
zhaochenyang20: Sorry for my dumb question. If `non_blocking` means while updating the weights, the model is still doing inference, why should we pause it and after the update continue? This makes it no different from the previous.

test/srt/rl/test_update_weights_from_tensor.py
ShawnY112358: In the old API, pause_generation would abort all running requests and return immediately, and no new requests would be processed afterward. #7419 

We now want pause_generation to not abort requests, but instead only pause the decoding process of currently running requests. To achieve this, we introduced the abort_all parameter:

When abort_all=True, pause_generation behaves as before—aborting all running requests.
When abort_all=False, it pauses the inference process in the scheduler's event loop without aborting any requests. The requests remain in either the running_queue or the waiting_queue, preserving their state for potential resumption later.

ShawnY112358: We cannot have the model updating parameters during forward passes. So in this unit test, we first pause the scheduler's event loop using a pause_generation request, execute non-blocking parameter updates, and then resume the scheduler's event loop.

Users have the flexibility to decide whether to flush the KV cache according to their specific requirements, and can resume inference by calling the continue_generation API after the cache has been flushed.（Note that if users want to flush the KV cache, users need to retract all running batches using the retract_all parameter within the pause_generation process to force all requests to re-prefill after resuming inference.otherwise, this will result in an error）

test/srt/rl/test_update_weights_from_disk.py
ShawnY112358: The pause_generation here does not abort all requests—it only pauses the decoding process of running_batch. After executing update_weights and continue_generation, sglang will immediately resume the inference process for the running requests. We must first stop the scheduler from decoding to enable non-blocking parameter updates. Otherwise, when scheduler.enable_overlap is True, there may still be samples performing forward computation while processing parameter updates.

python/sglang/srt/managers/io_struct.py
ShawnY112358: > or abort and return all the unfinished requests, then update parameters.

When non_blocking is set to True, the running request does not return immediately. Instead, it remains in the inference engine while the scheduler's event loop pauses decoding. Inference resumes only after the parameter synchronization is complete.

zhaochenyang20:
Will review it soon.

python/sglang/srt/managers/io_struct.py
JD-ETH: don't we have 3 cases now? 
1. wait for all requests in queue and update 
2. cancel current requests with partial completion and update 
3. keep current requests, update inbetween forward, and update. 

Can we still have all the options available in this struct?

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -1456,7 +1456,7 @@ def check_decode_mem(
         return self._is_available_size_sufficient(num_tokens)
 
     def retract_decode(
-        self, server_args: ServerArgs
+        self, server_args: ServerArgs, retract_all=False
     ) -> Tuple[List[Req], float, List[Req]]:
         """Retract the decoding requests when there is not enough memory."""
```
JD-ETH, line 1527: I feel like you are abusing this function. Just retracting everything in process sounds straight forward, why do you have to reuse this logic?

JD-ETH (commented):
Thanks for implementing this! This is a feature that's we are very excited about. However, I have the following concerns about the interaction design. 

Instead of the pause - update - continue; I would like a single call of update_nonblock: 

`ret = self.run_update_weights(named_tensors, non_blocking=True, flush_cache=False)`

compared to what's in the current PR
```
            ret = self.pause_generation()
            ret = self.run_update_weights(named_tensors, non_blocking=True)
            self.assertTrue(ret["success"])
            ret = self.continue_generation()
```

1. The key of this feature for me is to enable a true non-blocking weight update that is completely invisible to the inference caller. inference caller will not be aware of underlying updates, beyond metadata being returned; and this allows a weight update caller to work completely independently from the service caller. 

From what I see it seems possible to just send the _pause_engine signal directly to the scheduler through weight_update call without even changing the pause_engine API? Obviously I don't know the overlap scheduler well enough, but handling this seems possible. 

Following that "invisible" paradigm, the retract_all logic will need to be implicitly handled given the flush_cache flag, and we have one less failure point for the user. 
- non_blocking + flush_cache: pause generation, flush_cache, weight_update, send all current running requests back to prefill stage, continue 
- non_blocking only: pause generation, weight_update, continue. 
Personally I'm not sure how much people like option 2 --- You are now left with a dirty cache without knowing which version of checkpoint they came from, do we have an actual use case that would prefer this? 


Let me know what you think

python/sglang/srt/managers/io_struct.py
ShawnY112358: 1. Simply set non_blocking=False.
2. Call pause_generation first, then call the update_weights API—this approach is independent of the non_blocking parameter. Of course, if you use the pause_generation API introduced in this PR, you’ll need to set abort_all=True.
3. First call pause_generation with abort_all=False, then call update_weights with non_blocking=True. If you want to implement a single-call non-blocking weight synchronization, you wouldn’t need to call pause_generation at all—just call update_weights directly with non_blocking=True.

python/sglang/srt/managers/schedule_batch.py
ShawnY112358: That makes a lot of sense—I’ll go ahead and update this part.

ShawnY112358:
> sorry, busy with my stuff at school. Ping me if not reply in 24h.

ding～

python/sglang/srt/managers/io_struct.py
zhaochenyang20: In this sense:

1. Set `non_blocking= False`, the weights refit will wait till all the remaining requests finish, then begin the update.
2. First `pause_generation` with `abort_all=True`, then all the requests will be returned. Then call the update weights.
3. Directly call `update_weights` with `non_blocking=True`, then the engine will pause decoding, wait for the parameters to be updated, and then continue decoding with the old KV cache immediately.


Am I understanding these right? If so, could we add these three conditions and usage to the comments to let users know?

BTW, what shall happen if `pause_generation` but `abort_all=False`?

test/srt/rl/test_update_weights_from_tensor.py
zhaochenyang20: > Note that if users want to flush the KV cache, users need to retract all running batches using the retract_all parameter within the pause_generation process to force all requests to re-prefill after resuming inference.otherwise, this will result in an error

Great, could you add this in the comment of the change. This shall be important to users.

JD-ETH:
> I agree with this discussion. Let me set these APIs in this, would you guys agree: @JD-ETH @ShawnY112358
> 
> Let's rename `non_blocking` to `block_generation`.
> 
> 1. Wait for all the running requests to finish, then do the update weights:
> 
> ```python
>  ret = self.run_update_weights(named_tensors, block_generation=False)
> ```
> 
> 2. Stop and return all the generated content, and update the weights.
> 
> ```python
> ret = self.pause_generation(abort_all=True)
> ret = self.run_update_weights(named_tensors, block_generation=False/True)
> ```
> 
> Note that, to continue rollout these unfinished requests, we shall manually send these requests back to the server. Simply calling the `self.continue_generation()` does not affect.
> 
> 3. Stop current decoding, update the weights, and continue decoding after refitting immediately with the old kv cache.
> 
> ```python
> ret = self.run_update_weights(named_tensors, block_generation=True, flush_cache=False)
> ```
> 
> 4. Stop current decoding, update the weights, flush the KV cache with new parameters, and continue decoding:
> 
> ```python
> ret = self.run_update_weights(named_tensors, block_generation=True, flush_cache=True)
> ```
> 
> Would this be clearer?

This is a nice design! I am just unhappy with the "block_generation" naming --- this still somewhat intersects with the `pause_generation` API. 

What about just calling it `force` or `immediate`? Because that's what it is essentially doing --- it's running weight update immediately by jumping the queue on the scheduler.

JD-ETH:
> > Thanks for implementing this! This is a feature that's we are very excited about. However, I have the following concerns about the interaction design.
> > Instead of the pause - update - continue; I would like a single call of update_nonblock:
> > `ret = self.run_update_weights(named_tensors, non_blocking=True, flush_cache=False)`
> > compared to what's in the current PR
> > ```
> >             ret = self.pause_generation()
> >             ret = self.run_update_weights(named_tensors, non_blocking=True)
> >             self.assertTrue(ret["success"])
> >             ret = self.continue_generation()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 1. The key of this feature for me is to enable a true non-blocking weight update that is completely invisible to the inference caller. inference caller will not be aware of underlying updates, beyond metadata being returned; and this allows a weight update caller to work completely independently from the service caller.
> > 
> > From what I see it seems possible to just send the _pause_engine signal directly to the scheduler through weight_update call without even changing the pause_engine API? Obviously I don't know the overlap scheduler well enough, but handling this seems possible.
> > Following that "invisible" paradigm, the retract_all logic will need to be implicitly handled given the flush_cache flag, and we have one less failure point for the user.
> > 
> > * non_blocking + flush_cache: pause generation, flush_cache, weight_update, send all current running requests back to prefill stage, continue
> > * non_blocking only: pause generation, weight_update, continue.
> >   Personally I'm not sure how much people like option 2 --- You are now left with a dirty cache without knowing which version of checkpoint they came from, do we have an actual use case that would prefer this?
> > 
> > Let me know what you think
> 
> Question 2: Regarding "dirty" KV cache
> 
> Some research has shown that in partial rollouts, RL training can still converge properly even without flushing the KV cache. Our recent experiments have also observed similar findings.
> 
> Not flushing the cache avoids the computational overhead of re-running the prefill phase for cached prompts, which can provide a noticeable speed advantage—especially when weight synchronization happens frequently.

thanks for that! Do you still periodically flush it though, or not at all throughout the run and hope the old ones will get evicted naturally?

ShawnY112358:
> I agree with this discussion. Let me set these APIs in this, would you guys agree: @JD-ETH @ShawnY112358
> 
> Let's rename `non_blocking` to `block_generation`.
> 
> 1. Wait for all the running requests to finish, then do the update weights:
> 
> ```python
>  ret = self.run_update_weights(named_tensors, block_generation=False)
> ```
> 
> 2. Stop and return all the generated content, and update the weights.
> 
> ```python
> ret = self.pause_generation(abort_all=True)
> ret = self.run_update_weights(named_tensors, block_generation=False/True)
> ```
> 
> Note that, to continue rollout these unfinished requests, we shall manually send these requests back to the server. Simply calling the `self.continue_generation()` does not affect.
> 
> 3. Stop current decoding, update the weights, and continue decoding after refitting immediately with the old kv cache.
> 
> ```python
> ret = self.run_update_weights(named_tensors, block_generation=True, flush_cache=False)
> ```
> 
> 4. Stop current decoding, update the weights, flush the KV cache with new parameters, and continue decoding:
> 
> ```python
> ret = self.run_update_weights(named_tensors, block_generation=True, flush_cache=True)
> ```
> 
> Would this be clearer?

Oh, I just remembered there's another issue with using the single-call approach...
The update_weights_from_tensor API is often called multiple times in succession by users to complete a single weight synchronization. This could cause SGLang to perform inference during the parameter synchronization process, leading to inconsistent or incorrect results. Do you have any suggestions for a better solution? @JD-ETH

python/sglang/srt/managers/scheduler.py
```diff
@@ -966,6 +971,9 @@ def event_loop_normal(self):
             recv_reqs = self.recv_requests()
             self.process_input_requests(recv_reqs)
 
+            if self._engine_paused:
+                continue
+
             batch = self.get_next_batch_to_run()
```
JD-ETH, line 964: I worry about the logic changes here. 

1. you are skipping after the process_input_requests. Are we preprocessing requests (CPU) continuously in this case? shouldn't we skip the loop entirely? 
2. you are skipping the self_check_during_idle() as well. What does it do? Don't you want to keep it?

JD-ETH (commented):
I need someone with more knowledge about the control loop, lol @zhaochenyang20

JD-ETH:
@ShawnY112358 This is a good point --- often the updates are bucketed and during `force` updates we could see the model forward now happens with partial model weights being updated. 

I can see 2 ways going forward:

1. just ignore this. I actually worry less about the result being nonsensical, since LLMs are pretty robust. However I am concerned about the overhead --- since we are repeatedly hijacking the scheduler queue this could cause significantly induce performance overhead that remove the gains completely with larger models. This probably needs to be profiled. 

2. Fix pause_generation/continue_generation. From what I read now, the current pause_generation behavior is really just "abort_all". is not really "pausing" the current generation in its entire state, but rather just cancelling all pending requests. The logic you implemented looks more like a true pause that I would even prefer as the default solution. It's counterintuitive that a user is supposed to keep the current states of a "pause" themselves. 

========================================================================
Actually, I just saw your initial comment, which aligns with this perfectly, sorry I missed it! 


Thus I would prefer splitting the PR into 2 like you said. 
1. a "force" weight update that immediately jumps the queue in between forward calls, with a cache flush flag.  
2. current `pause_generation` --> `abort_all`, and your proposal as `continue_generation`

@zhaochenyang20  wdyt?

python/sglang/srt/managers/scheduler.py
ShawnY112358: 1. Here, we only want to stop the inference logic—other requests still need to be processed. Skipping all requests outright would also prevent the update_weights request from being handled.
2. From my point of view, self_check_during_idle is designed to check for memory leaks when SGLang's inference request queue is empty. However, in the self._engine_paused state, there may still be partially processed inference requests in the queue, so performing self_check at this time would be inappropriate.

ShawnY112358:
> @ShawnY112358 I wonder why the default retract behavior can't work here.
> 
> the more important distinction for me in A is that the service has stopped and caller attempts will get http request error. Does `run_update_weights` with `force=False` need the waiting queue to be empty also? Why is that? the behavior in 11673 looks the cleanest, what if we have a stop_service where all running requests get moved to the waiting queue together with the partial completion? In that case, we don't even need the user to worry about handling the partial requests?
> 
> In this paradigm, Option A:
> 
> ```
> # retract partial completion to waiting queue
> self.stop_service(retract_all=True)
> ret = self.run_update_weights(named_tensors, force=False/True, flush_cache=True/False)
> # The service will operate on the tasks from the waiting queue automatically
> self.start_service()
> ```
> 
> This looks even better, right?

This is the implementation of writer_lock in aio_rwlock.py. The update_weights operation with force=False needs to acquire the writer_lock, while generation requests need to acquire the reader_lock. Acquiring the writer_lock requires waiting for all currently held reader_locks to be released. Since requests in the waiting_queue also hold the reader_lock, they will block update_weights from acquiring the writer_lock.
```python
    async def acquire_writer(self):
        async with self._lock:
            # Increment the count of writers waiting
            self._waiting_writers += 1
            try:
                # Wait while either a writer is active or readers are present
                while self._writer_active or self._readers > 0:
                    await self._cond.wait()
                self._writer_active = True
            finally:
                # Decrement waiting writers only after we've acquired the writer lock
                self._waiting_writers -= 1
```

JD-ETH:
Summarizing the question: 
For Option A, executing weight update while service is marked down temporarily, we want to retract all pending and running requests into a "waiting queue", and restore them automatically when service is back up. 
Should look like this for the user: 
```
# retract partial completion to waiting queue
self.stop_service(retract_all=True)
ret = self.run_update_weights(named_tensors, force=False/True, flush_cache=True/False)
# The service will operate on the tasks from the waiting queue automatically
self.start_service(restore_retract=True)
```

However we can't use the current waiting queue due to lock contention preventing the weight update. Could we create a new "staging queue" for retraction and restoring purposes?

ShawnY112358:
> Summarizing the question: For Option A, executing weight update while service is marked down temporarily, we want to retract all pending and running requests into a "waiting queue", and restore them automatically when service is back up. Should look like this for the user:
> 
> ```
> # retract partial completion to waiting queue
> self.stop_service(retract_all=True)
> ret = self.run_update_weights(named_tensors, force=False/True, flush_cache=True/False)
> # The service will operate on the tasks from the waiting queue automatically
> self.start_service(restore_retract=True)
> ```
> 
> However we can't use the current waiting queue due to lock contention preventing the weight update. Could we create a new "staging queue" for retraction and restoring purposes?

If we require update_weights to acquire the lock, generation requests would need to be returned at least as far back as the tokenizer_manager, which would make the code extremely complex. I think it's better to inform users that in such cases, they should simply set force=True.

ShawnY112358:
> A couple of small questions about this PR:
> 
> 1. During the pause state, it is possible for new requests to be arriving.  I assume that they are placed in the existing request queue?  Do any changes need to be made to that data structure in order to properly support online request update?
> 2. If a pause request is issued with abort/retract set to True, then the intent is for in-flight requests to be re-issued (starting from prefill) after the update is complete.  I'm curious about how the metadata is traversed to collect the set of in-flight requests that will be re-issued.  Can you share a pointer to that logic?



> A couple of small questions about this PR:
> 
> 1. During the pause state, it is possible for new requests to be arriving.  I assume that they are placed in the existing request queue?  Do any changes need to be made to that data structure in order to properly support online request update?
> 2. If a pause request is issued with abort/retract set to True, then the intent is for in-flight requests to be re-issued (starting from prefill) after the update is complete.  I'm curious about how the metadata is traversed to collect the set of in-flight requests that will be re-issued.  Can you share a pointer to that logic?

As requested by @JD-ETH, the pause state has already been removed in this PR, and a separate PR will be submitted for it later. Let me first explain these questions to you based on the design in the new PR：

1. In the paused state, new generation requests are held in the tokenizer_manager and not dispatched to the scheduler, while other types of requests (e.g., weight updates) are still processed by the scheduler. The scheduler simply refrains from performing forward passes; therefore, we only need to skip the forward pass logic within the scheduler’s event loop.
https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/managers/tokenizer_manager.py#L1191
https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/managers/tokenizer_manager.py#L441

2. When pausing generation with abort=True, the in-flight requests are returned to the user, and SGLang no longer takes responsibility for rescheduling partially generated requests. If retract=True is chosen instead, the requests are placed back into the waiting_queue. Once the pause state ends, the scheduler’s event loop will invoke get_next_batch_to_run → get_new_batch_prefill, which automatically retrieves these partially generated requests from the waiting_queue and resumes them from the prefill stage.
https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/managers/scheduler.py#L1709C26-L1709C47

python/sglang/srt/entrypoints/http_server.py
```diff
@@ -1086,19 +1068,19 @@ async def separate_reasoning_request(obj: SeparateReasoningReqInput, request: Re
...
         status_code=200,
     )
 
 
 @app.post("/continue_generation")
-async def continue_generation(request: Request):
+async def continue_generation(obj: ContinueGenerationReqInput, request: Request):
     """Continue generation."""
-    await _global_state.tokenizer_manager.continue_generation()
+    await _global_state.tokenizer_manager.continue_generation(obj)
```
JD-ETH, line 1104: do you still need this? ContinueGenerationReqInput

python/sglang/srt/managers/io_struct.py
```diff
@@ -1064,6 +1008,17 @@ class FlushCacheReqOutput(BaseReq):
     success: bool
 
 
+@dataclass
+class PauseGenerationReqInput(BaseReq):
+    abort_all: bool = True
```
JD-ETH: looking at your implementation, abort and retract are mutually exclusive. Can you do Literal string here? 
`mode: Literals['abort', 'retract']`

And add some explaination what they do

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -1187,21 +1143,18 @@ def abort_request(self, rid: str = "", abort_all: bool = False):
...
-                is_locked = await self.model_update_lock.is_locked()
-                if not is_locked:
-                    break
-                await asyncio.sleep(1.0)
 
-    async def continue_generation(self):
+    async def continue_generation(self, obj: ContinueGenerationReqInput):
         async with self.is_pause_cond:
             self.is_pause = False
+            await self.send_to_scheduler.send_pyobj(obj)
```
JD-ETH, line 1269: why do you need this? Isnt' it empty?

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -1216,22 +1169,17 @@ async def update_weights_from_disk(
             obj.load_format = self.server_args.load_format
         logger.info("Start update_weights. Load format=%s", obj.load_format)
 
+        if obj.force:
```
JD-ETH: I see you added a new parameter to circumvent the lock. I don't think additional parameter should be necessary here.

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -1216,22 +1169,17 @@ async def update_weights_from_disk(
...
+            return await self._wait_for_model_update_from_disk(obj)
+
         if obj.abort_all_requests:
             self.abort_request(abort_all=True)
 
         if True:  # Keep this redundant check to simplify some internal code sync
```
JD-ETH, line 1292: Instead you can skip this lock check if you identify the engine is in a paused state. like `if not self.is_pause`?

JD-ETH (changes requested):
Thanks for the update! Some final comments and I think it looks good to me. Let's remove `force` all together. 

It should look like this for non_blocking: 
```
engine.pause_generation(mode='retract')
engine.update_weights(model_params)
engine.continue_generation()
```
and this for queue clear 
```
engine.pause_generation(mode='abort')
engine.update_weights(model_params)
engine.continue_generation()
```

python/sglang/srt/entrypoints/http_server.py
ShawnY112358: The _request_dispatcher in the scheduler needs to invoke the appropriate handler function based on the type of the incoming request object.

python/sglang/srt/managers/io_struct.py
ShawnY112358: Perhaps in certain scenarios we need abort_all=False and retract=False to avoid re-running prefill. Alternatively, we could use the default value `pause_only` to represent the case where abort_all=False and retract_all=False？

python/sglang/srt/managers/tokenizer_manager.py
ShawnY112358: Oh, right—I'll update it.

python/sglang/srt/managers/tokenizer_manager.py
ShawnY112358: For the same reason we keep the ContinueGenerationReqInput data structure — because requests sent to the scheduler need to be routed by _request_dispatcher to the appropriate handler function based on their type.

python/sglang/srt/managers/io_struct.py
JD-ETH: I see. I'd say those are 3 distinct cases. So Let's do Literals['abort', 'retract', 'in_place'] instead? We should give user the full visibility. In terms of the default value, I was thinking retract? 

Could you add test case for in_place as well? From the test you had so far they look like only supporting 'retract' mode.

JD-ETH: in this case, after "in_place" pause, and if we update weight with "flush_cache", the continue generation will error out, right?

JD-ETH: I'm fine if we don't handle it right now, but if you could assert an error in the test case it would be very insightful I think.

ShawnY112358: https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/managers/scheduler.py#L2260
Since flush_cache already includes logic to check for running requests (and issues a warning instead of flushing if any are present), it would be cleaner and more consistent to reuse that existing logic rather than duplicating the check elsewhere.

ShawnY112358: Regarding the default value, I set `abort` as the default so that users who are accustomed to the old `pause_generation` interface won't need to modify their calls.

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -1216,22 +1169,18 @@ async def update_weights_from_disk(
             obj.load_format = self.server_args.load_format
         logger.info("Start update_weights. Load format=%s", obj.load_format)
 
+        async with self.is_pause_cond:
+            if self.is_pause:
+                return await self._wait_for_model_update_from_disk(obj)
+
```
JD-ETH, line 1291: can you move it to after the abort_all_requests? Though unlikely, we still want the abort_all_requests to take effect if set to True during weight update call.

JD-ETH (approved):
This is clean! just one small nit. Thanks for the nice work. @zhaochenyang20 the test case is pretty clear now.

python/sglang/srt/managers/tokenizer_manager.py
ShawnY112358: Done.

JD-ETH (approved):
waiting on someone with access. LGTM

python/sglang/srt/managers/scheduler.py
```diff
@@ -2510,6 +2520,29 @@ def abort_request(self, recv_req: AbortReq):
...
 
+    def pause_generation(self, recv_req: PauseGenerationReqInput):
+        self._engine_paused = True
+
+        if self.enable_overlap and self.last_batch:
+            # Process the results of the last batch
+            tmp_batch, tmp_result = self.result_queue.popleft()
+            self.process_batch_result(tmp_batch, tmp_result)
+            self.last_batch = None
+            self.cur_batch = None  # TODO: to confirm
```
JD-ETH: can you confirm this works? Let's not add TODOs

python/sglang/srt/managers/tokenizer_communicator_mixin.py
```diff
@@ -402,6 +402,13 @@ async def update_weights_from_distributed(
         if obj.abort_all_requests:
             self.abort_request(abort_all=True)
 
+        async with self.is_pause_cond:
```
JD-ETH, line 408: Let's add a comment what this does --- immediately update the weights if the engine is in paused state

JD-ETH (changes requested):

python/sglang/srt/managers/scheduler.py
ShawnY112358: Forgot to remove

python/sglang/srt/managers/io_struct.py
```diff
@@ -1064,6 +1064,36 @@ class FlushCacheReqOutput(BaseReq):
     success: bool
 
 
+@dataclass
+class PauseGenerationReqInput(BaseReq):
+    """
+    abort: Abort all requests currently being processed.
```
zhaochenyang20: Abort all requests currently being processed.

-> 

Abort and return all requests currently being processed.

python/sglang/srt/managers/io_struct.py
```diff
@@ -1064,6 +1064,36 @@ class FlushCacheReqOutput(BaseReq):
...
+
+    in_place: Pause the scheduler's event_loop from performing inference;
+            only non-inference requests (e.g., control commands) will be handled.
+            Note: In 'inplace' mode, flush_cache will fail if there are any requests
+            in the running_batch.
+
+    retract: Pause the scheduler's event loop from performing inference;
+            only non-inference requests will be handled, and all currently running
+            requests will be retracted back to the waiting_queue.
```
zhaochenyang20, line 1082: What will happen in `in_place`? The requests in the engine will be paused and stay in the `event_loop`, then continue generation after `ContinueGenerationReqInput` with the old kv cache?

Also, for `retract`, the requests in the engine will be paused and removed from the `event_loop` into `waiting_queue`, and the KV cache will be flushed and recomputed after `ContinueGenerationReqInput`?

If my understanding is correct, please make a more adequate description of this PR.

At the same time, it seems that `in_place` and `retract` are counterparts; should we only keep one parameter like `retract=True`?

zhaochenyang20 (changes requested):

python/sglang/srt/managers/io_struct.py
zhaochenyang20: Sorry for my late reply, but I'm just not sure we should put the `abort` here.

If we have an abort parameter here, the `abort_request` and `pause_generation` seem to be intersecting. This makes the API ambiguous.

Could we only leave the `pause` to do the pause, and let users choose whether to recompute the KV cache?

zhaochenyang20 (changes requested):

python/sglang/srt/managers/io_struct.py
zhaochenyang20: I see that the current `pause` API can do the `abort_all`, which feels a bit strange to me. Let me discuss with the team.

test/srt/rl/test_update_weights_from_disk.py
```diff
@@ -231,6 +231,105 @@ def test_update_weights_abort_all_requests(self):
...
+                # ensure the decode has been started
+                time.sleep(2)
+
+                new_model_path = DEFAULT_SMALL_MODEL_NAME_FOR_TEST.replace(
+                    "-Instruct", ""
+                )
+                ret = self.pause_generation(mode)
+                ret = self.run_update_weights(
+                    new_model_path, flush_cache=mode == "retract"
+                )
```
zhaochenyang20: Indeed, we cannot validate whether the KV cache is indeed recomputed. We can leave a comment here, stating that “we can not validate whether the KV cache is indeed recomputed, we just trust the `retract` parameter.”

test/srt/rl/test_update_weights_from_distributed.py
```diff
@@ -627,5 +672,78 @@ def test_update_weights_from_distributed(self):
...
+            "model.layers.5.mlp.gate_proj.weight",
+            "model.layers.6.mlp.up_proj.weight",
+            "model.layers.7.mlp.down_proj.weight",
+            "model.layers.8.post_attention_layernorm.weight",
+            "model.norm.weight",
+            "lm_head.weight",
+        ]
+
+        for tp_size, dp_size, model_name, backend in test_suits:
+            modes = ["in_place", "retract"]
```
zhaochenyang20: `modes = ["in_place", "retract"]` -> `pause_generation_modes = ["in_place", "retract"]`.

test/srt/rl/test_update_weights_from_distributed.py
```diff
@@ -627,5 +672,78 @@ def test_update_weights_from_distributed(self):
...
+            for mode in modes:
```
zhaochenyang20: `for pause_generation_mode in pause_generation_modes:`

test/srt/rl/test_update_weights_from_distributed.py
```diff
@@ -627,5 +672,78 @@ def test_update_weights_from_distributed(self):
...
+                test_update_weights_from_distributed(
+                    tp_size,
+                    dp_size,
+                    model_name,
+                    backend,
+                    model_state_dict_shapes[model_name],
+                    truncate_size,
+                    checking_parameters,
+                    mode=mode,
```
zhaochenyang20: `pause_generation_mode`, please do not use `pause_generation_mode=pause_generation_mode`.

zhaochenyang20 (changes requested):

test/srt/rl/test_update_weights_from_disk.py
```diff
@@ -231,6 +231,105 @@ def test_update_weights_abort_all_requests(self):
...
+            self.base_url + "/generate",
+            json={
+                "text": "The capital of France is",
+                "sampling_params": {
+                    "temperature": 0,
+                    "max_new_tokens": max_new_tokens,
+                    "ignore_eos": True,
+                },
+            },
+        )
```
zhaochenyang20: In the same time, as we want to test the KV cache pool, we do not always send the same request to the server, otherwise it will always use the same KV cache. Send a request like `"text": "Question: {random.int},The capital of France is",`

test/srt/rl/test_update_weights_from_tensor.py
```diff
@@ -167,6 +177,112 @@ def test_update_weights_from_tensor_load_format_flattened_bucket(self):
...
+
+    @classmethod
+    def tearDownClass(cls):
+        kill_process_tree(cls.process.pid)
+
+    def run_decode(self, max_new_tokens=32):
+        response = requests.post(
+            self.base_url + "/generate",
+            json={
+                "text": "The capital of France is",
```
zhaochenyang20: change the `text` also.

test/srt/rl/test_update_weights_from_distributed.py
```diff
@@ -282,8 +292,25 @@ def init_process_sgl(
...
 
-    torch.cuda.synchronize()
-    time_begin_update = time.perf_counter()
+    if mode in ["inplace", "retract"]:
+
+        def run_decode(max_new_tokens=32):
+            response = requests.post(
+                url + "/generate",
+                json={
+                    "text": "The capital of France is",
```
zhaochenyang20: change the `text` also.

test/srt/rl/test_update_weights_from_tensor.py
```diff
@@ -167,6 +177,112 @@ def test_update_weights_from_tensor_load_format_flattened_bucket(self):
...
+                ],
+                "flush_cache": flush_cache,
+            },
+        )
+        ret = response.json()
+        return ret
+
+    def test_update_weights(self):
+        modes = ["in_place", "retract"]
+        for mode in modes:
```
zhaochenyang20: also, please do not use `mode` here.

test/srt/rl/test_update_weights_from_disk.py
```diff
@@ -231,6 +231,105 @@ def test_update_weights_abort_all_requests(self):
...
+            json={},
+        )
+        ret = response.json()
+        return ret
+
+    def test_update_weights(self):
+        origin_model_path = self.get_model_info()
+        print(f"[Server Mode] origin_model_path: {origin_model_path}")
+
+        modes = ["in_place", "retract"]
```
zhaochenyang20: also, please do not use `mode` here.

zhaochenyang20 (changes requested):

test/srt/rl/test_update_weights_from_distributed.py
zhaochenyang20: In the same time, please move the `pause_generation_mode` into the `test_suits`, make it like:

`for tp_size, dp_size, model_name, backend, pause_generation_mode in test_suits:`

The same for other update weights unit tests.

And, please help me change the codes as follows: (someone seems to be abusing my previous setting for random covering of the tests, basically, local tests should cover more)

```python
        if is_in_ci():
            backend = random.choose("server", "engine")
            pause_generation_mode = random.choose("in_place", "retract")
            test_suits = [
                (1, 1, DEFAULT_SMALL_MODEL_NAME_FOR_TEST, backend),
            ]
        else:

            test_suits = [
                (1, 1, DEFAULT_MODEL_NAME_FOR_TEST, "Sever", random.choose("in_place", "retract")), (1, 1, DEFAULT_MODEL_NAME_FOR_TEST, "Engine", random.choose("in_place", "retract"))
            ]

            if torch.cuda.device_count() >= 4:
                test_suits.append(
                    (1, 2, DEFAULT_MODEL_NAME_FOR_TEST, "Engine", random.choose("in_place", "retract")), (2, 1, DEFAULT_MODEL_NAME_FOR_TEST, "Sever", random.choose("in_place", "retract")),
                )

            if torch.cuda.device_count() >= 5:
                test_suits.append(
                    (2, 2, DEFAULT_MODEL_NAME_FOR_TEST, random.choose("Engine", "Sever"), random.choose("in_place", "retract")),
                )
```

Do it the same as other tests, adding new parameters in the `test_suites`.

zhaochenyang20 (changes requested):

python/sglang/srt/managers/io_struct.py
ShawnY112358: Discussed this with @JD-ETH . Currently, pause only supports these three modes. If we use the abort and retract parameters for control, the retract parameter becomes meaningless when abort=True.

test/srt/rl/test_update_weights_from_disk.py
ShawnY112358: I didn't quite understand your point. Here, we just want to clarify that if the user wants to flush the cache, they must call pause_generation in retract mode.

test/srt/rl/test_update_weights_from_distributed.py
ShawnY112358: The pause_generation API does not expose an interface for Engine mode

python/sglang/srt/managers/io_struct.py
zhaochenyang20: Please add these details in the comments:

> What will happen in in_place? The requests in the engine will be paused and stay in the event_loop, then continue generation after ContinueGenerationReqInput with the old kv cache?
> Also, for retract, the requests in the engine will be paused and removed from the event_loop into waiting_queue, and the KV cache will be flushed and recomputed after ContinueGenerationReqInput?
> If my understanding is correct, please make a more adequate description of this PR.

zhaochenyang20 (changes requested):

test/srt/rl/test_update_weights_from_distributed.py
```diff
@@ -627,5 +673,100 @@ def test_update_weights_from_distributed(self):
...
+        assert torch.cuda.device_count() >= 2, "At least 2 GPUs are required"
+        # test_suits : tp, dp, model_name, backend
+        if is_in_ci():
+            pause_generation_mode = random.choice(["in_place", "retract"])
+            test_suits = [
+                (
+                    1,
+                    1,
+                    DEFAULT_SMALL_MODEL_NAME_FOR_TEST,
+                    "Server",
```
zhaochenyang20: `if_in_ci`, also choose from `["Server", "Engine"]`

zhaochenyang20 (approved):
Great job

python/sglang/srt/managers/tokenizer_communicator_mixin.py
```diff
@@ -404,6 +404,14 @@ async def update_weights_from_distributed(
         if obj.abort_all_requests:
             self.abort_request(abort_all=True)
 
+        # Immediately update the weights if the engine is in paused state
+        async with self.is_pause_cond:
+            if self.is_pause:
+                result = (await self.update_weights_from_distributed_communicator(obj))[
+                    0
+                ]
+                return result.success, result.message
```
merrymercy, line 413: There are some other steps.

Please also do `success, message = _Communicator.merge_results(results)` and 

```
        if success and obj.weight_version is not None:
            self._update_weight_version_if_provided(obj.weight_version)
            message += f" Weight version updated to {obj.weight_version}."
```

python/sglang/srt/managers/tokenizer_communicator_mixin.py
```diff
@@ -457,6 +465,12 @@ async def update_weights_from_tensor(
         if obj.abort_all_requests:
             self.abort_request(abort_all=True)
 
+        # Immediately update the weights if the engine is in paused state
+        async with self.is_pause_cond:
+            if self.is_pause:
+                result = (await self.update_weights_from_tensor_communicator(obj))[0]
+                return result.success, result.message
```
merrymercy, line 472: Same, please do the post-processing correctly instead of directly returning the `result` from rank 0.

Code skeleton
```
if self.is_pause:
    result = await self.update_weights_from_tensor_communicator(obj)
else:
    async with self.model_update_lock.writer_lock:
    result = ...

# post processing
```

PR #9911, opened 2025-09-02
[NVIDIA] Build CUDA 13 sgl-kernel

johnnynunez:
This allow to build CUDA 13 sgl-kernel in CI

cc: @zhyncs

PR #10075, opened 2025-09-05
fix memory Jetson Orin/Thor/Spark

johnnynunez:
For devices where the GPU uses system RAM as its memory, the torch.cuda.mem_get_info() function only returns the amount of memory that is unused at the moment. This can underestimate the real available memory, since it ignores the portion currently stored in cache. To get a more accurate picture, we check the system’s reported available memory instead.

cc @zhyncs


Ying1123:
included in #9911

Issue #10077, opened 2025-09-05
[Bug] unify get_ip methods and netifaces missing


jinmingyi1998:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Describe the bug

ImportError: Environment variable SGLANG_LOCAL_IP_NIC requires package netifaces, please install it through 'pip install netifaces'

### Reproduction

PD with mooncake, set Environment variable SGLANG_LOCAL_IP_NIC

### Environment

docker sglang:v0.5.0rc0-cu126

jimmy-evo:
related PR:
#8356  why remove `pip install netifaces`
https://github.com/sgl-project/sglang/pull/7237#discussion_r2149914159
@fzyzcjy 
I prefer to install netifaces in docker image (not write in project.toml), or i have to maintain another branch to install this.

also, https://github.com/SamuelYvon/netifaces-2 is an alternative


@ShangmingCai  there are too many duplicate and similar `get_ip_xxx`
`get_local_ip_auto` can get ip with `SGLANG_HOST_IP` in `get_ip()`

and `get_ip` do the same thing as `def get_local_ip_by_remote() -> str:`

jimmy-evo:
I would like to submit a PR.
I'd also like to hear your opinions.

PR #10081, opened 2025-09-05
[Feature]feat(get_ip): unify get_ip_xxx

jimmy-evo:
whenever want to get local machine's ip, use `get_local_ip_auto`
#10077

## Motivation

get_ip_xxx method is not unified and similar.

## Modifications

use `get_local_ip_auto` instead
before modify, there are 4 cases:
- get_ip
> get_ip just read env `SGLANG_HOST_IP` (most people don't set), then do the same thing as `get_local_ip_by_remote()`
- get_local_ip_by_remote **NOT CHANGED**
- get_local_ip_auto
> actually is `get_local_ip_by_nic`
- get_local_ip_by_nic **NOT CHANGED**
after:
only one entry `get_local_ip_auto`
in `get_local_ip_auto`, try three ways in order:
- get_ip
- get_local_ip_by_remote 
- get_local_ip_by_nic

also can set a fallback value such as "0.0.0.0" if use this as server's host

## Checklist


python/sglang/srt/utils.py
```diff
@@ -1938,48 +1938,11 @@ def set_uvicorn_logging_configs():
...
-        pass
-
-    warnings.warn(
-        "Failed to get the IP address, using 0.0.0.0 by default."
-        "The value can be set by the environment variable"
-        " SGLANG_HOST_IP or HOST_IP.",
-        stacklevel=2,
-    )
-    return "0.0.0.0"
+    return None
```
ShangmingCai, line 2012: Not sure whether this change is safe, why was the following code deleted?

jimmy-evo: > Not sure whether this change is safe, why was the following code deleted?

these code is duplicate to get_local_ip_remote

jimmy-evo: > Not sure whether this change is safe, why was the following code deleted?

@ShangmingCai  This `get_ip` method does "read SGLANG_HOST_IP" and `get_local_ip_by_remote`, which is duplicated.

`get_ip` should just read SGLANG_HOST_IP
and There is already a "get_local_ip_by_remote" here.

python/sglang/srt/distributed/device_communicators/shm_broadcast.py
```diff
@@ -191,7 +191,9 @@ def __init__(
         self.n_remote_reader = n_remote_reader
 
         if connect_ip is None:
-            connect_ip = get_ip() if n_remote_reader > 0 else "127.0.0.1"
+            connect_ip = (
+                get_local_ip_auto("0.0.0.0") if n_remote_reader > 0 else "127.0.0.1"
+            )
```
ShangmingCai, line 196: @yizhang2077 Can you check whether this is correct?

python/sglang/srt/utils.py
```diff
@@ -2246,16 +2209,9 @@ def bind_or_assign(target, source):
...
-        get_local_ip_by_nic(interface)
-        if interface is not None
-        else get_local_ip_by_remote()
-    )
-
-
-def get_local_ip_by_nic(interface: str) -> str:
+def get_local_ip_by_nic(interface: str = None) -> Optional[str]:
+    if (interface or os.environ.get("SGLANG_LOCAL_IP_NIC", None)) is None:
+        return None
```
ShangmingCai: Does this change correct? Even if `(interface or os.environ.get("SGLANG_LOCAL_IP_NIC", None)) is None` is False, but interface is not assigned a value? Also, this `or` logic looks really weird. I think there does not exist a case where SGLANG_LOCAL_IP_NIC is not set while `interface` is not None and `get_local_ip_by_nic` is called.

jimmy-evo: > Does this change correct? Even if `(interface or os.environ.get("SGLANG_LOCAL_IP_NIC", None)) is None` is False, but interface is not assigned a value? Also, this `or` logic looks really weird. I think there does not exist a case where SGLANG_LOCAL_IP_NIC is not set while `interface` is not None and `get_local_ip_by_nic` is called.

```
>>> interface,os.environ.get("SGLANG_LOCAL_IP_NIC", None)
(None, None)
>>> interface or os.environ.get("SGLANG_LOCAL_IP_NIC", None)
>>> interface or os.environ.get("SGLANG_LOCAL_IP_NIC", None) is None
True
>>> interface = ""
>>> interface or os.environ.get("SGLANG_LOCAL_IP_NIC", None) is None
True
```

jimmy-evo:
@ShangmingCai 
you're correct, `interface` is not assigned, fixed:
```
>>> interface = 'interface'
>>> import os
>>> if not (interface := interface or os.environ.get("SGLANG_LOCAL_IP_NIC", None)):
...     print('none!')
...     
>>> print(interface)
interface
>>> 
>>> 
>>> 
>>> interface = ''
>>> if not (interface := interface or os.environ.get("SGLANG_LOCAL_IP_NIC", None)):
...     print('none!')
...     
none!
>>> 
>>> 
>>> 
>>> os.environ['SGLANG_LOCAL_IP_NIC'] = 'env'
>>> 
>>> if not (interface := interface or os.environ.get("SGLANG_LOCAL_IP_NIC", None)):
...     print('none!')
...     
>>> interface
'env'
>>> 
```

python/sglang/srt/utils.py
```diff
@@ -2309,7 +2263,48 @@ def get_local_ip_by_remote() -> str:
...
+        str: The detected local IP address, or the fallback value if detection fails.
+
+    Raises:
+        ValueError: If IP detection fails and no fallback value is provided.
+
+    Note:
+        The function tries detection methods in the following order:
+        1. Direct IP detection via get_ip()
+        2. Remote connection method via get_local_ip_by_remote()
+        3. Network interface enumeration via get_local_ip_by_nic()
```
ShangmingCai: It seems this order is not aligned with the implementation.

jimmy-evo: > It seems this order is not aligned with the implementation.

@ShangmingCai fix

python/sglang/srt/distributed/device_communicators/shm_broadcast.py
yizhang2077: LGTM

hnyls2002 (approved):
LGTM. CI triggered and auto-merge is enabled.

ShangmingCai:
No need to merge main, I can rerun the failed CI.

PR #10093, opened 2025-09-06
[doc] add walkthrough for implementing and hosting a simple llama wrapper m…

glenliu21:
…odel

## Motivation

For newcomers to the SRT library, the steps to create a local model and then host it via the Offline Engine aren't clear. In particular, the docs located at `docs/supported_models/support_new_models.md` are quite sparse.

## Modifications

I added a section in the docs that explains how to create a simple Llama Wrapper model using SRT and the steps that are required to download the model from Huggingface and then host it via the Offline Engine.

### DIFF
--- a/docs/supported_models/support_new_models.md
+++ b/docs/supported_models/support_new_models.md
@@ -135,6 +135,182 @@ ModelRegistry.models.update(import_new_model_classes())
 launch_server(server_args)
 ```
 
+## Example: Implementing and Serving a Llama Wrapper Model
+
+Below is an introductory, step-by-step walkthrough on how to implement a new model end-to-end in SGLang and then run it via the [Offline Engine](https://github.com/sgl-project/sglang/blob/main/docs/basic_usage/offline_engine_api.ipynb).
+
+### Implementing Our Model
+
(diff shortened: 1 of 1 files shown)


docs/supported_models/support_new_models.md
```diff
@@ -135,6 +135,182 @@ ModelRegistry.models.update(import_new_model_classes())
...
+from typing import Optional
+from sglang.srt.layers.logits_processor import LogitsProcessorOutput
+from sglang.srt.layers.quantization.base_config import QuantizationConfig
+from sglang.srt.model_executor.forward_batch_info import ForwardBatch, PPProxyTensors
+
+from sglang.srt.models.llama import LlamaForCausalLM
+```
+
+Next, we declare a new `class` for our model and have it inherit from `LlamaForCausalLM`, which allows our model to access `LlamaForCausalLM`'s predefined modules and layers, such as `LlamaAttention` and `LlamaMLP`.
+Note that almost all model implementations take in `config` and `quant_config` as arguments for their `__init__` method; `config` and `quant_config` are passed in via [`model_loader/loader.py`](https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/model_loader/loader.py#L205).
```
Kangyan-Zhou: Can we use a permanent link with a specific commit instead of main since the code changes very often?

hnyls2002 (approved):

PR #10101, opened 2025-09-06
Optimize nvfp4 block scaled gemm kernel when M is small.

HydraQYH:
## Motivation
Until now, the CUTLASS-based nvfp4 block-scaled GEMM kernel used a uniform performance config for all MNKs. This config performed well when MNKs were large, but it resulted in redundant computation when M was small. Therefore, we have optimized this for scenarios with small M. In this PR, we would like to highlight several sgl-kernel specific optimizations:

- Fine-tuned ClusterShape: ClusterShape affects the direction of TMA Multicast.
  - If ClusterShapeM > 1, the ThreadBlock Cluster will use TMA Multicast to load B. The ThreadBlocks in the Cluster will share B but load different A. ClusterMmaTileShape is MmaTileShape expanded in the M direction to ClusterShapeM times.
  - If ClusterShapeN > 1, the ThreadBlock Cluster will use TMA Multicast to load A. The ThreadBlocks in the Cluster will share A but load different B. ClusterMmaTileShape is MmaTileShape expanded in the N direction to ClusterShapeN times.
  - For 1SM strategy, ClusterMmaTileShape is (MmaTileShapeM * ClusterShapeM, MmaTileShapeN * ClusterShapeN, MmaTileShapeK)
  - For 2SM strategy, ClusterMmaTileShape is (MmaTileShapeM * (ClusterShapeM/2), MmaTileShapeN * ClusterShapeN, MmaTileShapeK)
  - When M is small, we no longer set ClusterShapeM greater than 1 to avoid redundant calculations, but still use ClusterShapeN greater than 1 to achieve higher TMA multicast.
- Blindly using `cutlass::epilogue::collective::EpilogueTileAuto` may lead to register spilling, so we allocate exactly one (128, 64) output tile per Epilogue Warp.
- We use Dynamic Clusters to ensure high SM occupancy.

Reference:

- https://github.com/NVIDIA/cutlass/blob/main/media/docs/cpp/blackwell_functionality.md
- https://github.com/NVIDIA/cutlass/blob/main/examples/73_blackwell_gemm_preferred_cluster/blackwell_gemm_preferred_cluster.cu

## Modifications
sgl-kernel/csrc/gemm/nvfp4_scaled_mm_kernels.cu

## Accuracy Tests
image

## Benchmarking and Profiling
Before:
image
After:
image

## Checklist


HydraQYH:
@zhyncs The compilation bug of CUDA 12.4 has been fixed. Ready to merge.

PR #10104, opened 2025-09-06
Fix run time error in dsv3-fp8 model on mi35x

kkHuang-amd:
## Motivation
Got run-time error in running dsv3-fp8 model on mi35x.
The error is caused by the activation tensor is changed to tuple of tensors type for the next mxfp4 linear computation. But the fp8 does not need to do this change

## Checklist


HaiShaw (approved):
LGTM

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -249,7 +249,11 @@ def forward(
         if (self.tp_size == 1) and x.shape[0] == 0:
             return x
 
-        if gemm_output_zero_allocator != None and x.shape[0] <= 256:
+        if (
+            gemm_output_zero_allocator is not None
+            and x.shape[0] <= 256
+            and self.gate_up_proj.weight.dtype == torch.uint8
```
zhyncs, line 255: Is it necessary to add this condition on nv gpu

kkHuang-amd: un-neccessary on nv gpu

HaiShaw: L252-L260 were introduced only for HIP/gfx95x.
@kkHuang-amd would you please either add a comment for this block, or add an explicit condition?

PR #7448, opened 2025-06-22
Fix for fp8 quantization failure of qwen 2.5 VL 7B model.

PanJason:
## Motivation

Fix this bug https://github.com/sgl-project/sglang/issues/6828 of qwen 2.5 VL 7B where fp8 quantization was reported to fail in the visual model part.

## Modifications

As written in the commit message, two modifications are performed.
1. Make `per_token_quant_fp8.cu` work with non-16-aligned hidden dim. I split the non-aligned hidden dim into 3 parts, the prologue part, the aligned part, and the tail part, since the address may start from a non-128-bit aligned address.
2. Make `fp8_gemm_kernel` work with non-16-aligned K and N supposing the mat mul is `[M, K] x [K, N]`. The extension of N is unlikely to trigger perf regression as we can just initialize the output matrix to be larger but aligned and return a view. The unaligned K is the problematic part (hidden dim 3420 in case of qwen2.5 VL 7B visual LLM). In this PR, the solution is to initialize two empty matrices whose K is aligned and perform two additional copies from the original `mat_a` and `mat_b` respectively. I don't know where this is appropriate, as it slows down the inference and requires double the memory. Two other options are available as I stated in the commit message:
a. Initialize the weight matrix directly to have a 16-aligned (3424 in this case) hidden dim with zero padding.
b. Disable quantization for the vision model part of qwen 2.5 VL. 

Feedback is welcome. Thanks!

## Checklist

- [x] Format your code according to the [Code Formatting with Pre-Commit](https://docs.sglang.ai/references/contribution_guide.html#code-formatting-with-pre-commit).
- [x] Add unit tests as outlined in the [Running Unit Tests](https://docs.sglang.ai/references/contribution_guide.html#running-unit-tests-adding-to-ci).
- [ ] Update documentation / docstrings / example tutorials as needed, according to [Writing Documentation](https://docs.sglang.ai/references/contribution_guide.html#writing-documentation-running-docs-ci).
- [ ] Provide throughput / latency benchmark results and accuracy evaluation results as needed, according to [Benchmark and Profiling](https://docs.sglang.ai/references/benchmark_and_profiling.html) and [Accuracy Results](https://docs.sglang.ai/references/accuracy_evaluation.html).
- [ ] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co-author when merging the PR.
- [ ] Please feel free to join our Slack channel at https://slack.sglang.ai to discuss your PR.

PR #10112, opened 2025-09-06
fix: fp8 quantization failure of qwen 2.5 VL 7B model

PanJason:
## Motivation

Fix this bug https://github.com/sgl-project/sglang/issues/6828 of qwen 2.5 VL 7B where fp8 quantization was reported to fail in the visual model part.

## Modifications

1. Make the initialization of the intermediate size in qwen 2.5 VL 7B 8-aligned
2. Change both `weight_loader` and `weight_loader_v2` in `linear.py` to add zero padding when loading weights that are not 8-aligned. After this fix, both `w8a8_fp8` and `fp8` can work
3. Change `w8a8_int8.py` to fix the runtime crash when using `w8a8_int8`

We don't change the kernel as the previous version did #7448

## Remaining work
While fixing the quantization failure for qwen2_5_VL model, I found two other issues. Since they are not exactly related to VL model, I created two other issues. The fix of them will go to different patch series
1. When sending two concurrent requests to qwen2_5_VL model, rotary_embedding crashed regardless of quantization, this is reported in #10623 
2. When fixing w8a8_int8, I found qwen2.5 7B instruct base model generates gibberish. This is reported in #10626 

## Checklist


python/sglang/srt/layers/linear.py
```diff
@@ -626,9 +626,38 @@ def weight_loader(
...
+                            )
+                        else:
+                            # All padding
+                            pad_shape = list(loaded_weight.shape)
+                            pad_shape[output_dim] = shard_size
+                            loaded_weight = torch.zeros(
+                                pad_shape,
+                                dtype=loaded_weight.dtype,
+                                device=loaded_weight.device,
+                            )
```
JustinTong0323: [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). Refactor it into a helper function.

JustinTong0323 (changes requested):

PanJason:
Refactor as requested. Now fix fp8 and w8a8_int8

PanJason:
@JustinTong0323 @mickqian Could you please check this PR?

b8zhong:
Can you check if the MMMU accuracy is recovered?

Swipe4057:
Hi, I tested it on Qwen2.5-7B-VL in w8a8_fp8 quantization and it works without errors, checked on local corporate benchmark - the accuracy dropped by about 3 %.

PanJason:
> Can you check if the MMMU accuracy is recovered?

Can you point me to where the MMMU accuracy tests are? Are they local or run in CI?

PanJason:
> Hi, I tested it on Qwen2.5-7B-VL in w8a8_fp8 quantization and it works without errors, checked on local corporate benchmark - the accuracy dropped by about 3 %.

Is 3% comparing before and after quantization?

Swipe4057:
> > Hi, I tested it on Qwen2.5-7B-VL in w8a8_fp8 quantization and it works without errors, checked on local corporate benchmark - the accuracy dropped by about 3 %.
> 
> Is 3% comparing before and after quantization?

yes

b8zhong:
> Can you point me to where the MMMU accuracy tests are? Are they local or run in CI?

https://github.com/sgl-project/sglang/blob/main/benchmark/mmmu/README.md

PanJason:
@b8zhong Hi I just tested the benchmark on my local machine with no quantization, `w8a8_fp8` quantization and `fp8` quantization. The following is the command I used for server. I dropped `--enable-mixed-chunk` because of #10623 .
```bash
python3 -m sglang.launch_server --model-path Qwen/Qwen2.5-VL-7B-Instruct  --tp 1 --dp 1 --port 8001 --host 0.0.0.0 \ 
--mem-fraction-static 0.9 --schedule-conservativeness 0.3 --grammar-backend xgrammar --enable-metrics \ 
--allow-auto-truncate --enable-multimodal --mm-attention-backend fa3 --attention-backend flashinfer \ 
--schedule-policy fcfs
```

On the client side, the same concurrency level as https://github.com/sgl-project/sglang/blob/main/benchmark/mmmu/README.md
```bash
python3 bench_sglang.py --port 8001 --concurrency 16
```
## Results:
| Quantization    | Overall accuracy |
| -------- | ------- |
| No quantization  | 0.516    |
| w8a8_fp8 |  0.506     |
| fp8    |  0.507    |

## Test env
[test.txt](https://github.com/user-attachments/files/22460767/test.txt)

Let me know if anything else needs to be done for this patch

mickqian (approved):

JustinTong0323 (approved):
LGTM.

BBuf (approved):

PR #10115, opened 2025-09-06
[Do not merge] Add pass manager to unify kernel fusion pattens

BBuf:
## Motivation

I think we’ve reached the point where this abstraction is necessary. For example, the `fused_allreduce_residual_rmsnorm`(It can also handle `fp8/fp4` quant) pattern has already required duplicate logic implementations for three different models. Moreover, models like DeepSeek V3 and GPT-OSS now contain numerous kernel fusion patterns. The worst part is that many of these patterns could be reused across similar model architectures, yet we’re currently forced to repeatedly modify Python code for each model implementation.

We can manage kernel fuse pattern by `torch.fx`, this pr implement a `torch.fx` Pass Manager to do `silu_mul` kernel fuse. Please keep the discussion going.

## Checklist

### DIFF
--- a/python/sglang/srt/fx_pass_manager.py
+++ b/python/sglang/srt/fx_pass_manager.py
@@ -1,123 +1,15 @@
-# Copyright 2023-2024 SGLang Team
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-#     http://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-# ==============================================================================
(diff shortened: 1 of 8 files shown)


DevashishLal-CB:
Got the RFC done https://github.com/sgl-project/sglang/issues/10118

python/sglang/srt/fx_pass_manager.py
```diff
@@ -0,0 +1,157 @@
...
+    def is_applicable(self, graph: fx.Graph) -> bool:
+        for node in graph.nodes:
+            if self._is_silu_node(node):
+                return True
+        return False
+
+    def apply(self, graph: fx.Graph) -> bool:
+        modified = False
+
+        for node in list(graph.nodes):
```
DevashishLal-CB: any reason for not using the inductor pattern matcher ?

BBuf: I also think we should use the inductor pattern matcher. However, we need to use direct_register_custom_op to register more operators from sgl-kernel. I look forward to you submitting a PR soon so we can work together to advance this effort. PR 10115 is just a Python-level experiment that I won’t merge - let’s go with the inductor pattern matcher approach as the standard

Issue #10070, opened 2025-09-05
[Bug] Deadlocks when running Qwen3-32B-FP8 with hicache


SzymonOzog:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Describe the bug

When running Qwen-32B-FP8 with hierarchical cache enabled, the worker deadlocks upon hitting the cache, one TP thread is already running the model while the other one is stuck eternally synchronizing the cuda stream:
TP0:
```
Thread 424435 (idle): "MainThread"
    all_reduce (torch/distributed/distributed_c10d.py:2931)
    wrapper (torch/distributed/c10d_logger.py:81)
    writing_check (hiradix_cache.py:187)
    check_hicache_events (hiradix_cache.py:378)
    get_new_batch_prefill (scheduler.py:1608)
    get_next_batch_to_run (scheduler.py:1548)
    event_loop_overlap (scheduler.py:814)
    decorate_context (torch/utils/_contextlib.py:120)
    run_scheduler_process (scheduler.py:2669)
    run (multiprocessing/process.py:108)
    _bootstrap (multiprocessing/process.py:314)
    _main (multiprocessing/spawn.py:135)
    spawn_main (multiprocessing/spawn.py:122)
    <module> (<string>:1)
```
TP1:
```
Thread 427258 (active): "MainThread"
    synchronize (torch/cuda/streams.py:231)
    resolve_last_batch_result (tp_worker_overlap_thread.py:218)
    process_batch_result_decode (scheduler_output_processor_mixin.py:211)
    process_batch_result (scheduler.py:1866)
    event_loop_overlap (scheduler.py:839)
    decorate_context (torch/utils/_contextlib.py:120)
    run_scheduler_process (scheduler.py:2669)
    run (multiprocessing/process.py:108)
    _bootstrap (multiprocessing/process.py:314)
    _main (multiprocessing/spawn.py:135)
    spawn_main (multiprocessing/spawn.py:122)
[...]

PR #10127, opened 2025-09-07
[Refactor] Remove Hicache Load & Write threads

DarkSharpness:
## Motivation

We actually don't need OS primitives and threads to implement hicache loading & writing. Instead, we can just use `cudaEvent` to synchronize between hicache operations and model execution.

## Modifications

1st commit, we remove a potential race condition in tp_worker_overlap. Now the the consumer index will be updated exactly once (in the overlap thread). We also fix a little typing errors by the way.

2nd commit, we remove the load thread for hicache. We use `cudaEvent` instead of OS level synchronization. In addition, for batches without hicache loading, we will skip the loading check (`cudaEvent` synchronization) to ensure there's no overhead.

3rd commit, we further remove the write thread for hicache. We also optimize the unnecessary TP sync in `writing_check` when there are no ongoing writes.

## Checklist

### DIFF
--- a/python/sglang/srt/managers/cache_controller.py
+++ b/python/sglang/srt/managers/cache_controller.py
@@ -18,7 +18,7 @@ import math
 import threading
 import time
 from queue import Empty, Full, PriorityQueue, Queue
-from typing import TYPE_CHECKING, List, Optional, Set, Tuple
+from typing import TYPE_CHECKING, List, NamedTuple, Optional, Set, Tuple
 
 import torch
 
@@ -133,6 +133,9 @@ class CacheOperation:
     @staticmethod
     def merge_ops(ops: List[CacheOperation]) -> CacheOperation:
         assert len(ops) > 0
+        if len(ops) == 1:
+            return ops[0]
+
         host_indices = torch.cat([op.host_indices for op in ops])
         device_indices = torch.cat([op.device_indices for op in ops])
         node_ids = []
@@ -166,6 +169,12 @@ class CacheOperation:
 
     def __lt__(self, other: CacheOperation):
         return self.priority < other.priority
+
+
+class HiCacheAck(NamedTuple):
+    start_event: torch.cuda.Event
+    finish_event: torch.cuda.Event
+    node_ids: List[int]
 
 
 class TransferBuffer:
@@ -390,11 +399,11 @@ class HiCacheController:
         ]:
             raise ValueError(f"Invalid write policy: {write_policy}")
 
-        self.write_queue = PriorityQueue[CacheOperation]()
+        # self.write_queue = PriorityQueue[CacheOperation]()
         self.load_queue: List[CacheOperation] = []
-
-        self.ack_write_queue = Queue()
-        self.ack_load_queue: List[Tuple[torch.cuda.Event, List[int]]] = []
+        self.write_queue: List[CacheOperation] = []
+        self.ack_load_queue: List[HiCacheAck] = []
+        self.ack_write_queue: List[HiCacheAck] = []
 
         self.stop_event = threading.Event()
         self.write_buffer = TransferBuffer(self.stop_event)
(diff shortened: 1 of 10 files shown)


SzymonOzog:
Amazing! This also fixes https://github.com/sgl-project/sglang/issues/10070

python/sglang/srt/managers/tp_worker.py
```diff
@@ -230,6 +234,9 @@ def forward_batch_generation(
     ) -> Tuple[
         Union[LogitsProcessorOutput, torch.Tensor], Optional[torch.Tensor], bool
     ]:
+        # update the consumer index of hicache to the running batch
+        self.set_hicache_consumer(model_worker_batch.hicache_consumer_index)
+
```
xiezhq-hermann, line 239: I think postpone the set to here would increase the likelihood of io stall

python/sglang/srt/managers/cache_controller.py
```diff
@@ -502,90 +558,67 @@ def load(
...
         return device_indices
 
-    def move_indices(self, host_indices, device_indices):
+    def move_indices(self, op: CacheOperation):
+        host_indices, device_indices = op.host_indices, op.device_indices
         # move indices to GPU if using kernels, to host if using direct indexing
         if self.io_backend == "kernel":
-            return host_indices.to(self.mem_pool_device.device), device_indices
+            if not host_indices.is_cuda:
+                host_indices = host_indices.pin_memory().to(
```
xiezhq-hermann: no need to specify pin_memory() here

python/sglang/srt/managers/cache_controller.py
```diff
@@ -43,39 +43,63 @@
...
+        self.load_events[layer_index].record()
+
+    def wait(self, layer_index: int):
+        assert 0 <= layer_index < self._num_layers
+        if layer_index in self.wait_set:
+            return
+        self.wait_set.add(layer_index)
+        stream = torch.cuda.current_stream()
+        # make current stream wait for the load event of the layer
+        self.load_events[layer_index].wait(stream)
```
xiezhq-hermann: is it only for waiting for the copy of indices? if so we should only do a wait outside instead of here

python/sglang/srt/managers/cache_controller.py
```diff
@@ -502,90 +558,67 @@ def load(
...
+            # this is because we need to guarantee that these tensors are
+            # still alive when the load stream is executing.
+            if host_indices.is_cuda:
+                host_indices.record_stream(self.load_stream)
+            if device_indices.is_cuda:
+                device_indices.record_stream(self.load_stream)
+
+        self.ack_load_queue.append(
+            HiCacheAck(
+                start_event=producer_event.start_event,
```
xiezhq-hermann, line 575: why do need to return this start_event in the ack queue?

python/sglang/srt/managers/cache_controller.py
```diff
@@ -43,39 +43,63 @@
...
+
+    def complete(self, layer_index: int):
+        assert 0 <= layer_index < self._num_layers
+        self.load_events[layer_index].record()
+
+    def wait(self, layer_index: int):
+        assert 0 <= layer_index < self._num_layers
+        if layer_index in self.wait_set:
+            return
+        self.wait_set.add(layer_index)
```
xiezhq-hermann: I don't get what does this `wait_set` mean and do

python/sglang/srt/managers/cache_controller.py
xiezhq-hermann: I don't understand why we do wait(stream) here instead of simply wait for the completion of the event

python/sglang/srt/managers/cache_controller.py
```diff
@@ -43,39 +43,63 @@
...
+        self.events = [LayerEvent(num_layers) for _ in range(self.num_counters)]
+        self.producer_index = -1
+        self.consumer_index = -1
 
     def update_producer(self):
-        self.producer_index = self.next_producer()
+        self.producer_index = (self.producer_index + 1) % self.num_counters
+        assert self.events[
+            self.producer_index
+        ].finish_event.query(), (
```
xiezhq-hermann, line 77: I think there is no need to check it here, instead we should have things reset after the `finished_event` got queried.

python/sglang/srt/managers/cache_controller.py
xiezhq-hermann: does it actually mean the completed_set of layers?

xiezhq-hermann:
@DarkSharpness I have done some further cleaning, can you help do a test?

xiezhq-hermann (approved):

PR #10132, opened 2025-09-07
Refactor kv_cache_scheme handling for quantization

mmangkad:
## Motivation

Currently, the model loader crashes when trying to load certain quantized models.

The root cause is that the `from_config` methods in `ModelOptFp4Config` and `ModelOptFp8Config` rigidly expect the `kv_cache_scheme` parameter in `config.json` to be a **dictionary**. However, some models define this value as a simple **string** (e.g., `"FP8"`). This leads to a crash when the code attempts to call `.get()` on the string value, raising the following error:

```
  File ".../srt/model_loader/weight_utils.py", line 156, in get_quant_config
    return quant_cls.from_config(hf_quant_config)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../srt/layers/quantization/modelopt_quant.py", line 592, in from_config
    and kv_cache_scheme.get("type") == "float"
        ^^^^^^^^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'get'
```

This fragility prevents users from loading these valid models out-of-the-box and forces them to manually edit the configuration files. This PR makes the configuration parsing more robust to improve user experience and broaden model compatibility.

## Modifications

  - Refactored `kv_cache_scheme` handling in both `ModelOptFp4Config` and `ModelOptFp8Config` to gracefully accept either a dictionary or a string value.
  - Added type-checking to safely parse the configuration, ensuring the loader no longer crashes on models that use the shorthand string format.


mmangkad:
Hi @zhyncs, is it possible to merge this? Thanks!

LorrinWWW:
@mmangkad can you fix lint error by `pre-commit run --all-files`

mmangkad:
> @mmangkad can you fix lint error by `pre-commit run --all-files`

@LorrinWWW done

PR #10133, opened 2025-09-07
Introduce naming convention in `io_struct` and base sglang io classes.

hnyls2002:
This PR introduces the base classes of IO dataclasses: `BaseReq` and `BaseBatchReq`, and a naming convention for all the IO structs inside `io_struct.py`.

After this, we can:
- Assign an RID for each request, no matter what type they are.
- Better debugging tracing use this RID.
- Route requests through RID when there are multiple HTTP workers (#11172)
- Add some common fields in the future.


CatherineSue (approved):
Changes related to grpc_request_manager.py LGTM

PR #10143, opened 2025-09-08
metrics: support customer labels specified in request header

acelyc111:
## Motivation

The SGLang inference service may servering multi-tenants or used in different scenarios. In the current monitoring implementation, we can't distinguish the request/response feature (e.g. `prompt_tokens_histogram` metric).

## Modifications

Add new labels for tokenizer metrics to distinguish tenants or scenarios. The labels are specified in requests' HTTP header specified by `--tokenizer-metrics-custom-labels-header`(default is `x-customer-labels`) in a dict form and limited by server args `--tokenizer-metrics-allowed-customer-labels`. Then the output tokenizer metrics are labeled by extra labels specified in requests, or empty if not specified.

For example:
start server:
```
    --model-path /root/.cache/modelscope/hub/models/Qwen/Qwen3-0.6B \
    --port 30200 \
    --base-gpu-id 1 \
    --enable-metrics \
    --enable-metrics-for-all-schedulers \
    --collect-tokens-histogram \
    --tokenizer-metrics-allowed-customer-labels customer_label1  # new added
```

request:
```
curl -X POST http://localhost:30200/v1/chat/completions \
    -H 'Content-Type:application/json' \
    -H 'x-customer-labels: {"customer_label1": "v1"}' \  # new header
    -d '{"messages": [{"role": "user",
                       "content": "when 1+1=3?"}],
         "max_tokens": 1024,
         "stream": true}'
```

metrics:
```
...
sglang:prompt_tokens_histogram_count{customer_label1="",model_name="/root/.cache/modelscope/hub/models/Qwen/Qwen3-0.6B"} 1.0
sglang:prompt_tokens_histogram_bucket{customer_label1="v1",le="100.0",model_name="/root/.cache/modelscope/hub/models/Qwen/Qwen3-0.6B"} 1.0
sglang:prompt_tokens_histogram_bucket{customer_label1="v1",le="300.0",model_name="/root/.cache/modelscope/hub/models/Qwen/Qwen3-0.6B"} 1.0
sglang:prompt_tokens_histogram_bucket{customer_label1="v1",le="500.0",model_name="/root/.cache/modelscope/hub/models/Qwen/Qwen3-0.6B"} 1.0
sglang:prompt_tokens_histogram_bucket{customer_label1="v1",le="700.0",model_name="/root/.cache/modelscope/hub/models/Qwen/Qwen3-0.6B"} 1.0
sglang:prompt_tokens_histogram_bucket{customer_label1="v1",le="1000.0",model_name="/root/.cache/modelscope/hub/models/Qwen/Qwen3-0.6B"} 1.0
...
```

## Checklist


whybeyoung:
LGTM

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -122,6 +123,23 @@ def _convert_to_internal_request(
             else:
                 prompt_kwargs = {"input_ids": processed_messages.prompt_ids}
 
+        customer_labels = None
+        if self.tokenizer_manager.server_args.tokenizer_metrics_custom_labels:
+            raw_labels = (
+                json.loads(raw_request.headers.get("x-customer-labels"))
```
whybeyoung: "x-customer-labels" can be a configurable item. This is because in certain scenarios, gateway headers may need to comply with specific rules, which can vary across different scenarios.

python/sglang/srt/entrypoints/openai/serving_completions.py
```diff
@@ -85,6 +87,23 @@ def _convert_to_internal_request(
         else:
             prompt_kwargs = {"input_ids": prompt}
 
+        customer_labels = None
+        if self.tokenizer_manager.server_args.tokenizer_metrics_custom_labels:
+            raw_labels = (
+                json.loads(raw_request.headers.get("x-customer-labels"))
```
whybeyoung: the same as above

python/sglang/srt/entrypoints/openai/serving_chat.py
whybeyoung: btw, json.loads may throw exception, should be hanlde

python/sglang/srt/server_args.py
```diff
@@ -1210,6 +1211,15 @@ def add_cli_args(parser: argparse.ArgumentParser):
             "to record request metrics separately. This is especially useful when dp_attention is enabled, as "
             "otherwise all metrics appear to come from TP 0.",
         )
+        parser.add_argument(
+            "--tokenizer-metrics-custom-labels",
```
whybeyoung: Here, --tokenizer-metrics-custom-labels refers to the configuration of allowed labels, right? Could we rename it to a term that better aligns with its functional semantics, such as --tokenizer-metrics-allowed-labels?

whybeyoung (changes requested):

python/sglang/srt/server_args.py
acelyc111: Renamed it to --tokenizer-metrics-allowed-customer-labels

whybeyoung (approved):

hnyls2002:
@CatherineSue, could you check the OpenAI-related changes?

slin1237:
why do we modify http server for this?
the responsibility boundary between model engine and production/service component becomes blurry

whybeyoung:
> why do we modify http server for this? the responsibility boundary between model engine and production/service component becomes blurry

I have encountered similar scenarios related to this PR, and many others companies  also have similar requirements. Specifically, the engine side should provide ways to add custom labels to requests. In MaaS (Model as a Service) scenarios, this allows the upper layer to conduct data analysis on requests from different dimensions. for instance, I may need to count the request traffic characteristics of different users. In essence, this is essentially adding labels to metrics. @slin1237

acelyc111:
@slin1237  It's a requirement from real usage cases, and this is a flexible implementation, users can disable this feature completely.

The way to pass extra info through HTTP header is lighter, e.g. the openai sdk pass organization/project through http header, https://github.com/openai/openai-python/blob/847ff0b83841d9262ba0d9c4fdf46f0478004ad0/src/openai/_client.py#L329

slin1237 (approved):

zhyncs (approved):

PR #10153, opened 2025-09-08
Add support for AutoRound quantized models

WeiweiZhang1:
This pr is to support models quantized by AutoRound [github](https://github.com/intel/auto-round) [paper](https://arxiv.org/abs/2309.05516),

AutoRound delivers significantly higher accuracy at extremely low bit-widths (e.g., 2-bit) and offers broader compatibility across models (LLMs and VLMs), quantization formats, and configurations. You can check out our github/paper or this [blog post](https://huggingface.co/blog/autoround).

AutoRound has been integrated into [vllm](https://github.com/vllm-project/vllm), [pytorch/ao](https://github.com/pytorch/ao) and [Hugging Face Transformers](https://github.com/huggingface/transformers). Several Hugging Face Spaces offer models quantized with AutoRound, including [OPEA](https://huggingface.co/OPEA), [Kaitchup](https://huggingface.co/kaitchup), and [fbaldassarri](https://huggingface.co/fbaldassarri).

Known issues
Mixed bits support is limited
Mixed-bit quantization is currently limited. Since vLLM fuses layers (e.g., QKV), applying different bit-widths to components within the same fused layer can lead to incompatibility issues.

Limited Support for Quantized MoE Models
Quantized MoE models may encounter inference issues due to kernel limitations (e.g., lack of support for mlp.gate layer quantization). To avoid such errors, please skip quantizing 'gate' layers when processing quantization to MoE modules.

Quantized vlms support is limited
Qwen2.5-VL-7B : auto_round:auto_gptq format accuracy is close to zero. gptq model has the ‘The output size is not aligned with the quantized weight shape’ issue. auto_round:auto_awq and awq format are fine


python/sglang/srt/layers/quantization/auto_round.py
```diff
@@ -0,0 +1,394 @@
...
+            raise ValueError(
+                f"Unsupported backend: {backend},  "
+                f"currently only support  {self.SUPPORTED_BACKENDS}"
+            )
+
+        self.weight_bits = weight_bits
+        self.group_size = group_size
+        self.sym = sym
+        self.packing_format = packing_format
+        self.block_name_to_quantize = (
```
BBuf, line 68: Can we add a check `if block_name_to_quantize is not None`.

python/sglang/srt/layers/quantization/auto_round.py
```diff
@@ -0,0 +1,394 @@
...
+            raise ValueError(
+                f"Unsupported data_type: {data_type},"
+                f" currently only support  {self.SUPPORTED_DTYPES}"
+            )
+        if packing_format not in self.SUPPORTED_FORMATS:
+            raise ValueError(
+                f"Unsupported packing_format: {packing_format}, "
+                f"currently only support  {self.SUPPORTED_FORMATS}"
+            )
+        if backend not in self.SUPPORTED_BACKENDS:
```
BBuf, line 58: Should we check `group_size` value too?

python/sglang/srt/layers/quantization/auto_round.py
```diff
@@ -0,0 +1,394 @@
...
+            }
+            use_marlin = (weight_bits, sym) in GPTQ_TYPE_MAP and check_marlin_supported(
+                GPTQ_TYPE_MAP[(weight_bits, sym)], group_size, has_zp=not sym
+            )
+            if isinstance(layer, FusedMoE):
+                use_marlin = use_marlin and check_moe_marlin_supports_layer(
+                    layer, group_size
+                )
+        else:
+            use_marlin = False
```
BBuf, line 337: Line `303-337` is the same as line `221-256` ? Can we simplify it.

python/sglang/srt/layers/quantization/auto_round.py
WeiweiZhang1: block_name_to_quantize could be None. split() is only applied if it is a string, and the current behavior is as expected

python/sglang/srt/layers/quantization/auto_round.py
WeiweiZhang1: only the import statements and print/debug logs are the same here, so it is not recommended to abstract them into a separate function

python/sglang/srt/layers/quantization/auto_round.py
```diff
@@ -0,0 +1,394 @@
...
+
+class AutoRoundConfig(QuantizationConfig):
+    """Config class for AutoRound.
+    Reference: https://arxiv.org/pdf/2309.05516
+    """
+
+    SUPPORTED_BITS = {2, 3, 4, 8}
+    SUPPORTED_DTYPES = {"int"}
+    SUPPORTED_FORMATS = {"auto_round:auto_gptq", "auto_round:auto_awq"}
+    SUPPORTED_BACKENDS = {"auto", "gptq", "gptq:marlin", "awq", "awq:marlin", "marlin"}
```
AniZpZ, line 29: add supported group_size?

python/sglang/srt/layers/quantization/auto_round.py
```diff
@@ -0,0 +1,394 @@
...
+        self.packing_format = packing_format
+        self.block_name_to_quantize = (
+            block_name_to_quantize.split(",")
+            if isinstance(block_name_to_quantize, str)
+            else block_name_to_quantize
+        )
+        self.extra_config = extra_config
+        self.data_type = data_type
+        self.backend = backend
+        self.pack_factor = Fraction(32, weight_bits)
```
AniZpZ, line 76: it seems nerver used

wenhuach21: it will be used in some checkers as we leverages the 3rd party kernels for cuda, e.g. https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/layers/quantization/gptq.py#L424

python/sglang/srt/layers/quantization/auto_round.py
wenhuach21: Different backends/kernels support different `group_size` values, and specific kernels may be updated to support new ones. Therefore, it is preferable to rely on the runtime checker within each kernel

python/sglang/srt/layers/quantization/auto_round.py
WeiweiZhang1: Sorry, just realized I didn’t send out the reply to this comment successful.
As my peer mentioned, different backends/kernels support different group_size values, and specific kernels may be updated to support new ones in the future. Therefore, it’s preferable to rely on the runtime checker within each kernel rather than introducing hardcoded restrictions here.

python/sglang/srt/layers/quantization/auto_round.py
wenhuach21: Another reason is that some backends, such as Torch, may support many group_size values, which are not easy to list here, especially since we don’t have the exact data.

mingfeima:
enable CI

BBuf (approved):

FlamingoPg:
Looks good, some conflict and lint need to fix

AniZpZ (approved):

FlamingoPg (approved):
LGTM

wenhuach21:
@AniZpZ @BBuf  @FlamingoPg  @zhyncs  Thank you all for your kind reviews and helpful support.

PR #10162, opened 2025-09-08
Split toml files for sglang package

ZailiWang:
## Motivation

Split the `.toml` files for `sglang` package, to better facilitate the planned per-device `sgl-whl` building.

## Modifications

- Simplified the default `pyproject.toml` file for building whls only supporting cuda, created `pyproject_[cpu|hip|xpu|hpu|npu].toml` files for other devices;
- Removed the `[project.optional-dependencies]` hierarchical dependencies. Only 1 `dependencies` list in each `.toml` file (except test related packages);
- Commands in the doc & Dockerfiles require updating. Updated CPU for now, may collaborate to update that of other devices with the respective admins.

## Accuracy Tests

N/A

## Benchmarking and Profiling

N/A

## Checklist

### DIFF
--- a/docker/Dockerfile.xeon
+++ b/docker/Dockerfile.xeon
@@ -32,16 +32,17 @@ ENV CONDA_PREFIX=/sgl-workspace/miniforge3
 
 RUN pip config set global.index-url https://download.pytorch.org/whl/cpu && \
     pip config set global.extra-index-url https://pypi.org/simple && \
-    pip install intel-openmp
 
 RUN git clone https://github.com/sgl-project/sglang.git && \
     cd sglang && \
     git checkout ${VER_SGLANG} && \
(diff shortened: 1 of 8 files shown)


hnyls2002:
It seems to be too complex and too much duplicate code sharing between different files. Do you have other methods to do this?

ZailiWang:
Hi @hnyls2002 , the toml split was requested by @zhyncs . It indeed introduces duplications, but it seems this is inevitable if we need to split the package configs into device specific ones. I searched official `pip install` tutorials but failed to find anything about hierarchical config file support.

Let's keep the content of the PR as-is until we get a conclusion whether this is what we need, or we can find any better approaches.

ZailiWang:
Will take care of CPU/XPU backends only

PR #10168, opened 2025-09-08
[FIX] Fix Prometheus multiproc metrics cleanup

zhanghaotong:
## Motivation

When using SGLang with the environment variable PROMETHEUS_MULTIPROC_DIR pre-set, I noticed that
```python
prometheus_multiproc_dir = tempfile.TemporaryDirectory(
    dir=os.environ["PROMETHEUS_MULTIPROC_DIR"]
)
``` 
does not work as expected. 

The Prometheus client still writes data directly into the directory specified by PROMETHEUS_MULTIPROC_DIR, and this data is not cleaned up when the program exits. 

For example, when PROMETHEUS_MULTIPROC_DIR is set to /tmp/test,
image

A new subdirectory is created under /tmp/test, but the data is still written directly to /tmp/test instead of into the subdirectory. 

image

Furthermore, when the main process exits, although the subdirectory is removed, the data is still not cleaned up, which is not the expected behavior.

image

## Modifications

To fix this, I made some modifications to ensure that the data generated by the Prometheus client is stored in a temporary subdirectory under PROMETHEUS_MULTIPROC_DIR, 

image

and that this data is automatically cleaned up when the engine shuts down.
image

## Checklist

### DIFF
--- a/python/sglang/srt/utils/common.py
+++ b/python/sglang/srt/utils/common.py
@@ -1446,7 +1446,7 @@ def set_prometheus_multiproc_dir():
         )
     else:
         prometheus_multiproc_dir = tempfile.TemporaryDirectory()
-        os.environ["PROMETHEUS_MULTIPROC_DIR"] = prometheus_multiproc_dir.name
+    os.environ["PROMETHEUS_MULTIPROC_DIR"] = prometheus_multiproc_dir.name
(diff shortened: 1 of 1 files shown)


zhanghaotong:
@merrymercy Could you please help review this PR when you have a moment? Thank you!

ChrisYangAI:
> @merrymercy Could you please help review this PR when you have a moment? Thank you!

The same problem showed in our production environment. Could you please help review this? Thx. @merrymercy

PR #10169, opened 2025-09-08
[3/N]DP refactor: Improve dp rank scheduling in PD disaggregation mode.

hnyls2002:
Related to #10174

### DIFF
--- a/python/sglang/srt/server_args.py
+++ b/python/sglang/srt/server_args.py
@@ -223,6 +223,8 @@ class ServerArgs:
     # Data parallelism
     dp_size: int = 1
     load_balance_method: str = "round_robin"
+    # FIXME: remove this after dp rank scheduling is fully supported with PD-Disaggregation
+    prefill_round_robin_balance: bool = False
 
     # Multi-node distributed serving
     dist_init_addr: Optional[str] = None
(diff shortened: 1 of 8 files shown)


python/sglang/srt/server_args.py
```diff
@@ -1384,6 +1394,12 @@ def add_cli_args(parser: argparse.ArgumentParser):
                 "minimum_tokens",
             ],
         )
+        parser.add_argument(
+            "--prefill-round-robin-balance",
+            default=ServerArgs.prefill_round_robin_balance,
+            action="store_true",
+            help="Prefill is round robin balanced. This is used to promise decode server can get the correct dp rank.",
+        )
```
ShangmingCai, line 1402: Should this be `--decode-round-robin-balance`? Cause decode_dp_rank is not required at the prefill side, but prefill_dp_rank is required for decode ranks during bootstrapping. So we should let the prefill do the % while decode can enable whatever the load balance strategy is, and know the prefill_dp_rank through %.

hnyls2002: Oh, this argument is just a reminder to suggest the user set the prefill instance's load balance method to `round_robin`. I just did some renaming and clarification, not changing the original logic.

PR #10180, opened 2025-09-08
Fix chunked prefix cache for nvfp4

wenscarl:
co-authored by @elfiegg 

## Motivation
To address issue  [9806](https://github.com/sgl-project/sglang/issues/9806)

## Accuracy Tests

```
server.sh
  --model-path nvidia/DeepSeek-R1-0528-FP4 \
  --trust-remote-code \
  --attention-backend trtllm_mla \ or flashinfer \
  --disable-radix-cache \
  --max-running-requests 256 \
  --chunked-prefill-size 2048 \
  --mem-fraction-static 0.89 \
  --max-prefill-tokens 32768 \
  --disable-cuda-graph \
  --tp 8 \
  --dp 8 \
  --enable-dp-attention \
  --quantization modelopt_fp4 

client.sh
python3 benchmark/gsm8k/bench_sglang.py --num-shots 8 --num-questions 1310 --parallel 1310

flashinfer attention backend
before:
Accuracy: 0.510
Invalid: 0.480

after:
Accuracy: 0.948
Invalid: 0.000


trtllm_mla attention backend
before:
Accuracy: 0.421
Invalid: 0.429

after:
Accuracy: 0.952
Invalid: 0.001
```
## Checklist


python/sglang/srt/models/deepseek_v2.py
```diff
@@ -1863,6 +1863,8 @@ def _chunked_prefix_attn_mha(
             k[..., self.qk_nope_head_dim :] = k_pe
 
             output, lse = self.attn_mha(q, k, v, forward_batch, save_kv_cache=False)
+			if forward_batch.num_prefix_chunks == 1:
```
Fridge003: I feel this is not correct. Since this operation will ignore the passed in accum_output
And also we can't make sure the number of prefix chunk is 1

wenscarl (commented):
Work around cutlass kernel for chunked prefix

python/sglang/srt/layers/attention/flashinfer_mla_backend.py
```diff
@@ -232,7 +232,12 @@ def __init__(
 
         self.fmha_backend = "auto"
         if is_sm100_supported():
-            self.fmha_backend = "cutlass"
+            # TODO(shuw): Flashinfer CUTLASS MHA kernel has accuracy issue
```
zhyncs: hi @yzh119 can you help take a look

Fridge003:
@wenscarl Will the accuracy of dpsk-r1 nvfp4 be back to normal after changing flashinfer kernel to fa2 version?

wenscarl:
> @wenscarl Will the accuracy of dpsk-r1 nvfp4 be back to normal after changing flashinfer kernel to fa2 version?
Yes. It's in the description.

python/sglang/srt/layers/attention/flashinfer_mla_backend.py
wenscarl: https://github.com/flashinfer-ai/flashinfer/issues/1663 to track.

Fridge003 (approved):
LGTM

python/sglang/srt/layers/attention/flashinfer_mla_backend.py
yzh119: Adding `head_dim_vo` field in plan function should address the issue.

elfiegg:
can we route the logic based on quantization temporarily ? FP8 model works well with cutlass backend, and using FA2 the perf would drop to 1/2.

python/sglang/srt/layers/attention/flashinfer_mla_backend.py
zhyncs: @wenscarl can you try this

wenscarl: @zhyncs it's not `head_dim_vo` though I missed that in the reproducer. The cause is still from kernel for certain shape. It's not directly related to chunked_prefix_cache since without disabling it, we still see accuracy drop.  Running 10x and compare to fa2.
```
python3 benchmark/gsm8k/bench_sglang.py   --num-questions 40   --parallel 40   --num-shots 1
FA2:
0.925 0.95 0.97 1.0 0.975 0.975 0.05 0.95 0.9
CUTLASS:
0.775 0.77 0.675 0.825 0.8 0.774 0.65 0.675 0.8 0.675
```

@elfiegg to provide more.

zhyncs:
```bash
pre-commit run --all-files
```

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -1095,6 +1108,7 @@ def _dispatch_mla_subtype():
                     )
                     or sum_extend_prefix_lens == 0
                 )
+                and not skip_chunked_mha
```
Fridge003: If we skip_chunked_mha, can we dispatch it to `AttnForwardMethod.MHA` instead?

wenscarl: Yes. Updated

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -1096,6 +1109,8 @@ def _dispatch_mla_subtype():
                     or sum_extend_prefix_lens == 0
                 )
             ):
+                if skip_chunked_mha:
+                    return AttnForwardMethod.MHA
                 return AttnForwardMethod.MHA_CHUNKED_KV
             else:
                 return _dispatch_mla_subtype()
```
elfiegg, line 1116: Can you add the same logic for TRTLLM too?

elfiegg (commented):
Thanks for the quick fix!

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -1083,6 +1083,19 @@ def _dispatch_mla_subtype():
...
+
+            original_mode = getattr(forward_batch, "_original_forward_mode", None)
+            skip_chunked_mha = False
+            # TODO(shuw@nvidia.com) Only flashinfer cutlass and cutlass_mla backend has accuracy issue on blackwell.
+            if (
+                original_mode is not None
+                and original_mode.is_decode()
+                and is_sm100_supported()
+                and self.current_attention_backend in ("cutlass_mla", "flashinfer")
+                and _is_fp4_quantization_enabled()
```
elfiegg: We might not need to differentiate fp4 quant now as `original_mode is not None original_mode.is_decode()` is triggering problematic kernels

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -1083,6 +1083,19 @@ def _dispatch_mla_subtype():
             disable_ragged = (
                 attention_backend == "flashinfer" or attention_backend == "flashmla"
             ) and self.flashinfer_mla_disable_ragged
+
+            original_mode = getattr(forward_batch, "_original_forward_mode", None)
+            skip_chunked_mha = False
+            # TODO(shuw@nvidia.com) Only flashinfer cutlass and cutlass_mla backend has accuracy issue on blackwell.
+            if (
```
elfiegg: You should probably directly do
```
return _dispatch_mla_subtype()
```
instead of MHA when hit this branch. You want to call MLA kernel in this case which doesn't involve any MHA at all

kushanam:
This PR already includes the changes from https://github.com/sgl-project/sglang/pull/10178

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -1083,6 +1083,18 @@ def _dispatch_mla_subtype():
             disable_ragged = (
                 attention_backend == "flashinfer" or attention_backend == "flashmla"
             ) and self.flashinfer_mla_disable_ragged
+
+            original_mode = getattr(forward_batch, "_original_forward_mode", None)
+            skip_chunked_mha = False
+            # TODO(shuw@nvidia.com) Only flashinfer cutlass and cutlass_mla backend has accuracy issue on blackwell.
```
Fridge003: In the TODO comment we can add link to the issue: https://github.com/sgl-project/sglang/issues/9806

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -1083,6 +1083,18 @@ def _dispatch_mla_subtype():
...
+            if (
+                original_mode is not None
+                and original_mode.is_decode()
+                and is_sm100_supported()
+                and self.current_attention_backend in ("cutlass_mla", "flashinfer")
+            ):
+                skip_chunked_mha = True
```
Fridge003: we can directly return _dispatch_mla_subtype() here and remove the `skip_chunked_mha` variable

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -1095,15 +1107,34 @@ def _dispatch_mla_subtype():
...
                 )
+                and not skip_chunked_mha
             ):
                 return AttnForwardMethod.MHA_CHUNKED_KV
             else:
                 return _dispatch_mla_subtype()
         elif attention_backend == "trtllm_mla":
+            # TODO(shuw@nvidia.com) Same reason as mentioned in flashinfer branch.
+            original_mode = getattr(forward_batch, "_original_forward_mode", None)
+            skip_chunked_mha = False
```
Fridge003: This line can be removed

Fridge003:
@wenscarl Please fix lint with
```
pre-commit run --all-files
```

fzyzcjy:
looking forward to this

PR #10196, opened 2025-09-09
[Draft] Add new LoRA backend & kernels based on chunked SGMV 

lifuhuang:
This PR is for demonstration purpose only. 

It will be broken into two smaller PRs for code review.


python/sglang/srt/lora/backend/base_backend.py
```diff
@@ -93,21 +93,62 @@ def run_gate_up_lora(
...
-        from sglang.srt.lora.backend.triton_backend import TritonLoRABackend
-
-        return TritonLoRABackend
-    elif name == "flashinfer":
-        raise ValueError(
-            "FlashInfer LoRA backend has been deprecated, please use `triton` instead."
-        )
-    else:
-        raise ValueError(f"Invalid backend: {name}")
+    from sglang.srt.lora.backend.chunked_backend import ChunkedSgmvLoRABackend
```
lifuhuang, line 138: Hack just to trigger CI

PR #10201, opened 2025-09-09
[4/N]DP refactor: support watching mode `get_load` and shortest queue strategy

hnyls2002:
### Watching mode communicator
- Watching Mode: If there is already an in-flight request, use the in-flight request's output.
- Queueing Mode: If there is an in-flight request, wait until the request has been finished, and then send the next request.

### Why choose the watching mode?

- The load's changing interval must be larger than a scheduler iteration. Requesting `get_load` too frequently is meaningless, and all the concurrent requests are actually acquiring the same value.

### Design
budget mmd


hnyls2002:
The load report mechanism will be integrated into the request (streaming)response. I will bring this up in future PRs and remove the periodic polling logic.

python/sglang/srt/managers/data_parallel_controller.py
```diff
@@ -127,6 +168,31 @@ def __init__(
...
+    def init_dispatcher(self):
+        self._request_dispatcher = TypeBasedDispatcher(
+            [
+                (TokenizedGenerateReqInput, self.dispatching),
+                (TokenizedEmbeddingReqInput, self.dispatching),
+                (BlockReqInput, self.send_to_all_workers),
+                (WatchLoadUpdateReq, self.handle_load_update_req),
+            ]
+        )
+        self._request_dispatcher.add_fallback_fn(self.send_control_message)
```
zhyncs, line 194:
```
    self._request_dispatcher.add_fallback_fn(self.send_control_message)
AttributeError: 'TypeBasedDispatcher' object has no attribute 'add_fallback_fn'
```

PR #10214, opened 2025-09-09
[Fix] KV-cache eviction mismatch across PP ranks in DeepSeek V3/R1

qhsc:
## Motivation

Fix KV-cache size mismatch across PP ranks that causes crashes with radix cache.
In a PP=2 setup with the DeepSeek V3/R1 model, the maximum total tokens reserved for the KV cache in pp0 and pp1 are not equal. When radix cache is enabled, after some time, the prefix cached num of tokens for the same req will be in-consistent among pp ranks due to asymmetric cache evict, which will lead to system crash.

image

Working with @jianfei-wangg.

## Checklist


yizhang2077 (approved):
LGTM

Issue #10218, opened 2025-09-09
[Feature] Add Observability for KV Eviction/Reload and Chunked Prefill (new_token_ratio, eviction/load_back durations, prefill loop count)


ShawnKung:
### Checklist

- [x] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 2. Please use English, otherwise it will be closed.

### Motivation

In production, we observed intermittent throughput drops and tail-latency spikes during bursts of long-context requests. We lacked visibility into whether the regressions were compute-bound or due to KV cache movement between GPU and CPU. To address this, we added four metrics to improve diagnosis and tuning:

- sglang:new_token_ratio (Gauge): Tracks the proportion of newly generated tokens relative to reused tokens, helping distinguish steady decoding vs. frequent context rebuilds or cache misses.
- sglang:eviction_duration_secon
- ds (Histogram): Time spent evicting memory from GPU to CPU, exposing memory-pressure and paging overhead.
- sglang:load_back_duration_seconds (Histogram): Time spent loading memory from CPU back to GPU, highlighting thrash and hot-cache misses.
- sglang:chunked_prefill_loop_count (Histogram): Number of loops in chunked prefill, indicating how fragmented or oversized prefill segments are under load.

With these, we were able to pinpoint periods where eviction/load-back dominated latency (e.g., P95 eviction ~80 ms, load-back ~120 ms during spikes) and where prefill loops frequently exceeded expected counts, signaling suboptimal chunk sizing. After tuning chunk sizes and eviction thresholds, we saw improved throughput and reduced p99 latency in our internal deployments.

Example usages:
  - Alert if eviction_duration_seconds or load_back_duration_seconds P95 > 50 ms sustained for N minutes.
  - Track new_token_ratio drops below an expected band (e.g., <0.3) during bursts, signaling excessive cache rebuilds.
  - Watch chunked_prefill_loop_count > 5 as a heuristic for suboptimal prefill chunking.

I will follow up with a PR to add these metrics.

<img width="1165" height="919" alt="Image" src="https://github.com/user-attachments/assets/11d382bc-ea09-40a6-906a-a9c507f6172f" />

PR #10225, opened 2025-09-09
feat(metrics): add scheduler and hiradix cache metrics (#10218)

ShawnKung:
## Motivation

See [Issue #10218](https://github.com/sgl-project/sglang/issues/10218) for detailed background and motivation.

## Checklist


python/sglang/srt/server_args.py
```diff
@@ -1266,6 +1269,27 @@ def add_cli_args(parser: argparse.ArgumentParser):
             default=ServerArgs.collect_tokens_histogram,
             help="Collect prompt/generation tokens histogram.",
         )
+        parser.add_argument(
+            "--bucket-eviction-duration",
+            type=float,
+            nargs="+",
+            default=ServerArgs.bucket_eviction_duration,
```
xiezhq-hermann: I think we shouldn't be putting them all into server args, but to have a default value and maybe configure through environment variables or a json file.

xiezhq-hermann:
Thanks a lot for the PR, these are all reasonable metrics to collect. My only complain is that we probably do not want to have all the minor configs into server config parameters as I mentioned in the comment.

python/sglang/srt/server_args.py
ShawnKung: Good idea! Putting everything into server args is indeed a bit verbose. I’m going to revise it and push a new commit.

ShawnKung:
> Thanks a lot for the PR, these are all reasonable metrics to collect. My only complain is that we probably do not want to have all the minor configs into server config parameters as I mentioned in the comment.

Good idea! I've moved them into env vars. Thanks for the suggestion.

python/sglang/srt/managers/scheduler.py
```diff
@@ -1695,6 +1701,7 @@ def get_new_batch_prefill(self) -> Optional[ScheduleBatch]:
             # only record queue time when enable_metrics is True to avoid overhead
             for req in can_run_list:
                 req.queue_time_end = time.perf_counter()
+                req.prefill_loop_count += 1
```
xiezhq-hermann: can you elaborate what this loop count is for? it actually feels quite similar to what the queuing time is already doing

python/sglang/srt/managers/scheduler_metrics_mixin.py
```diff
@@ -59,7 +60,29 @@ def init_metrics(
             }
             if dp_rank is not None:
                 labels["dp_rank"] = dp_rank
-            self.metrics_collector = SchedulerMetricsCollector(labels=labels)
+            self.metrics_collector = SchedulerMetricsCollector(
+                labels=labels,
+                bucket_eviction_duration=self.get_histogram_conf_from_env(
```
xiezhq-hermann: I think we can do the environment variable reading inside the `collector.py`

python/sglang/srt/mem_cache/hiradix_cache.py
```diff
@@ -293,6 +304,11 @@ def evict(self, num_tokens: int):
                 assert node.backuped
                 self._evict_backuped(node)
 
+        if num_evicted > 0 and self.scheduler_metrics_collector is not None:
+            self.scheduler_metrics_collector.observe_eviction_duration(
```
xiezhq-hermann: is this to measure the time taken for eviction to complete? is it primarily for understanding the eviction overhead or other purpose I am not aware here. This overhead will vary a lot based on the tokens required for eviction, simply recording the time probably wouldn't make too much sense

python/sglang/srt/managers/scheduler.py
ShawnKung: > can you elaborate what this loop count is for? it actually feels quite similar to what the queuing time is already doing

@xiezhq-hermann This is meant to capture something slightly different from queuing time.
- Queuing time = how long a request waits before prefill starts.
- Prefill loop count = how many rounds a request goes through during prefill.
The main purpose is to check if our chunking configuration is reasonable. For example, in some deployment scenarios, if most prefill requests go through multiple rounds, it may indicate that the chunked_prefill_size could be increased to reduce prefill rounds.

python/sglang/srt/managers/scheduler_metrics_mixin.py
ShawnKung: > I think we can do the environment variable reading inside the `collector.py`

good point

python/sglang/srt/mem_cache/hiradix_cache.py
ShawnKung: > is this to measure the time taken for eviction to complete? is it primarily for understanding the eviction overhead or other purpose I am not aware here. This overhead will vary a lot based on the tokens required for eviction, simply recording the time probably wouldn't make too much sense

@xiezhq-hermann You make a good point.
My initial idea was mainly to have a quick observable signal in cases where the system slows down noticeably, we can immediately see whether eviction or write-back is taking up significant time.
I think we could enhance the story by also recording **evicted/write-back tokens and total bytes**. With these, we could compute average time per token and effective eviction throughput, which would give a more complete picture of the overhead.
What do you think?

python/sglang/srt/managers/scheduler.py
xiezhq-hermann: I am a bit confused, when a certain request got chunked, the next chunk will also come from this request, which means the number of rounds of a request got processed when chunking happens is basically length / chunk size, which can be easily analysed with input length and the chunk size instead of recording a runtime metrics.

python/sglang/srt/mem_cache/hiradix_cache.py
xiezhq-hermann: I think recording number of tokens got evicted will be a helpful metrics to understand memory activity. e.g., we can plot the # evicted tokens over time to see the peak and low, etc.

xiezhq-hermann: overhead of eviction might not be something as interesting, but I can see that to be useful though. we are pushing for replacing hiradix tree by a cpp version, and then this overhead might be more insignificant.

python/sglang/srt/managers/scheduler.py
ShawnKung: @xiezhq-hermann I agree with your point that in theory the chunk count can be derived as length / chunk_size. However, since there’s currently no metric that reflects the actual chunk_size, when I need to compute the request’s chunk count I would have to hardcode the chunk size in the expression. That makes it less flexible and not aligned with the real deployment setup, where the chunk size may vary. Having a runtime metric is therefore more convenient and more direct in practice.

python/sglang/srt/mem_cache/hiradix_cache.py
ShawnKung: @xiezhq-hermann I agree — adding the number of evicted tokens is a good idea, I’ll include that metric.
As for the eviction time, while it may become less relevant once the c++ version is in place, I think it’s still useful at this stage. In our case, we may also develop a custom distributed storage backend for the hiradix tree, and having this metric framework in place will make it easier to extend with additional metrics we need later. That said, if you think the eviction time metric is unnecessary for the community, I can drop it.

ShawnKung:
@xiezhq-hermann hello, I've add these two new metric of evicted and load back token numbers,
- `sglang:evicted_tokens_total`: A new Counter metric that tracks the total number of tokens evicted from the GPU to the CPU.
- `sglang:load_back_tokens_total`: A new Counter metric that tracks the total number of tokens loaded back from the CPU to the GPU.

xiezhq-hermann:
thank you @ShawnKung for the continuing update, I have two final suggestions:
1. I still find the loop_count rather less meaningful, it is easier to just log the chunk size when start the serving engine and querying this over and over again is unnecessary
2. it feels rather weird to pass the scheduler_metrics_collector to every radix implementation, how about we just create a collector in the base radix cache and every derived class can just use it?

ShawnKung:
> thank you @ShawnKung for the continuing update, I have two final suggestions:
> 
> 1. I still find the loop_count rather less meaningful, it is easier to just log the chunk size when start the serving engine and querying this over and over again is unnecessary
> 2. it feels rather weird to pass the scheduler_metrics_collector to every radix implementation, how about we just create a collector in the base radix cache and every derived class can just use it?

@xiezhq-hermann Thank you for the suggestions. I've removed the `sglang:prefill_loop_count` metric since it was not very meaningful.  
I've also extracted a new metric collector `RadixCacheMetricsCollector` and a base cache class `BasePrefixCacheMetricsMixin` for cache-related metrics.

ShawnKung:
@xiezhq-hermann  Hi, just wanted to check in — I've addressed the latest comments.  
Please let me know if there are any other issues I should fix.
If everything looks good, I'll resolve the conflicts so it can be merged.
Thanks!

xiezhq-hermann:
> @xiezhq-hermann Hi, just wanted to check in — I've addressed the latest comments. Please let me know if there are any other issues I should fix. If everything looks good, I'll resolve the conflicts so it can be merged. Thanks!

I will handle the rest and get it merged soon thanks!

MMuzzammil1:
@ShawnKung thanks for the PR. Do you think it would also be nice to have metrics for the `evict_host` as well (either a flag for if the host memory is exhausted/ or simple log of tokens that are getting evicted from the host memory). This could be a good indicator to asses the memory status of L2 cache.

taegeonum:
@ShawnKung When do you think it is going to be merged? We'd like to use this feature to track cache/eviction related metrics.

ShawnKung:
@xiezhq-hermann Hi, just checking in — seems there’s more interest in this feature now.  
Is there any plan or timeline for merging?
Happy to help with anything needed to get it merged. Thanks!

MMuzzammil1 (commented):
Wouldn't this metrics_collector be overwritten by the HiRadixMetrics collector? I believe this should be fixed.

python/sglang/srt/mem_cache/hiradix_cache.py
```diff
@@ -135,6 +140,7 @@ def __init__(
             page_size,
             disable=False,
             eviction_policy=eviction_policy,
+            server_args=server_args,
```
MMuzzammil1: Wouldn't the `StorageMetricsCollector` (https://github.com/sgl-project/sglang/pull/10225/files#diff-8d800527a4772206625ba4f8e0fb4a6746e736098fdab17126e7335252b97a6bR122) be overwritten by the HiRadixMetrics collector? I believe this should be fixed.

ShawnKung: > Wouldn't the `StorageMetricsCollector` (https://github.com/sgl-project/sglang/pull/10225/files#diff-8d800527a4772206625ba4f8e0fb4a6746e736098fdab17126e7335252b97a6bR122) be overwritten by the HiRadixMetrics collector? I believe this should be fixed.

yes, I'm going to fix it, thanks!

xiezhq-hermann:
@ShawnKung I made some changes to it can you take a look and let me know how do you feel about it, thanks!

ShawnKung:
> @ShawnKung I made some changes to it can you take a look and let me know how do you feel about it, thanks!

@xiezhq-hermann  Thanks! The changes look good to me.

xiezhq-hermann (approved):

hnyls2002 (approved):

Issue #10226, opened 2025-09-09
EAGLE `verify` step re-computed the top-p sampling mask


KivenChen:
Following previous optimization, #9404, a conditional branch was introduced to avoid an expensive `top_p_renorm_prob` when we don't need top-p sampling. However, we suspect that this conditional branch might introduce another overhead b/c the top-p sampling mask is re-computed.

Attached is our flamegraph of running EAGLE-3 on H20, where the "is_nonzero" call becomes the overhead.

<img width="1599" height="933" alt="Image" src="https://github.com/user-attachments/assets/b86423a9-808f-434d-ab49-1f4cbb80747a" />

PR #10227, opened 2025-09-09
remove eagle3 top_p is_none_zero

jiapingW:
## Motivation

Fix the bug https://github.com/sgl-project/sglang/issues/10226. We directly use the value of need_top_p_sampling to reduce the non-zero judgment once. 

## Checklist


KivenChen:
Thanks for the prompt fix! I think this will bring significant performance gain.

jiapingW:
> Thanks for the prompt fix! I think this will bring significant performance gain.

I hope this can solve your problem. My initial idea was to delete the top_p_renorm big GPU kernel here in flashinfer.

KivenChen:
> > Thanks for the prompt fix! I think this will bring significant performance gain.
> 
> I hope this can solve your problem. My initial idea was to delete the top_p_renorm big GPU kernel here in flashinfer.

I can see this is quite heavy, although a fast path was introduced last month: https://github.com/flashinfer-ai/flashinfer/blame/c7e5583c93fd4646f04a26c49615b1fe729c8379/include/flashinfer/sampling.cuh#L1582

Do you plan to replace it in `eagle_utils`? Or how do you like to further optimize this kernel?

jiapingW:
> > > Thanks for the prompt fix! I think this will bring significant performance gain.
> > 
> > 
> > I hope this can solve your problem. My initial idea was to delete the top_p_renorm big GPU kernel here in flashinfer.
> 
> I can see this is quite heavy, although a fast path was introduced last month: https://github.com/flashinfer-ai/flashinfer/blame/c7e5583c93fd4646f04a26c49615b1fe729c8379/include/flashinfer/sampling.cuh#L1582
> 
> Do you plan to replace it in `eagle_utils`? Or how do you like to further optimize this kernel?

I haven't tried the new flashinfer kernel yet. My opinion is this: this operation and **renorm_top_k** can be performed both inside the attention backend like flashinfer and outside of it, I think running it inside the backend still incurs some overhead, and running it externally might be easier to manage across different backends and have lower overhead.

PR #6318, opened 2025-05-15
feat: limit peak memory usage when computing logprobs

aftersnow:
## Motivation

While computing input and output token log probabilities, we frequently encounter CUDA out-of-memory (OOM) errors, even after reducing the --mem-fraction-static to below 0.6.

Like this: 

```
  File "/sglang/python/sglang/srt/layers/logits_processor.py", line 311, in forward
    logits = self._get_logits(pruned_states, lm_head, logits_metadata)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/sglang/python/sglang/srt/layers/logits_processor.py", line 461, in _get_logits
    logits = logits[:, : self.config.vocab_size].float()
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 6.05 GiB. GPU 0 has a total capacity of 79.10 GiB of which 5.94 GiB is free. Process 428408 has 73.15 GiB memory in use. Of the allocated memory 65.99 GiB is allocated by PyTorch, with 80.15 MiB allocated in private pools (e.g., CUDA Graphs), and 4.45 GiB is reserved by PyTorch but unallocated. If reserved but unallocated memory is large try setting PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True to avoid fragmentation.  See documentation for Memory Management  (https://pytorch.org/docs/stable/notes/cuda.html#environment-variables)
```

And this:

```
  File "/sglang/python/sglang/srt/layers/logits_processor.py", line 375, in forward
    input_logprobs = self.compute_temp_top_p_normalized_logprobs(
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/sglang/python/sglang/srt/layers/logits_processor.py", line 547, in compute_temp_top_p_normalized_logprobs
    return torch.nn.functional.log_softmax(last_logits, dim=-1)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch/nn/functional.py", line 2248, in log_softmax
    ret = input.log_softmax(dim)
          ^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #10230, opened 2025-09-09
Add Logprobs unit test with a loose threshold

PrinsYin:
# Add Logprobs Unittest

## Motivation
We want to merge [#6318](https://github.com/sgl-project/sglang/pull/6318), so this PR adds a dedicated unittest suite to verify that the changes do not break **logprobs computation and interfaces**.  
The main goal is to catch indexing/corner-case issues and ensure stability/accuracy of logprobs against baseline data.

---

## Modifications
- Added `test/srt/test_logprobs.py`
  - Implements:
    - **`TestLogprobsDense`**: runs against `DEFAULT_SMALL_MODEL_NAME_FOR_TEST`
  - Configurable batch sizes, sample counts, and temperatures
  - Randomly samples records per run for coverage
  - Compares `input_top_logprobs` and `output_top_logprobs` against baselines
  - Validates both **max diff** and **mean diff** against thresholds
  - Writes per-config results into `$GITHUB_STEP_SUMMARY`

- Added to `run_suite.py` so it runs in CI
- Added retry logic for downloading Hugging Face baselines
- Enabled selective `return_logprob` in batches to check partial-request cases
- Cleaned up unsafe access and added tolerance-based failure reporting

---

## Baseline Data

WE ARE NOT TESTING MOE in THIS PR, the data below is just for the reference

- **Version**: `v0.5.1.post3`
- **Dense baseline**: [sglang_baseline_2000.pkl](https://huggingface.co/datasets/font-info/logprobs/blob/main/sglang_baseline_2000.pkl)
- **MoE baseline**: [sglang_baseline_moe.pkl](https://huggingface.co/datasets/font-info/logprobs/blob/main/sglang_baseline_moe.pkl)
- Generated ~1000 samples with `v0.5.1.post3`, uploaded to Hugging Face
- Each test run samples a few hundred records based on config
- Prompt length controlled to ~2000 tokens

---

## Tolerance Settings
**Dense (CUDA):**
- Max diff ≤ **1.5**
- Mean diff ≤ **0.1**

**Dense (ROCm):**
- Max diff ≤ **1.4**
- Mean diff ≤ **0.1**

**MoE (NVIDIA):**
- Max diff ≤ **10**
- Mean diff ≤ **0.3**

**MoE (ROCm):**
- Max diff ≤ **9.0**
- Mean diff ≤ **0.5**

---

## GitHub Summary Reporting
Each run appends a section like:


yushengsu-thu:
Add AMD baseline and thresholds


- Dense:
Max diff ≤ 1.4
Mean diff ≤ 0.1 (per sample)

- MoE:
Max diff ≤ 9.0
Mean diff ≤ 0.5 (per sample)

PrinsYin:
@narutolhy @zhaochenyang20 can you take a moment to review this? thanks!

test/srt/run_suite.py
```diff
@@ -91,6 +91,7 @@ class TestFile:
         TestFile("test_no_chunked_prefill.py", 108),
         TestFile("test_no_overlap_scheduler.py", 234),
         TestFile("test_original_logprobs.py", 200),
+        TestFile("test_logprobs.py", 200),
```
narutolhy: This file sorts the test files alphabetically. test_logprobs.py can be added near the beginning of  `test_l`

test/srt/test_logprobs.py
```diff
@@ -0,0 +1,307 @@
...
+    },
+    {
+        "batch_size": 20,
+        "num_samples": 1000,
+        "logprob_sample_ratio": 0.6,
+        "temperature": 1.0,
+    },
+]
+
+os.environ["RETURN_ORIGINAL_LOGPROB"] = "True"
```
narutolhy: Why do we need to turn on this switch? After turning it on, only the original logprobs will be returned. The different temperatures added in TEST_CONFIGS will not affect the returned logprobs.

test/srt/test_logprobs.py
```diff
@@ -0,0 +1,307 @@
...
+from sglang.test.test_utils import (
+    DEFAULT_SMALL_MODEL_NAME_FOR_TEST,
+    write_github_step_summary,
+)
+
+if torch.version.hip is not None:
+    print("Running on AMD ROCm GPU")
+    MOE_INPUT_PKL_URL = "https://huggingface.co/datasets/yushengsu/logprobs/resolve/main/sglang_baseline_moe_v0.5.1.pkl"
+    MOE_TOLERANCE_MAX_DIFF = 9.0
+    MOE_TOLERANCE_MEAN_DIFF = 0.5
```
Hecate0821:
```suggestion
 
```
Configs never used

test/srt/test_logprobs.py
```diff
@@ -0,0 +1,307 @@
...
+elif torch.version.cuda is not None:
+    print("Running on NVIDIA CUDA GPU")
+    MOE_INPUT_PKL_URL = "https://huggingface.co/datasets/font-info/logprobs/resolve/main/sglang_baseline_moe.pkl"
+    MOE_TOLERANCE_MAX_DIFF = 10
+    MOE_TOLERANCE_MEAN_DIFF = 0.3
```
Hecate0821:
```suggestion
 
```
Configs never used

test/srt/run_suite.py
PrinsYin: thanks！

test/srt/test_logprobs.py
PrinsYin: thanks！

test/srt/test_logprobs.py
PrinsYin: I think we enable this so that logprobs stay independent of temperature and we can purely test consistency across different batch/ratio cases in this test? @zhaochenyang20

test/srt/test_logprobs.py
```diff
@@ -0,0 +1,297 @@
...
+    DENSE_INPUT_PKL_URL = "https://huggingface.co/datasets/yushengsu/logprobs/resolve/main/sglang_baseline_v0.5.1.pkl"
+    DENSE_TOLERANCE_MAX_DIFF = 1.4
+    DENSE_TOLERANCE_MEAN_DIFF = 0.1
+elif torch.version.cuda is not None:
+    print("Running on NVIDIA CUDA GPU")
+    DENSE_INPUT_PKL_URL = "https://huggingface.co/datasets/font-info/logprobs/resolve/main/sglang_baseline_2000.pkl"
+    DENSE_TOLERANCE_MAX_DIFF = 1.5
+    DENSE_TOLERANCE_MEAN_DIFF = 0.1
+else:
+    print("No GPU backend (CPU only)")
```
zhaochenyang20, line 31: else raise error

test/srt/test_logprobs.py
```diff
@@ -0,0 +1,297 @@
...
+        "num_samples": 1000,
+        "logprob_sample_ratio": 0.5,
+        "temperature": 2.0,
+    },
+    {
+        "batch_size": 20,
+        "num_samples": 1000,
+        "logprob_sample_ratio": 0.6,
+        "temperature": 1.0,
+    },
```
zhaochenyang20: shall we have a temperature `0`, `0.5` `1.0`, `2.0`? No need to have a `10.0`

test/srt/test_logprobs.py
zhaochenyang20: Well. I think we do not need 4 group of tests. We only need one group of it. `num_samples 1000, ratio 0.5`.

test/srt/test_logprobs.py
```diff
@@ -0,0 +1,297 @@
...
+    @classmethod
+    def setUpClass(cls):
+        """Set up the test class - initialize the engine once for all tests."""
+        print(f"Launching SGLang Engine with {DENSE_MODEL_NAME}...")
+        cls.engine = sgl.Engine(
+            model_path=DENSE_MODEL_NAME,
+            random_seed=42,
+            skip_tokenizer_init=True,
+            mem_fraction_static=0.6,
+            max_running_requests=1,
```
zhaochenyang20: `mem_fraction_static` should be higger.

why setting `max_running_requests`? We only need to set batch size when we are making the baseline. For the test, do not do this.

test/srt/test_logprobs.py
```diff
@@ -0,0 +1,297 @@
...
+            )
+            for e1, e2 in zip(arrA, arrB):
+                if not e1 or not e2:
+                    continue
+                dmapA = {tid: lp for lp, tid, _ in e1}
+                dmapB = {tid: lp for lp, tid, _ in e2}
+                common = dmapA.keys() & dmapB.keys()
+                for tid in common:
+                    diffs.append(abs(dmapA[tid] - dmapB[tid]))
+        if not diffs:
```
zhaochenyang20: this if should have an else

test/srt/test_logprobs.py
```diff
@@ -0,0 +1,297 @@
...
+                )
+                print(
+                    f"Will return logprobs for {logprob_count} samples (ratio: {config['logprob_sample_ratio']})"
+                )
+
+                all_max, all_mean = [], []
+                logprob_returned_count = 0
+
+                for i in range(0, len(test_records), config["batch_size"]):
+                    batch = test_records[i : i + config["batch_size"]]
```
zhaochenyang20: why do we need to have a batch size. Could we send all the 1000 requests all at once?

test/srt/test_logprobs.py
```diff
@@ -0,0 +1,297 @@
...
+                            input_top_logprobs = (
+                                meta_info.get("input_top_logprobs")
+                                if meta_info
+                                else None
+                            )
+
+                            self.assertIsNotNone(
+                                input_top_logprobs,
+                                f"return_logprob enabled on this sample, but input_top_logprobs is None (length: {len(input_top_logprobs) if input_top_logprobs is not None else 'N/A'})",
+                            )
```
zhaochenyang20: also assert these we do not want log probs actaully do not return anything.

zhaochenyang20 (changes requested):

test/srt/test_logprobs.py
PrinsYin: oops, my bad

test/srt/test_logprobs.py
PrinsYin: It is asserted on line 236

test/srt/test_logprobs.py
PrinsYin: ofc!

test/srt/test_logprobs.py
narutolhy: I feel that if this is added, there is no need to compare the effects of different temperatures, because different temperatures do not affect the return value.

test/srt/test_logprobs.py
```diff
@@ -0,0 +1,261 @@
...
+TOP_K = 20
+MAX_RETRIES = 3
+RETRY_DELAY = 2
+
+# Test configurations
+TEST_CONFIG = {
+    "num_samples": 1000,
+    "logprob_sample_ratio": 0.5,
+    "temperature": 1.0,
+}
```
zhaochenyang20: No need to use a dict for this. Just as these:

```python
TOP_K = 20
MAX_RETRIES = 3
RETRY_DELAY = 2
```

zhaochenyang20 (approved):

test/srt/test_logprobs.py
```diff
@@ -0,0 +1,261 @@
...
+
+            except Exception as e:
+                print(f"Attempt {attempt + 1}/{MAX_RETRIES} failed: {e}")
+                if attempt == MAX_RETRIES - 1:
+                    raise Exception(
+                        f"Failed to load data after {MAX_RETRIES} attempts: {e}"
+                    )
+                time.sleep(RETRY_DELAY)
+
+    def compare_meta(self, metaA, metaB):
```
zhaochenyang20: do not name it as A or B. Use something like `baseline`.

zhaochenyang20 (changes requested):

zhaochenyang20:
Tested on H200 for 5 times:

```python
max Δ=0.624243
mean Δ=0.0032411
logprobs returned for 500 samples (expected: 500)
.
max Δ=0.624243
mean Δ=0.00346705
logprobs returned for 500 samples (expected: 500)
.
max Δ=0.624243
mean Δ=0.00395164
logprobs returned for 500 samples (expected: 500)

max Δ=0.624243
mean Δ=0.00399972
logprobs returned for 500 samples (expected: 500)

  loop = asyncio.get_event_loop()
max Δ=0.499887
mean Δ=0.00326239
logprobs returned for 500 samples (expected: 500)
```

merrymercy (approved):

zhaochenyang20 (approved):

PR #10232, opened 2025-09-09
fix: idle scheduler thread occupy 100% CPU. Add short sleep and cut CPU usage to 2%.

joybinchen:
## Motivation

The Sglang scheduler thread always occupy 100% cpu, even when it's idle and processing no request.
Try to cut unnecessary workload in the scheduler loop checking the queue when there is no request. 

## Modifications

Take a quick sleep when there is no request in the queue.

## Benchmarking and Profiling

Small delay(<1ms) increased to the TTFT when switching from idle state.
No effect to the overall throughput.

## Checklist

### DIFF
--- a/python/sglang/srt/managers/scheduler.py
+++ b/python/sglang/srt/managers/scheduler.py
@@ -1026,6 +1026,9 @@ class Scheduler(
                     except zmq.ZMQError:
                         break
                     recv_reqs.append(recv_rpc)
+
+                if not recv_reqs:
+                    time.sleep(0.001)
             else:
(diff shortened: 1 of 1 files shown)


hnyls2002 (changes requested):
Why not put the sleep into `self_check_during_idle`. When there are requests to process but no new requests coming, your modification would slow down each iteration by 0.001s (Unless in overlap mode).

joybinchen:
Thank you for quick responsive reply. It's a fake requirement. I should use --sleep-on-idle argument when running sglang serve.

Issue #10215, opened 2025-09-09
[Bug] cutlass_w4a8_moe_mm failed on hopper cu126 with cu129 whl


zhyncs:
### Checklist

- [ ] 1. I have searched related issues but cannot get the expected help.
- [ ] 2. The bug has not been fixed in the latest version.
- [ ] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [ ] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [ ] 5. Please use English, otherwise it will be closed.

### Describe the bug

fix https://github.com/sgl-project/sglang/blob/9ab72f9895c7ebd3b7cab06b76bc84e1976ff66b/sgl-kernel/tests/test_cutlass_w4a8_moe_mm.py#L141-L148

Starting from sgl-kernel v0.3.9, we default to using cu129. It performs well on cu128 and cu129, as well as in most cases on cu124 and cu126, with the exception of cutlass_w4a8_mm failing.

### Reproduction

N/A

### Environment

N/A

rainj-me:
Failed logs:

```bash
        # Compare results
        try:
>           torch.testing.assert_close(c, c_ref, rtol=1e-2, atol=0.1)
E           AssertionError: Tensor-likes are not close!
E
E           Mismatched elements: 28659 / 114688 (25.0%)
E           Greatest absolute difference: 1.5625 at index (6, 6778) (up to 0.1 allowed)
E           Greatest relative difference: 258048.0 at index (2, 4016) (up to 0.01 allowed)

test_cutlass_w4a8_moe_mm.py:246: AssertionError
------------------------------ Captured stdout call ------------------------------

Testing with batch_size=16, k=256, n=7168, num_experts=8
  FAILURE: tensors are NOT close.
    Max absolute difference: 1.5625
    Mean absolute difference: 0.07568359375
    AssertionError: Tensor-likes are not close!

```

FlamingoPg:
> Failed logs:
> 
>         # Compare results
>         try:
> >           torch.testing.assert_close(c, c_ref, rtol=1e-2, atol=0.1)
> E           AssertionError: Tensor-likes are not close!
> E
> E           Mismatched elements: 28659 / 114688 (25.0%)
[...]

PR #10241, opened 2025-09-10
pass a_scale from fp8 quant result instead of hard code to 1.0f 

rainj-me:
coauthor with @yicwang  @ayrnb 

## Motivation

Issue: #10215 

Fix the w4afp8 test case failures on hopper

## Modifications

- Fix the w4afp8 kernel pass the quantization a_scale instead of hard coding to 1.0f
- Fix the failed test cases

## Accuracy Tests

- Unit tests

```bash
 python test_cutlass_w4a8_moe_mm.py
======================================================================= test session starts =======================================================================
platform linux -- Python 3.12.11, pytest-8.4.1, pluggy-1.6.0
rootdir: /sgl-workspace/sglang/sgl-kernel
configfile: pyproject.toml
plugins: typeguard-4.4.4, anyio-4.10.0
collected 405 items                                                                                                                                               

test_cutlass_w4a8_moe_mm.py ............................................................................................................................... [ 31%]
........................................................................................................................................................... [ 69%]
...........................................................................................................................                                 [100%]

======================================================================= 405 passed in 3.09s =======================================================================
```

```bash
curl http://127.0.0.1:8000/v1/chat/completions -H "Content-Type: application/json" -d '{"model": "/data01/models/DeepSeek-R1-W4AFP8", "temperature": 0, "messages": [ {"role": "user", "content": "How to travel to New York from San Francisco"}], "max_tokens": 200, "temperature": 0}'
{"id":"726e6ae588d944d38454749054b44dd4","object":"chat.completion","created":1757473837,"model":"/data01/models/DeepSeek-R1-W4AFP8","choices":[{"index":0,"message":{"role":"assistant","content":"Okay, so I need to figure out how to travel from San Francisco to New York. Let me start by thinking about the different ways people usually travel between cities. The main options are flying, driving, taking a train, or a bus. Maybe there are other less common methods too, like cycling or walking, but those seem impractical for such a long distance. Let me go through each option one by one.\n\nFirst, flying. That's probably the fastest and most common way. I should check how long the flight takes. I think a direct flight from San Francisco International Airport (SFO) to New York's airports like JFK or Newark (EWR) would take around 5 to 6 hours. There are also other airports nearby, like Oakland or San Jose, but SFO is the main one. On the New York side, there's JFK, LaGuardia (LGA), and Newark. I should mention the airlines that operate these routes, like United, Delta","reasoning_content":null,"tool_calls":null},"logprobs":null,"finish_reason":"length","matched_stop":null}],"usage":{"prompt_tokens":14,"total_tokens":214,"completion_tokens":200,"prompt_tokens_details":null,"reasoning_tokens":0},"metadata":{"weight_version":"default"}}
```

- GSM8K

## Checklist


zhyncs:
when we build on cu124, does it also work?

rainj-me:
> when we build on cu124, does it also work?

I haven't tested with Cuda 12.4, but only tested with Cuda 12.6. Is there anyway I can download the wheel pkg from the CI?

rainj-me:
> when we build on cu124, does it also work?

I have tried the cuda 12.4 pkg installed on cuda 12.6 env, works fine. Same as cuda 12.9 pkg.

PR #10251, opened 2025-09-10
[1/4] Refactor LoRA to support backend-specific batch preprocessing.

lifuhuang:
## Motivation

This is the 1st PR extracted from: https://github.com/sgl-project/sglang/pull/10196

The main goal is to extract LoRA batch preparation logic from LoRAManager to LoRABackend, such that different backends could have their customized preprocessing logic, as required by the batched SGMV that we plan to introduce in a follow-up PR.

## Checklist


python/sglang/srt/lora/backend/base_backend.py
```diff
@@ -1,22 +1,22 @@
...
+from sglang.srt.model_executor.forward_batch_info import ForwardBatch
 
 
 class BaseLoRABackend:
     """Base class for different Lora backends.
        Each backend has its own implementation of Lora kernels.
 
     Args:
-        name: name of backend
         batch_info: information of current batch for use
```
Fridge003: We can change the explanation on Args to `max_lora_per_batch` and `device` instead

python/sglang/srt/lora/utils.py
```diff
@@ -10,19 +10,19 @@
...
 
-    # Indice pointers of each sequence in shape (bs + 1, )
+    # Indice pointers of each sequence in shape (num_segments + 1, )
     seg_indptr: torch.Tensor
 
-    # Maximum sequence length of current batch
-    max_len: int
-
-    # The index of lora adapter used by each sequence, in shape (bs,)
+    # The index of lora adapter used by each sequence, in shape (num_segments,)
```
Fridge003: Maybe we can change here to `...used by each segment`

python/sglang/srt/lora/utils.py
```diff
@@ -31,6 +31,15 @@ class LoRABatchInfo:
     # scaling of each lora adapter, in shape (lora_num,)
     scalings: torch.Tensor
 
+    # Lengths of each segments in shape (num_segments,)
+    seg_lens: Optional[torch.Tensor]
+
+    # Maximum sequence length of current batch
```
Fridge003: Is it "Maximum segment length"?

python/sglang/srt/lora/backend/triton_backend.py
```diff
@@ -8,12 +10,14 @@
     sgemm_lora_b_fwd,
 )
 from sglang.srt.lora.utils import LoRABatchInfo
+from sglang.srt.model_executor.forward_batch_info import ForwardBatch
 
 
 class TritonLoRABackend(BaseLoRABackend):
+    name = "triton"
```
Fridge003, line 17: Maybe we can remove this?

lifuhuang: it's used as a user friendly string in an error message in lora.py. But yeah, ideally we should just use this in server_args for consistency, maybe something for the future.

Fridge003 (approved):
LGTM

PR #10280, opened 2025-09-10
[router] add benchmark for regular router and pd router

key4ng:
## Motivation
pd router use 2 prefill workers and 2 decode workers
regular router use 2 workers, both worker dp=2

## Checklist


key4ng:
weird result: The pd router's performance is relatively worse than the regular router. both of them using 4 A10 
Screenshot 2025-09-11 at 10 17 58 AM

It could be as expected

> on a single A10 box, prefill/decode disaggregation usually makes TTFT worse. You’re paying an extra KV-cache handoff cost between GPUs (and often through host memory) that a “regular” router avoids. With only 2 prefill + 2 decode, batching also gets fragmented, so you lose throughput while adding transfer overhead

We can take it as a functional test, but not performance test

slin1237 (approved):

PR #10286, opened 2025-09-10
[2/4] Introduce Chunked-SGMV kernels and corresponding LoRA backend for improved performance

lifuhuang:
(Huge thanks to @Fridge003 for the extremely valuable discussions and insights)

## Motivation

The current LoRA solution has been suffering from disproportionate performance degradation compared with base model as batch size grows. For example, when benchmarking the current implementation, the median e2e latency with LoRA is appx 2x larger than the base model at request rate 4 and 3x larger at request rate 8.

During investigation, we found out that the current LoRA kernels and the previously deprecated flashinfer kernels are both dividing computations by batch size (e.g., grid dim = `[max_seq_len * cdiv(k_dim, BLOCK_K) , batch size]`), which suffers from the following problems:
1. **Prefill**: prefill batches with high variance in sequence length causes very high max_seq_len, therefore lots of no-op CTAs are launched, wasting GPU resources. 
2. **Decode**:  as all sequence in a batch have lengths 1, effectively 15/16 of each CTA's compute power is wasted (because 16 is the smallest M for tensor cores).

## Modifications

To overcome this issues, a new solution (Chunked SGMV) is introduced, based on Punica-style gather + chunking: 
- Step 1 - Gather: group all input tokens by LoRA weight to facilitate batching. This is identical to the SGMV solution in Punica.
- Step 2 - Chunking: splitting the tokens into constant-sized chunks (e.g., 16) at LoRA batch preprocessing phase to reduce fragmentation & launched CTA numbers when token distribution across groups are highly skewed.

With this approach, we observed significant perf improvement across the board.

## Benchmarking and Profiling

Benchmark command: 
```
    --model-path meta-llama/Llama-3.1-8B-Instruct \
    --disable-radix-cache \
    --lora-paths lora=algoprog/fact-generation-llama-3.1-8b-instruct-lora \
    --lora-backend csgmv \
    --port 30000
python3 -m sglang.bench_serving --backend sglang  --num-prompt 1000 --request-rate <2,8,16,inf>  --dataset-name sharegpt --lora-name lora 
```
<table>
<tr>
<td>image</td>
<td>image</td>
</tr>
<tr>
<td>image</td>
<td>image</td>
</tr>
</table>

### Median E2E Latency (ms)
| Request Rate | 2 | 8 | 16 | inf |
|--------------|------|------|-------|-------|
| Baseline | 1156 | 2424 | 8002 | 51212 |
| CSGMV | 1050 (**-9%**) | 1452 (**-40%**) | 2382 (**-70%**) | 9905 (**-81%**) |

---

### Median ITL (ms)
| Request Rate | 2 | 8 | 16 | inf |
|--------------|------|------|------|------|
| Baseline | 10.1 | 18.5 | 59.7 | 118.2 |
| CSGMV | 8.9 (**-12%**) | 9.73 (**-47%**) | 12.3 (**-79%**) | 36.4 (**-69%**) |

---

### Median TTFT (ms)
| Request Rate | 2 | 8 | 16 | inf |
|--------------|------|------|------|-------|
| Baseline | 47 | 55 | 98 | 15157 |
| CSGMV | 47 (**0%**) | 50 (**-9%**) | 75 (**-23%**) | 3128 (**-79%**) |

---

### Total Throughput (tokens/s)
| Request Rate | 2 | 8 | 16 | inf |
|--------------|------|------|------|------|
| Baseline | 942 | 3557 | 5286 | 4930 |
| CSGMV | 942 (**0%**) | 3665 (**+3%**) | 6582 (**+25%**) | 8005 (**+62%**) |

## Accuracy Tests

* testing with phi4mm shows normal output:
image

## Checklist


python/sglang/srt/lora/triton_ops/chunked_sgmv_expand.py
```diff
@@ -0,0 +1,212 @@
...
+    # and accumulate
+    pid_n = tl.program_id(axis=0)
+    n_offset = tl.arange(0, BLOCK_N) + pid_n * BLOCK_N + slice_start
+    k_offset = tl.arange(0, BLOCK_K)
+
+    x_ptrs = (
+        x
+        + slice_id * cur_rank * x_stride_1
+        + (s_offset_physical[:, None] * x_stride_0 + k_offset[None, :] * x_stride_1)
+    )
```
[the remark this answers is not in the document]
lifuhuang: It's a good point but I don't have a good way to address it -- the solution proposed by gemini likewise introduce data dependency on runtime variable `slice_id` and `x_stride_1`.

Considering this is consistent with the current triton backend, I will do more research and benchmark the perf difference carefully in a future PR.

Fridge003:
@lifuhuang Can you please post the commands for accuracy/benchmark tests?

python/sglang/srt/lora/triton_ops/chunked_sgmv_expand.py
```diff
@@ -0,0 +1,212 @@
...
+    slice_offsets,
+    # Meta parameters
+    NUM_SLICES: tl.constexpr,
+    MAX_RANK: tl.constexpr,  # K = R
+    BLOCK_S: tl.constexpr,
+    BLOCK_N: tl.constexpr,
+    BLOCK_K: tl.constexpr,
+):
+    """
+    This kernel performs chunked SGMV (Sparse General Matrix-Vector) for LoRA
```
ispobock: qq: SGMV is not `Segmented Gather Matrix-Vector multiplication`?

python/sglang/srt/lora/triton_ops/chunked_sgmv_expand.py
```diff
@@ -0,0 +1,212 @@
...
+
+    # Get dims
+    s = x.shape[0]
+    input_dim = x.shape[1]
+    max_lora_rank = lora_weight_b.shape[-1]
+    output_dim = lora_weight_b.shape[-2]
+    num_slices = len(slice_offsets) - 1
+    assert input_dim == num_slices * max_lora_rank
+
+    # TODO (lifuhuang): fine-tune per operation
```
ispobock, line 167: `num_stages` and `num_warps` can also be tuned.

lifuhuang:
> @lifuhuang Can you please post the commands for accuracy/benchmark tests?

Attached.

test/srt/lora/test_chunked_sgmv_backend.py
```diff
@@ -0,0 +1,740 @@
...
+        List[int],
+        List[str],
+    ]:
+        """Create test batch with specified composition and mode"""
+        seq_lengths = self.generate_sequence_lengths(
+            batch_size, batch_mode, 1, self.max_seq_len
+        )
+        if batch_composition == BatchComposition.UNIFORM:
+            lora_assignments = ["lora_A"] * batch_size
+        elif batch_composition == BatchComposition.MIXED:
```
Fridge003, line 436: Can we add `None` to the mixed batch, so we can cover the case when None and assigned_lora are run together

python/sglang/srt/lora/backend/base_backend.py
```diff
@@ -143,10 +143,10 @@ def get_backend_from_name(name: str) -> BaseLoRABackend:
         from sglang.srt.lora.backend.triton_backend import TritonLoRABackend
 
         return TritonLoRABackend
-    # elif name == "csgmv":
-    #     from sglang.srt.lora.backend.chunked_backend import ChunkedSgmvLoRABackend
+    elif name == "csgmv":
```
Fridge003, line 146: Can we set csgmv to the default lora backend? Since it's better than triton backend

python/sglang/srt/lora/backend/base_backend.py
```diff
@@ -143,10 +143,10 @@ def get_backend_from_name(name: str) -> BaseLoRABackend:
...
+        from sglang.srt.lora.backend.chunked_backend import ChunkedSgmvLoRABackend
```
Fridge003, line 147: We also need to update the lora_backend part in documents, including `lora.ipynb` and `server_arguments.md`.

This can be left for the next PR

python/sglang/srt/lora/backend/chunked_backend.py
```diff
@@ -0,0 +1,239 @@
...
+            seg_weight_indices, non_blocking=True
+        )
+        batch_info.seg_indptr[: num_segments + 1].copy_(seg_indptr, non_blocking=True)
+        batch_info.permutation[: len(permutation)].copy_(permutation, non_blocking=True)
+
+        self.batch_info = batch_info
+
+    @staticmethod
+    def _get_permutation(seq_weight_indices, forward_batch: ForwardBatch):
+        with torch.device("cpu"):
```
Fridge003, line 221: Can you add an example for this function, since it takes some time to understand its function from pure code.

python/sglang/srt/lora/backend/chunked_backend.py
```diff
@@ -0,0 +1,239 @@
...
+            permutation = torch.empty(
+                (len(row_weight_indices),), dtype=torch.long, pin_memory=True
+            )
+            torch.argsort(row_weight_indices, stable=True, out=permutation)
+            weights_reordered = row_weight_indices[permutation]
+
+            return permutation, weights_reordered
+
+    def _get_segments_info(self, weights_reordered: torch.Tensor):
+        with torch.device("cpu"):
```
Fridge003, line 278: The same, can you please add an example

python/sglang/srt/lora/triton_ops/chunked_sgmv_shrink.py
```diff
@@ -0,0 +1,176 @@
...
+
+    assert x.is_contiguous()
+    assert weights.is_contiguous()
+    assert len(x.shape) == 2
+    assert len(weights.shape) == 3
+
+    # Block shapes
+    BLOCK_S = 16
+    BLOCK_N = 16
+    BLOCK_K = 256
```
Fridge003, line 138: We can add a TODO for split-k

python/sglang/srt/lora/triton_ops/chunked_sgmv_shrink.py
```diff
@@ -0,0 +1,176 @@
...
+
+    S = x.shape[0]
+    N = weights.shape[1]
+    K = weights.shape[2]
+    assert x.shape[-1] == K
+
+    num_segments = batch_info.num_segments
+    grid = (
+        triton.cdiv(N, BLOCK_N),
+        batch_info.bs if batch_info.use_cuda_graph else num_segments,
```
Fridge003, line 148: I feel currently we are just asserting the forward batch is in decoding mode iff cuda_graph_enabled. 
But this is not true if we are going to support speculative decoding with lora in the future. Discoupling decode forward mode and cuda graph usage in lora might become one part of work in the future.

python/sglang/srt/lora/backend/chunked_backend.py
```diff
@@ -0,0 +1,239 @@
...
+from sglang.srt.lora.utils import LoRABatchInfo
+from sglang.srt.model_executor.forward_batch_info import ForwardBatch
+
+
+class ChunkedSgmvLoRABackend(BaseLoRABackend):
+    name = "csgmv"
+
+    def __init__(self, max_loras_per_batch: int, device: torch.device):
+        super().__init__(max_loras_per_batch, device)
+        self.segment_size = 16  # TODO (lifuhuang): make it configurable?
```
Fridge003, line 28: I think we need to make it configurable. We also need to do some experiments on how changing segment_size will affect performance

python/sglang/srt/lora/triton_ops/chunked_sgmv_expand.py
```diff
@@ -0,0 +1,211 @@
...
+    BLOCK_M = 16
+    BLOCK_K = 16
+    BLOCK_N = 64
```
Fridge003, line 170: Will a larger BLOCK_N help?

lifuhuang: Using the benchmark command I shared, larger BLOCK_N made result worse, but probably it has to do with the experiment setup (e.g., result might vary across different batch sizes, seq len, etc.)

I created a follow-up task: https://github.com/sgl-project/sglang/issues/10420 to further tune all meta parameters. Perhaps before that we should look into the launch overheads first, as it currently seems to impact perf more than the kernel itself.

python/sglang/srt/lora/backend/base_backend.py
lifuhuang: Let me do it in a separate PR -- I updated document and notebook for the new csgmv backend such that users can override themselves if they are interested in trying it out. I want to do more testing and update all our existing tests to use csgmv before making it default. WDYT?

Fridge003 (approved):
LGTM

python/sglang/srt/lora/backend/chunked_backend.py
lifuhuang: Good point. I did some experiment offline, will send a separate PR to support heuristically determine chunk size per batch and corresponding server args.

PR #10293, opened 2025-09-11
reduce vLLM's dependence on XPU

huaiyuzh:
reduce vLLM's dependence on XPU.


python/sglang/srt/distributed/parallel_state.py
```diff
@@ -1746,6 +1748,9 @@ def in_the_same_node_as(pg: ProcessGroup, source_rank: int = 0) -> List[bool]:
 
 
 def monkey_patch_vllm_parallel_state(reverse: bool = False):
+    if _is_xpu:
+        return
+
```
mingfeima, line 1867: what will changes in this file affect? tensor parallel for multi-gpu?

python/sglang/srt/layers/quantization/compressed_tensors/compressed_tensors.py
```diff
@@ -38,16 +38,20 @@
...
+from sglang.srt.utils import is_xpu
+
+_is_xpu = is_xpu()
+VLLM_AVAILABLE = False
+if not _is_xpu:
+    try:
+        from vllm.model_executor.layers.quantization.compressed_tensors.schemes.compressed_tensors_wNa16 import (
+            WNA16_SUPPORTED_BITS,
+            CompressedTensorsWNA16,
+        )
```
mingfeima: try to make minimal changes to framework.

we only change the parts that matters for our current stage targets. we don't really need compressed_tensors for intel gpus right now, do we?

python/sglang/srt/layers/quantization/compressed_tensors/compressed_tensors_moe.py
```diff
@@ -48,12 +49,15 @@
...
+VLLM_AVAILABLE = False
+if not _is_xpu:
+    try:
+        import vllm
 
-    VLLM_AVAILABLE = True
-except ImportError:
-    VLLM_AVAILABLE = False
+        VLLM_AVAILABLE = True
+    except ImportError:
```
mingfeima: same as above.

python/sglang/srt/layers/quantization/compressed_tensors/schemes/compressed_tensors_w8a16_fp8.py
```diff
@@ -15,22 +15,32 @@
     CompressedTensorsScheme,
 )
 from sglang.srt.layers.quantization.utils import convert_to_channelwise
+from sglang.srt.utils import is_xpu
+
+_is_xpu = is_xpu()
```
mingfeima: same as before

python/sglang/srt/layers/quantization/w4afp8.py
```diff
@@ -17,10 +17,11 @@
 from sglang.srt.layers.quantization.fp8 import Fp8LinearMethod
 from sglang.srt.layers.quantization.unquant import UnquantizedLinearMethod
 from sglang.srt.layers.quantization.utils import is_layer_skipped
-from sglang.srt.utils import is_npu, set_weight_attrs
+from sglang.srt.utils import is_npu, is_xpu, set_weight_attrs
```
mingfeima: `w4afp8` is our prioritized quantization config. You may align internally with Hengyu.

python/sglang/srt/layers/rotary_embedding.py
```diff
@@ -102,8 +104,10 @@ def __init__(
             cache = cache.to(dtype)
 
         if (
-            not (_is_cuda or _is_npu) or self.head_size not in [64, 128, 256, 512]
-        ) and not (_is_cpu and _is_cpu_amx_available):
+            (not (_is_cuda or _is_npu) or self.head_size not in [64, 128, 256, 512])
+            and not (_is_cpu and _is_cpu_amx_available)
+            and not _is_xpu
```
mingfeima: this is getting too complexed, can we simplify this logic, make it easier for maintainance.

mingfeima (changes requested):
remove parts that does not matter to intel gpu at our current stages.

we don't have to match all the features to CUDA gpus at the moment, just focus on the minimal requirements.

mingfeima:
> reduce vLLM's dependence on XPU.

need to be more specific on what kind of changes this PR applies.

and also the results, say:
before - you run xxx will end up with runtime error
after - you run xxx will NOT end up with runtime error

and a more general concept - why we want to decouple vllm? for UX purposes and etc.

python/sglang/srt/distributed/parallel_state.py
huaiyuzh: yes, it is for distributed GPU. We said before that XPU does not need to support monkey_patch_vllm_parallel_state.

mingfeima:
close this one as this has been covered in other commits.

PR #8118, opened 2025-07-17
[feat] Support tp mode for DeepSeek-R1-W4AFP8

chenxijun1029:
## Motivation

Support tp mode for DeepSeek w4a8 model, which has a better performace than ep mode.

## Modifications

1. Add W4AFp8MoEMethod and associated `create_weights`, `process_weights_after_loading` function and `apply` function. In the apply function, we use the same cutlass_w4a8_moe kernel as ep moe uses.
2. Add some tile shape and cluster shape config for tp moe in `cutlass_w4a8_moe` kernel.
3. Add a router logic in w4afp8 quant config and method. When "enable_ep_moe" found in global_server_args_dict, we use ep mode, else tp.

Co-author: @yuhyao <827623970@qq.com>

## Benchmark
We run DeepSeek-R1-W4AFP8 on 8*H20 with tp8, comparing to run DeepSeek-R1 on 8*H20 with ep8.
Test configuration: 

### ISL1000, OSL1000
input/output len = 1000/1000, qps=128, max_concurrency=128, num_prompt=256.
The results are shown below:

TP
```
============ Serving Benchmark Result ============
Backend:                                 sglang    
Traffic request rate:                    128.0     
Max request concurrency:                 128       
Successful requests:                     256       
Benchmark duration (s):                  159.00    
Total input tokens:                      256000    
Total generated tokens:                  256000    
Total generated tokens (retokenized):    254696    
Request throughput (req/s):              1.61      
Input token throughput (tok/s):          1610.09   
Output token throughput (tok/s):         1610.09   
Total token throughput (tok/s):          3220.18   
Concurrency:                             127.52    
----------------End-to-End Latency----------------
Mean E2E Latency (ms):                   79201.47  
Median E2E Latency (ms):                 78956.21  
---------------Time to First Token----------------
Mean TTFT (ms):                          6547.98   
Median TTFT (ms):                        6612.11   
P99 TTFT (ms):                           11687.82  
---------------Inter-Token Latency----------------
Mean ITL (ms):                           72.73     
Median ITL (ms):                         68.05     
P95 ITL (ms):                            72.42     
P99 ITL (ms):                            73.05     
Max ITL (ms):                            11148.65  
==================================================
```

While EP:
```
============ Serving Benchmark Result ============
Backend:                                 sglang    
Traffic request rate:                    128.0     
Max request concurrency:                 128       
Successful requests:                     256       
Benchmark duration (s):                  161.37    
Total input tokens:                      256000    
Total generated tokens:                  256000    
[...]

PR #10299, opened 2025-09-11
[fix CI] Fix logical condition in fused MoE layer for compressed tensor quantization

BBuf:
### Background

The bug is caused by https://github.com/sgl-project/sglang/pull/8118 cc @chenxijun1029 

The current code in `fused_moe_triton/layer.py` has a logical operator precedence issue in the condition check for input scales validation (lines 615-620). The problematic condition was:

```python
if (
    "compressed" in self.quant_method.__class__.__name__.lower()
    or "w4afp8" in self.quant_config.get_name()
    and (param.data[expert_id] != 1).any()
    and ((param.data[expert_id] - loaded_weight).abs() > 1e-5).any()
):
```

Due to operator precedence (`and` has higher precedence than `or`), this condition is actually parsed as:

```python
if (
    "compressed" in self.quant_method.__class__.__name__.lower()
    or (
        "w4afp8" in self.quant_config.get_name()
        and (param.data[expert_id] != 1).any()
        and ((param.data[expert_id] - loaded_weight).abs() > 1e-5).any()
    )
):
```

This means that **any** model using compressed tensors quantization will unconditionally trigger the ValueError, regardless of whether the input scales are actually equal or not.

### Error Observed

When loading models with compressed tensors quantization (e.g., `neuralmagic/Mixtral-8x7B-Instruct-v0.1-FP8`), the following error occurs:

```
ValueError: input_scales of w1 and w3 of a layer must be equal. But got 1.0 vs. tensor([0.1011], device='cuda:1', dtype=torch.bfloat16)
```

### Solution

Add parentheses to ensure correct operator precedence:

```python
if (
    ("compressed" in self.quant_method.__class__.__name__.lower()
    or "w4afp8" in self.quant_config.get_name())
    and (param.data[expert_id] != 1).any()
    and ((param.data[expert_id] - loaded_weight).abs() > 1e-5).any()
):
```

Now the condition correctly checks:
- The quantization method is either `compressed` OR `w4afp8` 
- AND the parameter data is not equal to 1
- AND the difference between existing and loaded weights exceeds the threshold

This ensures the validation only runs when appropriate and doesn't incorrectly fail for valid compressed tensor models.

### Testing

This fix resolves the model loading failure for compressed tensor quantized models like `neuralmagic/Mixtral-8x7B-Instruct-v0.1-FP8` in the nightly GSM8K evaluation tests.

### DIFF
--- a/python/sglang/srt/layers/moe/fused_moe_triton/layer.py
+++ b/python/sglang/srt/layers/moe/fused_moe_triton/layer.py
@@ -613,8 +613,8 @@ class FusedMoE(torch.nn.Module):
             loaded_weight = loaded_weight.to(param.data.device)
 
             if (
-                "compressed" in self.quant_method.__class__.__name__.lower()
-                or "w4afp8" in self.quant_config.get_name()
+                ("compressed" in self.quant_method.__class__.__name__.lower()
+                or "w4afp8" in self.quant_config.get_name())
                 and (param.data[expert_id] != 1).any()
                 and ((param.data[expert_id] - loaded_weight).abs() > 1e-5).any()
             ):
--- a/python/sglang/srt/layers/moe/fused_moe_triton/layer.py
+++ b/python/sglang/srt/layers/moe/fused_moe_triton/layer.py
@@ -613,8 +613,10 @@ class FusedMoE(torch.nn.Module):
             loaded_weight = loaded_weight.to(param.data.device)
 
             if (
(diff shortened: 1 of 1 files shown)


merrymercy (approved):

zhyncs:
https://github.com/sgl-project/sglang/actions/runs/17636575709/job/50113894911 @BBuf

PR #10300, opened 2025-09-11
refactor zero copy

pansicheng:
## Checklist


xiezhq-hermann:
I think the new design makes sense:
1. pass host memory pool to storage backend so it would be easier to implement logics around in the backend instead of cache controller
2. at the same time the host memory pool can still provide public methods such as get pointers
@huangtingwei9988 can you take a look and coordinate change on Mooncake.
Hey @mkhazraee, @vvenkates27, can you take a look as well see does it help on zero-copy implementation on NIXL?

python/sglang/srt/mem_cache/storage/mooncake_store/mooncake_store.py
```diff
@@ -189,6 +201,145 @@ def register_buffer(self, buffer: torch.Tensor) -> None:
             logger.error("Failed to register buffer to Mooncake Store: %s", err)
             raise TypeError("Mooncake Store Register Buffer Error.") from err
 
+    def _get_mha_buffer_meta(self, keys, indices):
```
xiezhq-hermann, line 207: can this reuse logics in `get_page_buffer_list`? If that does not suits the need here we should make the public methods better so the backends does not have to reimplement the same things over and over again

python/sglang/srt/mem_cache/storage/mooncake_store/mooncake_store.py
```diff
@@ -189,6 +201,145 @@ def register_buffer(self, buffer: torch.Tensor) -> None:
...
+            return [
+                (
+                    (k_res == 0 and v_res == 0)
+                    if is_set_operate
+                    else (k_res > 0 and v_res > 0)
+                )
+                for k_res, v_res in kv_pairs
+            ]
+
+    def batch_get_v1(
```
xiezhq-hermann, line 253: why do we need two versions of the batch get and set.

python/sglang/srt/managers/cache_controller.py
```diff
@@ -630,41 +622,16 @@ def append_host_mem_release(self, host_indices: torch.Tensor):
...
-            target_sizes=buffer_sizes,
-        )
-        if get_result != len(hash_values):
-            logger.warning(
-                f"Prefetch operation {operation.request_id} failed or partially failed."
-            )
-        if get_result != 0:
-            operation.increment(get_result * self.page_size)
+    def _generic_page_get_v1(self, operation, hash_values, host_indices):
+        results = self.storage_backend.batch_get_v1(hash_values, host_indices)
```
xiezhq-hermann, line 626: It's probably better to have a single generic page set function and backends can choose their batch size if needed instead if having two implementations.

python/sglang/srt/mem_cache/storage/mooncake_store/mooncake_store.py
huangtingwei9988: > can this reuse logics in `get_page_buffer_list`? If that does not suits the need here we should make the public methods better so the backends does not have to reimplement the same things over and over again

@xiezhq-hermann  The zero copy of mooncake store and 3fs is quite different. Even if I reuse `get_page_buffer_list`, I still need to reorganize the input of mooncake store's zero copy. At the same time, `_get_mha_buffer_meta` has better performance for mooncake.

python/sglang/srt/mem_cache/storage/mooncake_store/mooncake_store.py
huangtingwei9988: > why do we need two versions of the batch get and set.

cc @pansicheng

pansicheng: @xiezhq-hermann We currently have two versions of batch_[get/set] and _generic_page_[get/set] because the main logic and interface parameters have been modified. However, only the 3fs and mooncake backends have implemented this new path. To ensure compatibility and prevent the nixl and file backends from breaking, we are temporarily maintaining both interfaces. Once all backends have adopted the new implementation, we can deprecate the old one and clean up the code to keep only a single version.

python/sglang/srt/mem_cache/storage/mooncake_store/mooncake_store.py
```diff
@@ -177,8 +182,15 @@ def warmup(self):
         assert self.store.put(warmup_key, warmup_value) == 0
         assert self.store.is_exist(warmup_key) == 1
         assert self.store.get(warmup_key) == warmup_value
-
-    def register_buffer(self, buffer: torch.Tensor) -> None:
+        # delete warmup key
+        self.store.remove(warmup_key)
```
ykwd: Once a key is obtained, it has a period of lease (5 sec in default), during which the removal of this key will fail. On the other hand, this key will automatically be evicted by the LRU algorithm in the future. So no need to remove it here. @huangtingwei9988

python/sglang/srt/managers/cache_controller.py
```diff
@@ -630,41 +622,16 @@ def append_host_mem_release(self, host_indices: torch.Tensor):
...
+        for i in range(len(hash_values)):
+            if not results[i]:
+                logger.warning(
+                    f"Prefetch operation {operation.request_id} failed to retrieve page {hash_values[i]}."
+                )
+                break
+            if not operation.increment(self.page_size):
```
ykwd: We could just count how many pages succeed, then increment succeed_num * self.page_size in the end. Each time the increment is called, it will lock a mutex, which is expensive.

python/sglang/srt/managers/cache_controller.py
ykwd: The batch size issue is already handled within the `_page_transfer` and `_page_backup` functions. The main difference between `generic_xxx_v1` and `generic_xxx` lies in whether data copy operations are involved.

python/sglang/srt/mem_cache/storage/mooncake_store/mooncake_store.py
huangtingwei9988: > Once a key is obtained, it has a period of lease (5 sec in default), during which the removal of this key will fail. On the other hand, this key will automatically be evicted by the LRU algorithm in the future. So no need to remove it here. @huangtingwei9988

@ykwd ok, thanks!

python/sglang/srt/managers/cache_controller.py
pansicheng: Fixed! PTAL

xiezhq-hermann (approved):

Issue #9163, opened 2025-08-13
[Feature] Support Cache Salting for secure prefix caching


pyc96:
### Checklist

- [x] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 2. Please use English, otherwise it will be closed.

### Motivation

Similar to https://github.com/vllm-project/vllm/issues/16016, to mitigate issues like [Leaking Secrets from Prefix Caches](https://arxiv.org/html/2411.18191v1), it will be great if SGLang can support similar single-barrier one salt per request pattern. So that for multi tenant deployment we can easily isolate the cache between different users.

PR #10317, opened 2025-09-11
Refactors radix cache for extra key support

JustinTong0323:
## Motivation
In specific scenarios, it is necessary to isolate the kvcache for certain requests. For instance, requests utilizing different LoRA adapters, or requests for which we do not desire sharing the kvcache with others(ref https://github.com/sgl-project/sglang/issues/9163).

This PR refactors the Key in the Radix cache by adding an `extra_key` field. Only requests/tokens in the cache that have the same `extra_key` can share the prefix KV cache.

I would implement the logic for cache salting in following PR.

## Modifications

This commit refactors the radix cache to support an extra key (e.g., lora_id, cache_salt) for classifying requests.

The changes include:

- Adds `extra_key` to the `Req` class 
- Replace raw token_ids key to `BaseKey` class  to facilitate request classification.
- Modifies the `match_prefix` method in `RadixCache` and `SchedulePolicy` to incorporate the `BaseKey` when matching prefixes.
- **Removes the LoRARadixCache**, as the base RadixCache can now handle LoRA IDs through the `extra_key` mechanism.

## TODO List:
- [x] Add support for HiRadixCache
- [x] Support SWA radix cache
- [x] Compatible for cpp/lmc
- [ ] Support radix cache cpp for extra keys (need modify cpp codes, in following PR
- [ ] others I may not considered

## Checklist


test/srt/test_swa_unittest.py
```diff
@@ -4,7 +4,8 @@
 
 from sglang.srt.mem_cache.allocator import SWAKVPool, SWATokenToKVPoolAllocator
 from sglang.srt.mem_cache.memory_pool import ReqToTokenPool
-from sglang.srt.mem_cache.radix_cache import SWARadixCache
+from sglang.srt.mem_cache.radix_cache import BaseKey
```
ispobock: Do we have unit test for radix cache? Maybe we can create one.

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -40,14 +40,38 @@
     from sglang.srt.managers.schedule_batch import Req
 
 
+class BaseKey:
```
ispobock: qq: Why call it base key? How about putting it to base_prefix_cache?

test/srt/test_swa_unittest.py
JustinTong0323: We have `sglang/test/srt/test_radix_attention.py`

python/sglang/srt/mem_cache/radix_cache.py
JustinTong0323: I think there could be additional Key schemas inheriting from this base class in the future. These schemas could contain more information, like for hicache, we could attach data such as location (I'm not familiar with hicache, this is just a random thought).

JustinTong0323: Also I noticed BaseKey is not used in chunked cache, so I suppose it should placed in radix cache.

test/srt/test_swa_unittest.py
ispobock: I see. It seems an e2e test, not unit test. Unit test will test the basic functions like insert/match/eviction.

python/sglang/srt/mem_cache/radix_cache.py
ispobock: @xiezhq-hermann what do you think?

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -192,12 +236,12 @@ def match_prefix(self, key: List[int], **kwargs) -> MatchResult:
...
         )
 
-    def insert(self, key: List, value=None, chunked=False):
+    def insert(self, key: BaseKey, value=None, chunked=False):
         if self.disable:
             return 0
 
         if value is None:
-            value = [x for x in key]
+            value = torch.tensor([x for x in key.token_ids], dtype=torch.int64)
```
JustinTong0323: Note:  this line is to fix a minor error, if value is `list`, an error would be raised when execute `torch.cat(value)`

test/srt/test_swa_unittest.py
JustinTong0323: Added.

python/sglang/srt/mem_cache/radix_cache.py
JustinTong0323: but it seems not used... not sure if we could directly delete this `if value is None:`

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -137,29 +181,29 @@ def __init__(
...
         self.protected_size_ = 0
         self._record_all_cleared_event()
 
-    def match_prefix(self, key: List[int], **kwargs) -> MatchResult:
+    def match_prefix(self, key: BaseKey, **kwargs) -> MatchResult:
         """Find the matching prefix from the radix tree.
         Args:
-            key: A list of token IDs to find a matching prefix.
         Returns:
+            key: A BaseKey contains a list of token IDs to find a matching prefix.
```
Fridge003: Here we need some explanation on extra_key

Fridge003:
@JustinTong0323 Please fix the conflict

Fridge003:
Seems really clean and tidy to me.
I feel the key of this PR is to make sure it doesn't break all kinds of features (hi-cache, swa-radix...) So please make sure this PR isn't merged until all the related tests can pass

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -203,12 +274,12 @@ def match_prefix(self, key: List[int], **kwargs) -> MatchResult:
...
         )
 
-    def insert(self, key: List, value=None, chunked=False):
+    def insert(self, key: BaseKey, value=None, chunked=False):
         if self.disable:
             return 0
 
         if value is None:
-            value = [x for x in key]
+            value = torch.tensor(key.token_ids, dtype=torch.int64)
```
JustinTong0323, line 282: Note: this line is to fix a minor error, if value is list, an error would be raised when execute` torch.cat(value) `
But it seems not used... not sure if we could directly delete this `if value is None:`

Fridge003 (approved):
LGTM

python/sglang/srt/mem_cache/radix_cache.py
xiezhq-hermann: I think called it as radix_key might be more straightforward, it is specific to radix trees and can contain different information, including the token ids and the extra information/

xiezhq-hermann (approved):
Looks good to me, would suggest to rename `BaseKey` to `RadixKey` though

python/sglang/srt/mem_cache/radix_cache.py
hnyls2002: @JustinTong0323 Do not call this class `BaseKey`. Name it `RadixKey` or others.

python/sglang/srt/mem_cache/radix_cache.py
hnyls2002: This value will not be None unless in the `__main__` function in this file. Only used for testing, I think.

hnyls2002 (approved):

PR #10334, opened 2025-09-11
Refactor(prefix_cache): refactor init prefix cache, server args

jimmy-evo:
## Motivation

there are too many args and prefix cache implementation.

some of them are confilict.

too many if-else for `disable_radix_cache` and `enable_hicache`
and check between `enable_lmcache` and `disable_radix_cache` is forgot

use `prefix_cache_backend` and discard `disable_radix_cache` and `enable_hicache`

## Modifications

introduce new arg: --prefix-cache-backend

## Checklist

### DIFF
--- a/python/sglang/srt/managers/scheduler.py
+++ b/python/sglang/srt/managers/scheduler.py
@@ -135,6 +135,10 @@ from sglang.srt.managers.session_controller import Session
 from sglang.srt.managers.tp_worker import TpModelWorker
 from sglang.srt.managers.tp_worker_overlap_thread import TpModelWorkerClient
 from sglang.srt.managers.utils import DPBalanceMeta, validate_input_length
+from sglang.srt.mem_cache.base_prefix_cache import (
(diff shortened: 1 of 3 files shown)


jinmingyi1998:
@xiezhq-hermann @hnyls2002 
My goal is to optimize server configuration arguments.

Initially, simply disabling or enabling certain features and then deriving the final behavior required users to have a deep understanding of the code logic. This is especially challenging now that there are so many caching methods available.

for example
```
    prefix_cache_backend: Literal[
        "radix",
        "radix_cpp",
        "chunk",
        "hierarchical",
        "hierarchical_cpp",
        "lora",
        "lmcache",
        "swa_radix",
        "swa_chunk",
        "none",
    ] = "radix"
```

hnyls2002:
@jinmingyi1998 No, we prefer using simply combinations of `--enable-xxx` instead of `--prefix-cache-bakend`. Some users would not know which exact cache they use.

PR #10371, opened 2025-09-12
[NVIDIA] Adding Specific Family Blackwell

johnnynunez:
- Article: https://developer.nvidia.com/blog/nvidia-blackwell-and-nvidia-cuda-12-9-introduce-family-specific-architecture-features/
- This make compatible with < 12.9 CUDA and add Family Specific for >=CUDA 12.9
- Added 11.0f (Jetson Thor, new era of Robotics)
- With family specific make compatbile with GB300, Spark and Thor, reducing binaries size.

cc @zhyncs


johnnynunez:
cc @pavanimajety @Aidyn-A

johnnynunez:
@hlu1 @zhyncs @Aidyn-A

sgl-kernel/CMakeLists.txt
```diff
@@ -216,22 +216,28 @@ endif()
...
-if ("${CUDA_VERSION}" VERSION_GREATER_EQUAL "12.8" OR SGL_KERNEL_ENABLE_SM100A)
-    list(APPEND SGL_KERNEL_CUDA_FLAGS
-        "-gencode=arch=compute_100a,code=sm_100a"
-        "-gencode=arch=compute_120a,code=sm_120a"
-    )
+if ("${CUDA_VERSION}" VERSION_GREATER_EQUAL "12.8" OR SGL_KERNEL_ENABLE_SM100F)
+    if ("${CUDA_VERSION}" VERSION_GREATER_EQUAL "12.9")
+        list(APPEND SGL_KERNEL_CUDA_FLAGS
+            "-gencode=arch=compute_100f,code=sm_100"
+            "-gencode=arch=compute_120f,code=sm_120"
```
Aidyn-A: Shouldn't it be like this?
```suggestion
            "-gencode=arch=compute_100f,code=sm_100f"
            "-gencode=arch=compute_120f,code=sm_120f"
```

sgl-kernel/CMakeLists.txt
```diff
@@ -216,22 +216,28 @@ endif()
...
+        )
+    endif()
 
     # refer sm_121, sm_110 and sm_101 description  https://github.com/pytorch/pytorch/pull/156176
     if ("${CUDA_VERSION}" VERSION_GREATER_EQUAL "13.0")
         list(APPEND SGL_KERNEL_CUDA_FLAGS
-            "-gencode=arch=compute_103a,code=sm_103a"
-            "-gencode=arch=compute_110a,code=sm_110a"
-            "-gencode=arch=compute_121a,code=sm_121a"
+            "-gencode=arch=compute_110f,code=sm_110"
```
Aidyn-A: Same here:
```suggestion
            "-gencode=arch=compute_110f,code=sm_110f"
```

Aidyn-A (changes requested):

sgl-kernel/CMakeLists.txt
```diff
@@ -216,22 +216,28 @@ endif()
...
-if ("${CUDA_VERSION}" VERSION_GREATER_EQUAL "12.8" OR SGL_KERNEL_ENABLE_SM100A)
-    list(APPEND SGL_KERNEL_CUDA_FLAGS
-        "-gencode=arch=compute_100a,code=sm_100a"
-        "-gencode=arch=compute_120a,code=sm_120a"
-    )
+if ("${CUDA_VERSION}" VERSION_GREATER_EQUAL "12.8" OR SGL_KERNEL_ENABLE_SM100F)
+    if ("${CUDA_VERSION}" VERSION_GREATER_EQUAL "12.9")
+        list(APPEND SGL_KERNEL_CUDA_FLAGS
+            "-gencode=arch=compute_100f,code=sm_100f"
+            "-gencode=arch=compute_120f,code=sm_120f"
```
hlu1, line 228: Is this the right choice? This would prevent us from using 100a or 103a features in sgl-kernel. 120f should be ok because it's likely to add 120a kernel to sgl-kernel.

johnnynunez: @hlu1 @Aidyn-A (correct me if i am wrong). Blackwell family f activate the features for all minors.
It is equivalent but more versatile

- a = architecture-specific (strict, no forward compatibility). Targets a single, exact architecture. Code built with this is not forward-compatible with other architectures.

- f = family-specific (safer portability within a major compute capability). Enables features shared across the whole GPU family, so code remains portable across different variants of that major compute capability

so, in summary:
f (family-specific) features are guaranteed across the family (same major CC) and are often Tensor Core–related; code remains portable within 10.x/12.x, but not across different major families.

hlu1: 100f refers to the common features that are shared by all current and future sm_10x GPUs, meaning that to use 103a features, we still need to compile with sm_103a.

johnnynunez: Depends that you want. 
f activate only the new fp4 format(3xfp4)
while a activates fp4 and 3xfp4. Depends that you want. If not, close this PR.

PR #10374, opened 2025-09-12
decoding batch cuda graph disable warning

BBuf:
## Motivation

图片

## Checklist

### DIFF
--- a/python/sglang/srt/managers/cuda_graph_warner.py
+++ b/python/sglang/srt/managers/cuda_graph_warner.py
@@ -22,8 +22,10 @@ class NoCudaGraphWarner:
     def __init__(self):
         self.consecutive_no_cuda_graph_count = 0
 
-    def __call__(self, can_run_cuda_graph: bool):
-        if not can_run_cuda_graph:
+    def on_step(self, can_run_cuda_graph: bool):
+        if can_run_cuda_graph:
+            self.consecutive_no_cuda_graph_count = 0
+        else:
             self.consecutive_no_cuda_graph_count += 1
             if self.consecutive_no_cuda_graph_count % 5 == 0:
                 logger.warning(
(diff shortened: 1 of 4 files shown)


python/sglang/srt/managers/scheduler_metrics_mixin.py
```diff
@@ -219,6 +220,17 @@ def log_decode_stats(
             f"#queue-req: {len(self.waiting_queue)}, "
         )
 
+        if not can_run_cuda_graph:
```
fzyzcjy: nit: would be great to make this a separate class, e.g.

```
class NoCudaGraphWarner:
  def __init__():
     self.consecutive_no_cuda_graph_count=0
  def __call__():
     the main logic
```

to avoid making the already huge Scheduler even bigger

python/sglang/srt/managers/cuda_graph_warner.py
```diff
@@ -0,0 +1,35 @@
...
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class NoCudaGraphWarner:
+    def __init__(self):
+        self.consecutive_no_cuda_graph_count = 0
+
+    def __call__(self, can_run_cuda_graph: bool):
```
fzyzcjy: nit: maybe call it on_step etc (I originally thought it can be named `__call__` but now seeing code it seems that is not super clear)

python/sglang/srt/managers/cuda_graph_warner.py
```diff
@@ -0,0 +1,35 @@
...
+        if not can_run_cuda_graph:
```
fzyzcjy: optional nit: maybe put the positive branch (if can_run_cuda_graph) first

fzyzcjy (approved):
only two tiny optional nits and LGTM

python/sglang/srt/managers/cuda_graph_warner.py
BBuf: done.

python/sglang/srt/managers/cuda_graph_warner.py
BBuf: done.

BBuf:
图片

hebiao064 (approved):

PR #10382, opened 2025-09-12
Benchmark: Improve path of check_chat_template

Muqi1029:
## Motivation

When using `base_url` to request remote servers, users typically provide a **tokenizer path** to handle prompt tokenization and ensure prompt length. However, the current `check_chat_template` checks the **model path** from the HTTP request, which may not match the official model identifier (e.g., just `"DeepSeek-V3"`), leading to lookup errors.

Since the `chat_template` is usually defined in the **tokenizer path**, it makes more sense to prioritize checking the tokenizer path provided by the user. This way, users don’t need to explicitly pass model arguments, making the experience simpler and more intuitive.

## Modifications
1. Move `tokenizer_id` earlier in the workflow so it can be used by `check_chat_template`.
2. Rename the argument in `check_chat_template` for clarity and consistency.

## Checklist

### DIFF
--- a/python/sglang/bench_serving.py
+++ b/python/sglang/bench_serving.py
@@ -1930,9 +1930,9 @@ async def benchmark(
     return result | result_details
 
 
-def check_chat_template(model_path):
+def check_chat_template(tokenizer_id):
     try:
-        tokenizer = AutoTokenizer.from_pretrained(model_path, trust_remote_code=True)
+        tokenizer = AutoTokenizer.from_pretrained(tokenizer_id, trust_remote_code=True)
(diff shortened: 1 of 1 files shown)


python/sglang/bench_serving.py
```diff
@@ -1930,9 +1930,9 @@ async def limited_request_func(request_func_input, pbar):
     return result | result_details
 
 
-def check_chat_template(model_path):
+def check_chat_template(tokenizer_id):
```
hnyls2002:
```suggestion
def check_chat_template(tokenizer_path):
```

hnyls2002 (approved):

PR #10385, opened 2025-09-12
[Router]fix: fix get_load missing api_key

jimmy-evo:
## Motivation

fix `/get_load` 401 Error

`get_load` need api_key

worker metadata should has api_key

each worker may have different api_key in future

## Modifications

add api_key to worker metadata

add api_key in header for `/get_load`

## Checklist


hnyls2002:
@jinmingyi1998 Please resolve the conflicts. and @slin1237 Could you review on this?

sgl-router/src/server.rs
```diff
@@ -281,15 +281,16 @@ async fn v1_responses_list_input_items(
...
-struct UrlQuery {
+struct AddWorkerQuery {
     url: String,
+    api_key: Option<String>,
 }
 
 async fn add_worker(
     State(state): State<Arc<AppState>>,
-    Query(UrlQuery { url }): Query<UrlQuery>,
+    Query(AddWorkerQuery { url, api_key }): Query<AddWorkerQuery>,
```
slin1237, line 292: this would be breaking change for existing users
should we really do that?

slin1237: if the problem is really just because the worker has apikey setup
we should probably update
```
async fn create_worker(
    State(state): State<Arc<AppState>>,
    Json(config): Json<WorkerConfigRequest>,
) -> Response {
```
to include apikey as part of `workerConfigRequest`

sgl-router/src/core/worker.rs
slin1237, line 1: we should probably change this to builder pattern at some point
adding a single field required 440 lines of change is insane

sgl-router/src/routers/grpc/pd_router.rs
```diff
@@ -346,7 +348,11 @@ impl RouterTrait for GrpcPDRouter {
 
 #[async_trait]
 impl WorkerManagement for GrpcPDRouter {
-    async fn add_worker(&self, _worker_url: &str) -> Result<String, String> {
```
slin1237, line 356: we plan to deprecate this
lets not add them here
we already have api key as part of router input
lets do that?

sgl-router/src/core/worker.rs
```diff
@@ -22,7 +22,8 @@ static WORKER_CLIENT: LazyLock<reqwest::Client> = LazyLock::new(|| {
 pub trait Worker: Send + Sync + fmt::Debug {
     /// Get the worker's URL
     fn url(&self) -> &str;
-
+    /// Get the worker's API key
+    fn api_key(&self) -> &Option<String>;
```
slin1237, line 28: on a second thought
perhaps lets not do api key here?
router already has api key as part of the startup command
we can just pass that to overall pd router

slin1237 (approved):

PR #10389, opened 2025-09-12
fix(disagg): IPv6-ready bootstrap host resolution

dmitrygx:
## Motivation

The disaggregation bootstrap process currently assumes IPv4-style addressing, which breaks in environments where only IPv6 is available or when hostnames resolve to IPv6.  
This PR adds **IPv6 support** to the bootstrap registration logic so that multi-node setups can reliably connect in dual-stack or IPv6-only networks.  

## Modifications

- Updated `_register_to_bootstrap` to correctly handle:
  - Explicit IPv6 literals with or without brackets (e.g., `[2001:db8::1]`, `[2001:db8::1]:30001`).  
  - Hostnames resolving to IPv4 or IPv6, using `socket.getaddrinfo` with `AF_UNSPEC` to let the OS select the appropriate address family.  
  - Automatic wrapping of IPv6 literals in brackets when constructing URLs (`{host}:{port}` format).  
  - Graceful fallback to the raw host/FQDN if resolution fails.  
- Preserved compatibility with existing IPv4 and FQDN-based configurations.  

## Accuracy Tests

N/A – this change only affects **bootstrap registration and networking logic**, not model execution or inference accuracy.  

## Benchmarking and Profiling

N/A – bootstrap is not performance-critical compared to inference execution.  
Manual testing confirmed no measurable overhead when resolving hosts via `socket.getaddrinfo`.  

## Checklist

- [x] Format code with pre-commit.  
- [x] Add unit tests:
  - IPv4 literal  
  - IPv6 literal (with/without brackets)  
  - FQDN resolving to IPv4/IPv6  
  - Invalid host fallback case  
- [x] Update documentation with IPv6 bootstrap examples.  
- [ ] Provide accuracy/benchmark results – not applicable to this PR.

### DIFF
--- a/python/sglang/srt/disaggregation/common/conn.py
+++ b/python/sglang/srt/disaggregation/common/conn.py
@@ -80,7 +80,23 @@ class CommonKVManager(BaseKVManager):
                 else:
                     host, _ = self.dist_init_addr.rsplit(":", 1)
             else:
-                host = socket.gethostbyname(self.dist_init_addr.rsplit(":", 1)[0])
+                # IPv6-ready: resolve FQDN/IPv4/IPv6 via AF_UNSPEC and
+                # bracket-wrap IPv6 for safe "{host}:{port}" URL building.
+                _raw_host = self.dist_init_addr.rsplit(":", 1)[0]
+                try:
+                    infos = socket.getaddrinfo(
+                        _raw_host,
+                        None,
+                        family=socket.AF_UNSPEC,
+                        type=socket.SOCK_STREAM,
+                    )
+                    host = infos[0][4][0]  # let OS policy pick v4/v6
+                except socket.gaierror:
+                    # Fallback to the literal/FQDN if resolution fails now.
+                    host = _raw_host
+                # Wrap IPv6 literal with brackets so f"{host}:{port}" is valid.
+                if ":" in host and not host.startswith("["):
+                    host = f"[{host}]"
         else:
             # single node: bootstrap server's host is same as http server's host
             host = self.bootstrap_host


ShangmingCai (commented):
Isn't `--dist-init-addr` something that the user passes in through server args? Maybe it can be solved by users modifying the command. Not sure if it's necessary to add these fallback processes?

dmitrygx:
> Isn't `--dist-init-addr` something that the user passes in through server args? Maybe it can be solved by users modifying the command. Not sure if it's necessary to add these fallback processes?

@ShangmingCai I use SGLang in the cloud w/o IPv4 support and we need to specify hostname of pods instead of their IPv6/IPv4 address which could be changed.
`gethostbyname()` doesn't support IPv6, it could be used for IPv4 only. In contrast, `getaddrinfo()` supports both IPv4 and IPv6. So, it is worth changing it.

ShangmingCai (commented):
Seems OK, but a little bit hacky to me. Do you have time to check on this modification? @thefacetakt (the ipv6 support PR contributor)

thefacetakt:
Hi, @ShangmingCai

Dmitry is a colleague of mine, so if it works on his machine, it works on mine. :)

I do have a couple small nitpicks:

1. The current code handles the case of `<ipv6 without brackets>:<port>`. Not sure if that's needed -- this form isn't valid according to [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986).
2. As gemini pointed out, it could be good to handle the case of just `<ipv6 without brackets>`. It looks like a simple `addr = maybe_wrap_ipv6(self.dist_init_addr)` at the top of the method, and then using `addr` instead of `self.dist_init_addr` later would suffice.
3. Personally, I find `host = infos[0][4][0]` a bit scary to read. Maybe add a comment like:

```
# getaddrinfo returns a list of 5-tuples:
# (family, type, proto, canonname, sockaddr)
# For AF_INET: sockaddr is (address, port)
# For AF_INET6: sockaddr is (address, port, flowinfo, scope_id)
```

Thanks!

hnyls2002:
Thanks @dmitrygx! IPv6-ready bootstrap resolution is native on main now, so I'm closing this as
  superseded. Please reopen if I've missed a case.

PR #10393, opened 2025-09-12
Add script for testing deterministic

Fridge003:
## Motivation

Modified based on `send_one.py`

This PR is merged into #10417

Usage:

First launch server
```bash
python3 -m sglang.launch_server --model-path Qwen/Qwen3-8B --disable-radix-cache
```

Then run script 
```bash
python3 -m sglang.test.test_deterministic --n-trials 100
```

Currently the number of unique samples is more than 1. After supporting batch invariant ops, the number of unique samples should be equal to 1.

## Checklist

### DIFF
--- a/python/sglang/test/test_deterministic.py
+++ b/python/sglang/test/test_deterministic.py
@@ -1,5 +1,5 @@
 """
-Send the same prompt for multiple times and test if the results are deterministic.
+Batch the same prompt in random batch sizes, and test if the results are consistent across different trials.
 
 Usage:
 python3 -m sglang.test.test_deterministic --n-trials <numer_of_trials>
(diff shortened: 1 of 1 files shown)


DevashishLal-CB:
I was looking into determinism for sglang and was wondering if we can modify the scheduler to create deterministic batches

sending one prompt works but I want to test cases with bs>1 also it would be faster than sending 4 prompts individually for (bs=4)

from what I understand after looking at [handle_batch_request](https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/managers/scheduler.py#L1327) is that each prompt gets added to the `_wating_queue` at the same time the event loop might pick up the requests before all the prompts have been added, we probably can improve determinism by having a server arg or env var that enforces atomic updates to the `_waiting_queue`

Fridge003:
> I was looking into determinism for sglang and was wondering if we can modify the scheduler to create deterministic batches
> 
> sending one prompt works but I want to test cases with bs>1 also it would be faster than sending 4 prompts individually for (bs=4)
> 
> from what I understand after looking at [handle_batch_request](https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/managers/scheduler.py#L1327) is that each prompt gets added to the `_wating_queue` at the same time the event loop might pick up the requests before all the prompts have been added, we probably can improve determinism by having a server arg or env var that enforces atomic updates to the `_waiting_queue`

We have a plan to suppport batch invariant ops. From my understanding the effect of batching will disappear after using batch invariant ops?

PR #10397, opened 2025-09-13
fix: tool parse in large streaming chunk beginning with normal content

JustinTong0323:
## Motivation
This pull request addresses a bug in the base_format_detector where it would fail to parse streaming tool calls if there was any text in the buffer before the bot_token.

When processing **large streaming chunk** like `unit.\n\n<tool_call>\n`, current parser would fail in calculating the start id.

The previous implementation incorrectly assumed the bot_token would be at the beginning of the string, causing the JSON parser to fail with a JSONDecodeError when it encountered non-JSON text.

The fix replaces the faulty logic with a call to current_text.find(self.bot_token) to correctly locate the start of the tool call payload. This makes the parsing more robust and ensures that only the JSON portion of the buffer is processed, resolving the decoding errors.

## Checklist

### DIFF
--- a/python/sglang/srt/function_call/base_format_detector.py
+++ b/python/sglang/srt/function_call/base_format_detector.py
@@ -165,6 +165,10 @@ class BaseFormatDetector(ABC):
                 tool_call_pos = current_text.find(self.bot_token)
                 if tool_call_pos != -1:
                     start_idx = tool_call_pos + len(self.bot_token)
+                elif self.current_tool_id > 0 and current_text.startswith(
+                    self.tool_call_separator
+                ):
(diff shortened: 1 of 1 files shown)


CatherineSue:
What does `unit.\n\n<tool_call>\n` mean? Is `unit.\n\n` a follow up of the previous chunk?

JustinTong0323:
> What does `unit.\n\n<tool_call>\n` mean? Is `unit.\n\n` a follow up of the previous chunk?

Yes, to reproduce just following the tool parse doc for qwen3next thinking

CatherineSue (approved):

PR #10413, opened 2025-09-13
Fix(SRT): Correctly initialize kv_head_num for MLA backend

### LINKED ISSUE #10381: [Bug] AttributeError: 'KVArgs' object has no attribute 'kv_head_num'
### Checklist

### Describe the bug

  File "/sgl-workspace/sglang/python/sglang/srt/managers/scheduler.py", line 2615, in run_scheduler_process
    scheduler.event_loop_overlap_disagg_prefill()
  File "/usr/local/lib/python3.12/dist-packages/torch/utils/_contextlib.py", line 120, in decorate_context
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/sgl-workspace/sglang/python/sglang/srt/disaggregation/prefill.py", line 344, in event_loop_overlap_disagg_prefill
    self.process_batch_result_disagg_prefill(tmp_batch, tmp_result)
  File "/sgl-workspace/sglang/python/sglang/srt/disaggregation/prefill.py", line 474, in process_batch_result_disagg_prefill
    self.send_kv_chunk(req, last_chunk=False, end_idx=req.tmp_end_idx)
  File "/sgl-workspace/sglang/python/sglang/srt/disaggregation/prefill.py", line 620, in send_kv_chunk
    req.disagg_kv_sender.send(page_indices)
  File "/sgl-workspace/sglang/python/sglang/srt/disaggregation/nixl/conn.py", line 584, in send
    new_xfer_handles = self.kv_mgr.add_transfer_request(
  File "/sgl-workspace/sglang/python/sglang/srt/disaggregation/nixl/conn.py", line 451, in add_transfer_request
    kv_xfer_handle = self.send_kvcache_slice(
                     ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/sgl-workspace/sglang/python/sglang/srt/disaggregation/nixl/conn.py", line 264, in send_kvcache_slice
    num_kv_heads = self.kv_args.kv_head_num
                   ^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'KVArgs' object has no attribute 'kv_head_num'

### Reproduction

sglang TP+DP+EP PD
same with https://github.com/sgl-project/sglang/issues/9787

### Environment

sglang:v0.5.2-cu126

avchauzov:
Resolves AttributeError in disaggregation by ensuring kv_head_num is always initialized in KVArgs for both MHA and MLA backends.

- For MHA, kv_head_num is derived from token_to_kv_pool.head_num.
- For MLA, it is calculated from model_config.num_attention_heads and tensor parallel size, as MLATokenToKVPool lacks a head_num attribute.

Adds unit and integration tests to verify the fix and prevent regressions. Closes #10381

## Motivation

This pull request fixes a critical `AttributeError` bug that occurs when using the disaggregation feature with the MLA backend. The goal is to improve the stability and reliability of the runtime by ensuring that the `KVArgs` object is always correctly initialized, preventing server crashes.

## Modifications
`
python/sglang/srt/disaggregation/prefill.py`: Modified the `_init_kv_manager` method to add conditional logic. It now correctly initializes the `kv_head_num` attribute for both MHA and MLA backends, addressing the root cause of the bug.

`python/sglang/test/attention/test_kv_head_initialization.py`: Added a new unit test to verify the initialization logic in isolation for both MHA and MLA backends.

`test/srt/test_disaggregation_kv_head_fix.py`: Added a new integration test to ensure the fix works within the full disaggregation runtime and prevents the original crash.

## Checklist

### DIFF
--- a/python/sglang/srt/disaggregation/prefill.py
+++ b/python/sglang/srt/disaggregation/prefill.py
@@ -42,6 +42,7 @@ from sglang.srt.disaggregation.utils import (
     poll_and_all_reduce,
     prepare_abort,
 )
+from sglang.srt.layers.dp_attention import get_attention_tp_size
 from sglang.srt.managers.schedule_batch import FINISH_LENGTH, Req, ScheduleBatch
 from sglang.srt.model_executor.forward_batch_info import ForwardMode, PPProxyTensors
 from sglang.srt.utils import (
@@ -132,6 +133,13 @@ class PrefillBootstrapQueue:
         kv_args.kv_item_lens = kv_item_lens
         if not self.is_mla_backend:
             kv_args.kv_head_num = self.token_to_kv_pool.head_num
+        else:
+            # For MLA backend, we need to calculate head_num from model config
+
+            kv_args.kv_head_num = (
+                self.scheduler.model_config.num_attention_heads
+                // get_attention_tp_size()
+            )
         kv_args.page_size = self.token_to_kv_pool.page_size
 
         kv_args.aux_data_ptrs, kv_args.aux_data_lens, kv_args.aux_item_lens = (
--- a/python/sglang/test/attention/test_kv_head_initialization.py
+++ b/python/sglang/test/attention/test_kv_head_initialization.py
@@ -1,0 +1,264 @@
+import unittest
+from unittest.mock import Mock, patch
+
+import torch
+
+from sglang.srt.disaggregation.prefill import PrefillBootstrapQueue
+from sglang.srt.disaggregation.utils import TransferBackend
+from sglang.test.test_utils import CustomTestCase
+
+
+class MockTokenToKVPool:
+    """Mock for token_to_kv_pool with configurable head_num."""
+
+    def __init__(self, head_num=32, page_size=1):
+        self.head_num = head_num
+        self.page_size = page_size
+        self.start_layer = 0
+
+    def get_contiguous_buf_infos(self):
+        """Mock method for getting buffer info."""
+        return [], [], []
+
+
+class MockMLATokenToKVPool:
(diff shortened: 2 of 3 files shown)


python/sglang/srt/disaggregation/prefill.py
```diff
@@ -52,6 +52,7 @@
 from sglang.srt.utils import (
     DynamicGradMode,
     broadcast_pyobj,
+    get_attention_tp_size,
```
thefacetakt: I think this import might be incorrect. The correct import should be:

```
from sglang.srt.layers.dp_attention import get_attention_tp_size
```

avchauzov: Done.

python/sglang/srt/disaggregation/prefill.py
```diff
@@ -132,6 +133,13 @@ def _init_kv_manager(self) -> BaseKVManager:
         kv_args.kv_item_lens = kv_item_lens
         if not self.is_mla_backend:
             kv_args.kv_head_num = self.token_to_kv_pool.head_num
+        else:
+            # For MLA backend, we need to calculate head_num from model config
+
+            kv_args.kv_head_num = (
+                self.scheduler.model_config.num_attention_heads
+                // get_attention_tp_size()
+            )
```
ShangmingCai, line 142: If I remember correctly, only `send_kvcache_slice` of mooncake/nixl requires `kv_args.kv_head_num`, and this function will never be called when using MLA. Can you elaborate on which case triggers the `AttributeError`?

ShangmingCai (commented):
Please move tests to the test/manual folder since they are not registered in the sglang CI run suite.

Issue #10278, opened 2025-09-10
[Feature] Support deterministic inference with Batch Invariant Ops


Fridge003:
### Checklist

- [ ] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [ ] 2. Please use English, otherwise it will be closed.

### Motivation

-[ ] Support torch flex attention backend #9947
-[ ] Enable batch invariant mode

Reference:
https://thinkingmachines.ai/blog/defeating-nondeterminism-in-llm-inference/
https://github.com/thinking-machines-lab/batch_invariant_ops

skyzh:
Hey, I'm interested in working on part of the issue but I have no idea where to start. Does "Support torch flex attention backend" mean that we need to wait for upstream support for deterministic kernels before we can integrate it with sglang? Would be great if you can have some pointers to the code we're going to patch in sglang. Thanks!

Fridge003:
> Hey, I'm interested in working on part of the issue but I have no idea where to start. Does "Support torch flex attention backend" mean that we need to wait for upstream support for deterministic kernels before we can integrate it with sglang? Would be great if you can have some pointers to the code we're going to patch in sglang. Thanks!

Thanks for your interest~ We feel appreciated if you can help us

There are several kernels we need for deterministic inference:
RMSNorm, Matmul, Attention

For RMSNorm and Matmul kernels, we can directly apply the ones provided in https://github.com/thinking-machines-lab/batch_invariant_ops/tree/main/batch_invariant_ops, with a single operation `set_batch_invariant_mode(True)`.

But for deterministic attention kernels, we need to work on how to obtain them.  For flex attention we don't need to wait for upstream support since it's already supported. But for other kernels like flashinfer or triton, we need some extra development on how to make them deterministic.

Currently SGLang doesn't support flex attention as the attention backend. One way of achieving deterministic inference would be first supporting flex attention, and then apply some simple modifications similar to https://github.com/vllm-project/vllm/pull/24583. The main blocker for this approach is to integrate flex attention into sglang.

There has been a related PR #9947, which is a naive example of supporting flex attention in sglang. However it's not usable due to slow speed. To make flex attention usable, we might need some extra management on attention masks, so that the kernel is compiled once for each forward batch.

skyzh:
[...]

PR #10417, opened 2025-09-14
[POC] Support deterministic inference

Fridge003:
co-author: @hebiao064 , @Qiaolin-Yu

## Motivation

#10278

## Modifications

1. (Flashinfer) Apply the batch invariant attention kernels from https://github.com/flashinfer-ai/flashinfer/pull/1675 
2. (Flashinfer) Only use BatchPrefillWithPagedKVCacheWrapper for prefilling (even when there is no prefix). Fix size of split-kv to 2048 for decode and 4096 for prefill.
3. Use forward_native for norm kernels
4. Enable batch invariant mode from https://github.com/thinking-machines-lab/batch_invariant_ops/tree/main
5. To handle chunked prefill batches,we choose to truncate sequence at multiple of split size. So that the edges of blocks are always aligned with split size.
6. Support cuda graph by disabling split-kv during graph capturing and graph replay for flashinfer
7. Generate a fixed sampling seed for each prompt, so temperature > 0 can also achieve determinism

## Reproduction

Environment: H200, Cuda12.6, sglang 0.5.2, torch 2.8.0, Python 3.12.11

1. For flashinfer, we need to install its latest main branch from source, under the instruction of https://github.com/flashinfer-ai/flashinfer?tab=readme-ov-file#install-from-source
2. Batch invariant ops needs to be built from https://github.com/thinking-machines-lab/batch_invariant_ops/tree/main
3. Apply the modification in this PR

Launch qwen3-8b:
```python3
python3 -m sglang.launch_server --model-path Qwen/Qwen3-8B --attention-backend flashinfer --disable-radix-cache 
```

Test determinism with single prompt and different batch sizes:
```bash
python3 -m sglang.test.test_deterministic --test-mode single

# Main branch
Total samples: 50, Unique samples: 6

# This PR
Total samples: 50, Unique samples: 1
```

Test determinism with mixture of short prompts and long prompts in each batch:
```bash
# Requires running multiple times
python3 -m sglang.test.test_deterministic --test-mode mixed

Prompt 1: total samples: 644, Unique samples: 1
Prompt 2: total samples: 423, Unique samples: 1
Long prompt: total samples: 208, Unique samples: 1
```

Test determinism with multiple prompts with different lengths of common prefix:
```bash

Prompt 0 with prefix length 1: total samples: 327, Unique samples: 1
Prompt 1 with prefix length 511: total samples: 316, Unique samples: 1
Prompt 2 with prefix length 2048: total samples: 326, Unique samples: 1
Prompt 3 with prefix length 4097: total samples: 306, Unique samples: 1
```

## Accuracy Tests

GSM8K

```bash

# Main Branch
Accuracy: 0.902
Invalid: 0.000
Latency: 105.043 s
Output throughput: 1631.791 token/s

# This PR
Accuracy: 0.901
Invalid: 0.000
Latency: 125.780 s
Output throughput: 1366.029 token/s
```

GPQA
```bash
python3 -m sglang.test.run_eval --port 30000 --eval-name gpqa --num-examples 198 --max-tokens 4096

# Main Branch
'score:std': np.float64(0.40881022918884974), 'score': np.float64(0.21212121212121213)

# This PR
'score:std': np.float64(0.40520665017240837), 'score': np.float64(0.20707070707070707)
```

## Temperature > 0
For launching, we need to specify pytorch sampling backend temporarily
```
python3 -m sglang.launch_server --model-path Qwen/Qwen3-8B --attention-backend flashinfer --disable-radix-cache --sampling-backend pytorch
```

For testing, we set temperature to a value greater than 0
```
python3 -m sglang.test.test_deterministic --test-mode mixed --temperature 0.2

# Results
Prompt 1: total samples: 530, Unique samples: 1
Prompt 2: total samples: 527, Unique samples: 1
Long prompt: total samples: 218, Unique samples: 1
```

## Checklist

### DIFF
--- a/python/sglang/srt/layers/attention/flashinfer_backend.py
+++ b/python/sglang/srt/layers/attention/flashinfer_backend.py
@@ -129,7 +129,7 @@ class FlashInferAttnBackend(AttentionBackend):
         if global_workspace_buffer is None:
             # different from flashinfer zero_init_global_workspace_buffer
             global_workspace_buffer = torch.empty(
-                1024 * 1024 * 1024,
+                2048 * 1024 * 1024,
                 dtype=torch.uint8,
                 device=model_runner.device,
             )
--- a/python/sglang/srt/layers/attention/flashinfer_backend.py
+++ b/python/sglang/srt/layers/attention/flashinfer_backend.py
@@ -39,6 +39,7 @@ if TYPE_CHECKING:
     from sglang.srt.layers.radix_attention import RadixAttention
     from sglang.srt.model_executor.model_runner import ModelRunner
 
+
 if is_flashinfer_available():
     from flashinfer import (
         BatchDecodeWithPagedKVCacheWrapper,
@@ -87,14 +88,15 @@ class FlashInferAttnBackend(AttentionBackend):
         super().__init__()
 
         # Parse constants
-        self.decode_use_tensor_cores = should_use_tensor_core(
-            kv_cache_dtype=model_runner.kv_cache_dtype,
-            num_attention_heads=model_runner.model_config.num_attention_heads
-            // get_attention_tp_size(),
-            num_kv_heads=model_runner.model_config.get_num_kv_heads(
-                get_attention_tp_size()
-            ),
-        )
+        # self.decode_use_tensor_cores = should_use_tensor_core(
+        #     kv_cache_dtype=model_runner.kv_cache_dtype,
+        #     num_attention_heads=model_runner.model_config.num_attention_heads
+        #     // get_attention_tp_size(),
+        #     num_kv_heads=model_runner.model_config.get_num_kv_heads(
+        #         get_attention_tp_size()
+        #     ),
+        # )
+        self.decode_use_tensor_cores = True
         self.max_context_len = model_runner.model_config.context_len
         self.skip_prefill = skip_prefill
         self.is_multimodal = model_runner.model_config.is_multimodal
@@ -127,7 +129,7 @@ class FlashInferAttnBackend(AttentionBackend):
         if global_workspace_buffer is None:
             # different from flashinfer zero_init_global_workspace_buffer
             global_workspace_buffer = torch.empty(
-                global_config.flashinfer_workspace_size,
+                1024 * 1024 * 1024,
                 dtype=torch.uint8,
                 device=model_runner.device,
             )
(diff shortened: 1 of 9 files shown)


python/sglang/srt/layers/attention/flashinfer_backend.py
```diff
@@ -848,6 +855,8 @@ def __init__(self, model_runner: ModelRunner, attn_backend: FlashInferAttnBacken
             assert self.attn_backend.num_wrappers == 1
             self.update = self.update_single_wrapper
 
+        self.fixed_split_size = 4096
```
Edenzzzz, line 869: if you don't use mixed chunk, prefill and decode wrapper can have different `fixed_split_size`

Fridge003: yes，we are not using mixed chunk

python/sglang/srt/layers/attention/flashinfer_backend.py
```diff
@@ -346,6 +348,8 @@ def init_forward_metadata_capture_cuda_graph(
                 decode_wrappers=decode_wrappers,
                 encoder_lens=encoder_lens,
                 spec_info=spec_info,
+                fixed_split_size=-1,
+                disable_split_kv=True,
```
Edenzzzz, line 352: For small bs, you might not need to disable split kv (e.g. bs 1-4)
For example, each SM launches 2 blocks, and on H100 for bs=1, split size 4096, it takes 1 million context length to break this limit. 
But this is optional and can be left for future work.
image

hebiao064:
FA3 Backend are batch invariant as well after this PR

**Single Mode**

Total samples: 50, Unique samples: 1

**Mixed Mode**
Prompt 1: total samples: 600, Unique samples: 1
Prompt 2: total samples: 485, Unique samples: 1
Long prompt: total samples: 190, Unique samples: 1

**Prefix Mode**
Prompt 0 with prefix length 1: total samples: 311, Unique samples: 1
Prompt 1 with prefix length 511: total samples: 319, Unique samples: 1
Prompt 2 with prefix length 2048: total samples: 323, Unique samples: 1
Prompt 3 with prefix length 4097: total samples: 322, Unique samples: 1


And after I disabled enable_batch_invariant_mode, single mode will fail

**Single Mode**

Total samples: 50, Unique samples: 5

hebiao064:
Triton Backend need some work
Single Mode
Total samples: 50, Unique samples: 4

python/sglang/srt/layers/attention/flashinfer_backend.py
hebiao064: should we also consider the num of heads?
Related comment: https://github.com/Dao-AILab/flash-attention/issues/609#issuecomment-1766860116

Edenzzzz: Yes this is more like an optional heuristic

PR #9834, opened 2025-08-31
perf: Avoid unnecessary data type conversions for DeepSeek-V3 on Blackwell

jinyangyuan-nvidia:
Optimize the performance of DeepSeek-V3 on Blackwell for small batch sizes by

- Change the output data type of router GEMM to FP32
- Convert the data type of `correction_bias` in the object initialization
- ~~Use the `cutlass` backend for the first GEMM in shared experts~~ (only benefits small batches)

The performance has been tested on B200 using the following commands.
```
python3 -m sglang.launch_server \
    --model-path ${path_model} \
    --trust-remote-code \
    --quantization modelopt_fp4 \
    --tp 8 \
    --moe-runner-backend flashinfer_trtllm \
    --attention-backend trtllm_mla \
    --enable-flashinfer-allreduce-fusion

for batch_size in 1 2 4 8 16 32 64 128 256; do
    python3 -m sglang.bench_serving \
        --model ${path_model} \
        --dataset-name random \
        --backend sglang-oai \
        --random-range-ratio 1 \
        --random-input-len 1200 \
        --random-output-len 256 \
        --max-concurrency ${batch_size} \
        --num-prompts $(( batch_size * 10 ))
done
```

The tested median TTFT and median ITL are listed below. Because TTFT has very large variance (comparing it is not very meaningful), the performance is assessed using ITL.

<table>
    <thead>
        <tr>
            <th rowspan=3>Batch Size</th>
            <th rowspan=2, colspan=2>Original</th>
            <th colspan=6>Avoid Dtype Conversion</th>
        </tr>
        <tr>
            <th colspan=2>cuDNN, cuDNN</th>
            <th colspan=2>CUTLASS, cuDNN</th>
            <th colspan=2>CUTLASS, CUTLASS</th>
        </tr>
        <tr>
            <th>TTFT (ms)</th>
            <th>ITL (ms)</th>
            <th>TTFT (ms)</th>
            <th>ITL (ms)</th>
            <th>TTFT (ms)</th>
            <th>ITL (ms)</th>
            <th>TTFT (ms)</th>
            <th>ITL (ms)</th>
        </tr>
    </thead>
    <tbody>
<tr><td align=center>1</td> <td align=center>72.63</td> <td align=center>7.41</td> <td align=center>72.21</td> <td align=center>7.15</td> <td align=center>74.54</td> <th align=center>7.07</th> <td align=center>69.93</td> <td align=center>7.15</td></tr>
[...]

PR #10426, opened 2025-09-14
Fix correction bias undefined behavior for nvfp4 models

fzyzcjy:
## Motivation

TODO tomorrow: (1) add assertions at kernel level (2) check shall we make it bf16 etc (3) test e2e

## Checklist

### DIFF
--- a/python/sglang/srt/layers/moe/topk.py
+++ b/python/sglang/srt/layers/moe/topk.py
@@ -675,7 +675,7 @@ def biased_grouped_topk_gpu(
     ):
         topk_weights, topk_ids = moe_fused_gate(
             gating_output.to(dtype=torch.float32),
-            correction_bias.to(dtype=torch.float32),
+            correction_bias,
             num_expert_group,
             topk_group,
             topk,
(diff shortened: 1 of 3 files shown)


python/sglang/srt/layers/moe/topk.py
```diff
@@ -675,7 +675,7 @@ def biased_grouped_topk_gpu(
     ):
         topk_weights, topk_ids = moe_fused_gate(
             gating_output.to(dtype=torch.float32),
-            correction_bias,
+            correction_bias.to(dtype=torch.float32),
```
zhyncs: In which use case will we have accuracy issues?

fzyzcjy: it is more about UB in nvfp4 ckpt, not check in detail yesterday, I will check today.

ch-wan: Is it related to #9834?

fzyzcjy: yes I think so, I am going to investigate it today

zhyncs (approved):

zhyncs:
fix dsv3 fp4 cutlass moe etp

PR #10465, opened 2025-09-15
feat: support pip install sglang

zhyncs:
## Motivation

- We should support `pip install sglang` on NVIDIA GPU.
- If we don't back up `pyproject_other.toml`, other platforms will encounter errors during installation.

## Checklist

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -82,6 +82,12 @@ test = [
   "pytest",
   "tabulate",
 ]
+tracing = [
+    "opentelemetry-sdk",
+    "opentelemetry-api",
+    "opentelemetry-exporter-otlp",
+    "opentelemetry-exporter-otlp-proto-grpc",
+]
 all = ["sglang[test]"]
 blackwell = ["nvidia-cutlass-dsl==4.1.0", "sglang[test]"]
 dev = ["sglang[test]"]
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -63,7 +63,6 @@ dependencies = [
   "torchvision",
   "cuda-python",
   "flashinfer_python==0.3.1",
-  "nvidia-cutlass-dsl==4.1.0",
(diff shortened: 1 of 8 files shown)


python/pyproject.toml
```diff
@@ -10,163 +10,114 @@ readme = "README.md"
...
-all_hpu = ["sglang[srt_hpu]", "sglang[openai]", "sglang[anthropic]", "sglang[decord]"]
-all_cpu = ["sglang[srt_cpu]", "sglang[openai]", "sglang[anthropic]", "sglang[decord]"]
-all_npu = ["sglang[srt_npu]", "sglang[openai]", "sglang[anthropic]", "sglang[decord]"]
-
-dev = ["sglang[all]", "sglang[test]"]
-dev_hip = ["sglang[all_hip]", "sglang[test]"]
-dev_xpu = ["sglang[all_xpu]", "sglang[test]"]
-dev_hpu = ["sglang[all_hpu]", "sglang[test]"]
-dev_cpu = ["sglang[all_cpu]", "sglang[test]"]
+all = ["sglang[test]"]
```
zhyncs, line 91: This is just to be compatible with previous usage, hoping to remove it in the future.

python/pyproject.toml
```diff
@@ -10,163 +10,113 @@ readme = "README.md"
...
+blackwell = ["nvidia-cutlass-dsl==4.1.0", "sglang[test]"]
```
zhyncs, line 92: After 4.1.0, we should move `nvidia-cutlass-dsl` into `dependencies`.

python/pyproject_other.toml
zhyncs, line 1: This file is same as `python/pyproject.toml` in latest main.

zhyncs:
After this PR, other platforms can maintain their own pyproject.toml without necessarily using pyproject_other.toml.

PR #10489, opened 2025-09-16
Pipeline parallel idle optimization

gholmes829:
## Motivation

Online serving with pipeline parallelism currently suffers from a big problem where same-node GPUs with `pp_rank > 0` spin incessantly **while idle**, incurring near 100% utilization and wasting significant power. 

I first observe this on my 2x3090 single-node system with `--pp-size 2` and PP layer partition split 3:1. While idle, 2nd GPU (PP rank 1) draws constant 150W and 100% utilization.

This may not seem like a whole lot when compared to data center setups, but it leaves the gpu unusable for other tasks (despite only using ~8GB VRAM in my case), causes fans to whine nonstop, spikes my electricity bill, and overall contributes to avoidable wear and tear.

I traced this undesirable behavior down to how non-zero-rank PP stages poll recv which internally calls NCCL's `cudaStreamSynchronize`. Profiling with Nsys confirms this.

The solution proposed in this PR is actually quite simple and non-invasive. I discuss more in following section, but we can solve the problem by adding a touch of idle-friendly CPU sync via ZMQ.

This reducing power consumption while idle from ~150W to ~30W, enables GPU to enter proper sleep state, and generally frees it for other tasks since it's not stuck spinning at 100%.

The theoretical performance impact is truly negligible. In practice, I was actually happily surprised to observe modest performance **gains** of ~5% (discussed below).

I will stop rambling, but please advise if I can provide any more detail and/or if any revisions are needed.

CCing pipeline parallelism owner @Ying1123 for review as per contribution guide.
  

## Modifications

  - **Added `--enable-pp-sleep-on-idle` flag**: Opt-in feature for when `--pp-size > 1` and `pp_size_per_node > 1`. I personally don't see any reason not to just bake it in, but figured it's better to start defensively.
  - **Implemented CPU-only wake signaling**: Uses ZMQ PUSH/PULL sockets to coordinate PP stages without GPU involvement
    - PP rank N sends wake signal before sending work to PP rank N+1
    - PP rank N+1 waits on ZMQ socket instead of NCCL polling when idle
  - **Modified files**:
    - `scheduler.py`: Added wake socket initialization and idle wait logic
    - `server_args.py`: Added new flag and IPC mappings for wake sockets
    - `engine.py` & `data_parallel_controller.py`: Setup wake socket connections for same-node PP ranks
    - `docs/advanced_features/server_arguments.md`: Documented new flag
    - `test/srt/test_pp_single_node.py`: Added basic test

## Accuracy Tests

N/A -- no changes to anything that would affect model accuracy.

## Benchmarking and Profiling

I've run a slew of benchmarks and tests to catch any performance regressions or sync issues.

For this section, I show results comparing "flag" (`--enable-pp-sleep-on-idle` enabled) vs "baseline" (code without any of my changes).

For these tests, I use dense Qwen3-32B on single node and `--pp-size 2` with PP layer partition split 3:1.

I run benchmarks with following:
``` bash
docker compose --env-file .env -f docker-compose.sglang.yml run --rm \
    --entrypoint "python -m sglang.bench_serving" sglang \
    --backend sglang-oai \
    --host sglang \
    --port 8080 \
    --dataset-name random \
    --num-prompts 256 \
    --random-input-len 1024 \
    --random-output-len 512 \
    --random-range-ratio 0.2 \
    --max-concurrency 16 \
    --request-rate 16 \
    --warmup-requests 16 \
    --apply-chat-template \
    --extra-request-body '{"temperature":0.7,"top_p":0.8,"top_k":20,"min_p":0,"chat_template_kwargs":{"enable_thinking":false}}'
```

As this is running, I use `nvidia-smi` to collect data for analysis. Lets start by looking at results of one benchmark comparison:

pr-comparison-annotated-focused

I should have extended the window to cover more idle time, but take a close look before and after the benchmark runs. For GPU 1 (`pp_rank > 0`) note the distinct and unrelenting elevated plateau of power and utilization on the baseline (right). Now note how with the flag enabled, the usage and power actually drop to true idling state while idle (left).

Now let's check on performance, I was initially concerned that the solution might hurt performance due to sync overhead. Here are 3 trials (6 benchmarks) comparing baseline (without my changes) vs flag (include my changes and set `--enable-pp-sleep-on-idle`).

| Metric | Baseline 1 | Baseline 2 | Baseline 3 | **Baseline Avg** | Flag 1 | Flag 2 | Flag 3 | **Flag Avg** | **% Diff (Flag vs Baseline)** |
|--------|------------|------------|------------|------------------|---------|---------|---------|--------------|-------------------------------|
| **Benchmark duration (s)** | 596.03 | 613.21 | 562.59 | **590.61** | 609.87 | 542.24 | 527.16 | **559.76** | **-5.2%** |
| **Request throughput (req/s)** | 0.43 | 0.42 | 0.46 | **0.44** | 0.42 | 0.47 | 0.49 | **0.46** | **+4.5%** |
| **Input token throughput (tok/s)** | 261.14 | 253.82 | 276.66 | **263.87** | 255.21 | 287.04 | 295.25 | **279.17** | **+5.8%** |
| **Output token throughput (tok/s)** | 135.37 | 131.58 | 143.42 | **136.79** | 132.30 | 148.80 | 153.06 | **144.72** | **+5.8%** |
| **Total token throughput (tok/s)** | 396.51 | 385.40 | 420.07 | **400.66** | 387.51 | 435.84 | 448.31 | **423.89** | **+5.8%** |
| **Mean E2E Latency (ms)** | 36,130.29 | 37,106.19 | 33,928.42 | **35,721.63** | 36,987.83 | 32,748.57 | 31,817.27 | **33,851.22** | **-5.2%** |
| **Median E2E Latency (ms)** | 35,254.99 | 36,312.87 | 33,274.76 | **34,947.54** | 36,250.98 | 31,236.85 | 30,624.33 | **32,704.05** | **-6.4%** |
| **Mean TTFT (ms)** | 1,411.06 | 1,473.27 | 1,356.12 | **1,413.48** | 1,459.29 | 1,305.98 | 1,276.88 | **1,347.38** | **-4.7%** |
| **Median TTFT (ms)** | 1,085.16 | 1,126.55 | 1,058.60 | **1,090.10** | 1,152.60 | 985.95 | 937.75 | **1,025.43** | **-5.9%** |
| **P99 TTFT (ms)** | 6,593.62 | 6,557.62 | 6,396.49 | **6,515.91** | 6,469.77 | 6,282.74 | 6,403.28 | **6,385.26** | **-2.0%** |
| **Mean ITL (ms)** | 110.52 | 113.42 | 103.68 | **109.21** | 113.09 | 100.08 | 97.25 | **103.47** | **-5.3%** |
| **Median ITL (ms)** | 69.21 | 70.94 | 64.84 | **68.33** | 71.93 | 61.90 | 59.96 | **64.60** | **-5.5%** |
| **P95 ITL (ms)** | 217.64 | 214.15 | 183.85 | **205.21** | 214.90 | 176.12 | 173.14 | **188.05** | **-8.4%** |
| **P99 ITL (ms)** | 1,289.18 | 1,359.18 | 1,193.78 | **1,280.71** | 1,328.65 | 1,174.56 | 1,126.17 | **1,209.79** | **-5.5%** |

**Notable takeaways:**
- While idle (e.g. the time between sglang being ready and me actually starting the benchmark, and the time between benchmark completing and me killing the container) "flag" goes truly idle ~30W ~0% util whereas "baseline" eats 2nd GPU at constant ~150W and ~100% util.
- Benchmarks show increase in throughput by ~5.8%, and reduction to latency by 5-6% on average

I honestly wasn't expecting performance gains here, I'm mostly in this for the true idle. Thinking about it though, it Is plausible that less NCCL polling overhead makes everything more efficient with less context switching, or maybe it has something to do with avoiding thermal throttling caused by the unnecessary polling. Or maybe its not statistically significant. But in any case, the point here it to show this implementation does not hurt performance.

In any case, I'd argue this PR is huge for anyone wanting to run PP efficiently, especially with asymmetric partitioning and shared workloads. And as a ~100 LOC opt-in change without concessions, I look forward to this helping drive sglang even further ahead of the competition.

## Checklist

### DIFF
--- a/docs/advanced_features/server_arguments.md
+++ b/docs/advanced_features/server_arguments.md
@@ -423,6 +423,7 @@ Please consult the documentation below and [server_args.py](https://github.com/s
 | `--enable-mscclpp` | Enable using mscclpp for small messages for all-reduce kernel and fall back to NCCL. | `False` | bool flag (set to enable) |
 | `--enable-torch-symm-mem` | Enable using torch symm mem for all-reduce kernel and fall back to NCCL. Only supports CUDA device SM90 and above. SM90 supports world size 4, 6, 8. SM10 supports world size 6, 8. | `False` | bool flag (set to enable) |
 | `--disable-overlap-schedule` | Disable the overlap scheduler, which overlaps the CPU scheduler with GPU model worker. | `False` | bool flag (set to enable) |
+| `--enable-pp-sleep-on-idle` | Enable pipeline parallel sleep while idle, preventing costly GPU polling during idle periods. | `False` | bool flag (set to enable) |
 | `--enable-mixed-chunk` | Enabling mixing prefill and decode in a batch when using chunked prefill. | `False` | bool flag (set to enable) |
 | `--enable-dp-attention` | Enabling data parallelism for attention and tensor parallelism for FFN. The dp size should be equal to the tp size. Currently DeepSeek-V2 and Qwen 2/3 MoE models are supported. | `False` | bool flag (set to enable) |
 | `--enable-dp-lm-head` | Enable vocabulary parallel across the attention TP group to avoid all-gather across DP groups, optimizing performance under DP attention. | `False` | bool flag (set to enable) |
--- a/python/sglang/srt/entrypoints/engine.py
+++ b/python/sglang/srt/entrypoints/engine.py
@@ -25,6 +25,7 @@ import multiprocessing as mp
 import os
 import random
 import signal
+import tempfile
 import threading
 import time
 from typing import AsyncIterator, Callable, Dict, Iterator, List, Optional, Tuple, Union
@@ -891,6 +892,20 @@ def _launch_scheduler_processes(
 
         for pp_rank in pp_rank_range:
             for tp_rank in tp_rank_range:
+                # Compute mappings needed so PP stages may sleep on idle
+                if server_args.enable_pp_sleep_on_idle and tp_rank == tp_rank_range[0]:
+                    # Add notifier-listener pair if this rank precedes another on the same node
+                    if (pp_rank // pp_size_per_node) == (
+                        (pp_rank + 1) // pp_size_per_node
+                    ):
+                        notifier_listener_ipc_name = port_args.pp_idle_wakeup_listeners_ipc_names.setdefault(
+                            (None, pp_rank + 1),
+                            f"ipc://{tempfile.NamedTemporaryFile(delete=False).name}",
+                        )
+                        port_args.pp_idle_wakeup_notifiers_ipc_names[None, pp_rank] = (
+                            notifier_listener_ipc_name
+                        )
+
                 reader, writer = mp.Pipe(duplex=False)
                 gpu_id = (
                     server_args.base_gpu_id
--- a/python/sglang/srt/managers/data_parallel_controller.py
+++ b/python/sglang/srt/managers/data_parallel_controller.py
@@ -17,6 +17,7 @@ import faulthandler
 import logging
 import multiprocessing as mp
 import signal
+import tempfile
 import threading
 import time
 from enum import Enum, auto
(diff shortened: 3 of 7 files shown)


gholmes829:
Edit: rebased on `main` to sync branch, added annotations to figure in description for clarity, and description grammar fixes -- no new code changes since originally published

ShangmingCai (commented):
With this added `server_args.enable_pp_sleep_on_idle`, the impl looks safe to me. My opinion is that this could be an experimental feature for the energy-friendly use case. However, a detailed review from the main author of SGLang PP support, @Ying1123, is still needed.

gholmes829:
@ShangmingCai thanks for the feedback! Looking forward to the review from @Ying1123.

gholmes829:
@XucSh sure thing.

I've re-applied my changes on top of latest main and pushed over original commits, since the conflicts with new mixin pattern were pretty significant.

From what I can tell, my changes seem to be compatible with the new async and PD disaggregation changes, but please call out if not the case.

Also fixed a gap in the original where PP sleep-on-idle only worked with DP attention - now works for regular DP+PP as well.

Haven't had time to test locally just yet, so lets see how the pipeline tests go. Looks like they are rate limited, so I will trigger them again in a bit.

Looking forward to the review!

gholmes829:
@ShangmingCai @XucSh do y'all have any guidance on review timeline for this PR?

No immediate rush on my account -- just want to follow up now that we are post async PP integration as it would be nice to see this feature cleanly merged before more changes require more rebase resolution work.

Please advise if anything else is needed from my end, thanks!

PR #10497, opened 2025-09-16
[Structural Output] Get prepared for the upcoming support for XGrammar's latest structural tag format

aaronyeeio:
## Motivation

Enhance support for XGrammar's latest structural tag format in documentation and backend. Added new code examples in the Jupyter notebook and updated the backend to handle both new and deprecated structural tag formats. Introduced a new response format class for compatibility with the updated API.
See: https://github.com/mlc-ai/xgrammar/pull/420

## Modifications

Modified: `protocol.py` and `xgrammar_backend.py` for compatibility through all entrypoints.

Docs also updated. Check docs for the usage.

## Accuracy Tests

Currently XGrammar has not published a new version with this PR yet and I believe they soon will. So please test it with the latest build of XGrammar.

## Checklist


python/sglang/srt/entrypoints/openai/protocol.py
```diff
@@ -126,6 +126,9 @@ class StructuralTagResponseFormat(BaseModel):
     structures: List[StructuresResponseFormat]
     triggers: List[str]
 
+class StructuralTagResponseFormatV1(BaseModel):
+    type: Literal["structural_tag"]
+    format: Dict[str, Any]
```
Ubospica, line 131: You can just import StructuralTag from xgrammar, that's also a pydantic BaseModel and all its fields are type checked.

Ubospica (commented):
Thank you so much for your support of the new Structural Tag! #10691 is also working on this, so it would be great if we could have a joint effort. The change in the jupyter notebook makes sense to me.

PR #10510, opened 2025-09-16
check  master server for mooncake store

huangtingwei9988:
## Motivation
When deploying Mooncake Store online, the startup order of Mooncake Store master and sglang may not be guaranteed. Therefore, a check server is added. When Mooncake Store Master has not yet been fully started successfully, sglang waits for the master for 10 minutes to prevent rapid failure when executing `store.setup`

cc @ykwd @stmatengss 

co-author @zhaoyongke
## Checklist


python/sglang/srt/mem_cache/storage/mooncake_store/mooncake_store.py
```diff
@@ -171,6 +177,52 @@ def __init__(self, storage_config: HiCacheStorageConfig = None):
...
+
+            if check_health_resp.text != "OK":
+                logger.info("waiting Mooncake store master started, cost_time: %s", time.time() - start_time)
+                time.sleep(3)
+                continue
+
+            health_result = True
+
+            try:
+                check_segments_resp = requests.get(segments_url)
```
stmatengss: Why need to check all segments?

python/sglang/srt/mem_cache/storage/mooncake_store/mooncake_store.py
```diff
@@ -1,16 +1,19 @@
...
 from typing import Any, List, Optional
+import requests
 
 import torch
 
 from sglang.srt.mem_cache.hicache_storage import HiCacheStorage, HiCacheStorageConfig
 
 DEFAULT_GLOBAL_SEGMENT_SIZE = 4 * 1024 * 1024 * 1024  # 4 GiB
 DEFAULT_LOCAL_BUFFER_SIZE = 16 * 1024 * 1024  # 16 MB
+SETUP_TIMEOUT = 600 # 10min
```
stmatengss: Ten minutes is too long to start Mooncake Master. I guess 60 seconds might be preferable.

python/sglang/srt/mem_cache/storage/mooncake_store/mooncake_store.py
stmatengss: The logic should be like this:
```
While the interval in setup_timeout
    if check(health):
        if check(check_segments):
             is_success = true
             break
```

ykwd:
It’s a good idea to prevent startup failures caused by the master server starting late. However, I have some concerns about the current solution:

* The HTTP server port used to query master information is hard-coded as `9003`. This is problematic because the port number is configurable when starting the HTTP server.
* Potential deadlocks may occur. The SGLang server waits for the master to become active and for at least one global segment to be registered before it creates the store client and registers its own segment. But if users choose not to run a separate store service—instead relying entirely on segments provided by the SGLang server—this could lead to deadlocks.
* The `check_server` function sends HTTP requests directly to `self.config.master_server_address`. However, in HA mode, `master_server_address` actually points to the etcd endpoints. In this case, the Mooncake client first queries etcd through this address to obtain the actual master server address.

For these reasons, I think it would be better to implement the waiting logic inside the Mooncake store client. Whether or not to wait for the master to start could then be exposed as an optional client parameter.

stmatengss:
Checking segments is enough to ensure a healthy status. It can simplify the implementations.

python/sglang/srt/mem_cache/storage/mooncake_store/mooncake_store.py
zhaoyongke: In our case, init a huge segment(TB level) will take minutes, 60s will always timeout

python/sglang/srt/mem_cache/storage/mooncake_store/mooncake_store.py
zhaoyongke: > Why need to check all segments?

Warmup will put a key-value into store, so at least a segment should be mounted before warmup.

zhaoyongke:
> * The HTTP server port used to query master information is hard-coded as `9003`. This is problematic because the port number is configurable when starting the HTTP server.

Set a unified env variable for both sglang and mooncake config?

> * Potential deadlocks may occur. The SGLang server waits for the master to become active and for at least one global segment to be registered before it creates the store client and registers its own segment. But if users choose not to run a separate store service—instead relying entirely on segments provided by the SGLang server—this could lead to deadlocks.

Exactly. We should put self.check_server() after ret_code = self.store.setup(...) in this case

> * The `check_server` function sends HTTP requests directly to `self.config.master_server_address`. However, in HA mode, `master_server_address` actually points to the etcd endpoints. In this case, the Mooncake client first queries etcd through this address to obtain the actual master server address.

Known issue. Now only support non-HA mode.

ykwd:
> > * The HTTP server port used to query master information is hard-coded as `9003`. This is problematic because the port number is configurable when starting the HTTP server.
> 
> Set a unified env variable for both sglang and mooncake config?

Mooncake master already supports configuring this parameter. Any way to allow configuring this in sglang side should work.
> 
> > * Potential deadlocks may occur. The SGLang server waits for the master to become active and for at least one global segment to be registered before it creates the store client and registers its own segment. But if users choose not to run a separate store service—instead relying entirely on segments provided by the SGLang server—this could lead to deadlocks.
> 
> Exactly. We should put self.check_server() after ret_code = self.store.setup(...) in this case

`check_server` is called before `store.setup` to ensure the success of the latter. Putting it after `store.setup` may not help with this purpose.

> 
> > * The `check_server` function sends HTTP requests directly to `self.config.master_server_address`. However, in HA mode, `master_server_address` actually points to the etcd endpoints. In this case, the Mooncake client first queries etcd through this address to obtain the actual master server address.
> 
> Known issue. Now only support non-HA mode.

Currently, both HA and Non-HA mode are available to users. If we only support non-HA mode, then HA mode users will no longer be able to start the mooncake backend, as `check_server` will not pass under non-HA mode.

To ensure compatibility with HA mode and to make the “wait-for-master” logic easier to maintain, I recommend implementing it on the Mooncake client side. If we really want to handle this on the SGLang side, then at the very least, it should be made an optional feature.

stmatengss (approved):
LGTM

stmatengss (changes requested):
It seems we can't pass the test in local run.

huangtingwei9988:
> It seems we can't pass the test in local run.

@stmatengss After discussing with @ykwd, I added an option to open it if necessary.

stmatengss (approved):

stmatengss:
@ykwd Could you double check this changes？

hzh0425 (approved):

python/sglang/srt/mem_cache/storage/mooncake_store/mooncake_store.py
```diff
@@ -63,10 +71,16 @@ def from_file() -> "MooncakeStoreConfig":
                 config.get("global_segment_size", DEFAULT_GLOBAL_SEGMENT_SIZE)
             ),
             # Zero copy interface does not need local buffer
-            local_buffer_size=DEFAULT_LOCAL_BUFFER_SIZE,
+            local_buffer_size=config.get(
+                "local_buffer_size", DEFAULT_LOCAL_BUFFER_SIZE
```
ykwd: Why do we need local_buffer_size parameter?

zhaoyongke: Keep this to be compatible with non-zerocopy calls, such as warmup().

ykwd: Currently, the default 16 MB setting should be sufficient for the warmup, and there are no other parts of the code that use the local buffer.
Perhaps we could consider hiding this parameter to avoid confusing users.

huangtingwei9988: okk @ykwd

xiezhq-hermann (approved):

PR #10512, opened 2025-09-16
[HiCache] Configurable and Dynamic Prefetch Timeout

ykwd:
## Motivation

In real-world deployments, **timeout** serves as a practical prefetch policy. On the one hand, prefetch latency depends on factors such as hardware conditions, the model, and the number of tokens to prefetch. On the other hand, the prefetch deadline is constrained by the SLO. Therefore, allowing users to flexibly configure the prefetch timeout parameter—and dynamically determine its value based on the number of tokens to prefetch per request—is highly valuable.

This PR introduces three new parameters in `hicache-storage-backend-extra-config`:

* `prefetch_threshold`
* `prefetch_timeout_base`
* `prefetch_timeout_per_ki_token`

The timeout for a request is calculated as:

```
timeout = prefetch_timeout_base + prefetch_timeout_per_ki_token * num_token_to_fetch / 1024
```

Here, `prefetch_timeout_base` represents overheads unrelated to token count, such as scheduling and synchronization.

cc. @xiezhq-hermann @huangtingwei9988 

## Checklist


huangtingwei9988:
@ykwd Give an example of hicache-storage-backend-extra-config？

python/sglang/srt/mem_cache/hiradix_cache.py
```diff
@@ -121,6 +131,73 @@ def __init__(
...
+                f"prefetch_timeout_base must be a number, got {type(prefetch_timeout_base).__name__}. Converting to float."
+            )
+        prefetch_timeout_base = float(prefetch_timeout_base)
+
+        # Validate prefetch_timeout_per_ki_token is a float
+        if not isinstance(prefetch_timeout_per_ki_token, (int, float)):
+            logger.warning(
+                f"prefetch_timeout_per_ki_token must be a number, got {type(prefetch_timeout_per_ki_token).__name__}. Converting to float."
+            )
+        prefetch_timeout_per_ki_token = float(prefetch_timeout_per_ki_token)
```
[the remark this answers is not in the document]
ykwd: This is intentional, because if the given parameter is incorrect, we should stop rather than continue serving. Otherwise, the system may behave differently from users’ expectations—silently and without notice—which is even more undesirable.

ykwd:
> @ykwd Give an example of hicache-storage-backend-extra-config？

```
--hicache-storage-backend-extra-config '{"prefetch_threshold":512, "prefetch_timeout_base": 0.5, "prefetch_timeout_per_ki_token": 0.2}'
```

Together with other parameters:
```
--hicache-storage-backend-extra-config '{"master_server_address": "127.0.0.1:9821", "local_hostname": "localhost", "metadata_server": "http://127.0.0.1:8199/metadata", "global_segment_size": 4294967296, "local_buffer_size": 16777216, "protocol": "rdma", "device_name": "erdma_0,erdma_1","prefetch_threshold":512, "prefetch_timeout_base": 0.5, "prefetch_timeout_per_ki_token": 0.2}'
```

python/sglang/srt/mem_cache/hiradix_cache.py
```diff
@@ -127,6 +137,73 @@ def __init__(
...
+                extra_config = json.loads(storage_backend_extra_config)
+            except Exception as e:
+                logger.error(f"Invalid backend extra config JSON: {e}")
+                # Cannot continue if extra config is invalid
+                raise e
+
+        # Extract parameters from extra_config if they exist, otherwise use defaults
+        prefetch_threshold = 256  # tokens
+        prefetch_timeout_base = 1  # seconds
+        prefetch_timeout_per_ki_token = 0.25  # seconds
```
xiezhq-hermann: any justification on this number?

python/sglang/srt/mem_cache/hiradix_cache.py
```diff
@@ -127,6 +137,73 @@ def __init__(
...
+            prefetch_timeout_per_ki_token = extra_config.pop(
+                "prefetch_timeout_per_ki_token", prefetch_timeout_per_ki_token
+            )
+
+        # Validate prefetch_threshold is an integer
+        if not isinstance(prefetch_threshold, int):
+            logger.warning(
+                f"prefetch_threshold must be an integer, got {type(prefetch_threshold).__name__}. Converting to int."
+            )
+            prefetch_threshold = int(prefetch_threshold)
```
xiezhq-hermann: this might fail if provided invalid strings, we might want to use try and except clause

python/sglang/srt/mem_cache/hiradix_cache.py
```diff
@@ -127,6 +137,73 @@ def __init__(
...
+
+        # Validate prefetch_timeout_base is a float
+        if not isinstance(prefetch_timeout_base, (int, float)):
+            logger.warning(
+                f"prefetch_timeout_base must be a number, got {type(prefetch_timeout_base).__name__}. Converting to float."
+            )
+        prefetch_timeout_base = float(prefetch_timeout_base)
```
xiezhq-hermann: same here

python/sglang/srt/mem_cache/hiradix_cache.py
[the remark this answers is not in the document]
xiezhq-hermann: I mean, that case you should directly raise error instead of try convert it into int and float. The error message of conversion failure might be confusing

python/sglang/srt/mem_cache/hiradix_cache.py
```diff
@@ -507,7 +584,10 @@ def can_terminate_prefetch(self, operation: PrefetchOperation):
             can_terminate = completed
         elif self.prefetch_stop_policy == "timeout":
             can_terminate = completed or (
-                time.monotonic() - operation.start_time > self.prefetch_timeout
+                # If hash_value has not been computed in timeout_base seconds, terminate it.
+                time.monotonic() - operation.start_time
+                > self.prefetch_timeout_base
+                + len(operation.hash_value) * self.prefetch_timeout_per_page
```
xiezhq-hermann: is this base + bonus timeout the only structure we will be using for adaptive timeout?

huangtingwei9988: > is this base + bonus timeout the only structure we will be using for adaptive timeout?

I think it can be abstracted into a method, and then users can customize their implementation

ykwd:
Updated based on our discussion this Tuesday:

1. If the input parameter has the wrong type, directly raise an error instead of trying to convert it into `int` or `float`.
2. Refactored the timeout check into a separate method, so we can add alternative conditions more easily.
3. Added comments to clarify the meaning of default values for the related parameters.

xiezhq-hermann (approved):

ykwd:
Thanks for improving the code style and comments. The changes look great to me.

PR #10520, opened 2025-09-16
Support calling functions with RPC without arguments

nathanrchn:
## Motivation

It was not possible to call functions with less than 2 arguments using rpc. This PR fixes that and allow to have optional parameters.

## Checklist


hnyls2002:
Thanks @nathanrchn! RPC calls without arguments are handled on main now (a no-arg fallback in the
  scheduler), so I'm closing this. Please reopen if I've missed a case.

PR #10530, opened 2025-09-16
skip vision_model for lora

gongwei-130:
## Motivation
when multimodal is enabled,  visual_model module will be added to Lora for Llama4, and would introduce assert failure of shape not matched.

## Checklist

### DIFF
--- a/python/sglang/srt/lora/lora_manager.py
+++ b/python/sglang/srt/lora/lora_manager.py
@@ -415,6 +415,9 @@ class LoRAManager:
         replace_submodule(self.base_model, module_name, lora_module)
         return lora_module
 
+    def should_skip_lora(self, module_name):
+        return module_name.find("vision_model.model") != -1
+
     def init_lora_modules(self):
         # Look-up table that essentially maps (layer_index, module_name) to the corresponding LoRA module.
         self.lora_modules: List[Dict[str, BaseLayerWithLoRA]] = [
@@ -433,7 +436,7 @@ class LoRAManager:
                 continue
 
             # Skip vision model
-            if module_name.find("vision_model.model") != -1:
+            if self.should_skip_lora(module_name):
                 continue
 
             # The module should be converted if it is included in target_names
--- a/python/sglang/srt/lora/lora_manager.py
+++ b/python/sglang/srt/lora/lora_manager.py
@@ -415,7 +415,7 @@ class LoRAManager:
         replace_submodule(self.base_model, module_name, lora_module)
         return lora_module
 
-    def should_skip_lora(self, module_name):
+    def should_skip_lora_for_vision_model(self, module_name):
         return module_name.find("vision_model.model") != -1
 
     def init_lora_modules(self):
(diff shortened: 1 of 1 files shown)


python/sglang/srt/lora/lora_manager.py
```diff
@@ -415,6 +415,9 @@ def set_lora_module(self, module_name, module):
         replace_submodule(self.base_model, module_name, lora_module)
         return lora_module
 
+    def should_skip_lora(self, module_name):
+        return module_name.find("vision_model.model") != -1
```
JustinTong0323, line 420: The names of visual modules can be diverse. For example, [qwen2.5vl](https://huggingface.co/Qwen/Qwen2.5-VL-3B-Instruct/blob/main/model.safetensors.index.json) is named `⁠visual.block`, and [internvl3.5](https://huggingface.co/OpenGVLab/InternVL3_5-8B/blob/main/model.safetensors.index.json) series is named ⁠`vision_model.encoder`. So perhaps a better way can be found to capture the content here.

python/sglang/srt/lora/lora_manager.py
```diff
@@ -415,6 +415,9 @@ def set_lora_module(self, module_name, module):
         replace_submodule(self.base_model, module_name, lora_module)
         return lora_module
 
+    def should_skip_lora(self, module_name):
```
JustinTong0323: function naming could be more specific, like `maybe_skip_lora_for_vision_model`

zhyncs:
```suggestion
    def should_skip_lora_for_vision_model(self, module_name):
```

python/sglang/srt/lora/lora_manager.py
```diff
@@ -432,6 +435,10 @@ def init_lora_modules(self):
             ) and not self.base_model.should_apply_lora(module_name):
                 continue
 
+            # Skip vision model
+            if self.should_skip_lora(module_name):
```
zhyncs:
```suggestion
            if self.should_skip_lora_for_vision_model(module_name):
```

python/sglang/srt/lora/lora_manager.py
```diff
@@ -415,6 +415,9 @@ def set_lora_module(self, module_name, module):
         replace_submodule(self.base_model, module_name, lora_module)
         return lora_module
 
+    def should_skip_lora_for_vision_model(self, module_name):
+        return module_name.find("vision_model.model") != -1
```
zhyncs, line 420:
```suggestion
        # TODO: support different vision models
        return module_name.find("vision_model.model") != -1
```

python/sglang/srt/lora/lora_manager.py
zhyncs: I think that this is only for Llama 4 for now. Let's cover more models in follow up.

zhyncs (approved):

zhyncs:
@gongwei-130 Please add unit test for this use case.

PR #10534, opened 2025-09-16
chore: Initial support for input config files

kushanam:
Reducing and organizing the input parameters for server commands

This PR adds support for YAML configuration files to SGLang server, allowing users to specify server arguments through config files instead of (or in addition to) command-line arguments. This feature follows the same pattern as vLLM's config file support.

python/sglang/srt/config_parser.py: New module containing the configuration argument parser
python/sglang/srt/server_args.py:
Added --config argument to the argument parser
test/srttest_config_integration.py: test suite

N/A

N/A


python/sglang/srt/config_parser.py
```diff
@@ -0,0 +1,190 @@
...
+        return args and not args[0].startswith("-")
+
+    def _merge_with_subcommand(
+        self, before_config: List[str], config_args: List[str], after_config: List[str]
+    ) -> List[str]:
+        """Merge config args when subcommand is present."""
+        subcommand = before_config[0]
+        remaining_args = before_config[1:]
+
+        if subcommand == "serve":
```
mickqian: Could you add some doc about the introduced serve command?

test/srt/test_config_integration.py
```diff
@@ -0,0 +1,143 @@
...
+@pytest.fixture
+def merger():
+    """Fixture providing a ConfigArgumentMerger instance."""
+    return ConfigArgumentMerger()
+
+
+def test_config_parser(merger):
+    """Test the config parser functionality."""
+    # Create a temporary config file
+    config_data = {
```
mickqian, line 25: Could you test this new feature with more complex inputs, e.g. combination with some boolean args ?

python/sglang/srt/config_parser.py
kushanam: removed the argument, it was intended to support OAI type of server run, but better be a subject for another PR.

test/srt/test_config_integration.py
kushanam: Added More test Arguments. Please check and see if it satisfies the requirements,

mickqian (approved):
Seems good, while it surely requires more test, but good for now. cc @zhyncs

kushanam:
gentle ping @zhyncs

python/sglang/srt/server_args_config_parser.py
```diff
@@ -0,0 +1,146 @@
+"""
+Configuration argument parser for command-line applications.
```
merrymercy, line 2: rename it to `server_config_parser.py` or `server_args_config_parser.py`

merrymercy (approved):

python/sglang/srt/server_args_config_parser.py
kushanam: Done.

Ouna-the-Dataweaver:
Just so you know - currently docs have that info on using --config, but if you install the latest 0.5.3rc0 using uv, as said in the same docs - this MR is not included there, and config parsing obviously isn't working.

PR #10541, opened 2025-09-17
fix: kv events with tp > 1

ishandhanani:
## Checklist


python/sglang/srt/mem_cache/kv_events.py
zhyncs, line 1: Please do not change the location as it may break the forks. Thank you.

ishandhanani: Got it - moving it back

python/sglang/srt/managers/scheduler.py
```diff
@@ -540,9 +540,12 @@ def __init__(
 
         # Init metrics stats
         self.init_metrics(tp_rank, pp_rank, dp_rank)
-        self.init_kv_events(server_args.kv_events_config)
         self.init_dp_balance(dp_balance_meta)
 
+        # Only emit KV events from tp rank 0
+        if tp_rank == 0:
```
sanshang-nv: Is using `self.enable_kv_cache_events` better? you already use condition `tp_rank == 0` when assigning value to it.

BTW, here is only considering Tensor Parallel. How about other parallelism (e.g., PP, EP, DP, etc.)? Or we only need to consider TP?

ishandhanani: 1. Fixed
2. At the moment we only consider events from tp rank 0

sanshang-nv:
LGTM

ShangmingCai (approved):
LGTM

PR #10550, opened 2025-09-17
Use jsonschema to constrain required or specific tool choice

TJ5:
## Motivation

EBNF constraints for tool choice are hard to read and extend. They must be individually created for each model's tool call format. Instead, we can use a much more easily extended json schema constraint. Json schema enables a much more simple constraint for required or named tool choices through passing constraints such as `minItems: 1`, which very easily enforces at least one tool call. The model output is constrained to json, so it can also bypass a parser for non-stream requests entirely. 

## Modifications

 - Created `get_json_schema_constraint`, a util function which outputs a json schema given the incoming tools and tool choice.
 - Created `json_array_parser`, which parses streaming output from requests for which tool_choice is set to 'required' or a named tool. 
 - Modified non-stream tool requests in `_process_tool_calls` to skip using a parser entirely and just extract the call information directly from the json output.
 - Added validations: Tools should specify a name and parameters. Parameters should be a valid json schema. If a tool choice is named to point to a specific tool, make sure that tool actually exists. We return a **400 BadRequestError** if any of these aren't true.
 - Updated the `_partial_json_loads()` function used by the base streaming parser. Since the incoming chunks for tool calls are an array, we have to also catch IndexErrors as well as JsonDecodeErrors to correctly load them and avoid a crash.
 - Note: for all json schemas, we use an **array schema**. If `tool_choice` is `required`, we specify `minItems: 1` so that at least one tool call is returned. If the tool_choice is a specific function, we specify `minItems: 1` and `maxItems: 1` so that exactly one tool call is returned, and it is the desired function per the tool choice.
 - Fix OpenAI protocl: `Function.name` should be required.

## Note on whitespace

Certain models (mistral, llama4) seem to generate excess whitespace in their tool calls with json schema. See: https://github.com/sgl-project/sglang/issues/8250 . Mistral in particular fails one of the existing test_tool_choice tests due to creating enough newline characters to hit the maximum token limit. Patching https://github.com/sgl-project/sglang/pull/8919 fixes this.

### Tool Choice tests
   - Updated `test_error_handling_invalid_tool_choice test` to expect a 400 BadReqestError if we try to choose an invalid function
   - Added `test_invalid_tool_missing_name`, `test_invalid_tool_missing_parameters`, and `test_invalid_json_schema_in_tool` to expect a 400 BadRequestError if the user-submitted tools are malformed. For example: 
   
```python
{
                "type": "function",
                "function": {
                    "name": "test_function",
                    "description": "Test function with invalid JSON schema",
                    "parameters": {
                        "type": "object",
                        "properties": {
                            "invalid_field": {
                                "type": "unknown_type",  # Invalid type
                                "description": "This field has an invalid type",
                            }
                        },
                        "required": ["invalid_field"],
                    },
                },
            } 
```
  This isn't a valid json schema and will return a 400
   - The existing tests, which test a correct response for required tool choice requests, named tool choice requests, streaming and non-streaming also cover the json-schema constrained generation.
 - **Json schema tests**
   - New test file test_json_schema_constraints
   - Generates a variety of json schema constraints and verifies that they are both valid and accurately constrain the model. 
   - Also compares with the old ebnf constraint
 - **Function Call Parser Tests**
   - Added a class of tests to the existing function call parser tests to test the json array parser, which is used for streaming required/named tool choice requests
   - Test with malformed json, empty inputs, whitespace handling, separators in the middle of and at the end of chunks, and with multiple tool calls
   - Examples:
      - whitespace: chunk1: `' \n\n [{"name": "get_weather", "parameters": '`, chunk2: `'{"location": "Tokyo"}}]'`
      - malformed json: `'[{"name": "get_weather", "parameters": {"location": "Tokyo"'`

### Other tests done

Tested on various models:
| Model             | Required | Named Tool | Required Multiple Calls | Named Streaming | Required Streaming | Required Multiple Streaming |
|-------------------|----------|------------|-------------------------|-----------------|--------------------|-----------------------------|
| Deepseek V3       | ✅        | ✅          | ✅                       | ✅               | ✅                  | ✅                           |
| Deepseek V3.1     | ✅        | ✅          | ✅                       | ✅               | ✅                  | ✅                           |
| Qwen3             | ✅        | ✅          | ✅                       | ✅               | ✅                  | ✅                           |
| Qwen3-Coder       | ✅        | ✅          | ✅                       | ✅               | ✅                  | ✅                           |
| Llama-3.1-8B      | ✅        | ✅          | ✅                       | ✅               | ✅                  | ✅                           |
| Llama-4-Scout     | ✅        | ✅          | ✅                       | ✅               | ✅                  | ✅                           |
| Llama-4-Maverick  | ✅        | ✅          | ✅                       | ✅               | ✅                  | ✅                           |
| Mistral-v0.3      | ✅        | ✅          | ⚠️ (whitespace)          | ✅               | ✅                  | ⚠️ (whitespace)              |
| GPT-OSS-120B      | ✅        | ✅          | ✅                       | ✅               | ✅                  | ✅                           |

#### Deepseek-v3 Know Issue
 - Attempted to reproduce the tool calling bug reported for deepseekv3 here: https://github.com/sgl-project/sglang/issues/9311
 - Used the reporter's script. 
 
Output: (top: this PR, bottom: **main brach**) 
 - **Behavior:** Generated output contains two tool calls
 - The correct response has three tool calls, not two. However, the initial bug reported where the tool calls were malformed with empty URL fields is not present. 
 - The current main branch only returned one tool call.

Screenshot 2025-09-25 at 4 59 19 PM
Screenshot 2025-09-25 at 4 59 40 PM

#### gpt-oss-120b Test
 - Also tested on gpt-oss-120b using the same script above. We expect 3 tool calls, get only one though. This is not an issue to do with this PR however as the same output is given if tool call is optional, using the old gpt-oss detector.

Output (for both tool_choice: required and auto): 
Screenshot 2025-09-25 at 5 01 27 PM

## Checklist


python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -952,13 +1030,31 @@ async def _process_tool_call_stream(
...
+                    tool_call_parser=self.tool_call_parser,
+                )
+
         parser = parser_dict[index]
 
-        normal_text, calls = parser.parse_stream_chunk(delta)
+        # Handle both FunctionCallParser and JsonArrayParser
+        if isinstance(parser, JsonArrayParser):
+            result = parser.parse_streaming_increment(delta, request.tools)
+            normal_text, calls = result.normal_text, result.calls
```
CatherineSue, line 1089: Does python not support parsing NamedTuple by default?

python/sglang/srt/function_call/json_array_parser.py
```diff
@@ -0,0 +1,139 @@
...
+            full_json = self._buffer + json_part
+            if full_json.startswith(self.bot_token):
+                full_json = full_json[len(self.bot_token) :]
+            elif full_json.startswith(self.tool_call_separator):
+                full_json = full_json[len(self.tool_call_separator) :]
+            if full_json.endswith(self.eot_token):
+                full_json = full_json[: -len(self.eot_token)]
+
+            try:
+                json.loads(full_json)
```
CatherineSue: A bit confused with this block.

Is this still incremental streaming?

python/sglang/srt/function_call/json_array_parser.py
```diff
@@ -0,0 +1,139 @@
...
+    ) -> StreamingParseResult:
+        """
+        Parse JSON array tool calls with improved separator detection.
+        Uses brace counting to find valid separators that are outside JSON objects.
+        """
+        # If we previously withheld a valid tool-call separator, prepend it
+        # when we see the next JSON object start
+        if self._pending_separator:
+            # Check if the text (after removing leading whitespace) starts with '{'
+            if new_text.lstrip().startswith("{"):
```
CatherineSue: I don't think this necessarily means the start of a tool call chunk?

python/sglang/srt/entrypoints/openai/serving_chat.py
TJ5: It does but in this case parse_streaming_increment is returning a `StreamingParseResult` object, not a tuple

test/srt/openai_server/basic/test_serving_chat.py
```diff
@@ -354,11 +354,12 @@ def test_kimi_k2_non_streaming_tool_call_id_format(self):
                 {"type": "function", "function": {"name": "get_weather"}},
             ]
 
-            tool_calls, remaining_text, _ = self.chat._process_tool_calls(
+            result = self.chat._process_tool_calls(
```
CatherineSue: do we need this change? same as above, I thought puython can do parsing for the named tuple

python/sglang/srt/function_call/json_array_parser.py
CatherineSue: What if I have a json string like `{"key": "value with } brace"}`?

test/srt/function_call/test_json_schema_constraint.py
```diff
@@ -0,0 +1,318 @@
+"""
```
CatherineSue, line 1: Please add this to run_suite.py

test/srt/openai_server/function_call/test_json_array_parser_integration.py
```diff
@@ -0,0 +1,397 @@
+"""
+Integration tests for JsonArrayParser with tool_choice functionality
```
CatherineSue: Should this be merged with `test_tool_choice.py`?

python/sglang/srt/function_call/json_array_parser.py
```diff
@@ -0,0 +1,123 @@
...
+        """
+        Build an EBNF grammar for constrained generation.
+        This is not used for JSON schema constraints as they are handled
+        by the constraint backends directly.
+        """
+        raise NotImplementedError(
+            "EBNF generation is not supported for JSON schema constraints."
+        )
+
+    def parse_streaming_increment(
```
CatherineSue, line 49: Overall, the state management of pending_separator and brace count is convolutated. 

A few test cases I can think of that could lead to error:

1. chunks: 
``` 
[
      '{"name": "test", "param": "has } inside"}',
      ',',
      '{"name": "next"}'
]
```

2. chunks:
```
[
      '{"name": "foo", "parameters": {}}',
      ',{"name": "bar"}'
  ]
```

3. chunks:
```
[
      '{"name": "foo"}',
      ',',
      '{"name": "bar"}'
  ]
```

4. chunks:
```
[
      '{"name": "test", "parameters": {}',
      '},{"name": "next"}'
  ]
```

python/sglang/srt/function_call/json_array_parser.py
TJ5: If I'm understanding correctly, you're pointing out that an open brace could be the start of an unrelated json string that's not a tool call. This isn't a problem for a few reasons. First, this if statement is nested inside a check for self._pending_separator, which requires both that a preceding chunk terminated with a tool call separator and that the next chunk starts with brace_count=0, meaning that we are no longer within any currently streaming tool call.

If we pass to the base class: {tool call 1} [ separator], the fact that we have a full, valid json object followed by in this case a comma causes the partial json loader to crash. So if the full text is {tool call 1} [ separator ] {unrelated json object}, we should still be moving the separator to the start of the unrelated json object to prevent a crash. The overall content doesn't change, we just move the separator to the next chunk.

python/sglang/srt/function_call/json_array_parser.py
TJ5: Rewrote this whole section to just use the base class's parser without changes. Modified _partial_json_loads() to get around issues with whitespace and with the ']' character terminating certain chunks. All of these cases have been added to test_function_call_parser as well, and all pass.

python/sglang/srt/function_call/json_array_parser.py
TJ5: Repeat of other comment, but have overhauled this whole section so we are just using the base parser now. Modified _partial_json_loads() to get around previous issues with whitespace and with chunks terminating with the ']' character.

python/sglang/srt/function_call/utils.py
```diff
@@ -1,10 +1,21 @@
...
-from typing import Any, Tuple
+from json.decoder import WHITESPACE
+from typing import Any, List, Literal, Optional, Tuple, Union
 
 import partial_json_parser
 from partial_json_parser.core.options import Allow
 
+from sglang.srt.entrypoints.openai.protocol import Tool, ToolChoice
+
+_WS = " \t\r\n"
```
CatherineSue: Do we still need this function?

python/sglang/srt/function_call/function_call_parser.py
```diff
@@ -177,9 +178,13 @@ def get_structure_constraint(
         ):
             strict_tag = self.get_structure_tag()
             return ("structural_tag", strict_tag)
-        elif tool_choice == "required" or isinstance(tool_choice, ToolChoice):
-            ebnf = self.get_ebnf(tool_choice)
-            return ("ebnf", ebnf) if ebnf is not None else None
+        elif (
+            tool_choice == "required"
+            or isinstance(tool_choice, ToolChoice)
+            or (isinstance(tool_choice, dict) and "function" in tool_choice)
```
CatherineSue: Do we need to check dict instance?

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -74,6 +78,27 @@ def _validate_request(self, request: ChatCompletionRequest) -> Optional[str]:
...
+        # Validate tool definitions
+        if request.tools:
+            for i, tool in enumerate(request.tools):
+                if not hasattr(tool, "function") or not tool.function:
+                    return f"Tool {i} is missing required 'function' field."
+                if not hasattr(tool.function, "name") or not tool.function.name:
+                    return f"Tool {i} function is missing required 'name' field."
+                if (
+                    not hasattr(tool.function, "parameters")
+                    or tool.function.parameters is None
```
CatherineSue: Why could parameters not be None?

TJ5: Updated this: parameters should always be a valid json schema, so we use jsonschema.Draft202012Validator to check that it is. Otherwise return a 400

python/sglang/srt/function_call/function_call_parser.py
TJ5: We don't, removed.

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -188,6 +217,19 @@ def _process_messages(
                 tool_call_constraint = parser.get_structure_constraint(
                     request.tool_choice
                 )
+            # Handle JSON schema constraint directly for required or named tool choice
+            if (
+                request.tool_choice == "required"
+                or isinstance(request.tool_choice, ToolChoice)
+                or (
+                    isinstance(request.tool_choice, dict)
```
CatherineSue: Is dict needed here?

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -952,13 +1034,31 @@ async def _process_tool_call_stream(
...
+            # Use JSON detector directly for required or named tool choice
+            if (
+                request.tool_choice == "required"
+                or isinstance(request.tool_choice, ToolChoice)
+                or (
+                    isinstance(request.tool_choice, dict)
+                    and "function" in request.tool_choice
+                )
+            ):
+                # Store JsonArrayParser directly for JSON schema constraints
```
CatherineSue: please remove this comment as it is obivious

python/sglang/srt/function_call/utils.py
```diff
@@ -50,3 +55,83 @@ def _is_complete_json(input_str: str) -> bool:
...
+            all_defs = {}
+            for tool in tools:
+                if tool.function.parameters is None:
+                    continue
+                # Make a copy to avoid modifying original
+                params = tool.function.parameters.copy()
+                defs = params.pop("$defs", {})
+                for def_name, def_schema in defs.items():
+                    if def_name in all_defs and all_defs[def_name] != def_schema:
+                        raise ValueError(
```
CatherineSue: Who is responsible for catching this ValueError? Based on the code flow, this ValueError will be captured here: https://github.com/sgl-project/sglang/blob/c3faf2d6e673650b275bbefad9070706efddf3d3/python/sglang/srt/entrypoints/openai/serving_base.py#L65-L71

This is a 500 instead of 400. Is that intended?

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -74,6 +79,30 @@ def _validate_request(self, request: ChatCompletionRequest) -> Optional[str]:
...
+                return "Tools cannot be empty if tool choice is set to a specific tool."
+            tool_name = request.tool_choice.function.name
+            tool_exists = any(tool.function.name == tool_name for tool in request.tools)
+            if not tool_exists:
+                return f"Tool '{tool_name}' not found in tools list."
+
+        # Validate tool definitions
+        if request.tools:
+            for i, tool in enumerate(request.tools):
+                if not hasattr(tool, "function") or not tool.function:
```
CatherineSue: why would tool not have function parameter? this is a pydantic model, right?

python/sglang/srt/entrypoints/openai/serving_chat.py
CatherineSue: why is parameter should always be a valid json schema?

CatherineSue: If we want to enforce this, better to make it the same as OpenAI's error message: 

Image

This is making the CI failed: https://github.com/sgl-project/sglang/actions/runs/17986382981/job/51165645456?pr=10550

Based on the CI error, the error message handling could be refined
Screenshot 2025-09-24 at 12 20 47 PM

Please fix the CI.

CatherineSue:
It seems this test is failing:
https://github.com/sgl-project/sglang/actions/runs/17986382981/job/51165645440?pr=10550

Screenshot 2025-09-24 at 12 26 20 PM

Could you take a look?

python/sglang/srt/entrypoints/openai/serving_chat.py
TJ5: I'll fix the error message. Answering your earlier question, I saw it should be a json schema from openai's documentation on function calling:
image
https://platform.openai.com/docs/guides/function-calling

CatherineSue: Can you also check if we need to have `hasattr(tool.function, "parameters")`?
Because tool is a pydantic model, pydantic auto validates the fields of the passed request.

TJ5: We don't need to check if we need a "function" field (I haven't pushed this change yet to get rid of the check, as of writing this) but pydantic doesn't catch if the parameters field doesn't exist. With that said, the Draft Schema validator also covers this case, so we don't need to check with `hasattr(tool.functino, "parameters")`

TJ5:
Regarding the failing test in test_tool_choice, this is the whitespace issue encountered with Mistral mentioned in the PR Description. The actual model output for the failing test (test_multi_tool_scenario_required) is `'[ {\n  "name": "get_weather",\n  "parameters": {"location": "Tokyo"}\n}\n\n\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t\t...` The tab characters extend to the maximum token length, which causes the test to fail since the tool call JSON output does not properly terminate. This is fixed by patching https://github.com/sgl-project/sglang/pull/8919

python/sglang/srt/entrypoints/openai/serving_chat.py
TJ5: The test fails because it had a json schema that used `"type": "int"` instead of `"type": "integer"`, which is illegal for json schema. I modified the test to fix it.

Swipe4057:
Hi! Thanks for this awesome merge request! If it's not too much trouble, could you point me to which function or code is most likely to have bugs in gpt-oss function calling? (I'm new to the sglang codebase)

I wrote about the issue here:
https://github.com/sgl-project/sglang/issues/10738

I'm trying to fix it, but haven't been able to isolate the problem yet, and the current gpt-oss tests completely miss function calling bugs.

python/sglang/srt/function_call/utils.py
TJ5: It's not. What I did to change this was I moved the validation to its own util function and call that with the rest of the validation logic in serving_chat.py.

python/sglang/srt/entrypoints/openai/serving_chat.py
CatherineSue: Do we need the check here?

https://github.com/sgl-project/sglang/pull/10550/files#diff-20a5bbe903ddc72a708cae54cfc9bd1ba42f3693a90dab09fc10904329bb6377R95-R96

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -74,6 +82,34 @@ def _validate_request(self, request: ChatCompletionRequest) -> Optional[str]:
...
+                    return (
+                        f"Tool {i} function has invalid 'parameters' schema: {str(e)}"
+                    )
+
+            # Check for conflicting tool definitions when tool_choice is required or specific
+            if request.tool_choice == "required" or isinstance(
+                request.tool_choice, ToolChoice
+            ):
+                try:
+                    validate_tool_definitions(request.tools)
```
CatherineSue: qq: this function is called for both required and ToolChoice here. But it is only called for required in utils.py. Is it intended?

python/sglang/srt/function_call/utils.py
```diff
@@ -50,3 +55,98 @@ def _is_complete_json(input_str: str) -> bool:
...
+    all_defs = {}
+    for tool in tools:
+        if tool.function.parameters is None:
+            continue
+        # Make a copy to avoid modifying original
+        params = tool.function.parameters.copy()
+        defs = params.pop("$defs", {})
+        for def_name, def_schema in defs.items():
+            if def_name in all_defs and all_defs[def_name] != def_schema:
+                raise ValueError(
```
CatherineSue, line 80: Could we explore if there is a way to make this more concise? Right now this function is almost the same as `get_tool_schema_defs`. Just without ValueError raise.

python/sglang/srt/entrypoints/openai/serving_chat.py
CatherineSue: Could we move this check to https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/entrypoints/openai/protocol.py#L395 ?

CatherineSue:
Regarding the test_multi_tool_scenario_required for mistral, maybe mark it as optional or flaky for now.

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -75,6 +83,33 @@ def _validate_request(self, request: ChatCompletionRequest) -> Optional[str]:
...
+                try:
+                    Draft202012Validator.check_schema(tool.function.parameters)
+                except SchemaError as e:
+                    return (
+                        f"Tool {i} function has invalid 'parameters' schema: {str(e)}"
+                    )
+
+            # Check for conflicting tool definitions when tool_choice is required.
+            # Specific tool requests only have one tool so conflicts are not possible
+            if request.tool_choice == "required":
```
CatherineSue: Let's remove this here. Revert the function back to `get_validate_tool_schema_defs`. And add a try-catch `ValueError` in serving_base.py instead. 

https://github.com/sgl-project/sglang/blob/e56c64bfaf341aafe927c43f292f11f99f746ba8/python/sglang/srt/entrypoints/openai/serving_base.py#L41-L71

python/sglang/srt/function_call/utils.py
```diff
@@ -50,3 +55,90 @@ def _is_complete_json(input_str: str) -> bool:
         return True
     except JSONDecodeError:
         return False
+
+
+def get_and_validate_tool_schema_defs(tools: List[Tool], validate: bool) -> dict:
```
CatherineSue: See comment above. Let's revert this back to `get_tool_schema_defs`. And we only call the function at L139 below.

python/sglang/srt/function_call/utils.py
```diff
@@ -50,3 +55,90 @@ def _is_complete_json(input_str: str) -> bool:
...
+    Get the JSON schema constraint for the specified tool choice.
+
+    Args:
+        tool_choice: The tool choice specification
+
+    Returns:
+        JSON schema dict, or None if no valid tools found
+    """
+
+    def get_tool_schema(tool):
```
CatherineSue: Can we move this function out? Basically no nested functions.

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -867,14 +918,59 @@ def _process_tool_calls(
...
+                    )
+                return ToolCallProcessingResult(tool_calls, "", finish_reason)
+            except json.JSONDecodeError as e:
+                logger.error(f"Tool call parsing error: {e}")
+                return ToolCallProcessingResult(None, text, finish_reason)
+
+        # Use parser since output is not constrained by JSON schema
         parser = FunctionCallParser(tools, self.tool_call_parser)
         if parser.has_tool_call(text):
+            # Set finish reason to tool_calls since we detected tool calls
```
CatherineSue: let's remove this.

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -867,14 +918,59 @@ def _process_tool_calls(
...
             if finish_reason["type"] == "stop":
                 finish_reason["type"] = "tool_calls"
                 finish_reason["matched"] = None
+
```
CatherineSue: remove this line as well. Keep minimal changes

python/sglang/srt/entrypoints/openai/serving_base.py
```diff
@@ -62,6 +62,21 @@ async def handle_request(
             return self.create_error_response(
                 message=e.detail, err_type=str(e.status_code), status_code=e.status_code
             )
+        except ValueError as e:
+            # Check if this is a tool schema conflict error
```
CatherineSue: ValueError can be treated as 400 in general.

Let's remove the check and just do:

```
                return self.create_error_response(
                    message=str(e),
                    err_type="BadRequest",
                    status_code=400,
                )
```

CatherineSue:
Overall LGTM! Thanks for the great work. Let's wait for the CI results. Then I will approve.

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -75,6 +80,23 @@ def _validate_request(self, request: ChatCompletionRequest) -> Optional[str]:
...
 
+        if request.tool_choice is not None and not isinstance(request.tool_choice, str):
+            if not request.tools:
+                return "Tools cannot be empty if tool choice is set to a specific tool."
+            tool_name = request.tool_choice.function.name
+            tool_exists = any(tool.function.name == tool_name for tool in request.tools)
+            if not tool_exists:
+                return f"Tool '{tool_name}' not found in tools list."
+
+        # Validate tool definitions
```
CatherineSue, line 91: This might reduce the indentations

```suggestion
for i, tool in enumerate(request.tools or []):
    if tool.function.parameters is None:
        continue
    try:
        Draft202012Validator.check_schema(tool.function.parameters)
    except SchemaError as e:
        return f"Tool {i} function has invalid 'parameters' schema: {e}"
```

CatherineSue: Note: we can't commit directly as I didn't wrap all the lines.

CatherineSue:
unit-test-backend-1-gpu(5) is failing due to HF 403: https://github.com/sgl-project/sglang/actions/runs/18052643506/job/51379091807?pr=10550

CatherineSue (approved):

JustinTong0323:
Note: The only not passed CI is failed due to an error on main, fix in #10977

slin1237 (approved):

YosanHo:
not work for deepseek-v3.2 with tool_choice='required'

request:
```json
{
    "messages": [
        {
            "role": "system",
            "content": "You are a research assistant. Use tools to gather information systematically."
        },
        {
            "role": "user",
            "content": "Find detailed information about the best electric mountain bikes in 2025"
        },
        {
            "role": "assistant",
            "tool_calls": [
                {
                    "id": "call_search123",
                    "type": "function",
                    "function": {
                        "name": "web_search",
                        "arguments": "{\"query\": \"best electric mountain bikes 2025\"}"
                    }
                }
            ]
        },
        {
            "role": "tool",
            "tool_call_id": "call_search123",
            "content": "[{\"title\":\"Best Electric Bikes of 2025\",\"url\":\"https://electricbikereview.com/best-electric-bikes/\",\"description\":\"We have tested over 100 e-bikes this year\"},{\"title\":\"The 5 Best Electric Mountain Bikes of 2025\",\"url\":\"https://www.outdoorgearlab.com/topics/biking/best-electric-mountain-bike\",\"description\":\"Expert testing of top e-MTBs\"},{\"title\":\"2025 E-MTB Guide\",\"url\":\"https://bikeradar.com/best-electric-mountain-bikes/\",\"description\":\"Complete buyers guide\"}]"
        }
    ],
    "model": "DeepSeek-V3.2",
    "temperature": 0.3,
    "stream": false,
    "tool_choice": "required",
    "tools": [
        {
            "type": "function",
            "function": {
                "name": "web_search",
                "description": "Search the web for information",
                "parameters": {
                    "type": "object",
                    "properties": {
                        "query": {"type": "string", "description": "Search query"}
                    },
                    "required": ["query"],
                    "additionalProperties": false
                }
            }
        },
        {
            "type": "function",
            "function": {
                "name": "scrape_url",
                "description": "Scrape content from a URL",
                "parameters": {
                    "type": "object",
                    "properties": {
                        "url": {"type": "string", "description": "URL to scrape"}
                    },
                    "required": ["url"],
                    "additionalProperties": false
                }
            }
        }
    ]
}
```

response with tool_choice='required'
```json
{
  "id": "3d3e4117f6024c369ec97c5b95e6458a",
  "object": "chat.completion",
  "created": 1760168470,
  "model": "DeepSeek-V3.2",
  "choices": [
    {
      "index": 0,
      "message": {
        "role": "assistant",
        "content": "Let me search for more specific information about electric mountain bikes in 2025:<｜tool▁calls▁begin｜><｜tool▁call▁begin｜>web_search<｜tool▁sep｜>{\"query\": \"\\\"2025 electric mountain bike\\\" reviews comparison specs\"}<｜tool▁call▁end｜><｜tool▁calls▁end｜>",
        "reasoning_content": null,
        "tool_calls": null
      },
      "logprobs": null,
      "finish_reason": "tool_calls",
      "matched_stop": null
    }
  ],
  "usage": {
    "prompt_tokens": 391,
    "total_tokens": 430,
    "completion_tokens": 39,
    "prompt_tokens_details": null,
    "reasoning_tokens": 0
  },
  "metadata": {
    "weight_version": "default"
  }
}
```

response with tool_choice='auto'
```json
{
  "id": "52430dc469364a1b931181e4f559bdf7",
  "object": "chat.completion",
  "created": 1760168563,
  "model": "DeepSeek-V3.2",
  "choices": [
    {
      "index": 0,
      "message": {
        "role": "assistant",
        "content": "Let me search for more specific information about electric mountain bikes in 2025:",
        "reasoning_content": null,
        "tool_calls": [
          {
            "id": "call_04c10242c55e422ea6833f9f",
            "index": 0,
            "type": "function",
            "function": {
              "name": "web_search",
              "arguments": "{\"query\": \"\\\"2025 electric mountain bike\\\" reviews comparison specs\"}"
            }
          }
        ]
      },
      "logprobs": null,
      "finish_reason": "tool_calls",
      "matched_stop": null
    }
  ],
  "usage": {
    "prompt_tokens": 391,
    "total_tokens": 430,
    "completion_tokens": 39,
    "prompt_tokens_details": null,
    "reasoning_tokens": 0
  },
  "metadata": {
    "weight_version": "default"
  }
}
```

PR #10555, opened 2025-09-17
prepare for sglang+verl

lbk-sys:
## Motivation
run sglang for verl

At that time, the NPU commercial law version was unable to support the adaptation of the current UUID scheme, so the change was temporarily avoided. It is expected that a replacement scheme will be available by the end of October.

## Modifications

At present, this PR is used to integrate verl+sglang+npu capabilities, while other modifications are incorporated into sgl kernel npu.

## Checklist


python/sglang/srt/torch_memory_saver_adapter.py
```diff
@@ -3,11 +3,16 @@
...
 from contextlib import contextmanager, nullcontext
+from sglang.srt.utils import (
+    is_npu,
+)
 
 try:
     import torch_memory_saver
 
     _memory_saver = torch_memory_saver.torch_memory_saver
+    if is_npu():
```
ping1jing2: `_is_npu = is_npu()`

PR #10566, opened 2025-09-17
feat: Add Non-intrusive Tensor Dumping for Model Inference

guoyuhong:
## Motivation

Inference engines often require the ability to compare intermediate results during the inference process. Currently, the existing dumping functionality (`sglang/srt/debug_utils/dumper.py`) requires intrusive modifications to model code — for example, `sglang/srt/models/grok.py` contains numerous model-specific calls to `dump_to_file`. This approach is not user-friendly, as it forces users to modify model code every time they want to debug. A non-intrusive solution for tensor dumping is highly desirable.

## Modifications

This PR introduces a new file `tensor_dump_forward_hook.py`, which leverages PyTorch's `register_forward_hook` mechanism. After the model is loaded, calling `register_forward_hook_for_model` once will attach tensor-dumping hooks to all layers of the model. During subsequent inference debugging, the outputs of all forward passes can be recorded and saved to files.

Usage:
Use the `--debug-tensor-dump-output-folder` argument to specify the output directory for dumped tensors.

- Each GPU rank generates its own subdirectory named: `f"TP{tp_rank}_PP{pp_rank}_Rank{rank}_pid{pid}"`.

- Each complete forward pass of the model produces a `.pt` file named: `f"Pass{pass_num}.pt"`.
These `.pt` files can be loaded via `torch.load()` and contain a dictionary of key-value pairs. The keys correspond to operator/module names in the model (similar to those in `model.safetensors.index.json`), and the values are the output tensors from the respective operators.

## Accuracy Tests

- Added unit test: test_tensor_dump_forward_hook.py.
    image

- Tested successfully with the Qwen model. Example results are as follows:
    - `python -m sglang.launch_server --model-path Qwen/Qwen2-7B-Instruct --port 8000 --host 0.0.0.0  --debug-tensor-dump-output-folder /home/admin/sglang_dump --tp-size 2 `: `--debug-tensor-dump-output-folder` will set `disable_cuda_graph` & `skip_server_warmup`.
    - make a request to the server:
    image

    - tensors are dumped to the target directory:
    image

    - 18 tokens have 18 files:
    image

    - load these file and do the work you need:
    image

## Checklist

### DIFF
--- a/python/sglang/srt/debug_utils/tensor_dump_forward_hook.py
+++ b/python/sglang/srt/debug_utils/tensor_dump_forward_hook.py
@@ -1,0 +1,149 @@
+"""
+This file provides a function `register_forward_hook_for_model` that registers a forward hook on every operator of the model.
+After registration, during model inference, all tensors generated throughout the forward pass will be recorded.
+
+Usage:
+Specify the output directory for dumping tensors using the argument `--debug-tensor-dump-output-folder`.
+A separate directory will be created for each GPU rank, named in the format `f"TP{tp_rank}_PP{pp_rank}_Rank{rank}_pid{pid}"`.
+Each complete forward pass of the model generates a `.pt` file named `f"Pass{pass_num}.pt"`, which can be loaded using `torch.load`.
+The file contains a series of key-value pairs, where the keys correspond to operator names in the model
+(similar to those in model.safetensors.index.json), and the values are the outputs produced by the respective operators.
+"""
+
+import logging
+import os
+from pathlib import Path
+
+import torch
+
+from sglang.srt.layers.logits_processor import LogitsProcessorOutput
+from sglang.srt.model_executor.forward_batch_info import ForwardBatch, PPProxyTensors
+
+logger = logging.getLogger(__name__)
+
+
+class TensorDumper:
+    def __init__(
+        self, dump_dir: str, dump_layers: int, tp_size: int, tp_rank: int, pp_rank: int
+    ):
+        self._dump_layers = dump_layers
+        self._forward_pass_id = 0
+        self._pid = os.getpid()
+        self._current_tensors = {}
+        self._base_dir = Path(dump_dir)
+        rank = tp_size * pp_rank + tp_rank
+        self._process_dir = (
+            self._base_dir / f"TP{tp_rank}_PP{pp_rank}_Rank{rank}_pid{self._pid}"
+        )
+        self._process_dir.mkdir(parents=True, exist_ok=True)
+
+    def get_dump_dir(self):
+        return str(self._process_dir)
+
+    def add_tensor(self, name, tensor_item):
+        if isinstance(tensor_item, (tuple, list)):
+            tensors = [t.cpu() for t in tensor_item if t is not None]
+            if len(tensors) == 1:
+                self._current_tensors[name] = tensors[0]
+            else:
+                self._current_tensors[name] = tensors
+        elif isinstance(tensor_item, torch.Tensor):
+            self._current_tensors[name] = tensor_item.cpu()
+        elif isinstance(tensor_item, LogitsProcessorOutput):
+            self._current_tensors[name] = tensor_item.next_token_logits.cpu()
+        elif isinstance(tensor_item, ForwardBatch):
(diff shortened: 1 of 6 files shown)


python/sglang/srt/debug_utils/tensor_dump_forward_hook.py
```diff
@@ -0,0 +1,118 @@
...
+            f"Dump {self._forward_pass_id:05d}th pass to {tensor_file_for_pass}"
+        )
+        torch.save(self._current_tensors, str(tensor_file_for_pass))
+        self._current_tensors = {}
+        self._forward_pass_id += 1
+
+    def _add_hook_recursive(self, model, prefix, dump_module_name="model"):
+        has_dump_module = False
+        for name, module in model._modules.items():
+            do_dump = False
```
BBuf: Can `do_dump` be renamed to `top_level_model` ?

BBuf (approved):
qq: We can usually detect numerical bugs by checking the first few layers during model inference. Could we add an optional parameter to control the depth when registering forward hooks, allowing users to specify the number of layers to monitor?

python/sglang/srt/debug_utils/tensor_dump_forward_hook.py
guoyuhong: Done

guoyuhong:
> qq: We can usually detect numerical bugs by checking the first few layers during model inference. Could we add an optional parameter to control the depth when registering forward hooks, allowing users to specify the number of layers to monitor?

I've added a new server argument `--debug-tensor-dump-layers` to allow users to specify which layers should be dumped. Most models follow a structure similar to this:
```text
XxxxForCausalLM
     (model): XxxxModel
          (layers): ModuleList
```
With this layout, the default implementation works out of the box for the majority of models. For models with a different module hierarchy, users can customize the behavior by setting the environment variables `TENSOR_DUMP_TOP_LEVEL_MODULE_NAME` and `TENSOR_DUMP_LAYERS_MODULE_NAME` to match their model's structure.


Here is the example with `--debug-tensor-dump-layers=4`:
1. Dump files has smaller size:
image

2. Only required layers are dumped:
image

guoyuhong:
@zhyncs do you have any more comments?

fzyzcjy (approved):

test/srt/debug_utils/test_tensor_dump_forward_hook.py
```diff
@@ -0,0 +1,87 @@
...
+
+    dir_path = dumper.get_dump_dir()
+    inp = torch.randn(4, TEST_HIDDEN_SIZE, dtype=torch.bfloat16) * 0.01
+    result = model(inp.cuda())
+    data = torch.load(f"{dir_path}/Pass00000.pt")
+    assert "model.layernorm" in data
+    assert "model.mlp.down_proj" in data
+    assert torch.allclose(
+        data["model.mlp.down_proj"], result.cpu(), rtol=1e-5, atol=1e-5
+    )
```
ispobock, line 89: It seems no entry point?
ref: https://github.com/sgl-project/sglang/blob/42f8ea4030947bcb5431464370492daee6735e3d/test/srt/test_abort.py#L114

guoyuhong: Thank you for pointing out this issue. I saw that the CI already had this case and forgot to follow the convention.

python/sglang/srt/server_args.py
```diff
@@ -3365,6 +3374,12 @@ def add_cli_args(parser: argparse.ArgumentParser):
             default=ServerArgs.debug_tensor_dump_output_folder,
             help="The output folder for dumping tensors.",
         )
+        parser.add_argument(
+            "--debug-tensor-dump-layers",
+            type=int,
+            default=-1,
+            help="The layer number for dumping tensors.",
+        )
```
merrymercy, line 3382: can you make it a list?
```
debug_tensor_dump_layers: Optional[List[int]] = None
```

```
        parser.add_argument(
            "--debug-tensor-dump-layers",
            type=int,
            nargs="+",
            help="The layer ids to dump. Dump all layers if not specified.",
        )
```

PR #10576, opened 2025-09-17
fix: support GPT OSS tool calls ending with `<|return|>`

gaelmuller:
## Motivation

When using gpt-oss-120b with SGLang with the Chat Completion API and tools, I often get an empty response containing only `reasoning_content` (no `content` nor `tool_calls`).

When looking at what is returned by the model, the tool call is followed by the `<|return|>` end token rather than  expected `<|call|>`.

When parsing this output with OpenAI's harmony library, this is parsed as a tool call so I think this should be supported.

## Modifications

* Keep all end tokens (not only `<|call|>`) since they are used by `HarmonyParser`
* Use a tool regex rather than the `<|call|>` end token in `HarmonyParser` to determine if an `Event` should be considered as a tool call or as text
* Update the tool call regex to accept `<|return|>` as a valid end token for tool calls

## Checklist

## Additional information

The [tool parser](https://docs.sglang.ai/advanced_features/tool_parser.html) documentation states that:

> The gpt-oss tool parser filters out analysis channel events and only preserves normal text. This can cause the content to be empty when explanations are in the analysis channel. To work around this, complete the tool round by returning tool results as role="tool" messages, which enables the model to generate the final content.

I did not understand this comment so I'm unclear how this is related.

### DIFF
--- a/python/sglang/srt/function_call/gpt_oss_detector.py
+++ b/python/sglang/srt/function_call/gpt_oss_detector.py
@@ -32,7 +32,7 @@ class GptOssDetector(BaseFormatDetector):
 
         # Pattern to extract function name and JSON from tool_call event content
         self.tool_extract_pattern = re.compile(
-            r"to=([a-zA-Z_][a-zA-Z0-9_.-]*)\s*<\|constrain\|>json<\|message\|>(.*?)(?:<\|call\|>|$)",
+            r"to=([a-zA-Z_][a-zA-Z0-9_.-]*)\s*<\|constrain\|>json<\|message\|>(.*?)(?:<\|call\|>|<\|return\|>|$)",
             re.DOTALL,
         )
 
--- a/python/sglang/srt/function_call/gpt_oss_detector.py
+++ b/python/sglang/srt/function_call/gpt_oss_detector.py
@@ -32,7 +32,7 @@ class GptOssDetector(BaseFormatDetector):
 
         # Pattern to extract function name and JSON from tool_call event content
         self.tool_extract_pattern = re.compile(
-            r"to=([a-zA-Z_][a-zA-Z0-9_.-]*)\s*<\|constrain\|>json<\|message\|>(.*?)(?:<\|call\|>|<\|return\|>|$)",
+            r"to=([a-zA-Z_][a-zA-Z0-9_.-]*)\s*(?:<\|constrain\|>json)?<\|message\|>(.*?)(?:<\|call\|>|<\|return\|>|$)",
             re.DOTALL,
         )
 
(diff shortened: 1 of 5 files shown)


Swipe4057:
Is anyone using codex https://github.com/openai/codex? Currently, I'm having problems calling functions and seeing no responses in the interface when inferencing oss20b with sglang. I'll try to figure out the cause of this bug.

Please write when mr is ready, and I will test it!

gaelmuller:
> Is anyone using codex https://github.com/openai/codex? Currently, I'm having problems calling functions and seeing no responses in the interface when inferencing oss20b with sglang. I'll try to figure out the cause of this bug.
> 
> Please write when mr is ready, and I will test it!

I am using custom agent logic rather than Codex but FYI the PR is ready, I'm already using it to fix my tool call issues with gpt-oss-120b.

Swipe4057:
> > Is anyone using codex https://github.com/openai/codex? Currently, I'm having problems calling functions and seeing no responses in the interface when inferencing oss20b with sglang. I'll try to figure out the cause of this bug.
> > Please write when mr is ready, and I will test it!
> 
> I am using custom agent logic rather than Codex but FYI the PR is ready, I'm already using it to fix my tool call issues with gpt-oss-120b.

With the 20b model and your mr, I get these responses in the content field. Is this not the case with you?
image

gaelmuller:
> With the 20b model and your mr, I get these responses in the content field. Is this not the case with you? image

Seems like this should be extracted correctly. Are you setting the parser with `--tool-call-parser gpt-oss` ?

Swipe4057:
Yes, I specify --tool-call-parser gpt-oss. And similarly with the reasoning parser.
I took your scripts and replaced them in the official Docker image at the path /sglang_workspace/sglang/python/sglang/srt/.
After the changes it got better, but here's another debugging example:
image
(I set the task of studying the files in the folder and writing a readme file, Continue + VS Code)

Swipe4057:
There seems to be a problem with the final call token.
image

Swipe4057:
Apparently, there's a bug with removing and analyzing the initial <|start|>assistant<|channel|>

<|start|>assistant<|channel|>commentary to=tool.create_new_file <|constrain|>json<|message|>{"filepath":"new/README.md","contents":"# Project Title\n\nA brief description of what this project does and who it's for.\n\n## Getting Started\n\nInstructions to get the project up and running locally.\n\n## Usage\n\nExamples of how to use the project.\n\n## Contributing\n\nGuidelines for contributing to this project.\n\n## License\n\nSpecify the license under which the project is released."}<|call|>

PR #10577, opened 2025-09-17
Fix the GPT function calling regex to allow dash in the name

antoine-roux:
## Motivation

Since LobeChat is using dash in its tool specification (something like "functions.lobe-chat_searchOnline", the regex does not match.

## Modifications

This **short** (maybe the shortest) PR, fixes that : one character in a regex.


JustinTong0323 (approved):
LGTM

PR #10423, opened 2025-09-14
[1/N] Introduce Mooncake Backend and Mooncake EP to Support Elastic EP

UNIDY2002:
## Motivation

**Why do we need a new backend?**

The Mooncake Backend is a collective communication backend that provides fault tolerance while remaining seamlessly compatible with PyTorch. To the best of our knowledge, this is the first backend that supports fault tolerance for both:
- `torch.distributed` primitives, and
- EP all-to-all primitives.

This lays the foundation for fault-tolerant distributed MoE inference (#8961).

## Modifications

The Mooncake Backend and Mooncake EP can be enabled via two options:

**Mooncake Backend**
- Set `--dist-backend mooncake`.
- This replaces NCCL, Gloo, etc. used by `torch.distributed` with Mooncake.
- Implemented by registering `"mooncake"` as a custom backend in `torch.distributed.init_process_group`.
- Minimal code changes are required (non-intrusive integration).

**Mooncake EP**
- Set `--moe-a2a-backend mooncake`.
- This replaces the DeepEP all-to-all backend with Mooncake EP.
- Implemented by adding a new class `MooncakeEPDispatcher` in `token_dispatcher/mooncake.py`.

## Accuracy Tests

- Added a new unittest: `ep/test_mooncake_ep_small.py` to validate correctness of the new features.
- All existing unittests remain unaffected and should continue to pass.

**Caveat:**

The Mooncake package with the new backend support will be published soon.

For now, you may download the wheel named with "+ep" from the CI artifacts of https://github.com/kvcache-ai/Mooncake/pull/805.

## Benchmarking and Profiling

According to our unit tests, enabling the Mooncake Backend and Mooncake EP may slightly decrease performance. This is expected, since the Mooncake Backend is not yet fully optimized; optimization work will be carried out separately.

We compared `test_mooncake_ep_small.py` with `test_deepep_small.py`.

**Throughput (token/s):**

| TestCase       | test\_mooncake\_ep\_small.py | test\_deepep\_small.py |
| -------------- | ---------------------------- | ---------------------- |
| TestPureDP     | 1581.135                     | 2011.321               |
| TestHybridDPTP | 1324.707                     | 1773.038               |
| TestTP         | 1484.145                     | 1668.714               |

**Latency (s):**

| TestCase       | test\_mooncake\_ep\_small.py | test\_deepep\_small.py |
| -------------- | ---------------------------- | ---------------------- |
| TestPureDP     | 11.561                       | 9.063                  |
[...]

PR #8961, opened 2025-08-08
Elastic EP Support (Milestone 1 & 2)

UNIDY2002:
# SUPERSEDED BY #15771

Progress tracker:
- https://github.com/sgl-project/sglang/pull/10423
- https://github.com/sgl-project/sglang/pull/10606
- https://github.com/sgl-project/sglang/pull/11657
- https://github.com/sgl-project/sglang/pull/17374
- https://github.com/sgl-project/sglang/pull/12068
- https://github.com/sgl-project/sglang/pull/15771

Drafts:
- https://github.com/sgl-project/sglang/pull/11837

## Motivation

In the context of MoE models like DeepSeek V3, GPU resources must be dynamically expanded or reduced in response to the request rate. Additionally, the system must maintain correctness even when some GPUs fail, which is common in large-scale inference scenarios. Any changes or recovery in configuration should be completed within an acceptable time frame.

This PR aims to introduce elastic EP support for SGLang, as a part of #8210.

## Roadmap

**Milestone 1: fault tolerance in the case of masking out a certain rank**
In this stage, we simulate a broken GPU by masking out a certain rank during inference. We expect most results (*num_healthy_ranks* out of *num_total_ranks*) to be correct. This milestone serves as a proof of concept for fault tolerance in the DeepSeek model, without requiring changes to the SGLang scheduler engine.

**Milestone 2: fault tolerance in the case of killing some ranks**
In this stage, we plan to enhance the SGLang's scheduler to make it aware of the underlying faults and dynamically redirect requests to healthy ranks. This requires adjustments to the P2P communication logic within the scheduler and other necessary modifications. This milestone will provide true fault tolerance.

**Milestone 3: scaling up and scaling down**
In this stage, we plan to introduce the ability to dynamically adjust GPU usage based on the system administrator's configuration, enabling seamless scaling of resources. This will include support for scaling the world size during inference, ensuring that the system remains efficient and responsive.

### Enhance the DeepEP Library to Support Fault Detection and Fault Tolerance

The current implementation of DeepEP assumes that all ranks are healthy and reachable. To handle faulty conditions, we modified the library to detect and manage faults. Specifically, we introduced a time limit for P2P communication during token dispatch, marking a peer as failed if a timeout occurs during the P2P receiving phase. This failure status is propagated to upper-layer modules to support EPLB's needs. To ensure consistency across all ranks, a master rank is elected to monitor the health of all peer ranks, which is critical for the EPLB module to gather statistical information of all the healthy ranks.

[...]

PR #10606, opened 2025-09-18
[2/N] Added the core structure of elastic EP and the eplb algorithm with faulty rank

HanHan009527:
## Motivation

1. Integrating Mooncake's fault-awareness, we need to adjust the eplb algorithm and model loading logic to enable the forward pass to bypass faulty ranks.

base on #10423
check our next pr #11657 and full draft #8961 (update on 10.16) to test the effect of fault redundancy

The ut part is modified to facilitate testing on machines with different ibdev names

## Modifications
1. Adding the core structure of the elastic ep module to bridge forward propagation and scheduling.
2. Add an EPLB algorithm that supports fault rank.

## Accuracy Tests
Test results are available from our full draft version. #8961 

## Checklist


ShangmingCai:
CC: @zhihui1084

python/sglang/test/test_disaggregation_utils.py
```diff
@@ -100,41 +103,57 @@ def tearDownClass(cls):
...
+    ]
+    logger.info("Resolved rdma_all_devices=%s", rdma_all_devices)
+
+    n_rdma = len(rdma_all_devices)
     # 1. Get visible GPU indices
     cuda_visible_devices = os.getenv("CUDA_VISIBLE_DEVICES")
     if not cuda_visible_devices:
         warnings.warn("CUDA_VISIBLE_DEVICES is not set. Using default RDMA devices.")
-        return "mlx5_roce0,mlx5_roce4"
+        return ",".join(_pick_default_pair(rdma_all_devices))
```
ShangmingCai: IB device topo detection and auto fault tolerance feature of mooncake is buggy, I think we only need one IB from each NUMA, it is enough for the current CI.

HanHan009527: This part is to enable the volcano h20 to perform the test. The mellanox roce card of that machine is named  to mlx5_1, etc. Let me move this part to another pr

python/sglang/srt/elastic_ep/elastic_ep.py
```diff
@@ -0,0 +1,77 @@
...
+        )
+
+    @classmethod
+    def healthy_rank_state(
+        cls, *, ep_size: Optional[int], device: Optional[torch.device]
+    ) -> torch.Tensor:
+        size = ep_size if ep_size is not None else torch.distributed.get_world_size()
+        dev = device if device is not None else cls._select_device()
+
+        return torch.ones(size, dtype=torch.int32, device=dev)
```
ShangmingCai, line 74: Should this dtype be changed to` torch.int64` to align with the future usage?

python/sglang/srt/elastic_ep/elastic_ep.py
```diff
@@ -0,0 +1,77 @@
...
+            last_active_ranks=active.clone(),
+            active_ranks_cpu=active.detach().cpu().clone(),
+        )
+
+    @classmethod
+    def healthy_rank_state(
+        cls, *, ep_size: Optional[int], device: Optional[torch.device]
+    ) -> torch.Tensor:
+        size = ep_size if ep_size is not None else torch.distributed.get_world_size()
+        dev = device if device is not None else cls._select_device()
```
ShangmingCai, line 72: nit: maybe give `ep_size` and `device` a default value: `None`

ShangmingCai (approved):
LGTM. Changes are clean.

python/sglang/srt/elastic_ep/elastic_ep.py
UNIDY2002: Mooncake EP currently uses int32. BTW, what does "future usage" refer to?

ShangmingCai: @UNIDY2002 Just wonder if it should align with `log2phy`, which is int64. But I am not sure.

UNIDY2002: I had a quick verification. `active_ranks` has two usages in `rebalance_experts`: (1) `num_active_ranks = active_ranks.sum().item()`; (2) `active_ranks_list = active_ranks.tolist()`, so I think using int32 for `active_ranks` may be okay.

python/sglang/srt/elastic_ep/elastic_ep.py
```diff
@@ -0,0 +1,77 @@
...
+    _instance: Optional[ElasticEPState] = None
+    _lock = threading.Lock()
+
+    @classmethod
+    def instance(cls) -> ElasticEPState:
+        return cls._instance
+
+    @classmethod
+    def init(cls, server_args: ServerArgs):
+        with cls._lock:
```
fzyzcjy: nit: shall we init it in one single thread and only once, then the code can be simplified here

fzyzcjy (commented):
LGTM only a nit

python/sglang/srt/elastic_ep/elastic_ep.py
HanHan009527: get, I will remove this lock. This part just follows the customary writing style of singleton, which is originally single-threaded.

python/sglang/srt/elastic_ep/elastic_ep.py
HanHan009527: done

PR #10629, opened 2025-09-18
refactor: use registry for _get_attention_backend_from_str

zhyncs:
## Motivation

Right now, `_get_attention_backend_from_str` is implemented as a large if-elif chain. Every time we add a new backend, we need to modify this function, which is a textbook violation of the Open/Closed Principle (OCP) — the code is not extension-friendly. In particular, if we want to add a private backend, we have to patch this function, even though the actual change is minimal. Introducing a registry mechanism would eliminate this issue by allowing new backends to be registered without touching the core function, making the design cleaner and more extensible.
These registries don’t actually need to be written in one place. I only put them together here to avoid modifying the existing backend code. In fact, a new private backend can be written separately, we just need to make sure it is imported.

## Checklist


zhyncs:
@iforgetmyname can you help take a look? https://github.com/sgl-project/sglang/actions/runs/17843042129/job/50736799499?pr=10629 thanks!

PR #10634, opened 2025-09-19
Fix fast decode plan for flashinfer v0.4.0rc1 and upgrade sgl-kernel 0.3.11

Fridge003:
## Motivation

In #10587, updating flashinfer to v0.4.0rc1 will cause some bugs, which is caused by some extra arguments added in https://github.com/flashinfer-ai/flashinfer/pull/1661 and  https://github.com/flashinfer-ai/flashinfer/pull/1675.

To solve this, we just need to pass the default values of these extra arguments in `fast_decode_plan` of flashinfer backend.

## Checklist


zhyncs (approved):

zhyncs:
I've tested b200 ut locally and it works well, I think that the ci failed because of the new flashinfer jit cache. I need the new version and let's merge first.

PR #2699, opened 2025-01-02
[Feature] Support regex as a stopping condition

mickqian:
## Motivation

Addressing the issue #2007.

## Modifications

1. Adds `stop_regex_strs` field to `SamplingParams`
2. Implement stop logic in `Req::check_finished`

## Checklist

- [x] Format your code according to the [Contributor Guide](https://github.com/sgl-project/sglang/blob/main/docs/references/contributor_guide.md).
- [x] Add unit tests as outlined in the [Contributor Guide](https://github.com/sgl-project/sglang/blob/main/docs/references/contributor_guide.md).
- [x] Update documentation as needed, including docstrings or example tutorials.

This PR is based on top of #2035, with modifying the stopping logic.

Issue #2007, opened 2024-11-11
[Feature] Regex stop condition


SinanAkkoyun:
### Checklist

- [X] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [X] 2. Please use English, otherwise it will be closed.

### Motivation

Hi! This would be awesome, as it would address following problems:
- most custom stopping conditions can be expressed as a regex
- handling custom stopping in a streaming response does not work as quickly as a backend based stopping condition would

(at least in my testing, a NodeJS client (with breaking the AsyncGenerator for await loop) can not stop the SGLang streaming generation the same way it works with the official OpenAI API)

I hope this is easy to implement

merrymercy:
Contribution is welcome!

glenliu21:
@merrymercy @mickqian Hi all, is this issue still open? I see that there have been some PRs on it that have been opened and closed. But if it's still open, I'm happy to be assigned to it and work on it!

mickqian:
Yes it is, you can start from my pr

glenliu21:
@mickqian Awesome, thanks! Please assign it to me, and I'll start working on it right away.

PR #10635, opened 2025-09-19
[Feature] support regex strings as a stopping condition

glenliu21:
## Motivation

In this PR, I address the issue raised in #2007. This PR builds off of #2699.

## Modifications

1. Add `stop_regex` as a field to `SamplingParams`.
2. Write a function that gets a strict upperbound on the number of tail tokens that need to be decoded given the provided `stop_regex` strings.
3. Implement the stopping logic in `Req::check_finished`.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added regex-based stop conditions via a new stop_regex parameter across generation APIs and OpenAI-compatible endpoints; responses now stop when patterns match and surface a matched-regex finish reason.

* **Documentation**
  * Sampling parameters guide updated to document stop_regex usage and behavior.

* **Tests**
  * Added tests for stop_regex in completions/chat and for max-length estimation of complex regex patterns.


python/sglang/srt/managers/schedule_batch.py
```diff
@@ -797,6 +810,19 @@ def check_finished(self):
                     self.finished_reason = FINISH_MATCHED_STR(matched=stop_str)
                     return
 
+        # Check stop regex
```
hnyls2002: When the `stop_str` and `stop_regex_strs` are both set, only get `tail_str` once.

hnyls2002 (changes requested):

whybeyoung:
can  you resolve the confilicts ? @glenliu21

JustinTong0323 (commented):
overall lgtm, could you please resolve the conflicts so we can run CI?

glenliu21:
Merge conflicts have been fixed, and it seems like failing tests aren't related to my changes.

cc @whybeyoung @JustinTong0323 @hnyls2002

hnyls2002 (approved):
LGTM

hnyls2002:
@whybeyoung, Are you also interested in handling the streaming cases for regex stopping? Similar to string matching, we need to check whether a suffix of the current tailing string can be accepted by a grammar.

JustinTong0323 (approved):

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -735,8 +748,17 @@ def init_incremental_detokenize(self):
...
+        if (
+            len(self.sampling_params.stop_strs) > 0
+            or len(self.sampling_params.stop_regex_strs) > 0
+        ):
+            max_len_tail_str = max(
+                self.sampling_params.stop_str_max_len + 1,
+                self.sampling_params.stop_regex_max_len + 1,
+            )
+
+        tail_len = min((max_len_tail_str + 1), len(self.output_ids))
```
hanming-lu, line 761: Is this a bug that it has one extra `+1` than before? `max_len_tail_str` has `+1` already

glenliu21: Thanks for spotting this! Put up a fix in #18584.

PR #10637, opened 2025-09-19
fix: radix cache memory accounting

skyzh:
## Motivation

There seems to be two bugs in the radix cache:

- In the case of page_size != 1 we might get key == 0 after truncating the key; in that case, we should return an empty match in advance
- `self.evictable_size_` when adding/removing the leaf should both use value instead of key

## Checklist

### DIFF
--- a/python/sglang/srt/mem_cache/radix_cache.py
+++ b/python/sglang/srt/mem_cache/radix_cache.py
@@ -595,7 +595,7 @@ class RadixCache(BasePrefixCache):
             new_node.key = key
             new_node.value = value
             node.children[child_key] = new_node
-            self.evictable_size_ += len(value)
+            self.evictable_size_ += len(key)
             self._record_store_event(new_node)
         return total_prefix_length
 
--- a/python/sglang/srt/mem_cache/radix_cache.py
+++ b/python/sglang/srt/mem_cache/radix_cache.py
@@ -267,6 +267,17 @@ class RadixCache(BasePrefixCache):
         """
         key.token_ids = self.key_convert_fn(key.token_ids)
 
+        def empty_match_result():
+            return MatchResult(
+                device_indices=torch.empty(
+                    (0,),
+                    dtype=torch.int64,
+                    device=self.device,
+                ),
+                last_device_node=self.root_node,
+                last_host_node=self.root_node,
+            )
+
         if self.disable or len(key) == 0:
             return empty_match_result()
 
--- a/python/sglang/srt/mem_cache/radix_cache.py
+++ b/python/sglang/srt/mem_cache/radix_cache.py
@@ -268,19 +268,14 @@ class RadixCache(BasePrefixCache):
         key.token_ids = self.key_convert_fn(key.token_ids)
 
         if self.disable or len(key) == 0:
-            return MatchResult(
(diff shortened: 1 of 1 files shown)


zhaochenyang20:
fix the lint plz @skyzh

skyzh:
thanks :) lint should work now

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -479,7 +486,7 @@ def _delete_leaf(self, node):
             if v == node:
                 break
         del node.parent.children[k]
-        self.evictable_size_ -= len(node.key)
+        self.evictable_size_ -= len(node.value)
```
Edenzzzz: maybe it should just be `len(node.key)`? value is the page-sized aligned token indices into the page table, key is the list of token ids. Actually they should have the same len
https://github.com/sgl-project/sglang/blob/c68487b9617ae212ad86f16ec98073237a82beaa/python/sglang/srt/mem_cache/radix_cache.py#L279

skyzh:
@Edenzzzz thanks for the reviews :) there's also:

```
            if node.lock_ref == 0:
                self.evictable_size_ -= len(node.value)
                self.protected_size_ += len(node.value)
```

in the same file, which sounds like we are moving something from evictable to protected, should I also use `node.key` here?

Edenzzzz:
cc @xiezhq-hermann I feel these two can be used interchangeably?

python/sglang/srt/mem_cache/radix_cache.py
xiezhq-hermann: they should have the same length and actually it was len(value) and got changed to this in a recent PR, let's keep it len(key) for consistency

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -188,10 +189,16 @@ def match_prefix(self, key: List[int], **kwargs) -> MatchResult:
...
 
+        if self.disable or len(key) == 0:
+            return empty_match_result()
+
         if self.page_size != 1:
             page_aligned_len = len(key) // self.page_size * self.page_size
             key = key[:page_aligned_len]
 
+        if len(key) == 0:
+            return empty_match_result()
```
xiezhq-hermann, line 289: it's good we do a guardrail here, but is this causing a bug? I believe there are protection within `_match_prefix_helper` as well.

python/sglang/srt/mem_cache/radix_cache.py
skyzh: yep - I'll push a fix

python/sglang/srt/mem_cache/radix_cache.py
skyzh: kind of.. this comes from my experiment to enable deterministic radix cache by simply setting the page_size_ in radix cache to the split size of prefill - it yields several issues. So this might just work fine with the current code.

My other argument is that we are going to return empty anyways, so we can skip a bunch of code here and do a shortcircuit path to directly return empty?

skyzh:
ready for review again :) thanks! cc @xiezhq-hermann

xiezhq-hermann (approved):

PR #10639, opened 2025-09-19
[PoC] make radix cache deterministic

skyzh:
## Motivation

This patch is based on https://github.com/sgl-project/sglang/pull/10678.

The patch (tries) adding determinism to the radix cache. I overrode the split size with 1024 so that the model fits into my GPU.

```
python3 -m sglang.launch_server --model-path Qwen/Qwen3-0.6B --attention-backend flashinfer --enable-deterministic-inference
python3 -m sglang.test.test_deterministic --test-mode prefix --temperature 0.7 --n-trials 12
```

Before:

```
Prompt 0 with prefix length 1: total samples: 18, Unique samples: 1
Prompt 1 with prefix length 511: total samples: 17, Unique samples: 3
Prompt 2 with prefix length 2048: total samples: 17, Unique samples: 5
Prompt 3 with prefix length 4097: total samples: 26, Unique samples: 4
```

After:

```
Prompt 0 with prefix length 1: total samples: 20, Unique samples: 1
Prompt 1 with prefix length 511: total samples: 18, Unique samples: 1
Prompt 2 with prefix length 2048: total samples: 22, Unique samples: 1
Prompt 3 with prefix length 4097: total samples: 18, Unique samples: 1
```

## Modifications

The patch mainly modifies the match_prefix logic:

- If it's called from the scheduler, always return the value to the multiply of SPLIT_SIZE.
- Otherwise, insert into the tree as normal.

The algorithm is two-pass:

- In the first pass, do the prefix matching as before.
- If we need to align to the split size, in the second pass, we use the nodes we collected from the previous pass to split exactly at the split size and return the node.

Note that this would do two splits in one match_prefix: one during the normal match to split at any prefix match, and another during we split at the split_size point; but it seems to work as desired. Please let me know if this doesn't work (memory leak?)

We can also do this in one pass and I don’t have a strong preference of which approach to take — feel free to suggest.

## Checklist

### DIFF
--- a/python/sglang/srt/layers/attention/flashinfer_backend.py
+++ b/python/sglang/srt/layers/attention/flashinfer_backend.py
@@ -137,10 +137,10 @@ class FlashInferAttnBackend(AttentionBackend):
         if self.enable_deterministic:
             self.decode_use_tensor_cores = True
             self.prefill_split_tile_size = get_int_env_var(
-                "SGLANG_FLASHINFER_PREFILL_SPLIT_TILE_SIZE", 4096
+                "SGLANG_FLASHINFER_PREFILL_SPLIT_TILE_SIZE", 1024
             )
             self.decode_split_tile_size = get_int_env_var(
-                "SGLANG_FLASHINFER_DECODE_SPLIT_TILE_SIZE", 2048
+                "SGLANG_FLASHINFER_DECODE_SPLIT_TILE_SIZE", 1024
             )
             self.disable_cuda_graph_kv_split = True
             global_config.flashinfer_workspace_size = 2048 * 1024 * 1024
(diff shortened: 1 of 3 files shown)


skyzh:
the patch seems to work on my machine -- still resolving the conflicts and avoid having the `SPLIT_SIZE` global variable.

JustinTong0323:
Hi, Mr. Chi! 🤣 Thank you so much for your contribution! I've just refactored the key for the radix cache. Could you please resolve the conflict? Really appreciate it

skyzh:
Yep no worries I need to wait for the upstream patch before I can resolve 🤣

skyzh:
looks like GitHub automatically closed this patch as it's based on a deleted branch ; will reopen one

zhyncs:
Sorry lol

skyzh:
reopened at https://github.com/sgl-project/sglang/pull/10721

Issue #10640, opened 2025-09-19
[Bug] SGL Frontend Language breaks when n>1


itsmeknt:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Describe the bug

I am trying to use the SGLang frontend language to create an sgl.function, like this:

```python
@sgl.function
def sgl_func(s: ProgramState, question: str) -> None:
    s += sgl.user(question)
    s += sgl.separate_reasoning(sgl.assistant(sgl.gen("final_answer", max_tokens=64000, n=3)))
```
Notice that `sgl.gen()` specifies `n=3`, which is to create 3 distinct responses.

I print my results here:
```python
def test():
    response = sgl_func.run("My mermaid diagram is having an issue adding parenthesis inside a label. Give me 10 possible solutions to fix this") 
    print(response.get_var("final_answer"))
```

but get this error:
```
/home/kevin/workspace/hpitt/.venv/lib/python3.12/site-packages/sglang/lang/interpreter.py:410: UserWarning: Error in stream_executor: Traceback (most recent call last):
  File "/home/kevin/workspace/hpitt/.venv/lib/python3.12/site-packages/sglang/lang/interpreter.py", line 408, in _thread_worker_func
    self._execute(expr)
  File "/home/kevin/workspace/hpitt/.venv/lib/python3.12/site-packages/sglang/lang/interpreter.py", line 451, in _execute
    self._execute(x)
  File "/home/kevin/workspace/hpitt/.venv/lib/python3.12/site-packages/sglang/lang/interpreter.py", line 446, in _execute
    self._execute_gen(other)
  File "/home/kevin/workspace/hpitt/.venv/lib/python3.12/site-packages/sglang/lang/interpreter.py", line 574, in _execute_gen
    comp, meta_info = self.backend.generate(
                      ^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #10642, opened 2025-09-19
Fix: frontend language to allow to handle JSONArray responses when n>1

itsmeknt:
## Motivation

Provide a fix for issue #10640 

In summary: when using the SGL frontend language, if you specify `n > 1` in `sgl.gen()`, the function currently throws an error (see issue #10640 for detailed error log). This is because in `sglang/lang/backend/runtime_endpoint.py`, the response handling assumes the response is a JSON (i.e. `n=1`) but did not handle if it were a JSONArray (i.e. `n>1`).

## Modifications

Just a few lines of code in `sglang/lang/backend/runtime_endpoint.py`.

I check `isinstance(obj, list)`, and handle JSONArrays differently than JSONs.

## Accuracy Tests

n/a

## Benchmarking and Profiling

n/a

## Checklist


itsmeknt:
I'm new to this project, how do I add the `run-ci` label to this PR? It seems to make the tests fail otherwise

mickqian:
> Yes, `mickqian`, we can simplify this logic by unifying the handling of single JSON objects and JSON arrays. By ensuring `obj` is always treated as a list, we can remove the `else` block and make the code more concise and easier to read. This aligns with Python's PEP 8 guidelines for readability.
> 
> Here's a suggestion:
> 
> ```python
>         # Ensure obj is always a list for consistent processing
>         if not isinstance(obj, list):
>             obj = [obj]
> 
>         if not obj:
>             return [], []
> 
>         comps, meta_infos = zip(
>             *((item["text"], item["meta_info"]) for item in obj)
>         )
>         return list(comps), list(meta_infos)
> ```
> 
> This change maintains the current functionality while reducing code duplication and improving clarity. The severity of this suggestion is `low` as it's a stylistic improvement.

@itsmeknt Could you apply this suggestion?

itsmeknt:
@mickqian I implemented the simplification and pushed, but some of the CI tests were failing due to incorrect type handling.

What I realized is that the simplified code changed the return type behavior. If `obj` is a JSON, the function should return `(str, dict[str, Any])` to be compatible with the old behavior.

The simplified code made the function always return a `(list[str], list[dict[str, Any]])` whether `obj` was a JSON or JSONArray, which broke some of the integration tests that expected a `(str, dict[str, Any])`.

itsmeknt:
@mickqian gemini doesnt seem to be responding.. perhaps try again?

Issue #10453, opened 2025-09-15
[Feature] Simple service discovery for mini_lb


hnyls2002:
### Checklist

- [x] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 2. Please use English, otherwise it will be closed.

### Motivation

Only for debugging.

PR #10655, opened 2025-09-19
Support service auto service-discovery for minilb on Kubernetes

whybeyoung:
Motivation: implement for https://github.com/sgl-project/sglang/issues/10453

Main work: Align the service discovery mode of the router Rust version in the Kubernetes cluster
Mainly support minilb to run in some debugging environments

image

you  use cmd:
```
 python -m sglang_router.launch_router --mini-lb --port 8000 --host 0.0.0.0  --pd-disaggregation --service-discovery --service-discovery-namespace aiservice        --service-discovery-port 30000  --prefill-selector pd_role=prefill  --decode-selector pd_role=decode

```
Before using this feature, you must have a Kubernetes cluster and the label selector for prefill decode corresponding to the same namespace. Additionally, you need to configure the correct RBAC.

for example:
```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  name: pod-watcher
  namespace: aiservice
rules:
- apiGroups: [""]
  resources: ["pods"]
  verbs: ["get", "list", "watch"]
---
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: pod-watcher-binding
  namespace: aiservice
subjects:
- kind: ServiceAccount
  name: default   # sa that you use in prefill/decode pod
  namespace: aiservice
roleRef:
  kind: Role
  name: pod-watcher
  apiGroup: rbac.authorization.k8s.io
```
CC @ShangmingCai @hnyls2002


sgl-router/setup.py
```diff
@@ -4,6 +4,19 @@
...
+    "setproctitle",
+    "aiohttp",
+    "orjson",
+    "uvicorn",
+    "fastapi",
+]
+
+# Add kubernetes dependency for non-Rust builds (Python-only mini_lb)
+if no_rust:
+    install_requires.append("kubernetes>=18.0.0")
```
slin1237, line 18: k8s has a lot of api dependency management overhead, we had that in rust, now because miniLB is part of the router package, this means Python k8s package will be another of concern. 
are we really sure of that?

sgl-router/setup.py
```diff
@@ -17,6 +30,12 @@
     )
 
 setup(
+    name="sglang-router",
```
slin1237, line 33: these are already in project.toml file

slin1237:
in that regard
I would like to understand, 
- shall we replicate all features and capabilities? 
- the role of each (rust and py)
- Maintainability moving forward, we understand the overhead of never ending k8s api dependency upgrade

because we are setting a precedent here. I'm a little confused and concerned

whybeyoung:
First of all, minilb will never be a viable option for production use. It only needs to support automatic discovery in the Kubernetes (k8s) environment for the purpose of debugging within that environment.

Secondly, I will consider adjusting the setup such that users are only prompted to install Kubernetes dependencies when using minilb, rather than making Kubernetes a dependency of the entire project.

Do you think the above approach is feasible? @slin1237

slin1237:
what's blocking using router for debugging.

ShangmingCai:
> what's blocking using router for debugging.

@Simon-Li I am testing Hicache + PD + mooncake store. In some cases, timeouts are triggered, causing requests to be aborted. At this time, mini lb without any advanced features like load balancing and fault detection may be useful.

whybeyoung:
> what's blocking using router for debugging.

 totally understand your concern, haha. This needs to be brought to @ShangmingCai  attention for review.

Generally speaking, I think the correctness of some PD functions can be verified with a simple Python proxy. Before the router is fully officially released, could we first allow customers to use minilb to check for issues?

On a positive note, the feedback on the router has indeed become more and more stable, and some debugging work can also be done in Rust. All in all, Master Simo, don’t worry about it—everything is for the betterment of the Rust router!

PR #10661, opened 2025-09-19
[router] preserve order of json params using preserve_order feature

fgebhart:
## Motivation

This change was found to the be fix to a problem we identified when using tool calling together with the sglang rust router. The symptoms of the issue are degraded performance in the quality of the outputs of the models (deepseek in this case, but I assume it does not matter).
Consider the following prompt body which would be sent to the sglang router

<details>

<summary>Example prompt body snipped</summary>

```python
body = {
    "messages": [
        {
            "content": "You are a helpful assistant ...",
            "role": "system",
        },
        {"role": "user", "content": "How to build a paper plane?"},
    ],
    "model": model,
    "tool_choice": "required",
    "tools": [
        {
            "type": "function",
            "function": {
                "name": "intent_analysis",
                "description": "...",
                "parameters": {
                    "properties": {
                        "true_intent": {
                            "description": "baa",
                            "type": "string",
                        },
                        "primary_objective": {
                            "description": "baz",
                            "type": "string",
                        },
                        "hidden_complexities": {
                            "description": "baz",
                            "items": {"type": "string"},
                            "type": "array",
                        },
                        "clarifications": {
                            "description": "baa",
                            "items": {"type": "string"},
                            "type": "array",
                        },
                    },
                    "required": [
                        "true_intent",
                        "primary_objective",
                        "hidden_complexities",
                        "clarifications",
                    ],
                    "type": "object",
                    "additionalProperties": False,
                    "$defs": {
                        "Constraint": {
                            "description": "baz",
                            "properties": {
                                "temporal": {
                                    "description": "...",
                                    "anyOf": [{"type": "string"}, {"type": "null"}],
                                },
                                "geographical": {
                                    "description": "...",
                                    "anyOf": [{"type": "string"}, {"type": "null"}],
                                },
                                "depth": {
                                    "description": "...",
                                    "type": "string",
                                },
                            },
                            "type": "object",
                            "additionalProperties": False,
                            "required": [
                                "temporal",
                                "geographical",
                                "depth",
                            ],
                        }
                    },
                },
                "strict": True,
            },
        }
    ],
}
```
</details>

Without the change in this PR, the order of elements (in the above example: `["true_intent", "primary_objective", "hidden_complexities", "clarifications"]`) in the json payload generated from `serde_json` would not be preserved but rather sorted alphabetically. We found this to have undesired effects on the output quality of the LLM, as the input would be different. 

On a side note; the JSON standard does not preserve order of elements in a set like this. Hence one could argue that JSON might not be the best option for this purpose, but this is yet another discussion. The problem however, stems from the fact that `serde_json` does apply ordering to the input given by the user. To prevent this (alphabetical) ordering we propose to enable the `preserve_order` feature of `serde_json`. Doing so we achieved the desired output quality, which is in line with the output of sglang without the rust router. 

Let me know what you think. Happy to discuss!

## Modifications

I'm changing the `cargo.toml` of the sglang router only to enable the `preserve_order` feature of `serde_json`.

## Accuracy Tests

I did not (yet) touch any tests at all. Let me know what you think about testing this change.

## Benchmarking and Profiling

No benchmarking performed.

## Checklist

~~- [ ] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).~~
~~- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.ai/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.ai/developer_guide/contribution_guide.html#benchmark-the-speed).~~

### DIFF
--- a/sgl-router/Cargo.toml
+++ b/sgl-router/Cargo.toml
@@ -24,7 +24,10 @@ axum = { version = "0.8.4", features = ["macros", "ws", "tracing"] }
 tower = { version = "0.5", features = ["full"] }
 tower-http = { version = "0.6", features = ["trace", "compression-gzip", "cors", "timeout", "limit", "request-id", "util"] }
 serde = { version = "1.0", features = ["derive"] }
-serde_json = "1.0"
+serde_json = { version = "1.0", default-features = false, features = [
+    "std",
+    "preserve_order",
+] }
 bytes = "1.8.0"
 rand = "0.9.2"
 reqwest = { version = "0.12.8", features = ["stream", "blocking", "json"] }


slin1237:
@fgebhart you guys have started using tool parser from in our router?
Tool parser isn't integrated fully in any of the code base at the moment
I'm curious how you guys are using it, assuming you are importing this in your own rust code?
I'm mostly super curious about the experience that you guys have had if above is true!

fgebhart:
We indeed do have a custom rust (OpenAI compatible) api sitting infront of sglang, but that does not apply any special magic for tool parsing. Up until recently we passed requests from our api to the OpenAI sglang API directly which worked flawlessly.

Then we introduced the sglang router into the loop and realized the degraded output quality. Eventually we figured out that the changed order of elements in the json was the root cause for this.

slin1237 (approved):

slin1237:
@fgebhart may i ask what performance degradation number we are seeing?

fgebhart:
@slin1237 

The performance impact was not on the lines of tokens throughput or similar, but rather in the output quality of the model. We ran a bunch of tests to get some statistics, which could be summarized like the following.

* Running against the openai sglang interface, the tool call was performed successfully in 10 out of 10 cases.
* Running against the sglang routner, the tool call was performed successfully in ~3 out of 10 cases.

With this fix we are again back to 10/10 via the router.

Hope that helps. Thanks for the quick review & merge!

PR #10646, opened 2025-09-19
feat: Add checkpoint-engine IPC integration

BraveY:
Implement IPC-based weight updates for checkpoint-engine compatibility:

This allows efficient model weight updates via IPC without server restart.

## Motivation
Implement IPC-based weight updates for [checkpoint-engine](https://github.com/MoonshotAI/checkpoint-engine) compatibility.

## Modifications

- Add SGLangCheckpointEngineWorkerExtension worker class
- Implement update_weights_from_ipc across scheduler/tokenizer/worker layers
- Add collective_rpc endpoint for vLLM API compatibility
- Support ZMQ communication with device UUID management
- Include post-loading hooks and error handling

## Accuracy Tests

Test command is similar with test update weights on vllm in [checkpoint-engine](https://github.com/MoonshotAI/checkpoint-engine)'s readme.  Just replace the command to start vLLM with the command to start SGLang.

Single Node:
Start sglang
```
python -m sglang.launch_server --model-path /workspace/Qwen/Qwen3-4B/ --tensor-parallel-size 2 --port 19730 --load-format dummy --mem-fraction-static 0.7
```
Update weigths by checkpoint-engine's script. 

```
torchrun --nproc-per-node 2 examples/update.py --update-method broadcast --checkpoint-path /workspace/Qwen/Qwen3-4B/  --inference-parallel-size 2
```

The weight update process of DeepSeek-R1 has been tested on a single machine and two machines. After the weight update is completed, the inference results of the model are verified to be normal.

Co-author: @stmatengss @XucSh @zxpdemonio
Thanks to @weixiao-huang for help.
## Checklist

- [ ] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.ai/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.ai/developer_guide/contribution_guide.html#benchmark-the-speed).

PR #10667, opened 2025-09-19
[Feature] Support loading weights from ckpt engine connector

stmatengss:
## Motivation

Motivated by https://github.com/sgl-project/sglang/pull/8215, we aim to integrate [ckpt-engine](https://github.com/moonshotai/checkpoint-engine)into SGLang to accelerate model loading and weight synchronization.

A proposal is in #10464, and this PR can support both co-locate/disaggregation deployment and TP.

Usage: 
```
pip install 'checkpoint-engine[p2p]'  # install checkpoint engine
```

~~Fake sglang server (only occupying model weights).~~
```
python3 -m sglang.launch_server --model [model-name] --tp 2 --page-size 1 --mem-fraction-static 0 --enable-ckpt-engine 
```

~~New sglang instance~~
```
python3 -m sglang.launch_server --model [model-name]  --tp 2 --mem-fraction-static 0.4 --load-format ckpt_engine --port [port_num]
```

Running Methods:

sglang
```
python3 -m sglang.launch_server --model /opt/models/Qwen/Qwen3-8b --tp 8 --load-format ckpt_engine --port 30001
```

checkpoint engine
```
torchrun --nproc-per-node 8 ckptengine_update.py --update-method all --checkpoint-path /opt/models/Qwen/Qwen3-8b/
```

Co-author: @XucSh @zxpdemonio @BraveY
Thanks to @weixiao-huang for help.

## Checklist


python/sglang/srt/connector/ckpt_engine.py
```diff
@@ -0,0 +1,191 @@
...
+                    self._merge_and_store(gate_key, gate_tensor, item["name"], tensor)
+                else:
+                    self.pending_weights[item["name"]] = tensor
+
+            else:
+                weights.append((item["name"], tensor))
+        return weights
+
+    # Implemented as a no-op to make BaseConnector interface consistent.
+    def weight_iterator(
```
weixiao-huang, line 195: Could we reuse the method `checkpoint_engine.worker.update_weights_from_ipc` to make code more simple like below
```python
    def _process_gate_up_proj(self, named_tensors: list[Tuple[str, torch.Tensor]]):
        for name, tensor in named_tensors:
            if "mlp.gate_proj.weight" in name:
                up_key = name.replace("gate_proj", "up_proj")
                if up_key in self.pending_weights:
                    up_tensor = self.pending_weights.pop(up_key)
                    self._merge_and_store(name, tensor, up_key, up_tensor)
                else:
                    self.pending_weights[name] = tensor

            elif "mlp.up_proj.weight" in name:
                gate_key = name.replace("up_proj", "gate_proj")
                if gate_key in self.pending_weights:
                    gate_tensor = self.pending_weights.pop(gate_key)
                    self._merge_and_store(gate_key, gate_tensor, name, tensor)
                else:
                    self.pending_weights[name] = tensor

            else:
                yield name, tensor
        for key, tensor in self.final_state_dict.items():
            yield key, tensor


    def weight_iterator(self, rank: int = 0) -> Generator[Tuple[str, torch.Tensor], None, None]:
        from checkpoint_engine.worker import update_weights_from_ipc
        if self.socket is None:
            self.get_socket_handle(rank)

        update_weights_from_ipc(
            self.zmq_ctx,
            self.zmq_handle,
            rank,
            run=self._process_gate_up_proj,
        )
```

python/sglang/srt/connector/ckpt_engine.py
```diff
@@ -0,0 +1,191 @@
...
+            tensor_data = buffer[offset : offset + size]
+            tensor = tensor_data.view(dtype=dtype).view(shape)
+
+            if "mlp.gate_proj.weight" in item["name"]:
+                up_key = item["name"].replace("gate_proj", "up_proj")
+                if up_key in self.pending_weights:
+                    up_tensor = self.pending_weights.pop(up_key)
+                    self._merge_and_store(item["name"], tensor, up_key, up_tensor)
+                else:
+                    self.pending_weights[item["name"]] = tensor
```
weixiao-huang: weights are received by using a fixed bucket size. So all weights may be split to multiple turn to be received. This tensor data from buffer may be changed in the next update turn. A workaround method is to `tensor.clone()` but will occupy more GPU memory. Is it necessary for this tensor to be saved in `pending_weights`?

python/sglang/srt/connector/ckpt_engine.py
```diff
@@ -0,0 +1,191 @@
...
+            else:
+                weights.append((item["name"], tensor))
+        return weights
+
+    # Implemented as a no-op to make BaseConnector interface consistent.
+    def weight_iterator(
+        self, rank: int = 0
+    ) -> Generator[Tuple[str, torch.Tensor], None, None]:
+        if self.socket is None:
+            self.get_socket_handle(rank)
```
weixiao-huang: `self.zmq_handle` will be changed in each `ps.update` method in the future's `checkpoint-engine==0.1.2` version. see https://github.com/MoonshotAI/checkpoint-engine/blob/03ff7e7268d614b5c5d3af7388e541fc181bd892/checkpoint_engine/ps.py#L812-L818 since `self._zmq_addr_counter += 1` will trigger for each `ps.update`. So `self.zmq_handle` should be updated by using `self.get_zmq_handle()` in each request

python/sglang/srt/managers/scheduler.py
```diff
@@ -324,6 +325,13 @@ def __init__(
             self.enable_overlap = False
             logger.info("Overlap scheduler is disabled for embedding models.")
 
+        # TODO: May change it to somewhere
+        os.environ["RANK"] = str(self.tp_rank)
```
weixiao-huang: Maybe I'll add `rank` and `world_size` args in `ParameterServer`, see https://github.com/MoonshotAI/checkpoint-engine/pull/20/files

XucSh:
Tested on Qwen3-0.6B (TP1 & TP2) and Qwen3-8B（TP4） models. New inference instances function correctly after the weight update.

python/sglang/srt/managers/scheduler.py
XucSh: Got it. We're planning to make this a startup argument for SGLang and are working on it.

python/sglang/srt/connector/ckpt_engine.py
XucSh: Thanks, that's a helpful reminder. We are refactoring the static port to be dynamically negotiated, and we'll be sure to include this point.

python/sglang/srt/connector/ckpt_engine.py
XucSh: In practice, this pending buffer is unused. The associated code will be removed in a subsequent commit.

python/sglang/srt/connector/ckpt_engine.py
XucSh: That's an excellent suggestion. I will address this in the next commit.

XucSh:
User can test with below command
1.  python3 -m sglang.launch_server --model /opt/models/Qwen/Qwen3-8b --tp 8 --load-format ckpt_engine --port 30001
2. torchrun --nproc-per-node 8 ckptengine_update.py --update-method all --checkpoint-path /opt/models/Qwen/Qwen3-8b/

XucSh:
Testing was conducted on an 8*A10 GPU machine with the following results:

For the Qwen3-0.6B model with TP=2, the loading time for the Checkpoint Engine was 3.72s, compared to 5.35s for loading from disk.
For the Qwen3-8B model with TP=4, the loading time for the Checkpoint Engine was 16s, compared to 85s for loading from disk.

stmatengss:
#10646 is related to this PR

whybeyoung:
LGTM

whybeyoung:
any  usage doc update ？

python/sglang/srt/connector/ckpt_engine.py
```diff
@@ -0,0 +1,182 @@
...
+from sglang.srt.utils import init_custom_process_group
+
+logger = logging.getLogger(__name__)
+
+# TODO: using dynamic port
+CKPTENGINE_PORT = 33001
+
+
+def _get_physical_gpu_id(rank: int) -> str:
+    result = subprocess.run(["nvidia-smi", "-L"], capture_output=True, text=True)
```
whybeyoung: maybe we should use python lib ?

python/sglang/srt/connector/ckpt_engine.py
```diff
@@ -0,0 +1,182 @@
...
+import torch.distributed as dist
+import zmq
+
+from sglang.srt.connector import BaseConnector
+from sglang.srt.utils import init_custom_process_group
+
+logger = logging.getLogger(__name__)
+
+# TODO: using dynamic port
+CKPTENGINE_PORT = 33001
```
whybeyoung: maybe need add to port_args

python/sglang/srt/connector/ckpt_engine.py
```diff
@@ -0,0 +1,182 @@
...
+        self.buffer: Optional[torch.Tensor] = None
+        self.local_rank = None
+        self.final_state_dict = OrderedDict()
+        self.pending_weights: Dict[str, torch.Tensor] = {}
+
+    def get_zmq_handle(self, tp_rank: int):
+        # FIXME: There needs a local rank
+        self.device_uuid = _get_physical_gpu_id(tp_rank)
+        socket = zmq.Context().socket(zmq.PULL)
+        socket.bind(f"tcp://*:{CKPTENGINE_PORT + tp_rank}")
```
whybeyoung: emmm, I don't particularly recommend continuing to use implicit calculation for port number increment here, as it can easily lead to port conflicts. Could we use the dynamic allocate method instead?

python/sglang/srt/connector/ckpt_engine.py
```diff
@@ -0,0 +1,182 @@
...
+
+                # Handle weight metadata payload
+                assert isinstance(payload, list)
+
+                model.load_weights(self._extract_weights(payload, buffer))
+
+                torch.cuda.synchronize()
+                self.socket.send(b"")
+        except Exception as e:
+            logger.error(f"Error in IPC weight update on device {rank}: {e}")
```
whybeyoung, line 231: can we  add some strategies to allow falling back to loading weights from the filesystem.

python/sglang/srt/connector/ckpt_engine.py
weixiao-huang: Maybe this code is a solution https://github.com/MoonshotAI/checkpoint-engine/blob/main/checkpoint_engine/ps.py#L257

python/sglang/srt/connector/ckpt_engine.py
stmatengss: We need a dynamic port selection strategy which is simlar to PD implementation.

python/sglang/srt/connector/ckpt_engine.py
stmatengss: Connectors currently lack a fallback path. Consider adding a fallback path for all connectors in the next PR.

python/sglang/srt/connector/ckpt_engine.py
XucSh: We are currently developing and testing the related code.

python/sglang/srt/entrypoints/http_server.py
```diff
@@ -793,6 +794,38 @@ async def update_weights_from_distributed(
...
+    if success:
+        return ORJSONResponse(
+            content,
+            status_code=HTTPStatus.OK,
+        )
+    else:
+        return ORJSONResponse(
+            content,
+            status_code=HTTPStatus.BAD_REQUEST,
+        )
```
JustinTong0323:
```suggestion
    status_code = HTTPStatus.OK if success else HTTPStatus.BAD_REQUEST
    return ORJSONResponse(
        content,
        status_code=status_code,
    )
```

python/sglang/ckptengine_update.py
JustinTong0323, line 1: This is not a good place to put this file

stmatengss: Is it ok to move it to `/scripts`?

python/sglang/srt/connector/ckpt_engine.py
XucSh: Refactoring...

python/sglang/ckptengine_update.py
weixiao-huang: This scripts seems to be the same as https://github.com/MoonshotAI/checkpoint-engine/blob/main/examples/update.py. Could we write a code to get this python file and exec it? Just like
```bash
wget https://raw.githubusercontent.com/MoonshotAI/checkpoint-engine/refs/heads/main/examples/update.py
python3 update.py --help
```

BraveY: > This scripts seems to be the same as https://github.com/MoonshotAI/checkpoint-engine/blob/main/examples/update.py. Could we write a code to get this python file and exec it? Just like
> 
> ```shell
> wget https://raw.githubusercontent.com/MoonshotAI/checkpoint-engine/refs/heads/main/examples/update.py
> python3 update.py --help
> ```

In fact, We need to modify the logic of this script because executing it directly allocates most of the GPU memory to the communication buffer, causing Sglang to run out of memory and fail to start. Therefore, the original script logic cannot be used directly.

stmatengss: Got it. will fix

python/sglang/srt/connector/ckpt_engine.py
stmatengss: Should we provide a ckpt_engine_port in server args for users?

python/sglang/ckptengine_update.py
stmatengss: > This scripts seems to be the same as https://github.com/MoonshotAI/checkpoint-engine/blob/main/examples/update.py. Could we write a code to get this python file and exec it? Just like
> 
> ```shell
> wget https://raw.githubusercontent.com/MoonshotAI/checkpoint-engine/refs/heads/main/examples/update.py
> python3 update.py --help
> ```

We can temporarily maintain this file within sglang, then merge it into the main checkpoint engine repository for easier maintenance.

stmatengss:
Will reimplement the checkpoint engine connector after merging PR #11755.

ByronHsu (commented):
High level questions:

1. Checkpoint engine needs to load from disk to cpu, and use the pipeline to send to inference engines. disk-to-cpu is still on the critical path. What is the time we save here?
2. Can you provide examples of how to use in online serving cases? For example, I add 10 new engines, how can i use checkpoint engine to make them load faster.
3. Can you provide examples of how to use in RL case?  How to do efficient broadcast to all inference engines.

Happy to chat online. You can find me at ByronHsu in sglang slack.

XucSh:
> High level questions:
> 
> 1. Checkpoint engine needs to load from disk to cpu, and use the pipeline to send to inference engines. disk-to-cpu is still on the critical path. What is the time we save here?
> 2. Can you provide examples of how to use in online serving cases? For example, I add 10 new engines, how can i use checkpoint engine to make them load faster.
> 3. Can you provide examples of how to use in RL case?  How to do efficient broadcast to all inference engines.
> 
> Happy to chat online. You can find me at ByronHsu in sglang slack.

The checkpoint service is a persistent process that holds weights in memory (GPU/CPU). Each new instance (specifically, each TP rank) has its own ParameterService（checkpoint engine） from which it fetches weights at startup.

PR #10693, opened 2025-09-20
[BugFix] port finding and avoid hardcoded ipv4 host 

xk-huang:
## Motivation

This PR fixes occasional NCCL rendezvous deadlocks and repeated `nccl_port` collisions when launching multi-GPU runs with `--dp > 1` (The bug is introduced in https://github.com/sgl-project/sglang/pull/7418). The previous logic tweaked the `nccl_port` by adding/subtracting small constants, which could still land on a busy/ephemeral port or cause different processes to race for the same value.

## Modifications
* Avoid hardcoded ipv4 localhost: as some machine only has ipv6, directly initializing `torch.dist` on ipv4 localhost would cause a stuck.
* In `python/sglang/srt/server_args.py`, during `PortArgs` initialization:
  * Remove the previous `±(42|43)` adjustments that could still collide or select a closed port which are all not available.
  * **`nccl_port` should also be increased in when `--dp > 1`**. Otherwise the port collision happens. Add a simple availability probe loop using `is_port_available(nccl_port)` and increment until a free port is found.
  
This keeps the selection deterministic relative to the server’s base port while adding jitter and a concrete availability check to avoid deadlocks. (Change diff is in the commit view.)

## Accuracy Tests

No changes to model kernels, scheduling, or outputs—only to rendezvous port selection. To be safe, I validated on the following setups:

* **Single node / 8 GPUs (NCCL):** multiple back-to-back launches with varying `--port` showed no deadlocks; all runs reached `dist.init_process_group()` and completed a small `all_reduce` sanity test.
* **Explicit `--nccl-port`:** when provided, the code respects the user value (no changes), and initialization succeeds if the port is free; fails fast if it’s actually bound by another process (as expected when `--dp > 1`).

Observed identical token-level outputs across runs compared to main (no accuracy deltas).

## Benchmarking and Profiling

Negligible overhead.

## Checklist

### DIFF
--- a/python/sglang/srt/model_executor/model_runner.py
+++ b/python/sglang/srt/model_executor/model_runner.py
@@ -621,7 +621,7 @@ class ModelRunner:
         if self.server_args.dist_init_addr:
             dist_init_method = f"tcp://{self.server_args.dist_init_addr}"
         else:
-            dist_init_method = f"tcp://127.0.0.1:{self.dist_port}"
+            dist_init_method = f"tcp://{self.server_args.host}:{self.dist_port}"
         set_custom_all_reduce(not self.server_args.disable_custom_all_reduce)
         set_mscclpp_all_reduce(self.server_args.enable_mscclpp)
 
--- a/python/sglang/srt/server_args.py
+++ b/python/sglang/srt/server_args.py
@@ -2788,9 +2788,14 @@ class PortArgs:
                 if nccl_port < 60000:
                     nccl_port += 42
                 else:
-                    nccl_port -= 43
+                    nccl_port = server_args.port + random.randint(100, 1000)
         else:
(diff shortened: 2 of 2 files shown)


jinmingyi1998:
I think this pr want to solve the nccl port conflict issue. but this method is wrong.

The root cause of port conflicts is the simultaneous launch of multiple processes within an extremely short time frame, which can result in duplicates even when using random values.

Another reason is that the `is_port_available` method does not determine availability as reliably as expected. The same port may return `True` when checked by different processes.

I recommend always setting `dist_init_addr` manually.

hnyls2002:
Thanks @xk-huang! Both the host fallback and free-port handling landed on main, so I'm closing this as
  superseded. Please reopen if I've missed a case.

PR #10705, opened 2025-09-21
feat: unify dockerfiles

ishandhanani:
## Motivation

SGLang should be the best across all platforms!! This PR unifies the current Dockerfile and Dockerfile.gb200 to stem from the same file. This simplifies maintenance for version bumps and makes CI/release easier.

In a follow up PR we should
1. test the sgl-kernel release -> pypi directly
2. update the dockerfile to pip install it vs installing from the repo
3. delete the gb200 dockerfile

# Tests

- [x] x86 dockerfile (on h100)
- [x] aarch64 dockerfile (on gb200)


zhyncs:
can we delete https://github.com/sgl-project/sglang/blob/main/docker/Dockerfile.gb200

zhyncs:
btw we can also upload arm whl to pypi https://pypi.org/project/sgl-kernel/0.3.11/

zhyncs:
So we don't need to install sgl-kernel from sgl-project/whl

ishandhanani:
1. I will delete gb200 dockerfile after I run tests! Or maybe it might be better to do in a follow up cleanup PR just to ensure everything goes smoothly

2. Sure - check my recent commits I think that should be the way. Anyway to test this?

ishandhanani:
~Note if https://github.com/sgl-project/sglang/pull/10732 goes in before this PR, we should bump to `0.3.12`~

Addressed in [7d1d108](https://github.com/sgl-project/sglang/pull/10705/commits/7d1d108fccf20fcda6753f327f00fe4fd3191e03)

docker/Dockerfile
```diff
@@ -1,10 +1,11 @@
-ARG CUDA_VERSION=12.9.1
-FROM nvidia/cuda:${CUDA_VERSION}-cudnn-devel-ubuntu22.04 as base
+ARG CUDA_VERSION=12.6.1
```
zhyncs: Default should be `12.9.1`

.github/workflows/release-docker-dev.yml
```diff
@@ -6,15 +6,55 @@ on:
...
 jobs:
-  build-dev:
+  build-dev-x86:
     if: ${{ github.repository == 'sgl-project/sglang' }}
     runs-on: ubuntu-22.04
     strategy:
       matrix:
         variant:
-          - version: 12.9.1
+          - version: 12.6.1
```
zhyncs:
```suggestion
          - version: 12.9.1
```

docker/Dockerfile
ishandhanani: Fixed

.github/workflows/release-docker.yml
```diff
@@ -65,25 +65,58 @@ jobs:
...
             echo "Unsupported build type"
             exit 1
           fi
 
-          docker buildx build --output type=image,compression=zstd . -f docker/Dockerfile --build-arg CUDA_VERSION=${{ matrix.cuda_version }} --build-arg BUILD_TYPE=${{ matrix.build_type }} -t lmsysorg/sglang:${tag}${tag_suffix} --no-cache
-
-          if [ "${{ matrix.cuda_version }}" = "12.6.1" ]; then
-            docker tag lmsysorg/sglang:${tag}${tag_suffix} lmsysorg/sglang:latest${tag_suffix}
-            docker push lmsysorg/sglang:latest${tag_suffix}
+          if [ "${{ matrix.variant.cuda_version }}" = "12.6.1" ]; then
```
zhyncs:
```suggestion
          if [ "${{ matrix.variant.cuda_version }}" = "12.9.1" ]; then
```

docker/Dockerfile
```diff
@@ -81,24 +82,33 @@ RUN python3 -m pip install --no-cache-dir --upgrade pip setuptools wheel html5li
       12.9.1) CUINDEX=129 ;; \
       *) echo "Unsupported CUDA version: $CUDA_VERSION" && exit 1 ;; \
     esac \
+ && if [ "$CUDA_VERSION" = "12.8.1" ]; then \
```
zhyncs: we don't need this

ishandhanani: fixed

zhyncs:
ref https://github.com/sgl-project/sglang/actions/runs/17937527600

zhyncs (approved):

zhyncs:
ref https://github.com/sgl-project/sglang/actions/runs/17937575539

ishandhanani:
> ref https://github.com/sgl-project/sglang/actions/runs/17937575539

Need to kick this off again so that sgl-kernel 0.3.12 is used

PR #10709, opened 2025-09-21
[4/4] Introduce CachedKernel to reduce CSGMV kernel launch overheads by 60%

lifuhuang:
## Motivation
After previous optimizations, currently LoRA kernel perf itself has significantly reduced, making the triton kernel launch time one of the main sources of overheads from LoRA. This PR caches compiled triton to bypass the generic but slow cache lookup mechanism in Triton. More relevant discussions can be found here: https://github.com/triton-lang/triton/issues/2637 . 

Many thanks to @ispobock for sharing the idea!

#### Before (319 us)

image

#### After (123.4 us)
image

### E2E Latency

Compared with the latest main (CSGMV v2), E2E latency at bs (8-64) is reduced by 4% - 9.4%

image

### DIFF
--- a/python/sglang/utils.py
+++ b/python/sglang/utils.py
@@ -624,7 +624,7 @@ class CachedKernel:
 
 def cached_triton_kernel(key_fn=None):
     """
-    Decorator that enables key-based caching for Triton kernels using a key function. 
+    Decorator that enables key-based caching for Triton kernels using a key function.
 
     It essentially bypasses Triton's built-in caching mechanism, allowing users to
     define their own caching strategy based on kernel parameters. This helps reduce
--- a/python/sglang/utils.py
+++ b/python/sglang/utils.py
@@ -624,7 +624,12 @@ class CachedKernel:
 
 def cached_triton_kernel(key_fn=None):
     """
-    Decorator that enables key-based caching for Triton kernels using a key function.
+    Decorator that enables key-based caching for Triton kernels using a key function. 
+
+    It essentially bypasses Triton's built-in caching mechanism, allowing users to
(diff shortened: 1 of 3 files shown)


python/sglang/utils.py
```diff
@@ -540,3 +543,114 @@ def resolve_obj_by_qualname(qualname: str) -> Any:
...
+            if cached_kernel is None:
+                # First time: compile and cache the kernel
+                cached_kernel = self.fn[grid](*args, **kwargs)
+                self.kernel_cache[cache_key] = cached_kernel
+                return cached_kernel
+            else:
+                # Use cached kernel
+                all_args = self._build_args(args, kwargs)
+                cached_kernel[grid](*all_args)
+                return cached_kernel
```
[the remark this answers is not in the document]
lifuhuang: This is wrong. Invocation actually returns compiled kernel instead of None.

zhyncs (approved):

python/sglang/utils.py
```diff
@@ -540,3 +543,114 @@ def resolve_obj_by_qualname(qualname: str) -> Any:
     module_name, obj_name = qualname.rsplit(".", 1)
     module = importlib.import_module(module_name)
     return getattr(module, obj_name)
+
+
+class CachedKernel:
```
zhyncs, line 548: Can we implement this in https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/utils.py instead

merrymercy: yes, please move  it to srt/utils.py
the `sglang/utils.py` is mostly for the language frontend

PR #10712, opened 2025-09-21
Fix NIXL OBJ desciptors

tshmilnvidia:
Fix mistake in creation of Nixl descriptors when using OBJ backend.


hnyls2002:
@trevor-m @ishandhanani Could you please take a look?

PR #10720, opened 2025-09-22
Intoduce cpu tensor as metadata to avoid blocking gpu kernel launch

AHEADer:
## Motivation
We observe some .item() operations in kv cache allocator, which will do cuda sync and block gpu kernel launching and make generation steps slower. We remove these operations by introducing some metadata tensor on cpu.

## Checklist


zhyncs (approved):

hnyls2002:
@AHEADer Great PR. A few things to discuss:
- It seems that we all agree the GPU tensors of `seq_lens` and `prefix_lens` are unnecessary during the scheduling/launch stage. In fact, keeping only GPU tensors for `seq_lens` without CPU copies goes against the original intention of `ScheduleBatch`—what we actually need in the scheduling stage are CPU values. Instead of maintaining a `seq_lens_cpu_cache`, we should maintain a `seq_lens_gpu_cache`.

- It looks like there are improvements when overlapping scheduling is disabled, which only applies in the speculative decoding scenario. However, the critical sync of [`acc_length_cpu`](https://github.com/sgl-project/sglang/pull/10720/files#diff-537e788456dd547527db3f78b6ee097a0adcd5437bf385a4ceeb84b465015cfaR401-R655) is not addressed in your case (and of course cannot be resolved). My guess is that the major improvement mainly comes from acceleration in the prefill stage, right?

AHEADer:
> @AHEADer Great PR. A few things to discuss:
> 
> * It seems that we all agree the GPU tensors of `seq_lens` and `prefix_lens` are unnecessary during the scheduling/launch stage. In fact, keeping only GPU tensors for `seq_lens` without CPU copies goes against the original intention of `ScheduleBatch`—what we actually need in the scheduling stage are CPU values. Instead of maintaining a `seq_lens_cpu_cache`, we should maintain a `seq_lens_gpu_cache`.
> * It looks like there are improvements when overlapping scheduling is disabled, which only applies in the speculative decoding scenario. However, the critical sync of [`acc_length_cpu`](https://github.com/sgl-project/sglang/pull/10720/files#diff-537e788456dd547527db3f78b6ee097a0adcd5437bf385a4ceeb84b465015cfaR401-R655) is not addressed in your case (and of course cannot be resolved). My guess is that the major improvement mainly comes from acceleration in the prefill stage, right?

You are kind of correct, this improves speculative decoding cases. `seq_lens_cpu_cache` is not introduced in my pr, you suggest I need to fix it? `accept_length_cpu` is just for updating `seq_lens_cpu`. We may have methods to avoid block syncing for accept_length_cpu, but will not in this pr?

hnyls2002 (approved):

PR #10721, opened 2025-09-22
make radix cache deterministic

skyzh:
## Motivation

The patch (tries) adding determinism to the radix cache. Part of https://github.com/sgl-project/sglang/issues/10278.

## Modifications

The patch mainly modifies the match_prefix logic:

- If it's called from the scheduler, always return the value to the multiply of SPLIT_SIZE.
- Otherwise, insert into the tree as normal.

The algorithm is two-pass:

- In the first pass, do the prefix matching as before.
- If we need to align to the split size, in the second pass, we use the nodes we collected from the previous pass to split exactly at the split size and return the node.

Note that this would do two splits in one match_prefix: one during the normal match to split at any prefix match, and another during we split at the split_size point; but it seems to work as desired. Please let me know if this doesn't work (memory leak?)

This patch also changes the test size for test_determinism's prefix size so that it can create more possibility of doing prefix matches in the cache.

## Accuracy Tests

```
python3 -m sglang.test.test_deterministic --test-mode prefix --temperature 0.7
```

gives 1 unique sample in all testable backends with Qwen-8B

## Checklist


skyzh:
slightly updated the test to have larger prefix so that we can trigger the code path to split at SPLIT_SIZE=1024

```
Testing Trial 12 with batch size 12, # prefix length 1: 3, # prefix length 4097: 2, # prefix length 5120: 2, # prefix length 8191: 5,
Prompt 0 with prefix length 1: total samples: 22, Unique samples: 1
Prompt 1 with prefix length 4097: total samples: 11, Unique samples: 1
Prompt 2 with prefix length 5120: total samples: 19, Unique samples: 3
Prompt 3 with prefix length 8191: total samples: 26, Unique samples: 3
```

Now we're getting a few non-determinism but still better than without the SPLIT_SIZE code path at all.

skyzh:
well, this non-determinism doesn't come from the radix cache: with `--disable-radix-cache` I can still get:

```
Testing Trial 12 with batch size 12, # prefix length 1: 4, # prefix length 4097: 2, # prefix length 5120: 2, # prefix length 8191: 4,
Prompt 0 with prefix length 1: total samples: 21, Unique samples: 1
Prompt 1 with prefix length 4097: total samples: 20, Unique samples: 1
Prompt 2 with prefix length 5120: total samples: 15, Unique samples: 3
Prompt 3 with prefix length 8191: total samples: 22, Unique samples: 2
```

with my slightly modified prefix lengths. guess we have some issues with other parts of the code.

skyzh:
(or`SGLANG_FLASHINFER_PREFILL_SPLIT_TILE_SIZE=1024` is not practical to test this)

skyzh:
This patch also includes https://github.com/sgl-project/sglang/pull/10637; I can close that one and keep this

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -168,13 +168,16 @@ def __init__(
...
+        enable_deterministic_inference: bool = False,
     ):
         self.req_to_token_pool = req_to_token_pool
         self.token_to_kv_pool_allocator = token_to_kv_pool_allocator
         self.page_size = page_size
         self.disable = disable
         self.enable_kv_cache_events = enable_kv_cache_events
         self.kv_event_queue = []
+        self.enable_deterministic_inference = enable_deterministic_inference
+        self.split_size = 4096  # see `init_deterministic_inference_config`
```
Fridge003: Can you get the split size from `init_deterministic_inference_config` ? 
I feel `init_deterministic_inference_config` can be moved to `python/sglang/srt/utils` as a util function

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -210,7 +213,9 @@ def reset(self):
         self.protected_size_ = 0
         self._record_all_cleared_event()
 
-    def match_prefix(self, key: RadixKey, **kwargs) -> MatchResult:
+    def match_prefix(
+        self, key: RadixKey, align_split_size: bool = True, **kwargs
```
Fridge003: The default value of `align_split_size` is better to be False, since we only need it for deterministic inference.

python/sglang/srt/managers/scheduler.py
```diff
@@ -753,6 +753,7 @@ def init_memory_pool_and_cache(self):
                     disable=server_args.disable_radix_cache,
```
Fridge003, line 846: https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/server_args.py#L1100

These few lines in server_args.py can be removed after this PR?

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -455,16 +470,56 @@ def _dfs_helper(node: TreeNode):
...
-    def _match_prefix_helper(self, node: TreeNode, key: RadixKey):
+    def _match_prefix_helper(
+        self, node: TreeNode, key: RadixKey, align_split_size: bool = True
+    ):
         node.last_access_time = time.monotonic()
 
         child_key = self.get_child_key_fn(key)
 
         value = []
+        match_history = [node]
```
Fridge003: We only need match_history when align_split_size is True. We can put it under `if align_split_size`

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -455,16 +470,56 @@ def _dfs_helper(node: TreeNode):
...
+                    value.append(node.value)
+            # return the root node as the corresponding node doesn't exist yet
+            # and the previously computed node is not at the split boundary
+            return [], match_history[0]
+
         while len(key) > 0 and child_key in node.children.keys():
             child = node.children[child_key]
             child.last_access_time = time.monotonic()
             prefix_len = self.key_match_fn(child.key, key)
+            match_history.append(child)
```
Fridge003: Also, we only need this when align_split_size is True

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -528,7 +590,7 @@ def _insert_helper(self, node: TreeNode, key: RadixKey, value):
             new_node.key = key
             new_node.value = value
             node.children[child_key] = new_node
-            self.evictable_size_ += len(value)
+            self.evictable_size_ += len(key)
```
Fridge003, line 669: Not sure about this line.
How do you think @xiezhq-hermann

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -455,16 +470,56 @@ def _dfs_helper(node: TreeNode):
...
+
+        if align_split_size and len(key) < self.split_size:
+            # fast path: directly return the root node if the split point is 0
+            return value, node
+
+        # use the access history to first find a split point at split_size and then return the value and node at that point.
+        def reconstruct_at_split_point(match_history, value_len):
+            # reverse the search process to find the last node right above the split_size, split here
+            split_point = value_len // self.split_size * self.split_size
+            print("value_len=", value_len, "split_point=", split_point)
```
Fridge003: please remove the printing log here

Fridge003:
Also please post the result of 
```bash
python3 -m sglang.test.test_deterministic --test-mode single
python3 -m sglang.test.test_deterministic --test-mode mixed
```
on both flashinfer and triton backends

When do the testing, please make sure the align size is the same as prefill_split_size/prefill_truncation_size

Fridge003:
@hnyls2002  @xiezhq-hermann 
This code changes some of the logics in radix cache, can you please help checking together.
Especially regarding potential memory leaks

skyzh:
```
python3 -m sglang.launch_server --model-path Qwen/Qwen3-8B --attention-backend flashinfer --enable-deterministic-inference
```

Flashinfer:

single - Total samples: 50, Unique samples: 1 ; the logic added in this patch is not triggered b/c all sequences are < split size, and therefore the cache always return un-match on prefill as part of the fast path (I assume this is expected behavior?)

mixed

```
Prompt 1: total samples: 583, Unique samples: 1
Prompt 2: total samples: 468, Unique samples: 1
Long prompt: total samples: 224, Unique samples: 1
```

same, I don't see the logic of reconstruction is triggered, probably due to the prompt length is too small

prefix

```
Prompt 0 with prefix length 1: total samples: 321, Unique samples: 1
Prompt 1 with prefix length 511: total samples: 332, Unique samples: 1
Prompt 2 with prefix length 2048: total samples: 311, Unique samples: 1
Prompt 3 with prefix length 4097: total samples: 311, Unique samples: 1
```

"prefix length 4097" does not yield a call to radix cache with >= 4096 (split size) requests, probably because of tokenization?

triton:

single - Total samples: 50, Unique samples: 1


Maybe it makes sense for me to increase the prompt length so that we can actually test the behavior of radix cache?

---

I'm resolving the comments meanwhile, thanks for reviews :)

python/sglang/srt/mem_cache/radix_cache.py
skyzh: The rationale here is that callers of match_prefix either comes from:

- decode, via `cache_unfinished_req`
- prefill, via direct call to `match_prefix`

if we put `align_split_size=True` here, we don't need to touch other code in the codebase; otherwise, we would find all callsites and set `align_split_size` to the desired value.

No strong preference; I feel either would work.

skyzh:
okay I need to do a benchmark again 🤣 I lost the changes to remove disable radix cache logic after rebasing

skyzh:
^ still not hitting the radix cache due to prompt length too small, should we update the tests?

python/sglang/srt/mem_cache/radix_cache.py
skyzh: https://github.com/sgl-project/sglang/pull/10721/commits/50cc9436ead57a20901c72c464be47df8a0d645e

Fridge003:
> ^ still not hitting the radix cache due to prompt length too small, should we update the tests?

Yes, we should try some longer prefices to test radix cache logic

skyzh:
with https://github.com/sgl-project/sglang/pull/10721/commits/e207d8ccc8143d61f1c16d41acaaca64eddd6f7c I can confirm the cache gets a lot of hits,

trition:

```
Testing Trial 50 with batch size 50, # prefix length 1: 13, # prefix length 2048: 11, # prefix length 10000: 10, # prefix length 20000: 16,
Prompt 0 with prefix length 1: total samples: 305, Unique samples: 1
Prompt 1 with prefix length 2048: total samples: 334, Unique samples: 1
Prompt 2 with prefix length 10000: total samples: 302, Unique samples: 1
Prompt 3 with prefix length 20000: total samples: 334, Unique samples: 1
```

flashinfer:

```
Testing Trial 50 with batch size 50, # prefix length 1: 13, # prefix length 2048: 10, # prefix length 10000: 13, # prefix length 20000: 14,
Prompt 0 with prefix length 1: total samples: 296, Unique samples: 1
Prompt 1 with prefix length 2048: total samples: 312, Unique samples: 1
Prompt 2 with prefix length 10000: total samples: 350, Unique samples: 1
Prompt 3 with prefix length 20000: total samples: 317, Unique samples: 1
```

Fridge003:
@skyzh Can you also provide some data for performance? Like how much it degrades from normal mode?

python/sglang/srt/mem_cache/radix_cache.py
xiezhq-hermann: this should be equivalent but be mindful about a recent refactoring: https://github.com/sgl-project/sglang/pull/10317

xiezhq-hermann:
Thank you @skyzh for the PR, it is indeed an interesting and important feature. But I am not sure whether enforcing a split is the best way of achieving this. Does it really guarantee determinism or more like reducing variances, especially if the split length is small. Would a large page size somehow achieving similar results?

skyzh:
Hi @xiezhq-hermann,

- when split size is small: yes, I tried split size like 128 and it can reduce non-determinism compared with not considering the split size at all. However, there're other parts of the system that is not determinism with the split size is very small.
- large page size: yes. The initial implementation of this patch simply sets page_size to split_size. But it leads to some memory mismatch (leak) errors. Before we hit those errors it gives good results. So I think it's better to consider both page_size and split_size in this patch.

skyzh:
(still gathering perf data, but my impression was that there're no significant speed different before/after enable if split_size is small; in other words, as we always align the result with the split size, when it's large, we have to decode more)

skyzh:
perf data: no significant difference with Qwen3-8B:

```
python3 -m sglang.test.test_deterministic --test-mode prefix --temperature 0.7
```

* with radix cache with deterministic, batch = 10: 3.3927013874053955 seconds
* with radix cache w/o deterministic, batch = 10: 3.073150634765625 seconds

hebiao064:
@Fridge003 @skyzh is it ready to merge?

skyzh:
@hebiao064 not yet - I have mid-confidence on whether this patch makes sense

Fridge003:
@hebiao064 Let's wait for opinions from @hnyls2002

python/sglang/test/test_deterministic.py
```diff
@@ -259,7 +259,7 @@ def test_deterministic(args):
 
     elif args.test_mode == "prefix":
         # In prefix mode, we create prompts from the same long prompt, with different lengths of common prefix.
-        len_prefix = [1, 511, 2048, 4097]
+        len_prefix = [1, 2048, 10000, 20000]
```
Fridge003: Just came up with one thing. The length of the long prompt is like 6500, will length 10000 and 20000 create the same prefix?

python/sglang/srt/mem_cache/radix_cache.py
hnyls2002: I suggest renaming align_split_size to something that reflects its usage in cache_unfinished_req (e.g., `do_not_align` or `is_cache_unfinished`), so that the flag clearly decouples the logic of alignment from the check for unfinished cached requests.

python/sglang/srt/mem_cache/radix_cache.py
hnyls2002: @skyzh The logic is the same, but why do you make this change?

hnyls2002 (changes requested):

python/sglang/srt/mem_cache/radix_cache.py
skyzh: This is part of https://github.com/sgl-project/sglang/pull/10637 - mostly for nit purposes; will rebase

python/sglang/test/test_deterministic.py
skyzh: I added an assertion `assert all(i <= len(LONG_PROMPT) for i in len_prefix)` and it seems to be true. This is the text size not the token size?

python/sglang/srt/mem_cache/radix_cache.py
skyzh: rebased and the diff doesn't exist

python/sglang/srt/mem_cache/radix_cache.py
skyzh: I took `is_cache_unfinished`

skyzh:
addressed the comments and ready for review again :) thanks!

python/sglang/test/test_deterministic.py
Fridge003: Ah, I think the length 6500 is token size. Then there is no mistake then

hnyls2002 (approved):

hnyls2002 (approved):

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -234,7 +239,9 @@ def reset(self):
         self.protected_size_ = 0
         self._record_all_cleared_event()
 
-    def match_prefix(self, key: RadixKey, **kwargs) -> MatchResult:
+    def match_prefix(
+        self, key: RadixKey, is_cache_unfinished: bool = False, **kwargs
+    ) -> MatchResult:
```
hanming-lu, line 244: let's not break API and use `kwargs` for `is_cache_unfinished`

python/sglang/srt/managers/scheduler.py
```diff
@@ -849,6 +846,7 @@ def init_memory_pool_and_cache(self):
...
                     enable_kv_cache_events=self.enable_kv_cache_events,
                     eviction_policy=server_args.radix_eviction_policy,
+                    enable_deterministic_inference=server_args.enable_deterministic_inference,
```
hanming-lu, line 849: It will be great to add asserts for other radix cache

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -534,16 +544,58 @@ def _dfs_helper(node: TreeNode):
...
+            not is_cache_unfinished and self.enable_deterministic_inference
+        )
+        match_history = [node] if align_split_size else None
+
+        if align_split_size and len(key) < self.split_size:
+            # fast path: directly return the root node if the split point is 0
+            return value, node
+
+        # use the access history to first find a split point at split_size and then return the value and node at that point.
+        def reconstruct_at_split_point(match_history, value_len):
```
hanming-lu, line 565: let's move this definition outside of `def _match_prefix_helper()`

PR #10735, opened 2025-09-22
Organize spec-related data structures

hnyls2002:
A step to merge #9334 into main and clarify all data structures for separate decoding/overlap scheduling.

Depends on #10715

***

- Introduce `SpecInput` class to unify `EagleDraftInput`, `EagleVerifyInput`, `LookaheadVerifyInput`, and more speculative algorithms' inputs
- Introduce `SpecInputType` to avoid direct assert `isinstance` in different attention backends and runners.
- Separate eagle info types out of `eagle_utils`.
- Move current `eagle_utils` to `spec_utils`.

This refactor is to support multiple speculative decoding algorithms better, making backends(attention, runner) not aware of the spec input type.


python/sglang/srt/disaggregation/decode_schedule_batch_mixin.py
```diff
@@ -149,7 +149,7 @@ def process_prebuilt_extend(
             hidden_states = torch.stack(hidden_states_list, dim=0).to(self.device)
 
             # local import to avoid circular import
-            from sglang.srt.speculative.eagle_utils import EagleDraftInput
+            from sglang.srt.speculative.spec_info import EagleDraftInput
```
zhyncs: Can we keep the old file to avoid breaking the downstream fork?

zhyncs (changes requested):

python/sglang/srt/disaggregation/decode_schedule_batch_mixin.py
hnyls2002: I can remove inputs of Eagle (`EagleDraftInput`, `EagleVerifyInput`) back to eagle_utils, but will still get an abstraction class `SpecInput` to unify them.

hnyls2002: We have to split the common part of speculative utils and speculative dataclasses into another module, finally. Current `spec_info` and `eagle_utils` are too messy.

python/sglang/srt/speculative/spec_info.py
```diff
@@ -35,3 +81,699 @@ def from_string(name: str):
...
+                    capture_hidden_mode=CaptureHiddenMode.LAST,
+                )
+
+            return EagleVerifyOutput(
+                draft_input=draft_input,
+                logits_output=logits_output,
+                verified_id=verified_id,
+                accept_length_per_req_cpu=accept_length_cpu,
+                accepted_indices=accept_index,
+            )
```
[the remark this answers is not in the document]
hnyls2002: Yes, because it is too complex, I choose not to modify it for now.

PR #10751, opened 2025-09-22
Add auth to get server info

Muqi1029:
## Motivation

Getting server information requires proper authorization.

## Modifications

Simply add the necessary headers to the HTTP GET request to ensure authorized access.

## Checklist


hnyls2002:
@Muqi1029 Could you also help to shift the use environ from `sglang.srt.environ`?

```
def get_auth_headers() -> Dict[str, str]:
    openai_api_key = os.environ.get("OPENAI_API_KEY")
    if openai_api_key:
        return {"Authorization": f"Bearer {openai_api_key}"}
    else:
        api_key = os.environ.get("API_KEY")
        if api_key:
            return {"Authorization": f"{api_key}"}
        return {}
```

Muqi1029:
@hnyls2002  Thanks for your review and advice!

However, `bench_serving.py` is intended to be a lightweight client script for sending HTTP requests.

If we import `sglang.srt.environ`, the script would introduce an unnecessary dependency on sglang. In scenarios where a user wants to run benchmarks in a minimal Python environment (e.g., inside a container or pod), installing sglang may not be straightforward. The idea is that users can simply copy this script and run it independently. For this reason, I think it may be better to stick with plain Python functions here.

JustinTong0323 (approved):
LGTM.

JustinTong0323:
@hnyls2002 Could you help merge it directly? It is a simple fix and wouldn't have side effects.

PR #10770, opened 2025-09-23
Clean up server args

merrymercy:
- Do not use "Step 1", "Step 2", ... , as it will need lots of changes if we insert something in the middle
- rename `model_specific_adjustments` -> `_handle_model_specific_adjustments`
- Sort the names and functions better
- Deprecate old arguments `enable_ep_moe`, `enable_deepep_moe`, ..


zhyncs (commented):
```
Deprecate old arguments enable_ep_moe, enable_deepep_moe, ..
```

@ch-wan @merrymercy Should we maintain internal logic for compatibility reasons

merrymercy:
These old arguments `enable_ep_moe` are marked as deprecated with warnings for several weeks, so I think it is okay to deprecate them. We still raised helpful error messages for users to migrate.

PR #10780, opened 2025-09-23
Fix [test]: Env:SGLANG_TORCH_PROFILER_DIR for pytest.

singhalshubham03:
## Motivation

Running sglang unit tests on XPU currently fails under pytest because the SGLANG_TORCH_PROFILER_DIR environment variable is only set in the __main__ block. When executed via pytest, the __main__ block is not triggered, leading to missing profiler output directory setup. However, when running with unittest directly, the test passes since the variable is set in that execution path.
## Modifications

Moved the os.environ["SGLANG_TORCH_PROFILER_DIR"] = OUTPUT_DIR assignment into the setUpClass method of TestStartProfile.

Ensures the environment variable is always set, regardless of whether the test is executed via pytest or unittest.
## Accuracy Tests

Verified that the test now passes when executed with both pytest and unittest.
## Benchmarking and Profiling

No accuracy or performance regressions introduced.

Only affects test setup to make profiling directory consistently available.
## Checklist

### DIFF
--- a/python/sglang/srt/environ.py
+++ b/python/sglang/srt/environ.py
@@ -126,6 +126,7 @@ class Envs:
     SGLANG_DISABLE_REQUEST_LOGGING = EnvBool(False)
     SGLANG_SIMULATE_ACC_LEN = EnvFloat(-1)
     SGLANG_SIMULATE_ACC_METHOD = EnvStr("multinomial")
+    SGLANG_TORCH_PROFILER_DIR = EnvStr("/tmp")
 
     # Model Parallel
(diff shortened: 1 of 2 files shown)


hnyls2002:
@singhalshubham03 Use `sglang.environ` to set the envs.

singhalshubham03:
> @singhalshubham03 Use `sglang.environ` to set the envs.

Resolved

hnyls2002:
@singhalshubham03 lint failed....

singhalshubham03:
lint passed

PR #10787, opened 2025-09-23
fix: draft model IMA by overide max_positional_embeddings

JustinTong0323:
## Motivation

Fix #10713
Root cause is the rope cache calculated by model's `max_positional_embeddings`, which is default to [2048](https://huggingface.co/lmsys/sglang-EAGLE-LLaMA3-Instruct-8B/blob/main/config.json#:~:text=%22max_position_embeddings%22%3A-,2048,-%2C). So if the input(after batching) longer then this number, IMA occurs.
 
https://github.com/sgl-project/sglang/blob/ea338676b5024188db7f20e54a80b06c8effd72e/python/sglang/srt/layers/rotary_embedding.py#L132-L141

## Checklist


zhyncs:
@JustinTong0323 can you add ut for this

JustinTong0323:
Maybe merge with #10788

python/sglang/srt/configs/model_config.py
```diff
@@ -210,6 +210,11 @@ def __init__(
                 ):
                     logger.warning(msg)
                     self.context_len = context_length
+                    if is_draft_model:
```
hnyls2002, line 213: `get_bool_env_var("SGLANG_ALLOW_OVERWRITE_LONGER_CONTEXT_LEN")` get environ var from `sglang.srt.environ`

hnyls2002 (changes requested):

hnyls2002 (approved):

PR #10788, opened 2025-09-23
Fix: Dynamic RoPE Cache Expansion to Prevent Position-ID Out-of-Bounds in EAGLE + Long-Sequence Workloads

YAMY1234:
### **Title**

Fix: Dynamic RoPE Cache Expansion to Prevent Position-ID Out-of-Bounds in EAGLE + Long-Sequence Workloads

---

### **Background**

Running SGLang with **EAGLE speculative decoding** and long sequences occasionally triggers

```
RuntimeError: BatchQKApplyRotaryPosIdsCosSinCache failed with error code an illegal memory access was encountered
```

Root cause analysis shows that the crash occurs **whenever a generated `position_id` exceeds the length of the RoPE `cos_sin_cache` tensor**, which is initialized to `max_position_embeddings`.

* Mismatch between the target and draft model’s `max_position_embeddings` increases the likelihood, the *fundamental* issue is that the cache can be too short for real runtime positions, especially with speculative multi-step decoding and batching.

---

### **What This PR Does**

* **Thread-Safe Dynamic Expansion**

  * Adds `_ensure_cos_sin_cache_length()` to `RotaryEmbedding` for on-demand, cache growth.

* **Device-Aware Initialization**

  * Builds initial `cos_sin_cache` directly on the current CUDA device to prevent CPU–GPU copy overhead.

---

### **Expansion Policy**

The recommended pre-expansion formula (handled in the model runner before graph capture) is:

```
reserve_len = max(target_max_pos, draft_max_pos)
            + speculative_num_steps * speculative_num_draft_tokens
            + safety_margin  # (~1k)
```

This covers:

* Target/draft model declared maxima,
* Multi-step speculative decoding,
* A configurable buffer for future changes.

---

### **Results & Verification**

* Stress-tested with **1,200+ input tokens**, `speculative_num_steps=5`, and high concurrency.
* No illegal memory access observed.
* Single, one-time memory allocation; no runtime re-allocations during inference.
* Negligible performance impact—graph capture proceeds without synchronization penalties.

---

### **Why This Matters**

Any under-provisioned RoPE cache will still break under long or speculative workloads.
This PR makes SGLang **robust regardless of model mismatch** and future-proofs against larger context windows.

---

**In short:**

> **Before:** RoPE cache length fixed at model config → runtime overflow possible.
> **After:** RoPE cache auto-expands safely and transparently → no illegal memory access.

---

**Closes:** #10713 (and related long-sequence RoPE overflow issues)

---


python/sglang/srt/model_executor/model_runner.py
```diff
@@ -838,6 +841,61 @@ def load_model(self):
                 f"TP rank {self.tp_rank} could finish the model loading, but there are other ranks that didn't finish loading. It is likely due to unexpected failures (e.g., OOM) or a slow node."
             ) from None
 
+    def _reserve_rope_cache_for_long_sequences(self):
```
hnyls2002: Move this function to utils.

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -838,6 +841,61 @@ def load_model(self):
...
+        """Pre-expand RoPE cache for long sequences and speculative decoding."""
+        import os
+
+        SAFETY_FACTOR = int(os.getenv("SGLANG_SPEC_EXPANSION_SAFETY_FACTOR", "2"))
```
hnyls2002: Use environ with `sglang.srt.environ`

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -838,6 +841,61 @@ def load_model(self):
...
+        MARGIN = int(os.getenv("SGLANG_ROPE_CACHE_SAFETY_MARGIN", "1024"))
```
hnyls2002: margin too long? use a smaller proper value.

hnyls2002 (changes requested):

python/sglang/srt/model_executor/model_runner.py
YAMY1234: Done!

python/sglang/srt/model_executor/model_runner.py
YAMY1234: Done!

hnyls2002 (approved):

PR #10804, opened 2025-09-23
[Feature] Sglang Tracing: Fine-Grained Tracking for Request Latency - Part 2

sufeng-buaa:
## Motivation

The PR is response to https://github.com/sgl-project/sglang/issues/8965. For details on the motivation and visual output, please refer to the issue.

## Modifications
The PR have split the patch into two parts.
First part is here: https://github.com/sgl-project/sglang/pull/9962, which has been merged.

This is the second part, which include:
* Request tracing support for PD disaggregation.
* Request tracing support for DP attention scenarios.
* A script for converting OpenTelemetry data to Perfetto data 

Router tracing is currently only implemented for mini_lb, since it is implemented in Python. Support for the Rust-based router will be added in a follow-up commit, after I finish implementing the tracing package in Rust.

Building upon Part 1, to accommodate PD disaggregation, I have upgraded the original three-level span structure to a four-level hierarchy by adding a top-level bootstrap_room_span. While the previous three-level structure could still achieve the original design goals, this change was made with future extensibility in mind. Specifically, we may want to attach certain attribute information to the request root span in the future; however, OpenTelemetry does not allow adding attributes to spans that are propagated from other nodes. Therefore, the updated span hierarchy is as follows:
```
bootstrap room span
├── router req root span
|    └── router thread span
|          └── slice span
├── prefill req root span
|    ├── tokenizer thread span
|    |     └── slice span
|    └── scheduler thread span
|          └── slice span
└── decode req root span
      ├── tokenizer thread span
      |    └── slice span
      └── scheduler thread span
           └── slice span
```

## Checklist

### DIFF
--- a/python/sglang/srt/disaggregation/decode.py
+++ b/python/sglang/srt/disaggregation/decode.py
@@ -787,7 +787,7 @@ class DecodeTransferQueue:
                     )
                     self.tree_cache.cache_finished_req(decode_req.req)
                     trace_slice_end(
-                        "quick_finish", decode_req.req.rid, thread_finish_flag=True
+                        RequestStage.DECODE_QUICK_FINISH, decode_req.req.rid, thread_finish_flag=True
                     )
                 else:
                     transferred_reqs.append(decode_req.req)
--- a/python/sglang/srt/disaggregation/decode.py
+++ b/python/sglang/srt/disaggregation/decode.py
@@ -58,7 +58,12 @@ from sglang.srt.mem_cache.memory_pool import (
     ReqToTokenPool,
     SWAKVPool,
 )
-from sglang.srt.tracing.trace import trace_event, trace_slice, trace_slice_end, trace_slice_batch
+from sglang.srt.tracing.trace import (
+    trace_event_batch,
+    trace_slice,
+    trace_slice_batch,
+    trace_slice_end,
+)
 from sglang.srt.utils import get_int_env_var, require_mlp_sync
 from sglang.srt.utils.torch_memory_saver_adapter import TorchMemorySaverAdapter
 
@@ -972,8 +977,7 @@ class SchedulerDisaggregationDecodeMixin:
 
         if ret:
             attrs = {"bid": hex(id(ret)), "batch_size": ret.batch_size()}
-            for req in ret.reqs:
-                trace_event("schedule", req.rid, attrs=attrs)
+            trace_event_batch("schedule", ret.reqs, attrs=attrs)
         return ret
 
     def get_new_prebuilt_batch(self: Scheduler) -> Optional[ScheduleBatch]:
--- a/python/sglang/srt/disaggregation/decode.py
+++ b/python/sglang/srt/disaggregation/decode.py
@@ -971,8 +971,9 @@ class SchedulerDisaggregationDecodeMixin:
                 ret = self.running_batch if not self.running_batch.is_empty() else None
 
         if ret:
+            attrs = {"bid": hex(id(ret)), "batch_size": ret.batch_size()}
             for req in ret.reqs:
-                trace_event("schedule", req.rid)
+                trace_event("schedule", req.rid, attrs=attrs)
         return ret
 
     def get_new_prebuilt_batch(self: Scheduler) -> Optional[ScheduleBatch]:
--- a/python/sglang/srt/disaggregation/decode.py
+++ b/python/sglang/srt/disaggregation/decode.py
@@ -58,6 +58,7 @@ from sglang.srt.mem_cache.memory_pool import (
     ReqToTokenPool,
     SWAKVPool,
 )
+from sglang.srt.tracing.trace import trace_event, trace_slice, trace_slice_end, trace_slice_batch
 from sglang.srt.utils import get_int_env_var, require_mlp_sync
 from sglang.srt.utils.torch_memory_saver_adapter import TorchMemorySaverAdapter
 
@@ -313,6 +314,7 @@ class DecodePreallocQueue:
             )
 
             req.add_latency(RequestStage.DECODE_PREPARE)
+            trace_slice_end(RequestStage.DECODE_PREPARE, req.rid, auto_next_anon=True)
             self.queue.append(
                 DecodeRequest(req=req, kv_receiver=kv_receiver, waiting_for_input=False)
             )
@@ -528,6 +530,9 @@ class DecodePreallocQueue:
                 time.perf_counter()
             )
             decode_req.req.add_latency(RequestStage.DECODE_BOOTSTRAP)
+            trace_slice_end(
+                RequestStage.DECODE_BOOTSTRAP, decode_req.req.rid, auto_next_anon=True
+            )
 
         self.queue = [
             entry for i, entry in enumerate(self.queue) if i not in indices_to_remove
@@ -776,8 +781,17 @@ class DecodeTransferQueue:
                         [decode_req.req], decode_req.req.return_logprob
                     )
                     self.tree_cache.cache_finished_req(decode_req.req)
+                    trace_slice_end(
+                        "quick_finish", decode_req.req.rid, thread_finish_flag=True
+                    )
                 else:
                     transferred_reqs.append(decode_req.req)
+                    trace_slice_end(
+                        RequestStage.DECODE_TRANSFERRED,
+                        decode_req.req.rid,
+                        auto_next_anon=True,
+                    )
+
             elif poll in [
                 KVPoll.Bootstrapping,
                 KVPoll.WaitingForInput,
@@ -823,6 +837,7 @@ class SchedulerDisaggregationDecodeMixin:
                     self.stream_output(
                         batch.reqs, any(req.return_logprob for req in batch.reqs)
                     )
+                    trace_slice_batch(RequestStage.DECODE_FAKE_OUTPUT, batch.reqs)
                     if prepare_mlp_sync_flag:
                         self._prepare_idle_batch_and_run(None)
                 else:
@@ -872,6 +887,7 @@ class SchedulerDisaggregationDecodeMixin:
                     self.stream_output(
                         batch.reqs, any(req.return_logprob for req in batch.reqs)
                     )
+                    trace_slice_batch(RequestStage.DECODE_FAKE_OUTPUT, batch.reqs)
                     if prepare_mlp_sync_flag:
                         batch_, batch_result = self._prepare_idle_batch_and_run(
                             None, delay_process=True
@@ -954,6 +970,9 @@ class SchedulerDisaggregationDecodeMixin:
                 self.running_batch = self.update_running_batch(self.running_batch)
                 ret = self.running_batch if not self.running_batch.is_empty() else None
 
+        if ret:
+            for req in ret.reqs:
+                trace_event("schedule", req.rid)
         return ret
 
     def get_new_prebuilt_batch(self: Scheduler) -> Optional[ScheduleBatch]:
--- a/python/sglang/srt/disaggregation/decode.py
+++ b/python/sglang/srt/disaggregation/decode.py
@@ -60,7 +60,6 @@ from sglang.srt.mem_cache.memory_pool import (
 )
 from sglang.srt.tracing.trace import (
     trace_event_batch,
-    trace_slice,
     trace_slice_batch,
     trace_slice_end,
 )
@@ -787,7 +786,9 @@ class DecodeTransferQueue:
                     )
                     self.tree_cache.cache_finished_req(decode_req.req)
                     trace_slice_end(
-                        RequestStage.DECODE_QUICK_FINISH, decode_req.req.rid, thread_finish_flag=True
+                        RequestStage.DECODE_QUICK_FINISH,
+                        decode_req.req.rid,
+                        thread_finish_flag=True,
                     )
                 else:
                     transferred_reqs.append(decode_req.req)
--- a/python/sglang/srt/managers/schedule_batch.py
+++ b/python/sglang/srt/managers/schedule_batch.py
@@ -392,13 +392,20 @@ class MultimodalInputs:
 
 
 class RequestStage(str, enum.Enum):
-    # prefill
+    # Tokenizer
+    TOKENIZE = "tokenize"
+    TOKENIZER_DISPATCH = "dispatch"
+
+    # common/non-disaggregation
     PREFILL_WAITING = "prefill_waiting"
+    REQUEST_PROCESS = "request_process"
+    DECODE_LOOP = "decode_loop"
+    PREFILL_FORWARD = "prefill_forward"
+    PREFILL_CHUNKED_FORWARD = "chunked_prefill"
 
     # disaggregation prefill
     PREFILL_PREPARE = "prefill_prepare"
     PREFILL_BOOTSTRAP = "prefill_bootstrap"
-    PREFILL_FORWARD = "prefill_forward"
     PREFILL_TRANSFER_KV_CACHE = "prefill_transfer_kv_cache"
 
     # disaggregation decode
@@ -406,6 +413,7 @@ class RequestStage(str, enum.Enum):
     DECODE_BOOTSTRAP = "decode_bootstrap"
     DECODE_WAITING = "decode_waiting"
     DECODE_TRANSFERRED = "decode_transferred"
+    DECODE_FAKE_OUTPUT = "fake_output"
 
 
 class Req:
--- a/python/sglang/srt/managers/schedule_batch.py
+++ b/python/sglang/srt/managers/schedule_batch.py
@@ -417,6 +417,7 @@ class RequestStage(str, enum.Enum):
     DECODE_WAITING = "decode_waiting"
     DECODE_TRANSFERRED = "decode_transferred"
     DECODE_FAKE_OUTPUT = "fake_output"
+    DECODE_QUICK_FINISH = "quick_finish"
 
 
 class Req:
--- a/python/sglang/srt/managers/schedule_batch.py
+++ b/python/sglang/srt/managers/schedule_batch.py
@@ -396,6 +396,9 @@ class RequestStage(str, enum.Enum):
     TOKENIZE = "tokenize"
     TOKENIZER_DISPATCH = "dispatch"
 
+    # DP controller
+    DC_DISPATCH = "dc_dispatch"
+
     # common/non-disaggregation
     PREFILL_WAITING = "prefill_waiting"
     REQUEST_PROCESS = "request_process"
--- a/python/sglang/srt/managers/scheduler.py
+++ b/python/sglang/srt/managers/scheduler.py
@@ -1376,7 +1376,7 @@ class Scheduler(
             self._prefetch_kvcache(req)
             self.waiting_queue.append(req)
             req.time_stats.wait_queue_entry_time = time.perf_counter()
-            trace_slice_end("process req", req.rid, auto_next_anon=True)
+            trace_slice_end(RequestStage.REQUEST_PROCESS, req.rid, auto_next_anon=True)
         elif self.disaggregation_mode == DisaggregationMode.PREFILL:
             self._prefetch_kvcache(req)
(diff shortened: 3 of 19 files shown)


python/sglang/srt/disaggregation/decode.py
```diff
@@ -697,6 +710,12 @@ def event_loop_normal_disagg_decode(self: Scheduler):
                     self.stream_output(
                         batch.reqs, any(req.return_logprob for req in batch.reqs)
                     )
+                    for req in batch.reqs:
+                        trace_slice(
+                            RequestStage.DECODE_FAKE_OUTPUT,
+                            reqs.rid,
+                            auto_next_anon=True,
+                        )
```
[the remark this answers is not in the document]
sufeng-buaa: fixed

sgl-router/py_src/sglang_router/mini_lb.py
```diff
@@ -91,11 +100,40 @@ async def generate(
...
+                for bootstrap_room in modified_request["bootstrap_room"]:
+                    trace_slice_end(
+                        "mini lb luanch", bootstrap_room, auto_next_anon=True
+                    )
+            elif isinstance(modified_request["bootstrap_room"], int):
+                trace_slice_end(
+                    "mini lb luanch",
+                    modified_request["bootstrap_room"],
+                    auto_next_anon=True,
+                )
```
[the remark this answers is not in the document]
sufeng-buaa: fixed

sgl-router/py_src/sglang_router/mini_lb.py
```diff
@@ -91,11 +100,31 @@ async def generate(
...
+                ),
+                session.post(
+                    f"{decode_server}/{endpoint}",
+                    json=modified_request,
+                    headers=headers,
+                ),
             ]
 
+            for bootstrap_room in bootstrap_room_list:
+                trace_slice_end("mini lb luanch", bootstrap_room, auto_next_anon=True)
```
acelyc111:
```suggestion
                trace_slice_end("mini lb launch", bootstrap_room, auto_next_anon=True)
```

acelyc111: Other places are the same.

sgl-router/py_src/sglang_router/mini_lb.py
```diff
@@ -131,10 +168,36 @@ async def stream_results():
...
+                        json=modified_request,
+                        headers=headers,
+                    ),
                 ]
+                if isinstance(modified_request["bootstrap_room"], list):
+                    for bootstrap_room in modified_request["bootstrap_room"]:
+                        trace_slice_end(
+                            "mini lb luanch", bootstrap_room, auto_next_anon=True
+                        )
+                for bootstrap_room in bootstrap_room_list:
```
acelyc111, line 199: If `modified_request["bootstrap_room"]` is a list, the `bootstrap_room_list` is a list with the same content, then the following `trace_slice_end` will be executed twice, is it expected?

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -413,20 +413,31 @@ def merge(self, other: MultimodalInputs):
 
 
 class RequestStage(str, enum.Enum):
-    # prefill
+    # Tokenizer
+    TOKENIZE = "tokenize"
+    TOKENIZER_DISPATCH = "dispatch"
+
+    # DP controller
+    DC_DISPATCH = "dc dispatch"
```
acelyc111: It make sense to add fine grained stages, and these new stage names are also easy to be added as Prometheus label names as they did. But there are some limitation for naming, for example, spaces are not allowed, see https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels. It would be better to keep compatibility with Prometheus naming rules.

sgl-router/py_src/sglang_router/mini_lb.py
sufeng-buaa: Ah sorry, I didn't completely clean up the obsolete code. 
Fixed.

python/sglang/srt/managers/schedule_batch.py
sufeng-buaa: Fixed

sufeng-buaa:
Hi @ishandhanani , could you please review this PR when you have a moment? Thanks!

sgl-router/py_src/sglang_router/mini_lb.py
sufeng-buaa: Fixed

lun-4:
hello! I have a question, I've looked at both #9962 and this PR but it's unclear to me if there's support for receiving trace/span IDs via the `traceparent` header (through [opentelemetry context propagation](https://opentelemetry.io/docs/languages/python/propagation/), manually or automatically via [fastapi instrumentation](https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/fastapi/fastapi.html)) added or planned to be added in some future PR? in my deployment I'm using sglang behind a service which does authentication and other roles, and it already initializes a lot of important span metadata that would be welcome to link to sglang's own traces.

EDIT: looks like not. just made a quick patch for it in https://github.com/sgl-project/sglang/pull/11074

merrymercy:
Please also pay extreme attention to overhead https://github.com/sgl-project/sglang/pull/9962#discussion_r2393423009

sufeng-buaa:
> hello! I have a question, I've looked at both #9962 and this PR but it's unclear to me if there's support for receiving trace/span IDs via the `traceparent` header (through [opentelemetry context propagation](https://opentelemetry.io/docs/languages/python/propagation/), manually or automatically via [fastapi instrumentation](https://opentelemetry-python-contrib.readthedocs.io/en/latest/instrumentation/fastapi/fastapi.html)) added or planned to be added in some future PR? in my deployment I'm using sglang behind a service which does authentication and other roles, and it already initializes a lot of important span metadata that would be welcome to link to sglang's own traces.
> 
> EDIT: looks like not. just made a quick patch for it in #11074

 Is this PR https://github.com/sgl-project/sglang/pull/10808 what you mean?

sufeng-buaa:
> Please also pay extreme attention to overhead [#9962 (comment)](https://github.com/sgl-project/sglang/pull/9962#discussion_r2393423009)

OK, thank you for the review. I will fix it as soon as possible.

sufeng-buaa:
> Please also pay extreme attention to overhead [#9962 (comment)](https://github.com/sgl-project/sglang/pull/9962#discussion_r2393423009)

@merrymercy  I have updated the code according to your suggestions. Could you please take a look again?

ishandhanani:
Hi @sufeng-buaa - curious on if you've looked at https://grafana.com/oss/tempo/. We use grafana for our metrics in SGL already and tempo plays well with the grafana stack. 

No need to change/update anything just sharing in case it's useful :)

ishandhanani (approved):
Overall this PR looks good to me. 

In the previous PR, you shared some numbers on how much overhead each span added. Could you run the same setup here to ensure that that has not changed. Additionally - can you share/provide some code on how you setup this benchmarking to measure the overhead? 

It would be good to have as a spot test in the future

sufeng-buaa:
> Hi @sufeng-buaa - curious on if you've looked at https://grafana.com/oss/tempo/. We use grafana for our metrics in SGL already and tempo plays well with the grafana stack.
> 
> No need to change/update anything just sharing in case it's useful :)

Thanks for sharing! I'm aware of Grafana, but I do not know its support for trace span visualization. However, as long as it can integrate with OpenTelemetry Collector, we won't need to modify our trace instrumentation code. We'd only need to switch our frontend tool from Jaeger to Tempo.

sufeng-buaa:
> Overall this PR looks good to me.
> 
> In the previous PR, you shared some numbers on how much overhead each span added. Could you run the same setup here to ensure that that has not changed. Additionally - can you share/provide some code on how you setup this benchmarking to measure the overhead?
> 
> It would be good to have as a spot test in the future

ok, I'll post some overhead data here. I don't have a dedicated benchmark for measuring overhead – I simply record timestamps before/after each instrumentation point to evaluate per-span cost, and enable torch profiling to monitor for unexpected overhead. In addition, I use sglang.bench_serving to test whether TTFT, TPOT, and throughput change with tracing enabled/disabled. The steps to enable trace can be found in docs/references/production_request_trace.md.

sufeng-buaa:
### Instrumentation Overhead Evaluation
The test environment is the same as https://github.com/sgl-project/sglang/pull/9962

Compared to https://github.com/sgl-project/sglang/pull/9962, the overhead of single-span generation and trace context propagation across threads remains unchanged. This section adds support for request tracing in the PD Disaggregation scenario, with an overhead of approximately 60 μs for trace context propagation across nodes (from mini_lb to prefill or decode nodes).

Additionally, the span structure has been extended to four levels, increasing the overhead of trace_req_start() to approximately 95 μs.

python/sglang/srt/disaggregation/decode.py
```diff
@@ -675,8 +685,17 @@ def pop_transferred(self) -> List[Req]:
                         [decode_req.req], decode_req.req.return_logprob
                     )
                     self.tree_cache.cache_finished_req(decode_req.req)
+                    trace_slice_end(
+                        "quick_finish", decode_req.req.rid, thread_finish_flag=True
```
acelyc111: How about defining a new enum value for "quick_finish"?

acelyc111 (approved):
LGTM. We've tested on our environment and it works.

image

python/sglang/srt/disaggregation/decode.py
sufeng-buaa: ok, I have added it.

python/sglang/srt/managers/scheduler.py
```diff
@@ -2303,13 +2308,10 @@ def process_batch_result(
     ):
         if batch.forward_mode.is_decode():
             self.process_batch_result_decode(batch, result)
-            if self.enable_trace:
-                trace_slice_batch("decode loop", batch.reqs)
+            trace_slice_batch(RequestStage.DECODE_LOOP, batch.reqs)
```
ShangmingCai, line 2020: Why removing `if self.enable_trace`? 
IIUC, are we trying to use the global param `tracing_enabled` in all the utils of tracing module:
```python
    if not tracing_enabled:
        return
```
to avoid placing `if self.enable_trace` or `if self.server_args.enable_trace` everywhere?

sufeng-buaa: I have moved `if enable_trace` to `trace_slice_batch`.

There are still some places where tracing functions need to be executed in batch, but it seems unnecessary to write a dedicated batch function for these cases. To avoid performing loop-based condition checks when tracing is disabled, an `if self.enable_trace` guard is added at the outer layer.

`self.server_args.enable_trace` cannot be deleted anyway, so there is no need to import the global variable `tracing_enable` specifically?

ShangmingCai: OK, as long as it won't bring a performance drop when tracing is not enabled. If a simple `if` can do this, we should avoid checking it in function calling.

python/sglang/srt/tracing/trace.py
```diff
@@ -569,10 +653,26 @@ def trace_slice_batch(
...
+    name: str,
+    reqs: List[Req],
+    ts: Optional[int] = None,
+    attrs: Dict[str, Any] = None,
+):
+    if not tracing_enabled:
+        return
+
+    for req in reqs:
+        trace_event(name, req.rid, ts=ts, attrs=attrs)
```
acelyc111, line 687: It should be noticed that the larger the batch size is, the more overhead it would cause, they are in linear relationship. Maybe we can optimize this in following patches.

ishandhanani: This is a good thought. Any ideas @sufeng-buaa ?

sufeng-buaa: Yes, I've also thought about this issue. I considered generating only one span per batch, but since requests in a batch are not fixed, I can only create spans and events at the request level for now, and I haven't yet figured out the best way to address this. Currently, I've added a conditional check at the outer level so that it at least doesn't impact performance when tracing is disabled. Moreover, thanks to the overlap scheduling mechanism, there's no significant overhead when batch size is less than 128, which I believe covers most practical scenarios.

ShangmingCai:
@sufeng-buaa Can you resolve the conflicts? I think fixing some imports should be work.

sufeng-buaa:
> @sufeng-buaa Can you resolve the conflicts? I think fixing some imports should be work.

fixed.

ollybbmonster:
> ### Instrumentation Overhead Evaluation
> The test environment is the same as #9962
> 
> Compared to #9962, the overhead of single-span generation and trace context propagation across threads remains unchanged. This section adds support for request tracing in the PD Disaggregation scenario, with an overhead of approximately 60 μs for trace context propagation across nodes (from mini_lb to prefill or decode nodes).
> 
> Additionally, the span structure has been extended to four levels, increasing the overhead of trace_req_start() to approximately 95 μs.

Hi, did you consider the exporter thread overhead? I'm not sure if it's caused by GIL, but it seems that the exporter thread blocks the scheduler thread when encoding spans in otel sdk, making the overall overhead not fully overlapped in many scenarios.

sufeng-buaa:
> > ### Instrumentation Overhead Evaluation
> > The test environment is the same as #9962
> > Compared to #9962, the overhead of single-span generation and trace context propagation across threads remains unchanged. This section adds support for request tracing in the PD Disaggregation scenario, with an overhead of approximately 60 μs for trace context propagation across nodes (from mini_lb to prefill or decode nodes).
> > Additionally, the span structure has been extended to four levels, increasing the overhead of trace_req_start() to approximately 95 μs.
> 
> Hi, did you consider the exporter thread overhead? I'm not sure if it's caused by GIL, but it seems that the exporter thread blocks the scheduler thread when encoding spans in otel sdk, making the overall overhead not fully overlapped in many scenarios.

Great Question — I actually ran into this issue during high-load testing and spent several weeks tracking it down. The main problem was that the exporter's asynchronous export cycle was too long, causing a sudden spike in generating a large amount of garbage, which in turn triggered prolonged GC pauses that blocked the scheduler thread.
I've since addressed this by tuning the `schedule_delay_millis` and `max_export_batch_size` parameters to make exports more frequent but smaller (https://github.com/sgl-project/sglang/pull/10804/commits/bbd10e78a06b8caa255d50ddcf3f088a96637d6f). This helps prevent garbage collection spikes and significantly reduces the risk of blocking the scheduler. While this may increase CPU usage slightly, in most LLM deployment environments CPU resources are typically underutilized, so the trade-off is well worth the improved latency and stability.

python/sglang/srt/tracing/trace.py
```diff
@@ -194,7 +198,9 @@ def process_tracing_init(otlp_endpoint, server_name):
         )
 
         processor = BatchSpanProcessor(
-            OTLPSpanExporter(endpoint=otlp_endpoint, insecure=True)
+            OTLPSpanExporter(endpoint=otlp_endpoint, insecure=True),
+            schedule_delay_millis=500,
+            max_export_batch_size=64,
```
acelyc111: How about using environment variables?

acelyc111:
> > > ### Instrumentation Overhead Evaluation
> > > The test environment is the same as #9962
> > > Compared to #9962, the overhead of single-span generation and trace context propagation across threads remains unchanged. This section adds support for request tracing in the PD Disaggregation scenario, with an overhead of approximately 60 μs for trace context propagation across nodes (from mini_lb to prefill or decode nodes).
> > > Additionally, the span structure has been extended to four levels, increasing the overhead of trace_req_start() to approximately 95 μs.
> > 
> > 
> > Hi, did you consider the exporter thread overhead? I'm not sure if it's caused by GIL, but it seems that the exporter thread blocks the scheduler thread when encoding spans in otel sdk, making the overall overhead not fully overlapped in many scenarios.
> 
> Great Question — I actually ran into this issue during high-load testing and spent several weeks tracking it down. The main problem was that the exporter's asynchronous export cycle was too long, causing a sudden spike in generating a large amount of garbage, which in turn triggered prolonged GC pauses that blocked the scheduler thread. I've since addressed this by tuning the `schedule_delay_millis` and `max_export_batch_size` parameters to make exports more frequent but smaller ([bbd10e7](https://github.com/sgl-project/sglang/commit/bbd10e78a06b8caa255d50ddcf3f088a96637d6f)). This helps prevent garbage collection spikes and significantly reduces the risk of blocking the scheduler. While this may increase CPU usage slightly, in most LLM deployment environments CPU resources are typically underutilized, so the trade-off is well worth the improved latency and stability.

Another approach is enable [sampling](https://opentelemetry.io/docs/concepts/sampling/), unfortunately, the Python SDK doesn't support this feature, maybe we can do this in SGLang application layer.

python/sglang/srt/tracing/trace.py
sufeng-buaa: good idea. The default value can be set to a low value.

sufeng-buaa:
> > > > ### Instrumentation Overhead Evaluation
> > > > The test environment is the same as #9962
> > > > Compared to #9962, the overhead of single-span generation and trace context propagation across threads remains unchanged. This section adds support for request tracing in the PD Disaggregation scenario, with an overhead of approximately 60 μs for trace context propagation across nodes (from mini_lb to prefill or decode nodes).
> > > > Additionally, the span structure has been extended to four levels, increasing the overhead of trace_req_start() to approximately 95 μs.
> > > 
> > > 
> > > Hi, did you consider the exporter thread overhead? I'm not sure if it's caused by GIL, but it seems that the exporter thread blocks the scheduler thread when encoding spans in otel sdk, making the overall overhead not fully overlapped in many scenarios.
> > 
> > 
> > Great Question — I actually ran into this issue during high-load testing and spent several weeks tracking it down. The main problem was that the exporter's asynchronous export cycle was too long, causing a sudden spike in generating a large amount of garbage, which in turn triggered prolonged GC pauses that blocked the scheduler thread. I've since addressed this by tuning the `schedule_delay_millis` and `max_export_batch_size` parameters to make exports more frequent but smaller ([bbd10e7](https://github.com/sgl-project/sglang/commit/bbd10e78a06b8caa255d50ddcf3f088a96637d6f)). This helps prevent garbage collection spikes and significantly reduces the risk of blocking the scheduler. While this may increase CPU usage slightly, in most LLM deployment environments CPU resources are typically underutilized, so the trade-off is well worth the improved latency and stability.
> 
> Another approach is enable [sampling](https://opentelemetry.io/docs/concepts/sampling/), unfortunately, the Python SDK doesn't support this feature, maybe we can do this in SGLang application layer.

can be considered in next patch.

ShangmingCai (approved):
LGTM

ShangmingCai:
CC: @slin1237 Do you have time to take another look since there are some modifications in sglang_router as well?

slin1237 (approved):

zhyncs:
not related https://github.com/sgl-project/sglang/actions/runs/18828338947/job/53832111756?pr=10804

cooool

zhyncs (approved):

jimmy-evo:
@slin1237 this break router image. in router image there is no sglang.srt. got import error

https://github.com/sgl-project/sglang/blob/42f8ea4030947bcb5431464370492daee6735e3d/sgl-router/py_src/sglang_router/mini_lb.py#L21

sufeng-buaa:
> @slin1237 this break router image. in router image there is no sglang.srt. got import error
> 
> https://github.com/sgl-project/sglang/blob/42f8ea4030947bcb5431464370492daee6735e3d/sgl-router/py_src/sglang_router/mini_lb.py#L21

Indeed,  it will cause an import error if the image not installed sglang.
Let me figure out a way to fix it as soon.

sufeng-buaa:
> @slin1237 this break router image. in router image there is no sglang.srt. got import error
> 
> https://github.com/sgl-project/sglang/blob/42f8ea4030947bcb5431464370492daee6735e3d/sgl-router/py_src/sglang_router/mini_lb.py#L21

I pushed a fix patch, link is here https://github.com/sgl-project/sglang/pull/12338

PR #10805, opened 2025-09-23
Added memoization storage for logprobs detokenize

echo-rain:
## Motivation

In the current top logprobs logic, all logprobs_val will be detokenized at the end of each decoding round. When the sequence length is long, it will cause excessive CPU load.And when the tokenizer uses a single-core CPU for parallel processing threads, the slower tokenizer processing process will block the processing of other parallel requests.Therefore, it is necessary to add a cache of detokenize logprobs to avoid repeated calculations and thus reduce the CPU load.

## Modifications
1.  A new field is added to ReqState to store the logprobs that have been detokenized.
2. Modify the detokenize logic to change repeated detokenize to only use incremental logprobs for detokenize

## Benchmarking and Profiling
baseline benchmark

After Added memoization storage for logprobs detokenize

## Checklist

### DIFF
--- a/python/sglang/srt/managers/tokenizer_manager.py
+++ b/python/sglang/srt/managers/tokenizer_manager.py
@@ -144,6 +144,13 @@ class ReqState:
     output_token_ids_logprobs_val: List = dataclasses.field(default_factory=list)
     output_token_ids_logprobs_idx: List = dataclasses.field(default_factory=list)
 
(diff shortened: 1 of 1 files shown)


hnyls2002:
Thanks @echo-rain! logprobs detokenize is already memoized on main (delta-only), so I'm closing this as
  superseded. Please reopen if there's still redundant work.

PR #10806, opened 2025-09-23
add lws helm deploy

LuyuZhang00:
## Motivation
Here is the simplified SGLang deployment using LWS, reduced from multiple YAML files to a Helm command

## Modifications
add docs/references/multi_node_deployment/lws_pd_helm/


hnyls2002:
@whybeyoung How do you like it?

PR #10808, opened 2025-09-23
[Feature] Propagate Trace Headers into Root Span for OpenTelemetry Cross-Service Context

zhanghaotong:
## Motivation

Propagate trace headers from upstream requests into the engine trace’s root span to enable OpenTelemetry cross-service context propagation.

## Modifications

Extract traceparent and tracestate from HTTP request headers, construct a trace_headers dict, and pass it as the context when starting the request’s root span.

## Checklist

### DIFF
--- a/python/sglang/srt/managers/io_struct.py
+++ b/python/sglang/srt/managers/io_struct.py
@@ -227,6 +227,9 @@ class GenerateReqInput(BaseReq):
     # Whether to return entropy
     return_entropy: bool = False
 
+    # OpenTelemetry trace headers
+    trace_headers: Optional[Dict[str, str]] = None
+
     def contains_mm_input(self) -> bool:
         return (
             has_valid_data(self.image_data)
--- a/python/sglang/srt/managers/io_struct.py
+++ b/python/sglang/srt/managers/io_struct.py
@@ -227,9 +227,6 @@ class GenerateReqInput(BaseReq):
     # Whether to return entropy
     return_entropy: bool = False
 
-    # tracing context
-    trace_context: Optional[Dict] = None
-
     def contains_mm_input(self) -> bool:
         return (
             has_valid_data(self.image_data)
--- a/python/sglang/srt/managers/io_struct.py
+++ b/python/sglang/srt/managers/io_struct.py
@@ -227,8 +227,8 @@ class GenerateReqInput(BaseReq):
     # Whether to return entropy
     return_entropy: bool = False
 
-    # OpenTelemetry trace headers
-    trace_headers: Optional[Dict[str, str]] = None
+    # tracing context
+    trace_context: Optional[Dict] = None
 
     def contains_mm_input(self) -> bool:
         return (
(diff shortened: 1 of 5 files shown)


zhanghaotong:
This PR will need to be rebased to accommodate the PD separation once [#10804](https://github.com/sgl-project/sglang/pull/10804) has been merged.
cc @sufeng-buaa

zhanghaotong:
I've implemented support for trace part2; Could you please help review this PR? @sufeng-buaa

sufeng-buaa:
> I've implemented support for trace part2; Could you please help review this PR? @sufeng-buaa

ok, let me take a look.

python/sglang/srt/managers/io_struct.py
```diff
@@ -175,6 +175,9 @@ class GenerateReqInput(BaseReq):
     # Whether to return entropy
     return_entropy: bool = False
 
+    # tracing context
+    trace_context: Optional[Dict] = None
```
sufeng-buaa: How about renaming `trace_context` to `external_trace_header`? That would distinguish it from the trace_context in my tracing package, and also make its purpose clearer.

sgl-router/py_src/sglang_router/mini_lb.py
```diff
@@ -376,7 +377,7 @@ async def get_model_info():
 
 
 @app.post("/generate")
-async def handle_generate_request(request_data: dict):
+async def handle_generate_request(request_data: dict, raw_request: Request):
```
sufeng-buaa: The mini_lb is just a testing tool, so it probably won't be integrated into any upstream services. Maybe we don’t need to add this feature. What do you think?

python/sglang/srt/managers/io_struct.py
sufeng-buaa: To keep the protocol struct stable, maybe we can avoid extending `GenerateReqInput` and just use a temporary variable for trace_context?

zhanghaotong: Good idea, I will rename it.

zhanghaotong: Okay, I found that I can pass the context down directly using a temporary variable.

sgl-router/py_src/sglang_router/mini_lb.py
zhanghaotong: In fact, our router is a custom implementation, so we can keep the part as a reference for anyone who wants to build their own router?

zhanghaotong:
@stmatengss Hi ~ Could you please help review this PR?

sgl-router/py_src/sglang_router/mini_lb.py
zhanghaotong: Ok, i have already revert it.

sufeng-buaa:
LGTM. I have verified it in my machine. @ShangmingCai Could you take a look?

ShangmingCai (approved):
LGTM

PR #10846, opened 2025-09-24
Fix eagle radix cache

ispobock:
## Motivation

```
python3 -m sglang.launch_server --model meta-llama/Llama-3.1-8B-Instruct --speculative-algo EAGLE3 \
    --speculative-draft-model-path lmsys/sglang-EAGLE3-LLaMA3.1-Instruct-8B  \
    --speculative-num-steps 2 --speculative-eagle-topk 1 --speculative-num-draft-tokens 3 \
    --dtype float16
    
cd benchmark/mtbench
python3 bench_sglang_eagle.py --parallel 1 --num-questions 10
```
main:
```
w/o radix cache:
#questions: 10, Throughput: 269.86 token/s, Acceptance length: 2.39

w/ radix cache:
#questions: 10, Throughput: 255.74 token/s, Acceptance length: 2.26
```
this PR:
```
w/o radix cache:
#questions: 10, Throughput: 269.45 token/s, Acceptance length: 2.39

w/ radix cache:
#questions: 10, Throughput: 270.85 token/s, Acceptance length: 2.39
```

compatibility test:
- [x] page size 16
- [x] page size 16 + chunked prefill 64
- [x] page size 2 + chunked prefill 64
- [x] page size  1 + chunked prefill 64
- [x] HiCache (this fix is not adapted for HiCache, but doesn't break current HiCache)
- [x] multi prompts in one batch (chunked prefill corner case, ref: #10991)


JustinTong0323 (approved):
Awesome!

ispobock:
The page_size>1 still have some issue, will fix it later.

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -293,29 +319,39 @@ def cache_finished_req(self, req: Req):
...
             page_aligned_kv_indices = kv_indices[:page_aligned_len].to(
                 dtype=torch.int64, copy=True
             )
             self.token_to_kv_pool_allocator.free(kv_indices[page_aligned_len:])
         else:
-            page_aligned_len = len(kv_indices)
+            page_aligned_len = actual_kv_len
             page_aligned_kv_indices = kv_indices.to(dtype=torch.int64, copy=True)
+            if self.is_eagle:
+                self.token_to_kv_pool_allocator.free(kv_indices[page_aligned_len:])
```
xiezhq-hermann, line 338: why do we release it here?

ispobock: In EAGLE case, the kv of the last token will not be inserted into the tree. We need to release it manually.

ispobock:
@xiezhq-hermann @merrymercy This PR is ready for review.

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -278,8 +297,15 @@ def insert(self, key: RadixKey, value=None, chunked=False):
         if self.disable:
             return 0
 
+        key.token_ids = self.key_convert_fn(key.token_ids)
+
         if value is None:
             value = torch.tensor(key.token_ids, dtype=torch.int64)
+
+        if self.is_eagle:
```
xiezhq-hermann, line 305: hiradix (and other trees like swa) override the `insert` function, would that be a problem since eagle worker shared the same tree?

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -278,8 +297,15 @@ def insert(self, key: RadixKey, value=None, chunked=False):
...
+            # Make sure the value len equal to the EAGLE bigram key len
+            value = value[: len(key)]
+
         return self._insert_helper(self.root_node, key, value)
 
     def cache_finished_req(self, req: Req):
```
xiezhq-hermann, line 311: while `hiradix` does not, swa tree override this implementation as well

python/sglang/srt/mem_cache/radix_cache.py
ispobock: Yes, in current design, we need to adapt this change to other trees like swa and hiradix if they override these functions. This PR just makes the main radix tree ready. HiCache and swa need extra work and test to make them ready.

python/sglang/srt/mem_cache/radix_cache.py
ispobock: The swa cache inherits from BaseRadixCache, so it seems all the changes should be implemented again on it. HiCache is from RadixCache, we just need to do some adaptation on it with less override. But for HiCache, the main thing I'm concerning is that the chunked prefill size is a little changed. If the chunked prefill size is 64, actually only 63 bigram keys are inserted to the tree. Maybe it's not efficient for cache offloading with block.

xiezhq-hermann: My understanding is what we are doing is primarily to resolve conflict with eagle workers since it shares the same radix tree but has its own pool, but not to have hicache support for eagle workers, i.e., eagle workers to fetch kv caches from host memory, which seems unnecessary and potentially complicated. Is it correct?

ispobock: I agree that the kv cache for eagle worker is unnecessary to store into host memory since it's only one layer. If we use HiCache only for target model, can we still share the kv indices between target and draft pool?

xiezhq-hermann: yes I think it should be fine just wanted to confirm that we are aligned on this

ispobock: cc: @merrymercy

PR #10847, opened 2025-09-24
fix: split deepseek_v3_fp4 ut

zhyncs:
## Checklist


test/srt/run_suite.py
```diff
@@ -155,6 +155,7 @@ class TestFile:
     "per-commit-4-gpu-b200": [
         # TestFile("test_gpt_oss_4gpu.py", 600),
         TestFile("test_deepseek_v3_fp4_4gpu.py", 3600),
```
ShangmingCai, line 157: Maybe we should decrease the estimated time of this test after split: `TestFile("test_deepseek_v3_fp4_4gpu.py", 1800),`

PR #10850, opened 2025-09-24
Add XPU device support to BaseMultimodalProcessor

vshekhawat-hlab:
## Motivation

Enable support for Intel XPU in BaseMultimodalProcessor.
Previously, only CUDA and NPU devices were supported. This PR extends device handling so that multimodal processing can run seamlessly on Intel XPUs while maintaining existing NPU exceptions (for Qwen processors) and CUDA fallback.

## Modifications

Imported is_xpu and added _is_xpu initialization.
Updated device selection logic to:
- Prefer XPU if available.
- Use NPU (except for Qwen2_5_VLProcessor and Qwen3VLProcessor).
- Default to CUDA otherwise as before.

## Accuracy Tests

No changes to model accuracy expected (only device placement logic modified).

## Checklist


JustinTong0323 (approved):
LGTM.

python/sglang/srt/multimodal/processors/base_processor.py
```diff
@@ -234,14 +235,17 @@ def process_mm_data(
             and isinstance(processor.image_processor, BaseImageProcessorFast)
             and not self.server_args.disable_fast_image_processor
         ):
-            if not _is_npu:
+            if _is_xpu:
```
mickqian: can you write a new method `decide_device_for_processor`?

vshekhawat-hlab: We already have get_device defined in utils.py, so we can reuse it instead of creating a new one. Below is the updated code snippet. Would you like me to make this change?

```

device = get_device()
if device == "npu" and processor.__class__.__name__ in {
    "Qwen2_5_VLProcessor", "Qwen3VLProcessor"
}:
    # Note: For Qwen-VL, the processor has a reshape issue due to dimension restrictions on Ascend.
    pass
else:
    kwargs["device"] = device

```

vshekhawat-hlab: Hi @mickqian,

Should i go with above change?

PR #10851, opened 2025-09-24
Support w4a16 nvfp4 quantization for CompressedTensors

LHXuuu:
## Motivation
While the llm-compressor library has already implemented this quantization method, sglang currently lacks support for w4a16 nvfp4 quantization.

## Modifications

- Added support for w4a16 nvfp4 quantization.

## Checklist


python/sglang/srt/layers/quantization/compressed_tensors/compressed_tensors.py
```diff
@@ -44,6 +44,8 @@
 from sglang.srt.layers.quantization.unquant import UnquantizedLinearMethod
 
 try:
+    from vllm.model_executor.layers.quantization.compressed_tensors.schemes import (
```
AniZpZ: please do not import anymore vllm dependency in quantization module

PR #10854, opened 2025-09-24
[feat]: enhance fused MOE kernel with TMA support

liusy58:
## Motivation

This commit introduces TMA (Tensor Memory Accelerator) support for the fused MoE (Mixture of Experts) kernel to improve memory access performance on modern GPU architectures.

## Benchmarking and Profiling
This optimization delivers measurable performance improvements in production scenarios:
* 5% TTFT reduction for DeepSeek-R1 deployment
* Configuration: 4000 input tokens, 1500 output tokens, TP=8

 

## Checklist

### DIFF
--- a/python/sglang/srt/layers/moe/fused_moe_triton/fused_moe_triton_kernels.py
+++ b/python/sglang/srt/layers/moe/fused_moe_triton/fused_moe_triton_kernels.py
@@ -40,6 +40,8 @@ elif _is_hip:
 
 padding_size = 128 if bool(int(os.getenv("SGLANG_MOE_PADDING", "0"))) else 0
 
+USE_TMA = get_bool_env_var("USE_TMA")
+
 
 @triton.jit
 def write_zeros_to_output(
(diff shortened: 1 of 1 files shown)


xu-yfei:
The solution is similar to that in PR #10567. In addition to modifying B with TMA, we also applied TMA modifications to input A of the second MOE. We found that the performance improvement after TMA modification of input A is more significant. Moreover, the optimal configuration of TMA will also change.

Alcanderian:
Let's compare the performance with #10567 @liusy58

Jacki1223:
> The solution is similar to that in PR #10567. In addition to modifying B with TMA, we also applied TMA modifications to input A of the second MOE. We found that the performance improvement after TMA modification of input A is more significant. Moreover, the optimal configuration of TMA will also change.

Thank you for sharing this insight. I have a follow-up question: since you observed a bigger gain from applying TMA to input A alone, why not simply apply TMA to both A and B? In our own experiments we consistently see an additional, non-negligible boost when TMA is enabled on both MOE inputs, and the overhead is minimal.

xu-yfei:
> > The solution is similar to that in PR #10567. In addition to modifying B with TMA, we also applied TMA modifications to input A of the second MOE. We found that the performance improvement after TMA modification of input A is more significant. Moreover, the optimal configuration of TMA will also change.
> 
> Thank you for sharing this insight. I have a follow-up question: since you observed a bigger gain from applying TMA to input A alone, why not simply apply TMA to both A and B? In our own experiments we consistently see an additional, non-negligible boost when TMA is enabled on both MOE inputs, and the overhead is minimal.

Yes, we have applied TMA to both inputs A and B of the second MoE (down proj), while no TMA was used for the first one (gateup proj). The performance improvement we observed for the first MoE is relatively small (approximately 1%).

dongyibo:
May I ask if there are any merger plans? @liusy58

PR #10248, opened 2025-09-10
enable rmsnorm on XPU

huaiyuzh:
enable rmsnorm on XPU.

PR #10858, opened 2025-09-24
[Intel GPU] fix import error to run DeepSeek-V2-Lite model with BF16 on XPU

polisettyvarma:
## Motivation

Enable deepseek model on XPU

## Modifications

import cuda specific kernels when cuda is available

## Accuracy Tests

`python -m sglang.bench_one_batch  --batch-size 8  --input 1024  --output 1024  --model "deepseek-ai/DeepSeek-V2-Lite"  --dtype bfloat16 --tp 1  --trust-remote-code  --mem-fraction-static 0.8  --device xpu --correctness --json-model-override-args '{"num_hidden_layers": 8}'`
above command gives garbage output due to reduced model, if enough memory is available - full model can be tried

## Benchmarking and Profiling

Not Applicable

## Checklist


python/sglang/srt/layers/moe/cutlass_w4a8_moe.py
```diff
@@ -3,12 +3,18 @@
...
+
+_is_cuda = is_cuda()
+
+if _is_cuda:
+    from sgl_kernel import (
+        cutlass_w4a8_moe_mm,
+        get_cutlass_w4a8_moe_mm_data,
+    )
+
+from sgl_kernel import sgl_per_tensor_quant_fp8, silu_and_mul
```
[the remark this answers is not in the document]
polisettyvarma: not right way to fix this

polisettyvarma:
@Edwardf0t1 @kushanam @BBuf @ch-wan @HaiShaw @ispobock @zhyncs @Ying1123 @merrymercy @Alcanderian @mingfeima please review

mingfeima:
I don't think that we have implemented w4a8 moe on intel gpus yet. Even we have done this job. We most likely don't need this patch, we should align kernel APIs with existing cuda ones and directly import from `sgl-kernel`.

I suggest we do the enabling job in this order:
* name our target model, say Llama3-8B
* finish kernel level optimizations from sgl-kernel
* make modifications in sglang python files

we should avoid:
* modify every bit of corners that has `cuda`

python/sglang/srt/layers/moe/cutlass_w4a8_moe.py
```diff
@@ -3,12 +3,18 @@
...
-from sgl_kernel import (
-    cutlass_w4a8_moe_mm,
-    get_cutlass_w4a8_moe_mm_data,
-    sgl_per_tensor_quant_fp8,
-    silu_and_mul,
-)
+
+from sglang.srt.utils import is_cuda
+
+_is_cuda = is_cuda()
```
deepvars: should we use is_cuda() here or is_cuda_alike ? Or may be just check not XPU here ?

polisettyvarma: changed to is_cuda_alike

deepvars (approved):

mingfeima:
duplicated with https://github.com/sgl-project/sglang/pull/10248/files

changes in `cutlass_w4a8_moe.py` is not needed at the moment:
* right now intel xpu does not really support `--quantization w4afp8` recipe, report an error when parsing the input arguments makes more sense.
* the w4afp8 cutlass kernels are WIP right now, and once the kernels are ready, we are much likely to dispatch in `/python/sglang/srt/layers/quantization/w4afp8.py` which will just by pass `cutlass_w4a8_moe.py`

so my suggestion is that we directly make changes in sglang after the w4afp8 kernels are ready in `sgl-kernel-xpu`.

polisettyvarma:
i have removed duplicate code
@mingfeima quantization is None by default and got this issue when running with bf16

mingfeima (approved):

mingfeima:
@polisettyvarma rebase please.

polisettyvarma:
> @polisettyvarma rebase please.

i have done it, thanks

polisettyvarma:
@iforgetmyname please review

polisettyvarma:
@Kangyan-Zhou please review this PR

PR #10871, opened 2025-09-24
[Bugfix][DO NOT MERGE] Free req.req_pool_idx when it is finished in the mixed_chunk mode to solve memory leak

Muqi1029:
## Motivation

When using mixed chunking, a request may complete in the first prefill stage. In the original code, only `out_cache_loc` is freed, while `req_pool_idx` remains allocated. As a result, the scheduler’s `check_memory` function detects this as a memory leak and raises an error, which causes the entire SGLang Engine to fail and significantly impacts its stability.

## Modifications

Free `req_pool_idx` in `process_batch_result_prefill`.

## Checklist


MichelleWu351:
not work, still have memory leak

Muqi1029:
@MichelleWu351  Maybe the situation of memory leak is not the same as yours. I test it in the latest sglang, it still happens.

Muqi1029:
I can easily reproduce this bug about memory leak:

Using the latest sglang:
### Server
```shell
python -m sglang.launch_server --model-path Qwen/Qwen3-8B --port 8888 --enable-mixed-chunk
```

### Client
```shell
import requests
import json

url = "http://127.0.0.1:8888/generate"

headers = {
    "Content-Type": "application/json",
    "Authorization": "Just Keep Me"
}

data = {
    "input_ids": [1] * 40957,
}

response = requests.post(url, headers=headers, json=data)

if response.status_code == 200:
    result = response.json()
    print(json.dumps(result, ensure_ascii=False, indent=2))
else:
    print(f"Error: {response.status_code}")
    print(response.text)

```

### Result of server:
A bad request will make sglang engine fail to check memory leak, leading to the following error, making the whole sglang engine down:
```
[2025-10-25 07:57:04] The server is fired up and ready to roll!
[2025-10-25 07:57:12 TP0] Input length (40957 tokens) exceeds the maximum allowed length (40954 tokens). Use a shorter input or enable --allow-auto-truncate., self.rid='4781a1d1797f4ef78da4d4e4f3949168'
[2025-10-25 07:57:12 TP0] Prefill batch [10], #new-seq: 1, #new-token: 1, #cached-token: 0, token usage: 0.00, #running-req: 0, #queue-req: 0,
[2025-10-25 07:57:12] [http_server] Error: Input length (40957 tokens) exceeds the maximum allowed length (40954 tokens). Use a shorter input or enable --allow-auto-truncate.
[2025-10-25 07:57:12 TP0] Scheduler hit an exception: Traceback (most recent call last):
  File "/root/projects/dev_project/sglang/python/sglang/srt/managers/scheduler.py", line 2769, in run_scheduler_process
  File "/root/.envs/dev/lib/python3.12/site-packages/torch/utils/_contextlib.py", line 120, in decorate_context
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/root/projects/dev_project/sglang/python/sglang/srt/managers/scheduler.py", line 1018, in event_loop_overlap
    self.self_check_during_idle()
  File "/root/projects/dev_project/sglang/python/sglang/srt/managers/scheduler_runtime_checker_mixin.py", line 214, in self_check_during_idle
    self.check_memory()
  File "/root/projects/dev_project/sglang/python/sglang/srt/managers/scheduler_runtime_checker_mixin.py", line 149, in check_memory
    self._check_req_pool()
  File "/root/projects/dev_project/sglang/python/sglang/srt/managers/scheduler_runtime_checker_mixin.py", line 135, in _check_req_pool
    raise ValueError(msg)
ValueError: req_to_token_pool memory leak detected!available_size=4095, total_size=4096
```

### Reason
A request with an `input_ids` length below the model configuration limit can bypass the length check in the HTTP server:

https://github.com/sgl-project/sglang/blob/a4a3d8239365d2fc16e2dc707605373ea3f35ded/python/sglang/srt/managers/tokenizer_manager.py#L607-L616

However, it fails later in the engine:

https://github.com/sgl-project/sglang/blob/a4a3d8239365d2fc16e2dc707605373ea3f35ded/python/sglang/srt/managers/scheduler.py#L1369-L1381

This causes the request to be marked as aborted. The aborted request is then added to the `waiting_queue` as if it were a normal one by setting its `input_ids` to `[0]`. As a result, its `out_cache_loc` and request slot are allocated but never released:

https://github.com/sgl-project/sglang/blob/a4a3d8239365d2fc16e2dc707605373ea3f35ded/python/sglang/srt/managers/scheduler_output_processor_mixin.py#L85-L89

Eventually, this leads to a `check_memory` failure, which brings down the entire engine and reduces SGLang’s robustness to malformed or edge-case requests.

Muqi1029:
DO NOT MERGE since it may be incompatible with retracted req, I am fixing it now

PR #10874, opened 2025-09-24
EPLB: prefer to use physical experts in the same gpu or node

acelyc111:
## Motivation

Similar to https://github.com/sgl-project/sglang/pull/9849, when it's possible to use physical experts on the same node, we'd prefer to route to these GPU ranks in dynamic mode.

## Modifications

When setting `--ep-dispatch-algorithm dynamic`, select the experts on the same GPU at first if possible, then select experts on the same node if possible, at last, select the experts randomly as a fall back method.

## Checklist

### DIFF
--- a/python/sglang/srt/eplb/expert_location.py
+++ b/python/sglang/srt/eplb/expert_location.py
@@ -113,6 +113,7 @@ class ExpertLocationMetadata:
         server_args: ServerArgs,
         model_config: ModelConfig,
         physical_to_logical_map,
+        gpu_id: int = None,
     ):
         if not isinstance(physical_to_logical_map, torch.Tensor):
             physical_to_logical_map = torch.tensor(physical_to_logical_map)
@@ -125,8 +126,11 @@ class ExpertLocationMetadata:
 
         model_config_for_expert_location = common["model_config_for_expert_location"]
         logical_to_all_physical_map = _compute_logical_to_all_physical_map(
-            physical_to_logical_map,
+            server_args=server_args,
+            physical_to_logical_map=physical_to_logical_map,
             num_logical_experts=model_config_for_expert_location.num_logical_experts,
+            num_gpus=common["ep_size"],
+            gpu_id=gpu_id,
         )
 
         return ExpertLocationMetadata._init_raw(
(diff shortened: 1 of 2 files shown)


python/sglang/srt/eplb/expert_location.py
```diff
@@ -321,6 +329,29 @@ def _compute_logical_to_all_physical_map(
...
+                    num_local_gpu_physical_experts=num_local_gpu_physical_experts,
+                    gpu_id=gpu_id,
+                    num_gpus_per_node=num_gpus_per_node,
+                    num_local_node_physical_experts=num_local_node_physical_experts,
+                )
+                if nearest_expert != -1:
+                    # Replace with the nearest physical expert
+                    logical_to_all_physical_map[layer_id][logical_expert_id] = [
+                        nearest_expert
+                    ]
```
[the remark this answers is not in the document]
acelyc111: It's fair for all ranks.

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -353,7 +353,11 @@ def initialize(self, min_per_gpu_memory: float):
 
         if not self.is_draft_worker:
             set_global_expert_location_metadata(
-                compute_initial_expert_location_metadata(server_args, self.model_config)
+                compute_initial_expert_location_metadata(
+                    server_args=server_args,
+                    model_config=self.model_config,
+                    gpu_id=self.moe_ep_rank,
```
fzyzcjy: maybe call it moe_ep_rank. the "gpu_id" means the gpu index within one single node.

python/sglang/srt/eplb/expert_location.py
```diff
@@ -321,6 +330,29 @@ def _compute_logical_to_all_physical_map(
                 physical_expert_id
             )
 
+    if gpu_id is not None:
```
fzyzcjy: maybe add a brief comments on here or somewhere else explaning "prefer same gpu"

fzyzcjy (approved):
LGTM only nits

python/sglang/srt/model_executor/model_runner.py
acelyc111: done

PR #10875, opened 2025-09-24
feat(reasoning): improve enable thinking from request

jimmy-evo:
## Motivation

read from request's "enable_thinking" or "thinking", 
should also filtered with `reasoning_parser`

----

there is a real case:
to compat dpsk and qwen, both `chat_template_kwargs.enable_thinking` and `chat_template_kwargs.thinking`  will  pass to server, default is False.
if enable thinking for dpsk, `chat_template_kwarge` in request will be like `enable_thinking=True, thinking =False`, then run into `enable_thinking == False` instead of `thinking == True`

## Modifications

add reasoning parser check with `enable_thinking`

## Checklist


python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -933,11 +934,11 @@ def _get_enable_thinking_from_request(self, request: ChatCompletionRequest) -> b
         """
         if hasattr(request, "chat_template_kwargs") and request.chat_template_kwargs:
             # For Qwen3 models, `enable_thinking` is supported.
-            if request.chat_template_kwargs.get("enable_thinking") is not None:
-                return request.chat_template_kwargs.get("enable_thinking")
+            if self.reasoning_parser in ["qwen3-thinking", "qwen3", "glm45"]:
```
JustinTong0323: qwen3-thinking is force thinking one, remove.

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -58,6 +58,7 @@ def __init__(
         super().__init__(tokenizer_manager)
         self.template_manager = template_manager
         self.tool_call_parser = self.tokenizer_manager.server_args.tool_call_parser
+        self.reasoning_parser = self.tokenizer_manager.server_args.reasoning_parser
```
JustinTong0323, line 67: Could you also help replace all occurrences of `self.tokenizer_manager.server_args.reasoning_parser` to `self.reasoning_parser`? Thanks~

JustinTong0323 (commented):
For more robust consideration, this scenario should be handled before passing to sglang... But it's also OK to enhance here, thanks for your contribution!

CatherineSue (approved):

CatherineSue:
is the CI failure related to this change?

PR #10883, opened 2025-09-25
[minor] Remove deprecated function `get_ip`

merrymercy:


EduardDurech:
@merrymercy can you not make deprecated methods instead of removing? This is used by veRL https://github.com/volcengine/verl/blob/main/verl/workers/rollout/sglang_rollout/sglang_rollout.py#L41C5-L41C11

PR #10192, opened 2025-09-09
[PD-HiCache]: Support Async Offloading KVCache In Decode Side

hzh0425:
## Motivation
In the PD architecture, the HiRadixTree is not enabled by default on the Decode side. Therefore, a new module named 'decode_kvcache_offload_manager' has been added. This module is responsible for managing the offload process of KVCache from Device to Host to Remote on the Decode side.

image

Co Author: @ShangmingCai 
## Modifications
```
request finished
   │
   ▼
offload_kv_cache(req)
   │
   ├─ get token_indices
   └─ HiCacheController.write()      
) ─────────────────────────────────┐
                                   │ GPU → Host
                                   ▼
                         ongoing_offload[ack_id]
                                   │
                                   ▼
                         check_offload_progress()
                                   │
                          ┌────────┴────────┐
                          ▼                 ▼
              _check_offload_progress   _check_backup_progress
                          │                 │
                          ▼                 ▼
                   ack_write_queue     ack_backup_queue
                          │                 │
                          ▼                 ▼
              ┌─ tree_cache.cache_     host_mem_pool.
              │  finished_req()       free()
              │                       
              ▼                       
         _trigger_backup() 
              │
              ▼
    write_storage() ──────┐
                          │ Host → Storage
                          ▼
              Remote Storage
```

## Checklist

- [ ] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.ai/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.ai/developer_guide/contribution_guide.html#benchmark-the-speed).

PR #10884, opened 2025-09-25
[Feature] PD-HiCache support backup skip function

leihuang-sketch:
## Motivation
Regarding the function of PD separating [decode and writing back kv cache to storage](https://github.com/sgl-project/sglang/pull/10192), prefill performs a backup and decode performs an incremental write back, which can cause the following situations
1. Pre fill and decode incremental fragmentation on physical storage
2. Writing prefill and decode once each is somewhat redundant

## Modifications
Prefill: Add the `--hicache-backup` skip option for prefill skipping backups and entering read-only mode
Decode: Decode uses `--disaggregation-decode-enable-offload-kvcache`, which can fully and continuously backup the entire request's kv cache

## Checklist


hzh0425:
> * Pre fill and decode incremental fragmentation on physical storage
> * Writing prefill and decode once each is somewhat redundant

Actually, although PD will write once each time, D will only write the incremental part, so the impact might be minimal. Moreover, the actual bandwidth usage remains the same.

Secondly, I don't quite understand what you mean by "fragmentation"?

thanks

leihuang-sketch:
> > * Pre fill and decode incremental fragmentation on physical storage
> > * Writing prefill and decode once each is somewhat redundant
> 
> Actually, although PD will write once each time, D will only write the incremental part, so the impact might be minimal. Moreover, the actual bandwidth usage remains the same.
> 
> Secondly, I don't quite understand what you mean by "fragmentation"?
> 
> thanks

@hzh0425 Fragmentation refers to the situation where the prefix is written to the decode increment, and there are still many requests continuously written in between. At this point, the offset between the prefix and the decode increment will be very large, especially in high concurrency scenarios
I think it would be more appropriate for the entire request to be written by the decode side for membership

PR #9687, opened 2025-08-27
[Do not merge][Bug Fix] Resolve CUDA illegal memory access issue in speculative decoding

ryang-max:
## Motivation

https://github.com/sgl-project/sglang/issues/6309
https://github.com/sgl-project/sglang/issues/9481
In verify process of speculative decoding with cuda_graph enabled, sometimes it throws "CUDA error: an illegal memory access was encountered" issue.

**Reproduction steps from #9481  **
```
python3 -m sglang.launch_server --model-path XiaomiMiMo/MiMo-7B-RL --enable-memory-saver \
--trust-remote-code --skip-server-warmup --speculative-algorithm EAGLE --speculative-num-steps 20  \
--speculative-eagle-topk 10 --speculative-num-draft-tokens 100  --mem-fraction 0.7

#benchmark/gsm8k
python3 bench_sglang.py --num-questions 2000 --parallel 17
```

## Modifications

Add padding for tree_mask to fit for cuda_graph_bs

## Checklist

- [x] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.ai/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.ai/developer_guide/contribution_guide.html#benchmark-the-speed).

Issue #8296, opened 2025-07-23
[Bug] illegal memory access


lethean1:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Describe the bug

Without the use of CUDA Graphs, the implementation of speculative decoding has a 20% probability of encountering illegal memory access.

the log is shown below

````
Found 1 data files. Loading...
Total rows loaded: 166667
Filtering for English language and turn 1...
Rows after filtering: 90040
Parsing conversations...
Loaded 89863 prompts and responses.
Loading safetensors checkpoint shards:   0% Completed | 0/4 [00:00<?, ?it/s]
Loading safetensors checkpoint shards:  25% Completed | 1/4 [00:05<00:16,  5.62s/it]
Loading safetensors checkpoint shards:  50% Completed | 2/4 [00:11<00:11,  5.81s/it]
Loading safetensors checkpoint shards:  75% Completed | 3/4 [00:17<00:05,  5.79s/it]
Loading safetensors checkpoint shards: 100% Completed | 4/4 [00:18<00:00,  3.97s/it]
Loading safetensors checkpoint shards: 100% Completed | 4/4 [00:18<00:00,  4.63s/it]

Loading pt checkpoint shards:   0% Completed | 0/1 [00:00<?, ?it/s]
Loading pt checkpoint shards: 100% Completed | 1/1 [00:02<00:00,  2.34s/it]
Loading pt checkpoint shards: 100% Completed | 1/1 [00:02<00:00,  2.34s/it]

[rank1]:[E723 14:57:44.922882154 ProcessGroupNCCL.cpp:1895] [PG ID 2 PG GUID 3 Rank 1] Process group watchdog thread terminated with exception: CUDA error: an illegal memory access was encountered
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1
[...]

Issue #6309, opened 2025-05-15
[Bug] eagle2【CUDA error: an illegal memory access was encountered】


dongyibo:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Describe the bug

I was running eagle2 online（based on sglang=0.4.5，sgl-kernel=0.0.5 ，flashinfer-python=0.2.5）, and after a while, I encountered an cuda error saying "an illegal memory access was encountered". Here are the details of the error. Could you help me take a look?

---------------------
File "/root/python/betelgeuse/srt/managers/scheduler.py", line 2582, in run_scheduler_process
    scheduler.event_loop_normal()
  File "/usr/local/lib/python3.10/dist-packages/torch/utils/_contextlib.py", line 116, in decorate_context
    return func(*args, **kwargs)
  File "/root/python/betelgeuse/srt/managers/scheduler.py", line 702, in event_loop_normal
    result = self.run_batch(batch)
  File "/root/python/betelgeuse/srt/managers/scheduler.py", line 1657, in run_batch
    ) = self.draft_worker.forward_batch_speculative_generation(batch)
  File "/root/python/betelgeuse/srt/speculative/eagle_worker.py", line 254, in forward_batch_speculative_generation
    logits_output, verify_output, model_worker_batch = self.verify(
  File "/root/python/betelgeuse/srt/speculative/eagle_worker.py", line 490, in verify
    logits_output,_,_,_ = self.target_worker.forward_batch_generation(
  File "/root/python/betelgeuse/srt/managers/tp_worker.py", line 292, in forward_batch_generation
    logits_output = self.model_runner.forward(forward_batch)
  File "/usr/local/lib/python3.10/dist-packages/torch/utils/_contextlib.py", line 116, in decorate_context
    return func(*args, **kwargs)
  File "/root/python/betelgeuse/srt/model_executor/model_runner.py", line 1257, in forward
    return self.cuda_graph_runner.replay(
[...]

PR #10892, opened 2025-09-25
Fix CUDA illegal memory access issues in speculative decoding

ur4t:
## Motivation

Fix CUDA illegal memory issues in EAGLE CUDA graph runners like:
#6309 #8296 #8336 #9888

This PR aims to fix the root cause of IMA issues, unlike #9687 which attempts to mitigate with padding.

## Modifications

Reset `self.positions` in EAGLE CUDA graph runners when `raw_bs` and `capture_bs` differ.

## Checklist


ur4t:
@ryang-max I'm not sure which approach is better or even are these two PR are working on problems caused by the same root cause, please take a look.

ryang-max:
> @ryang-max I'm not sure which approach is better or even are these two PR are working on problems caused by the same root cause, please take a look.

I think this PR may give a better solution for this issue! But we need to consider whether this will introduce extra cost in parallelism, and have test on its correctness (will not decrease accept-len). Any test result on it?
cc @guapisolo @ispobock

ur4t:
> But we need to consider whether this will introduce extra cost in parallelism, and have test on its correctness (will not decrease accept-len). Any test result on it?

Tested with `bench_speculative.py` with recommended arguments; before/after this PR with model/config/cli/machine determined:
- `Accept length`s are exactly the same.
- Performances are almost the same; which one is better depends on the condition of the machine.

ryang-max:
> > But we need to consider whether this will introduce extra cost in parallelism, and have test on its correctness (will not decrease accept-len). Any test result on it?
> 
> Tested with `bench_speculative.py` with recommended arguments; before/after this PR with model/config/cli/machine determined:
> 
> * `Accept length`s are exactly the same.
> * Performances are almost the same; which one is better depends on the condition of the machine.

https://github.com/sgl-project/sglang/pull/9687#issuecomment-3332511182

Ensure that this PR fix the issue, maybe try with arguments mentioned in #9687 description more than 5 times with no ima

ur4t:
> Ensure that this PR fix the issue, maybe try with arguments mentioned in #9687 description more than 5 times with no ima

No IMA during 10 runs with this fix. Without it, 1 run is enough to trigger IMA.

> I'm not sure which approach is better or even are these two PR are working on problems caused by the same root cause, please take a look.

In addition to `qo_indptr_host = qo_indptr.to("cpu")`, IMA may occur at score_list concat after replay, which takes hours to reproduce. Therefore, I was not quite sure whether they are caused by the same root cause and whether padding could make sense. Gladly, they are, and both fixed by this PR.

Ch3ngY1:
> > > But we need to consider whether this will introduce extra cost in parallelism, and have test on its correctness (will not decrease accept-len). Any test result on it?
> > 
> > 
> > Tested with `bench_speculative.py` with recommended arguments; before/after this PR with model/config/cli/machine determined:
> > 
> > * `Accept length`s are exactly the same.
> > * Performances are almost the same; which one is better depends on the condition of the machine.
> 
> [#9687 (comment)](https://github.com/sgl-project/sglang/pull/9687#issuecomment-3332511182)
> 
> Ensure that this PR fix the issue, maybe try with arguments mentioned in #9687 description more than 5 times with no ima

Hi all, 
we tested this PR and reached the following conclusions:

1. **Without this PR**, IMA occurs within 10 minutes, when using the following settings with cudagraph:
```
    --speculative-num-steps 10 \
    --speculative-eagle-topk 10 \
    --speculative-num-draft-tokens 100 \
```
2. **With this PR applied**, we have benchmarked our service for more than 24 hours without encountering IMA.

hnyls2002 (approved):
LGTM and the changes are solid.

hnyls2002 (approved):

jianfei-wangg:
> > @ryang-max I'm not sure which approach is better or even are these two PR are working on problems caused by the same root cause, please take a look.
> 
> I think this PR may give a better solution for this issue! But we need to consider whether this will introduce extra cost in parallelism, and have test on its correctness (will not decrease accept-len). Any test result on it? cc @guapisolo @ispobock

I wonder why this PR is better compared to #9687 ? #9687  fix target_model cuda graph and this PR fix draft_model cuda graph,  two fixes are both needed?

ur4t:
> I wonder why this PR is better compared to #9687 ? #9687 fix target_model cuda graph and this PR fix draft_model cuda graph, two fixes are both needed?

@jianfei-wangg Because actually #9687 is not to fix faults in target_model cuda graph, but to mitigate faults in draft_model cuda graph via padding `forward_batch.spec_info` which is only accessible in `cuda_graph_runner.py`.

PR #10909, opened 2025-09-25
model: Support Hybrid Mamba2 NemotronHForCausalLM (nvidia/NVIDIA-Nemotron-Nano-9B-v2)

netanel-haber:
Support the `NemotronHForCausalLM` architecture, which can include any combination of *Mamba2*, MLP and normal self-attention layers.

## Motivation
Support the `NemotronHForCausalLM` architecture, which includes https://huggingface.co/nvidia/NVIDIA-Nemotron-Nano-9B-v2:
> The model uses a hybrid architecture consisting primarily of Mamba-2 and MLP layers combined with just four Attention layers. For the architecture, please refer to the [Nemotron-H tech report](https://arxiv.org/abs/2504.03624).

TODOs:
- [x] Measure runtime of mamba kernel unittests and update run_test.
- [x] Run a GSM8K to check the Qwen3-Next + Falcon haven't been impacted by the changes.
- [x] Perhaps get official copy for Nemotron-Nano-V2 support matrix entry.

---------------------------
## Checklist


test/srt/run_suite.py
```diff
@@ -123,6 +125,9 @@ class TestFile:
         TestFile("test_vlm_input_format.py", 300),
         TestFile("test_vision_openai_server_a.py", 724),
         TestFile("test_vision_openai_server_b.py", 446),
+        TestFile("layers/attention/mamba/test_causal_conv1d.py", TODO),  # TODO: Measure
```
yizhang2077: add time here?

ispobock: Could you update the estimated time?

python/sglang/srt/layers/attention/mamba/mamba2_mixer.py
```diff
@@ -0,0 +1,458 @@
...
+from .mamba2_metadata import Mamba2Metadata
+from .mamba2_mixer_rms_norm_gated import Mixer2RMSNormGated
+from .mamba_ssm import selective_state_update
+from .ssd_combined import mamba_chunk_scan_combined
+
+# Added by the IBM Team, 2024
+
+
+# Adapted from transformers.models.mamba.modeling_mamba.MambaMixer
+class Mamba2Mixer(nn.Module):
```
ispobock: Some mamba components are already added in https://github.com/sgl-project/sglang/pull/10988. Could you rebase main branch and remove the common parts?

ispobock:
@netanel-haber Could you also resolve the conflicts? Thanks!

python/sglang/srt/mem_cache/memory_pool.py
```diff
@@ -107,17 +110,39 @@ def clear(self):
...
+        cache_params: Mamba2CacheParams,
         device: str,
         speculative_num_draft_tokens: Optional[int] = None,
     ):
+        conv_state_shape = cache_params.shape.conv
+        temporal_state_shape = cache_params.shape.temporal
+        conv_dtype = cache_params.dtype.conv
+        ssm_dtype = cache_params.dtype.temporal
+        num_mamba_layers = len(cache_params.layers)
+        # TODO: why +1?
```
mickqian: is this still necessary?

python/sglang/srt/layers/attention/mamba/mamba2_mixer.py
netanel-haber: Done. I merged the two Mamba2Mixers, and reused kernels added to .ops by falcon merge. PR has therefore been reduced from 6k LOC to ~3K LOC (1K of which is unittests lifted almost wholesale from VLLM).
I'll actually run the merge result on Sunday and change minor things if anything broke that I hadn't noticed, and go over my TODOs in my pr description, including the unittest times.
I added a TODO to the PR description to check Falcon hasn't been damaged either.

test/srt/run_suite.py
netanel-haber: See: https://github.com/sgl-project/sglang/pull/10909#discussion_r2401580981

python/sglang/srt/layers/attention/mamba/ops/ssd_state_passing.py
```diff
@@ -13,17 +13,18 @@
...
-#         triton.Config({"BLOCK_SIZE": 64}),
-#         triton.Config({"BLOCK_SIZE": 128}),
-#         triton.Config({"BLOCK_SIZE": 256}),
-#         triton.Config({"BLOCK_SIZE": 512}),
-#         triton.Config({"BLOCK_SIZE": 1024}),
-#         triton.Config({"BLOCK_SIZE": 2048}),
-#     ],
-#     key=["dim"],
-# )
+@triton.autotune(
```
yizhang2077: remove autotune here, the same as other kernel

test/srt/run_suite.py
```diff
@@ -11,6 +11,8 @@ class TestFile:
     estimated_time: float = 60
 
 
+TODO = 500
```
yizhang2077: remove TODO here

python/sglang/srt/layers/attention/mamba/causal_conv1d_triton.py
```diff
@@ -672,7 +672,9 @@ def _causal_conv1d_update_kernel(
         + (conv_state_batch_coord * stride_conv_state_seq)
         + conv_state_token_offset * stride_conv_state_tok
         + (idx_feats * stride_conv_state_dim)[None, :]
-        + ((idx_tokens + 1) * stride_conv_state_tok)[:, None]
+        + ((idx_tokens + (1 if IS_SPEC_DECODING else seqlen)) * stride_conv_state_tok)[
+            :, None
```
yizhang2077, line 676: why do we need modify triton kernel here? cc @hebiao064

netanel-haber: https://github.com/vllm-project/vllm/pull/24680
As per fix introduced to VLLM - for non spec decoding models (like nanov2)

python/sglang/srt/layers/attention/mamba/ops/ssd_state_passing.py
netanel-haber: Why remove the autotune?

yizhang2077: autotune will cause hang in profile stage. I think the best approach is to add a config, and finetune offline. Then we add it this config into a folder. We do `fusedMoE` in this way. But if we do in this way there are so many config/environments  to tune and it is hard to maintain. So I think we can choose an any proper config if the triton ops is not heavy as a trade-off.

python/sglang/srt/layers/attention/mamba/causal_conv1d_triton.py
yizhang2077: I think we need ask @hebiao064

ispobock:
@netanel-haber could you resolve the conflicts?

python/sglang/srt/layers/attention/mamba/ops/ssd_state_passing.py
netanel-haber: Removed autotune code.

python/sglang/srt/layers/attention/mamba/causal_conv1d_triton.py
hebiao064: @netanel-haber tho we adapted the kernel from vllm (which is also adapted from FLA repo), SGLang is tacking the Spec Decoding differently than vllm.

In SGLang, IS_SPEC_DECODING will always be False even under Target Verify mode, we allocated a conv states pool with 4 (as the qwen3-next conv state setup) + spec_num_steps (typically 3 or 4), and after verification is done, we sliced the 4 + x (x = num_accepted tokens) and put it back to the pool.

So we never used the spec decoding related logic in the triton kernel, I believe your PR and QWen3-Next will be good with or without this change.


I tentatively agree with keep your change to keep synced with vLLM, but please try test spec decoding with your model and also QWen3-Next

netanel-haber: I checked, and `IS_SPEC_DECODING` is always `False` for Qwen3-Next, even if speculative decoding is running, as you said.
Anyway, testing with or without the change yields equivalent results for Qwen3-Next with both spec decoding on and off [and for NanoV2, without spec decoding].

```python
test/srt/models/test_qwen3_next_models.py::TestQwen3Next::test_gsm8k
metrics={'accuracy': np.float64(0.965)}

test/srt/models/test_qwen3_next_models.py::TestQwen3NextMTP::test_gsm8k
metrics={'accuracy': np.float64(0.96)}

test/srt/models/test_nvidia_nemotron_nano_v2.py
metrics={'accuracy': np.float64(0.885)}

…/sglang git checkout main -- python/sglang/srt/layers/attention/mamba/causal_conv1d_triton.py

test/srt/models/test_qwen3_next_models.py::TestQwen3Next::test_gsm8k
metrics={'accuracy': np.float64(0.94)}

test/srt/models/test_qwen3_next_models.py::TestQwen3NextMTP::test_gsm8k
metrics={'accuracy': np.float64(0.965)}

test/srt/models/test_nvidia_nemotron_nano_v2.py
metrics={'accuracy': np.float64(0.875)}
```


So, I think that since you agree that parity with VLLM has merit, and that we observe no accuracy change, I'm leaning towards keeping the change.
I'll wait for further comment from you here before I resolve

ispobock (approved):
Overall LGTM. @netanel-haber just need to resolve the conflict and get @yizhang2077 @hebiao064 's approval. Then it can be merged.

python/sglang/srt/layers/attention/mamba/causal_conv1d_triton.py
netanel-haber: Resolving at this point, if @hebiao064 wants to, he can unresolve

yizhang2077 (approved):
LGTM overall, once all ut about linear models have passed I think we could merge it

test/srt/run_suite.py
```diff
@@ -127,6 +127,10 @@ class TestFile:
         TestFile("test_vlm_input_format.py", 300),
         TestFile("test_vision_openai_server_a.py", 724),
         TestFile("test_vision_openai_server_b.py", 446),
+        TestFile("layers/attention/mamba/test_causal_conv1d.py", 85),
+        TestFile("layers/attention/mamba/test_mamba_ssm.py", 85),
+        TestFile("layers/attention/mamba/test_mamba_ssm_ssd.py", 220),
+        TestFile("models/test_nvidia_nemotron_nano_v2.py", 180),
```
merrymercy, line 133: 1. These tests won't run by CI. You need to add this line https://github.com/sgl-project/sglang/blob/55b14656e656c0a0be3d9a24e69f624c957429e7/python/sglang/srt/mem_cache/storage/hf3fs/test_hf3fs_utils.py#L43
2. If some layer-wise unit tests take a long time to run in CI, we can consider skipping them in our per-commit CI (but still save the files in the repo for your local debugging) and only run the whole model CI.

test/srt/models/test_nvidia_nemotron_nano_v2.py
```diff
@@ -0,0 +1,44 @@
...
+            data_path=None,
+            num_questions=200,
+            max_new_tokens=512,
+            parallel=128,
+            host="http://127.0.0.1",
+            port=int(self.base_url.split(":")[-1]),
+        )
+        metrics = run_eval(args)
+        print(f"{metrics=}")
+        self.assertGreater(metrics["accuracy"], 0.87)
```
merrymercy, line 44: We need to add this. Otherwise, it won't run by the CI.

https://github.com/sgl-project/sglang/blob/55b14656e656c0a0be3d9a24e69f624c957429e7/test/srt/test_abort.py#L113-L114

python/sglang/srt/layers/attention/attention_registry.py
```diff
@@ -166,36 +173,41 @@ def create_dual_chunk_flash_attn_backend(runner):
...
 
-    # wrap for hybrid GDN models
-    if runner.is_hybrid_gdn:
-        from sglang.srt.utils import is_blackwell, is_npu
-
-        if is_blackwell():
-            assert (
-                runner.server_args.attention_backend == "triton"
-                or runner.server_args.attention_backend == "trtllm_mha"
-            ), "triton or trtllm_mha backend are the only supported backends on Blackwell GPUs for hybrid GDN models, use --attention-backend triton or --attention-backend trtllm_mha to specify the backend."
```
samuellees, line 186: Here disable trtllm_mha

PR #10920, opened 2025-09-25
Enables draft_extend in the TRT-LLM MLA backend

pranavm-nvidia:
## Motivation

Enabling draft_extend in the TRT-LLM MLA backend allows us to use it for MTP without requiring any fallback to the FlashInfer backend. 

## Modifications

Removes the logic that would cause draft_extend to fall back to the FlashInfer backend.

## Accuracy Tests

Accuracy reports are not useful for this change since the accuracy is not affected by the correctness of the draft model; if the draft model is wrong, this will be caught during `target_verify` and the final model output will be correct. Only the inference speed is impacted by the correctness of the draft model, so I have only included performance benchmark results for this PR. 

## Checklist


pranavm-nvidia:
Looks like there are some issues with my implementation (the accept length decreases slightly). Looking into that now.

PR #10973, opened 2025-09-26
Add `RequestMetricsExporter` utility to export request-level metrics

scottjlee:
## Motivation

When SGLang is used in production, it is critical to track information at a request level (e.g. performance or quality metrics), and also be able to later analyze this data in a convenient manner. For instance, one may want to analyze a set of requests served by an SGLang instance, and want to only analyze a subset (e.g. requests with high TTFT). Currently, the best way to accomplish this is to parse the log output, and filter on desired requests. This is limiting, since the data is not outputted in a structured, standardized format in the logs ready for analysis.

Ideally, we'd like to have the option to export request-level metrics in a structured format, so that they can be easily exported to various data sources depending on use case (file on disk, data streams, DB, etc).

## Modifications

- Adds an optional functionality to export request-level information to an external datasource. This is implemented in the `RequestMetricsExporter` abstract base class.
- Also include a concrete implementation `FileRequestMetricsExporter`, which writes the request-level data to files on disk. This is disabled by default, and can be enabled with the launch command parameter `--export-metrics-to-file`.

### Example Usage
```
python3 -m sglang.launch_server --host 0.0.0.0 --port 8001 --enable-metrics --log-requests --log-requests-level 0 \
--model-path meta-llama/Llama-3.1-8B-Instruct --export-metrics-to-file --export-metrics-to-file-dir sglang_request_metrics/
```

This will output request-level metrics to JSONL-format files under `sglang_request_metrics/` directory (as configured in launch server params). Below is a sample file output for several requests:
```
{"request_parameters": "{\"rid\": \"b22015035c5143a3a571f0d2e0de8d0b\", \"return_logprob\": false, \"logprob_start_len\": -1, \"top_logprobs_num\": 0, \"return_text_in_logprobs\": true, \"stream\": true, \"log_metrics\": true, \"return_hidden_states\": false, \"background\": false, \"no_logs\": false, \"return_bytes\": false, \"return_entropy\": false}", "id": "b22015035c5143a3a571f0d2e0de8d0b", "finish_reason": {"type": "length", "length": 100}, "prompt_tokens": 16, "weight_version": "default", "completion_tokens": 100, "cached_tokens": 0, "spec_accept_rate": 0.4603174603174603, "spec_accept_length": 2.380952380952381, "spec_verify_ct": 42, "e2e_latency": 0.45075368881225586}
{"request_parameters": "{\"rid\": \"1466795eff804efeb82e970d8cca331c\", \"return_logprob\": false, \"logprob_start_len\": -1, \"top_logprobs_num\": 0, \"return_text_in_logprobs\": true, \"stream\": true, \"log_metrics\": true, \"return_hidden_states\": false, \"background\": false, \"no_logs\": false, \"return_bytes\": false, \"return_entropy\": false}", "id": "1466795eff804efeb82e970d8cca331c", "finish_reason": {"type": "stop", "matched": 2}, "prompt_tokens": 16, "weight_version": "default", "completion_tokens": 49, "cached_tokens": 14, "spec_accept_rate": 0.23809523809523808, "spec_accept_length": 1.75, "spec_verify_ct": 28, "e2e_latency": 0.259782075881958}
{"request_parameters": "{\"rid\": \"9bd45c49ee6c4b41b12f05e91199547f\", \"return_logprob\": false, \"logprob_start_len\": -1, \"top_logprobs_num\": 0, \"return_text_in_logprobs\": true, \"stream\": true, \"log_metrics\": true, \"return_hidden_states\": false, \"modalities\": [], \"background\": false, \"no_logs\": false, \"return_bytes\": false, \"return_entropy\": false}", "id": "9bd45c49ee6c4b41b12f05e91199547f", "finish_reason": {"type": "stop", "matched": 2}, "prompt_tokens": 23, "weight_version": "default", "completion_tokens": 32, "cached_tokens": 21, "spec_accept_rate": 0.42857142857142855, "spec_accept_length": 2.2857142857142856, "spec_verify_ct": 14, "e2e_latency": 0.14313340187072754}
```

Some possible use cases for these output files:
- Import the JSONL file into your favorite data analysis environment or tool, for further analysis.
- Read and forward data in these files to external data sources, to ingest into data systems. For example, one can use [Vector](https://vector.dev/) to watch for new logs for files in configured directory, and forward the data to various destinations.

## Accuracy Tests

Not expected to impact model output/accuracy.

## Checklist


python/sglang/srt/managers/request_metrics_exporter.py
```diff
@@ -0,0 +1,119 @@
...
+            # Get the log file path for the current time.
+            current_time = datetime.now()
+            hour_suffix = current_time.strftime("%Y%m%d_%H")
+            log_filename = f"sglang-request-metrics-{hour_suffix}.log"
+            log_filepath = os.path.join(self.export_dir, log_filename)
+
+            metrics_data = self._format_output_data(obj, out_dict)
+
+            def write_file():
+                with open(log_filepath, "a", encoding="utf-8") as f:
```
acelyc111: Have you evaluate the overhead of open and close the file for every request? How about keeping the file handler for an hour?

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -1123,6 +1130,40 @@ def configure_logging(self, obj: ConfigureLoggingReq):
         logging.info(f"Config logging: {obj=}")
         self.log_request_metadata = self.get_log_request_metadata()
 
+    def _create_request_metrics_exporters(self) -> List[RequestMetricsExporter]:
```
acelyc111: How about moving theses code to the newly added request_metrics_exporter.py file, add a "manager" to manage the exporters? It would minimize the changes in tokenizer_manager.py.

acelyc111 (commented):
BTW, could you please give a sample to show how to use the output data?

python/sglang/srt/managers/request_metrics_exporter.py
scottjlee: I ran some local benchmarks with `bench_serving.py` and there was no significant impact to TPS and TTFT (possibly because the `write_file()` method is called async). I think your suggestion with keeping the file handler open is more efficient, let me re-implement to that.

scottjlee:
> BTW, could you please give a sample to show how to use the output data?

@acelyc111 thanks for the review, added more details and examples under the `Example Usage` section in the PR description.

acelyc111 (approved):
LGTM

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -288,6 +289,11 @@ def __init__(
         self.dump_requests_threshold = 1000
         self.dump_request_list: List[Tuple] = []
         self.log_request_metadata = self.get_log_request_metadata()
+        # Initialize performance metrics loggers with proper skip names
+        _, obj_skip_names, out_skip_names = self.log_request_metadata
+        self.request_metrics_exporter_manager = RequestMetricsExporterManager(
+            self.server_args, obj_skip_names, out_skip_names
+        )
```
ShangmingCai, line 331: Should these lines be wrapped in a
`if self.server_args.export_metrics_to_file:` condition?

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -951,6 +957,11 @@ async def _wait_one_response(
                         msg = f"Finish: obj={dataclass_to_string_truncated(obj, max_length, skip_names=skip_names)}, out={dataclass_to_string_truncated(out, max_length, skip_names=out_skip_names)}"
                     logger.info(msg)
 
+                # Asynchronously write performance metrics for this request using the exporter manager.
+                asyncio.create_task(
+                    self.request_metrics_exporter_manager.write_record(obj, out)
+                )
+
```

python/sglang/srt/managers/tokenizer_manager.py
scottjlee: since `--export-metrics-to-file` is corresponding to the specific `FileRequestMetricsExporter`, we don't want to gate this `RequestMetricsExporterManager` initialization here; in the case no exporter server launch args are enabled, we initialize a `RequestMetricsExporterManager` with no `._exporters` created, which has minimal footprint.

python/sglang/srt/managers/tokenizer_manager.py
scottjlee: good point, we definitely want to gate this part. i added a `RequestMetricsExporterManager.exporter_enabled()` method returning whether there is at least one `RequestMetricsExporterManager` enabled, and use it to gate this. now, this part will only run if there's at least one metrics exporter enabled via server launch args.

PR #10990, opened 2025-09-27
relax h200 cuda graph max bs in tp <= 4 to avoid oom

BBuf:


merrymercy:
move to https://github.com/sgl-project/sglang/pull/10975/files#diff-700b5118b493d60d7b5994857f5f1e6a7e842ad702392b8ab199945764dfc8ed

PR #10930, opened 2025-09-26
fix: make inference deterministic for large TP

JustinTong0323:
## Motivation
Close #10785
This PR enables deterministic inference for models using TP > 1.

## Modifications
Set `NCCL_ALGO="allreduce:tree"` and disable custom all-reduce when TP > 1.

## Accuracy Tests
Model Type | TP size | Backend | Status
-- | -- | -- | --
Dense & MoE | 2, 4, 8 | FlashInfer | ✅ Passed
Dense & MoE | 2, 4, 8 | Triton | ✅ Passed
Dense & MoE | 4, 8 | FlashAttention-3 | ✅ Passed
Dense | 2 | FlashAttention-3 | ✅ Passed
MoE | 2 | FlashAttention-3 | ❌ Partial Failed (Corner Case)¹

Tests passed for both Dense/MoE models (TP=2,4,8) across all backends (fa3, flashinfer, triton) using `sglang.test.test_deterministic`.

However, we found one exception: the `prefix` test failed for ~~MoE~~Qwen3-30B-A3B with TP=2 and the fa3 backend, generating two unique outputs for prefix length 1.(But `mistralai/Mixtral-8x7B-v0.1` Could pass)

> We use `Qwen3-32B` for dense and `Qwen3-30B-A3B` for MoE

## Reproduce
1. Launch server:
  ```
  python3 -m sglang.launch_server \
      --model Qwen/Qwen3-30B-A3B \
      --port 30000 \
      --enable-deterministic-inference \
      --attention-backend fa3 \
      --tp-size 2
  ```
2. Run test:
```
python3 -m sglang.test.test_deterministic --test-mode prefix
```

You can see result like:
```
Prompt 0 with prefix length 1: total samples: xxx, Unique samples: 2
```

## Checklist

- [ ] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
[...]

PR #10994, opened 2025-09-27
[logprobs] Enable local deterministic logrprobs testing with strict threshold

PrinsYin:
**Motivation**

Numerical logprobs are highly sensitive to GPU type, kernel backend, PyTorch version, and even kernel launch order. This makes CI-based validation unreliable and noisy.

This PR introduces a **local, developer-run logprobs test** that ensures:

- bitwise consistency of logprobs under deterministic kernels
- reproducible diffs before & after code changes

---

### Updated Test Logic

1. **Baseline Generation**
   - Fetch 1000 samples from the ShareGPT dataset
   - Randomly choose prompt start positions  
   - Run one‑step generation with deterministic kernels  
   - Save per‑sample metadata into `sglang_baseline_local.pkl`

2. **Comparison Phase**
   - Load the local baseline and re‑run inference on the same prompts  
   - For each sample:
     - Compare per‑token logprobs for overlapping top‑k tokens  
     - Assert numerical deltas within `1e‑5` tolerance  
     - Verify that `return_logprob` flags behave correctly  
   - The test fails fast with detailed per‑sample diffs if thresholds are exceeded
---

**Key Changes**

- 🔄 **Refactored** `test_logprobs.py` into a standalone test runner with two modes:
  - `gen`: Generate a local baseline from 1000 ShareGPT samples
  - `test`: Compare current outputs against the local baseline
- **Enabled deterministic execution** via `enable_deterministic_inference=True`
- **Enforced strict logprobs thresholds**:
  - `max Δ ≤ 1e-5`

- CLI usage:

```bash
# Step 1: generate baseline
python test/srt/test_logprobs.py gen

# Step 2: test after code changes
python test/srt/test_logprobs.py test
```

---

**Next Steps**

We recommend all logprobs-related PRs include test results using this local testing suite to ensure stability across future kernel changes.


zhaochenyang20 (approved):

Fridge003:
@PrinsYin Will changing to another attention backend (like triton or flashinfer) help with passing CI?
In #10930 we also spotted some internal indeterministic behavior of fa3 kernels.
Or lowering the temperature (maybe to 0) can also help?

Fridge003:
Also I feel that it's possible for deterministic feature to output different tokens from original output. Since deterministic might change the kernel behavior.

test/srt/test_logprobs.py
aftersnow: In deterministic mode, TOP_K / 2 can be changed to TOP_K

PrinsYin: thanks, make sense!

test/srt/test_logprobs.py
```diff
@@ -48,6 +48,8 @@ def setUpClass(cls):
         cls.engine = sgl.Engine(
             model_path=DENSE_MODEL_NAME,
             random_seed=42,
+            attention_backend="flashinfer",
```
merrymercy, line 219: try  triton?

PrinsYin:
@aftersnow can you help review this pr and use this sciprt to validate the correctness of #6318 ? thanks!

test/srt/test_logprobs.py
```diff
@@ -1,42 +1,208 @@
-import io
+"""
+Logprobs Accuracy Test for SGLang
+
+The test is designed to be run locally by users to
```
zhaochenyang20: Add  the reason that,

With determined/batch invariant kernels, we can make sure that with the same input for SGLang, we can get exactly the same results on its log probs. However, log probs are highly sensitive to GPU machines, kernels, torch version and a lot of other factors, so we cannot have a unified generated log probs baseline across different machines.

test/srt/test_logprobs.py
```diff
@@ -1,42 +1,208 @@
...
+verify logprobs accuracy before contributing changes on related code.
+We really appreciate your effort and contribution to SGLang!
+
+### Contributing
+
+When submitting changes that affect logprobs computation:
+1. Run this test locally to ensure accuracy
```
zhaochenyang20: 1. make baseline
2. run it locally
3. submit the results

test/srt/test_logprobs.py
```diff
@@ -1,42 +1,208 @@
...
+2. Include test results in your PR description
+
+
+## How to Run Locally
+### Usage
+
+This script supports two modes: **generation** and **testing**.
```
zhaochenyang20: it's too AI generated 😂 remove the `###` and `**` in the doc strings.

test/srt/test_logprobs.py
```diff
@@ -1,42 +1,208 @@
...
+python test/srt/test_logprobs.py test
+```
+This will test your changes against the locally generated baseline from Step 1.
+
+### Interpreting Results
+
+The test will output:
+- `max Δ`: Maximum difference across all logprobs
+- `mean Δ`: Mean difference across all logprobs
+- Pass/fail status based on tolerance thresholds
```
zhaochenyang20: Add what does this test do. Explaining that this test will generated and test:

generate:

1. using 1000 reqeusts to get log probs;

test:

1. the boundary of log probs requests;
2. the log probs is invariant;

zhaochenyang20 (changes requested):

test/srt/test_logprobs.py
```diff
@@ -1,46 +1,44 @@
...
+With deterministic/batch invariant kernels, we can make sure that with the same input for SGLang,
+we can get exactly the same results on its log probs.
+However, log probs are highly sensitive to GPU machines, kernels, torch version and a lot of other factors,
+so we cannot have a unified generated log probs baseline across different machines.
+
+This test is designed to be run locally by contributors to
+verify logprobs accuracy before making changes on related code.
+When submitting changes that affect logprobs computation, please:
 1. Run this test locally to ensure accuracy
 2. Include test results in your PR description
```
zhaochenyang20: 1. generate baseline
2. run test
3. submit results

test/srt/test_logprobs.py
```diff
@@ -1,46 +1,44 @@
...
+
+======================
+What does this test do?
+This test will fetch 1000 requests from ShareGPT dataset, and generate log probs for each request and save it as a baseline.
+Then by running the test mode, it will test the accuracy of the log probs by comparing it with the baseline.
 
-## How to Run Locally
-### Usage
+This test makes sure that:
+- the boundary of log probs requests are correct;
```
zhaochenyang20: - the boundary of log probs requests are correct, eg, the index for tokens that required log probs are strictly followed

zhaochenyang20 (approved):
Great to go. Two left minor comments.

PR #11003, opened 2025-09-28
Fix FusedSetKVBufferArg  in RotaryEmbedding

merrymercy:


FlamingoPg (approved):

yuan-luo:
The assert in RotaryEmbedding is unnecessary.
```
    def forward_native(
        self,
        positions: torch.Tensor,
        query: torch.Tensor,
        key: torch.Tensor,
        offsets: Optional[torch.Tensor] = None,
        fused_set_kv_buffer_arg: Optional[FusedSetKVBufferArg] = None,
    ) -> Tuple[torch.Tensor, torch.Tensor]:
        """A PyTorch-native implementation of forward()."""
        assert (
            fused_set_kv_buffer_arg is None
        ), "fused_set_kv_buffer_arg is not supported for native implementation"

```

PR #11005, opened 2025-09-28
[HiCache]bug fix: fixed blank item in host_mem_release_queue

zhangzuo21:
After enabling HiCache L3 and running with multiple tensor parallelism (TP) processes, the sglang server would eventually hang under sustained high-intensity workloads. The issue arises because the append_host_mem_release function may receive an empty host_indices tensor. However, splitting an empty tensor produces a chunks object that is not truly empty—it contains a single empty chunk. As a result, the number of chunks released through drain_storage_control_queues is the same across ranks, but some ranks end up releasing empty chunks. Over time, this discrepancy accumulates, leading to significant differences in available free memory across HiCache instances on different ranks. Consequently, the results returned by budget_state() may diverge, which in turn causes add_one_req to yield inconsistent outcomes across ranks, ultimately leading to a deadlock.


huangtingwei9988 (approved):
LGTM

huangtingwei9988:
cc @xiezhq-hermann

python/sglang/srt/mem_cache/memory_pool_host.py
```diff
@@ -172,6 +172,7 @@ def alloc(self, need_size: int) -> Optional[torch.Tensor]:
 
     @synchronized
     def free(self, indices: torch.Tensor) -> int:
+        indices = indices.unique()
```
ykwd: Why is unique called here? Are there scenarios where the indices might not be unique?

python/sglang/srt/managers/cache_controller.py
```diff
@@ -577,7 +577,8 @@ def terminate_prefetch(self, operation):
     def append_host_mem_release(self, host_indices: torch.Tensor):
         chunks = host_indices.split(self.mem_pool_host.page_size)
         for chunk in chunks:
-            self.host_mem_release_queue.put(chunk)
+            if len(chunk) > 0:
+                self.host_mem_release_queue.put(chunk)
```
xiezhq-hermann: should we just do a check on `host_indices`? that would be easier and less operations

ykwd: Good idea. We will update this.

ykwd: Updated @xiezhq-hermann

xiezhq-hermann (approved):

PR #11007, opened 2025-09-28
fix(decode): adjust ServerArgs import to explicit module path

xiaguan:
main branch
```
    from sglang.srt.entrypoints.engine import _launch_subprocesses
  File "/work/sjy/sglang/python/sglang/srt/entrypoints/engine.py", line 45, in <module>
    from sglang.srt.managers.data_parallel_controller import (
  File "/work/sjy/sglang/python/sglang/srt/managers/data_parallel_controller.py", line 41, in <module>
    from sglang.srt.managers.scheduler import run_scheduler_process
  File "/work/sjy/sglang/python/sglang/srt/managers/scheduler.py", line 47, in <module>
    from sglang.srt.disaggregation.decode_kvcache_offload_manager import (
  File "/work/sjy/sglang/python/sglang/srt/disaggregation/decode_kvcache_offload_manager.py", line 7, in <module>
    from sglang import ServerArgs
```
After this fix, it works fine.

## Checklist


xiaguan:
cc @hzh0425

hzh0425 (approved):

hzh0425:
thanks  @xiaguan

xiezhq-hermann (approved):

hnyls2002:
@xiaguan Try installing with `--config-settings editable_mode=compat`. This error may be related to your environment and package installation.

JustinTong0323:
Analysis given by gemini:

The Problem: Import Shadowing
The root cause of the issue is import shadowing.

When you run git clone sglang, it creates a local folder named sglang. Because Python's import system (sys.path) searches the current working directory first, any Python script run from that location will find the local sglang folder and treat it as the package. This "shadows" the actual, installed version of the library located in site-packages, leading to import errors or unexpected behavior.

The Solution: editable_mode=compat
You are correct that `pip install -e . --config-settings editable_mode=compat` can fix this.

This command changes how the editable install works. Instead of using the modern import-hook system, it falls back to the legacy .egg-link method. This method directly adds the explicit path to your cloned source code into Python's sys.path.

By making the path to the intended package source explicit, it resolves the ambiguity. Python is no longer confused by the local folder because it now has a more specific entry in its search path pointing to the correct, editable installation. This effectively bypasses the shadowing problem.

PR #11015, opened 2025-09-28
fix: check if weights are already local before downloading

mickqian:
Note: simply checking if a local directory exists may lead to problems like, weights incomplete
## Checklist

### DIFF
--- a/python/sglang/srt/model_loader/weight_utils.py
+++ b/python/sglang/srt/model_loader/weight_utils.py
@@ -8,7 +8,6 @@ import hashlib
 import json
 import logging
 import os
-import queue
 import tempfile
 from collections import defaultdict
 from typing import (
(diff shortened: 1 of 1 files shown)


python/sglang/srt/model_loader/weight_utils.py
```diff
@@ -260,6 +260,77 @@ def download_weights_from_hf(
     Returns:
         str: The path to the downloaded model weights.
     """
+    if is_in_ci():
+        found_local_snapshot_dir = None
```
merrymercy: wrap the whole thing into a function

JustinTong0323:
Maybe you shall check this PR with Kangyan to avoid conflict: https://github.com/sgl-project/sglang/pull/10966

merrymercy (approved):

PR #11028, opened 2025-09-28
feat(hicache): support numa detect to reduce long tail latency

xiaguan:
Hi from [novita.ai](https://novita.ai/) team

## Motivation

For machines with multi-GPU, multi-NIC, and multi-NUMA setups, topology is super important.

It doesn't just slightly hurt performance—it can significantly impact long-tail latency (e.g., p99 and max).

In this PR, I added --hicache-numa-detect (disabled by default) to bind the TP workers to the correct NUMA node (both CPU and memory).

I also updated the benchmark script to report p99 and max latency, so we can clearly see long-tail issues.

Most of the performance gain comes from RDMA transfers—now we avoid cross-NUMA RDMA traffic entirely.

Here are the benchmark results: H800 × 8, 200Gbps RDMA × 8, TP=8, Qwen-235B with page size 64.

not set `--hicache-numa-detect`

```
All requests completed
Performance metrics summary:
  Total requests: 800 at 16.0 requests per second
  Average Prompt Length: 11231.94 tokens
  Average Output Length: 1.00 tokens
  Average TTFT: 0.86
  P90 TTFT: 0.77
  P99 TTFT: 14.59
  Median TTFT: 0.62
  Max TTFT: 18.98
  Average latency: 0.86
  P90 latency: 0.77
  P99 latency: 14.59
  Median latency: 0.62
  Max latency: 18.98
  Input token throughput: 49544.09 tokens per second
  Output token throughput: 4.41 tokens per second
  Request Throughput: 4.41 requests per second
  Cache Hit Rate: 0.794541
```
The hit ratio dropped slightly due to Mooncake's lease mechanism. Long-tail transfer times can exceed the lease duration, causing read failures.

Same machine, same setup script with `--hicache-numa-detect`

```
All requests completed
Performance metrics summary:
  Total requests: 800 at 16.0 requests per second
  Average Prompt Length: 11231.92 tokens
  Average Output Length: 1.00 tokens
  Average TTFT: 0.60
  P90 TTFT: 0.67
  P99 TTFT: 0.74
  Median TTFT: 0.61
  Max TTFT: 0.83
  Average latency: 0.60
  P90 latency: 0.67
  P99 latency: 0.74
  Median latency: 0.61
  Max latency: 0.83
  Input token throughput: 70869.97 tokens per second
  Output token throughput: 6.31 tokens per second
  Request Throughput: 6.31 requests per second
  Cache Hit Rate: 0.815939
```

## Checklist


xiaguan:
cc @huangtingwei9988

huangtingwei9988:
> cc @huangtingwei9988

@xiaguan Thank you so much！！

python/sglang/srt/mem_cache/storage/mooncake_store/mooncake_store.py
```diff
@@ -163,6 +164,11 @@ def __init__(self, storage_config: HiCacheStorageConfig = None):
             if ret_code:
                 logger.error(f"failed to setup mooncake store, error code: {ret_code}")
 
+            if per_tp_global_segment_size != 0:
+                self.replica_config = ReplicateConfig()
+                self.replica_config.replica_num = 1
+                self.replica_config.preferred_segment = self.store.get_hostname()
+
```
stmatengss: Replica configuration not required? Since the default replica config is replica_num = 1 and use local segment.

hzh0425 (approved):
LGTM

xiezhq-hermann (approved):

python/sglang/srt/utils/common.py
```diff
@@ -3773,3 +3780,62 @@ def get_or_create_event_loop():
...
+    prefix = "NUMA IDs of closest CPU:"
+
+    if not output_line.startswith(prefix):
+        raise RuntimeError(
+            f"Unexpected nvidia-smi output: '{output_line}'. "
+            f"Expected line starting with '{prefix}'."
+        )
+
+    try:
+        numa_id_str = output_line[len(prefix) :].strip()
```
hzh0425: How to handle this situation?
'NUMA IDs of closest CPU: 0-1'

xiaguan: I believe nvidia-smi -C only returns a single integer?
```
[-C | --get-numa-id-of-nearby-cpu]: Displays NUMA ID of the nearest CPU for a GPU
                 Used in conjunction with -i which must be a single device ID.
```

hzh0425: But this result is exactly what I obtained on the H20 machine..

hzh0425: If this situation occurs, does it mean that each device is connected to every NUMA node?

xiaguan: The issue couldn't be reproduced on my H2O, perhaps due to different CPU platforms.

I'll apply a simple fix by just taking the first one.

xiaguan: Could u plz give me your `nvidia-smi topo -m` ? Tnks

hzh0425: > The issue couldn't be reproduced on my H2O, perhaps due to different CPU platforms.
> 
> I'll apply a simple fix by just taking the first one.

A better approach might be to bind the first four GPUs to NUMA0 and the last four to NUMA1.

xiaguan: For certain AMD platforms, the machine may possess eight NUMA domains.

xiaguan: Fixed

huangtingwei9988 (approved):

PR #11041, opened 2025-09-29
Implement LRU eviction policy for LoRA adapters

ConnorLi96:
## Motivation

For addressing this feature: [[Feature] (2/2) Support LRU cache for LoRA eviction ](https://github.com/sgl-project/sglang/issues/8696)
 
This PR implements a configurable LRU (Least Recently Used) eviction policy for LoRA adapters to provide more intelligent memory management. Currently, SGLang only supports FIFO eviction, which may not be optimal for workloads where certain LoRA adapters are accessed more frequently than others. The LRU policy ensures that frequently used adapters remain in memory while less recently used ones are evicted first, potentially improving cache hit rates and overall performance.

## Modifications

- **New eviction policy framework**: Added `eviction_policy.py` module with abstract `EvictionPolicy` class
- **LRU implementation**: Implemented `LRUEvictionPolicy` using OrderedDict for O(1) access tracking
- **FIFO compatibility**: Maintained existing `FIFOEvictionPolicy` for backward compatibility
- **Server configuration**: Added `--lora-eviction-policy` argument to `ServerArgs` with choices ["fifo", "lru"]
- **Memory pool integration**: Updated `LoRAMemoryPool` to use configurable eviction policies
- **Manager coordination**: Modified `LoRAManager` to pass eviction policy to memory pool
- **Test runner support**: Extended `SRTRunner` to accept eviction policy parameter
- **Adapter pinning**: Implemented mechanism to prevent specific adapters from being evicted

All changes maintain full backward compatibility with default FIFO behavior.

## Accuracy Tests

This PR does not affect model outputs or inference accuracy.

## Benchmarking and Profiling

The LRU eviction policy is designed to improve cache efficiency for workloads with non-uniform adapter access patterns. Performance impact is minimal:
- **Memory overhead**: Negligible (OrderedDict for tracking access order)
- **Eviction decision**: O(1) time complexity for both LRU and FIFO
- **Access tracking**: O(1) time complexity per adapter use

Detailed benchmarking will be conducted with realistic workloads in future testing.

## Checklist


python/sglang/srt/lora/eviction_policy.py
```diff
@@ -0,0 +1,124 @@
...
+    """LRU eviction policy - evicts the least recently used adapter."""
+
+    def __init__(self):
+        self.access_order = OrderedDict()  # key=uid, value=last_access_time
+        self.total_accesses = 0
+        self.eviction_count = 0
+
+    def mark_used(self, uid: Optional[str]) -> None:
+        if uid is not None:
+            current_time = time.time()
```
lifuhuang: Please use `time.monotonic()`. Wall clock is not reliable for comparison purposes.

python/sglang/srt/lora/lora_manager.py
```diff
@@ -411,6 +415,7 @@ def init_memory_pool(self):
             max_lora_rank=self.max_lora_rank,
             target_modules=self.target_modules,
             base_model=self.base_model,
+            eviction_policy=eviction_policy,
         )
```
lifuhuang, line 437: nit: remove L408 for consistency?

python/sglang/srt/lora/mem_pool.py
```diff
@@ -189,31 +194,45 @@ def prepare_lora_batch(
...
+            # Collect eviction candidates (not in current batch)
+            candidates = set()
+            pinned_uids = set()
+            
             for buffer_id in range(self.max_loras_per_batch):
                 uid = self.buffer_id_to_uid[buffer_id]
+                if uid not in cur_uids and uid is not None:
+                    candidates.add(uid)
+                    lora_ref = lora_refs.get(uid)
+                    if lora_ref is not None and lora_ref.pinned:
```
lifuhuang: Do we need to pass the pinned_uids here or we can simplify the logic by only collecting entries that are not pinned?

python/sglang/srt/lora/eviction_policy.py
```diff
@@ -0,0 +1,124 @@
...
+                return uid
+        return None
+
+    def remove(self, uid: Optional[str]) -> None:
+        if uid is not None:
+            self.access_order.pop(uid, None)
+            logger.debug(f"Removed LoRA {uid} from LRU tracking")
+
+
+class FIFOEvictionPolicy(EvictionPolicy):
```
lifuhuang, line 90: can you double check the current behavior is actually FIFO? I remember we simply evict the lowest index ID that's not used but I could be wrong.

python/sglang/srt/lora/eviction_policy.py
ConnorLi96: make sense!

python/sglang/srt/lora/eviction_policy.py
ConnorLi96: ah, nice catch! original code: https://github.com/sgl-project/sglang/blob/ab62b135c18ad5d655aac52d9010004d461346ce/python/sglang/srt/lora/mem_pool.py#L209

I thought you want to implement a FIFO, but it just simply pop the lowest buffer index. I think we should probably deprecate it since random? just keep LRU and FIFO, and set LRU as default. WDYT?

python/sglang/srt/lora/mem_pool.py
ConnorLi96: make sense! we can remove it

python/sglang/srt/lora/lora_manager.py
ConnorLi96: done!

test/srt/lora/test_lora_eviction_policy.py
```diff
@@ -0,0 +1,160 @@
...
+Unit tests for LoRA eviction policies.
+Tests LRU, FIFO, and adapter pinning functionality.
+"""
+
+import unittest
+
+from sglang.srt.lora.eviction_policy import get_eviction_policy
+
+
+class TestLoRAEvictionPolicy(unittest.TestCase):
```
lifuhuang, line 25: Please add your test to: https://github.com/sgl-project/sglang/blob/e5281f84d5bcbdfd6d7790c5ab450429e11291a6/test/srt/run_suite.py#L20

python/sglang/srt/lora/lora_manager.py
```diff
@@ -68,6 +68,13 @@ def __init__(
         self.tp_size: int = tp_size
         self.tp_rank: int = tp_rank
 
+        # Store eviction policy from server args
+        self.eviction_policy = (
```
lifuhuang: Please ensure default value is set in server_args for maintainability, we should safely assume lora_eviction_policy is not None. I believe server_args should also never be None, can you double check?

python/sglang/srt/lora/mem_pool.py
```diff
@@ -54,6 +55,7 @@ def __init__(
         max_lora_rank: int,
         target_modules: Set[str],
         base_model: torch.nn.Module,
+        eviction_policy: str = "fifo",
```
lifuhuang: Please keep all default logic to a single place (ServerArgs) for maintainability.

python/sglang/srt/lora/eviction_policy.py
```diff
@@ -0,0 +1,115 @@
...
+            self.access_order.pop(uid, None)
+            self.access_order[uid] = current_time
+            self.total_accesses += 1
+            logger.debug(f"LoRA {uid} marked as used at {current_time}")
+
+    def select_victim(self, candidates: Set[Optional[str]]) -> Optional[str]:
+        """Select the least recently used adapter from candidates."""
+        # Iterate through access_order (oldest first) to find LRU victim
+        for uid in list(self.access_order.keys()):
+            if uid in candidates and uid is not None:
```
lifuhuang: Do we ever store `None` in ordered dict?

python/sglang/srt/lora/eviction_policy.py
lifuhuang: Sounds reasonable to me

python/sglang/srt/lora/eviction_policy.py
```diff
@@ -0,0 +1,115 @@
...
+class FIFOEvictionPolicy(EvictionPolicy):
+    """FIFO eviction policy - for backward compatibility."""
+
+    def __init__(self):
+        self.insertion_order = OrderedDict()  # key=uid, value=insertion_time
+        self.eviction_count = 0
+
+    def mark_used(self, uid: Optional[str]) -> None:
+        """For FIFO, we only track insertion, not usage."""
+        if uid is not None and uid not in self.insertion_order:
```
lifuhuang, line 101: nit: this is never used right?

ConnorLi96: ah, "not usage" in comment just for distinguishing FIFO from LRU behavior, we actually use it for tracks insertion (time) only, then we're able to select the oldest adapter for eviction

python/sglang/srt/lora/lora_manager.py
ConnorLi96: ah, make a lot of sense!

python/sglang/srt/lora/mem_pool.py
ConnorLi96: yes, done, all moved to `ServerArgs` now

python/sglang/srt/lora/eviction_policy.py
ConnorLi96: good catch, we removed all `None` before `select_victim`

test/srt/lora/test_lora_eviction_policy.py
ConnorLi96: ok, will do later after we pass entire integration test. And I think we can probably deprecate the `TestFile("lora/test_lora_eviction.py", 200),`? if this all good?

test/srt/lora/test_lora_eviction_policy.py
```diff
@@ -0,0 +1,248 @@
...
+                    )
+                    raise
+
+        print(f"\n✓ {eviction_policy.upper()} test passed (pinned={pinned})")
+
+    def test_lru_eviction(self):
+        """Test LRU eviction policy with actual inference."""
+        self._run_eviction_test("lru", pinned=False)
+
+    def test_lru_eviction_with_pinning(self):
```
ConnorLi96: x-post for viz

My expected behavior (with max_loras_per_batch=2) here is:
- Load adapter_1 (pinned) → Buffer: [adapter_1, EMPTY]
- Load adapter_2 → Buffer: [adapter_1, adapter_2]
- Load adapter_3 → Should evict adapter_2, keep adapter_1 (pinned) → Buffer: [adapter_1, adapter_3]

But I got error like
`ValueError: No available buffer slots found. Please ensure the number of active loras is less than max_loras_per_batch.
`
The root cause is that we have base model (uid=None) in the buffer, and looks like [None is a valid uid](https://github.com/sgl-project/sglang/blob/ab62b135c18ad5d655aac52d9010004d461346ce/python/sglang/srt/lora/mem_pool.py#L80) in our code, so the behavior becomes
- Buffer state: [None(base model is here!),   adapter_1 (pinned)]
- Try to load adapter_2 with cur_uids = {adapter_2}
- Eviction logic checks:
  - None not in cur_uids :white_check_mark: (not needed in current batch)
  - adapter_1 not in cur_uids :white_check_mark: but **pinned** :no_good: (cannot evict)
- Result: No eviction candidates found → Error!

My question here is can we just evict the base model (None) in the buffer? since we don't need the base model in this batch, which we can ensure via this condition check `if uid not in cur_uids`

from @Fridge003 suggestion in slack we can do that, but wanted get approve from you @lifuhuang

python/sglang/srt/lora/lora_manager.py
```diff
@@ -131,6 +134,16 @@ def load_lora_adapter(self, lora_ref: LoRARef) -> LoRAUpdateResult:
             lora_ref.lora_id not in self.loras
         ), f"LoRA adapter with ID {lora_ref.lora_id} is already loaded. This should have been verified before request is sent to the backend."
 
+        # Check if pinning this adapter would exceed capacity
```
lifuhuang: Is the validation here not enough? https://github.com/sgl-project/sglang/blob/bfe931ffd12d8db078e6d7c5ae0a60f677845fc8/python/sglang/srt/lora/lora_manager.py#L183

python/sglang/srt/lora/mem_pool.py
```diff
@@ -189,31 +194,49 @@ def prepare_lora_batch(
...
+            # 2. Memory pool is full, need to evict using policy
+            candidates = set()
+
             for buffer_id in range(self.max_loras_per_batch):
                 uid = self.buffer_id_to_uid[buffer_id]
 
-                # Evict unneeded lora
-                if uid not in cur_uids:
-                    # Skip pinned LoRAs
-                    # TODO (lifuhuang): we might consider supporting pinning base model (uid == None) in the future.
```
lifuhuang, line 203: Should we keep this TODO?

python/sglang/srt/lora/mem_pool.py
```diff
@@ -189,31 +194,49 @@ def prepare_lora_batch(
...
-                    self.buffer_id_to_uid[buffer_id] = EMPTY_SLOT
-                    return buffer_id
+                # Skip if this adapter is needed by current batch
+                if uid in cur_uids:
+                    continue
+
+                # Skip if this adapter is pinned
+                if uid is not None:
+                    lora_ref = lora_refs.get(uid)
+                    if lora_ref and lora_ref.pinned:  # None (base model) can be evicted
```
lifuhuang: out of curiosity, how is this comment related to the pinned check here? Can you explain more?

python/sglang/srt/lora/eviction_policy.py
lifuhuang: Sorry, I meant the `time.monotonic()` do we actually use it at all?

python/sglang/srt/lora/eviction_policy.py
```diff
@@ -0,0 +1,135 @@
...
+            return BASE_MODEL_UID
+
+        # Iterate through insertion_order (oldest first) to find FIFO victim
+        for uid in list(self.insertion_order.keys()):
+            if uid in candidates:
+                logger.debug(f"Selected LoRA {uid} for eviction (FIFO)")
+                self.eviction_count += 1
+                return uid
+
+        # Should never reach here if candidates is non-empty
```
lifuhuang, line 123: Use assertion instead?

test/srt/lora/test_lora_eviction_policy.py
```diff
@@ -0,0 +1,218 @@
...
+import unittest
+
+import torch
+
+from sglang.srt.lora.eviction_policy import get_eviction_policy
+from sglang.test.runners import SRTRunner
+from sglang.test.test_utils import CustomTestCase
+
+
+class TestLoRAEvictionPolicy(unittest.TestCase):
```
lifuhuang, line 25: Suggestion: 
1. Add more tests cases -- currently we only test 1 test case per policy, also the logic seems quite ad-hoc and repetitive.

can we abstract the ad-hoc tests into a reusable function and add more test cases?

e.g.
```
def _test_lora_eviction(max_lora_per_batch, access_order, expected_victims)
    ...

def test_lru_eviction_policy(self):
    _test_lora_eviction(8, ["lora1", "lora2", "lora1", ...], "lora1")
    _test_lora_eviction(8, ["lora1", "lora2", "lora1", ...], "lora2")
    _test_lora_eviction(16, ["lora1", "lora2", "lora1", ...], ...)
    _test_lora_eviction(16, ["lora1", "lora2", "lora1", ...], ...)

def test_fifo_eviction_policy(self):
    _test_lora_eviction(8, ["lora1", "lora2", "lora1", ...], ...)
    _test_lora_eviction(8, ["lora1", "lora2", "lora1", ...], ...)
    _test_lora_eviction(16, ["lora1", "lora2", "lora1", ...], ...)
    _test_lora_eviction(16, ["lora1", "lora2", "lora1", ...], ...)
```

2. avoid breaching module abstractions -- I don't see much value in "simulating" the pinned test here: we should either treat eviction policy as the UUT (Unit Under Test), in which case we should only consider possible inputs/outputs for eviction policy without caring about HOW these inputs were created in the first place (e.g., whether the `candidates` are results of pinning is not relevant to the logic of `select_victim`) **OR** if we truly want to validate the pinning logic, we should treat memory pool or its methods as the UUT and only validate inputs/outputs of that.

python/sglang/srt/lora/lora_manager.py
ConnorLi96: make sense, we should always leave a slot even we evict the base model

python/sglang/srt/lora/eviction_policy.py
ConnorLi96: done

python/sglang/srt/lora/eviction_policy.py
ConnorLi96: ah, i see, deleted

python/sglang/srt/lora/mem_pool.py
ConnorLi96: added it back!

python/sglang/srt/lora/mem_pool.py
ConnorLi96: ah, sorry to make confusion, since the base model cannot be pinned, so can be evicted, move the comment the top of this if condition

test/srt/lora/test_lora_eviction_policy.py
ConnorLi96: cool! updated!
- Expanded to 12 unit tests with reusable helper function `_test_eviction_policy`
- Removed pinning tests (breaches module abstraction - should be tested in mem_pool tests instead)
- Kept 2 basic integration tests for E2E validation

lifuhuang (approved):
LGTM. Great work!

ConnorLi96:
> LGTM. Great work!

nice, thank you so much for the guidance all the way! Can we add run-ci label for this PR? or we can just merge it directly.

PR #11062, opened 2025-09-29
ci: reduce and refactor vlm ut and combine test files

mickqian:
## Modifications

1. models that are less popular, or with similar architectures (qwen2VL to qwen2_5VL) are removed from pr-test
2. resume a mistakenly deleted ut (`test_mixed_batch`)

## Checklist


test/srt/test_vision_openai_server_b.py
```diff
@@ -1,250 +0,0 @@
...
-            cls.model,
-            cls.base_url,
-            timeout=DEFAULT_TIMEOUT_FOR_SERVER_LAUNCH,
-            other_args=[
-                "--trust-remote-code",
-                "--mem-fraction-static",
-                "0.70",
-            ],
-        )
-        cls.base_url += "/v1"
```
JustinTong0323, line 191: We shall keep this as the only audio test?

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -1793,11 +1793,23 @@ def get_model_worker_batch(
...
+            and attention_backend_str == "flashinfer"
+        ):
+            # for meta-llama/Llama-3.2-11B-Vision-Instruct
+            seq_lens_cpu = None
+        else:
+            seq_lens_cpu = (
+                seq_lens_cpu_cache
+                if seq_lens_cpu_cache is not None
+                else self.seq_lens.cpu()
+            )
```
JustinTong0323: this is ugly bro () Why we use this?

mickqian: Previously mllama-3.1 is not included in Pr-test, when I add it back in, the bug above appears

test/srt/test_vision_openai_server_a.py
```diff
@@ -128,213 +80,106 @@ def test_single_image_chat_completion(self):
...
+        "--enable-multimodal",
+        "--cuda-graph-max-bs",
+        "4",
+    ]
+
+
+class TestKimiVLServer(ImageOpenAITestMixin):
+    model = "moonshotai/Kimi-VL-A3B-Instruct"
+    other_args = [
+        "--trust-remote-code",
```
JustinTong0323: trust remote code is typically not harmful, we could move it to base class

test/srt/test_vision_openai_server_a.py
```diff
@@ -128,213 +80,106 @@ def test_single_image_chat_completion(self):
...
+
+
+class TestGemma3itServer(ImageOpenAITestMixin):
+    model = "google/gemma-3-4b-it"
+    other_args = [
+        "--trust-remote-code",
+        "--mem-fraction-static",
+        "0.70",
+        "--enable-multimodal",
+        "--cuda-graph-max-bs",
```
JustinTong0323: Could we add the cuda graph max bs = 4 to base class?

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -1847,9 +1847,23 @@ def get_model_worker_batch(
...
+            attention_backend_str = global_server_args_dict["decode_attention_backend"]
+        else:
+            attention_backend_str = global_server_args_dict["prefill_attention_backend"]
+        # Create seq_lens_cpu when needed
+        if (
+            not global_server_args_dict["use_mla_backend"]
+            and attention_backend_str == "flashinfer"
+        ):
+            # for meta-llama/Llama-3.2-11B-Vision-Instruct
+            seq_lens_cpu = None
```
JustinTong0323: maybe wrap the if/else to a _maybe_unset_seq_len_cpu helper function for better readability?
like:

```suggestion
def _maybe_unset_seq_len_cpu(self):
       # for meta-llama/Llama-3.2-11B-Vision-Instruct
        if self.forward_mode.is_decode_or_idle():
            attention_backend_str = global_server_args_dict["decode_attention_backend"]
        else:
            attention_backend_str = global_server_args_dict["prefill_attention_backend"]
        # Create seq_lens_cpu when needed
        if (
            not global_server_args_dict["use_mla_backend"]
            and attention_backend_str == "flashinfer"
        ):
            return True
        return False

if _maybe_unset_seq_len_cpu(self):
            seq_lens_cpu = None
```

python/sglang/srt/managers/schedule_batch.py
JustinTong0323: /gemini why there is a bug?

test/srt/test_vision_openai_server_a.py
mickqian: better not, otherwise some model not being able to load in real cases will be able to in CI

JustinTong0323 (approved):
LGTM

merrymercy (approved):

test/srt/test_vision_openai_server_a.py
```diff
@@ -50,49 +34,29 @@ def setUpClass(cls):
...
-                "--cuda-graph-max-bs",
-                "4",
-            ],
-        )
-        cls.base_url += "/v1"
+class TestQwen25VLServer(ImageOpenAITestMixin, VideoOpenAITestMixin):
+    model = "Qwen/Qwen2.5-VL-7B-Instruct"
+    extra_args = [
+        "--mem-fraction-static",
+        "0.35",
```
merrymercy: - Can we remove all `--mem-fraction-static` in this file (and nightly tests) so that we can test our auto mem-fraction-static logics?
- Can you tune our auto mem fraction static logic to match here https://github.com/sgl-project/sglang/blob/af96ca1136caef34d3cae9e4c36267a29abe5ba7/docs/advanced_features/hyperparameter_tuning.md?plain=1#L43-L50. Maybe we need to reserve a little bit more for VLMs.

mickqian: yes we should reserve more memories for VLMs. 
Although the most frequent reasons for specifying a low `--mem-fraction-static` is for long-inputs, e.g. video inputs.
Except from video input, we might find a strategy to set the `--mem-fraction-static` well

mickqian: `--mem-fraction-static` in this file all removed

merrymercy (approved):

PR #11065, opened 2025-09-29
Feat/add nixl pp support

shaharmor98:
## Checklist

### DIFF
--- a/python/sglang/srt/disaggregation/test_disaggregation_nixl_utils.py
+++ b/python/sglang/srt/disaggregation/test_disaggregation_nixl_utils.py
@@ -4,7 +4,6 @@ from sglang.test.test_disaggregation_utils import TestDisaggregationBase
 from sglang.test.test_utils import (
     DEFAULT_MODEL_NAME_FOR_TEST,
     DEFAULT_TIMEOUT_FOR_SERVER_LAUNCH,
-    DEFAULT_URL_FOR_TEST,
     popen_launch_pd_server,
 )
 
@@ -15,16 +14,7 @@ class TestDisaggregationNixl(TestDisaggregationBase):
     @classmethod
     def setUpClass(cls):
         cls.model = DEFAULT_MODEL_NAME_FOR_TEST
-        parsed_url = urlparse(DEFAULT_URL_FOR_TEST)
-        cls.base_host = parsed_url.hostname
-        base_port = str(parsed_url.port)
-        cls.lb_port = base_port
-        cls.prefill_port = f"{int(base_port) + 100}"
-        cls.decode_port = f"{int(base_port) + 200}"
-        cls.prefill_url = f"http://{cls.base_host}:{cls.prefill_port}"
-        cls.decode_url = f"http://{cls.base_host}:{cls.decode_port}"
-        cls.lb_url = f"http://{cls.base_host}:{cls.lb_port}"
-        print(f"{cls.base_host=} {cls.lb_port=} {cls.prefill_port=} {cls.decode_port=}")
+        super().setUpClass()
 
     @classmethod
     def start_prefill(cls, prefill_args):
--- a/python/sglang/srt/disaggregation/test_disaggregation_nixl_utils.py
+++ b/python/sglang/srt/disaggregation/test_disaggregation_nixl_utils.py
@@ -1,8 +1,5 @@
-import unittest
-from types import SimpleNamespace
 from urllib.parse import urlparse
 
-from sglang.test.few_shot_gsm8k import run_eval as run_eval_few_shot_gsm8k
 from sglang.test.test_disaggregation_utils import TestDisaggregationBase
 from sglang.test.test_utils import (
     DEFAULT_MODEL_NAME_FOR_TEST,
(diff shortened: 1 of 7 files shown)


test/srt/run_suite.py
```diff
@@ -151,6 +151,7 @@ class TestFile:
         TestFile("test_disaggregation.py", 499),
         TestFile("test_disaggregation_different_tp.py", 155),
         TestFile("test_disaggregation_pp.py", 60),
+        TestFile("test_disaggregation_nixl.py", 35 * 60),
```
ShangmingCai: The total time taken for this test is going to be too long. We are currently fixing the CI for main. The new rule requires that each test file must have an estimated run time of less than 500 seconds. Perhaps we need to split it up.

ShangmingCai (commented):
The overall changes LGTM. But the test needs to be split to improve the user experience of CI (maybe split the PP one). We want all tests completed in 30 minutes.

test/srt/test_disaggregation_nixl.py
```diff
@@ -0,0 +1,136 @@
...
+            "--disaggregation-ib-device",
+            disaggregation_ib_device,
+        ]
+
+    def test_gsm8k_accuracy(self):
+        """Test GSM8K accuracy with NIXL disaggregation"""
+        test_cases = [
+            (1, 1, 1),  # 1 prefill TP, 1 prefill PP, 1 decode TP
+            (1, 1, 2),  # 1 prefill TP, 1 prefill PP, 2 decode TP
+            (2, 2, 2),  # 2 prefill TP, 2 prefill PP, 2 decode TP
```
ShangmingCai: Maybe we can remove these cases since they are basic compared to the remained cases. If (1, 2, 2), (2, 1, 4), (2, 2, 4) are passed, then these cases should be passed as well.

test/srt/test_disaggregation_nixl.py
```diff
@@ -0,0 +1,136 @@
...
+                self.wait_server_ready(self.prefill_url + "/health")
+                self.wait_server_ready(self.decode_url + "/health")
+
+                self.launch_lb()
+
+                # Run GSM8K evaluation
+                args = SimpleNamespace(
+                    num_shots=5,
+                    data_path=None,
+                    num_questions=200,
```
ShangmingCai: Maybe we can reduce this to 100 as well to reduce the CI elapsed time, but anyway, we can observe the actual cost at first, then we adjust the params if needed.

test/srt/test_disaggregation_nixl.py
shaharmor98: Although they are basic, they nevertheless test functionality (`send_kvcache`) that the other more "complex" configurations aren't testing (`send_kvcache_slice`). So IMO they are needed as well

test/srt/test_disaggregation_nixl.py
shaharmor98: I can do this, so it basically should halve the test-running time, right?

hnyls2002:
@shaharmor98  Great to see the NIXL backend will have CI tests soon. Please install the nixl backend like mooncake does https://github.com/sgl-project/sglang/blob/main/scripts/ci/ci_install_dependency.sh#L77

test/srt/test_disaggregation_nixl.py
ShangmingCai: The time cost is still too large. Currently, 8 GPU CI is limited to 20 minutes, which 19+- minutes is very dangerous and will make this test flaky. I will disable the (1,1,1) test to see if the estimated time is stable within 20 minutes.

trevor-m (approved):
LGTM!

python/sglang/srt/disaggregation/test_disaggregation_nixl_utils.py
```diff
@@ -0,0 +1,79 @@
...
+        parsed_url = urlparse(DEFAULT_URL_FOR_TEST)
+        cls.base_host = parsed_url.hostname
+        base_port = str(parsed_url.port)
+        cls.lb_port = base_port
+        cls.prefill_port = f"{int(base_port) + 100}"
+        cls.decode_port = f"{int(base_port) + 200}"
+        cls.prefill_url = f"http://{cls.base_host}:{cls.prefill_port}"
+        cls.decode_url = f"http://{cls.base_host}:{cls.decode_port}"
+        cls.lb_url = f"http://{cls.base_host}:{cls.lb_port}"
+        print(f"{cls.base_host=} {cls.lb_port=} {cls.prefill_port=} {cls.decode_port=}")
```
ShangmingCai:
```suggestion
       super().setUpClass()
```

test/srt/test_disaggregation_variable_parallelism_nixl.py
```diff
@@ -0,0 +1,69 @@
...
+)
+
+
+class TestDisaggregationVariableParallelismNixl(TestDisaggregationNixl):
+    """Test NIXL disaggregation functionality with variable TP"""
+
+    def test_gsm8k_accuracy(self):
+        """Test GSM8K accuracy with NIXL disaggregation"""
+        test_cases = [
+            (1, 1, 2),  # 1 prefill TP, 1 prefill PP, 2 decode TP
```
ShangmingCai, line 16: This case can be deleted and extracted from this file, you can create a _test_disaggregation_nixl_basic.py_ file and add to "per-commit-2-gpu" suite. 8 gpu runner is expensive, so we have to minimize the cost for each run.

ShangmingCai: My bad, this is "1 prefill TP, 1 prefill PP, 2 decode TP", forget about this comment above.

python/sglang/srt/disaggregation/test_disaggregation_nixl_utils.py
shaharmor98: you're completely right, removed

PR #9436, opened 2025-08-21
[Performance] Batch Send from Tokenizer Manager.

sundar24295s:
## Motivation
- The previous [PR #5141](https://github.com/sgl-project/sglang/pull/5141) introduced batch tokenization for parallel processing.
- However, when sending a batch of prompts (e.g., 50), the tokenizer manager currently transmits them individually via ZMQ socket. This can cause the batch to be split (e.g., 10 + 40) at the Scheduler, increasing per-request latency even if `max-prefill-tokens` is set appropriately.
 

## Modifications
- Send the entire batch in a single ZMQ message so the Scheduler can admit the full batch into prefill together.
- Added two small structs to make the batch handling and dispatch logic explicit and clean.
- Note: this change is part of the `enable-tokenizer-batch-encode` server arg which is disabled by default.

## Benchmarking and Profiling
- Model Qwen3-0.6B
- Input Token Length = 300
- GPU Type = H100
- Traffic distribution = Poisson
- Benchmark file: `sglang/benchmark/score/bench_score.py`

```
(sglang-repo3) jobuser [ /shared/user/repos3/sglang ]$ python -m sglang.launch_server --model-path /shared/public/sharing/suramach/Qwen3-0.6B --port 30000 --host 0.0.0.0 --chunked-prefill-size -1 --enable-torch-compile --dtype float16 --max-prefill-tokens 30000 --mem-fraction-static 0.3 --enable-tokenizer-batch-encode  --disable-cuda-graph
```
### Single Request Latency Comparison

| Batch Size | Method      | avg_response_time_ms | p50_response_time_ms | p90_response_time_ms | p99_response_time_ms |
|------------|-------------|---------------------|---------------------|---------------------|---------------------|
| 50         | Baseline    | 70.39               | 70.39               | 70.39               | 70.39               |
| 50         | Batch Send  | 41.12               | 41.12               | 41.12               | 41.12               |
| 100        | Baseline    | 94.28               | 94.28               | 94.28               | 94.28               |
| 100        | Batch Send  | 61.30               | 61.30               | 61.30               | 61.30               |

**Latency Reduction:**  
- For batch size 50, average latency decreased by **41.5%** (from 70.39 ms to 41.12 ms).
- For batch size 100, average latency decreased by **34.9%** (from 94.28 ms to 61.30 ms).

## Baseline logs: From logs we see a single batch request getting split
When sending a batch of 100 prompts
```
[...]

PR #11091, opened 2025-09-30
support tokenized batch request

narutolhy:
## Motivation
The previous PR https://github.com/sgl-project/sglang/pull/9436 only supported batched requests when the inputs were provided as text.
This PR adds support for batched requests when all inputs are input_ids, so the execution path is consistent with the text input case and avoids unnecessary per-request overhead.

## Modifications

Added a new condition to detect when a batch of requests contains only input_ids.
Reused the existing batch handling logic:
Each request is still tokenized via _tokenize_one_request (to wrap into the correct internal format).
All requests are then dispatched together through _send_batch_request.
Generators are created for each request in the batch to collect responses as before.

## Accuracy Tests

Verified locally using the test that batched input_ids requests return identical outputs to equivalent text requests (same token counts, logprob behavior, etc.).
I can add a single test, but I don’t think it’s necessary because this test is only for this PR

## Benchmarking and Profiling

The speed improve is the same as https://github.com/sgl-project/sglang/pull/9436 

## Checklist

- [X] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.ai/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.ai/developer_guide/contribution_guide.html#benchmark-the-speed).

## Summary by CodeRabbit

* **Refactor**
  * Improved batching logic to automatically choose between batch and per-request tokenization based on request content.
  * Optimized handling for batches without text input, reducing unnecessary work and lowering latency.
  * Centralized validation before batch processing for more predictable behavior with mixed or multimodal inputs.


narutolhy:
This is my local test, test the result between send request one by one and batch send.
```python
# -*- coding: utf-8 -*-
import random
import unittest
from types import SimpleNamespace

import torch
import sglang as sgl
from sglang.test.test_utils import DEFAULT_SMALL_MODEL_NAME_FOR_TEST

def _batch_has_only_input_ids_like(batch_size: int, obj) -> bool:
    for i in range(batch_size):
        if obj[i].input_ids is None:
            return False
    return True


class TestBatchHasOnlyInputIdsPredicate(unittest.TestCase):
    def test_all_have_input_ids_returns_true(self):
        obj = [
            SimpleNamespace(input_ids=[1, 2, 3]),
            SimpleNamespace(input_ids=[]),
            SimpleNamespace(input_ids=[42]),
        ]
        self.assertTrue(_batch_has_only_input_ids_like(len(obj), obj))

    def test_any_missing_input_ids_returns_false(self):
        obj = [
            SimpleNamespace(input_ids=[1]),
            SimpleNamespace(input_ids=None),
            SimpleNamespace(input_ids=[3, 4]),
        ]
        self.assertFalse(_batch_has_only_input_ids_like(len(obj), obj))

    def test_ignores_other_fields(self):
        obj = [
            SimpleNamespace(input_ids=[1, 2], text="hi"),
            SimpleNamespace(input_ids=[3], prompt="p"),
            SimpleNamespace(input_ids=[4], input_embeds="whatever"),
        ]
        self.assertTrue(_batch_has_only_input_ids_like(len(obj), obj))


TOP_K = 10
TEMPERATURE = 0.0
BATCH_SIZE = 6
MAX_NEW_TOKENS = 3


class TestGenerateWithInputIdsOnlyPath(unittest.TestCase):
    @classmethod
    def setUpClass(cls):
        cls.model_name = DEFAULT_SMALL_MODEL_NAME_FOR_TEST
        cls.engine = sgl.Engine(
            model_path=cls.model_name,
            random_seed=1234,
            mem_fraction_static=0.85,
        )
        cls.tok = AutoTokenizer.from_pretrained(
            cls.model_name, trust_remote_code=True
        )

    @classmethod
    def tearDownClass(cls):
        cls.engine.shutdown()
        if torch.cuda.is_available():
            torch.cuda.empty_cache()

    def _make_batch(self):
        prompts = [
            "Hello",
            "Hello world",
            "The quick brown fox",
            "To be or not to be",
            "SGLang batching with ids",
            "input_ids only request",
        ]
        random.seed(42)
        random.shuffle(prompts)
        prompts = prompts[:BATCH_SIZE]
        input_ids = [self.tok.encode(p, add_special_tokens=False) for p in prompts]
        input_ids = [ids if ids else [self.tok.eos_token_id or 0] for ids in input_ids]
        return_logprob = [(i % 2 == 0) for i in range(len(input_ids))]
        sampling_params = [
            {"temperature": TEMPERATURE, "top_p": 1.0, "top_k": TOP_K, "max_new_tokens": MAX_NEW_TOKENS}
            for _ in input_ids
        ]
        return input_ids, return_logprob, sampling_params

    def test_generate_ids_only_basic(self):
        input_ids, return_logprob, sampling_params = self._make_batch()
        outputs = self.engine.generate(
            input_ids=input_ids,
            return_logprob=return_logprob,
            top_logprobs_num=TOP_K,
        )
        self.assertEqual(len(outputs), len(input_ids))
        for i, out in enumerate(outputs):
            self.assertIn("text", out)
            meta = out.get("meta_info") or {}
            has_lp = bool(meta.get("input_top_logprobs"))
            if return_logprob[i]:
                self.assertTrue(has_lp, f"sample {i} expected input_top_logprobs")
            else:
                self.assertFalse(has_lp, f"sample {i} should NOT return input_top_logprobs")

    def test_generate_ids_only_batch_vs_single(self):
        input_ids, return_logprob, sampling_params = self._make_batch()
        MAX_ALLOWED_TOKEN_LEN_DELTA = 0
        MAX_ALLOWED_TEXT_MISMATCH = 0
        LOGPROB_MAX_DIFF = 1e-3
        LOGPROB_MEAN_DIFF = 1e-4
        outputs_batched = self.engine.generate(
            input_ids=input_ids,
            return_logprob=return_logprob,
            top_logprobs_num=TOP_K,
        )
        self.assertEqual(len(outputs_batched), len(input_ids))
        outputs_single = []
        for ids, sp, need_lp in zip(input_ids, sampling_params, return_logprob):
            out = self.engine.generate(
                input_ids=[ids],
                sampling_params=[sp],
                return_logprob=[need_lp],
                top_logprobs_num=TOP_K,
            )
            self.assertEqual(len(out), 1)
            outputs_single.append(out[0])

        def _get_meta(o):
            return o.get("meta_info") or {}

        num_mismatched_text = 0
        all_lp_max, all_lp_mean = [], []
        for i, (bo, so) in enumerate(zip(outputs_batched, outputs_single)):
            bmeta, smeta = _get_meta(bo), _get_meta(so)
            b_ids = bmeta.get("output_token_ids") or []
            s_ids = smeta.get("output_token_ids") or []
            self.assertLessEqual(
                abs(len(b_ids) - len(s_ids)),
                MAX_ALLOWED_TOKEN_LEN_DELTA,
                f"sample {i}: output length mismatch, batched={len(b_ids)}, single={len(s_ids)}",
            )
            if bo.get("text") != so.get("text"):
                num_mismatched_text += 1
            b_tok_lp = bmeta.get("output_token_ids_logprobs")
            s_tok_lp = smeta.get("output_token_ids_logprobs")
            if b_tok_lp and s_tok_lp and len(b_tok_lp) == len(s_tok_lp):
                diffs = [abs(b - s) for b, s in zip(b_tok_lp, s_tok_lp)]
                if diffs:
                    all_lp_max.append(max(diffs))
                    all_lp_mean.append(sum(diffs) / len(diffs))
        self.assertEqual(
            num_mismatched_text,
            MAX_ALLOWED_TEXT_MISMATCH,
            f"{num_mismatched_text} samples had different texts between batched and single runs.",
        )
        if all_lp_max:
            self.assertLessEqual(
                max(all_lp_max), LOGPROB_MAX_DIFF,
                f"max logprob diff {max(all_lp_max):.3e} > {LOGPROB_MAX_DIFF}"
            )
            mean_of_means = sum(all_lp_mean) / len(all_lp_mean)
            self.assertLessEqual(
                mean_of_means, LOGPROB_MEAN_DIFF,
                f"mean logprob diff {mean_of_means:.3e} > {LOGPROB_MEAN_DIFF}"
            )


if __name__ == "__main__":
    unittest.main()

**Comment by narutolhy** [comment]
Hi @hebiao064 please take a look, thanks ⸜( ⌓̈ )⸝

File: python/sglang/srt/managers/tokenizer_manager.py [hunk 1]
```diff
@@ -820,6 +820,15 @@ def _validate_batch_tokenization_constraints(
...
                 )
 
+    def _batch_has_only_input_ids(
+        self, batch_size: int, obj: Union[GenerateReqInput, EmbeddingReqInput]
+    ) -> bool:
+        """Check if all requests in the batch provide input_ids only."""
+        for i in range(batch_size):
+            if obj[i].input_ids is None:
+                return False
+        return True
```
JustinTong0323: The validation seems not shows "only"?

narutolhy: My bad, I'm lazy, I'll add some other exclusions.

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -954,11 +970,21 @@ async def _handle_batch_request(
...
+                    self._validate_batch_tokenization_constraints(batch_size, obj)
+                    tokenized_objs = await self._batch_tokenize_and_process(
+                        batch_size, obj
+                    )
+                else:
+                    tokenized_objs = []
+                    for i in range(batch_size):
+                        tmp_obj = obj[i]
+                        tokenized_obj = await self._tokenize_one_request(tmp_obj)
+                        tokenized_objs.append(tokenized_obj)
```
JustinTong0323: The logic here is confused, shall be refactored..

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -771,6 +771,16 @@ async def _batch_tokenize_and_process(
         """Handle batch tokenization for text inputs only."""
         logger.debug(f"Starting batch tokenization for {batch_size} text requests")
 
+        if self._batch_has_only_input_ids(batch_size, obj):
+            # All requests already have input_ids, no need to tokenize
+            tokenized_objs = []
```
sundar24295s: Maybe we can make this cleaner
```
if self._batch_has_only_input_ids(batch_size, obj):
    # All requests already have input_ids, no need to tokenize
    return [await self._tokenize_one_request(obj[i]) for i in range(batch_size)]
```

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -771,6 +771,16 @@ async def _batch_tokenize_and_process(
         """Handle batch tokenization for text inputs only."""
         logger.debug(f"Starting batch tokenization for {batch_size} text requests")
 
+        if self._batch_has_only_input_ids(batch_size, obj):
```
sundar24295s: - Can we extend this for `input_embeds` as well similar to input_ids?
- Essentially, we can send a single zmq socket for text. ids and embeds for batch sequences sent by client.

python/sglang/srt/managers/tokenizer_manager.py
narutolhy: OK, I'll edit it, thanks.

python/sglang/srt/managers/tokenizer_manager.py
narutolhy: Got it, I'll add it.

hnyls2002 (approved):
LGTM

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -938,13 +970,8 @@ async def _handle_batch_request(
         generators = []
         rids = []
         if getattr(obj, "parallel_sample_num", 1) == 1:
-            if self.server_args.enable_tokenizer_batch_encode:
-                # Validate batch tokenization constraints
-                self._validate_batch_tokenization_constraints(batch_size, obj)
-
+            if self._should_use_batch_tokenization(batch_size, obj):
```
sundar24295s, line 978: - Idea is to batch send in one zmq socket for all text, token ids or embeddings as input, right?
- If that's the case, we can keep the check here as 
 `if self.server_args.enable_tokenizer_batch_encode:` and remove `_should_use_batch_tokenization` function completely.

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -755,6 +755,12 @@ async def _batch_tokenize_and_process(
         """Handle batch tokenization for text inputs only."""
         logger.debug(f"Starting batch tokenization for {batch_size} text requests")
 
+        if self._batch_has_only_tokenized_request(batch_size, obj):
```
sundar24295s: And over here we want would recommend writing like:

```
# If batch does not have text nothing to tokenize
# so lets construct the return object
if not self.batch_has_text(batch_size, obj):
   return [await self._tokenize_one_request(obj[i]) for i in range(batch_size)]

```

python/sglang/srt/managers/tokenizer_manager.py
narutolhy: `_should_use_batch_tokenization ` contains self.server_args.enable_tokenizer_batch_encode, Justin said in this way it is easier to understand. Do you think it is better to list all the conditions here or encapsulate them like this?
```python 
 def _should_use_batch_tokenization(self, batch_size, requests) -> bool:
        """Return True if we should run the tokenizer in batch mode.

        Current policy:
        - Respect explicit server flag `enable_tokenizer_batch_encode`.
        - Or, if every request only carries `input_ids` (no raw text/embeds), we still batch.
        """
        return (
            self.server_args.enable_tokenizer_batch_encode
            or self._batch_has_only_tokenized_request(batch_size, requests)
        )

narutolhy:
Hi @sundar24295s Do you have any other suggestions? Thanks

python/sglang/srt/managers/tokenizer_manager.py
sundar24295s: @narutolhy  I might be misunderstanding the intent of the PR — are you looking to change the default behavior for `non-text` inputs even when `server_args.enable_tokenizer_batch_encode` is set to false, or only when it’s true?

narutolhy: @sundar24295s Yes, I looking to change the default behavior for non-text inputs even when server_args.enable_tokenizer_batch_encode is set to false. My scenario is to input the token_id that has been tokenized, and I hope to batch send it at this time.

PR #10678, opened 2025-09-20
[2/2] Support deterministic inference for temperature > 0

Qiaolin-Yu:
## Motivation

Part 1 and background: #10645
### Launch Commands
```
python3 -m sglang.launch_server --model-path Qwen/Qwen3-8B --attention-backend flashinfer --enable-deterministic-inference
```

### 
Test Deterministic
```
python3 -m sglang.test.test_deterministic --test-mode single --temperature 0.7
python3 -m sglang.test.test_deterministic --test-mode mixed --temperature 0.7
python3 -m sglang.test.test_deterministic --test-mode prefix --temperature 0.7
```

Results
```
# Simple
Total samples: 50, Unique samples: 1

# Mixed
Prompt 1: total samples: 517, Unique samples: 1
Prompt 2: total samples: 492, Unique samples: 1
Long prompt: total samples: 266, Unique samples: 1

# Prefix
Prompt 0 with prefix length 1: total samples: 344, Unique samples: 1
Prompt 1 with prefix length 511: total samples: 308, Unique samples: 1
Prompt 2 with prefix length 2048: total samples: 295, Unique samples: 1
Prompt 3 with prefix length 4097: total samples: 328, Unique samples: 1
```

## Checklist

- [ ] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.ai/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.ai/developer_guide/contribution_guide.html#benchmark-the-speed).

PR #11095, opened 2025-09-30
convert test_deterministic into unit tests

skyzh:
## Motivation

Part of https://github.com/sgl-project/sglang/issues/10278

## Modifications

The patch converts the deterministic test into unit tests so that it's easier to detect issues in CI and avoid regression.

- add `n_start` parameter so that we can start with 10+ size batch, saving some time
- avoid supplying sampling seed as None as it would cause sglang server to stop (a bug to be fixed in the server?)
- add return value for test_deterministic: an array of unique responses for each prompt

## Checklist

- [ ] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.ai/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.ai/developer_guide/contribution_guide.html#benchmark-the-speed).

PR #11096, opened 2025-09-30
fix sampling_seed handling when deterministic is enabled

skyzh:
## Motivation

After https://github.com/sgl-project/sglang/pull/10678 lands the requests without sampling seed set would not work anymore with pytorch complaining:

```
[2025-09-30 05:01:31] Scheduler hit an exception: Traceback (most recent call last):
  File "/sgl-workspace/sglang/python/sglang/srt/managers/scheduler.py", line 2858, in run_scheduler_process
  File "/sgl-workspace/sglang/.venv/lib/python3.12/site-packages/torch/utils/_contextlib.py", line 120, in decorate_context
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/sgl-workspace/sglang/python/sglang/srt/managers/scheduler.py", line 921, in event_loop_overlap
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/sgl-workspace/sglang/python/sglang/srt/managers/scheduler.py", line 1755, in get_next_batch_to_run
    new_batch = self.get_new_batch_prefill()
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/sgl-workspace/sglang/python/sglang/srt/managers/scheduler.py", line 1939, in get_new_batch_prefill
    new_batch.prepare_for_extend()
  File "/sgl-workspace/sglang/python/sglang/srt/managers/schedule_batch.py", line 1413, in prepare_for_extend
    self.sampling_info = SamplingBatchInfo.from_schedule_batch(
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/sgl-workspace/sglang/python/sglang/srt/sampling/sampling_batch_info.py", line 100, in from_schedule_batch
    torch.tensor(
TypeError: 'NoneType' object cannot be interpreted as an integer
```

## Modifications

If sampling seed is not specified, use `DEFAULT_SAMPLING_SEED` that is 42.

## Checklist


Fridge003:
cc @Qiaolin-Yu

Qiaolin-Yu:
What's your launch command?

skyzh:
```
python3 -m sglang.launch_server --model-path Qwen/Qwen3-0.6B --attention-backend flashinfer --enable-deterministic-inference
```

skyzh:
(without setting temperature - the request by default has it set to `null` in JSON)

skyzh:
sorry, should post client cmd instead:
```
```

Qiaolin-Yu (dismissed):
I think the issue is from the test script. We should change this line https://github.com/sgl-project/sglang/blob/main/python/sglang/test/test_deterministic.py#L32 instead of changing the code.

Fridge003:
@skyzh Is this fix is covered in #11095 ?

skyzh:
yes - while test_deterministic shouldn't send `temperature`: null at the first place, I don't think it's a good idea to crash the server if `null` is provided wrongfully by the user. Qiaolin suggested https://github.com/sgl-project/sglang/pull/11076/files#diff-cf90e92cc4eded1a3d9c01e6f7d0c6abcfec835b427a4c86a9f77f9d426fc89e is the root cause and we should fix things in the server.

skyzh:
This patch now fixes both `test_deterministic` and the server. I suspect passing `null` to other parameters (like top_k, etc) would cause similar issues. I would revisit it later and see if we can get a full fix for everything like this.

(I wish we have something like Rust's serde 🤣)

skyzh:
cc @Qiaolin-Yu can you take a look again when you have time, thanks!

Fridge003 (approved):
This fix LGTM, how do you think @Qiaolin-Yu

python/sglang/srt/sampling/sampling_batch_info.py
```diff
@@ -98,7 +98,14 @@ def from_schedule_batch(cls, batch: ScheduleBatch, vocab_size: int):
         )
         sampling_seed = (
             torch.tensor(
-                [r.sampling_params.sampling_seed for r in reqs],
+                [
```
Qiaolin-Yu: I think we shouldn't change it here. We should put the default value in the init function of `sampling_params`

skyzh: i wonder if it really works because we have/the user might supply the "sampling_seed" field in JSON, so when dataclass deserializes it, it will put `None` here anyways? (not entirely sure, will try)

skyzh:
```
        if self.preferred_sampling_params:
            sampling_kwargs = {**self.preferred_sampling_params, **obj.sampling_params}
        else:
            sampling_kwargs = obj.sampling_params
        sampling_params = SamplingParams(**sampling_kwargs)
```

yeah there's a kwargs unpack.. that's why `None` gets passed into the init function in the first place

skyzh: https://github.com/sgl-project/sglang/pull/11096/commits/efb3fe6e3c16bf227e41699f52f50023eb59db79

skyzh:
well, it seems like we don't do any sanitizing when parsing the request: setting that field to string would also crash the server; I can do a refactor around request processing at some point to get a full fix. For now let's get this merged so that at least the test can run.

Qiaolin-Yu:
> well, it seems like we don't do any sanitizing when parsing the request: setting that field to string would also crash the server; I can do a refactor around request processing at some point to get a full fix. For now let's get this merged so that at least the test can run.

If the user passes in None, I think it's illegal? We could raise ValueError here: https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/sampling/sampling_params.py#L96

python/sglang/srt/sampling/sampling_batch_info.py
```diff
@@ -9,7 +9,7 @@
 
 import sglang.srt.sampling.penaltylib as penaltylib
 from sglang.srt.sampling.custom_logit_processor import CustomLogitProcessor
-from sglang.srt.sampling.sampling_params import TOP_K_ALL
+from sglang.srt.sampling.sampling_params import DEFAULT_SAMPLING_SEED, TOP_K_ALL
```
hebiao064: do we need to import DEFAULT_SAMPLING_SEED?

hebiao064:
i think we can simply fix the test.

If user doesn't pass sampling_seed, then it will be 42

Qiaolin-Yu:
> This patch now fixes both `test_deterministic` and the server. I suspect passing `null` to other parameters (like top_k, etc) would cause similar issues. I would revisit it later and see if we can get a full fix for everything like this.
> 
> (I wish we have something like Rust's serde 🤣)

I agree with this. I think we should just fix the test and throw ValueError if passing none @skyzh

Qiaolin-Yu:
> yes - while test_deterministic shouldn't send `temperature`: null at the first place, I don't think it's a good idea to crash the server if `null` is provided wrongfully by the user. Qiaolin suggested https://github.com/sgl-project/sglang/pull/11076/files#diff-cf90e92cc4eded1a3d9c01e6f7d0c6abcfec835b427a4c86a9f77f9d426fc89e is the root cause and we should fix things in the server.

Ohh I mean we should fix the test file itself. Sorry for the confusion. I think for server, throw ValueError when passing none is enough for this pr. We can discuss later for further error handling.

python/sglang/srt/sampling/sampling_batch_info.py
merrymercy: this is not needed

python/sglang/srt/sampling/sampling_params.py
```diff
@@ -55,7 +56,7 @@ def __init__(
         custom_params: Optional[Dict[str, Any]] = None,
         stream_interval: Optional[int] = None,
         logit_bias: Optional[Dict[str, float]] = None,
-        sampling_seed: int = 42,
+        sampling_seed: Optional[int] = DEFAULT_SAMPLING_SEED,
```
merrymercy: this change seems not needed. We can just use 42 as the default sampling seed and reject 'None' as input

skyzh:
thanks for the reviews :) Now the patch only contains changes to the test itself and some minimal checks in `sampling_params.py`

skyzh:
cc @Qiaolin-Yu if you have time could you please look again, thanks

Qiaolin-Yu (approved):
LGTM

hnyls2002 (approved):
LGTM. Just curious why we cannot set the default sampling_seed in `BenchArgs` to 42? @skyzh

PR #11106, opened 2025-09-30
[router] add logprob in grpc generate endpoint wip

slin1237:
## Checklist

### DIFF
--- a/sgl-router/src/grpc_client/sglang_scheduler.rs
+++ b/sgl-router/src/grpc_client/sglang_scheduler.rs
@@ -127,6 +127,16 @@ impl SglangSchedulerClient {
         let sampling_params =
             Self::build_sampling_params_from_plain(body.sampling_params.as_ref())?;
 
+        // logprob_start_len: position to start returning logprobs
+        // -1 = return all logprobs (both input and output)
+        // 0 = return from beginning
+        // N = skip first N tokens
+        let logprob_start = if body.return_logprob {
+            body.logprob_start_len.unwrap_or(0)
+        } else {
+            0
+        };
+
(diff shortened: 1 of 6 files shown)


sgl-router/src/protocols/spec.rs
```diff
@@ -1961,6 +1961,14 @@ pub struct GenerateRequest {
     #[serde(default)]
     pub return_logprob: bool,
 
+    /// Starting position for returning logprobs (0 = from beginning)
```
CatherineSue, line 1964: Please double check. -1 means from the beginning. 0 means the second.

sgl-router/src/grpc_client/sglang_scheduler.rs
```diff
@@ -127,6 +127,15 @@ impl SglangSchedulerClient {
         let sampling_params =
             Self::build_sampling_params_from_plain(body.sampling_params.as_ref())?;
 
+        // logprob_start_len: position to start returning logprobs
+        // -1 = return all logprobs (both input and output) - same as chat
+        // Use -1 to match chat completion behavior which works correctly
+        let logprob_start = if body.return_logprob {
```
CatherineSue, line 133: I don't think this is right.

See sampling_params.md

```
| logprob_start_len          | `Optional[Union[List[int], int]] = None`                                     | If return_logprob, the start location in the prompt for returning logprobs. Default is "-1", which returns logprobs for output tokens only. 
````

CatherineSue: In chat completions is `-1` because that will give `output` only. 

In completions if `request.echo` then it should be `0`

PR #11108, opened 2025-09-30
Add custom request-level attributes for logging

scottjlee:
## Motivation

When SGLang is deployed in production, it is often serving numerous requests from various users. For example, to better identify which logs belong to which users' requests, it can be useful to associate request logs with their corresponding user ID.

## Modifications
This feature can be enabled with the server launch arg `--log-requests-custom-attributes`. We add the option to pass request-level attributes in the request header field `x-request-custom-attributes` (header also configurable via `--log-requests-custom-attributes-header` flag), which is a JSON-format string containing key-value pairs of attributes associated with that request. These attributes are included in the [request level logging output](https://github.com/sgl-project/sglang/blob/87a92e459a63b33259b9609031f5d23fe3ee5c65/python/sglang/srt/managers/tokenizer_manager.py#L919-L921), so that requests can be easily identified and filtered by the attributes.

Sample request (see `x-request-custom-attributes` header field):
```
curl -s http://localhost:8001/v1/completions    
-H "Content-Type: application/json"     
-H 'x-request-custom-attributes: {"user_id": "scottjlee-dev", "external_request_id": "oBTZYgU-4YNCb4-980033869e849bc1"}'      
-d '{
  "model": "meta-llama/Llama-3.1-8B-Instruct",
  "prompt":  "Which European country has the largest land area? Answer in a single sentence.",
  "max_tokens": 100,
  "stream": true
  }'
```

Logs after this PR (now includes `external_request_id`, `user_id` fields extracted from header):
```
Finish: obj=GenerateReqInput(rid='7bbd93c6b6e34891a6e2f4589aad4d9d', http_worker_ipc=None, video_data=None, return_logprob=False, logprob_start_len=-1, top_logprobs_num=0, token_ids_logprob=None, return_text_in_logprobs=True, stream=True, log_metrics=True, return_hidden_states=False, modalities=None, session_params=None, lora_id=None, custom_logit_processor=None, bootstrap_host=None, bootstrap_port=None, bootstrap_room=None, bootstrap_pair_key=None, data_parallel_rank=None, background=False, conversation_id=None, priority=None, extra_key=None, no_logs=False, custom_labels=None, custom_request_attributes={'user_id': 'scottjlee-dev', 'external_request_id': 'oBTZYgU-4YNCb4-980033869e849bc1'}, return_bytes=False, return_entropy=False), out={'meta_info': {'id': '7bbd93c6b6e34891a6e2f4589aad4d9d', 'finish_reason': {'type': 'length', 'length': 100}, 'prompt_tokens': 16, 'weight_version': 'default', 'completion_tokens': 100, 'cached_tokens': 15, 'e2e_latency': 0.6665134429931641}, 'user_id': 'scottjlee-dev', 'external_request_id': 'oBTZYgU-4YNCb4-980033869e849bc1'}
```

## Accuracy Tests

Should not impact model outputs or accuracy, it just modifies request/output fields.

## Checklist


acelyc111:
It should be careful when changing the protocol, could you please take a look at https://github.com/sgl-project/sglang/pull/10143 to see if it can be reused?

scottjlee:
Thanks @acelyc111 , I took a look at the `custom_labels` feature, and I found that for this use case, we cannot directly use this due to high cardinality of labels. I first tried to re-use this parameter and add an extra toggleable feature which outputs these labels in the finished logging. However, the main problem with this approach is that for our intended use case, we would pass high cardinality fields such as request ID, customer ID, and user ID. If these high cardinality fields are also used in the metrics, this would cause slowdowns as more and more unique label values are added. Therefore, we'd like to also have a separate path to specify request details which should NOT end up as part of metrics.

I took inspiration from your implementation for the custom labels, and instead re-wrote my PR to use a similar approach of extracting custom request attributes from the header, and include it in the finished logging. This avoids directly modifying the protocol, but I'll still need to pass the new `log_requests_custom_attributes` as a field in the `GenerateReqInput`, `TokenizedGenerateReqInput`, and `EmbeddingReqInput` classes, similar to how `custom_labels` is passed. What do you think?

ShangmingCai (commented):
Not very sure about this feature, IMP, `user_id` or `external_request_id` means nothing to the engine, it might be helpful for debugging, but we can use `rid`, which can be assigned a unique value, to debug as well. Also, if these attributes are needed, I think it would be better to implement them at the sgl-router side. cc: @slin1237

scottjlee:
Heads up, I recently  updated the PR so it accepted an arbitrary set of labels via the header, not just request or user ID. Hopefully by generalizing the feature more, it can cover more use cases (e.g. experiment ID, routing info, other metadata, etc)

```
curl -s http://localhost:8001/v1/completions    
-H "Content-Type: application/json"     
-H 'x-request-custom-attributes: {"user_id": "scottjlee-dev", "external_request_id": "oBTZYgU-4YNCb4-980033869e849bc1", "blahblah": "foo"}'      
-d '{
  "model": "meta-llama/Llama-3.1-8B-Instruct",
  "prompt":  "Which European country has the largest land area? Answer in a single sentence.",
  "max_tokens": 100,
  "stream": true
  }'
```

scottjlee:
@ShangmingCai @slin1237 thoughts on this PR? I generalized the feature so that it now supports arbitrary string attributes associated with the request, so that requests can be identified by arbitrary keys/attributes -- i think this feature is very useful in the context of ingesting resulting logs/metrics from production SGLang deployments (allows additional identifying features to be embedded in per-request logs). thanks!

PR #11116, opened 2025-09-30
enable cudaProfilerApi for one batch benchmarking

lpc0220:
Enable cudaProfiler for one batch benchmarking. 

fix the bug:
output_len / 2 could be float.


python/sglang/bench_one_batch.py
```diff
@@ -80,6 +85,24 @@
 )
 
 
+def start_cuda_profiler(rank_print):
+    """Start CUDA Profiler API for nsys profiling."""
```
fzyzcjy: nit: what about removing this comment since this is self explanatory

python/sglang/bench_one_batch.py
```diff
@@ -80,6 +85,24 @@
...
+    try:
+        torch.cuda.cudart().cudaProfilerStart()
+        rank_print("CUDA Profiler started (nsys will begin capturing)")
+    except Exception as e:
+        rank_print(f"Failed to start CUDA profiler: {e}")
+
+
+def stop_cuda_profiler(rank_print):
+    """Stop CUDA Profiler API and signal nsys to dump traces."""
```
fzyzcjy: same

python/sglang/bench_one_batch.py
```diff
@@ -318,6 +356,20 @@ def _read_prompts_from_file(prompt_file, rank_print):
         return pf.readlines()
 
 
+def _get_torch_profiler_output_dir():
+    """Get the torch profiler output directory from environment variable or default to current directory."""
+    return os.environ.get("SGLANG_TORCH_PROFILER_DIR", ".")
```
fzyzcjy: iirc default is `/tmp`

I personally suggest bench_one_batch to be in sync with semantics of bench_serving in handling dirs, i.e. explicitly require users to specify -> https://github.com/sgl-project/sglang/blob/43f80884c59a386402e8870defb5601b1ef4b971/python/sglang/bench_serving.py#L2410-L2411

python/sglang/bench_one_batch.py
```diff
@@ -417,6 +472,8 @@ def latency_test_run_once(
 
     profiler = None
     if profile:
+        torch_profiler_dir = _get_torch_profiler_output_dir()
+        rank_print(f"Torch profiler traces will be saved to: {torch_profiler_dir}")
         profiler = torch.profiler.profile(
```
fzyzcjy: shall we have a flag to disable torch profile, since if we have nsys we may not want one more torch profile

python/sglang/bench_one_batch.py
```diff
@@ -463,10 +531,28 @@ def latency_test_run_once(
             )
             profiler.start()
 
+        if (
+            cuda_profiler
+            and cuda_profiler_stage in ["all", "decode"]
+            and i == output_len // 2
```
fzyzcjy: nit: `is_profiler_interesting_step = outputlen//2`

fzyzcjy (commented):
some nits

python/sglang/bench_one_batch.py
lpc0220: Done.

python/sglang/bench_one_batch.py
lpc0220: Done.

python/sglang/bench_one_batch.py
lpc0220: I changed the default to /tmp.

This is a legacy issue: bench_one_batch has different behavior than bench_serving for SGLANG_TORCH_PROFILER_DIR. I can understand it in some way. bench_one_batch doesn't launch a server and doesn't have to rely on this env var to enable profiler. So i'd suggest we make it more flexible here? i.e. SGLANG_TORCH_PROFILER_DIR is not a must. If users set it, we take it; otherwise, we choose /tmp. WDYT?

python/sglang/bench_one_batch.py
lpc0220: torch profile and nsys profile use different flags. they are exclusive.
torch profiler uses --profile as before.
nsys profiler uses --cuda_profiler.

python/sglang/bench_one_batch.py
lpc0220: I don't get here. What's it mean?

fzyzcjy:
```
is_profiler_interesting_step = outputlen//2

if ... and is_profiler_interesting_step:
  start profile, etc

**Comment by fzyzcjy** [comment]
text below not only apply to this pr, but as a general remark

### 1.

goal: have a *unified* API for users who use profiler.py, bench_serving.py, bench_one_batch.py, bench_one_batch_sevrer.py, /start_profile http api

step 1: make a table to briefly check about their current status

step 2: design api
e.g. `--profile --profile-activities CUDA_PROFILER`?
b/c e.g. bench_one_batch_server has --profile-steps, --profile-by-stage, ... thus we follow it

### 2.

great to extract a profile API

```
class ProfileWrapper:
  def start_profile(activities, ...) -> Tuple[bool indicating success, str indicating message]:
    the core code in scheduler_profiler_mixin.py :: start_profile, decoupled from that mixin

  def stop_profile:
    same
```

then we can simply call that single function in both scheduler_profiler_mixin.py and bench_one_batch.py.
Instead of currently, we re-write again the profile wrapper logic in bench_one_batch. And our current profile logic is a bit coupled w/ scheduler.
we also have extra benefit to make scheduler code cleaner

more concretely

```
class SchedulerProfilerMixin:
    def stop_profile() -> ProfileReqOutput | None:
        call the ProfileWrapper

  handle profiler_target_prefill_ct etc in this class
```

### 3

I think maybe it is good if we follow

* no breaking change (unless very necessary)
* code very clean

**Comment by fzyzcjy** [comment]
minimal change
* torch profile and nsys profile move to sibling
* api make it like above, --profile --profile-activity blahblah
* maybe extract a function about start profile and another function about stop file

mv L548 to L575

image

File: python/sglang/bench_one_batch.py [hunk 6]
```diff
@@ -447,32 +553,37 @@ def latency_test_run_once(
...
 
-    if profile:
-        profiler.stop()
-        trace_filename = f"{profile_filename_prefix}_batch{batch_size}_input{input_len}_output{output_len}_prefill.trace.json.gz"
-        _save_profile_trace_results(profiler, trace_filename)
-        rank_print(f"torch profiler chrome trace for prefill saved to {trace_filename}")
-
-    # Decode
     decode_latencies = []
+    step_of_interest = output_len // 2
```
fzyzcjy: tiny nit


```suggestion
    profile_step_of_interest = output_len // 2
```

python/sglang/bench_one_batch.py
```diff
@@ -399,7 +493,10 @@ def latency_test_run_once(
     log_decode_step,
     profile,
     profile_record_shapes,
+    profiler_activities,
     profile_filename_prefix,
+    profile_stage,
+    tp_rank,
```
fzyzcjy, line 513: tiny nit: do we need this (or can we obtain from dist.get_rank())

fzyzcjy (approved):

python/sglang/bench_one_batch.py
lpc0220: passing a parameter will avoid dup calls to dist.get_rank(). Let's keep it for now.

PR #11118, opened 2025-09-30
Create tester.py

molocule:
## Motivation

allow others to run sglang tests in a reliable and reproducible environment with multiple hardware options
this will allow for faster iteration and improvements on sglang from the community

## Modifications

add a simple tester file for running sglang tests on modal 

## Accuracy Tests
verified by running memory tests


hnyls2002:
@molocule, I think this script is better kept locally.

PR #11119, opened 2025-09-30
Add timing metrics for requests

scottjlee:
## Motivation
To better understand where a request is bottlenecked, it would be helpful to have a breakdown of how long a request spends in major phases of inference (validation, queue, prefill, decode). In addition, having timestamps of key points of the request lifecycle (request received, request scheduled, response sent) can be helpful to analyze latencies between external callers.

## Modifications
Add the following timing metrics breaking down how long a request spends in major phases of inference:
- `request_received_ts`: timestamp when the request is received
- `request_scheduled_ts`: timestamp when the request is dispatched to the scheduler
- `queue_time`: duration a request spent waiting in the scheduler's queue before being selected for execution
- `prefill_delay`: time between when a request is taken off the queue and when prefill processing actually starts
- `prefill_latency`: actual time spent executing the prefill computation phase
- `prefill_finished_ts`: timestamp when prefill phase completes and first token is generated. This is approximately close to the timestamp of the first token / TTFT event.
- `response_sent_ts`: timestamp when the response starts being sent back to the client (first chunk for streaming, complete response for non-streaming)
- `decode_finished_ts`: Timestamp when the entire request processing is complete
- `inference_time`: total time spent in inference processing from start to finish (prefill and decode)

Also adds the following related calculated performance metrics:
- `decode_throughput`: tokens generated per second during the decode phase
- `time_per_token`: average time to generate each token during decode phase

These timing metrics are collected at a per-request level, and reported in the final log output (example below):
```
[2025-09-30 23:16:53] Finish: obj=GenerateReqInput(video_data=None, rid='8a768a217ef145da9f8fe266ded97622', 
extra_key=None, return_logprob=False, logprob_start_len=-1, top_logprobs_num=0, token_ids_logprob=None, 
return_text_in_logprobs=True, stream=True, log_metrics=True, return_hidden_states=False, modalities=[], 
session_params=None, lora_id=None, custom_logit_processor=None, bootstrap_host=None, bootstrap_port=None, 
bootstrap_room=None, bootstrap_pair_key=None, validation_time=3.562029451131821e-05, data_parallel_rank=None, 
background=False, conversation_id=None, label=None, priority=None, return_bytes=False, custom_labels=None), out=
{'meta_info': {'id': '8a768a217ef145da9f8fe266ded97622', 'finish_reason': {'type': 'stop', 'matched': 128009}, 
'prompt_tokens': 50, 'weight_version': 'default', 'queue_time': 0.0003788946196436882, 'prefill_delay': 
0.0011139055714011192, 'prefill_latency': 0.00019742920994758606, 'completion_tokens': 51, 'cached_tokens': 1, 
'e2e_latency': 0.3615589141845703, 'request_received_ts': 1759274212.849985, 'request_scheduled_ts': 
1759274212.8502948, 'prefill_finished_ts': 1759274212.88563, 'response_sent_ts': 1759274212.88575, 
'decode_finished_ts': 1759274213.2115438, 'inference_time': 0.360590404830873, 'decode_throughput': 
156.48294216283838, 'time_per_token': 0.006390472892306599}}
```

## Accuracy Tests
Should not affect model outputs or accuracy.

## Checklist


hnyls2002:
@scottjlee LGTM. I think we can refactor this by introducing a shared base class that contains the common fields like timing metrics, retraction count, and prefix cache hit rate. Then all the request and output types — such as BatchMultimodalDecodeReq, BatchTokenIDOutput, BatchStrOutput, etc. — can simply inherit from that base class instead of redefining those fields individually.

cc @zhyncs

scottjlee:
Thanks for the suggestion, i refactored the added timing metrics and speculative decoding metrics into `RequestTimingMetricsMixin` and `SpeculativeDecodingMetricsMixin` classes, respectively. Note that because the output type classes now inherit from these common metrics dataclasses, I've had to re-order the parameters (attributes imported from dataclasses must be ordered at the front). @hnyls2002 could you take another look? Thanks!

If it looks good, I will also update the other outstanding PRs accordingly to also follow these mixin conventions.

sufeng-buaa:
Hi, I'm currently working on a refactoring to unify request stage tracing, metrics, and logging across SGLang—this includes consolidating how timestamps are collected and reported. https://github.com/sgl-project/sglang/issues/10916. I noticed there might be potential conflicts with the timing-related changes in your PR.
﻿
I plan to submit my patch this week. Would you be open to taking a look once it's up, so we can discuss how to align our approaches? Currently, observability in SGLang doesn't have a dedicated owner, and as more people contribute improvements, things are becoming a bit fragmented. I think a coordinated effort to integrate these pieces cleanly would be really valuable.
﻿
By the way, I couldn’t find your account on Slack—would you be up for continuing the discussion there? It might be easier to sync up in real time.
﻿
Looking forward to collaborating!

PR #11121, opened 2025-10-01
[router] feat: validate router spec against OpenAI spec and enable validation

tao12345666333:
## Motivation

fixes: https://github.com/sgl-project/sglang/issues/10919

## Checklist

### DIFF
--- a/sgl-router/benches/request_processing.rs
+++ b/sgl-router/benches/request_processing.rs
@@ -39,6 +39,8 @@ fn default_generate_request() -> GenerateRequest {
         parameters: None,
         sampling_params: None,
         return_logprob: false,
+        logprob_start_len: None,
+        return_text_in_logprobs: false,
         // SGLang Extensions
         lora_path: None,
         session_params: None,
--- a/sgl-router/src/protocols/spec.rs
+++ b/sgl-router/src/protocols/spec.rs
@@ -416,10 +416,7 @@ impl ChatCompletionRequest {
     /// - "auto" is the default if tools are present
     pub fn normalize_tool_choice(&mut self) {
         if self.tool_choice.is_none() {
-            let has_tools = self
-                .tools
-                .as_ref()
-                .map_or(false, |tools| !tools.is_empty());
+            let has_tools = self.tools.as_ref().is_some_and(|tools| !tools.is_empty());
             self.tool_choice = Some(if has_tools {
                 ToolChoice::Value(ToolChoiceValue::Auto)
             } else {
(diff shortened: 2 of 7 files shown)


slin1237:
for those validations
can we instead use validator? instead of forcing every handler to call this
these impl should be special or common implementation such as batch_size etc
for validation, perhaps like this

```rust
use validator::{Validate, ValidationError};

fn company_email(v: &str) -> Result<(), ValidationError> {
    if v.ends_with("@mycorp.com") { Ok(()) } else { Err(ValidationError::new("company_email")) }
}

#[derive(Deserialize, Validate)]
struct Invite {
    #[validate(email, custom(function = "company_email"))]
    email: String,
}
```

given the size of the spec
this is not a trivial task, but super tedious
we can break it down to different sections and have different folks to pick up different sections

tao12345666333:
@slin1237 Thanks for the suggestion! I agree that validating at the data model boundary (via derive-based validation) is cleaner than making every handler call validation explicitly.

Plan:

- Introduce model-level validation and call it once at the HTTP/gRPC boundary, so handlers only see normalized, validated inputs.
- Move existing invariants (e.g., tools/tool_choice consistency) into struct-level checks; keep normalization as a separate step before validation.
- Add shared validators for common params (e.g., batch_size, sampling ranges, penalties) so they’re enforced consistently across endpoints.

Given the size of the OpenAI spec, I propose:

- Land this PR as-is to keep it scoped to the spec-alignment fixes, and

- Follow up with a dedicated PR that introduces the validator and incrementally covers sections (Chat Completions, Responses, Embeddings/common params). I can open a tracking issue and break it down so others can pick up different parts.

If you prefer, I can also add the minimal plumbing in this PR (add the dependency and implement struct-level validation for the `tools/tool_choice` invariant) and do the rest in follow-ups. Let me know which approach you’d like. Thanks!

slin1237:
hey, sorry for the late reply
this PR seems ok, we still like to do a larger scope for clean up, and by then this code will likely get deleted as well
we shall focus on that

tao12345666333:
No worries about the late reply—thanks for getting back to me!

Totally understand, makes sense to tackle the bigger cleanup first. If there's anything I can help with on that front (e.g., contributing to the larger refactor), just let me know.

Looking forward to seeing it progress!

I'll close this one, if you want.

CatherineSue:
@tao12345666333 We were targeting a release yesterday so I moved some of the changes to #11588 . We can close this PR now. Appreciate the contribution!

PR #11123, opened 2025-10-01
Fix metrics and request tracing (TimeStats)

merrymercy:


python/sglang/srt/managers/scheduler_output_processor_mixin.py
```diff
@@ -707,6 +707,7 @@ def stream_output_generation(
                 and self.tp_rank == 0
                 and self.server_args.enable_request_time_stats_logging
             ):
+                print(f"{req.finished_reason=}")
```
merrymercy, line 710: need to drop this

PR #11131, opened 2025-10-01
Restore torch defaults between sgl-kernel tests

benbarsdell:
## Motivation

Currently, if a test changes the torch default device/dtype and then fails before resetting it, it can cause later tests to fail due to the unexpected default. This PR automatically restores the default settings around each test function.

## Modifications

Adds a conftest.py with an autoload fixture that restores the torch default device/dtype.

## Checklist

cc @kushanam


Kangyan-Zhou:
@benbarsdell Could you please help attach an example failure without this change?

benbarsdell:
E.g., if you add an `assert(False)` after [this](https://github.com/sgl-project/sglang/blob/97cd38e58d4c33a6cac002cd832e7e91c5b86e88/sgl-kernel/tests/test_kvcacheio.py#L54) line, you get errors from other tests such as:
```
FAILED test_lightning_attention_decode.py::test_lightning_attention_decode[64-64-96-96-dtype2] - RuntimeError: expected scalar type Float but found Half
FAILED test_merge_state.py::test_merge_state[128-32-2048] - RuntimeError: Half did not match Float
FAILED test_sampling.py::test_top_k_top_p_sampling_from_probs_logits_top_k_first_alignment[0.5-100-111-989] - AssertionError: assert tensor(False, device='cuda:0')
...
FAILED test_sampling.py::test_top_k_renorm_probs[500-128256-989] - AssertionError: The values for attribute 'dtype' do not match: torch.float16 != torch.float32.
FAILED test_sparse_flash_attn.py::test_convert_vertical_slash_indexes_mergehead[True] - AssertionError: assert False
```

Fridge003 (approved):

PR #11146, opened 2025-10-01
Add per-request prefix cache hit rate

scottjlee:
## Motivation

To better understand the effectiveness of the prefix cache, report per-request prefix cache hit rate.

## Modifications
Adds the `cache_hit_rate` metric tracked at the request level. These metrics are reported in the final log output (example below):
```
[2025-10-01 21:25:35] Finish: obj=GenerateReqInput(video_data=None, rid='24fa397b7cab4418b3f6a5c903ae310f', 
return_logprob=False, logprob_start_len=-1, top_logprobs_num=0, token_ids_logprob=None, 
return_text_in_logprobs=True, stream=True, log_metrics=True, return_hidden_states=False, modalities=[], 
session_params=None, lora_id=None, custom_logit_processor=None, bootstrap_host=None, bootstrap_port=None, 
bootstrap_room=None, bootstrap_pair_key=None, data_parallel_rank=None, background=False, conversation_id=None, 
priority=None, extra_key=None, no_logs=False, custom_labels=None, label=None, return_bytes=False), out={'meta_info': 
{'id': '24fa397b7cab4418b3f6a5c903ae310f', 'finish_reason': {'type': 'stop', 'matched': 128009}, 'prompt_tokens': 50, 
'weight_version': 'default', 'completion_tokens': 97, 'cached_tokens': 1, 'cache_hit_rate': 0.02, 'e2e_latency': 
0.6632676124572754}}
```

Additionally for streaming requests, we report the prefix cache hit rate in the first token response, which may be useful to the caller (e.g. dynamically adjust methods based on live cache hit rate)

## Accuracy Tests
Should not affect model outputs or accuracy.

## Checklist


python/sglang/srt/entrypoints/openai/protocol.py
```diff
@@ -255,12 +255,15 @@ class CompletionResponseChoice(BaseModel):
     finish_reason: Optional[Literal["stop", "length", "content_filter", "abort"]] = None
     matched_stop: Union[None, int, str] = None
     hidden_states: Optional[object] = None
+    cache_hit_rate: Optional[float] = None
```
ispobock: Is it an OAI compatible field?

scottjlee: i was trying to pass cache hit rate as part of the streaming response chunks, but i think there's no straightforward way of doing that without modifying the OAI classes. i decided to remove that part since it breaks compatibility, and it can be calculated by the caller using the existing `UsageInfo` information.

ShangmingCai (commented):
I think `cached_tokens` and `completion_tokens` should be enough. If the user or the router/middleware wants to obtain this `cache_hit_rate`, I think it can be easily calculated outside SGLang, so we should minimize the number of attributes if this newly added one actually doesn't carry more info.

scottjlee:
Fair point. I think some of our colleagues felt that during some debugging processes, it was convenient to be able to directly read the prefix cache hit rate from logs. But since it can be easily calculated by simply dividing two numbers in the log, I'll close the PR.

Issue #10513, opened 2025-09-16
[Bug] sgl-router not pass `prompt_tokens_details` in pd disagg mode


gogongxt:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Describe the bug

sgl-router not pass `prompt_tokens_details` in pd disagg mode

### Reproduction

Enable ` --enable-cache-report` when start sglang server.
Use curl request:
```plaintext
curl -X POST "http://localhost:58888/v1/chat/completions" \
  -H "Content-Type: application/json" \
  -d '{"model":"Qwen/Qwen2-1.5B-Instruct","messages":[{"role":"user","content":"What is the capital of France?"}]}'
```
Return json doesn't have `prompt_tokens_details`.

### Environment

any

gogongxt:
`mini lb` also has the same issue. 
The main problem is that the prefill ask JSON will be deprecated if logprobs are not used and only prefill json has the right `prompt_tokens_details` value.

gogongxt:
@slin1237

PR #11153, opened 2025-10-02
Fix missing prompt_tokens_details in PD NIXL cache reports

ai-jz:
## Motivation
- An issue in PD with NIXL backend was identified with symptom of PD cache reporting omitted `prompt_tokens_details` even with `--enable-cache-report`. It's similar to the symptom as reported in sgl-project/sglang#10513, but is a different issue not related to the router.
- Root cause: the NIXL backend only transferred the first aux metadata shard, leaving the decode worker's cached token counter at zero; this mirrors the issue captured in `results/diagnostics/pd_prompt_tokens_details_diag.md` and the router payload dump.

## Modifications
- Fix: iterate over every aux buffer in `NixlKVManager.send_aux`, mirroring mooncake's [implementation](https://github.com/sgl-project/sglang/blob/b0d25e72c401f37b55d689ddbf05b8c583afe854/python/sglang/srt/disaggregation/mooncake/conn.py#L600), and return all transfer handles so PD metadata (cached/prompt token breakdown) reaches the decode worker and router.

## Accuracy Tests

- Not applicable; changes do not touch model numerics.

## Benchmarking and Profiling

- Not applicable; the added test is hermetic and does not affect runtime behavior.

## Checklist

## Testing
- Relaunched prefill/decode workers with NIXL backend inside `lmsysorg/sglang:v0.5.2-cu126` with the PR:
  ```bash
  CUDA_VISIBLE_DEVICES=0 python -m sglang.launch_server \
    --model-path Qwen/Qwen2.5-Coder-7B-Instruct \
    --disaggregation-transfer-backend nixl \
    --enable-cache-report --log-requests --log-requests-level 3

  CUDA_VISIBLE_DEVICES=1 python -m sglang.launch_server \
    --model-path Qwen/Qwen2.5-Coder-7B-Instruct \
    --disaggregation-transfer-backend nixl \
    --enable-cache-report --log-requests --log-requests-level 3
  ```
- Confirmed the bug reproduces without the router by replaying the direct PD request script in `results/diagnostics/pd_prompt_tokens_details_diag.md` (decode `meta_info` lacked `prompt_tokens_details` before this patch and now includes the field).
- Started the router with PD handoff and sent the same chat request via `/v1/chat/completions`; after the patch the OpenAI-compatible response returns a populated `prompt_tokens_details` map instead of `null`.

### DIFF
--- a/test/srt/disaggregation/test_nixl_aux_transfer.py
+++ b/test/srt/disaggregation/test_nixl_aux_transfer.py
@@ -1,0 +1,96 @@
+import unittest
+from types import SimpleNamespace
+
+from sglang.srt.disaggregation.nixl.conn import NixlKVManager
+
+
+class FakeNixlAgent:
+    def __init__(self):
+        self.get_xfer_descs_calls = []
+        self.initialize_calls = []
+        self.transfer_calls = []
+        self._handle_counter = 0
+
+    def register_memory(self, *_args, **_kwargs):  # pragma: no cover - not used in test
+        return []
+
+    def get_xfer_descs(self, addrs, region):
+        self.get_xfer_descs_calls.append((tuple(addrs), region))
+        # Return descriptors mirroring inputs for assertions
+        return [(addr, length, region) for (addr, length, _gpu) in addrs]
+
+    def initialize_xfer(self, kind, src_descs, dst_descs, peer_name, notif):
+        self._handle_counter += 1
+        handle = f"handle-{self._handle_counter}"
+        self.initialize_calls.append(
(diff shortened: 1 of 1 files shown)


ishandhanani (approved):

ai-jz:
@ByronHsu @hnyls2002 Would like to have a look?

ai-jz:
It appears the same issue has been fixed by a later PR (https://github.com/sgl-project/sglang/pull/11375), which has been merged.

This PR is no longer of need.

PR #11170, opened 2025-10-02
Remove dp balance metadata and minimul token balance.

hnyls2002:
Deprecate this method.


ishandhanani:
Just curious on why @hnyls2002?

hnyls2002:
@ishandhanani The implementation is wrong (about the token updating, it is always delayed for 40 iters, and then the true tokens are changed) https://github.com/sgl-project/sglang/pull/11170/files#diff-9cdb34668cae0ef7eefead6ef788b5ab15c2bf954a84fa741bf0d05aabea8684L334

Also, we will introduce a timely and reliable method to report load (queuing requests, occupied tokens, KV cache usage, etc.). I will attach an issue later.

hnyls2002:
@ishandhanani here #11186

PR #11174, opened 2025-10-02
[Fix AMD CI] VRAM cleanup 

sunxxuns:
This PR addresses the issue where processes die but VRAM remains allocated, causing failures in subsequent AMD CI jobs.

## Changes Made

1. **Enhanced **:
   - Added VRAM checking function for both ROCm and NVIDIA systems
   - Added aggressive cleanup when VRAM is still allocated after killing SGLang processes
   - For ROCm: Kills all processes using  device

2. **Added **:
   - Standalone script to verify VRAM is clear
   - Exits with error code if VRAM is allocated
   - Used for verification in CI

3. **Updated AMD CI workflow** ():
   - Added 'Clean up GPU memory' step before installing dependencies
   - Added 'Verify VRAM is clear' step to ensure cleanup worked
   - Applied to all AMD CI jobs

## How it works

- Before each test job starts, the workflow now runs cleanup to kill any lingering SGLang processes
- Checks if VRAM is still allocated and performs additional cleanup if needed
- Verifies VRAM is clear before proceeding with dependency installation and tests
- This should prevent the 'process died but VRAM held' issue from affecting subsequent jobs

Fixes the issue seen in: https://github.com/sgl-project/sglang/actions/runs/18187939355/job/51776255653

### DIFF
--- a/test/srt/run_suite.py
+++ b/test/srt/run_suite.py
@@ -2,6 +2,7 @@ import argparse
 import glob
 from dataclasses import dataclass
 
+from sglang.srt.utils import is_hip
 from sglang.test.test_utils import run_unittest_files
 
 
@@ -368,7 +369,7 @@ if __name__ == "__main__":
     arg_parser.add_argument(
         "--timeout-per-file",
         type=int,
-        default=1200,
+        default=1500 if is_hip() else 1200,
         help="The time limit for running one file in seconds.",
     )
     arg_parser.add_argument(
--- a/.github/workflows/pr-test-amd.yml
+++ b/.github/workflows/pr-test-amd.yml
@@ -41,11 +41,8 @@ jobs:
         env:
           GITHUB_WORKSPACE: ${{ github.workspace }}
 
-      - name: Clean up GPU memory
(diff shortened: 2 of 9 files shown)


Kangyan-Zhou:
Can you check why in https://github.com/sgl-project/sglang/actions/runs/18208865074/job/51849604063?pr=11174 the script fails to clean up one GPU?

sunxxuns:
> Can you check why in https://github.com/sgl-project/sglang/actions/runs/18208865074/job/51849604063?pr=11174 the script fails to clean up one GPU?
this can be repro'ed, and exposed by this pr;  
we will update image to fix this flaky issue.

merrymercy (approved):

test/srt/run_suite.py
```diff
@@ -368,7 +369,7 @@ def auto_partition(files, rank, size):
     arg_parser.add_argument(
         "--timeout-per-file",
         type=int,
-        default=1200,
+        default=1500 if is_hip() else 1200,
```
merrymercy, line 372: we should not increase this timeout, instead, we should split the files and make each of them run faster.

PR #11178, opened 2025-10-02
fix self.enable_kv_cache_events

narutolhy:
## Motivation

`server_args.kv_events_config` defaults to None.
The original assignment:
`self.enable_kv_cache_events = server_args.kv_events_config and tp_rank == 0`
propagates None into `self.enable_kv_cache_events` when `kv_events_config` is None, because Python’s and returns the first falsy operand (None). This leads downstream to an assertion failure in `radix_cache_cpp.py:`
```python
assert (enable_kv_cache_events is False), "HiRadixCache does not support kv cache events yet"
```
Since `enable_kv_cache_events` was None (not False), the assertion triggers even though events were effectively “disabled”.

## Modifications
Make the intent explicit: only enable events when a config is provided and we’re on tp_rank == 0; otherwise set a strict boolean False.
```python
# Before
self.enable_kv_cache_events = server_args.kv_events_config and tp_rank == 0

# After
self.enable_kv_cache_events = (
    server_args.kv_events_config is not None and tp_rank == 0
)
```

## Accuracy Tests

No model logic changed; this is a control-flow/flag fix.
I print `self.enable_kv_cache_events ` before and after
before `self.enable_kv_cache_events`: None
after `self.enable_kv_cache_events`: False
## Benchmarking and Profiling

No runtime impact: single boolean expression change. No kernel or scheduling changes.

## Checklist

### DIFF
--- a/python/sglang/srt/managers/scheduler.py
+++ b/python/sglang/srt/managers/scheduler.py
@@ -298,8 +298,8 @@ class Scheduler(
         self.enable_metrics_for_all_schedulers = (
             server_args.enable_metrics_for_all_schedulers
         )
-        self.enable_kv_cache_events = (
-            server_args.kv_events_config is not None and tp_rank == 0
+        self.enable_kv_cache_events = bool(
+            server_args.kv_events_config and tp_rank == 0
         )
         self.enable_trace = server_args.enable_trace
         self.stream_interval = server_args.stream_interval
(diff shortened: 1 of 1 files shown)


JustinTong0323 (approved):

ishandhanani:
Good catch - thank you for this. Unfortunate CI didn't catch this

ishandhanani (approved):

PR #11207, opened 2025-10-04
fix: disable deepgemm on hopper cu124 temporarily

zhyncs:
## Checklist

### DIFF
--- a/python/sglang/srt/layers/quantization/deep_gemm_wrapper/configurer.py
+++ b/python/sglang/srt/layers/quantization/deep_gemm_wrapper/configurer.py
@@ -1,5 +1,7 @@
 import logging
 
+import torch
+
 from sglang.srt.utils import get_bool_env_var, get_device_sm, is_blackwell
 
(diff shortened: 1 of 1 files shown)


zhyncs:
@Fridge003 @fzyzcjy @Alcanderian

python/sglang/srt/layers/quantization/deep_gemm_wrapper/configurer.py
```diff
@@ -9,7 +11,9 @@ def _compute_enable_deep_gemm():
     sm_version = get_device_sm()
     if sm_version < 90:
         return False
-
+    if sm_version == 90 and torch.version.cuda is not None:
+        if torch.version.cuda != "12.8" and torch.version.cuda != "12.9":
+            return False
```
fzyzcjy, line 19: nit: shall we only disable when is_in_ci()

zhyncs: deepep tests will fail

Issue #10438, opened 2025-09-15
[Feature] support prefix caching for qwen 3 next


zhyncs:
### Checklist

- [ ] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [ ] 2. Please use English, otherwise it will be closed.

### Motivation

as titled

Swipe4057:
That would be cool!

zhyncs:
fixed by https://github.com/sgl-project/sglang/pull/11214

PR #11214, opened 2025-10-04
[Feature] Support mamba radix cache v0

yizhang2077:
## Motivation

ref https://github.com/sgl-project/sglang/issues/10438. add radix cache for mamba, we will implement page_size > 1 and [Marconi](https://arxiv.org/html/2411.19379v1) soon

Co-authored-by: hanming-lu <hanming@x.ai>
Co-authored-by: hzh0425 <hzh0425@apache.org>
Co-authored-by: thalahors <ericalcaide1@gmail.com>
## Modifications

ref: [doc](https://docs.google.com/document/d/1ZYBsZaz58NIIFmlOuQXqKPruIjv1ZJIUUrIj-sZlT4U/edit?tab=t.0#heading=h.f8tkurfljzli)

## Accuracy Tests
```
python3 -m sglang.launch_server --model Qwen/Qwen3-Next-80B-A3B-Instruct --tp 4  --chunked-prefill-size 64 --max-running-requests 16

python3 benchmark/gsm8k/bench_sglang.py --num-question 1000
Accuracy: 0.949
Invalid: 0.000
Latency: 346.597 s
Output throughput: 485.867 token/s
```

## Benchmarking and Profiling

```
# multi-turn benchmark

python3 -m sglang.bench_serving --backend sglang --dataset-name generated-shared-prefix --gsp-num-groups 50 --gsp-prompts-per-group 10 --gsp-system-prompt-len 10240 --gsp-question-len 256 --gsp-output-len 128 --max-concurrency 5  --port 30000

# without radix cache
 python3 -m sglang.launch_server --model Qwen/Qwen3-Next-80B-A3B-Instruct --tp 4 --disable-radix-cache

Max request concurrency:                 5         
Successful requests:                     500       
Benchmark duration (s):                  189.37    
Total input tokens:                      5521483   
Total generated tokens:                  64000     
Total generated tokens (retokenized):    63980     
Request throughput (req/s):              2.64      
Input token throughput (tok/s):          29157.63  
Output token throughput (tok/s):         337.97    
Total token throughput (tok/s):          29495.60  
Concurrency:                             5.00      
Mean E2E Latency (ms):                   1891.89   
Median E2E Latency (ms):                 1884.63   
Mean TTFT (ms):                          890.59    
Median TTFT (ms):                        1065.17   
P99 TTFT (ms):                           1180.80   
Mean ITL (ms):                           7.90      
Median ITL (ms):                         6.31      
P95 ITL (ms):                            6.57      
P99 ITL (ms):                            8.14      
Max ITL (ms):                            751.16    

# with radix cache
 python3 -m sglang.launch_server --model Qwen/Qwen3-Next-80B-A3B-Instruct --tp 4

Max request concurrency:                 5         
Successful requests:                     500       
Benchmark duration (s):                  141.38    
Total input tokens:                      5521483   
Total generated tokens:                  64000     
Total generated tokens (retokenized):    63985     
Request throughput (req/s):              3.54      
Input token throughput (tok/s):          39055.43  
Output token throughput (tok/s):         452.69    
Total token throughput (tok/s):          39508.12  
Concurrency:                             5.00      
Mean E2E Latency (ms):                   1412.49   
Median E2E Latency (ms):                 1337.07   
Mean TTFT (ms):                          507.33    
Median TTFT (ms):                        489.45    
P99 TTFT (ms):                           1336.39   
Mean ITL (ms):                           7.14      
Median ITL (ms):                         6.30      
P95 ITL (ms):                            6.85      
P99 ITL (ms):                            8.87      
Max ITL (ms):                            1484.16   
```
## Checklist


Swipe4057:
The error in https://github.com/sgl-project/sglang/actions/runs/18244671950/job/51951559880?pr=11214 is definitely not caused by this mr; it occurs in the main branch. Furthermore, we need to fix the line here that disabled the radix cache anyway: https://github.com/sgl-project/sglang/blob/85c1f7937781199203b38bb46325a2840f353a04/python/sglang/srt/model_executor/model_runner.py#L351

Swipe4057:
You need to fix the typo and rename the token_msg variables to token_usage_msg:

https://github.com/sgl-project/sglang/blob/b753f80ddf7aaf5b2cbf5f7c3733aeced87b8667/python/sglang/srt/managers/scheduler_metrics_mixin.py#L120

https://github.com/sgl-project/sglang/blob/b753f80ddf7aaf5b2cbf5f7c3733aeced87b8667/python/sglang/srt/managers/scheduler_metrics_mixin.py#L236

python/sglang/srt/mem_cache/mamba_radix_cache.py
```diff
@@ -0,0 +1,981 @@
...
+                evictable_size == lru_list_evictable_size
+            ), f"{self.mamba=}, total nodes: {total_nodes}, total lru plus 1: {total_lru_plus_1}, evictable size: {evictable_size} != lru list evictable size: {lru_list_evictable_size}"
+        except Exception as e:
+            msg = f"Mamba Radix tree sanity check failed, ping @yizhang2077: {e}"
+            logger.error(msg)
+            raise Exception(msg)
+
+
+class MambaRadixCache(BasePrefixCache):
+    def __init__(
```
ispobock, line 318: Does it compatible with MTP? [EAGLE fix](https://github.com/sgl-project/sglang/pull/10846) also should be applied to MambaRadixCache.

test/srt/test_mamba_unittest.py
```diff
@@ -0,0 +1,324 @@
...
+
+from sglang.srt.managers.schedule_batch import Req
+from sglang.srt.mem_cache.allocator import TokenToKVPoolAllocator
+from sglang.srt.mem_cache.mamba_radix_cache import MambaRadixCache
+from sglang.srt.mem_cache.memory_pool import HybridLinearKVPool, HybridReqToTokenPool
+from sglang.srt.mem_cache.radix_cache import RadixKey
+from sglang.srt.sampling.sampling_params import SamplingParams
+
+
+class TestMamba(unittest.TestCase):
```
ispobock, line 16: Do we have e2e test for qwen3-next w/ radix cache?

yizhang2077: we have qwen3next e2e ut with gsm8k (w/ radix cache by default)

Swipe4057:
During testing, I discovered that the server crashed when token_usage reached more than 0.99.

yizhang2077:
> During testing, I discovered that the server crashed when token_usage reached more than 0.99.

Do you have reproduce command? I think token_usage > 0.99 is an abnormal state. (It is too large and other models will crash as well in this state)

Swipe4057:
> > During testing, I discovered that the server crashed when token_usage reached more than 0.99.
> 
> Do you have reproduce command? I think token_usage > 0.99 is an abnormal state. (It is too large and other models will crash as well in this state)


reproduce command (server H100, tp-size=2):
python bench_serving.py \
 --host 127.0.0.1 \
 --port 30000 \
 --dataset-name generated-shared-prefix \
--gsp-system-prompt-len 1000 \
--gsp-question-prompt-len 2744 \
--gsp-output-prompt-len 820 \
--gsp-num-groups 8 \
--gsp-prompts-per-group 128

The root cause is incorrect memory availability checking in the Mamba pool. Instead of checking the Mamba pool, only the MHA pool is used, which leads to attempted memory allocation in a full Mamba pool and subsequent server crash due to None being returned from mamba_pool.alloc().

Clode Sonnet's recommendations:

1. `Scheduler.check_memory()`:
   - Check availability of both pools (MHA and Mamba) separately.
2. `PrefillAdder.budget_state()`:
   - For hybrid models with Mamba, check availability of both pools separately.
3. `ScheduleBatch.alloc_token_slots()`:
   - For Mamba pool, use req_to_token_pool.mamba_pool.alloc() instead of token_to_kv_pool_allocator.alloc().
4. `MambaRadixCache.match_prefix()`:
   - Add Mamba pool availability check before allocation.
5. `MambaRadixCache.evict_mamba()`:
   - Add verification that eviction will free sufficient memory.
6. `Scheduler._add_request_to_queue()`:
   - Add Mamba pool availability check before _prefetch_kvcache().
7. `HybridReqToTokenPool.alloc()`:
   - Check Mamba pool availability before allocation.
8. Add logging when Mamba pool memory is insufficient for diagnostics.

python/sglang/srt/mem_cache/mamba_radix_cache.py
yizhang2077: yes, I think maybe we can do it in another PR

yizhang2077:
@Swipe4057 Mamba pool controls memory capability by setting available size to 3x `max_running_requests ` [here](https://github.com/sgl-project/sglang/blob/047ab9a17718acb63d9985cb4f0a4d6dab40dc6b/python/sglang/srt/model_executor/model_runner.py#L1432) . We can control `mamba_usage` to around 0.66 at most during this benchmark. I have tried your benchmark in `Qwen3-Next-80B-A3B-Instruct-FP8` in H100 and it did not crash (but I found another bug)

Swipe4057:
> @Swipe4057 Mamba pool controls memory capability by setting available size to 3x `max_running_requests ` [here](https://github.com/sgl-project/sglang/blob/047ab9a17718acb63d9985cb4f0a4d6dab40dc6b/python/sglang/srt/model_executor/model_runner.py#L1432) . We can control `mamba_usage` to around 0.66 at most during this benchmark. I have tried your benchmark in `Qwen3-Next-80B-A3B-Instruct-FP8` in H100 and it did not crash (but I found another bug)

Run the service with the command and try testing again:
environment:
      - SGLANG_ENABLE_JIT_DEEPGEMM=1
    command:
      --model-path /data/models/Qwen3-Next-80B-A3B-Instruct
      --served-model-name Qwen3-Next-80B-A3B-Instruct
      --cuda-graph-max-bs 512
      --cuda-graph-bs 1 2 4 8 16 24 32 40 48 56 64 72 80 88 96 104 112 120 128 136 144 152 160 168 176 184 192 200 208 216 224 232 240 248 256 264 272 280 288 296 304 312 320 328 336 344 352 360 368 376 384 392 400 408 416 424 432 440 448 456 464 472 480 488 496 504 512
      --sleep-on-idle
      --port 8027
      --host 0.0.0.0
      --schedule-policy lof
      --random-seed 11111
      --context-length 131072
      --grammar-backend xgrammar
      --tool-call-parser qwen25
      --enable-metrics
      --quantization w8a8_fp8
      --allow-auto-truncate
      --mamba-ssm-dtype bfloat16
      --max-running-requests 1024
      --tp-size 2
      --ep-size 2
      --chunked-prefill-size 16384
      --prefill-attention-backend flashinfer
      --decode-attention-backend flashinfer
      --mem-fraction-static 0.86
      --max-running-requests 1024
      --api-key 123

Model: https://huggingface.co/Qwen/Qwen3-Next-80B-A3B-Instruct

yizhang2077:
@Swipe4057 I have tried and it is ok. Could you share your server log? (error and `mamba_usage` in log are important items)

Swipe4057:
You shouldn't limit the number of queries in operation with the new magic number 3 for all operating modes. You need to include the condition that radix cache is enabled.
https://github.com/sgl-project/sglang/blob/67d4e34913ae1992219d41f08e7eee826385e098/python/sglang/srt/model_executor/model_runner.py#L1430-L1432

Swipe4057:
Restart no longer occurs. I tested the new code, here are my results, I wrote the server launch command earlier:

main, disable radix:
input_throughput: 17906

mr, enable radix:
input_throughput: 18279

## Full:

### without radix cache
{'backend': 'sglang', 'dataset_name': 'generated-shared-prefix', 'request_rate': inf, 'max_concurrency': None, 'sharegpt_output_len': None, 'random_input_len': 1024, 'random_output_len': 1024, 'random_range_ratio': 0.0, 'duration': 225.5629655458033, 'completed': 1024, 'total_input_tokens': 4039046, 'total_output_tokens': 839680, 'total_output_tokens_retokenized': 839041, 'request_throughput': 4.5397523371010315, 'input_throughput': 17906.512224764232, 'output_throughput': 3722.596916422846, 'mean_e2e_latency_ms': 131048.26526600846, 'median_e2e_latency_ms': 114622.40997888148, 'std_e2e_latency_ms': 63244.031925246876, 'p99_e2e_latency_ms': 225238.2608978264, 'mean_ttft_ms': 102325.65577756759, 'median_ttft_ms': 91108.4556542337, 'std_ttft_ms': 62789.35906378496, 'p99_ttft_ms': 202928.22604222223, 'mean_tpot_ms': 35.070341255727556, 'median_tpot_ms': 33.90271993074225, 'std_tpot_ms': 5.1958526217693874, 'p99_tpot_ms': 45.5167203996517, 'mean_itl_ms': 35.07019149012473, 'median_itl_ms': 28.129231184720993, 'std_itl_ms': 245.95497367320698, 'p95_itl_ms': 29.61028926074505, 'p99_itl_ms': 35.236083529889754, 'concurrency': 594.9266685143978, 'accept_length': None}


### with radix cache
{'backend': 'sglang', 'dataset_name': 'generated-shared-prefix', 'request_rate': inf, 'max_concurrency': None, 'sharegpt_output_len': None, 'random_input_len': 1024, 'random_output_len': 1024, 'random_range_ratio': 0.0, 'duration': 220.95910985954106, 'completed': 1024, 'total_input_tokens': 4039046, 'total_output_tokens': 839680, 'total_output_tokens_retokenized': 839290, 'request_throughput': 4.634341623891111, 'input_throughput': 18279.608397080956, 'output_throughput': 3800.1601315907114, 'mean_e2e_latency_ms': 133114.31396250374, 'median_e2e_latency_ms': 145177.0340781659, 'std_e2e_latency_ms': 59586.806921843796, 'p99_e2e_latency_ms': 220620.25247588754, 'mean_ttft_ms': 94039.92438557907, 'median_ttft_ms': 100580.67605737597, 'std_ttft_ms': 57945.60860232475, 'p99_ttft_ms': 197400.40578043088, 'mean_tpot_ms': 47.70987738330239, 'median_tpot_ms': 42.28862401516483, 'std_tpot_ms': 28.60189182856475, 'p99_tpot_ms': 227.09880975949528, 'mean_itl_ms': 47.710699266049296, 'median_itl_ms': 30.049152672290802, 'std_itl_ms': 880.109046863313, 'p95_itl_ms': 34.64645054191351, 'p99_itl_ms': 279.4529473409058, 'concurrency': 616.8972059321408, 'accept_length': None}




## bench:
(server H100, tp-size=2):
python bench_serving.py
--host 127.0.0.1
--port 30000
--dataset-name generated-shared-prefix
--gsp-system-prompt-len 1000
--gsp-question-prompt-len 2744
--gsp-output-prompt-len 820
--gsp-num-groups 8
--gsp-prompts-per-group 128


## log:
[[2025-10-05 18](tel:2025-10-05 18):37:30 TP0 EP0] Decode batch. #running-req: 235, #full token: 818755, full token usage: 0.99, mamba num: 470, mamba usage: 0.46, cuda graph: True, gen throughput (token/s): 7752.32, #queue-req: 789,

[[2025-10-05 18](tel:2025-10-05 18):37:31 TP0 EP0] Decode batch. #running-req: 235, #full token: 828155, full token usage: 1.00, mamba num: 470, mamba usage: 0.46, cuda graph: True, gen throughput (token/s): 7751.90, #queue-req: 789,

[[2025-10-05 18](tel:2025-10-05 18):37:31 TP0 EP0] KV cache pool is full. Retract requests. #retracted_reqs: 1, #aborted_retracted_reqs: 0, #new_token_ratio: 0.0980 -> 0.7646

[[2025-10-05 18](tel:2025-10-05 18):37:31 TP1 EP1] KV cache pool is full. Retract requests. #retracted_reqs: 1, #aborted_retracted_reqs: 0, #new_token_ratio: 0.0980 -> 0.7646

[[2025-10-05 18](tel:2025-10-05 18):37:32 TP1 EP1] KV cache pool is full. Retract requests. #retracted_reqs: 1, #aborted_retracted_reqs: 0, #new_token_ratio: 0.7506 -> 0.7829

[[2025-10-05 18](tel:2025-10-05 18):37:32 TP0 EP0] KV cache pool is full. Retract requests. #retracted_reqs: 1, #aborted_retracted_reqs: 0, #new_token_ratio: 0.7506 -> 0.7829

[[2025-10-05 18](tel:2025-10-05 18):37:32 TP0 EP0] KV cache pool is full. Retract requests. #retracted_reqs: 1, #aborted_retracted_reqs: 0, #new_token_ratio: 0.7679 -> 0.8024

[[2025-10-05 18](tel:2025-10-05 18):37:32 TP1 EP1] KV cache pool is full. Retract requests. #retracted_reqs: 1, #aborted_retracted_reqs: 0, #new_token_ratio: 0.7679 -> 0.8024

[[2025-10-05 18](tel:2025-10-05 18):37:32 TP0 EP0] Decode batch. #running-req: 232, #full token: 826859, full token usage: 1.00, mamba num: 464, mamba usage: 0.45, cuda graph: True, gen throughput (token/s): 7658.65, #queue-req: 792,

[[2025-10-05 18](tel:2025-10-05 18):37:33 TP1 EP1] KV cache pool is full. Retract requests. #retracted_reqs: 1, #aborted_retracted_reqs: 0, #new_token_ratio: 0.7884 -> 0.8207

[[2025-10-05 18](tel:2025-10-05 18):37:33 TP0 EP0] KV cache pool is full. Retract requests. #retracted_reqs: 1, #aborted_retracted_reqs: 0, #new_token_ratio: 0.7884 -> 0.8207

[[2025-10-05 18](tel:2025-10-05 18):37:33 TP0 EP0] KV cache pool is full. Retract requests. #retracted_reqs: 1, #aborted_retracted_reqs: 0, #new_token_ratio: 0.8057 -> 0.8402

[[2025-10-05 18](tel:2025-10-05 18):37:33 TP1 EP1] KV cache pool is full. Retract requests. #retracted_reqs: 1, #aborted_retracted_reqs: 0, #new_token_ratio: 0.8057 -> 0.8402

[[2025-10-05 18](tel:2025-10-05 18):37:33 TP0 EP0] Decode batch. #running-req: 230, #full token: 828932, full token usage: 1.00, mamba num: 460, mamba usage: 0.45, cuda graph: True, gen throughput (token/s): 7786.25, #queue-req: 794,

[[2025-10-05 18](tel:2025-10-05 18):37:34 TP0 EP0] KV cache pool is full. Retract requests. #retracted_reqs: 1, #aborted_retracted_reqs: 0, #new_token_ratio: 0.8262 -> 0.8585

[[2025-10-05 18](tel:2025-10-05 18):37:34 TP1 EP1] KV cache pool is full. Retract requests. #retracted_reqs: 1, #aborted_retracted_reqs: 0, #new_token_ratio: 0.8262 -> 0.8585

[[2025-10-05 18](tel:2025-10-05 18):37:34 TP0 EP0] KV cache pool is full. Retract requests. #retracted_reqs: 1, #aborted_retracted_reqs: 0, #new_token_ratio: 0.8435 -> 0.8781

[[2025-10-05 18](tel:2025-10-05 18):37:34 TP1 EP1] KV cache pool is full. Retract requests. #retracted_reqs: 1, #aborted_retracted_reqs: 0, #new_token_ratio: 0.8435 -> 0.8781

[[2025-10-05 18](tel:2025-10-05 18):37:35 TP0 EP0] KV cache pool is full. Retract requests. #retracted_reqs: 1, #aborted_retracted_reqs: 0, #new_token_ratio: 0.8630 -> 0.8976

[[2025-10-05 18](tel:2025-10-05 18):37:35 TP1 EP1] KV cache pool is full. Retract requests. #retracted_reqs: 1, #aborted_retracted_reqs: 0, #new_token_ratio: 0.8630 -> 0.8976

[[2025-10-05 18](tel:2025-10-05 18):37:35] INFO:     127.0.0.1:33552 - "GET /health HTTP/1.1" 200 OK

[[2025-10-05 18](tel:2025-10-05 18):37:35 TP0 EP0] Decode batch. #running-req: 227, #full token: 827227, full token usage: 1.00, mamba num: 454, mamba usage: 0.44, cuda graph: True, gen throughput (token/s): 7641.70, #queue-req: 797,

[[2025-10-05 18](tel:2025-10-05 18):37:35 TP0 EP0] KV cache pool is full. Retract requests. #retracted_reqs: 1, #aborted_retracted_reqs: 0, #new_token_ratio: 0.8825 -> 0.9171

[[2025-10-05 18](tel:2025-10-05 18):37:35 TP1 EP1] KV cache pool is full. Retract requests. #retracted_reqs: 1, #aborted_retracted_reqs: 0, #new_token_ratio: 0.8825 -> 0.9171

[[2025-10-05 18](tel:2025-10-05 18):37:35 TP0 EP0] KV cache pool is full. Retract requests. #retracted_reqs: 1, #aborted_retracted_reqs: 0, #new_token_ratio: 0.9020 -> 0.9366

[[2025-10-05 18](tel:2025-10-05 18):37:36 TP1 EP1] KV cache pool is full. Retract requests. #retracted_reqs: 1, #aborted_retracted_reqs: 0, #new_token_ratio: 0.9020 -> 0.9366

[[2025-10-05 18](tel:2025-10-05 18):37:36 TP0 EP0] Decode batch. #running-req: 225, #full token: 828970, full token usage: 1.00, mamba num: 450, mamba usage: 0.44, cuda graph: True, gen throughput (token/s): 7525.47, #queue-req: 799,

[[2025-10-05 18](tel:2025-10-05 18):37:36 TP1 EP1] KV cache pool is full. Retract requests. #retracted_reqs: 1, #aborted_retracted_reqs: 0, #new_token_ratio: 0.9215 -> 0.9561

[[2025-10-05 18](tel:2025-10-05 18):37:36 TP0 EP0] KV cache pool is full. Retract requests. #retracted_reqs: 1, #aborted_retracted_reqs: 0, #new_token_ratio: 0.9215 -> 0.9561

[[2025-10-05 18](tel:2025-10-05 18):37:36 TP1 EP1] KV cache pool is full. Retract requests. #retracted_reqs: 1, #aborted_retracted_reqs: 0, #new_token_ratio: 0.9400 -> 0.9768

[[2025-10-05 18](tel:2025-10-05 18):37:36 TP0 EP0] KV cache pool is full. Retract requests. #retracted_reqs: 1, #aborted_retracted_reqs: 0, #new_token_ratio: 0.9400 -> 0.9768

[[2025-10-05 18](tel:2025-10-05 18):37:37 TP0 EP0] KV cache pool is full. Retract requests. #retracted_reqs: 1, #aborted_retracted_reqs: 0, #new_token_ratio: 0.9618 -> 0.9963

[[2025-10-05 18](tel:2025-10-05 18):37:37 TP1 EP1] KV cache pool is full. Retract requests. #retracted_reqs: 1, #aborted_retracted_reqs: 0, #new_token_ratio: 0.9618 -> 0.9963

[[2025-10-05 18](tel:2025-10-05 18):37:37 TP0 EP0] Decode batch. #running-req: 222, #full token: 826827, full token usage: 1.00, mamba num: 444, mamba usage: 0.43, cuda graph: True, gen throughput (token/s): 7609.35, #queue-req: 802,

[[2025-10-05 18](tel:2025-10-05 18):37:37 TP1 EP1] KV cache pool is full. Retract requests. #retracted_reqs: 1, #aborted_retracted_reqs: 0, #new_token_ratio: 0.9803 -> 1.0000

[[2025-10-05 18](tel:2025-10-05 18):37:37 TP0 EP0] KV cache pool is full. Retract requests. #retracted_reqs: 1, #aborted_retracted_reqs: 0, #new_token_ratio: 0.9803 -> 1.0000

[[2025-10-05 18](tel:2025-10-05 18):37:38 TP0 EP0] Prefill batch. #new-seq: 4, #new-token: 15775, #cached-token: 0, full token usage: 0.00, mamba usage: 0.00, #running-req: 220, #queue-req: 799,

yizhang2077:
> You shouldn't limit the number of queries in operation with the new magic number 3 for all operating modes. You need to include the condition that radix cache is enabled.
> 
> https://github.com/sgl-project/sglang/blob/67d4e34913ae1992219d41f08e7eee826385e098/python/sglang/srt/model_executor/model_runner.py#L1430-L1432

For disable radix cache, I think `max_mamba_cache_size` larger than `max_running_requests` is also fine.

Swipe4057:
[yizhang2077](https://github.com/yizhang2077) Do you have any idea why my test is so much worse than yours in performance?

Although in my test there are 8 groups of requests with the same system hints of 1000 tokens, in the log I see cached-token: 0

log:
[[2025-10-06 06](tel:2025-10-06 06):32:38 TP0 EP0] Prefill batch. #new-seq: 4, #new-token: 15821, #cached-token: 0, full token usage: 0.23, mamba usage: 0.24, #running-req: 125, #queue-req: 895,

[[2025-10-06 06](tel:2025-10-06 06):32:39 TP0 EP0] Prefill batch. #new-seq: 4, #new-token: 15801, #cached-token: 0, full token usage: 0.23, mamba usage: 0.25, #running-req: 129, #queue-req: 891,

[[2025-10-06 06](tel:2025-10-06 06):32:39 TP0 EP0] Prefill batch. #new-seq: 4, #new-token: 15785, #cached-token: 0, full token usage: 0.24, mamba usage: 0.26, #running-req: 133, #queue-req: 887,

[[2025-10-06 06](tel:2025-10-06 06):32:40 TP0 EP0] Prefill batch. #new-seq: 4, #new-token: 15748, #cached-token: 0, full token usage: 0.25, mamba usage: 0.26, #running-req: 137, #queue-req: 883,

[[2025-10-06 06](tel:2025-10-06 06):32:40 TP0 EP0] Prefill batch. #new-seq: 4, #new-token: 15791, #cached-token: 0, full token usage: 0.25, mamba usage: 0.27, #running-req: 141, #queue-req: 879,

[[2025-10-06 06](tel:2025-10-06 06):32:40 TP0 EP0] Prefill batch. #new-seq: 4, #new-token: 15719, #cached-token: 0, full token usage: 0.26, mamba usage: 0.28, #running-req: 145, #queue-req: 875,

[[2025-10-06 06](tel:2025-10-06 06):32:41 TP0 EP0] Prefill batch. #new-seq: 4, #new-token: 15655, #cached-token: 0, full token usage: 0.27, mamba usage: 0.29, #running-req: 149, #queue-req: 871,

[[2025-10-06 06](tel:2025-10-06 06):32:41 TP0 EP0] Prefill batch. #new-seq: 4, #new-token: 15803, #cached-token: 0, full token usage: 0.27, mamba usage: 0.29, #running-req: 153, #queue-req: 867,

[[2025-10-06 06](tel:2025-10-06 06):32:42 TP0 EP0] Prefill batch. #new-seq: 4, #new-token: 15746, #cached-token: 0, full token usage: 0.28, mamba usage: 0.30, #running-req: 157, #queue-req: 863,

[[2025-10-06 06](tel:2025-10-06 06):32:42 TP0 EP0] Prefill batch. #new-seq: 4, #new-token: 15783, #cached-token: 0, full token usage: 0.29, mamba usage: 0.31, #running-req: 161, #queue-req: 859,

[[2025-10-06 06](tel:2025-10-06 06):32:42 TP0 EP0] Prefill batch. #new-seq: 4, #new-token: 15755, #cached-token: 0, full token usage: 0.29, mamba usage: 0.32, #running-req: 165, #queue-req: 855,

[[2025-10-06 06](tel:2025-10-06 06):32:43 TP0 EP0] Prefill batch. #new-seq: 4, #new-token: 15784, #cached-token: 0, full token usage: 0.30, mamba usage: 0.33, #running-req: 169, #queue-req: 851,

[[2025-10-06 06](tel:2025-10-06 06):32:43 TP0 EP0] Prefill batch. #new-seq: 4, #new-token: 15750, #cached-token: 0, full token usage: 0.31, mamba usage: 0.33, #running-req: 173, #queue-req: 847,

[[2025-10-06 06](tel:2025-10-06 06):32:43 TP0 EP0] Prefill batch. #new-seq: 4, #new-token: 15765, #cached-token: 0, full token usage: 0.32, mamba usage: 0.34, #running-req: 177, #queue-req: 843,

Swipe4057:
> > You shouldn't limit the number of queries in operation with the new magic number 3 for all operating modes. You need to include the condition that radix cache is enabled.
> > https://github.com/sgl-project/sglang/blob/67d4e34913ae1992219d41f08e7eee826385e098/python/sglang/srt/model_executor/model_runner.py#L1430-L1432
> 
> For disable radix cache, I think `max_mamba_cache_size` larger than `max_running_requests` is also fine.

there will be OOM

hzh0425:
> [yizhang2077](https://github.com/yizhang2077) Do you have any idea why my test is so much worse than yours in performance?
> 
> Although in my test there are 8 groups of requests with the same system hints of 1000 tokens, in the log I see cached-token: 0

Hi  @Swipe4057 

It might be due to too many requests causing the cache in device to be evicted.

You can try adding --max-concurrency 5

Swipe4057:
> > [yizhang2077](https://github.com/yizhang2077) Do you have any idea why my test is so much worse than yours in performance?
> > Although in my test there are 8 groups of requests with the same system hints of 1000 tokens, in the log I see cached-token: 0
> 
> Hi @Swipe4057
> 
> It might be due to too many requests causing the cache in device to be evicted.
> 
> You can try adding --max-concurrency 5

Understood. Unfortunately, using `--max-concurrency 5` is not suitable for our production environment.

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -347,8 +347,6 @@ def initialize(self, min_per_gpu_memory: float):
                 self.is_hybrid = self.model_config.is_hybrid = True
 
         if self.is_hybrid_gdn:
-            logger.warning("Hybrid GDN model detected, disable radix cache")
-            self.server_args.disable_radix_cache = True
             if self.server_args.max_mamba_cache_size is None:
                 if self.server_args.max_running_requests is not None:
                     self.server_args.max_mamba_cache_size = (
```
hanming-lu: should it be `self.server_args.max_mamba_cache_size = 3 * self.server_args.max_running_requests`? Similar for the 512 below. let's make 3 a constant somewhere?

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -1430,7 +1428,8 @@ def init_memory_pool(
                 4096,
             )
         if self.is_hybrid_gdn:
-            max_num_reqs = min(max_num_reqs, self.server_args.max_mamba_cache_size)
+            # for mamba cache radix, it need be divided by 3 (magic number now). (yizhang2077)
```
hanming-lu: as discussed, we can explain the reason why we use 3 here? 
> here is likely why 2 won't work: 
before `self.dec_lock_ref(req.last_node)` in cache_unfinished_req, it is holding 3 states (i.e. matched, own copy, new cached copy)

python/sglang/srt/model_executor/model_runner.py
hanming-lu: I'll do in a separate PR that `max_mamba_cache_size` being relative to full kv token pool size instead of max running requests

python/sglang/srt/mem_cache/mamba_radix_cache.py
```diff
@@ -0,0 +1,986 @@
...
+        Args:
+            key: A RadixKey contains token IDs to find a matching prefix.
+        Returns:
+            A tuple of a tensor of matching prefix token IDs and
+            the last node that contains the prefix values. Note that
+            this API can modify the internal state of the Radix tree.
+            The last node create a new child if the prefix is shorter
+            than the last node's value.
+        """
+        cow_mamba: bool = kwargs.get("cow_mamba", False)
```
hanming-lu, line 369: what does `cow` refer to?

python/sglang/srt/mem_cache/mamba_radix_cache.py
```diff
@@ -0,0 +1,986 @@
...
+    def _evict_leaf_node(
+        self, x: TreeNode, is_evict_mamba: bool
+    ) -> Tuple[int, int, TreeNode, TreeNode]:
+        assert (
+            x.full_lock_ref == 0
+        ), f"leaf node with full lock must also have mamba lock, {x.id=} {x.full_lock_ref=}"
+        # 1. a leaf node, free full tokens and mamba
+        self.token_to_kv_pool_allocator.free(x.value)
+        full_num_evicted = len(x.value)
+        self.req_to_token_pool.mamba_pool.free(x.mamba_value)
```
hanming-lu, line 554: explicitly assert `x.mamba_value` is not None

python/sglang/srt/mem_cache/mamba_radix_cache.py
```diff
@@ -0,0 +1,986 @@
...
+        print(f"#full_tokens: {total_size}, #mamba_num: {total_mamba_size}")
+
+    def total_size(self) -> Tuple[int, int]:
+        return self._total_size_helper()
+
+    def _evict_leaf_node(
+        self, x: TreeNode, is_evict_mamba: bool
+    ) -> Tuple[int, int, TreeNode, TreeNode]:
+        assert (
+            x.full_lock_ref == 0
```
hanming-lu: nit: also assert `mamba_lock_ref == 0`

python/sglang/srt/mem_cache/mamba_radix_cache.py
```diff
@@ -0,0 +1,986 @@
...
+            if v == node:
+                break
+        del node.parent.children[k]
+        self.full_evictable_size_ -= len(node.key)
+        self.mamba_evictable_size_ -= len(node.mamba_value)
+
+    def _tombstone_internal_node(self, node: TreeNode) -> None:
+        assert len(node.children) != 0, f"Cannot tombstone a leaf node, {node.id=}"
+        self.mamba_evictable_size_ -= (
+            len(node.mamba_value) if node.mamba_value is not None else 0
```
hanming-lu: assert `node.mamba_value is not None` instead

python/sglang/srt/mem_cache/mamba_radix_cache.py
```diff
@@ -0,0 +1,986 @@
...
+        # evict lru leaf nodes until mamba_num_tokens is reached
+        while mamba_num_evicted < mamba_num and (self.mamba_lru_list.in_list(x)):
+            assert x.mamba_value is not None, f"node has no mamba value, {x.id=}"
+            assert x != self.root_node, f"root node is not evictable, {x.id=}"
+            assert x.mamba_lock_ref == 0, f"node is in use by mamba kv indices, {x.id=}"
+
+            if len(x.children) > 0:
+                # 1. an internal node, free mamba tokens.
+                self.req_to_token_pool.mamba_pool.free(x.mamba_value)
+                mamba_num_evicted += len(x.mamba_value)
```
hanming-lu, line 591: is there any case that `len(x.mamba_value) != 1`? should we put an assert?

python/sglang/srt/mem_cache/mamba_radix_cache.py
```diff
@@ -0,0 +1,986 @@
...
+        mamba_value_exist = False
+        if len(key):
+            new_node = TreeNode()
+            new_node.parent = node
+            new_node.key = key
+            new_node.value = value
+            new_node.mamba_value = mamba_value
+            self.full_lru_list.insert_mru(new_node)
+            self.full_evictable_size_ += len(value)
+            if mamba_value is not None:
```
hanming-lu: when would mamba_value be None? If won't happen, we should do assert instead? Similar for below

python/sglang/srt/mem_cache/mamba_radix_cache.py
```diff
@@ -0,0 +1,986 @@
...
+
+    def cache_finished_req(self, req: Req) -> None:
+        """Cache request when it finishes."""
+        if self.disable:
+            kv_indices = self.req_to_token_pool.req_to_token[
+                req.req_pool_idx,
+                : len(req.origin_input_ids) + max(len(req.output_ids) - 1, 0),
+            ]
+            self.token_to_kv_pool_allocator.free(kv_indices)
+            self.req_to_token_pool.free(req.req_pool_idx)
```
hanming-lu, line 434: Do we need to do `free_mamba_cache=True` here?

python/sglang/srt/mem_cache/mamba_radix_cache.py
```diff
@@ -0,0 +1,986 @@
...
+        token_ids = req.fill_ids
+        kv_indices = self.req_to_token_pool.req_to_token[
+            req.req_pool_idx, : len(token_ids)
+        ]
+        page_aligned_len = len(kv_indices)
+        page_aligned_kv_indices = kv_indices.to(dtype=torch.int64, copy=True)
+        page_aligned_token_ids = token_ids[:page_aligned_len]
+
+        # Radix Cache takes one ref in memory pool
+        # Note: the insert function already frees the overlapped kv_indices
```
hanming-lu: This Note seems not true here.

python/sglang/srt/mem_cache/mamba_radix_cache.py
```diff
@@ -0,0 +1,986 @@
...
+        if mamba_exist:
+            self.req_to_token_pool.mamba_pool.free(mamba_value_forked)
+
+        # The prefix indices could be updated, reuse it
+        new_indices, new_last_node, _, _ = self.match_prefix(
+            RadixKey(token_ids, req.extra_key)
+        )
+
+        if not mamba_exist:
+            assert torch.equal(new_last_node.mamba_value, mamba_value_forked)
```
hanming-lu, line 516: I am trying to understand this assertion, could you explain 1) why it only checks if not mamba_exist; 2) I feel this assertion is not valid, please correct me if wrong. 
- For example, we have 2 requests a,b with token ids a:[1,2] and b:[1,2,3]
- chunk prefill size is 2, and 
- for the prefill batch, req a is prefilling token id 2, req b is prefilling token id 1. 
- req a calls `cache_finished_req` first, then b calls `cache_unfinished_req` will match [1,2] instead of just [1]. Then this assertion will fail? If I'm correct, we should handle this case and copy over the matched state?

hanming-lu (commented):
Overall looks good! Most comments are minor, the one regarding the assert in `cache_unfinished_req` is more critical. Thanks!

python/sglang/srt/mem_cache/mamba_radix_cache.py
yizhang2077: it means "copy on write"

python/sglang/srt/mem_cache/mamba_radix_cache.py
yizhang2077: I used to explicitly set `free_mamba_cache=Fasle` since it is more uncommon and `free_mamba_cache` is True by default

python/sglang/srt/mem_cache/mamba_radix_cache.py
yizhang2077: yes, I think we need add assert here

python/sglang/srt/mem_cache/mamba_radix_cache.py
yizhang2077: 1) if `mamba_exist=True`,  `mamba_value_forked` will be free in previous, which will clean mamba cache inside, so it shouldn't add assertion here.
2) I add this assertion according to this: we insert `page_aligned_token_ids` as a key firstly and then call `match_prefix` with key `page_aligned_token_ids` immediately , in this case, I think the `last_node` returned by `match_prefix` must be the node generate by `insert`, so `new_last_node.mamba_value` must be equal to `mamba_value_forked`.

python/sglang/srt/mem_cache/mamba_radix_cache.py
yizhang2077: I think it will not happen, we do need add assertion here

python/sglang/srt/mem_cache/mamba_radix_cache.py
yizhang2077: I think we have added assertion in `evict_mamba` and we only need remove `if-else` here

python/sglang/srt/model_executor/model_runner.py
yizhang2077: done

python/sglang/srt/mem_cache/mamba_radix_cache.py
yizhang2077: done

python/sglang/srt/mem_cache/mamba_radix_cache.py
yizhang2077: remove it

python/sglang/srt/mem_cache/mamba_radix_cache.py
yizhang2077: done

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -1438,8 +1444,6 @@ def init_memory_pool(
                 ),
                 4096,
             )
-        if self.is_hybrid_gdn:
```
yizhang2077: I remove it since after this modification `max_num_reqs` equals to `max_running_requests` and it will be `max_mamba_cache_size // 3` by default

python/sglang/srt/model_executor/model_runner.py
yizhang2077: I remove this line and uses `MAMBA_CACHE_SIZE_MAX_RUNNING_REQUESTS_RATIO` to control it

python/sglang/srt/mem_cache/mamba_radix_cache.py
yizhang2077: >  then b calls cache_unfinished_req will match [1,2] instead of just [1]. Then this assertion will fail? If I'm correct, we should handle this case and copy over the matched state?

for your example, I think b also match [1] since the key for `match_prefix` is just [1]?

hanming-lu: You are right, the key for match_prefix is [1]. Makes sense. The current code looks good.

I have a feeling that this assertion is costly, it compares two relatively big tensors and just for assertion, do you think we should do it for debug purpose only?

hanming-lu (approved):
Logic looks good to me!

I am not sure about the status of crashing behavior from @Swipe4057 , if it is also fixed, we are good to merge.

Swipe4057:
> Logic looks good to me!
> 
> I am not sure about the status of crashing behavior from @Swipe4057 , if it is also fixed, we are good to merge.

I'll test the current code tomorrow.

hanming-lu:
seems like mamba tree cache sanity check is not running, let's add it?
```
def check_tree_cache(self):
        if self.is_hybrid and isinstance(self.tree_cache, SWARadixCache):
            self.tree_cache.sanity_check()
```

hanming-lu:
Scheduling related, no performance overhead if tested with `python3 -m sglang.bench_serving --backend sglang --dataset-name random --num-prompts 50 --random-input 24576 --random-output 1 --random-range-ratio 1`
See https://sgl-fru7574.slack.com/archives/C09DVT13FT8/p1759971353991489

> I found the mamba radix cache itself has ~12% prefill overhead. Not blocking PR merge, but we shouldn't settle on the 10% perf drop :) I tested with
> 
> ```
> python3 -m sglang.bench_serving --backend sglang --dataset-name random --num-prompts 50 --random-input 20480 --random-output 1 --random-range-ratio 1
> 
> w/ radix cache
> python3 -m sglang.launch_server --model Qwen/Qwen3-Next-80B-A3B-Instruct --tp 8
> Input token throughput (tok/s):          64229.48
> 
> w/o radix cache
> python3 -m sglang.launch_server --model Qwen/Qwen3-Next-80B-A3B-Instruct --tp 8 --disable-radix-cache
> Input token throughput (tok/s):          72806.69
> ```
> I am trying to experiment where it comes from. I tried:
> - comment out reset to 0 in MambaPool free
> - comment out copy_from in fork_from
> 
> Both don't make a difference.

hzh0425:
> I found the mamba radix cache itself has ~12% prefill overhead. Not blocking PR merge, but we shouldn't settle on the 10% perf drop :) I tested with


Currently, cache_unfinish_req will fork the mamba cache once, which might be introducing overhead.  @hanming-lu

python/sglang/srt/mem_cache/mamba_radix_cache.py
yizhang2077: `mamba_value` and `mamba_value_forked` are both indexes for big tensor, so I think they are not big. (length = 1)

Swipe4057:
> > I found the mamba radix cache itself has ~12% prefill overhead. Not blocking PR merge, but we shouldn't settle on the 10% perf drop :) I tested with
> 
> Currently, cache_unfinish_req will fork the mamba cache once, which might be introducing overhead. @hanming-lu

Please take a look at my test that I conducted earlier https://github.com/sgl-project/sglang/pull/11214#issuecomment-3369254479
If you specify the generated-shared-prefix dataset, the script will generate synthetic requests with specified system prompt lengths and user (variable) parts. I had 8 groups with identical system prompts.

My results (duplicating here) were as follows, on 2x H100 GPUs:
reproduce command (server H100, tp-size=2):
python bench_serving.py
--host 127.0.0.1
--port 30000
--dataset-name generated-shared-prefix
--gsp-system-prompt-len 1000
--gsp-question-prompt-len 2744
--gsp-output-prompt-len 820
--gsp-num-groups 8
--gsp-prompts-per-group 128

main, disable radix:
inputthroughput: 17906

mr, enable radix:
inputthroughput: 18279

I shared server logs, and all prefills looked like this:
Prefill batch. #new-seq: 4, #new-token: 15791, #cached-token: 0, full token usage: 0.25, mamba usage: 0.27, #running-req: 141, #queue-req: 879

That is, cached-token: 0! Although this is impossible, since cache matches should definitely exist) However, if you reduce the number of concurrent requests, for example to 5, or send requests one after another, cache matches do appear!

So I think there's a bug here.

I can't test again at the moment, our server is undergoing maintenance, I'll verify in the near future.


It would also be interesting to test something like this (--gsp-question-prompt-len 0):
python bench_serving.py
--dataset-name generated-shared-prefix
--gsp-system-prompt-len 1000
--gsp-question-prompt-len 0
--gsp-output-prompt-len 820
--gsp-num-groups 8
--gsp-prompts-per-group 128

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -1302,15 +1284,50 @@ def profile_max_num_token(self, total_gpu_memory: int):
...
+            # 2. mamba_state_memory / full_kv_cache_memory == server_args.mamba_full_memory_ratio
+            mamba_state_memory_raw = (
+                total_rest_memory
+                * server_args.mamba_full_memory_ratio
+                / (1 + server_args.mamba_full_memory_ratio)
+            )
+            # calculate the max_mamba_cache_size based on the given total mamba memory
+            server_args.max_mamba_cache_size = int(
+                (mamba_state_memory_raw * (1 << 30))
+                // config.mamba2_cache_params.mamba_cache_per_req
```
veden, line 1355: When using `--speculative-num-steps 3 --speculative-eagle-topk 1 --speculative-num-draft-tokens 4 --speculative-algo NEXTN` the `config.mamba2_cache_params.mamba_cache_per_req` is 0 and throws a divide by zero error when loading the draft MTP layer

```
  File "/ui/models/code/sglang/python/sglang/srt/model_executor/model_runner.py", line 1318, in handle_max_mamba_cache
    (mamba_state_memory_raw * (1 << 30))
ZeroDivisionError: float floor division by zero
```

yizhang2077:
> seems like mamba tree cache sanity check is not running, let's add it?
> 
> ```
> def check_tree_cache(self):
>         if self.is_hybrid and isinstance(self.tree_cache, SWARadixCache):
>             self.tree_cache.sanity_check()
> ```

open sanity_check

python/sglang/srt/model_executor/model_runner.py
hanming-lu: need some modifications to properly support spec dec. for spec dec reverted to previous behavior. Done.

yizhang2077: fix it, could you try again?

veden: Fixed, thank you.

ispobock (approved):

ispobock:
@yizhang2077 Could you resolve the conflicts? Then we can merge it.

Swipe4057:
[yizhang2077](https://github.com/yizhang2077) [hanming-lu](https://github.com/hanming-lu) [ispobock](https://github.com/ispobock) I retested your new code.,
Here's the command to start the server on 2xH100:

--model-path /data/models/Qwen3-Next-80B-A3B-Instruct --served-model-name Qwen3-Next-80B-A3B-Instruct --cuda-graph-max-bs 512 --cuda-graph-bs 1 2 4 8 16 24 32 40 48 56 64 72 80 88 96 104 112 120 128 136 144 152 160 168 176 184 192 200 208 216 224 232 240 248 256 264 272 280 288 296 304 312 320 328 336 344 352 360 368 376 384 392 400 408 416 424 432 440 448 456 464 472 480 488 496 504 512 --sleep-on-idle --port 8027 --host 0.0.0.0 --api-key ${SGLANG_API_KEY} --schedule-policy lof --random-seed 11111 --context-length 131072 --grammar-backend xgrammar --prefill-attention-backend flashinfer --decode-attention-backend flashinfer --tool-call-parser qwen25 --enable-metrics --quantization w8a8_fp8 --allow-auto-truncate --tp-size 2 --ep-size 2 --chunked-prefill-size 16384 --max-running-requests 1024 --mem-fraction-static 0.87 --mamba-ssm-dtype bfloat16 --mamba-full-memory-ratio 4

TEST-1:
reproduce command:
python bench_serving.py
--host 127.0.0.1
--port 30000
--dataset-name generated-shared-prefix
--gsp-system-prompt-len 1000
--gsp-question-prompt-len 2744
--gsp-output-prompt-len 820
--gsp-num-groups 8
--gsp-prompts-per-group 128

results:
main, disable radix:
input_throughput: 18518

mr, enable radix:
input_throughput: 16174 - the results are only getting worse !!!

logs:
Server startup is slow and 503 errors are occurring (this does not happen in main)
image

image

TEST-2:
reproduce command:
python bench_serving.py
--host 127.0.0.1
--port 30000
--dataset-name generated-shared-prefix
--gsp-system-prompt-len 1000
--gsp-question-prompt-len 0 - pay attention to this line and the results !!!
--gsp-output-prompt-len 1 - pay attention to this line and the results !!!
--gsp-num-groups 8
--gsp-prompts-per-group 128

results:
main, disable radix:
input_throughput: 45264

mr, enable radix:
input_throughput: 39668

As you can see, the last test sends 8 groups of identical prompts with a length of 1000 characters, without a variable part and with only 1 token for generation. Performance with radix cache enabled dropped dramatically!
And there is simply no cache match!

Swipe4057:
[yizhang2077](https://github.com/yizhang2077) Try this code to fix null cache hit:
```
def match_prefix(self, key: RadixKey, **kwargs) -> MatchResult:
    """Find the matching prefix from the radix tree."""
    cow_mamba: bool = kwargs.get("cow_mamba", False)
    req: Req = kwargs.get("req", None)

    if self.disable or len(key) == 0:
        return MatchResult(
            device_indices=torch.empty(
                (0,),
                dtype=torch.int64,
                device=self.device,
            ),
            last_device_node=self.root_node,
            last_host_node=self.root_node,
        )

    # Get the full matched prefix (including tombstone nodes)
    full_value, full_last_node, mamba_last_node = self._match_prefix_helper(key)

    # copy mamba state to req local space if cow is true
    if cow_mamba and mamba_last_node.mamba_value is not None:
        # for reqs without mamba cache
        if req.mamba_pool_idx is None:
            dst_index = self.req_to_token_pool.mamba_pool.alloc(1)
            # try to alloc again, protect mamba_last_node from eviction
            if dst_index is None:
                self.inc_lock_ref(mamba_last_node)
                self.evict_mamba(1)
                dst_index = self.req_to_token_pool.mamba_pool.alloc(1)
                self.dec_lock_ref(mamba_last_node)
                assert dst_index is not None, "Can not alloc mamba cache"
            src_index = mamba_last_node.mamba_value
            self.req_to_token_pool.mamba_pool.copy_from(src_index, dst_index)
            req.mamba_pool_idx = dst_index[0]
        else:
            src_index = mamba_last_node.mamba_value
            dst_index = req.mamba_pool_idx.unsqueeze(0)
            self.req_to_token_pool.mamba_pool.copy_from(src_index, dst_index)

    # Optimize: use torch.cat only if needed
    if full_value:
        value = torch.cat(full_value) if len(full_value) > 1 else full_value[0]
    else:
        value = torch.empty((0,), dtype=torch.int64, device=self.device)

    return MatchResult(
        device_indices=value,
        last_device_node=full_last_node,
        last_host_node=full_last_node,
    )

def _match_prefix_helper(
    self, key: RadixKey
) -> Tuple[List[torch.Tensor], TreeNode, TreeNode]:
    """
    Mamba prefix matching helper. Returns:
    - full_value: all matched full cache tokens (including through tombstone nodes)
    - full_last_node: last node with full cache match
    - mamba_last_node: last node with mamba_value (for COW)
    """
    node = self.root_node
    child_key = self.get_child_key_fn(key)
    value = []
    mamba_last_node = node  # Track last node with mamba state

        if prefix_len < len(child.key):
            new_node = self._split_node(child.key, child, prefix_len)
            value.append(new_node.value)
            node = new_node
            # Update mamba_last_node if new_node has mamba state
            if new_node.mamba_value is not None:
                mamba_last_node = new_node
            break
        else:
            value.append(child.value)
            # Update mamba_last_node before moving to child
            if child.mamba_value is not None:
                mamba_last_node = child
            node = child
            key = key[prefix_len:]
            if len(key):
                child_key = self.get_child_key_fn(key)

    # update time for matched nodes - single pass
    self.full_lru_list.reset_node_and_parents_mru(node, self.root_node)
    if mamba_last_node != self.root_node and mamba_last_node.mamba_value is not None:
        self.mamba_lru_list.reset_node_and_parents_mru(mamba_last_node, self.root_node)

    return value, node, mamba_last_node
```

TEST-1:

results:
main, disable radix:
input_throughput: 18518

mr+THIS FIX, enable radix:
input_throughput: 16174 up to 17993

TEST-2:

results:
main, disable radix:
input_throughput: 45264

mr+THIS FIX, enable radix:
input_throughput:  39668 up to 298052  !!!!!!!

hanming-lu:
@Swipe4057 let's merge this PR to add mamba radix cache functionality. welcome to make changes to improve performance!

hzh0425 (approved):

zhyncs (approved):

lisp2025:
/data/install/backup/sglang_main/python/sglang/srt/mem_cache/memory_pool.py", line 306, in alloc
    mid = self.mamba_pool.alloc(1)[0]
TypeError: 'NoneType' object is not subscriptable

on 20 requests,--max-mamba-cache-size 100000

Chen-0210:
@yizhang2077 Hi Is there a Slack channel for the mamba radix cache discussion?

PR #11216, opened 2025-10-04
Tiny add sanity check for torch empty tensors

fzyzcjy:
## Motivation

TODO: test

## Checklist


python/sglang/srt/managers/scheduler.py
```diff
@@ -367,6 +368,9 @@ def __init__(
                 context, zmq.PUSH, port_args.metrics_ipc_name, False
             )
 
+        if get_bool_env_var("SGLANG_SANITY_CHECK_TORCH_EMPTY"):
```
hnyls2002, line 371: BTW, you can use `sglang.srt.environ`

PR #11221, opened 2025-10-04
[ perf ] Replace json-> orjson in hot path

vincentzed:
## Accuracy Tests

```bash
# launch server with gptoss
# test with client.responses.create(...)
```
This is 1. model load config + response api + server arg (already check at launch).

## Checklist

### DIFF
--- a/python/sglang/srt/entrypoints/openai/serving_responses.py
+++ b/python/sglang/srt/entrypoints/openai/serving_responses.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 
 import asyncio
 import copy
-import json
 import logging
 import time
 from contextlib import AsyncExitStack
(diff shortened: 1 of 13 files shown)


JustinTong0323:
Check CI plz~

vincentzed:
I added more that actually happen >1 times, but didn't test all with UT.

python/sglang/srt/entrypoints/openai/serving_responses.py
```diff
@@ -1021,7 +1021,7 @@ def _send_event(event):
                 ):
                     function_name = previous_item.recipient[len("browser.") :]
                     action = None
-                    parsed_args = json.loads(previous_item.content[0].text)
+                    parsed_args = ororjson.loads(previous_item.content[0].text)
```
hnyls2002, line 1024: @vincentzed You should be more careful...

PR #11229, opened 2025-10-05
Update DeepGEMM repository tag to specific commit

merrymercy:


FlamingoPg (approved):

zaristei:
With respect to this change, was this the commit that this tag was pointing to prior? or is this a new commit being upgraded to?

FlamingoPg:
> With respect to this change, was this the commit that this tag was pointing to prior? or is this a new commit being upgraded to?

Just change tag from sgl to specific commit hash for stable. Never mind

PR #11231, opened 2025-10-05
EAGLE cache fix for SWARadixCache

ispobock:
## Motivation
follow-up of #10846

## Accept Length Test
start server with [this line](https://github.com/sgl-project/sglang/blob/8e66d87f0a931c1b14a1f8c36fb82bd04e98b0a4/python/sglang/srt/server_args.py#L774) commented:
```
python3 -m sglang.launch_server --model openai/gpt-oss-20b --attention-backend triton --speculative-algorithm EAGLE3 --speculative-draft-model-path zhuyksir/EAGLE3-gpt-oss-20b-bf16 --speculative-num-steps 2 --speculative-eagle-topk 1 --speculative-num-draft-tokens 3
```
run requests:
```
# requests to construct prefix tree
python3 -m sglang.test.send_one --prompt "You xxx" --max-new-tokens 3
python3 -m sglang.test.send_one --prompt "You are xxx" --max-new-tokens 3
python3 -m sglang.test.send_one --prompt "You are a xxx" --max-new-tokens 3
python3 -m sglang.test.send_one --prompt "You are a helpful xxx" --max-new-tokens 3
python3 -m sglang.test.send_one --prompt "You are a helpful, xxx" --max-new-tokens 3
python3 -m sglang.test.send_one --prompt "You are a helpful, respectful xxx" --max-new-tokens 3
python3 -m sglang.test.send_one --prompt "You are a helpful, respectful and xxx" --max-new-tokens 3
python3 -m sglang.test.send_one --prompt "You are a helpful, respectful and honest xxx" --max-new-tokens 3
python3 -m sglang.test.send_one --prompt "You are a helpful, respectful and honest assistant xxx" --max-new-tokens 3

# test request
python3 -m sglang.test.send_one --prompt "You are a helpful, respectful and honest assistant." --max-new-tokens 20
```
This PR w/ radix cache:
```
acc_length=1.33
speed=217.98 token/s
```
main w/ radix cache:
```
acc_length=1.18
speed=189.46 token/s
```
main w/o radix cache:
```
acc_length=1.33
speed=219.72 token/s
```


hanming-lu (approved):
On the high level makes sense. Finding it not straightforward to fully understand the +1 and -1 logics and their implications. Was there a doc for the radix cache changes?

python/sglang/srt/mem_cache/swa_radix_cache.py
```diff
@@ -422,25 +439,38 @@ def cache_finished_req(self, req: Req) -> None:
             return
 
         token_ids = (req.origin_input_ids + req.output_ids)[:-1]
+        all_token_len = len(token_ids)
+        actual_kv_len = all_token_len - 1 if self.is_eagle else all_token_len
```
hanming-lu, line 445: can we have more comments on the reason behind this -1?

python/sglang/srt/mem_cache/swa_radix_cache.py
```diff
@@ -422,25 +439,38 @@ def cache_finished_req(self, req: Req) -> None:
...
+                self.token_to_kv_pool_allocator.free(kv_indices[page_aligned_len:])
+
+        page_aligned_token_len = (
+            page_aligned_len + 1 if self.is_eagle else page_aligned_len
+        )
+
+        old_prefix_len = len(req.prefix_indices)
+        if self.is_eagle and old_prefix_len > req.last_matched_prefix_len:
+            # prefix_indices attached partial part (for page_size > 1) and one unmatched token (for EAGLE)
+            old_prefix_len -= 1
```
hanming-lu, line 468: Wishing for more comments here 🙏 The +1 and -1 logics and their implications are not straightforward to fully understand :(

python/sglang/srt/mem_cache/swa_radix_cache.py
ispobock: If we convert the key to bigram key, e.g. `[1,2,3,4] -> [(1,2), (2,3), (3,4)]`, the length will -1. (`len([(1,2), (2,3), (3,4)]) = len([1,2,3,4]) - 1`)
So for the corresponding kv length should also -1. Then we get the `actual_kv_len`, and use it to do later calculation and slicing.

python/sglang/srt/mem_cache/swa_radix_cache.py
ispobock: In the chunked prefill case, the chunked kv should be cached in the Radix cache. But in EAGLE case, the last token will not be inserted into the tree due to the shorter length of bigram key. But we still add it to `req.prefix_indices` ([ref](https://github.com/sgl-project/sglang/pull/11231/files#diff-360fa1b861122a508fd460b2bd865be9d248272b25fee7e444d795410fc2b302R551)), since the kv is still in the sequence. Here we do `old_prefix_len - 1` to just make sure the additional kv should be freed correctly, or we will get the memory leak.

PR #11238, opened 2025-10-06
reverse the amd ci test back to 1200s and split the 8-gpu deepseek job into two.

sunxxuns:
## Checklist


test/srt/run_suite.py
```diff
@@ -265,8 +266,11 @@ class TestFile:
     "per-commit-4-gpu-amd": [
         TestFile("test_pp_single_node.py", 150),
     ],
-    "per-commit-8-gpu-amd": [
-        TestFile("test_full_deepseek_v3.py", 250),
+    "per-commit-8-gpu-amd-basic": [
+        TestFile("test_deepseek_v3_basic.py", 150),
+    ],
+    "per-commit-8-gpu-amd-mtp": [
+        TestFile("test_deepseek_v3_mtp.py", 150),
```
merrymercy: We do not need to split them into two suite. Just with two files should be good, because the timeout is per file.

.github/workflows/pr-test-amd.yml
```diff
@@ -290,7 +290,7 @@ jobs:
         run: |
           bash scripts/ci/amd_ci_exec.sh python3 run_suite.py --suite per-commit-2-gpu-amd
 
-  unit-test-backend-8-gpu-amd:
+  unit-test-backend-8-gpu-amd-basic:
```
merrymercy: do not need to split into two jobs

test/srt/test_deepseek_v3_basic.py
```diff
@@ -0,0 +1,77 @@
+import unittest
```
merrymercy, line 1: Splitting into two files is good!

merrymercy (approved):

PR #11264, opened 2025-10-06
[CI] improve disaggregation CI.

hnyls2002:
For better running CI locally.


ShangmingCai (commented):
What about the decode instances' ib devices like  "mlx5_roce2,mlx5_roce3": 
```python
def start_decode(cls):
        decode_args = [
            "--trust-remote-code",
            "--disaggregation-mode",
            "decode",
            "--tp",
            "2",
            "--base-gpu-id",
            "2",
            "--disaggregation-ib-device",
            "mlx5_roce2,mlx5_roce3",
        ]
```
I can make a util to create a map for specific gpu index.

PR #11268, opened 2025-10-06
docker: add manifest to versioned docker releases

ishandhanani:
Goal is to run `docker pull lmsysorg/sglang:v0.5.3` and get the proper image corresponding to your machine. Also added in a `latest` as well @zhyncs

### DIFF
--- a/.github/workflows/release-docker.yml
+++ b/.github/workflows/release-docker.yml
@@ -47,32 +47,17 @@ jobs:
       - name: Build and Push AMD64
         run: |
           version=$(cat python/sglang/version.py | cut -d'"' -f2)
+          tag=v${version}-cu129-amd64
 
-          if [ "${{ matrix.variant.cuda_version }}" = "12.6.1" ]; then
-            cuda_tag="cu126"
-          elif [ "${{ matrix.variant.cuda_version }}" = "12.8.1" ]; then
-            cuda_tag="cu128"
-          elif [ "${{ matrix.variant.cuda_version }}" = "12.9.1" ]; then
-            cuda_tag="cu129"
-          else
-            echo "Unsupported CUDA version"
-            exit 1
-          fi
-
-          tag=v${version}-${cuda_tag}
-
-          if [ "${{ matrix.variant.build_type }}" = "all" ]; then
-            tag_suffix=""
-          else
-            echo "Unsupported build type"
-            exit 1
-          fi
-
-          if [ "${{ matrix.variant.cuda_version }}" = "12.9.1" ]; then
-            docker buildx build --platform linux/amd64 --push -f docker/Dockerfile --build-arg CUDA_VERSION=${{ matrix.variant.cuda_version }} --build-arg BUILD_TYPE=${{ matrix.variant.build_type }} -t lmsysorg/sglang:${tag}${tag_suffix} -t lmsysorg/sglang:latest --no-cache .
-          else
-            docker buildx build --platform linux/amd64 --push -f docker/Dockerfile --build-arg CUDA_VERSION=${{ matrix.variant.cuda_version }} --build-arg BUILD_TYPE=${{ matrix.variant.build_type }} -t lmsysorg/sglang:${tag}${tag_suffix} --no-cache .
-          fi
+          docker buildx build \
+            --platform linux/amd64 \
+            --push \
+            -f docker/Dockerfile \
+            --build-arg CUDA_VERSION=${{ matrix.variant.cuda_version }} \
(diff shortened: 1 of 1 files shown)


kyleliang-nv:
Can you also fix https://github.com/sgl-project/sglang/blob/main/docker/Dockerfile#L102 to add another OR condition for `all_aarch64`.
This line (https://github.com/sgl-project/sglang/blob/main/docker/Dockerfile#L115) should also be changed to 
check BUILD_TYPE for either `blackwell`, `blackwell_aarch64`, or `all_aarch64`

kyleliang-nv:
Another issue is this (https://github.com/sgl-project/sglang/blob/main/docker/Dockerfile#L143-L149) will fail if CUDA_VERISON=12.6.1.
You will want to set env `FLASH_MLA_DISABLE_SM100=1` for CUDA_VERSION less than 12.9.1, otherwise will get error like this.
```
#22 4.375 Processing /sgl-workspace/flash-mla
#22 4.375   Preparing metadata (setup.py): started
#22 4.375   Running command python setup.py egg_info
#22 4.833   /usr/local/lib/python3.12/dist-packages/torch/cuda/__init__.py:63: FutureWarning: The pynvml package is deprecated. Please install nvidia-ml-py instead. If you did not install pynvml directly, please report this to the maintainers of the package that installed pynvml for you.
#22 4.833     import pynvml  # type: ignore[import]
#22 5.573   W1006 05:49:58.407000 114 torch/utils/cpp_extension.py:118] No CUDA runtime is found, using CUDA_HOME='/usr/local/cuda'
#22 5.612   Traceback (most recent call last):
#22 5.612     File "<string>", line 2, in <module>
#22 5.612     File "<pip-setuptools-caller>", line 35, in <module>
#22 5.612     File "/sgl-workspace/flash-mla/setup.py", line 93, in <module>
#22 5.612       ] + get_features_args() + get_arch_flags() + get_nvcc_thread_args(),
#22 5.612                                 ^^^^^^^^^^^^^^^^
#22 5.612     File "/sgl-workspace/flash-mla/setup.py", line 39, in get_arch_flags
#22 5.613       assert DISABLE_SM100, "sm100 compilation for Flash MLA requires NVCC 12.9 or higher. Please set FLASH_MLA_DISABLE_SM100=1 to disable sm100 compilation, or update your environment."
#22 5.613              ^^^^^^^^^^^^^
#22 5.613   AssertionError: sm100 compilation for Flash MLA requires NVCC 12.9 or higher. Please set FLASH_MLA_DISABLE_SM100=1 to disable sm100 compilation, or update your environment.
```

.github/workflows/release-docker.yml
```diff
@@ -102,15 +87,48 @@ jobs:
...
-          tag=v${version}-${cuda_tag}
-          tag_suffix="-gb200"
+      - name: Create multi-arch manifests
+        run: |
+          version=$(cat python/sglang/version.py | cut -d'"' -f2)
 
-          docker buildx build --platform linux/arm64 --push -f docker/Dockerfile --build-arg CUDA_VERSION=${{ matrix.variant.cuda_version }} --build-arg BUILD_TYPE=${{ matrix.variant.build_type }} -t lmsysorg/sglang:${tag}${tag_suffix} --no-cache .
+          # Create versioned manifest
+          docker buildx imagetools create \
+            -t lmsysorg/sglang:v${version}-cu129 \
```
zhyncs: we don't need `-cu129` suffix

merrymercy (approved):

.github/workflows/release-docker.yml
merrymercy: maybe still keep it somewhere?
Later we might have -cu129,  -cu130 , which cannot be unified.

but for `lmsys/sglang:latest`, we can remove it and have a default cuda verison

zhyncs: e.g.
`lmsysorg/sglang:v0.5.3` for cu129
`lmsysorg/sglang:v0.5.3-cu130` for cu130

how about this

zhyncs:
```suggestion
            -t lmsysorg/sglang:v${version} \
```

ishandhanani: This makes sense. I think default for almost all users will be `cu129` therefore we don't need it. As we push on `cu130` lets keep it tagged as such

zhyncs: I am actually worried that users may be troubled, for example, this image can also be used on cu126/cu128/cu129 hopper, not just cu129.

PR #11269, opened 2025-10-06
bump sgl-kernel version to 3.14 for Dockerfile.b300

lpc0220:
SGLang pyproject.toml uses sgl-kernel==0.3.10, which will invalidate the 3.14 sgl-kernel compilation.

SGLang's pyproject.toml expects sgl-kernel==0.3.10
The compiled sgl-kernel is version 3.14.
This version mismatch will cause import/compatibility errors


zhyncs:
https://github.com/sgl-project/sglang/blob/4aeb193fbd12e25b21a9072432e8a8333c50513e/python/pyproject.toml#L56

PR #11295, opened 2025-10-07
Simplify multi-tokenizer

### LINKED ISSUE #11172: [Feature] Use `rid` to route requests in multi-http-worker mode.
### Checklist

### Motivation

After #10133 has been introduced, every request can have common fields, e.g., `rid`. So now we can remove the unly `MultiTokenizerWrapper` and uses a mapping from `rid` to ipc name to handle the request's routing/collecting without changing any logic inside scheduler/communicator.

### Related resources

_No response_

zhengkezhou1:
## Motivation

fix #11172

## Modification

This PR introduces `http_worker_ipc` field for `BaseReq` and `http_worker_ipcs` for `BaseBatchReq`.

- Former Register request is deprecated, we can directly get ipc name from each requests.
- Neither `rid` nor `MultTokenizerWrapper` carries ipc information.


python/sglang/srt/managers/io_struct.py
```diff
@@ -1382,7 +1382,8 @@ class MultiTokenizerRegisterReq(BaseBatchReq):
 
 @dataclass
 class MultiTokenizerWrapper:
-    # FIXME(lsyin): remove this
+    # DEPRECATED: This class should be removed after migration to rid-based routing
```
hnyls2002: Deprecate this in this PR

hnyls2002 (changes requested):

whybeyoung:
LGTM

zhengkezhou1:
Some tests failed, and I don't know why. So, I'm going to increase the timeout to try to figure it out

hnyls2002:
@zhengkezhou1 The CI times out because we still need something like `Wrapper`. You only handled the genration requets, but ignored other communicator requests. Try your latest modification with an `/get_server_info` request then you can reproduce the timeout.

PR #11304, opened 2025-10-07
[oai serving chat] Add argument `--sampling-defaults` and fix `ChatCompletionRequest` defaults

CatherineSue:
## Motivation

Currently, SGLang's `ChatCompletionRequest` uses `tempature=0.7` as a default. That is inconsistent with OpenAI API's default value, which is 1.0.

1.0 may be too high for generations and many models have a recommended sampling parameter setting in their `generation_config.json`.

This PR adds support for using a model's recommended default sampling parameters by passing `--sampling-defaults model`.

## Modifications

- Fixed `temperature` default value in `ChatCompletionRequest`
  - Set all sampling parameters field to `None`
  - Created `_DEFAULT_SAMPLING_PARAMS` in `ChatCompletionRequest`, similar to `ResponsesRequest`
- Added `--sampling-defaults` CLI parameter
  - Values: `"default"` (SGLang/OpenAI defaults) or `"model"` (use model's generation_config.json)
  - Updated value `default` to `openai` in fab44da
  - Default: `"model"`
- Added `get_default_sampling_params` in `ModelConfig` to get the sampling params based on `server_args.sampling_defaults`
- Move `_build_sampling_params()` to `ChatCompletionRequest.to_sampling_params()` to implement proper priority: **User explicit values > Model generation config > SGLang/OpenAI defaults**

### Default sampling parameters

```
--model /models/meta-llama/Llama-4-Maverick-17B-128E-Instruct-FP8 \
--port=8080 --tp-size=8 --context-length=20000 \
--attention-backend=fa3 --enable-multimodal \
--tool-call-parser=pythonic \
--log-requests --log-requests-level=1 \
--sampling-defaults=openai
```

#### Request

```
response_non_stream = client.chat.completions.create(
    model=model_name,
    messages=messages,
    stream=False,  # Non-streaming
    max_tokens=2000,
)
```

#### Log

Screenshot 2025-10-07 at 9 32 18 AM

### Model sampling parameters

```
--model /models/meta-llama/Llama-4-Maverick-17B-128E-Instruct-FP8 \
--port=8080 --tp-size=8 --context-length=20000 \
--attention-backend=fa3 --enable-multimodal \
--tool-call-parser=pythonic \
--log-requests --log-requests-level=1 \
--sampling-defaults=model
```

#### Request 1

Same as above

#### Log

The `Using default chat sampling` log appears twice because `OpenAIServingResponse` inherits from `OpenAIServingChat`

Screenshot 2025-10-07 at 9 17 51 AM

#### Request 2: pass `top_k=-1` in request 

Screenshot 2025-10-07 at 12 44 22 PM

## Checklist

### DIFF
--- a/python/sglang/srt/entrypoints/openai/serving_chat.py
+++ b/python/sglang/srt/entrypoints/openai/serving_chat.py
@@ -66,40 +66,6 @@ class OpenAIServingChat(OpenAIServingBase):
         self.tool_call_parser = self.tokenizer_manager.server_args.tool_call_parser
         self.reasoning_parser = self.tokenizer_manager.server_args.reasoning_parser
 
-        # Get default sampling parameters from model's generation config
-        self.default_sampling_params = (
-            self.tokenizer_manager.model_config.get_default_sampling_params()
-        )
-        if self.default_sampling_params:
-            logger.info(
-                f"Using default chat sampling params from model generation config: {self.default_sampling_params}",
-            )
-            self._override_chat_protocol_defaults()
-
-    def _override_chat_protocol_defaults(self) -> None:
-        """Override protocol-level defaults with model-specific sampling defaults."""
-        overrides_applied = {}
-        for param_name, param_default in self.default_sampling_params.items():
-            if param_default is None:
-                continue
-            field_info = ChatCompletionRequest.model_fields.get(param_name)
-            if field_info is None or field_info.is_required():
-                continue
-            current_default = field_info.default
-            if current_default == param_default:
-                continue
-            field_info.default = param_default
-            overrides_applied[param_name] = current_default
-        if overrides_applied:
-            ChatCompletionRequest.model_rebuild(force=True)
-            logger.debug(
-                "Overrode chat protocol defaults with model defaults: %s",
-                {
-                    name: {"previous": previous, "current": ChatCompletionRequest.model_fields[name].default}
-                    for name, previous in overrides_applied.items()
-                },
-            )
-
     def _request_id_prefix(self) -> str:
         return "chatcmpl-"
 
--- a/python/sglang/srt/entrypoints/openai/serving_chat.py
+++ b/python/sglang/srt/entrypoints/openai/serving_chat.py
@@ -74,31 +74,6 @@ class OpenAIServingChat(OpenAIServingBase):
             logger.info(
                 f"Using default chat sampling params from model generation config: {self.default_sampling_params}",
             )
-            self._override_chat_protocol_defaults()
-
-    def _override_chat_protocol_defaults(self) -> None:
-        """Override protocol-level defaults with model-specific sampling defaults."""
-        overrides_applied = {}
-        for param_name, param_default in self.default_sampling_params.items():
-            if param_default is None:
-                continue
-            field_info = ChatCompletionRequest.model_fields.get(param_name)
-            if field_info is None or field_info.is_required():
-                continue
-            current_default = field_info.default
-            if current_default == param_default:
-                continue
-            field_info.default = param_default
-            overrides_applied[param_name] = current_default
-        if overrides_applied:
-            ChatCompletionRequest.model_rebuild(force=True)
-            logger.debug(
-                "Overrode chat protocol defaults with model defaults: %s",
-                {
-                    name: {"previous": previous, "current": ChatCompletionRequest.model_fields[name].default}
-                    for name, previous in overrides_applied.items()
-                },
-            )
 
     def _request_id_prefix(self) -> str:
         return "chatcmpl-"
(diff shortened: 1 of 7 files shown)


python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -418,18 +427,39 @@ def _build_sampling_params(
...
 
+        # Map of request parameters to their Pydantic defaults
+        # If request value == default, we'll use generation config instead
+        request_defaults = {
+            "temperature": 1.0,
+            "top_p": 1.0,
+            "top_k": -1,
+            "min_p": 0.0,
+            "repetition_penalty": 1.0,
+        }
```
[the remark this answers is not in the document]
CatherineSue: Good suggestion, applied.

JustinTong0323:
It would change the behavior for all models, we still need to use the models' generation config by default I think.

python/sglang/srt/server_args.py
```diff
@@ -1866,6 +1867,15 @@ def add_cli_args(parser: argparse.ArgumentParser):
             default=ServerArgs.tool_call_parser,
             help=f"Specify the parser for handling tool-call interactions. Options include: {tool_call_parser_choices}.",
         )
+        parser.add_argument(
+            "--sampling-defaults",
+            type=str,
+            choices=["default", "model"],
```
JustinTong0323: we shall better change to "openai" and "model". and default to "model"

CatherineSue: Done

python/sglang/srt/server_args.py
```diff
@@ -1866,6 +1867,15 @@ def add_cli_args(parser: argparse.ArgumentParser):
...
             help=f"Specify the parser for handling tool-call interactions. Options include: {tool_call_parser_choices}.",
         )
+        parser.add_argument(
+            "--sampling-defaults",
+            type=str,
+            choices=["openai", "model"],
+            default=ServerArgs.sampling_defaults,
+            help="Where to get default sampling parameters. "
+            "'openai' uses SGLang/OpenAI defaults (temperature=1.0, top_p=1.0, etc.). "
+            "'model' uses the model's generation_config.json if available.",
```
JustinTong0323: Shall we add explanation here "default to model's own default sampling params"?

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -418,18 +427,35 @@ def _build_sampling_params(
...
+
+        # Helper to get parameter value with generation config fallback
+        def get_param(request_value, param_name):
+            if param_name in request_defaults:
+                if request_value != request_defaults[param_name]:
+                    return request_value
+                if param_name in self.default_sampling_params:
+                    return self.default_sampling_params[param_name]
+            return request_value
```
JustinTong0323: Seems not necessary? Could we just do override like:
```
foreach key:value in self.default_sampling_params:
    request.key = value
```

CatherineSue: If the parameter is set by the user, we should respect user's value first. We need request_defaults to know if user has set different values.

python/sglang/srt/server_args.py
CatherineSue: Added.

python/sglang/srt/entrypoints/openai/serving_chat.py
JustinTong0323: this way is a little bit not elegant to me, could we override the protocol schema default with the models' default?

CatherineSue: No. That is more difficult to do in pydantic. An easier way would be to set all above fields to None in ChatCompletionRequest. Then we override if the value is None, which means a user hasn't passed any value.

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -66,6 +66,15 @@ def __init__(
...
 
+        # Get default sampling parameters from model's generation config
+        self.default_sampling_params = (
+            self.tokenizer_manager.model_config.get_default_sampling_params()
+        )
+        if self.default_sampling_params:
+            logger.info(
+                f"Using default chat sampling params from model generation config: {self.default_sampling_params}",
+            )
+
```
JustinTong0323, line 76:
```suggestion
        if self.default_sampling_params:
            logger.info(
                f"Using default chat sampling params from model generation config: {self.default_sampling_params}",
            )
            self._override_chat_protocol_defaults()

    def _override_chat_protocol_defaults(self) -> None:
        """Override protocol-level defaults with model-specific sampling defaults."""

        overrides_applied = {}
        for param_name, param_default in self.default_sampling_params.items():
            if param_default is None:
                continue
            field_info = ChatCompletionRequest.model_fields.get(param_name)
            if field_info is None or field_info.is_required():
                continue
            current_default = field_info.default
            if current_default == param_default:
                continue
            field_info.default = param_default
            overrides_applied[param_name] = current_default

        if overrides_applied:
            ChatCompletionRequest.model_rebuild(force=True)
            logger.debug(
                "Overrode chat protocol defaults with model defaults: %s",
                {
                    name: {"previous": previous, "current": ChatCompletionRequest.model_fields[name].default}
                    for name, previous in overrides_applied.items()
                },
            )
```

JustinTong0323: Maybe this could help? (written by gpt) But I am not sure whether we should log the change

CatherineSue: I am unfamiliar with pydantic's model rebuilds. If this is a one-time change, feel free to apply and revert 91f10f6.

JustinTong0323: Seems not effective, debugging...

CatherineSue: I tried to apply it in b26395d, it doesn't seem to work:
Screenshot 2025-10-07 at 12 07 16 PM

CatherineSue: I did some digging, FastAPI caches the Pydantic model schemas at warmup, it won't catch the overrode here.

JustinTong0323: Yes... Sorry for misleading...

JustinTong0323: Seems hard to change here, just revert to the original solution, sorry,,,

CatherineSue: Tried it at the init stage of FastAPI startup, still doesn't work.

CatherineSue: No worries, it is good to learn about how pydantic model works

JustinTong0323 (approved):
Awesome

JustinTong0323:
```
======================================================================
ERROR: test_sampling_param_build (__main__.ServingChatTestCase)
  File "/public_sglang_ci/runner-l1c-gpu-1/_work/sglang/sglang/test/srt/openai_server/basic/test_serving_chat.py", line 196, in test_sampling_param_build
    params = self.chat._build_sampling_params(req, ["</s>"], None)
AttributeError: 'OpenAIServingChat' object has no attribute '_build_sampling_params'. Did you mean: 'default_sampling_params'?
```
We need to update the tests accordingly.

CatherineSue:
UT
Screenshot 2025-10-07 at 3 37 25 PM

voipmonitor:
@CatherineSue please check https://github.com/sgl-project/sglang/issues/11574 this PR caused chaning top_k to the 50 from -1 which might be wrong.

adarshxs:
https://huggingface.co/docs/transformers/main/en/main_classes/text_generation#transformers.GenerationConfig.top_k

@CatherineSue for models like [glm 4.5](https://huggingface.co/zai-org/GLM-4.5-Air-FP8/blob/main/generation_config.json), the top_k is [calculated](https://github.com/sgl-project/sglang/blob/27ef1459e6f47d348057031dde6740d94acf7df5/python/sglang/srt/models/glm4_moe.py#L418) rather than set in the config. top_k defaults to 50 in this case and not even -1

PR #11309, opened 2025-10-07
[DeepSeek-V3.2] Include indexer kv cache when estimating kv cache size

trevor-m:
## Motivation

For DS v3.2 to avoid OOM.

## Results

On 8xB200:
With this PR:
```
$ python -m sglang.launch_server --model deepseek-ai/DeepSeek-V3.2-Exp --tp 8 --dp 8 --enable-dp-attention
...                                                                                    
[2025-10-07 23:17:19 DP0 TP0] KV Cache is allocated. #tokens: 975424, KV size: 43.67 GB                                                                                         
[2025-10-07 23:17:19 DP0 TP0] Memory pool end. avail mem=37.71 GB  
```

With main (hardcoded mem-fraction-static=0.8) - KV size including index is actually 65.49 GB
```
[2025-10-07 22:29:11 DP0 TP0] KV Cache is allocated. #tokens: 1462784, KV size: 54.52 GB
[2025-10-07 22:29:11 DP0 TP0] Memory pool end. avail mem=15.38 GB
```

## Modifications

* Include indexer kv cache size in `profile_max_num_token`. I used the size calculation in `NSATokenToKVPool.__init__()` - can someone confirm if this is correct? It seems to be working.
* Include indexer kv cache size in allocation log

## Checklist

### DIFF
--- a/python/sglang/srt/model_executor/model_runner.py
+++ b/python/sglang/srt/model_executor/model_runner.py
@@ -1280,6 +1280,17 @@ class ModelRunner:
                 * num_layers
                 * torch._utils._element_size(self.kv_cache_dtype)
             )
+            # Add indexer KV cache overhead for NSA models (DeepSeek V3.2)
+            if is_deepseek_nsa(self.model_config.hf_config):
+                index_head_dim = get_nsa_index_head_dim(self.model_config.hf_config)
+                indexer_size_per_token = (
+                    index_head_dim
+                    + index_head_dim // NSATokenToKVPool.quant_block_size * 4
+                )
+                element_size = torch._utils._element_size(
+                    NSATokenToKVPool.index_k_with_scale_buffer_dtype
+                )
+                cell_size += indexer_size_per_token * num_layers * element_size
         else:
             cell_size = (
                 self.model_config.get_num_kv_heads(get_attention_tp_size())
--- a/python/sglang/srt/mem_cache/memory_pool.py
+++ b/python/sglang/srt/mem_cache/memory_pool.py
@@ -1177,7 +1177,9 @@ class MLATokenToKVPool(KVCache):
             dtype=torch.uint64,
             device=self.device,
         )
-        self._finalize_allocation_log(size)
+        if not use_nsa:
+            # NSA will allocate indexer KV cache later and then log the total size
+            self._finalize_allocation_log(size)
 
     def get_kv_size_bytes(self):
         assert hasattr(self, "kv_buffer")
@@ -1298,6 +1300,9 @@ class MLATokenToKVPool(KVCache):
 
 
 class NSATokenToKVPool(MLATokenToKVPool):
+    quant_block_size = 128
+    index_k_with_scale_buffer_dtype = torch.uint8
+
     def __init__(
         self,
         size: int,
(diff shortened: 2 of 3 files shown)


python/sglang/srt/model_executor/model_runner.py
```diff
@@ -1277,6 +1277,14 @@ def profile_max_num_token(self, total_gpu_memory: int):
                 * num_layers
                 * torch._utils._element_size(self.kv_cache_dtype)
             )
+            # Add indexer KV cache overhead for NSA models (DeepSeek V3.2)
+            if is_deepseek_nsa(self.model_config.hf_config):
+                index_head_dim = get_nsa_index_head_dim(self.model_config.hf_config)
+                quant_block_size = 128
```
Fridge003: Can we first assert that the token_to_kv_pool is a `NSATokenToKVPool` instance, and then get the block_size from `self.token_to_kv_pool`?

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -1277,6 +1277,14 @@ def profile_max_num_token(self, total_gpu_memory: int):
...
+                indexer_bytes_per_token = (
+                    index_head_dim + index_head_dim // quant_block_size * 4
+                )
+                cell_size += indexer_bytes_per_token * num_layers
```
Fridge003: Should be multiplied by the element size of `NSATokenToKVPool.index_k_with_scale_buffer`

python/sglang/srt/model_executor/model_runner.py
trevor-m: This function is called before `token_to_kv_pool` is initialized, so we can't do that.

Fridge003: Or can we add an util function that reads the block size from hf config file?

python/sglang/srt/model_executor/model_runner.py
trevor-m: Thank you for the review!
Seems this value is always uint8 (so would be multiply by 1): https://github.com/sgl-project/sglang/blob/f094e0a490a46c612642e1c8e1d0fc849e04be1c/python/sglang/srt/mem_cache/memory_pool.py#L1268

Since `token_to_kv_pool` is not initialized yet, should be add a static attribute to NSATokenToKVPool?
```
@staticmethod
def get_index_k_with_scale_buffer_dtype():
  return torch.uint8
```
Then we can use that when creating the buffer and also here with `get_index_k_with_scale_buffer_dtype().element_size()`.

Fridge003: Yeah I think that also works

python/sglang/srt/model_executor/model_runner.py
trevor-m: That sounds good, but I don't think I see that in the config file? https://huggingface.co/deepseek-ai/DeepSeek-V3.2-Exp/blob/main/config.json
Is the weight block size here?
```
"quantization_config": {
--
"activation_scheme": "dynamic",
"fmt": "e4m3",
"quant_method": "fp8",
"scale_fmt": "ue8m0",
"weight_block_size": [
128,
128
]
},
```

Fridge003: Or it can also be set from static attribut of KV Pool?

trevor-m: Thanks, done

python/sglang/srt/model_executor/model_runner.py
trevor-m: Thanks, done

Fridge003 (approved):
LGTM

PR #11313, opened 2025-10-08
Separate allocation logic from scheduler

cctry:
## Motivation
Preparation for mem_cache V2. 
The allocation logic is moved to `mem_cache/` from `schedule_batch.py`. Ideally, scheduling code should only interact with `tree_cache` in V1 and `memory_manager` in V2.

## Modifications

- create `mem_cache/common.py`  for allocation functions operating allocator and tree cache
- refactor the allocation to use two wrappers `alloc_for_extend` and `alloc_for_decode`
  - in `prepare_for_decode`, the increment of `seqlen` is moved after `alloc_for_decode` for clarity
  - in `prepare_for_extend`,  some allocation-needed fields are set before `alloc_for_extend`
- for spec decode, the allocation functions are unchanged and they are to be changed after spec V2
- in `bench_one_batch.py`, create a dummy `tree_cache` as the placeholder

## Checklist


merrymercy (approved):

xiezhq-hermann (approved):

python/sglang/srt/mem_cache/common.py
```diff
@@ -0,0 +1,479 @@
...
+    return result
+
+
+def alloc_token_slots(
+    tree_cache: BasePrefixCache,
+    num_tokens: int,
+    backup_state: bool = False,
+):
+    allocator = tree_cache.token_to_kv_pool_allocator
+    evict_from_tree_cache(tree_cache, num_tokens)
```
xiezhq-hermann, line 206: why evicting proactively here?

cctry: this is actually `evict_if_needed`
```
         if self.token_to_kv_pool_allocator.available_size() < num_tokens:
                if self.tree_cache is not None:
                    self.tree_cache.evict(num_tokens)
```

xiezhq-hermann: maybe we can rename it a bit, it's actually check for availability and evict if needed

cctry: what will be the case that we want to evict nodes regardless of the availability
 
i also feel the eviction policy is something non-trivial so `evict_from_tree_cache` will encapsulate complex logic from upper code

xiezhq-hermann: the current `self.tree_cache.evict ` is indeed to evict nodes to meet the requirement regardless of the availability, and I think we can probably keep eviction policy under the hood too like what we have for now

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -1695,28 +1496,6 @@ def prepare_for_decode(self):
...
 
-        # free memory
-        if isinstance(self.tree_cache, SWAChunkCache):
-            for req in self.reqs:
-                self.tree_cache.evict_swa(
-                    req, req.seqlen - 1, self.model_config.attention_chunk_size
-                )
-
-        # Allocate memory
-        if self.token_to_kv_pool_allocator.page_size == 1:
```
airMeng, line 1706: Hi, I find ```batch,token_to_kv_pool_allocator.page_size``` not always equals to ```batch.tree_cache.page_size```, then the paged config will go to the wrong allocation, breaks a lot of cases

python/sglang/bench_one_batch.py
```diff
@@ -257,11 +258,18 @@ def prepare_synthetic_inputs_for_latency_test(
 
 @torch.no_grad
 def extend(reqs, model_runner):
+    # Create dummy tree_cache for benchmarks (no prefix caching, just allocation)
+    dummy_tree_cache = SimpleNamespace(
+        page_size=1,
```
airMeng, line 263: why hard-code to 1 instead ```page_size=model_runner.server_args.page_size``` ?

python/sglang/srt/managers/schedule_batch.py
cctry: could you let me know when the page sizes are different? we can add tests to capture this in the future

python/sglang/bench_one_batch.py
cctry: Thanks for pointing out. wIll fix this

python/sglang/srt/managers/schedule_batch.py
airMeng: https://github.com/sgl-project/sglang/pull/11313#discussion_r2443904068 seems the page sizes should be the same and this is just a bug.

But the question is why we need to access the same page size from different places? echos the suggestions here https://github.com/sgl-project/sglang/pull/11645#issuecomment-3407535221

PR #11318, opened 2025-10-08
Support LoRA in bench_serving oai interface

lifuhuang:
## Motivation

Enable bench_serving to benchmark LoRA inference in SGLang and vLLM.

## Checklist


python/sglang/bench_serving.py
```diff
@@ -209,6 +209,11 @@ async def async_request_openai_completions(
             **request_func_input.extra_request_body,
         }
 
+        # hack to accommodate different LoRA conventions between SGLang and vLLM.
+        if request_func_input.lora_name:
+            payload["model"] = request_func_input.lora_name
```
Fridge003, line 214: I'm not sure, but isn't payload["model"] the base model?

lifuhuang: Yes. In vLLM, lora adapter is selected via model field (as if it was a different base model)

Fridge003 (approved):
LGTM

PR #11331, opened 2025-10-08
Deprecate `global_server_args_dict`

hnyls2002:
Deprecate the `global_server_args_dict`, just use `global_server_args: ServerArgs`.

- Only the scheduler process needs it
- Set the global value in `ModelRunner`
- Only access this global value by `get_global_server_args()` to avoid reading before setting the arguments.


python/sglang/srt/server_args.py
```diff
@@ -3329,6 +3329,9 @@ def adjust_mem_fraction_for_vlm(self, model_config):
         )
 
 
+global_server_args: Optional[ServerArgs] = None
```
fzyzcjy: iirc if we do `from ... import global_server_args`, then if later we do `global_server_args = blahblah`, the imported obj will still be old...

one way may be `def get_global_server_args(): return global_server_args`

ch-wan:
ci is broken: https://github.com/sgl-project/sglang/actions/runs/18446526755/job/52554088309

PR #11346, opened 2025-10-08
Modify SWAChunkCache to support SWA.

giantleap2012:
## Motivation

Add support for SWAChunkCache to support sliding window attention. Currently SWAChunkCache evicts the whole chunk when it reaches the attention_chunk_size. But for SWA, we will need to keep the most recent sliding_window_size tokens.

Also add a server config to schedule aggressively using the full pool when the model architecture has lots of local attention layers and sliding_window_size is much smaller than decoding length.

The changes here are mainly to accommodate the type of traffic
- A model architecture heavily rely on SWA layers, which means the SWA tokens pool will be much smaller than full tokens pool
- Long decoding length which far exceeds sliding_window_size

Please note no changes have been applied to SWARadixCache. As SWARadixCache is designed with concept of maximizing cache hit rate, making changes with evicting KV tokens falls out of sliding window right away will hurt the cache hit rate. I haven't come up a good design for radix cache support on this type of traffic (long decoding length, short sliding window size) yet. Will leave that for now.

## Modifications
Modify the SWAChunkCache to support SWA by keeping the most recent sliding_window_size tokens instead of evicting the whole chunk. If no sliding_window_size is passed in, the behavior is exactly the same as before.

Add a server config so that we can aggressively schedule a high batch size when the hybrid model architecture heavily rely on local attention and the sliding window size is far smaller the decoding length. 

## Checklist

### DIFF
--- a/python/sglang/srt/managers/schedule_policy.py
+++ b/python/sglang/srt/managers/schedule_policy.py
@@ -324,6 +324,7 @@ class PrefillAdder:
         rem_chunk_tokens: Optional[int],
         mixed_with_decode_tokens: int = 0,
         priority_scheduling_preemption_threshold: int = 0,
+        use_hybrid_full_pool: bool = False,
     ):
         self.page_size = page_size
         self.tree_cache = tree_cache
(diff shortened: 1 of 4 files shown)


hnyls2002:
Thanks @giantleap2012! SWAChunkCache SWA support has since landed on main, so I'm closing this as
  superseded. Please reopen if something's missing.

PR #11347, opened 2025-10-08
[GDN] Mamba radix cache ratio support

hanming-lu:
## Motivation
Depends on https://github.com/sgl-project/sglang/pull/11214

When mamba radix cache is enabled, allocate mamba pool size relative to full kv token pool size instead of max running requests.

## Accuracy Tests

```
python3 -m sglang.launch_server --model Qwen/Qwen3-Next-80B-A3B-Instruct --tp 4  --chunked-prefill-size 64 --max-running-requests 16

root@memx-azj-30-sr1:~/oss/sglang# python3 benchmark/gsm8k/bench_sglang.py --num-question 1000
/root/xai/.venv/lib/python3.13/site-packages/torch/cuda/__init__.py:63: FutureWarning: The pynvml package is deprecated. Please install nvidia-ml-py instead. If you did not install pynvml directly, please report this to the maintainers of the package that installed pynvml for you.
  import pynvml  # type: ignore[import]
100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1000/1000 [03:49<00:00,  4.36it/s]
Accuracy: 0.942
Invalid: 0.000
Latency: 230.713 s
Output throughput: 726.426 token/s
```

### DIFF
--- a/python/sglang/srt/model_executor/model_runner.py
+++ b/python/sglang/srt/model_executor/model_runner.py
@@ -358,29 +358,15 @@ class ModelRunner:
             if architectures and not any("Llama4" in arch for arch in architectures):
                 self.is_hybrid = self.model_config.is_hybrid = True
 
-        if config := self.mambaish_config:
-            if self.server_args.max_mamba_cache_size is None:
-                if self.server_args.max_running_requests is not None:
-                    self.server_args.max_mamba_cache_size = (
-                        self.server_args.max_running_requests
-                        * MAMBA_CACHE_SIZE_MAX_RUNNING_REQUESTS_RATIO
-                    )
+        # sets the max_mamba_cache_size based on the max_running_requests for hybrid gdn
+        # and disable radix cache
+        # For hybrid gdn with radix cache, it is set in the init_memory_pool function
+        if self.mambaish_config is not None and server_args.disable_radix_cache:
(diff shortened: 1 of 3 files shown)


python/sglang/srt/model_executor/model_runner.py
```diff
@@ -358,29 +358,15 @@ def initialize(self, min_per_gpu_memory: float):
...
-        if config := self.mambaish_config:
-            if self.server_args.max_mamba_cache_size is None:
-                if self.server_args.max_running_requests is not None:
-                    self.server_args.max_mamba_cache_size = (
-                        self.server_args.max_running_requests
-                        * MAMBA_CACHE_SIZE_MAX_RUNNING_REQUESTS_RATIO
-                    )
+        # sets the max_mamba_cache_size based on the max_running_requests for hybrid gdn
+        # and disable radix cache
+        # For hybrid gdn with radix cache, it is set in the init_memory_pool function
```
yizhang2077: I think we can move this logic into `handle_max_mamba_cache`?

hanming-lu: Done.

Swipe4057:
Is it taken into account that mamba_ssm_dtype can be either "float32" or "bfloat16"?

PR #11351, opened 2025-10-09
Refactor kv cache free

cctry:
## Motivation
Prepration for mem_cache V2.
This PR aims to simpify the kv cache free logic

## Modifications
- `cache_finished_req`
  -   add `is_insert` to skip inserting to tree
  - change the computation of `token_ids` to support preempting a prefill request.
- simplify `release_req` by using `cache_finished_req` with no insert
  - previous `release_req` does not consider the eagle radix tree fix cc @ispobock to confirm
- disable insertion for PD when transfer fails since kv cache is corrupted

## Checklist

### DIFF
--- a/python/sglang/srt/disaggregation/decode.py
+++ b/python/sglang/srt/disaggregation/decode.py
@@ -611,8 +611,8 @@ class DecodeTransferQueue:
                 self.scheduler.stream_output(
                     [decode_req.req], decode_req.req.return_logprob
                 )
-                # unlock the kv cache or it will have memory leak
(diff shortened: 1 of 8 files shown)


ispobock:
> previous release_req does not consider the eagle radix tree fix

Yes, the calculation of last_uncached_pos should be handled correctly in the EAGLE fix, or it may cause memory leak issue. This preempt case is not tested in the previous fix.

merrymercy (approved):

ByronHsu (approved):

PR #11365, opened 2025-10-09
[Bug Fix] prevent lora adapter from being loaded into LoRAManager if it is already loaded

glenliu21:
## Motivation

Currently, when the `LoRAManager` validates a new LoRA adapter, it doesn't check if the LoRA adapter's path has already been loaded. This leads to inconsistency between `LoRAManager` and `LoRARegistry` and also wasted computation.

To demonstrate an example:

1. Load `lora_a` -> both `LoRAManager` and `LoRARegistry` load `lora_a`.
2. Load `lora_a` again -> `LoRAManager` loads `lora_a` and its weights again. `LoRARegistry` rejects it.
3. Unload `lora_a` -> `LoRAManager` unloads one copy of `lora_a`, but still holds the other copy. `LoraRegistry` unloads its copy, and its current state correctly shows that no LoRA adapters are loaded, even though `LoRAManager` has one copy of `lora_a` loaded.

## Modifications

Added a check in `validate_new_adapter` and added to the unit tests in `test_lora_update.py`.

## Checklist


lifuhuang (approved):
LGTM

Fridge003 (approved):
LGTM

PR #11367, opened 2025-10-09
[tool call] Fix prev_tool_call_arr management in base_format_detector.py

CatherineSue:
## Motivation

There is a bug in the current `prev_tool_call_arr` management in `base_format_detector.py`.
When a tool call is complete at L258, `self.current_tool_call_id` increments by 1.

NOTE: the bug itself is not very harmless because it is hard to encounter (or notice).

### 1. Wrong `current_tool_call_id`

https://github.com/sgl-project/sglang/blob/84768d10175cb147da1e54b0eeb1e626ff50c276/python/sglang/srt/function_call/base_format_detector.py#L267-L271

However, at L301, `current_tool_call_id` is used to update in `prev_tool_call_arr`. 

https://github.com/sgl-project/sglang/blob/84768d10175cb147da1e54b0eeb1e626ff50c276/python/sglang/srt/function_call/base_format_detector.py#L301-L306

This would result a tool call that has completed, for instance, 0, be put in `self.prev_tool_call_arr[1]`. See the screenshot below. The current tool call 0 is completed, but the result is saved in `self.prev_tool_call_arr[1]`.

Screenshot 2025-10-08 at 7 47 01 PM

This may cause a defensive check in `_check_for_unstreamed_tool_args`. As at this moment, the length of `prev_tool_call_arr` is inconsistent with `streamed_args_for_tool`

https://github.com/sgl-project/sglang/blob/84768d10175cb147da1e54b0eeb1e626ff50c276/python/sglang/srt/entrypoints/openai/serving_chat.py#L1116-L1119

### 2. Empty `prev_tool_call_arr`

A second issue in the logic is that, when a tool call completes, base_format_detector.py#L267-L271 (see above) tries to clear the `prev_tool_call_arr[current_tool_id]`. However, `prev_tool_call_arr` is supposed to keep all the tool call data, so in `_check_for_unstreamed_tool_args`, it could do: 

https://github.com/sgl-project/sglang/blob/84768d10175cb147da1e54b0eeb1e626ff50c276/python/sglang/srt/entrypoints/openai/serving_chat.py#L1121-L1124

If we clear the tool call when it is complete, this check above will always get `{}`.

### 3. Only update `streamed_args_for_tool` when `is_complete`

Same as the above reason, `streamed_args_for_tool` should be updated as well even when `is_complete=True`

## Modifications

- Always update `self.prev_tool_call_arr[self.current_tool_id]` with `current_tool_call`.
- Doesn't clear `self.prev_tool_call_arr[self.current_tool_id]` when `is_complete`.

## Checklist


CatherineSue:
Need to wait for CI to complete

CatherineSue:
Looking into failed tests

CatherineSue:
function_call module passed. 
Screenshot 2025-10-08 at 11 12 01 PM
Screenshot 2025-10-08 at 11 20 33 PM

python/sglang/srt/function_call/base_format_detector.py
```diff
@@ -293,17 +301,15 @@ def parse_streaming_increment(
...
-                    self.prev_tool_call_arr.append({})
-                self.prev_tool_call_arr[self.current_tool_id] = current_tool_call
+                        # Update streamed_args_for_tool to track what we've sent
+                        # Use completing_tool_id for completed tools (before current_tool_id was incremented)
+                        # Use current_tool_id for ongoing tools
+                        update_tool_id = (
+                            completing_tool_id
+                            if is_current_complete
+                            else self.current_tool_id
+                        )
```
JustinTong0323: This update_tool_id seems the same as tool_index_to_use?

CatherineSue: Fixed

JustinTong0323 (approved):

PR #11368, opened 2025-10-09
Fix batch invariant ops

hebiao064:
## Motivation

Without this change, we will see many `NaN` from certain shape matmul, for example `32, 128 x 128, 1024`

### Problem
The persistent matmul kernel had a critical bug where computation and storage used misaligned tile IDs, causing incorrect results to be written to wrong memory locations, which caused gibberish result for our Grok 10B and v6 Mini model.

### Root Cause
```python
# Buggy: compute tile N, store to tile (N + NUM_SMS)
for tile_id in tl.range(start_pid, num_tiles, NUM_SMS):
    pid_m, pid_n = _compute_pid(tile_id, ...)      # Compute tile N
    # ... accumulate ...
    
    tile_id_c += NUM_SMS
    pid_m, pid_n = _compute_pid(tile_id_c, ...)    # ❌ Store to different tile!
    tl.store(c_ptrs, c, mask=c_mask)
```

### Solution
Use the same `pid_m, pid_n` for both computation and storage:
```python
# Fixed: compute and store to same tile
for tile_id in tl.range(start_pid, num_tiles, NUM_SMS):
    pid_m, pid_n = _compute_pid(tile_id, ...)      # Compute tile N
    # ... accumulate ...
    # Reuse pid_m, pid_n for storage
    tl.store(c_ptrs, c, mask=c_mask)               # ✅ Store to same tile
```

### Additional Changes
- Added explicit `bfloat16` and `float32` dtype handling

## Checklist


test/srt/batch_invariant/test_batch_invariant_ops.py
```diff
@@ -0,0 +1,161 @@
+import math
```
Fridge003, line 2: Maybe we should mark that this test is adapted from `https://github.com/thinking-machines-lab/batch_invariant_ops/blob/main/test_batch_invariance.py`

test/srt/batch_invariant/test_batch_invariant_ops.py
```diff
@@ -0,0 +1,161 @@
...
+        """
+        M, K, N = 32, 128, 1024
+        dtype = torch.float32
+
+        # Run without batch-invariant mode
+        with set_batch_invariant_mode(False):
+            difflist = self._run_multiple_iterations(
+                iters=5, M=M, K=K, N=N, dtype=dtype
+            )
+
```
Fridge003, line 160: Maybe print some differences (either to terminal or github log) when running without batch invariant mode?

hebiao064: addressed

test/srt/batch_invariant/test_batch_invariant_ops.py
hebiao064: done

Fridge003 (approved):
LGTM

PR #11380, opened 2025-10-09
[PD] Adjustable dispatch node count in fake mode for ideal expert distribution

ch-tiger1:
After multiple experiments and investigations, we found that the existing fake mode's expert distribution is relatively evenly distributed in terms of logical experts, but cannot achieve perfect expert distribution in terms of physical expert mapping. To simulate a perfect physical expert distribution, By modifying the following code logic. By default, all tokens are distributed  evenly  to  the current node. The number of nodes to which each batch of tokens is distributed can also be manually adjusted by using the --ep-dispatch-fake-num-node option to control the distribution. This ensures the optimal state and evaluates the maximum performance limit of expert distribution under PD separation. Experiments have shown that in the 4p9d isl-2000/osl-100 scenario, the maximum performance of a single node in the decoding phase can reach 26400 tok/s，expert balance close to 1, and the performance is stable.

### DIFF
--- a/python/sglang/srt/eplb/expert_location_dispatch.py
+++ b/python/sglang/srt/eplb/expert_location_dispatch.py
@@ -12,6 +12,7 @@
 # limitations under the License.
 # ==============================================================================
 
+import random
 from dataclasses import dataclass
 from typing import Literal, Optional
 
@@ -19,6 +20,8 @@ import torch
 
 from sglang.srt.eplb.expert_location import get_global_expert_location_metadata
 from sglang.srt.server_args import get_global_server_args
+
+DUMMY_MAX_TOPK_SHAPE = (256, 8)  # (max batch size, topk)
 
 
 @dataclass
@@ -31,6 +34,7 @@ class ExpertLocationDispatchInfo:
     # (num_logical_experts,)
     partial_logical_to_all_physical_map_num_valid: torch.Tensor
     num_physical_experts: int
+    precomputed_balanced_tensor_map: dict
 
     @classmethod
     def init_new(cls, layer_id: int):
@@ -40,6 +44,24 @@ class ExpertLocationDispatchInfo:
 
         if ep_dispatch_algorithm is None:
             return None
+
+        precomputed_balanced_tensor_map = {}
+        if ep_dispatch_algorithm == "fake":
+            nnodes = get_global_server_args().nnodes
+            node_rank = get_global_server_args().node_rank
+            device = get_global_server_args().device
+            num_physical_experts = expert_location_metadata.num_physical_experts
+            dispatch_node = get_global_server_args().fake_node
+            dummy_topk_ids_shape = DUMMY_MAX_TOPK_SHAPE
+            balanced_tensor = generate_balanced_expert_selection(
+                dummy_topk_ids_shape,
+                node_rank=node_rank,
+                nnodes=nnodes,
+                total_experts=num_physical_experts,
+                dispatch_node=dispatch_node,
+                device=device,
+            )
+            precomputed_balanced_tensor_map[dummy_topk_ids_shape] = balanced_tensor
 
         return cls(
             ep_dispatch_algorithm=ep_dispatch_algorithm,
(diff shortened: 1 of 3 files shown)


python/sglang/srt/eplb/expert_location_dispatch.py
```diff
@@ -107,3 +120,62 @@ def _topk_ids_logical_to_physical_dynamic(
...
+            remaining_list.append(base)
+        repeat.append(base.repeat(tokens_per_expert))
+    selected_experts = torch.cat(repeat)
+
+    if remaining_tokens_per_expert != 0:
+        sampled = random.sample(remaining_list, remaining_tokens_per_expert)
+        random_remaining = torch.cat(sampled)
+        selected_experts = torch.cat([selected_experts, random_remaining])
+
+    return selected_experts.view(topk_ids_shape)
```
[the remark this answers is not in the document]
ch-tiger1: In enable_deepep_moe mode, self.ep_size = self.tp_size, initialization _init_common has been judged assert num_physical_experts % ep_size == 0, Added assert judgment

python/sglang/srt/server_args.py
```diff
@@ -363,6 +363,7 @@ class ServerArgs:
     enable_expert_distribution_metrics: bool = False
     deepep_config: Optional[str] = None
     moe_dense_tp_size: Optional[int] = None
+    fake_node: int = 1
```
fzyzcjy: ep_dispatch_fake_num_node

python/sglang/srt/layers/moe/topk.py
```diff
@@ -806,14 +806,6 @@ def select_experts(
         topk_config.apply_routed_scaling_factor_on_output
     )
 
-    router_logits, correction_bias = (
```
fzyzcjy, line 920: what about keeping both the old and new algorithms

python/sglang/srt/eplb/expert_location_dispatch.py
fzyzcjy, line 1: > By modifying the following code logic. By default, all tokens are distributed to this node.

a bit confused, could you please elaborate your algorithm. it seems we should not distribute data to one single node, but to multi nodes etc.

ch-tiger1: Sorry, the description is a bit ambiguous. This is not a single node, it is the current node. By evenly distributing the number of tokens in each forward batch to each node, the physics expert on the current node only counts his own tokens, dispatch/combine communication overhead is minimized to achieve the optimal state, which can be used to evaluate the upper limit of performance and provide a reference value for tuning. Of course, the number of nodes distributed to the tokens here can also be adjusted through --ep-dispatch-fake-num-node to observe the ideal value across nodes. It can ensure that the expert balance is close to 1, and better evaluate the ideal state.

python/sglang/srt/layers/moe/topk.py
ch-tiger1: It doesn't make much sense

python/sglang/srt/server_args.py
ch-tiger1: done

python/sglang/srt/eplb/expert_location_dispatch.py
fzyzcjy: > the physics expert on the current node only counts his own tokens, communication overhead is minimized to achieve the optimal state

if I understand correctly, you mean each node do not communicate data w/ other nodes (if setting num_node=1)?

that does not seem good :/

ch-tiger1: Why? My understanding is that fake mode is used to simulate a perfect expert distribution, serving as an ideal state for evaluation. The more evenly distributed the experts, the better the performance. While the previous fake mode achieved a high degree of overall balance, it was imbalanced within each forward batch, resulting in suboptimal performance. After the modification, experiments showed that the balance of the expert distribution within each forward batch and across the entire inference process is now close to 1.

PR #11388, opened 2025-10-09
Replace pad with cat for better performance

yuan-luo:
## Motivation

Using concatenation tends to be faster than padding.

image

image

## Checklist

### DIFF
--- a/python/sglang/srt/models/qwen2_vl.py
+++ b/python/sglang/srt/models/qwen2_vl.py
@@ -407,7 +407,7 @@ class Qwen2VisionTransformer(nn.Module):
         cu_seqlens = torch.repeat_interleave(
             grid_thw[:, 1] * grid_thw[:, 2], grid_thw[:, 0]
         ).cumsum(dim=0, dtype=torch.int32)
-        cu_seqlens = F.pad(cu_seqlens, (1, 0), "constant", 0)
+        cu_seqlens = torch.cat([cu_seqlens.new_zeros(1), cu_seqlens])
 
         # transformers
         x = x.unsqueeze(1)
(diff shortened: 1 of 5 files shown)


mickqian (approved):

python/sglang/srt/models/qwen2_vl.py
```diff
@@ -407,7 +407,7 @@ def forward(
         cu_seqlens = torch.repeat_interleave(
             grid_thw[:, 1] * grid_thw[:, 2], grid_thw[:, 0]
         ).cumsum(dim=0, dtype=torch.int32)
-        cu_seqlens = F.pad(cu_seqlens, (1, 0), "constant", 0)
+        cu_seqlens = torch.cat([cu_seqlens.new_zeros(1), cu_seqlens])
```
BBuf, line 410: The code is:

```python
# compute cu_seqlens - move cu_seqlens to GPU and make it int32
        cu_seqlens = torch.cat(
            [
                torch.tensor([0], device=x.device, dtype=torch.int32),
                (grid_thw[:, 0] * grid_thw[:, 1] * grid_thw[:, 2])
                .cumsum(dim=0)
                .to(device=x.device, dtype=torch.int32),
            ]
        )
        cu_seqlens = torch.cat([cu_seqlens.new_zeros(1), cu_seqlens])
```

It'll cause `cu_seqlens = [0, 0, seq_len1, seq_len1+seq_len2, ...]` , is a bug?

yuan-luo: The original version `F.pad(cu_seqlens, (1, 0), "constant", 0)` also prepend a 0, the PR is just a pure functional replacement, doesn't change the logic.

BBuf (approved):
lgtm

fzyzcjy:
qq: curious is it only useful to reduce cpu time, or it is also helpful for gpu time

PR #11393, opened 2025-10-09
Add request type, HTTP status code as Prometheus metric labels

scottjlee:
## Motivation
To analyze SGLang deployed in production, it is useful to break down Prometheus metrics by several key attributes, such as the request type (`/v1/completions`, `/v1/chat/completions`, etc) and the HTTP status code (2xx, 4xx, 5xx). 

In addition, SGLang currently emits the `sglang:num_requests_total` metric, which tracks the number of total requests completed (success or failure); this does _not_ account for the total number of requests received by the engine, such as requests rejected at the validation stage due to bad request parameters. In production settings, it is useful to track the total number of requests received by the engine to monitor the health of the overall system (e.g. looking at the gap between # received requests vs # completed requests), which may also be used by external systems to make routing decisions.

## Modifications
Add the following optional labels to Prometheus metrics:
- `request_type`: Type of request, represented by the endpoint (`/v1/completions`, `/v1/chat/completions`, etc). This is disabled by default, and can be enabled with server launch arg `--metrics-label-request-type`
- `http_status`: 3-digit HTTP status code returned by the engine. This is disabled by default, and can be enabled with server launch arg `--metrics-label-http-status`

In addition, add the `sglang:num_requests_received_total` metric, which tracks the total number of requests received by the engine.

Example output from metrics endpoint with both of these flags enabled (relevant excerpt to avoid really long output). Before getting metrics, I made two successful and one incorrectly formatted requests to both of the `/v1/chat/completions` and `/v1/completions` endpoints. The `sglang:num_requests_total` metric only accounts for the successful requests, while the newly added `sglang:num_requests_received_total` metric tracks both successful and failed requests.

```
curl -s http://localhost:8001/metrics
...
# HELP sglang:prompt_tokens_total Number of prefill tokens processed.
# TYPE sglang:prompt_tokens_total counter
sglang:prompt_tokens_total{http_status="",model_name="meta-llama/Llama-3.1-8B-Instruct",request_type=""} 7.0
sglang:prompt_tokens_total{http_status="",model_name="meta-llama/Llama-3.1-8B-Instruct",request_type="/v1/chat/completions"} 100.0
sglang:prompt_tokens_total{http_status="",model_name="meta-llama/Llama-3.1-8B-Instruct",request_type="/v1/completions"} 32.0
# HELP sglang:generation_tokens_total Number of generation tokens processed.
# TYPE sglang:generation_tokens_total counter
sglang:generation_tokens_total{http_status="",model_name="meta-llama/Llama-3.1-8B-Instruct",request_type=""} 8.0
sglang:generation_tokens_total{http_status="",model_name="meta-llama/Llama-3.1-8B-Instruct",request_type="/v1/chat/completions"} 74.0
sglang:generation_tokens_total{http_status="",model_name="meta-llama/Llama-3.1-8B-Instruct",request_type="/v1/completions"} 200.0
# HELP sglang:num_requests_total Number of requests processed.
# TYPE sglang:num_requests_total counter
sglang:num_requests_total{http_status="",model_name="meta-llama/Llama-3.1-8B-Instruct",request_type=""} 1.0
sglang:num_requests_total{http_status="",model_name="meta-llama/Llama-3.1-8B-Instruct",request_type="/v1/chat/completions"} 2.0
sglang:num_requests_total{http_status="",model_name="meta-llama/Llama-3.1-8B-Instruct",request_type="/v1/completions"} 2.0
# HELP sglang:num_requests_received_total Number of requests received.
# TYPE sglang:num_requests_received_total counter
sglang:num_requests_received_total{http_status="200",model_name="meta-llama/Llama-3.1-8B-Instruct",request_type="/generate"} 1.0
sglang:num_requests_received_total{http_status="200",model_name="meta-llama/Llama-3.1-8B-Instruct",request_type="/v1/chat/completions"} 2.0
sglang:num_requests_received_total{http_status="400",model_name="meta-llama/Llama-3.1-8B-Instruct",request_type="/v1/chat/completions"} 1.0
sglang:num_requests_received_total{http_status="200",model_name="meta-llama/Llama-3.1-8B-Instruct",request_type="/v1/completions"} 2.0
sglang:num_requests_received_total{http_status="400",model_name="meta-llama/Llama-3.1-8B-Instruct",request_type="/v1/completions"} 1.0
# HELP sglang:cached_tokens_total Number of cached prompt tokens.
# TYPE sglang:cached_tokens_total counter
sglang:cached_tokens_total{http_status="",model_name="meta-llama/Llama-3.1-8B-Instruct",request_type="/v1/chat/completions"} 50.0
sglang:cached_tokens_total{http_status="",model_name="meta-llama/Llama-3.1-8B-Instruct",request_type="/v1/completions"} 16.0
```

## Accuracy Tests
Should not affect model outputs.

## Checklist


python/sglang/srt/entrypoints/openai/serving_base.py
```diff
@@ -42,6 +43,19 @@ async def handle_request(
...
+                        raw_request.url.path, "400"
+                    )
+                    labels = {
+                        **self.tokenizer_manager.metrics_collector.labels,
+                        **metric_labels,
+                        **request.custom_labels,
+                    }
+                    self.tokenizer_manager.metrics_collector.observe_one_received_request(
+                        labels
+                    )
```
JustinTong0323: DRY: This code snippet is repeated several times, it could definitely abstract to a general function like:
```
def _maybe_record_metrics(self, status_code):
    if self.tokenizer_manager.enable_metrics:
         .....
```
Then call it each time

JustinTong0323 (changes requested):

python/sglang/srt/entrypoints/openai/serving_base.py
scottjlee: good point, refactored this and other common logic in other files into shared utility methods.

python/sglang/srt/entrypoints/openai/serving_base.py
```diff
@@ -52,30 +57,51 @@ async def handle_request(
...
         except HTTPException as e:
-            return self.create_error_response(
+            error_response = self.create_error_response(
                 message=e.detail, err_type=str(e.status_code), status_code=e.status_code
             )
+            self._label_and_observe_one_received_request(
+                raw_request,
+                http_status=str(e.status_code),
+            )
+            return error_response
```
acelyc111: Is it possible to add a decorator for `handle_request`? so when it add new short returns or exceptions, the metric can be corrctly updated without changes.

scottjlee: I think if the engine returns new exceptions, we won't need to add to this section. since we handle `HTTPException` (extracting the specific code from the `e.status_code` attribute), if one implements a new exception / return code from engine side, it would be extracted from [here](https://github.com/sgl-project/sglang/pull/11393/files/af06e30131e69214c462796ef88422bce4a8b33e#diff-390f2cfcbe1324ba605ba394bb258c74e278a9677ad6a2e46abf5dada3c9a3daR81).

In addition, we cover for general `Exception`s (using general `500` code), so other general exceptions which do not fall into the explicitly checked types (`HTTPException`, `ValueError`) should still be covered. So I think that if new exceptions or return codes are implemented, it should be covered.

I also played around with a quick refactor. in this case, i think even if we use a decorator for `handle_request()`, i believe we'd still need to replicate the bulk of the logic in the try-except chain to extract the status, or manually specify general codes like 200, 400, 500. Refactoring into the decorator didn't significantly reduce complexity, so instead I unified the calls to `_label_and_observe_one_received_request()` so this method is easier to maintain. For new exceptions, we just need to add a block like

```
except <ExceptionType> as e:
    http_status = "xxx"
```

to set a more specific exception type other than `500`. Let me know if this is aligned with what you had in mind

acelyc111: I see, thanks for the clarification.

acelyc111 (approved):

PR #11399, opened 2025-10-09
[router] leverage RAII to actively cancel request during client disconnect

slin1237:
## Problem Statement

When HTTP clients disconnect during request processing, the router **did not send abort requests to backend gRPC servers at all**. This caused:

- **All modes affected**: Single-server, PD streaming, and PD non-streaming modes
- **Resource waste**: Backend servers continued processing abandoned requests indefinitely
- **No cleanup mechanism**: Client disconnects had no effect on server-side generation
- **Cost impact**: Wasted GPU compute on requests whose results would never be consumed

## Solution Overview

This PR **implements complete abort request support** across all serving modes using Rust's RAII (Resource Acquisition Is Initialization) pattern for automatic cleanup when clients disconnect.

### What is RAII?

**RAII** is a Rust programming pattern where resources are automatically cleaned up when they go out of scope:

```rust
struct AbortOnDropStream {
    aborted: AtomicBool,  // Track if request completed successfully
    client: GrpcClient,
    request_id: String,
}

impl Drop for AbortOnDropStream {
    fn drop(&mut self) {
        // Automatically called when stream goes out of scope
        if !self.aborted.load() {
            send_abort_to_server();  // Cleanup: abort the request
        }
    }
}
```

**Key principles:**
- **Automatic**: No manual cleanup needed - Drop runs when variable goes out of scope
- **Guaranteed**: Drop runs even if errors/panics occur (like try/finally in other languages)
- **Explicit opt-out**: Call `mark_completed()` to prevent abort on drop

**In our implementation:**
- Stream created → holds active gRPC connection to backend server
- Stream dropped (client disconnect, error, panic) → **automatically sends abort RPC**
- Stream explicitly marked complete → skip abort on drop (normal completion)

This ensures that **client disconnects always trigger server-side cleanup**, regardless of how the disconnect happens (graceful close, network error, timeout, etc.).
## Checklist


CatherineSue:
Cancellation test:

Screenshot 2025-10-09 at 3 32 51 PM

CatherineSue (approved):

YouNeedCryDear (approved):

PR #11404, opened 2025-10-10
[Quantization] Deepseek-R1 mixed static and dynamic mxfp4 quant in Quark

BowenBao:
Allows flexible adjustment of additional layers to be quantized at load time from existing pre-quantized Quark mxfp4 model.

Taking https://huggingface.co/amd/DeepSeek-R1-0528-MXFP4-ASQ as example, where all attention projection layers were excluded. Modifying the config.json as follows to enable additional `o_proj` layer quantization. Note that `online: true` is set as an additional field to mark that this layer is quantized online after loading.

```diff
76d75
<       "re:model.layers.*.self_attn.o_proj",
122c121,158
<     "layer_quant_config": {},
---
>     "layer_quant_config": {
>       "*.o_proj": {
>         "online": true,
>         "bias": null,
>         "input_tensors": {
>           "ch_axis": -1,
>           "dtype": "fp4",
>           "group_size": 32,
>           "is_dynamic": true,
>           "is_scale_quant": false,
>           "mx_element_dtype": null,
>           "observer_cls": "PerBlockMXObserver",
>           "qscheme": "per_group",
>           "round_method": "half_even",
>           "scale_calculation_mode": "even",
>           "scale_format": "e8m0",
>           "scale_type": "float",
>           "symmetric": null
>         },
>         "output_tensors": null,
>         "target_device": null,
>         "weight": {
>           "ch_axis": -1,
>           "dtype": "fp4",
>           "group_size": 32,
>           "is_dynamic": false,
>           "is_scale_quant": false,
>           "mx_element_dtype": null,
>           "observer_cls": "PerBlockMXObserver",
>           "qscheme": "per_group",
>           "round_method": "half_even",
>           "scale_calculation_mode": "even",
>           "scale_format": "e8m0",
>           "scale_type": "float",
>           "symmetric": null
>         }
>       }
>     },
152d187
< 
```

Accuracy comparison:
Static quantization only:
```
# gsm8k
Accuracy: 0.947
Invalid: 0.000
```

Static quantization + dynamic quantization of attention projections:
```
# gsm8k
Accuracy: 0.943
Invalid: 0.000
```

### DIFF
--- a/python/sglang/srt/layers/quantization/quark/schemes/quark_w4a4_mxfp4.py
+++ b/python/sglang/srt/layers/quantization/quark/schemes/quark_w4a4_mxfp4.py
@@ -28,12 +28,23 @@ class QuarkW4A4MXFP4(QuarkScheme):
         self,
         weight_quant_spec: dict[str, Any],
         input_quant_spec: dict[str, Any],
+        online: bool,
         layer_name: Optional[str] = None,
     ):
+        """
+        Initializes the quantization scheme for the layer.
+
+        Args:
+            weight_quant_spec (dict[str, Any]): Specification for weight quantization.
+            input_quant_spec (dict[str, Any]): Specification for input quantization.
+            online (bool): Whether to use online quantization, which loads full/half precision weight and quantize post loading.
+            layer_name (Optional[str], optional): Name of the layer for debugging purpose. Defaults to None.
+        """
         self.out_dtype = torch.get_default_dtype()
         self.qscheme = "per_group"
         self.weight_quant_spec = weight_quant_spec
         self.input_quant_spec = input_quant_spec
+        self.online = online
         self.layer_name = layer_name
 
     @classmethod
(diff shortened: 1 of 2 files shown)


python/sglang/srt/layers/quantization/quark/schemes/quark_w4a4_mxfp4.py
```diff
@@ -52,16 +87,25 @@ def create_weights(
...
 
         # WEIGHT
+        if self.weight_quant_spec.get("is_dynamic") is True:
+            # Create and load bfloat16 weight
+            # quantize to mxfp4 during `process_weights_after_loading`
+            pack_factor = 1
+            dtype = torch.bfloat16
+        else:
+            pack_factor = 2
+            dtype = torch.uint8
```
fxmarty-amd: I understand where this is coming from, as you add `"is_dynamic": true` in the config, but it seems not ideal to me - as weights quantization is static by definition.

Shouldn't there be an other field/arg/parameter used to tell which layers to quantize online?

python/sglang/srt/layers/quantization/quark/schemes/quark_w4a4_mxfp4.py
```diff
@@ -25,19 +25,54 @@
...
+
+        return w, mx_scales
+
     def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
-        return
+        if self.weight_quant_spec.get("is_dynamic") is True:
+            # NOTE: This step happens after `post_load_weights` for deepseek models.
+            # So it is safe for kv_b_proj quantization of w_kc and w_vc in `quark_post_load_weights`
+            # since kv_b_proj.weight at that point is still bfloat16.
+            w, mx_scales = self.mxfp4_quantize(layer.weight.data)
```
fxmarty-amd, line 77: any assert check we can add on `layer.weight.data` here?

BowenBao: adds one for dtype

python/sglang/srt/layers/quantization/quark/schemes/quark_w4a4_mxfp4.py
BowenBao: yea I have similar thoughts. Adjusted with an additional field `"online": true` for this case, to keep `"is_dynamic": false`. wdyt?

PR #10190, opened 2025-09-09
[HICache] introduce evict policy

XucSh:
Introduce customize evict policy for HiCache

Cc @stmatengss 

## Checklist

- [ ] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.ai/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.ai/developer_guide/contribution_guide.html#benchmark-the-speed).

PR #11406, opened 2025-10-10
[Feature]Add Node priority calculation for hicache offloading

TTThanos:
## Motivation

We notice that in recent main branch change, the CacheOperations in List will be concatenated to form a larger CacheOperation in merge_ops() which use the minimal priority in these CacheOperations represent the holistic priority. While, PriorityQueue also preserved as the container to store CacheOperation. When we change writing_queue to PriorityQueue[CacheOperation] rather List[CacheOperation], we expect the node with large value for its priority(frequently & recently accessed with big size) should be processed firstly.

## Modifications

Add --enable-backup-priority as a server parameter(to enable priority calculation when use PriorityQueue as writing_queue)
Add priority calculation function which consider frequency, recency and size of the node

## Accuracy Tests

image

## Checklist


python/sglang/srt/server_args.py
```diff
@@ -2420,6 +2421,12 @@ def add_cli_args(parser: argparse.ArgumentParser):
             default=ServerArgs.hicache_storage_backend_extra_config,
             help="A dictionary in JSON string format containing extra configuration for the storage backend.",
         )
+        parser.add_argument(
+            "--enable-backup-priority",
```
stmatengss: It should be `--hicache-enable-backup-priority`

python/sglang/srt/mem_cache/hiradix_cache.py
```diff
@@ -260,7 +275,10 @@ def _inc_hit_count(self, node: TreeNode, chunked=False):
         if not node.backuped:
             if node.hit_count >= self.write_through_threshold:
                 # write to host if the node is not backuped
-                self.write_backup(node)
+                priority = None
+                if self.enable_backup_priority:
+                    priority = self.calculate_node_value(node)
+                self.write_backup(node, priority)
```
stmatengss: To use priority to determine write-back, ensure the second value is a bool.
```
def write_backup(self, node: TreeNode, write_back=False):
```

python/sglang/srt/mem_cache/hiradix_cache.py
```diff
@@ -221,15 +223,28 @@ def clear_storage_backend(self) -> bool:
             logger.warning("Hierarchical cache storage backend is not enabled.")
             return False
 
-    def write_backup(self, node: TreeNode, write_back=False):
+    def calculate_node_value(self, node: TreeNode):
```
stmatengss, line 233: What's the difference between calculate_node_value and other eviction policies? See #10190

TTThanos: From my point, these are two sperate functions which serves two different phases of hicache(evict & offload). The eviction policy in #10190 relates to evict Radix-Tree Node which is used to decide the eviction order of the TreeNode when eviction is triggered, while calculate_node_value() computes the priority of the Nodes deciding the offload order of these Nodes.

python/sglang/srt/server_args.py
TTThanos: Resolved

python/sglang/srt/mem_cache/hiradix_cache.py
TTThanos: Resolved

python/sglang/srt/mem_cache/hiradix_cache.py
TTThanos: Hi @stmatengss , Could you spend some time to check my update, thansk :)

TTThanos (commented):
Delete redundant param in scheduler.py after modifying the server_args

PR #6572, opened 2025-05-24
Support custom_logit_processor for OpenAI API

ruokee:
## Motivation

This PR tries to support custom logit processor for OpenAI chat completions endpoint.

## Modifications

1. Added support for custom logit processors in the OpenAI API:
    - Added `custom_logit_processor` and `custom_params` fields to `ChatCompletionRequest`
    - Modified the OpenAI adapter to pass these parameters through to the sampling logic
2. Added tests including batch processing with mixed requests
3. Added documentation improvements:
    - Enhanced sampling parameters documentation with OpenAI client examples
    - Provides a implementation of thinking budget based on custom logit processor for reasoning models (DeepSeek-R1).

## Checklist

- [x] Format your code according to the [Code Formatting with Pre-Commit](https://docs.sglang.ai/references/contribution_guide.html#code-formatting-with-pre-commit).
- [x] Add unit tests as outlined in the [Running Unit Tests](https://docs.sglang.ai/references/contribution_guide.html#running-unit-tests-adding-to-ci).
- [x] Update documentation / docstrings / example tutorials as needed, according to [Writing Documentation](https://docs.sglang.ai/references/contribution_guide.html#writing-documentation-running-docs-ci).
- [ ] Provide throughput / latency benchmark results and accuracy evaluation results as needed, according to [Benchmark and Profiling](https://docs.sglang.ai/references/benchmark_and_profiling.html) and [Accuracy Results](https://docs.sglang.ai/references/accuracy_evaluation.html).
- [ ] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co-author when merging the PR.
- [ ] Please feel free to join our Slack channel at https://slack.sglang.ai to discuss your PR.

PR #11416, opened 2025-10-10
Support Thinking Budget (via custom_logit_processor for OpenAI API) [Fix #6572]

whybeyoung:
Update for  https://github.com/sgl-project/sglang/pull/6572

1. Added support for custom logit processors in the OpenAI API:
2. Added custom_logit_processor and custom_params fields to ChatCompletionRequest
3. Modified the OpenAI adapter to pass these parameters through to the sampling logic
4. Added tests (including batch processing with mixed requests)
5. Added documentation improvements:
6. Enhanced sampling parameters documentation with OpenAI client examples
7. Added built-in thinking budget for reasoning models (DeepSeek-R1) (implemented based on custom logit processor)

## Summary by CodeRabbit

- New Features
  - Support passing custom_logit_processor and custom_params in Chat/Completion requests for advanced sampling control.
  - Introduces a thinking-budget logit processor with presets for DeepSeek R1 and Qwen3 to limit “thinking” token generation.

- Documentation
  - Added DeepSeek R1 thinking budget guide with server setup, Python example, and request payload usage.
  - Expanded sampling parameters examples for OpenAI-compatible chat completions.
  - Updated FAQ to suggest increasing distributed timeout for long model loads/NCCL timeouts.

- Tests
  - Added tests covering custom logit processor usage, including deterministic sampling and mixed batch scenarios.


whybeyoung:
@slin1237   can you  review this ?

python/sglang/srt/sampling/custom_logit_processor.py
```diff
@@ -51,3 +51,74 @@ def __call__(
...
+            start_index = cur_ids.index(self.THINKING_START_TOKEN_ID)
+
+            # Count the number of tokens after the thinking start token
+            num_tokens_after_start = len(cur_ids) - start_index
+
+            if num_tokens_after_start < thinking_budget:
+                continue
+
+            # Ensure new line token before thinking end token
+            if req.output_ids[-1] != self.NEW_LINE_TOKEN_ID:
```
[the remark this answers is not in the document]
CatherineSue: seems like a valid check

docs/basic_usage/deepseek.md
```diff
@@ -235,6 +235,44 @@ Important Notes:
...
+            "content": "Question: Is Paris the Capital of France?",
+        }
+    ],
+    max_tokens=1024,
+    extra_body={
+        "custom_logit_processor": DeepSeekR1ThinkingBudgetLogitProcessor().to_str(),
+        "custom_params": {
+            "thinking_budget": 512,
+        },
+    },
```
JustinTong0323, line 271: This pattern is not friendly for users. Could we let the users could pass `extra_body={"thinking_budget":512}` directly?

whybeyoung: I understand that the definition of "extra" should refer to some parameters that are not part of the standard OpenAI API. After thinking about it, what you said also makes sense.

whybeyoung: Another important reason is that the implementation here references the sample implementation of DeterministicLogitProcessor — and in that example, it is also placed within custom_params. see https://github.com/sgl-project/sglang/blob/main/docs/basic_usage/sampling_params.md @JustinTong0323

ruokee: > This pattern is not friendly for users. Could we let the users could pass `extra_body={"thinking_budget":512}` directly?

In my actual production deployment, the external interface I provide to users is indeed `thinking_budget`. However, I translate it internally via a middleware layer, so the final parameters actually passed to the engine are still `custom_logit_processor` and `custom_params`.

I noticed that SGLang currently does not appear to support passing `middleware` via server arguments (unlike vLLM, which does), so I had to write some custom code to set up the middleware.

This PR mainly focuses on supporting the `custom_logit_processor` parameter. I think other improvements should be addressed in separate PRs.

JustinTong0323: Yeah, it makes sense to me, it could be processed by middleware.

JustinTong0323 (approved):

Issue #10863, opened 2025-09-24
[Bug] DeepseekV3 DP + TP Attn, EP Moe, TBO, allgather bug


ADaBenxiong:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Describe the bug

When I run sglang, tp_size = 4, dp_size = 2, enable-dp-attention, moe-dense-tp-size 1, enable-two-batch-overlap.
I encountered the following error.
File "/workspace/sglang/python/sglang/srt/models/deepseek_v2.py", line 1938, in op_comm_prepare_attn
    self.layer_communicator.prepare_attn(hidden_states, residual, forward_batch)
  File "/workspace/sglang/python/sglang/srt/layers/communicator.py", line 230, in prepare_attn
    hidden_states = self._communicate_simple_fn(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/sglang/python/sglang/srt/layers/communicator.py", line 384, in _scattered_to_tp_attn_full
    attn_tp_all_gather_into_tensor(
  File "/workspace/sglang/python/sglang/srt/layers/dp_attention.py", line 500, in attn_tp_all_gather_into_tensor
    return get_attention_tp_group().all_gather_into_tensor(output, input)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/sglang/python/sglang/srt/distributed/parallel_state.py", line 637, in all_gather_into_tensor
    torch.ops.sglang.reg_all_gather_into_tensor(
  File "/usr/local/lib/python3.12/dist-packages/torch/_ops.py", line 1243, in __call__
    return self._op(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/sglang/python/sglang/srt/distributed/parallel_state.py", line 160, in reg_all_gather_into_tensor
[...]

PR #11423, opened 2025-10-10
Init TBO with dp_padded batch

liquanfeng:
## Motivation

After the input tensor padding was introduced in #8280, the TBO initialization logic was not updated. It continued to initialize with the original, unpadded input shape, which directly led to the shape mismatch error described in #10863.

## Modifications

Moved the TBO initialization to occur *after* the input padding process. This ensures the TBO is created with the correct, padded tensor shape, resolving the shape mismatch.

## Checklist


liquanfeng:
PTAL, thanks! It seems label `run-ci` is necessary for ci. @Ying1123 @merrymercy @lolipopshock @zhyncs @hnyls2002

ch-wan (approved):

PR #11424, opened 2025-10-10
fix(utils): fix bugs and some warnings

jimmy-evo:
## Motivation

a lot red blocks and yellow blocks in IDE

## Modifications

fix some bug and warnings

## Checklist

### DIFF
--- a/python/sglang/srt/layers/quantization/base_config.py
+++ b/python/sglang/srt/layers/quantization/base_config.py
@@ -3,42 +3,15 @@ from __future__ import annotations
 
 import inspect
 from abc import ABC, abstractmethod
-from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Type
 
 import torch
-from torch import nn
 
 if TYPE_CHECKING:
     from sglang.srt.layers.moe.moe_runner import MoeRunnerConfig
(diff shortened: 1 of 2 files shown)


python/sglang/srt/layers/quantization/base_config.py
```diff
@@ -3,42 +3,15 @@
...
-        Expects create_weights to have been called before on the layer."""
-        raise NotImplementedError()
-
-    def process_weights_after_loading(self, layer: nn.Module) -> None:
-        """Process the weight after loading.
-
-        This can be used for example, to transpose weights for computation.
-        """
-        return
+from sglang.srt.utils.common import QuantizeMethodBase
```
hnyls2002, line 14: Why remove this?

jimmy-evo: > Why remove this?

@hnyls2002 
https://github.com/sgl-project/sglang/pull/11424/files#diff-c5c592375d9bc6dcf56a0ad17bef0fea258bdf30a519958783bdc4be51a38cfaR1062

here use it, move to utils/common.py

hnyls2002:
> a lot red blocks and yellow blocks in IDE

This should not be the standard of code adjustment; the community has basic lint to guarantee code style.

jimmy-evo:
fixed in #11922

Issue #5309, opened 2025-04-11
[Feature] Deep EPLB Integration Proposal (Draft)


yicwang:
### Checklist

- [x] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 2. Please use English, otherwise it will be closed.

### Motivation

# Background
## Deep EPLB
With Expert Parallelism, experts are assigned to different GPUs at startup time. The load of experts may vary depending on the scenario and is dynamically changing, which could imbalance GPU loads. DeepSeek adopted a redundant experts strategy and EPLB (https://github.com/deepseek-ai/EPLB) is the load balancing algorithm to achieve the balancing goal.

## Expert Distribution
With this PR: https://github.com/sgl-project/sglang/pull/4435 we did some initial analysis of expert distribution. We ran the DeepSeek-R1 model with EPMoE enabled on an 8 * H20 server. Both EP and TP are set to 8. For a sample request of "What is the capital of France?", and the expert usage heat map for the first MoE layer is as below. The summary statistics shows leveraging eplb could balance the workload for each GPU better.

![Image](https://github.com/user-attachments/assets/cf61d41c-e33c-4b7a-90fe-dda169ed2bb8)

With the above data, we tried to run against eplb.py. Take the first layer count number raw data as an example:
```
tensor([489, 185,  49, 476,  89, 168, 203, 213,  40, 180, 179, 105, 158,  28,
        165, 385,  51, 341, 254, 228, 160,  50, 105, 199, 202,  54,  82,  52,
         30, 293,  90, 342,  24,  72,  47, 158, 261, 444,  71,  64,  72, 230,
        213, 180, 119,  23, 108, 310, 210, 167, 100,  60,  28,  46, 269, 158,
         70, 189,  54,  49,  31, 155, 239, 121,  48,  89, 162, 154,  58,  81,
         68,  18, 111,  24, 224, 224, 142, 182,  69, 131, 135, 320, 161, 113,
         89, 293, 283,  69, 114, 403, 106,  53,  57, 128, 120,  35, 560, 190,
        119, 108, 190, 170, 124, 173, 171,  70,  40, 143,  65, 671, 102,  31,
        136,  30,  95,  52, 127, 117, 111,  95,  32,  43, 166, 187, 114,  47,
        195, 112, 189,  27,   9, 177,  38, 202,  25, 121, 105,  28,  15, 713,
[...]

PR #11432, opened 2025-10-10
[sgl-kernel][1/N]Support Expert Specialization Grouped GEMM

HydraQYH:
## Motivation
There are two motivations for implementing Expert Specialization Grouped GEMM.
### 1. CUTLASS MoE Benchmark
When I used [benchmark](https://github.com/sgl-project/sglang/blob/main/python/sglang/test/test_cutlass_moe.py) to test the performance of FP8 Blockwise Grouped GEMM, I found that the performance was poor when the batch size was 512 to 1024:
test_cutlass_moe
The reason is that at this time FP8 Blockwise Grouped GEMM uses a larger TiledMMAShape: https://github.com/sgl-project/sglang/blob/a1a20b4c7cb86f81296ece3c4cb265584be1c179/sgl-kernel/csrc/moe/fp8_blockwise_moe_kernel.cu#L488-L489
At this time, the number of tokens that each expert needs to process is very small (M is very small), and using a large TiledMMAShape will result in additional unnecessary calculations.
### 2. Unbalanced load
Inspired by this [PR](https://github.com/sgl-project/sglang/issues/5309), the load of experts may vary depending on the scenario and is dynamically changing. Furthermore, this distribution is usually uneven, typically, most experts process only a small number of tokens, while a few experts handle a very large number of tokens.
## Implement
The implementation idea of ​​Expert Specialization is to select the appropriate kernel based on the number of tokens processed by the expert. We launch multiple grouped GEMM kernels, each using a different Configuration. For a specific expert, we select an appropriate kernel based on the number of tokens it processes.
There is a detail here. Since grouped GEMM kernel requires num_groups as a **Host** parameter, this usually requires MemcpyD2H to implement, which will bring significant performance overhead. Therefore, we use another method - **Mask Problem Sizes** to avoid MemcpyD2H.

## Modifications
**Implement:**
sgl-kernel/CMakeLists.txt
sgl-kernel/csrc/common_extension.cc
sgl-kernel/csrc/expert_specialization/es_fp8_blockwise.cu
sgl-kernel/csrc/expert_specialization/es_fp8_blockwise_functor.cuh
sgl-kernel/csrc/expert_specialization/es_fp8_blockwise_launcher.cuh
sgl-kernel/csrc/expert_specialization/es_fp8_blockwise_traits.cuh
sgl-kernel/include/sgl_kernel_ops.h
sgl-kernel/python/sgl_kernel/\_\_init\_\_.py
sgl-kernel/python/sgl_kernel/expert_specilization.py
**Benchmark:**
sgl-kernel/benchmark/bench_es_fp8_blockwise_grouped_gemm.py
**Unitest:**
sgl-kernel/tests/test_es_fp8_blockwise_moe.py

## Accuracy Tests
pytest -s test_es_fp8_blockwise_moe.py
image

## Benchmarking and Profiling
I tested it using the benchmark script on the H200 and found that in some scenarios, the performance gain of Expert Specialization was significant:
image
After that, I modified this:https://github.com/sgl-project/sglang/blob/3b9d97f3350daf130bd79a0d51f97c52bf7edf8f/python/sglang/srt/layers/moe/cutlass_moe.py#L184
replacing `fp8_blockwise_scaled_grouped_mm`with`es_fp8_blockwise_scaled_grouped_mm`, and tested it further.
Before:
image
After:
image
Compared to the native `fp8_blockwise_scaled_grouped_mm` of sgl-kernel, a general performance improvement can be observed.

## TODO List
There is still a lot of work to be done next.
- [ ] Fine-tune the range of M for H20 and H20.
- [ ] Adapting CUDAGraph
- [ ] Adapt the Python side code for H20 and H200. This is because H20 and H200 have significant differences in configuration and applicable scenarios.
- [ ] Adapt to the SM100+ architecture.
## Co-Author
I had many discussions with @yuan-luo @FlamingoPg @BBuf and thank you very much.
## Original Project
https://github.com/HydraQYH/expert_specialization_moe
## Checklist


sgl-kernel/csrc/expert_specialization/es_fp8_blockwise_launcher.cuh
```diff
@@ -0,0 +1,312 @@
...
+  // Creat Scale Factor Layout Functor
+  using LayoutSFA = typename PerfConfigMiddleMH20::LayoutSFA;
+  using LayoutSFB = typename PerfConfigMiddleMH20::LayoutSFB;
+  struct Fp8BlockwiseGroupedGemmSFLayoutFunctor<PerfConfigMiddleMH20> sf_layout(
+      reinterpret_cast<LayoutSFA*>(layout_sfa.data_ptr()), reinterpret_cast<LayoutSFB*>(layout_sfb.data_ptr()));
+
+  int num_experts = (int)expert_offsets.size(0);
+  auto stream = at::cuda::getCurrentCUDAStream(a_tensors.device().index());
+  // Dispatch
+  if (out_tensors.dtype() == torch::kBFloat16) {
```
BBuf, line 54: Maybe we can simplify replicate code here

```c++

// BFloat16 branch
if (out_tensors.dtype() == torch::kBFloat16) {
    // ... create functors ...
    if (!is_h20_device) {
        // Hx00 config
    } else {
        // H20 config
    }
} else if (out_tensors.dtype() == torch::kFloat16) {
    // ... screate functors ...
    if (!is_h20_device) {
        // Hx00 config
    } else {
        // H20 config
    }
}
```

->

```c++
template <typename ElementD>
void dispatch_pre_compute(
    const torch::Tensor& out_tensors,
    const torch::Tensor& a_tensors,
    const torch::Tensor& b_tensors,
    // ... other params
    bool is_h20_device,
    cudaStream_t stream) {
    
    struct Fp8BlockwiseGroupedGemmOffsetFunctor<cutlass::float_e4m3_t, float, ElementD> of(...);
    
    if (!is_h20_device) {
        dispatch_problem_size_filters<PerfConfigLowMHx00, PerfConfigMiddleMHx00, PerfConfigHighMHx00>(...);
    } else {
        dispatch_problem_size_filters<PerfConfigLowMH20, PerfConfigMiddleMH20, PerfConfigHighMH20>(...);
    }
}
```

sgl-kernel/csrc/expert_specialization/es_fp8_blockwise_launcher.cuh
```diff
@@ -0,0 +1,312 @@
...
+  using MiddleMGemmHx00Traits = ExpertSpecializationSm90FP8BlockwiseGroupedGemmTraits<
+      OutType,
+      cutlass::layout::ColumnMajor,
+      PerfConfigMiddleMHx00>;
+  using HighMGemmH20Traits =
+      ExpertSpecializationSm90FP8BlockwiseGroupedGemmTraits<OutType, cutlass::layout::RowMajor, PerfConfigHighMH20>;
+  using HighMGemmHx00Traits =
+      ExpertSpecializationSm90FP8BlockwiseGroupedGemmTraits<OutType, cutlass::layout::RowMajor, PerfConfigHighMHx00>;
+
+  const std::string H20_device_type_str("NVIDIA H20");
```
BBuf, line 224: Better create a tool func in `sglang/sgl-kernel/include/utils.h` .

sgl-kernel/python/sgl_kernel/expert_specilization.py
```diff
@@ -0,0 +1,27 @@
+import torch
```
BBuf, line 1: File name should be `expert_specialization.py`?

sgl-kernel/python/sgl_kernel/__init__.py
```diff
@@ -244,6 +244,7 @@ def _find_cuda_home():
     rmsnorm,
     silu_and_mul,
 )
+from sgl_kernel.expert_specilization import es_fp8_blockwise_scaled_grouped_mm
```
BBuf, line 247:
```suggestion
from sgl_kernel.expert_specialization import es_fp8_blockwise_scaled_grouped_mm
```

zhyncs (approved):
Overall LGTM, we just need to address @BBuf's review comment

PR #11381, opened 2025-10-09
perf: optimize qwen-vl with symm mem allreduce

yuan-luo:
## Motivation

The All-Reduce in VL model inference has a high proportion.
image

Qwen2.5-VL-72B uses inplace all-reduce, by default is NCCL AR which is inefficient.

image

This PR introduces in-place Symm Mem All-Reduce for Qwen2.5-VL model. (Literally for any other models using in-place AR) Improved 2% e2e input throughput. Single AR latency drops from 446.9us to 209.1us.
For Qwen3-VL model, improved 3.9% e2e input throughput.

```
python -m sglang.launch_server \
  --model-path /home/admin/Qwen2.5-VL-7B-Instruct \
  --host 0.0.0.0 --port 30000 --trust-remote-code \
  --tp-size 4 --enable-cache-report  --log-level info \
  --max-running-requests 64 --mem-fraction-static 0.65 \
  --chunked-prefill-size 8192  --attention-backend fa3 \
  --chat-template qwen2-vl --mm-attention-backend fa3 
  --enable-torch-symm-mem
```
Main:
image

PR:
image

```
python -m python.sglang.bench_serving \
    --backend sglang \
    --dataset-name random-image \
    --random-output-len 1 \
    --num-prompts 768 --max-concurrency 16 --random-image-resolution 360p --profile 

MAIN:
============ Serving Benchmark Result ============
Backend:                                 sglang    
Traffic request rate:                    inf       
Max request concurrency:                 16        
Successful requests:                     768       
Benchmark duration (s):                  69.40     
Total input tokens:                      422368    
Total generated tokens:                  718       
Total generated tokens (retokenized):    677       
Request throughput (req/s):              11.07     
Input token throughput (tok/s):          6085.89   
Output token throughput (tok/s):         10.35     
Total token throughput (tok/s):          6096.24   
Concurrency:                             15.89     
----------------End-to-End Latency----------------
Mean E2E Latency (ms):                   1435.81   
Median E2E Latency (ms):                 1426.29   
---------------Time to First Token----------------
Mean TTFT (ms):                          1268.69   
Median TTFT (ms):                        1397.04   
P99 TTFT (ms):                           1804.01   
---------------Inter-Token Latency----------------
Mean ITL (ms):                           0.00      
Median ITL (ms):                         0.00      
P95 ITL (ms):                            0.00      
P99 ITL (ms):                            0.00      
Max ITL (ms):                            0.00      
==================================================

PR:
============ Serving Benchmark Result ============
[...]

PR #11436, opened 2025-10-10
Revert "perf: optimize qwen-vl with symm mem allreduce"

ch-wan:
Reverts sgl-project/sglang#11381


yuan-luo:
The reason is qwen2 doesn't support dp-attention, but this PR breaks this restriction. 
I'll fix it in a separate PR.

yuan-luo:
Reland in https://github.com/sgl-project/sglang/pull/11457

Issue #10384, opened 2025-09-12
[Bug] Incremental detokenization seems to be removed


hnyls2002:
https://github.com/sgl-project/sglang/pull/4354/files#diff-b4937569fc71f6ad215181b633b2f89c7183a2b4ac39e41fc22635599a9be7deL444

In the previous implementation of incremental decoding, we only sent the incremental part of the tokens to be decoded.

Maybe letting the scheduler be aware of `which parts are good and do not need to be sent anymore` is kind of tricky and time-consuming, but I think this feature still needs to be done, but in other ways.

PR #11451, opened 2025-10-11
small inefficiency in stop str decoding

b8zhong:
## Motivation

No need to scan the entire decoded text again

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stop-string handling during text generation to check only the current output tail, reducing false positives and preventing premature or delayed termination.
  * Yields more consistent and predictable completion lengths across prompts and batches.
  * Enhances reliability of stop behavior in edge cases without changing any user-facing configuration or APIs.


hnyls2002:
`decoded_text` is always `""`; this is a deprecated design, and there is no inefficiency issue. We can totally remove all fields related to incremental decoding or introduce a new incremental decoding. @vincentzed @b8zhong Do you have interest in this?

b8zhong:
@hnyls2002 I see, thanks for pointing it out. It was still being fwd to detokenizer in `python/sglang/srt/managers/scheduler_output_processor_mixin.py` so I thought this code path was still used.

I see your https://github.com/sgl-project/sglang/issues/10384. Was it removed due to a lack of performance gain? If not, I'd be interested in the re-integration, as streaming is an important use case for us.

PR #11458, opened 2025-10-11
[BugFix][Qwen3-VL]: fix cu_seqlens in qwen3-vl 

ZhengWG:
## Motivation

The implementation of cur_seqlens is inconsistent with the HF implementation, resulting in accuracy issues with multi-frame/multi-patch inputs.

## Modifications

Fixed the implementation of cur_seqlens to align with the original HF implementation: https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen3_vl/modeling_qwen3_vl.py#L725

## Checklist

## Summary by CodeRabbit

- Bug Fixes
  - Improved dtype handling to prevent tracing issues in certain environments.
  - Adjusted sequence length calculation to avoid potential errors while preserving behavior.

- Refactor
  - Optimized internal computation for better performance and stability.
  - No changes to the external API.


python/sglang/srt/models/qwen3_vl.py
```diff
@@ -452,11 +452,11 @@ def forward(
...
-        cu_seqlens = torch.cat(
-            [
-                torch.tensor([0], device=grid_thw.device),
-                (grid_thw[:, 0] * grid_thw[:, 1] * grid_thw[:, 2]).cumsum(dim=0),
-            ]
+        cu_seqlens = torch.repeat_interleave(
+            grid_thw[:, 1] * grid_thw[:, 2], grid_thw[:, 0]
+        ).cumsum(
+            dim=0,
+            dtype=grid_thw.dtype if torch.jit.is_tracing() else torch.int32,
```
mickqian: Could you remove `torch.jit.is_tracing()`?

JustinTong0323 (approved):
Local test passed. Seems no infect to accuracy test.

xiaomin-D:
Hello, @ZhengWG , Thanks for the fix! 

Just curious - I noticed HuggingFace uses `F.pad(cu_seqlens, (1, 0), value=0)` instead of `torch.cat`, I did a quick benchmark and F.pad is slightly faster (~3-5%), though the absolute difference is negligible since this only runs once per forward pass. 

Is there a specific reason for the current approach?

Ref: https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen3_vl/modeling_qwen3_vl.py#L764

ZhengWG:
> Hello, @ZhengWG , Thanks for the fix!
> 
> Just curious - I noticed HuggingFace uses `F.pad(cu_seqlens, (1, 0), value=0)` instead of `torch.cat`, I did a quick benchmark and F.pad is slightly faster (~3-5%), though the absolute difference is negligible since this only runs once per forward pass.
> 
> Is there a specific reason for the current approach?
> 
> Ref: https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen3_vl/modeling_qwen3_vl.py#L764

You can take a look at: https://github.com/sgl-project/sglang/pull/11388

xiaomin-D:
> > Hello, @ZhengWG , Thanks for the fix!您好，谢谢修复！
> > Just curious - I noticed HuggingFace uses `F.pad(cu_seqlens, (1, 0), value=0)` instead of `torch.cat`, I did a quick benchmark and F.pad is slightly faster (~3-5%), though the absolute difference is negligible since this only runs once per forward pass.只是好奇 - 我注意到 HuggingFace 使用 `F.pad(cu_seqlens, (1, 0), value=0)` 而不是 `torch.cat` ，我做了一个快速基准测试，F.pad 稍微快一点（~3-5％），但绝对差异可以忽略不计，因为它只在每次前向传递中运行一次。
> > Is there a specific reason for the current approach?当前的做法有什么具体原因吗？
> > Ref: https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen3_vl/modeling_qwen3_vl.py#L764参考：https: [//github.com/huggingface/transformers/blob/main/src/transformers/models/qwen3_vl/modeling_qwen3_vl.py#L764](https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen3_vl/modeling_qwen3_vl.py#L764)
> 
> You can take a look at: #11388您可以查看： #11388
thanks！

PR #11064, opened 2025-09-29
[Bugfix]Remove fused_set_kv_buffer_arg assertion.

Ximingwang-09:
Co-author @yuan-luo 
## Motivation
When starting SGLang with Qwen3-235b and enabling torch.compile, the following error occurs. In fact, `fused_set_kv_buffer_arg` is not consumed in `forward_native`, so we can safely remove related assertion.

**CMD**
```
python3 -m sglang.launch_server \
--model Qwen3-235B \
--trust-remote-code \
--tp-size 8 \
--enable-cache-report \
--dtype bfloat16 \
--max-running-requests 32 \
--mem-fraction-static 0.92 \
--chunked-prefill-size 16384 \
--context-length 262144 \
--prefill-attention-backend fa3 \
--decode-attention-backend flashinfer \
--enable-torch-compile \
--torch-compile-max-bs 32
```
**ERROR**
```
2025-09-29 17:38:27 ERROR 357260 [ TP3 scheduler.py:2885] Scheduler hit an exception: Traceback (most recent call last):
2025-09-29 17:38:27 ERROR 357260 [ TP3 scheduler.py:2885]   File "/opt/conda/lib/python3.10/site-packages/sglang/srt/managers/scheduler.py", line 2842, in run_scheduler_process
2025-09-29 17:38:27 ERROR 357260 [ TP3 scheduler.py:2885]     scheduler = Scheduler(
2025-09-29 17:38:27 ERROR 357260 [ TP3 scheduler.py:2885]   File "/opt/conda/lib/python3.10/site-packages/sglang/srt/managers/scheduler.py", line 356, in __init__
2025-09-29 17:38:27 ERROR 357260 [ TP3 scheduler.py:2885]     self.tp_worker = TpWorkerClass(
2025-09-29 17:38:27 ERROR 357260 [ TP3 scheduler.py:2885]   File "/opt/conda/lib/python3.10/site-packages/sglang/srt/managers/tp_worker_overlap_thread.py", line 66, in __init__
2025-09-29 17:38:27 ERROR 357260 [ TP3 scheduler.py:2885]     self.worker = TpModelWorker(
2025-09-29 17:38:27 ERROR 357260 [ TP3 scheduler.py:2885]   File "/opt/conda/lib/python3.10/site-packages/sglang/srt/managers/tp_worker.py", line 96, in __init__
2025-09-29 17:38:27 ERROR 357260 [ TP3 scheduler.py:2885]     self.model_runner = ModelRunner(
[...]

PR #9833, opened 2025-08-31
[Bugfix] GPT-OSS + torch.compile + --moe-runner-backend triton compatibility

b8zhong:
## Motivation

After https://github.com/sgl-project/sglang/pull/9014 was merged, when we enable torch.compile and call RotaryEmbedding it defaults to forward_native() which always passes in fused_set_kv_buffer_arg.

As well, https://github.com/sgl-project/sglang/issues/8833#issuecomment-3219286773

```
TypeError: RotaryEmbedding.forward_native() got an unexpected keyword argument 'fused_set_kv_buffer_arg'
```

```python
q, k = self.rotary_emb(
    positions, q, k,
    fused_set_kv_buffer_arg=_create_fused_set_kv_buffer_arg(...)
)
```

## Modifications

Add it

## Benchmarking and Profiling

`python3 -m sglang.launch_server --model openai/gpt-oss-20b --enable-torch-compile --moe-runner-backend triton`

```
...
[2025-08-30 17:35:53] max_total_num_tokens=644795, chunked_prefill_size=8192, max_prefill_tokens=16384, max_running_requests=2518, context_len=131072, available_gpu_mem=7.92 GB
[2025-08-30 17:35:54] INFO:     Started server process [1529]
[2025-08-30 17:35:54] INFO:     Waiting for application startup.
[2025-08-30 17:35:54] INFO:     Application startup complete.
[2025-08-30 17:35:54] INFO:     Uvicorn running on http://127.0.0.1:30000 (Press CTRL+C to quit)
[2025-08-30 17:35:55] INFO:     127.0.0.1:46299 - "GET /get_model_info HTTP/1.1" 200 OK
[2025-08-30 17:35:55] Prefill batch. #new-seq: 1, #new-token: 6, #cached-token: 0, token usage: 0.00, #running-req: 0, #queue-req: 0, 
[2025-08-30 17:36:03] INFO:     127.0.0.1:37720 - "POST /generate HTTP/1.1" 200 OK
[2025-08-30 17:36:03] The server is fired up and ready to roll!
```

> Actually there is a separate compatibility issue with torch compile with the default triton_kernel moe runner backend so I'll raise that issue separately.

```
torch.compile ON (Triton normal backend)
#Input tokens: 32768
#Output tokens: 16384
batch size: 32
input_len: 1024
output_len: 512
latency: 8.03 s
ttft: 1.76 s
last generation throughput: 2734.27 tok/s
input throughput: 18595.85 tok/s
[...]

PR #11468, opened 2025-10-11
Remove fused_set_kv_buffer_arg assert for torch.compile compatibility

yhyang201:
## Modifications

This PR removes the assertion that checks fused_set_kv_buffer_arg is None in rotary_embedding.py.
The check is currently incompatible with torch.compile, causing failures during compilation.

The root cause of the incompatibility is still under investigation. For now, removing this assertion allows models using torch.compile to run without hitting this error.

#11417 
#11344

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed a blocking runtime assertion that caused errors when using fused KV cache with rotary embeddings, allowing models to run in this configuration without interruption.

* **New Features**
  * Enabled compatibility with fused KV cache when using rotary embeddings, preserving existing positional behavior while permitting smoother execution and potential performance gains.


fzyzcjy:
> The root cause of the incompatibility is still under investigation

maybe print it to see what is its value and type (naive guess: is it some kind of torch compile wrappers, or some code path is changed under compile?). 

if it is just compile artifacts we may workaround it; o/w this may indicate a real bug

fzyzcjy:
btw see this when cleaning my inbox 

* https://github.com/sgl-project/sglang/pull/9833
* https://github.com/sgl-project/sglang/pull/11064

b8zhong:
To my understanding, it fwd to forward pytorch impl when patching with torch compile, but it expects the same api (just a guess)

PR #11487, opened 2025-10-12
init support for KTransformers Heterogeneous Computing

Atream:
## Motivation

KTransformers Integration to Support CPU/GPU Hybrid Inference for MoE Models

Only support CompressedTensor format. Will support other formats later.

## Modifications

Hack num_gpu_experts to reduce VRAM usage and GPU calculation. Will refactor to use  experts_map instead of num_gpu_experts.

Add dependency of kt_kernel (https://github.com/kvcache-ai/ktransformers)

## test command

```
"args": [
        "--host", "0.0.0.0",
        "--port", "30010",
        "--model", "/mnt/data2/models/Qwen3-Next-80B-A3B-Instruct-GPTQ4",
        "--amx-weight-path", "/mnt/data2/models/Qwen3-Next-80B-A3B-Instruct-INT4",
        "--cpuinfer", "60",
        "--subpool-count", "2",
        "--num-gpu-experts", "256",
        "--attention-backend", "triton",
        "--trust-remote-code",
        "--mem-fraction-static", "0.98",
        "--chunked-prefill-size", "4096",
        "--max-running-requests", "37",
        "--max-total-tokens", "37000",
        "--served-model-name", "Qwen3-Next-80B-A3B-Thinking",
        "--enable-mixed-chunk",
        "--tensor-parallel-size", "4",
        "--enable-p2p-check",
        "--disable-shared-experts-fusion",
        "--amx-method", "AMXINT4",
      ]
```

```
"args": [
        "--host", "0.0.0.0",
        "--port", "30010",
        "--model", "/mnt/data2/models/DeepSeek-R1-0528-GPU-weight",
        "--amx-weight-path", "/mnt/data2/models/DeepSeek-R1-0528-CPU-weight",
        "--cpuinfer", "80",
        "--subpool-count", "2",
        "--num-gpu-experts", "1",
        "--attention-backend", "triton",
        "--trust-remote-code",
        "--mem-fraction-static", "0.98",
        "--chunked-prefill-size", "4096",
        "--max-running-requests", "37",
        "--max-total-tokens", "37000",
        "--served-model-name", "DeepSeek-R1-0528-FP8",
        "--enable-mixed-chunk",
        "--tensor-parallel-size", "4",
        "--enable-p2p-check",
        "--disable-shared-experts-fusion",
        "--amx-method", "AMXINT4",
      ]
```

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added “compressed-tensors” quantization with FP8 (DeepSeek) support and expanded WNA16 type support.
  * Introduced AMX/Marlin MoE execution paths with optional CPU inference/offload and expert-partitioning controls.
  * Exposed new CLI options: --amx-weight-path, --amx-method, --cpu-embed, --cpuinfer, --subpool-count, --cpu-save, --num-gpu-experts, --enable-defer.
* **Performance**
  * CUDA graphs now propagate capture batch sizes to AMX integration for potential speed-ups.
* **Bug Fixes**
  * More robust MoE handling when local GPU experts are fewer than global experts; fixed incorrect hidden-state slicing.


python/sglang/srt/layers/moe/fused_moe_triton/layer.py
```diff
@@ -828,17 +840,13 @@ def forward(self, hidden_states: torch.Tensor, topk_output: TopKOutput):
...
         combine_input = self.quant_method.apply(
             layer=self,
             dispatch_output=dispatch_output,
         )
-
         final_hidden_states = self.dispatcher.combine(combine_input)
 
-        final_hidden_states = final_hidden_states[
-            ..., :origin_hidden_states_dim
-        ].contiguous()
```
ShangmingCai, line 835: Will these removed lines cause any side effects?

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -2813,6 +2821,9 @@ def __init__(
         self.config = config
         self.tp_size = get_tensor_model_parallel_world_size()
         self.quant_config = quant_config
+        CompressedTensorsConfig.DeepSeekFP8Config = Fp8Config(
+            True, "dynamic", None, [128, 128]
+        )
```
ShangmingCai: Please check whether these lines impact GLM4.5: https://github.com/sgl-project/sglang/blob/d88ac9bc9a6528942166ed9c61fba53cacd0b87c/python/sglang/srt/models/glm4_moe.py#L736

ShangmingCai: Should these lines be wrapped in some conditions? cc: @fzyzcjy

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -2952,21 +2963,26 @@ def post_load_weights(self, is_nextn=False, weight_names=None):
                 torch.float8_e4m3fn,
                 torch.float8_e4m3fnuz,
             ):
+                if hasattr(self.quant_config, "DeepSeekFP8Config"):
+                    cur_quant_config = self.quant_config.DeepSeekFP8Config
+                else:
+                    cur_quant_config = self.quant_config
```
ShangmingCai: Logic looks correct, but the name `cur_quant_config` and `if hasattr(self.quant_config, "DeepSeekFP8Config")` this condition is not generic. This might make the code more difficult to understand. Maybe use `quant_config` or `selected_quant_config`? And use another self-explained flag to replace `hasattr(self.quant_config, "DeepSeekFP8Config")`,

ShangmingCai (commented):
LGTM. Just some minor issues.

python/sglang/srt/layers/moe/fused_moe_triton/layer.py
Atream: add it back

python/sglang/srt/models/deepseek_v2.py
Atream: We add if branch and  test GLM-4-Air，it works.

python/sglang/srt/models/deepseek_v2.py
Atream: Renamed to selected_quant_config.
Quant config will be refactor to support hybrid quant in the near future.

ShangmingCai (approved):
LGTM

python/sglang/srt/layers/moe/fused_moe_triton/layer.py
```diff
@@ -1,6 +1,7 @@
 # Adapted from https://github.com/vllm-project/vllm/blob/a6221a144af772fd1a68fe7e627935dc53e81738/vllm/model_executor/layers/fused_moe/layer.py
 
 import logging
+import re
```
ShangmingCai:
```suggestion
import re
```
Unused import?

python/sglang/srt/server_args.py
```diff
@@ -1436,6 +1467,46 @@ def _handle_other_validations(self):
...
+        parser.add_argument(
+            "--num-gpu-experts",
+            type=int,
+            help="The number of GPU experts.",
+        )
+        parser.add_argument(
+            "--enable-defer",
+            action="store_true",
+            help="Enable Expert Deferral.",
+        )
```
fzyzcjy: qq: is all these only used for ktransformer, or also useful for other cases? if the former maybe we should do a prefix, o/w users may get confused

EDIT: seems some are good, but some e.g. "subpool-count" may be confusing if w/o context

fzyzcjy (commented):
only glanced, will check later

fzyzcjy:
oh i see this one

the current recommendation is to follow @hnyls2002's new utility class (and also prefix with SGLANG, thus maybe SGLANG_KT_sth

image

python/sglang/srt/server_args.py
```diff
@@ -1436,6 +1463,36 @@ def _handle_other_validations(self):
 
     @staticmethod
     def add_cli_args(parser: argparse.ArgumentParser):
+
+        # Ktransformer server args
+        parser.add_argument(
```
merrymercy: move this to the later part of this function, after hicache related args

python/sglang/srt/server_args.py
```diff
@@ -194,6 +195,13 @@ class ServerArgs:
     revision: Optional[str] = None
     model_impl: str = "auto"
 
+    # Ktransformers
+    kt_amx_weight_path: Optional[str] = None
```
merrymercy: move this after hicache related args

sgl-kernel/python/sgl_kernel/fused_moe.py
```diff
@@ -101,6 +101,7 @@ def fused_marlin_moe(
 
     if global_num_experts == -1:
         global_num_experts = E
+
```
fzyzcjy: nit: rm unrelated empty line changes like this to minimize diff

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -2855,6 +2864,10 @@ def __init__(
         self.config = config
         self.tp_size = get_tensor_model_parallel_world_size()
         self.quant_config = quant_config
+        if os.environ.get("KT_MOE_AMX_WEIGHT_PATH") is not None:
```
fzyzcjy: (same as the discussion about env before)

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -2855,6 +2864,10 @@ def __init__(
...
+            CompressedTensorsConfig.DeepSeekFP8Config = Fp8Config(
```
fzyzcjy, line 2869: confused why we let CompressedTensorsConfig (a general backend) understand what is "deepseek", shall we improve the abstraction

python/sglang/srt/layers/moe/fused_moe_triton/layer.py
```diff
@@ -527,6 +528,11 @@ def _weight_loader_physical(
             if expert_id == -1:
                 return
 
+        if hasattr(self.quant_method, "num_gpu_experts"):
```
fzyzcjy: what about `isinstance(quant_method, ...)`, since it may be possible in the future another very unrelated class happens to have this field

python/sglang/srt/layers/quantization/compressed_tensors/compressed_tensors.py
```diff
@@ -129,6 +133,10 @@ def get_quant_method(
         ):
             return UnquantizedLinearMethod()
         if isinstance(layer, LinearBase):
+            if CompressedTensorsConfig.DeepSeekFP8Config is not None:
```
fzyzcjy, line 136: (same as before - would be great to not have "deepseek" in such general logic)

python/sglang/srt/models/deepseek_v2.py
Atream: We aim to employ mixed-precision quantization: int4 for the MoE layers (with some experts remaining on the GPU in int4) and native FP8 for the rest. This calls for a new quant-config that supports mixed precision and for off-loading the experts in a format-agnostic way—work we will finish in the next PR. For now we are only developing and testing on the single model at https://huggingface.co/ISTA-DASLab/DeepSeek-V3-0324-GPTQ-4b-128g-experts; to get off-loading up and verified quickly we have taken this admittedly hacky approach.

fzyzcjy: isee, if this is temporary and will be soon removed then LGTM; o/w if it is permanent then looks not very good

python/sglang/srt/server_args.py
Atream: fixed

sgl-kernel/python/sgl_kernel/fused_moe.py
Atream: fixed

python/sglang/srt/server_args.py
Atream: fixed

python/sglang/srt/models/deepseek_v2.py
Atream: This is temporary and will be soon removed.

python/sglang/srt/layers/moe/fused_moe_triton/layer.py
Atream: fixed

Atream:
All reviews have been fixed. Hybrid QuantConfig will be supported in the next PR.

python/sglang/srt/layers/quantization/compressed_tensors/compressed_tensors_moe.py
```diff
@@ -683,3 +733,353 @@ def apply(
...
+    if cpuinfer is not None:
+        envs.SGLANG_KT_MOE_CPUINFER.set(cpuinfer)
+    if threadpool_count is not None:
+        envs.SGLANG_KT_THREADPOOL_COUNT.set(threadpool_count)
+    if amx_weight_path is not None:
+        envs.SGLANG_KT_MOE_AMX_WEIGHT_PATH.set(amx_weight_path)
+    if amx_method is not None:
+        envs.SGLANG_KT_AMX_METHOD.set(amx_method)
+    if chunked_prefill_size is not None:
+        envs.SGLANG_KT_MOE_CHUNKED_PREFILL_SIZE.set(chunked_prefill_size)
```
fzyzcjy, line 906: I personally feel we should use the global server_args variable instead of using a env and a serverargs

fzyzcjy (approved):
mainly checked the API and the parts modifying existing code, and the newly added code is not checked in detail. generally LGTM

python/sglang/srt/layers/moe/fused_moe_triton/layer.py
```diff
@@ -215,6 +219,8 @@ def __init__(
                 if not use_weight_loader_fused
                 else self.weight_loader_fused
             ),
+            intermediate_size_full=intermediate_size,
+            top_k=top_k,
```
ch-wan, line 223: These two should not be a part of `create_weights`. Please remove them in your next PR. Also, `intermediate_size_full` can be derived using `intermediate_size * `get_moe_tp_size()`.

PR #11504, opened 2025-10-12
[smol] improve model load speed on `xet` backend. 

vincentzed:
## Motivation

Two smol change. 

1. `HF_XET_HIGH_PERFORMANCE` increase the resource use to load model faster. https://huggingface.co/docs/huggingface_hub/en/package_reference/environment_variables#hfxethighperformance
2. Inform the user if we are not using `hf_transfer`. However, by default, in `huggingface_hub` >  `0.32.0` includes `hf_xet`. 
3. Fix bug where the option to disable hf_hub is not respected.

_Note: Where versions of huggingface_hub >= 0.30.0 and < 0.32.0 are installed, hf_xet must be installed explicitly: `pip install -U hf-xet`_
In most case the last point should never happen.

## Modifications
TODO: benchmark for the model load change. It can only improve the Xet load speed. And based on the project dep's, it only happens if hf_transfer is disabled. Previously, it was not possible to disable it because `enable_hf_transfer()` was always called.


vincentzed:
BTW, on a slow home network, i., 30 30MB/s, it will not make any difference.

python/sglang/srt/environ.py
```diff
@@ -206,6 +206,7 @@ class Envs:
     SGLANG_MOE_PADDING = EnvBool(False)
     SGLANG_CUTLASS_MOE = EnvBool(False)
     HF_HUB_DISABLE_XET = EnvBool(False)
+    HF_XET_HIGH_PERFORMANCE = EnvBool(True)
```
JustinTong0323, line 210: Why we add this ENV?

JustinTong0323:
Also please fix lint~

python/sglang/srt/environ.py
vincentzed: It's not enabled by default (false by default). Basically max resource to load faster on model load. It only make a difference on high speed networks

hnyls2002: If this is false by default, why would you set it true in SGLang by default. This is conflict with hf hebavior.

vincentzed: We enable hf transfer and install it which isn't default behaviour either. but I am not opinionated on this change.

> If you are running on a machine with high bandwidth, you can speed up downloads by allowing hf_xet to run on all CPU cores. hf_xet is a Rust-based package leveraging the new [Xet storage backend](https://huggingface.co/docs/hub/en/xet/index) to optimize file transfers with chunk-based deduplication. hf_xet is enabled by default but with lower performances to avoid bloating available CPU and bandwidth, which could degrade UX.
https://github.com/huggingface/hub-docs/blob/main/docs/hub/models-downloading.md

b8zhong:
Now set by default

PR #11510, opened 2025-10-12
[Bugfix] Fix Qwen3/DSV3/DSV3.2 model support

iforgetmyname:
## Motivation

This pr fixes very models that failed to start on npu after dsv3.2 support pr was merged. ci and image release infra was also improved.

## Modifications

- Fix mla prefill padding mismatch with actual seq lens in ascend_backend
- Fix dtype issue in AscendTokenToKVPool that breaks save kv ops
- Fix deepseek nsa condition
- Adapt default value of chunked_prefill_size and cuda_graph_max_bs
- Add packages to support DSV3.2

## Accuracy Tests

Will be covered by ci

## Benchmarking and Profiling

N/A

## Checklist


python/sglang/srt/mem_cache/memory_pool.py
```diff
@@ -1026,16 +1026,19 @@ def set_kv_buffer(
...
+            loc.view(-1, 1),
+            cache_k.view(-1, 1, self.head_num, self.head_dim),
+        )
+        torch_npu.npu_scatter_nd_update_(
+            self.v_buffer[layer_id - self.start_layer].view(
+                -1, 1, self.head_num, self.head_dim
             ),
-            slot_indices=loc,
+            loc.view(-1, 1),
+            cache_v.view(-1, 1, self.head_num, self.head_dim),
```
ping1jing2: we don't need this change, we can replace it with [this](https://github.com/sgl-project/sglang/commit/748900fa3f6937c8611c2599ad4dcca7286c26c3)

iforgetmyname: downcasting int64 to int32 is definitely not a good solution

python/sglang/srt/mem_cache/common.py
```diff
@@ -347,11 +347,7 @@ def alloc_for_extend(
     else:
         # Paged allocation - build last_loc
         last_loc = [
-            (
-                t[-1:]
-                if len(t) > 0
-                else torch.tensor([-1], device=batch.tree_cache.device)
-            )
+            (t[-1:] if len(t) > 0 else torch.tensor([-1], device=batch.device))
```
Alcanderian, line 350: @hnyls2002 @xiezhq-hermann Is this modification resaonable?

hnyls2002: Reasonable, they are the same. Maybe because previous v2-mem refactor removed the `device` attribute of `tree_cache`

hnyls2002 (approved):

PR #11364, opened 2025-10-09
chore: upgrade flashinfer 0.4.0

zhyncs:
## Checklist

- [ ] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.ai/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.ai/developer_guide/contribution_guide.html#benchmark-the-speed).

PR #11514, opened 2025-10-12
chore: remove flashinfer cleanup cache

zhyncs:
## Motivation

ref https://github.com/sgl-project/sglang/pull/11364#discussion_r2418077197

## Checklist


zhyncs:
FYI When we upgrade flashinfer v0.4.0, due to the incompatibility between versions, the cache of the old version will affect the use of the new version, so we added clear cache. However, it has been merged for several days now and there is no old version cache on the online runner, so we can delete this line.

PR #11519, opened 2025-10-12
[Fix] Skip visual layers when applying LoRA to Qwen2VL modules

anvdn:
# Motivation

The LoRA target modules (e.g. `qkv_proj`) exist in both the visual tower and the LM tower. However, LoRA adapters should only be applied to the latter as it is the default set-up when fine-tune this model (e.g. with `transformers`). The current logic only considers the module name and will error out at runtime while attempting to apply LoRA to the visual tower. 

## Modifications

- `python/sglang/srt/models/qwen2_vl.py`: implemented a logic skipping the modules from the visual tower

## Accuracy Tests

N/A

## Benchmarking and Profiling

N/A

## Checklist


anvdn:
@lifuhuang could you potentially take a look when you get the time? I see you've made similar changes for other models in the past: https://github.com/sgl-project/sglang/pull/8776

python/sglang/srt/models/qwen2_vl.py
```diff
@@ -442,6 +442,10 @@ class Qwen2VLForConditionalGeneration(nn.Module):
         "up_proj": ("gate_up_proj", 1),
     }
 
+    lora_pattern = re.compile(
+        r"^model\.layers\.(\d+)\.(?:self_attn|mlp)\.(?:qkv_proj|o_proj|down_proj|gate_up_proj)"
```
JustinTong0323: if only visual modules should be skipped, could we just match "visual"?

anvdn: Yep agreed that looks better.

anvdn: I think the name `should_apply_lora` kind of implies that each of the modules for which `True` is returned should have their corresponding adapter, which isn't the case if we merely exclude the visual tower. This is why I computed a full regex pattern.

python/sglang/srt/models/qwen2_vl.py
```diff
@@ -514,6 +513,10 @@ def _process_video_input(self, video_input: Qwen2VLVideoInputs) -> torch.Tensor:
     def get_input_embeddings(self):
         return self.model.embed_tokens
 
+    def should_apply_lora(self, module_name: str) -> bool:
```
Fridge003, line 516: Is it possible to filter the lora modules with a pattern regex,
like this one: https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/models/gemma3_mm.py#L159

anvdn: This is what I initially did but @JustinTong0323 suggested that I only match `"visual"` instead of building a full regex pattern for all target modules, so I updated the logic in the last commit. Which one do you guys prefer? I don't really mind.

Fridge003: okok, I missed @JustinTong0323's comment. Then it seems ok for me..

Fridge003 (approved):

Issue #10266, opened 2025-09-10
[Feature] Optimize LoRA Loading Mechanism to Decouple User Limits from CPU Memory Constraints


lw9527:
### Checklist

- [ ] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [ ] 2. Please use English, otherwise it will be closed.

### Motivation

Currently, the `--max-loaded-loras` parameter imposes a hard limit on both:
1. The number of LoRAs that can be loaded into CPU memory
2. The number of LoRAs a user is allowed to load

This creates a suboptimal user experience, as users may be prevented from loading enough LoRAs. 

We hope a more flexible approach where:
- The CPU memory constraint remains (to prevent OOM errors)
- User-facing limits on the number of LoRAs are removed
- LoRAs are dynamically loaded/unloaded from CPU memory based on user requests, within the bounds of available memory

This would allow users to request any number of LoRAs while the system automatically manages memory usage, resulting in a smoother experience without artificial restrictions.

**Expected Behavior:**
- users can submit requests for any number of LoRAs without arbitrary limits
- the `--max-loaded-loras` limits the number of Loras that can be loaded into CPU memory
- When max num Loras is reached, least recently used (or other smart eviction) LoRAs are unloaded to make space with new request.

glenliu21:
Can I please be assigned to this issue? I have experience with SGLang internals, and I'm new to LoRAs but happy to learn and contribute!

PR #11526, opened 2025-10-13
allow loras to be implicitly evicted and loaded based on max_loaded_loras

glenliu21:
## Motivation

This PR implements the feature in #10266.

## Modifications

This PR introduces two new features:
- `max_loaded_loras` is no longer a hard limit. That is, if a LoRA adapter is loaded and causes the limit to be exceeded, then the LRU LoRA adapter will be evicted. Furthermore, pinned LoRA adapters can't be implicitly evicted.
- Now that LoRA adapters can be implicitly evicted, to be consistent I also added functionality for LoRA adapters to be implicitly loaded. That is, if `generate` is called with a LoRA adapter that was previously loaded but now evicted, it will be reloaded. Note that LoRA adapters that haven't been explicitly loaded and then unloaded cannot be implicitly loaded.

## Accuracy Tests

Modified existing tests and added new unit tests.

## Checklist


python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -308,6 +309,12 @@ def __init__(
         # Please note that, unlike `model_update_lock`, this does not block inference, allowing
         # LoRA updates and inference to overlap.
         self.lora_update_lock = asyncio.Lock()
+        # A cache for mapping the lora_name for LoRA adapters that have been loaded at any
+        # point to their latest LoRARef objects, so that they can be
+        # dynamically loaded if needed for inference
+        self.lora_ref_cache: Dict[str, LoRARef] = {}
+        for lora_ref in self.server_args.lora_paths:
```
lw9527: This error occurs when the LoRA feature is not used.
for lora_ref in self.server_args.lora_paths:
TypeError: 'NoneType' object is not iterable

glenliu21: Thanks for pointing this out! It is fixed now.

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -408,6 +416,51 @@ async def generate_request(
...
+                    load_result = await self.load_lora_adapter(
+                        LoadLoRAAdapterReqInput(
+                            lora_name=new_lora_ref.lora_name,
+                            lora_path=new_lora_ref.lora_path,
+                            pinned=new_lora_ref.pinned,
+                        )
+                    )
+                    if (
+                        not load_result.success
+                        and "already exists" not in load_result.error_message
```
lw9527: Is this "already loaded"?

lifuhuang:
x-posting one open topic from our Slack thread:

> I think the change is reasonable, however, one part I have been debating myself is regarding where we should implement the eviction logic -- should it be in main process (tokenizer process) or TP worker process.
> For GPU eviction, the LRU logic is introduced by [@Connor Li](https://sgl-fru7574.slack.com/team/U096R96U00N) in TP worker.
> For CPU eviction (in your PR), the LRU logic is in tokenizer process.
> This is unideal because it introduced duplication in code/logic. Ideally we should have a single LRU logic and generalize it such that it can handle for both CPU and GPU. But at the same time, I am also aware that this MIGHT be a price we have to pay to sacrifice aesthetic value for better perf (I believe a large portion of the complexity was introduced by my prev change to support async lora loading :neutral_face:)
> I was hoping to think more carefully about this and see if we can find a design that to some extent achieve both maintainability and perf, but I haven't got a chance to really think it through

Perhaps we can continue our discussion in Slack.

python/sglang/srt/managers/tokenizer_manager.py
glenliu21: Thanks for catching this!

Fridge003 (approved):

PR #11528, opened 2025-10-13
Depreate `global_server_args_dict`

hnyls2002:


python/sglang/srt/model_executor/model_runner.py
```diff
@@ -278,15 +275,12 @@ def __init__(
...
-                "use_mla_backend": self.use_mla_backend,
-                "speculative_algorithm": self.spec_algorithm,
-            }
-        )
+        # Set the global server_args in the scheduler process
+        set_global_server_args_for_scheduler(server_args)
+        global_server_args = get_global_server_args()
+
+        # FIXME: hacky set `use_mla_backend`
+        global_server_args.use_mla_backend = self.use_mla_backend
```
kartikx, line 283: Hi @hnyls2002, can you help me understand why this is needed? I observe that this assignment over-writes the function `ServerArgs.use_mla_backend()` with a variable, which can cause exceptions later on if someone tries to call it. I was considering raising a fix for this.

PR #11602, opened 2025-10-14
make eagle3 be compatible with dp attention 

yizhang2077:
## Motivation

1. llama eagle3 be compatible with dp attention
2. make EAGLE3 be compatible with deepep when tp/dp = 1:1 (if tp/dp !=1, we need transfer embedding from scatter mode to dp attention full mode)
3. change draft model logits size from `vocab_size` to `draft_vocab_size`
## Checklist


ispobock (commented):
@yizhang2077 Could you add unit test for this change?

python/sglang/srt/models/llama_eagle3.py
```diff
@@ -124,6 +254,7 @@ def __init__(
             config.vocab_size,
             config.hidden_size,
             prefix=add_prefix("embed_tokens", prefix),
+            enable_tp=not is_dp_attention_enabled(),
```
ispobock, line 166: There are some points to discuss:
1. Should we also enable DP attention for the draft model when it’s enabled for the target model? Do we need an option for draft model to enable the DP attn?
2. If DP attention is enabled for the draft model, both target and draft must be in the same DP rank, otherwise KV cache may be inconsistent.
3. If enabled, how do performance and accuracy compare with TP?

yizhang2077: 1. draft model and target model shares the same server_args and tp/attn_tp communication groups, so we do not need to add extra parameter for enable dp attention.
2. since 1, we just need to use the same tp/attn_tp group then we can keep kv cache consistent

PR #11610, opened 2025-10-14
[Fix] sampling defaults: avoid inheriting HF defaults when using “model” sampling

### LINKED ISSUE #11574: [Bug] PR #11304 sets top_k = 50 by default which is wrong
### Checklist

### Describe the bug

#11304 changes the default top_k to 50 which might be wrong ? 

[2025-10-14 02:01:37] Using default chat sampling params from model generation config: {'repetition_penalty': 1.0, 'temperature': 1.0, 'top_k': 50, 'top_p': 1.0}

this is how I'm loading model: 

 NCCL_P2P_LEVEL=4 NCCL_DEBUG=INFO CUDA_VISIBLE_DEVICES=4,5,6,7 USE_TRITON_W8A8_FP8_KERNEL=1 SGL_ENABLE_JIT_DEEPGEMM=false PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True python -m sglang.launch_server --model /mnt/GLM-4.5-Air-FP8/  --tp 4 --speculative-algorithm EAGLE   --speculative-num-steps 3   --speculative-eagle-topk 1   --speculative-num-draft-tokens 4 --host 0.0.0.0 --port  4999 --mem-fraction-static 0.80  --context-length 128000  --enable-metrics  --attention-backend flashinfer  --tool-call-parser glm45    --reasoning-parser glm45   --served-model-name glm-4.5-air   --chunked-prefill-size 64736  --enable-mixed-chunk   --cuda-graph-max-bs 1024 --model-loader-extra-config '{"enable_multithread_load": true, "num_threads": 8}'  --sampling-defaults model

there is no top_k configuration in the glm which should set this to 50 - where this 50 is coming from? 

The side effect is that setting top_k to anything else than -1 causes token / sec speed drop by 30% when the eagle speculative is enabled 

### Reproduction

NCCL_P2P_LEVEL=4 NCCL_DEBUG=INFO CUDA_VISIBLE_DEVICES=4,5,6,7 USE_TRITON_W8A8_FP8_KERNEL=1 SGL_ENABLE_JIT_DEEPGEMM=false PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True python -m sglang.launch_server --model /mnt/GLM-4.5-Air-FP8/  --tp 4 --speculative-algorithm EAGLE   --speculative-num-steps 3   --speculative-eagle-topk 1   --speculative-num-draft-tokens 4 --host 0.0.0.0 --port  4999 --mem-fraction-static 0.80  --context-length 128000  --enable-metrics  --attention-backend flashinfer  --tool-call-parser glm45    --reasoning-parser glm45   --served-model-name glm-4.5-air   --chunked-prefill-size 64736  --enable-mixed-chunk   --cuda-graph-max-bs 1024 --model-loader-extra-config '{"enable_multithread_load": true, "num_threads": 8}'  --sampling-defaults model

[2025-10-14 02:01:37] Using default chat sampling params from model generation config: {'repetition_penalty': 1.0, 'temperature': 1.0, 'top_k': 50, 'top_p': 1.0}

### Environment

4x RTX 6000 PRO 
the latest sglang docker

adarshxs:
## Motivation
fixes: https://github.com/sgl-project/sglang/issues/11574
When `--sampling-defaults=model` is enabled, it would inadvertently use HuggingFace’s default sampling parameters [(e.g. top_k = 50)](https://huggingface.co/docs/transformers/main/en/main_classes/text_generation#transformers.GenerationConfig.top_k) when the model’s generation_config.json did not explicitly define them.

## Modifications

- Only includes sampling parameters from hf_generation_config if they differ from HuggingFace’s base defaults
- Omits parameters that are unset or equal to HF’s defaults
- Returns an empty dict if nothing is explicitly defined, so SGLang’s own fallback defaults are used

eg:
```bash
python3 -m sglang.launch_server --model-path meta-llama/Llama-3.2-1B --host 0.0.0.0 --sampling-defaults model
```

before:

```
[2025-10-14 12:27:18] INFO:     Started server process [10577]
[2025-10-14 12:27:18] INFO:     Waiting for application startup.
[2025-10-14 12:27:18] Using default chat sampling params from model generation config: {'repetition_penalty': 1.0, 'temperature': 0.6, 'top_k': 50, 'top_p': 0.9}
[2025-10-14 12:27:18] Using default chat sampling params from model generation config: {'repetition_penalty': 1.0, 'temperature': 0.6, 'top_k': 50, 'top_p': 0.9}
[2025-10-14 12:27:18] INFO:     Application startup complete.
[2025-10-14 12:27:18] INFO:     Uvicorn running on http://0.0.0.0:30000 (Press CTRL+C to quit)
```

after:
```
[2025-10-14 12:25:11] INFO:     Started server process [9684]
[2025-10-14 12:25:11] INFO:     Waiting for application startup.
[2025-10-14 12:25:11] Using default chat sampling params from model generation config: {'temperature': 0.6, 'top_p': 0.9}
[2025-10-14 12:25:11] Using default chat sampling params from model generation config: {'temperature': 0.6, 'top_p': 0.9}
[2025-10-14 12:25:11] INFO:     Application startup complete.
[2025-10-14 12:25:11] INFO:     Uvicorn running on http://0.0.0.0:30000 (Press CTRL+C to quit)
```

## Checklist

### DIFF
--- a/python/sglang/srt/configs/model_config.py
+++ b/python/sglang/srt/configs/model_config.py
@@ -686,13 +686,12 @@ class ModelConfig:
         Returns:
             A dictionary containing the non-default sampling parameters.
         """
+
         if self.sampling_defaults != "model":
             return {}
 
         if self.hf_generation_config is None:
             return {}
-
-        config = self.hf_generation_config.to_dict()
 
         available_params = [
             "repetition_penalty",
@@ -701,10 +700,18 @@ class ModelConfig:
             "top_p",
             "min_p",
         ]
+        default_sampling_params: dict[str, Any] = {}
 
-        default_sampling_params = {
-            p: config.get(p) for p in available_params if config.get(p) is not None
-        }
+        for param in available_params:
+            value = getattr(self.hf_generation_config, param, None)
+            if value is None:
+                continue
+            from transformers import GenerationConfig
+
+            base = GenerationConfig()
+            base_default = getattr(base, param, None)
+            if base_default is None or value != base_default:
+                default_sampling_params[param] = value
 
         return default_sampling_params
 
--- a/python/sglang/srt/configs/model_config.py
+++ b/python/sglang/srt/configs/model_config.py
@@ -703,10 +703,20 @@ class ModelConfig:
         default_sampling_params: dict[str, Any] = {}
 
         base = GenerationConfig()
+
+        explicit_keys = getattr(self.hf_generation_config, "_explicit_keys", set())
+
         for param in available_params:
             value = getattr(self.hf_generation_config, param, None)
             if value is None:
                 continue
+
+            # respect explicit keys from the JSON
+            if param in explicit_keys:
+                default_sampling_params[param] = value
+                continue
+
+            # otherwise, only include if it differs from HF base defaults
             base_default = getattr(base, param, None)
             if base_default is None or value != base_default:
                 default_sampling_params[param] = value
--- a/python/sglang/srt/configs/model_config.py
+++ b/python/sglang/srt/configs/model_config.py
@@ -20,7 +20,7 @@ from enum import Enum, IntEnum, auto
 from typing import Any, Dict, List, Optional, Set, Union
 
 import torch
-from transformers import PretrainedConfig
+from transformers import GenerationConfig, PretrainedConfig
 
 from sglang.srt.environ import envs
 from sglang.srt.layers.quantization import QUANTIZATION_METHODS
@@ -701,8 +701,6 @@ class ModelConfig:
             "min_p",
         ]
         default_sampling_params: dict[str, Any] = {}
-
-        from transformers import GenerationConfig
 
         base = GenerationConfig()
         for param in available_params:
--- a/python/sglang/srt/configs/model_config.py
+++ b/python/sglang/srt/configs/model_config.py
@@ -702,13 +702,12 @@ class ModelConfig:
         ]
         default_sampling_params: dict[str, Any] = {}
 
+        from transformers import GenerationConfig
+        base = GenerationConfig()
         for param in available_params:
             value = getattr(self.hf_generation_config, param, None)
             if value is None:
                 continue
-            from transformers import GenerationConfig
-
-            base = GenerationConfig()
             base_default = getattr(base, param, None)
             if base_default is None or value != base_default:
                 default_sampling_params[param] = value
--- a/python/sglang/srt/configs/model_config.py
+++ b/python/sglang/srt/configs/model_config.py
@@ -703,6 +703,7 @@ class ModelConfig:
         default_sampling_params: dict[str, Any] = {}
 
         from transformers import GenerationConfig
+
         base = GenerationConfig()
         for param in available_params:
             value = getattr(self.hf_generation_config, param, None)
--- a/python/sglang/srt/utils/hf_transformers_utils.py
+++ b/python/sglang/srt/utils/hf_transformers_utils.py
@@ -251,11 +251,27 @@ def get_generation_config(
     **kwargs,
 ):
     try:
-        return GenerationConfig.from_pretrained(
+        cfg = GenerationConfig.from_pretrained(
             model, trust_remote_code=trust_remote_code, revision=revision, **kwargs
         )
     except OSError as e:
         return None
+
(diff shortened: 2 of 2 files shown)


python/sglang/srt/configs/model_config.py
```diff
@@ -686,25 +686,32 @@ def get_default_sampling_params(self) -> dict[str, Any]:
...
             "top_p",
             "min_p",
         ]
-
-        default_sampling_params = {
-            p: config.get(p) for p in available_params if config.get(p) is not None
-        }
+        default_sampling_params: dict[str, Any] = {}
+
+        from transformers import GenerationConfig
```
CatherineSue: nit: please move imports at the top of the file

python/sglang/srt/configs/model_config.py
```diff
@@ -686,25 +686,32 @@ def get_default_sampling_params(self) -> dict[str, Any]:
...
+
+        base = GenerationConfig()
+        for param in available_params:
+            value = getattr(self.hf_generation_config, param, None)
+            if value is None:
+                continue
+            base_default = getattr(base, param, None)
+            if base_default is None or value != base_default:
```
CatherineSue, line 721: What does `value != base_default` mean here?

CatherineSue: What if I have a model which has a generation config.json that uses top_k = 50?

adarshxs: didn't really think about this. do you have any idea how we can handle this?

python/sglang/srt/configs/model_config.py
adarshxs: done

python/sglang/srt/configs/model_config.py
adarshxs: fixed

adarshxs:
@CatherineSue we pull the params from the generation_config json and check if they were explicitly set and give them priority. Solves the problem.

python/sglang/srt/configs/model_config.py
```diff
@@ -703,10 +703,20 @@ def get_default_sampling_params(self) -> dict[str, Any]:
...
+
+        explicit_keys = getattr(self.hf_generation_config, "_explicit_keys", set())
+
         for param in available_params:
             value = getattr(self.hf_generation_config, param, None)
             if value is None:
                 continue
+
+            # respect explicit keys from the JSON
+            if param in explicit_keys:
```
CatherineSue, line 715: What is the difference between the logic from L715 to L717 and the logic from L710 to L712?

CatherineSue (commented):
I am not certain about the changes in `hf_transformers_utils.py`. Need another peer review.

My overall thought on this issue is, 
1) if top_k = 50 affects the speculative decoding process, then we can revert the `--sampling-defaults` to `openai`.

2) if we think the model sampling params from generation_config.json is useful, then we can extract only params that are set in generation_config.json, if there are params that are not set, we fallback to the default of openai. In this case, `top_k=-1`.

python/sglang/srt/configs/model_config.py
adarshxs: The problem here is `hf_generation_config` fills in the missing fields with transformers internal defaults(top_k=50, temperature=1, etc)
It will return GenerationConfig that always looks complete, we cannot tell if that came from hf default or generation config. `_explicit_keys` tells which of those fields were explicitly present in the models own generation_config.json so we can set them manually

adarshxs:
1) yes top_k is harming specdec as of now. My theory is it's reducing token acceptance rate because it deviates from models greedy alignment. We might be better off defaulting to openai defaults first. 

2) yes this is exactly what's happening. We use explicit keys to derive those that are set and only pass them

python/sglang/srt/configs/model_config.py
CatherineSue: How did you fix this?
At this point, if we have a param that is not set by generation_config.json, we should use the default from openai.

adarshxs: @CatherineSue we manually check the generation_config.json for any keys set by it and store in explicit_keys. If not set, we default to openai values.

PR #11611, opened 2025-10-14
Support shared experts overlap in cutlass moe

fzyzcjy:
## Checklist


trevor-m (commented):
Hi @fzyzcjy - this looks like it's doing the same overlap as `forward_normal_dual_stream()`. Could you help explain the difference?

fzyzcjy:
@trevor-m forward_normal_dual_stream seems to overlap shared experts w/ routed experts, thus it is likely shared is oervlapped with dispatch and the first routed gemm. while this pr overlaps shared w/ combine, which has little overhead b/c combine takes few SM

trevor-m (approved):
Thank you for the explanation, and it looks good to me.

PR #11360, opened 2025-10-09
[BugFix] test_mla_fp8.py fails on Cublas 12.9

Liu-congo:
## Motivation
fix #11272 

## Modifications
replace the per_tensor_quant_mla_fp8 used in dsv2 with the to_float8 in sgl-kernel/tests/test_bmm_fp8.py

## Benchmarking
cmd: python3 test/srt/test_mla_fp8.py
raw implementation on cublas12.8.4.1
{'en': np.float64(0.86), 'en:std': np.float64(0.34698703145794946), 'group_latin': np.float64(0.86), 'group_latin:std': np.float64(0.34698703145794946), 'score:std': np.float64(0.34698703145794946), 'score': np.float64(0.86)}
Total latency: 26.120 s
Score: 0.860

raw implementation on cublas12.9.1.4
failed

current implementation on cublas12.8.4.1:
{'en': np.float64(0.888), 'en:std': np.float64(0.31536645351083237), 'group_latin': np.float64(0.888), 'group_latin:std': np.float64(0.31536645351083237), 'score:std': np.float64(0.31536645351083237), 'score': np.float64(0.888)}
Total latency: 14.972 s(very magic, adding 3 more test result: 30s, 26s, 29s, the true performance might be around 28s?)
Score: 0.888

current implementation on cublas12.91.4
{'en': np.float64(0.864), 'en:std': np.float64(0.3427885645700568), 'group_latin': np.float64(0.864), 'group_latin:std': np.float64(0.3427885645700568), 'score:std': np.float64(0.3427885645700568), 'score': np.float64(0.864)}
Total latency: 25.954 s
Score: 0.864

## Plan
I think the problem might caused by the triton implemented per_tensor_quant_mla_fp8, further 

## Checklist

- [ ] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
- [ x ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.ai/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.ai/developer_guide/contribution_guide.html#benchmark-the-speed).

Issue #11272, opened 2025-10-06
[Bug] test_mla_fp8.py fails on Cublas 12.9


Fridge003:
### Checklist

- [ ] 1. I have searched related issues but cannot get the expected help.
- [ ] 2. The bug has not been fixed in the latest version.
- [ ] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [ ] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [ ] 5. Please use English, otherwise it will be closed.

### Describe the bug

In the newest docker image, the version of `nvidia-cublas-cu12` is 12.9.1.4.
When running MLA fp8 CI test with
```python
python3 test/srt/test_mla_fp8.py
```
Following error on bmm kernel will happen during cuda graph capturing
```bash
[2025-10-06 20:13:00] Scheduler hit an exception: Traceback (most recent call last):
  File "/sgl-workspace/sglang/python/sglang/srt/model_executor/cuda_graph_runner.py", line 366, in __init__
    self.capture()
  File "/sgl-workspace/sglang/python/sglang/srt/model_executor/cuda_graph_runner.py", line 485, in capture
    ) = self.capture_one_batch_size(bs, forward)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/sgl-workspace/sglang/python/sglang/srt/model_executor/cuda_graph_runner.py", line 659, in capture_one_batch_size
    run_once()
  File "/sgl-workspace/sglang/python/sglang/srt/model_executor/cuda_graph_runner.py", line 648, in run_once
    logits_output_or_pp_proxy_tensors = forward(
                                        ^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch/utils/_contextlib.py", line 120, in decorate_context
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/sgl-workspace/sglang/python/sglang/srt/models/deepseek_v2.py", line 2562, in forward
    hidden_states = self.model(
                    ^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch/nn/modules/module.py", line 1773, in _wrapped_call_impl
[...]

PR #5370, opened 2025-04-14
[perf] experimental enhance fp8 per-tensor quant

Alcanderian:
## Motivation

1. fix `input_to_float8` precision
2. impl mla absrob specialized triton per tensor quant fp8
3. optimize out Tensor.aminmax which call Tensor.contiguous inside

## Unittest

```bash
python3 python/sglang/test/test_block_fp8.py            
test_per_tensor_quant_mla_fp8 (__main__.TestPerTensorQuantMlaFP8) ... ok
```

## Perf

About 22% boost🚀 on decode phase

### 1-batch

cmd: `python3 -m sglang.bench_one_batch --model RedHatAI/DeepSeek-Coder-V2-Lite-Instruct-FP8 --batch-size 1 --input-len 128 --output-len 128 --trust-remote-code`

- before
```
Benchmark ...
Prefill. latency: 0.07211 s, throughput:   1775.16 token/s
Decode.  latency: 0.00735 s, throughput:    136.03 token/s
Decode.  latency: 0.00735 s, throughput:    136.01 token/s
Decode.  latency: 0.00736 s, throughput:    135.86 token/s
Decode.  latency: 0.00732 s, throughput:    136.53 token/s
Decode.  latency: 0.00734 s, throughput:    136.17 token/s
Decode.  median latency: 0.00732 s, median throughput:    136.69 token/s
Total. latency:  1.003 s, throughput:    255.15 token/s
```

- after
```
Benchmark ...
Prefill. latency: 0.06900 s, throughput:   1855.09 token/s
Decode.  latency: 0.00632 s, throughput:    158.16 token/s
Decode.  latency: 0.00633 s, throughput:    157.97 token/s
Decode.  latency: 0.00631 s, throughput:    158.36 token/s
Decode.  latency: 0.00629 s, throughput:    159.01 token/s
Decode.  latency: 0.00632 s, throughput:    158.20 token/s
Decode.  median latency: 0.00628 s, median throughput:    159.35 token/s
Total. latency:  0.868 s, throughput:    295.07 token/s
```

### 64-batch

cmd: `python3 -m sglang.bench_one_batch --model RedHatAI/DeepSeek-Coder-V2-Lite-Instruct-FP8 --batch-size 64 --input-len 128 --output-len 128 --trust-remote-code`

- before
```
Benchmark ...
Prefill. latency: 0.10302 s, throughput:  79520.42 token/s
Decode.  latency: 0.00911 s, throughput:   7026.19 token/s
[...]

PR #11612, opened 2025-10-14
[BugFix] replace the input_to_float8 used in dsv2

Liu-congo:
## Motivation
this pr is a follow-up for #11272 and #11360 
in previous pr, i fix the ci failure using input_to_float8, after some further exploration, i find that the failure is caused by the improvement in #5549 
the function was first proposed in #5370 , 
in it's raw implementation:it create a "x_s = torch.zeros((1,), dtype=torch.float32, device=x.device)" when being called. x_s is the scaling factor for the input matrix, also the B_scale in bmm_fp8
then in #5549, it introduce a bumpallocator to allocate buffer beforehand to enhance throughput, however it caused the ci failure in test/srt/test_mla_fp8 under cublas12.9

## Changes
the lastest fix(#11360) for #11272 is adding version check for cublas12.9, and replace the per_tensor_quant_mla_fp8 with input_to_float8 if needed
now, i deprecated the input_to_float8 function. Substitute the 'zero_allocator.allocate(1)' with 'torch.zeros((1，), dtype=torch.float32, device=x.device)'.
i think the root cause of #11272 lies in cublas, but i'm not quite familiar with that : (, waiting for help to solve the issue throughly.

## Minimal Script to reproduce the issue in #11272 
the following script can reproduce the issue easily.
a core thing is that if the loop body executes only once, the bmm_fp8 function can work normally, however when it executes more than once, it would throw exception 'RuntimeError: bmm_fp8_internal_cublaslt failed: the requested functionality is not supported'. So far, i think it might have something to do with the pointer of q_nope_scale.
```python
import torch
from sglang.srt.layers.quantization.fp8_kernel import per_tensor_quant_mla_fp8
from sglang.srt.utils import BumpAllocator
from sgl_kernel import bmm_fp8

zero_allocator = BumpAllocator(16, dtype=torch.float32, device='cuda')
for i in range(16):
    q = torch.randn((256, 16, 192)).to(torch.float8_e4m3fn).cuda()
    q_nope, q_pe = q.split(
                    [128, 64], dim=-1
                )
    q_nope_val, q_nope_scale = per_tensor_quant_mla_fp8(
                        q_nope.transpose(0, 1),
                        zero_allocator.allocate(1),
                    )
    print(q_nope_scale.data_ptr(), q_nope_scale)
    w_kc = torch.randn((16, 512, 128)).to(torch.float8_e4m3fn).cuda()
    w_kc = w_kc.transpose(1, 2)
    w_scale = torch.Tensor([0.0009]).to(torch.float32).to('cuda')
    q_nope_out = bmm_fp8(
                    q_nope_val, w_kc, q_nope_scale, w_scale, torch.bfloat16
                )
```

## Checklist

- [ x ] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ x ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ x ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.ai/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.ai/developer_guide/contribution_guide.html#benchmark-the-speed).


Liu-congo:
@Fridge003 hello, could you please help to review the pr? many thanks

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -1576,14 +1575,14 @@ def forward_absorb_prepare(
...
             # TODO fix the per_tensor_quant_mla_fp8 for cublas 12.9
-            if _is_cublas_ge_129:
-                q_nope_val, q_nope_scale = input_to_float8(
-                    q_nope.transpose(0, 1), torch.float8_e4m3fn
-                )
-            else:
-                q_nope_val, q_nope_scale = per_tensor_quant_mla_fp8(
-                    q_nope.transpose(0, 1), zero_allocator.allocate(1)
-                )
+            q_nope_val, q_nope_scale = per_tensor_quant_mla_fp8(
```
Fridge003, line 1622: Please add a comment here that explains a little bit and  points to this PR

Liu-congo: ok, i've added now.

Fridge003 (approved):

PR #11616, opened 2025-10-14
Fix mamba radix cache eviction logic in `alloc_req_slots`

rogeryoungh:
## Motivation

This pull request addresses an issue with the cache eviction logic introduced in #11214. The eviction step is implemented in `schedule_batch.py`​​, but due to the recent refactoring in #11313 (which moved `alloc_req_slots` and its associated logic to `common.py`), the eviction logic wasn't being called. This issue was also reported in https://github.com/sgl-project/sglang/pull/11214#issuecomment-3396063269 .

This patch ensures that evictions are correctly called from the refactored alloc_req_slots function.

## Modifications

Moving eviction logic to `common.py`.

## Checklist


python/sglang/srt/mem_cache/common.py
```diff
@@ -330,6 +331,15 @@ def alloc_for_extend(
     bs = len(batch.reqs)
     prefix_tensors = [r.prefix_indices for r in batch.reqs]
 
+    if isinstance(batch.req_to_token_pool, HybridReqToTokenPool):
+        mamba_available_size = batch.req_to_token_pool.mamba_pool.available_size()
+        if mamba_available_size < bs:
```
yizhang2077: I think you need move this logic into `alloc_req_slots` in this file

rogeryoungh: updated

yizhang2077 (approved):
LGTM cc @hnyls2002

hzh0425 (approved):

hanming-lu (approved):

merrymercy (approved):

PR #11621, opened 2025-10-14
[router] Fix response api related spec

key4ng:
## Motivation
- remove to_sampling_params
- Remove DEFAULT_MODEL_NAME in rerank API, consolidate all default value for model field in all requests
- Remove generate_request_id for ResponsesRequest, add it in middleware.rs

## Checklist

### DIFF
--- a/sgl-router/src/middleware.rs
+++ b/sgl-router/src/middleware.rs
@@ -2,7 +2,7 @@ use axum::{
     body::Body, extract::Request, extract::State, http::header, http::HeaderValue,
     http::StatusCode, middleware::Next, response::IntoResponse, response::Response,
 };
-use rand::Rng;
+use rand::{Rng, RngCore};
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
 use std::time::Duration;
@@ -61,8 +61,21 @@ pub async fn auth_middleware(
     Ok(next.run(request).await)
 }
 
+/// Generate responses API ID with format "resp_" + 50 hex characters
+pub fn generate_responses_api_id() -> String {
+    let mut rng = rand::rng();
+    let mut bytes = [0u8; 25];
+    rng.fill_bytes(&mut bytes);
+    let hex_string: String = bytes.iter().map(|b| format!("{:02x}", b)).collect();
+    format!("resp_{}", hex_string)
+}
+
 /// Generate OpenAI-compatible request ID based on endpoint
 fn generate_request_id(path: &str) -> String {
+    if path.contains("/v1/responses") {
+        return generate_responses_api_id();
+    }
+
     let prefix = if path.contains("/chat/completions") {
         "chatcmpl-"
     } else if path.contains("/completions") {
--- a/sgl-router/src/middleware.rs
+++ b/sgl-router/src/middleware.rs
@@ -2,7 +2,7 @@ use axum::{
     body::Body, extract::Request, extract::State, http::header, http::HeaderValue,
     http::StatusCode, middleware::Next, response::IntoResponse, response::Response,
 };
-use rand::{Rng, RngCore};
+use rand::Rng;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
 use std::time::Duration;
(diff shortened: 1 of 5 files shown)


sgl-router/src/middleware.rs
```diff
@@ -61,8 +61,21 @@ pub async fn auth_middleware(
...
+    let mut bytes = [0u8; 25];
+    rng.fill_bytes(&mut bytes);
+    let hex_string: String = bytes.iter().map(|b| format!("{:02x}", b)).collect();
+    format!("resp_{}", hex_string)
+}
+
 /// Generate OpenAI-compatible request ID based on endpoint
 fn generate_request_id(path: &str) -> String {
+    if path.contains("/v1/responses") {
+        return generate_responses_api_id();
```
CatherineSue: Can we not use "resp_" and use L90 to L99?

slin1237:
do not merge until @CatherineSue approves

sgl-router/src/middleware.rs
key4ng: yeah we can, it's just the length of character doesn't match the openai format==

CatherineSue: I see. Emm is this a bug in L90 to L99? I thought default length is 24. If we need to have 0...25, we can fix it in L90 to L99. Keep it consistent for all endpoints

CatherineSue: Does resp_id by default have 48 chars? It seems chatcmpl has 24?

key4ng: it's by default having 50 chars after `resp_`

key4ng: wait a second, oai response api doesn't have request_id field, it seems a sglang concept

CatherineSue: ok, regarding the length, we could do something like:

```
fn generate_request_id(path: &str) -> String {
    let (prefix, length) = if path.contains("/v1/responses") {
        ("resp_", 48)  // or 50, not sure which
    } else if path.contains("/chat/completions") {
        ("chatcmpl-", 24)
    } else if path.contains("/completions") {
        ("cmpl-", 24)
    } else if path.contains("/generate") {
        ("gnt-", 24)
    } else {
        ("req-", 24)
    };

    // Generate a random string similar to OpenAI's format
    let chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
    let mut rng = rand::rng();
    let random_part: String = (0..length)
        .map(|_| {
            let idx = rng.random_range(0..chars.len());
            chars.chars().nth(idx).unwrap()
        })
        .collect();

    format!("{}{}", prefix, random_part)
}
```

key4ng: I guess we don't have to follow openai format if it's a sglang concept?
https://github.com/sgl-project/sglang/blob/c224a4c6cc466c70694ce2ba29f1b7ff777392cc/python/sglang/srt/entrypoints/openai/protocol.py#L916-L920

CatherineSue: it seems chatcmpl- by default is 29.
Screenshot 2025-10-14 at 12 04 44 PM

key4ng: I'm a little confused about the id and request_id
Screenshot 2025-10-14 at 12 08 39 PM

CatherineSue: how did you get request_id?

sgl-router/src/protocols/spec.rs
```diff
@@ -1823,7 +1724,10 @@ impl ResponsesResponse {
         usage: Option<UsageInfo>,
     ) -> Self {
         Self {
-            id: request.request_id.clone(),
+            id: request
+                .request_id
+                .clone()
+                .expect("request_id should be set by middleware"),
```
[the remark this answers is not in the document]
key4ng: it should throw error if the middleware is misconfigured

sgl-router/src/middleware.rs
key4ng: fixed it by using `resp-<24 char>`, pls have another look

CatherineSue (approved):

PR #11638, opened 2025-10-15
[router] add worker self discovery for metadata

slin1237:
- Auto-discover worker metadata for both HTTP and gRPC endpoints before registration, capturing model/tokenizer paths, weight version,   sampling flags, etc. via WorkerDiscovery helpers
  - Made create_basic_worker async so each worker merges discovery data with manual labels and reuses an already-connected gRPC client  when available
  - Let the dynamic /add_worker flow rely on the discovered worker metadata instead of forcing model_id="unknown" so policies pick up the actual model id automatically 
  - Added unit coverage for the new model-id derivation logic to guarantee consistent fallbacks from served name or model path 

output
```
curl localhost:8080/workers
{
   "workers":[
      {
         "id":"grpc://localhost:30000",
         "url":"grpc://localhost:30000",
         "model_id":"Llama-3.1-8B-Instruct",
         "priority":50,
         "cost":1.0,
         "worker_type":"regular",
         "is_healthy":true,
         "load":0,
         "connection_mode":"Grpc { port: None }",
         "tokenizer_path":"/raid/models/meta-llama/Llama-3.1-8B-Instruct",
         "metadata":{
            "tokenizer_path":"/raid/models/meta-llama/Llama-3.1-8B-Instruct",
            "max_req_input_len":"131066",
            "weight_version":"default",
            "model_type":"transformer",
            "max_context_length":"414382",
            "served_model_name":"/raid/models/meta-llama/Llama-3.1-8B-Instruct",
            "model_path":"/raid/models/meta-llama/Llama-3.1-8B-Instruct",
            "model_id":"Llama-3.1-8B-Instruct",
            "vocab_size":"128256",
            "is_generation":"true"
         }
      }
   ],
   "total":1,
   "stats":{
      "prefill_count":0,
      "decode_count":0,
      "regular_count":1
   }
}
```

## Checklist

### DIFF
--- a/sgl-router/src/core/worker_manager.rs
+++ b/sgl-router/src/core/worker_manager.rs
@@ -1228,9 +1228,8 @@ impl WorkerManager {
             Ok(server_info) => {
                 if let Some(model_id) = server_info.model_id {
                     if !model_id.is_empty() {
-                        discovery
-                            .labels
-                            .insert("model_id".to_string(), model_id.to_string());
+                        let normalized = Self::normalize_model_identifier(&model_id);
+                        discovery.labels.insert("model_id".to_string(), normalized);
                     }
                 }
                 if let Some(model_path) = server_info.model_path {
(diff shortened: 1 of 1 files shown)


sgl-router/src/core/worker_manager.rs
```diff
@@ -1084,6 +1121,306 @@ impl WorkerManager {
...
+            Ok(model_info) => {
+                if let Some(model_path) = model_info.get("model_path").and_then(|v| v.as_str()) {
+                    if !model_path.is_empty() {
+                        discovery
+                            .labels
+                            .insert("model_path".to_string(), model_path.to_string());
+                    }
+                }
+                if let Some(tokenizer_path) =
+                    model_info.get("tokenizer_path").and_then(|v| v.as_str())
```
key4ng, line 1149: could we use constant for `tokenizer_path` and other keys, or model_info can be a typed struct?

key4ng (approved):

Issue #11630, opened 2025-10-14
[Bug] RadixCacheCpp Error when using


narutolhy:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [ ] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Describe the bug

When I want to use RadixCacheCpp, it's failed.

allocator = tree_cache.token_to_kv_pool_allocator
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'RadixCacheCpp' object has no attribute 'token_to_kv_pool_allocatord

### Reproduction

`python` `test/srt/test_cpp_radix_cache.py`

### Environment

`python` `test/srt/test_cpp_radix_cache.py`

narutolhy:
Hi @cctry , When I tried to use the cpp version of radix tree, an error occurred. I think it might be related to this pr https://github.com/sgl-project/sglang/pull/11313.
@DarkSharpness is the feature owner of radix cpp tree.
Hi DarkSharpness, I saw you have created a test of radix tree. Could you add it to test/srt/run_suite.py? It will run when new pr want to merge after you add it. Thank you.

cctry:
Should be fixed in https://github.com/sgl-project/sglang/pull/11645

This also fixes another issue of type mismatch in `cache_finished_req` and `cache_unfinished_req` cc @DarkSharpness @xiezhq-hermann

PR #11645, opened 2025-10-15
Fix CPP Radix Cache and add test to CI

cctry:
## Motivation

Issue https://github.com/sgl-project/sglang/issues/11630 https://github.com/sgl-project/sglang/issues/13085

## Modifications
- Fix the naming issues introduced in https://github.com/sgl-project/sglang/pull/11313
- Add explict type conversion in `radix_cache_cpp.py` similar to other radix cache
- Add `test/srt/test_cpp_radix_cache.py` to CI.

## Checklist


DarkSharpness:
LGTM. But I don't think we should rely on the assumption that `tree_cache` must have some property called `token_to_kv_pool_allocator`.

Personally, I would choose to **explicitly** pass `token_to_kv_pool_allocator` to functions like `alloc_token_slots` and `evict_from_tree_cache`, instead of acquiring from the property of a `tree_cache`.

airMeng:
> LGTM. But I don't think we should rely on the assumption that `tree_cache` must have some property called `token_to_kv_pool_allocator`.
> 
> Personally, I would choose to **explicitly** pass `token_to_kv_pool_allocator` to functions like `alloc_token_slots` and `evict_from_tree_cache`, instead of acquiring from the property of a `tree_cache`.

https://github.com/sgl-project/sglang/pull/11313#discussion_r2443861779 seems I already encountered the same issue

hnyls2002 (approved):

PR #11655, opened 2025-10-15
[DeepseekV32] Enable flashmla_prefill kernel with fp8 kvcache

hlu1:
## Motivation

Add logics to dequant the kvcache from fp8 to bf16 in a separate kernel and use `flashmla_prefill` kernel with fp8 kvcache.

- [x] Verify accuracy and perf on both H200 and B200
- [x] Fuse topk transformations with the topk kernel.
- [x] Add a new triton fused kernel `dequantize_k_cache_paged`.
- [x] Add heuristics to switch flashmla_prefill/flashmla_decode automatically. Also add `flashmla_auto` mode and use it as the default mode for prefill when fp8 kvcache is enabled.
- [x] Adjust default settings for both fp8/bf16 kvcache on Blackwell. 
- [ ] Make it compatible with mtp. Currently it dispatches prefill to `flashmla_kv` with fp8 kvcache when spec decoding is detected.

flashmla_decode (before)

image

flashmla_prefill with no kvcache reuse or chunked prefill (after)
image

flashmla_prefill with kvcache reuse or chunked prefill (after)
image

## Accuracy Tests

gpqa (with an early fp4 checkpoint)
with fp8 kvcache
before: ['0.768', '0.823', '0.773', '0.783']
after: ['0.788', '0.798', '0.773', '0.798']

with bf16 kvcache
After: ['0.818', '0.818', '0.828', '0.758']

## Benchmarking and Profiling
With fp4 checkpoint:
```
python -m sglang.bench_one_batch_server --model-path $MODEL --tp 4 --dp 4 --enable-dp-attention --batch 64 --input-len 8192 --output-len 1  --nsa-prefill flashmla_prefill
```
before:
input throughput: 12536.68 tok/s

after:
input throughput: 14610.87 tok/s

## Checklist


python/sglang/srt/server_args.py
```diff
@@ -1030,7 +1037,8 @@ def _handle_model_specific_adjustments(self):
                     logger.warning("Setting KV cache dtype to fp8.")
```
Fridge003: Can we remove this limitation of fp8 kv cache on B200?
I mean when user doesn't specify the kv cache dtype(then kv cache dtype is set to `auto`), we can fix it to fp8. But if user wants to apply bf16 kvcache, he can also enable that through `--kv-cache-dtype bfloat16`?

python/sglang/srt/layers/attention/nsa/dequant_k_cache.py
```diff
@@ -159,5 +171,127 @@ def _dequantize_k_cache_fast_kernel(
...
+        dtype=torch.bfloat16,
+        device=quant_k_cache.device,
+    )
+
+    # cdiv(512 + 64, 128) = 5
+    num_blocks_per_token = triton.cdiv(dim_nope + dim_rope, group_size)
+    assert num_blocks_per_token == 5
+
+    assert dim_nope % group_size == 0
+    NUM_NOPE_BLOCKS = dim_nope // group_size  # 512 // 128 = 4
```
Fridge003: `NUM_NOPE_BLOCKS` and `num_tiles` are the same. Maybe remove duplication?

python/sglang/srt/layers/attention/nsa/dequant_k_cache.py
```diff
@@ -159,5 +171,127 @@ def _dequantize_k_cache_fast_kernel(
...
+
+    input_nope_q = quant_k_cache[:, :dim_nope]
+    # [:, 512:512+4*4] = [:, 512:528]
+    input_nope_s = quant_k_cache[:, dim_nope : dim_nope + num_tiles * 4].view(
+        torch.float32
+    )
+    # [:, 528:]
+    input_rope = quant_k_cache[:, dim_nope + num_tiles * 4 :].view(torch.bfloat16)
```
Fridge003, line 219: Just for checking, the indexing and view operation here will not launch kernels, is that correct?
If it launch its kernels, then we had better move the indexing and view operation to the triton kernel.

python/sglang/srt/layers/attention/nsa_backend.py
```diff
@@ -804,6 +879,21 @@ def forward_extend(
         elif NSA_PREFILL_IMPL == "flashmla_sparse":
             if q_rope is not None:
                 q_all = torch.cat([q_nope, q_rope], dim=-1)
+
+            # NSA_FLASHMLA_BACKEND_DECODE_COMPUTE_FP8 has no effect here,
+            # because the flashmla_prefill kernel doesn't support fp8 compute
```
Fridge003: should be `flashmla_sparse` kernel here

python/sglang/srt/server_args.py
```diff
@@ -1030,7 +1037,8 @@ def _handle_model_specific_adjustments(self):
...
 
                 if self.kv_cache_dtype == "fp8_e4m3":
-                    self.nsa_prefill_backend = "flashmla_kv"
+                    # flashmla_auto dispatches to flashmla_sparse/flashmla_kv based on hardware and heuristics
+                    self.nsa_prefill_backend = "flashmla_auto"
                     self.nsa_decode_backend = "flashmla_kv"
                     logger.warning(
                         "Setting NSA backend to flashmla_kv for FP8 KV Cache."
```
Fridge003: Should modify the warning log here.

python/sglang/srt/layers/attention/nsa_backend.py
```diff
@@ -181,6 +212,7 @@ def __init__(
         global NSA_PREFILL_IMPL, NSA_DECODE_IMPL
         NSA_PREFILL_IMPL = model_runner.server_args.nsa_prefill_backend
         NSA_DECODE_IMPL = model_runner.server_args.nsa_decode_backend
+        self.original_prefill_impl = model_runner.server_args.nsa_prefill_backend
```
Fridge003: We can use a variable `enable_auto_prefill_impl` here. And only trigger `self.set_nsa_prefill_impl` when `enable_auto_prefill_impl` is true. 
`self.original_prefill_impl` is a little confusing to me.

python/sglang/srt/layers/attention/nsa_backend.py
```diff
@@ -310,6 +348,23 @@ def init_forward_metadata(self, forward_batch: ForwardBatch):
                     )
                 ]
             )
+
+            page_table_1_flattened = torch.cat(
```
Fridge003: Are these operations only needed when transform method is `RAGGED`?
If so I think we need to skip them when method is `PAGED`

python/sglang/srt/layers/attention/nsa_backend.py
hlu1: Sounds good. I'll the name to `enable_auto_select_prefill_impl`. It reads slightly better

python/sglang/srt/layers/attention/nsa/dequant_k_cache.py
hlu1: Based on the pytorch implementation of `view_dtype` (https://github.com/pytorch/pytorch/blob/661a56002f5f747712378752cc172fafab7ee000/aten/src/ATen/native/TensorConversions.cpp#L872), it's a pure view op. 
I double checked nsys profile, it didn't trigger a copy before the kernel.
image

Fridge003:
@hlu1 The `Configuration Tips` section of [document](https://github.com/sgl-project/sglang/blob/main/docs/basic_usage/deepseek_v32.md) needs to be updated in a following PR after this PR get merged.

Fridge003 (approved):
Great work!

Fridge003:
@hlu1 Please fix the bug here https://github.com/sgl-project/sglang/actions/runs/18812434633/job/53676478829?pr=11655

hlu1:
> @hlu1 The `Configuration Tips` section of [document](https://github.com/sgl-project/sglang/blob/main/docs/basic_usage/deepseek_v32.md) needs to be updated in a following PR after this PR get merged.

Will do.



> @hlu1 Please fix the bug here https://github.com/sgl-project/sglang/actions/runs/18812434633/job/53676478829?pr=11655

Fixed.

Fridge003 (approved):
wonderful work!

PR #11659, opened 2025-10-15
docs: update sglang installation guide

zhyncs:
## Motivation

we don't need to use `[all]` anymore :)

## Checklist

### DIFF
--- a/docs/get_started/install.md
+++ b/docs/get_started/install.md
@@ -12,7 +12,7 @@ It is recommended to use uv for faster installation:
 ```bash
 pip install --upgrade pip
 pip install uv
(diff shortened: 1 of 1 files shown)


csahithi:
@zhyncs Thanks for this change! Just FYI, I've tried to install sglang using `uv pip install sglang --upgrade` and I land on `0.5.2`. 

On x86, If I add `--prerelease=allow` (which I believe is required because `torch_memory_saver` requires an rc version `0.0.9rc2`) then I land on `0.5.3rc0`. 
On ARM, I land only on `0.5.2`.
I can't seem to land on `0.5.3.post1` using this method.

ishandhanani:
^ yea i think we should keep the [all] for now

PR #11661, opened 2025-10-15
fix: replace buggy flashinfer fp4 gemm with sgl-kernel fp4 gemm

Qiaolin-Yu:
## Motivation

Flashinfer fp4 gemm is buggy. It will be super slow when the batch size is large.

image

And it's also slower than sgl-kernel fp4 gemm when bs=1

image

image

### Reproduce commands:
```
TRTLLM_ENABLE_PDL=1 python3 -m sglang.launch_server --model-path nvidia/DeepSeek-R1-FP4 --trust-remote-code --quantization modelopt_fp4 --tp 8  --speculative-algorithm=EAGLE   --disable-radix-cache

```

## Accuracy Tests

```
gsm8k

Accuracy: 0.955
Invalid: 0.000
Latency: 190.533 s
Output throughput: 710.890 token/s
```

## Checklist

### DIFF
--- a/python/sglang/srt/layers/quantization/modelopt_quant.py
+++ b/python/sglang/srt/layers/quantization/modelopt_quant.py
@@ -49,20 +49,21 @@ if TYPE_CHECKING:
     )
     from sglang.srt.single_batch_overlap import DownGemmOverlapArgs
 
+fp4_gemm = None
 if is_cuda():
+    from sgl_kernel import cutlass_scaled_fp4_mm as fp4_gemm
     from sgl_kernel import scaled_fp4_quant
 
+    enable_flashinfer_fp4_gemm = False
+
 try:
-    from flashinfer import mm_fp4 as fp4_gemm
     from flashinfer import reorder_rows_for_gated_act_gemm, shuffle_matrix_sf_a
 
-    enable_flashinfer_fp4_gemm = True
+    if not is_cuda():
+        from flashinfer import mm_fp4 as fp4_gemm
+
+        enable_flashinfer_fp4_gemm = True
 except ImportError:
-    if is_cuda():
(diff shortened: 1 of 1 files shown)


python/sglang/srt/layers/quantization/modelopt_quant.py
```diff
@@ -49,20 +49,21 @@
...
     from sgl_kernel import scaled_fp4_quant
 
+    enable_flashinfer_fp4_gemm = False
+
 try:
-    from flashinfer import mm_fp4 as fp4_gemm
     from flashinfer import reorder_rows_for_gated_act_gemm, shuffle_matrix_sf_a
 
-    enable_flashinfer_fp4_gemm = True
+    if not is_cuda():
```
zhyncs: When is_cuda is false, we should not use flashinfer fp4 gemm here.

Qiaolin-Yu: done

ispobock: do we need to keep flashinfer option here?

Qiaolin-Yu: I think we could merge this first. After flashinfer fixes this, we could revert this pr directly. wdyt?

zhyncs (changes requested):
actually we don't need to replace flashinfer fp4 gemm with sgl-kernel fp4 gemm, we can just use the `cutlass` backend for flashinfer fp4 gemm

zhyncs:
ref https://github.com/sgl-project/sglang/blob/cbac4997509f6a857655f7443de1a4833999f4e0/python/sglang/srt/layers/quantization/modelopt_quant.py#L862

PR #11664, opened 2025-10-15
Use trtllm_mla decode kernel for draft extend in speculative decoding

Qiaolin-Yu:
## Motivation

image
Before this pr, draft extend is using flashinfer_mla kernel. However, since flashinfer_mla doesn't support fp8, when using fp8 kv cache, it needs to dequantize kv to bf16, which is very slow (shown in figure).

After this pr,
image

```
python3 -m sglang.bench_serving --backend sglang-oai --dataset-name random --random-input-len 1024 --random-output-len 1024 --random-range-ratio 0.98 --num-prompts 5 --max-concurrency 1 --output-file res.jsonl
```

### Accuracy test
```

Accuracy: 0.962
Invalid: 0.000
Latency: 100.517 s
Output throughput: 1351.037 token/s
```

## Checklist


python/sglang/srt/layers/attention/trtllm_mla_backend.py
```diff
@@ -648,12 +655,17 @@ def forward_extend(
             )
 
             bmm1_scale = q_scale * k_scale * layer.scaling
-
-            seq_lens = (
-                forward_batch.seq_lens.to(torch.int32)
-                + forward_batch.spec_info.draft_token_num
-            )
-            max_seq_len = metadata.max_seq_len + forward_batch.spec_info.draft_token_num
+            if forward_batch.forward_mode.is_target_verify():
```
ispobock, line 949: can we move this update to the init_metadata (normal/cuda graph)?

python/sglang/srt/layers/attention/trtllm_mla_backend.py
```diff
@@ -372,6 +568,24 @@ def init_forward_metadata(self, forward_batch: ForwardBatch):
             self.forward_decode_metadata = TRTLLMMLADecodeMetadata(
                 block_kv_indices, max_seq_len_val
             )
+            if forward_batch.forward_mode.is_draft_extend():
+                max_seq = max(forward_batch.extend_seq_lens_cpu)
+                sum_seq_lens_q = forward_batch.spec_info.accept_length.sum().item()
```
ispobock: the item() will cause additional cpu sync, which will affect the cpu/gpu overlap
we can use accept_length cpu tensor in forward_batch  to do that.

python/sglang/srt/layers/attention/trtllm_mla_backend.py
```diff
@@ -372,6 +568,24 @@ def init_forward_metadata(self, forward_batch: ForwardBatch):
             self.forward_decode_metadata = TRTLLMMLADecodeMetadata(
                 block_kv_indices, max_seq_len_val
             )
+            if forward_batch.forward_mode.is_draft_extend():
+                max_seq = max(forward_batch.extend_seq_lens_cpu)
```
ispobock: tensor.max() is much faster than max(tensor)

python/sglang/srt/layers/attention/trtllm_mla_backend.py
```diff
@@ -457,6 +671,82 @@ def quantize_and_rope_for_fp8(
...
+            # Dynamic allocation for non-CUDA graph mode
+            output = torch.empty(
+                (total_tokens, tp_q_head_num, v_head_dim),
+                dtype=raw_out.dtype,
+                device=raw_out.device,
+            )
+
+        # Launch Triton kernel
+        BLOCK_SIZE = 64
+        grid = (batch_size * token_per_batch,)
```
ispobock: We can also parallelize it on head_num and head_dim to achieve better performance.

ispobock: It's better to have an unit test on this kernel in somewhere to make sure the correctness.

python/sglang/srt/layers/attention/trtllm_mla_backend.py
Qiaolin-Yu: somehow `accept_length_cpu` is not consistent with `accept_length`. I found `extend_seq_lens_cpu` is correct.

python/sglang/srt/layers/attention/trtllm_mla_backend.py
Qiaolin-Yu: but extend_seq_lens_cpu is a list

python/sglang/srt/layers/attention/trtllm_mla_backend.py
Qiaolin-Yu: done

ispobock (approved):

cicirori:
hi, just wandering why using padding here?

PR #11666, opened 2025-10-15
Manually flip deepep_mode for cuda_graph

zhuzilin:
## Motivation

When serving model with `--deepep-mode auto` and with cuda graph on, the current implementation won't flip the deepep mode from normal to low_latency and will trigger cuda illegal memory access in deepep or the following deepgemm kernel.

## Modifications

This PR is trying to add a manual flip by storing the deepep mode during capture and set it during replay.

## Checklist


fzyzcjy:
ci passed

python/sglang/srt/model_executor/cuda_graph_runner.py
```diff
@@ -655,6 +659,16 @@ def run_once():
...
             return logits_output_or_pp_proxy_tensors
 
+        # Resolve and pin DeepEP mode to avoid AUTO flipping between capture and replay
+        if get_moe_a2a_backend().is_deepep():
+            # Base runner handles decode/verify; not extend-in-batch
+            resolved = get_deepep_mode().resolve(is_extend_in_batch=False)
+            if resolved.is_low_latency():
+                DeepEPBuffer.set_dispatch_mode_as_low_latency()
+            else:
+                DeepEPBuffer.set_dispatch_mode_as_normal()
```
fzyzcjy: nit: extract to 

DeepEPBuffer.set_dispatch_mode(the mode variable)

python/sglang/srt/model_executor/cuda_graph_runner.py
```diff
@@ -655,6 +659,16 @@ def run_once():
             )
             return logits_output_or_pp_proxy_tensors
 
+        # Resolve and pin DeepEP mode to avoid AUTO flipping between capture and replay
+        if get_moe_a2a_backend().is_deepep():
```
fzyzcjy: the logic of

```
def init:
  self._captured_deepep_mode = None

def capture:
  buffer.set dispatch mode
  self.mode = mode

def replay:
  buffer.set dispatch mode
```

seems to occur 3 times

thus I personally think we can have:

```
class DeepEPCudaGraphRunnerAdapter:
  def init
  def capture
  def replay
```

and thus in cudagraphrunner, eagle runner, eagle extend runner, we only need to add a field of `self.blahblah=DeepEPCudaGraphRunnerAdapter()` and call it

fzyzcjy (commented):
LGTM, only nits

python/sglang/srt/layers/moe/token_dispatcher/deepep.py
```diff
@@ -235,6 +235,13 @@ def set_dispatch_mode_as_low_latency(cls):
             cls.clean_buffer()
         cls._dispatch_mode = DeepEPDispatchMode.LOW_LATENCY
 
+    @classmethod
+    def set_dispatch_mode(cls, mode: DeepEPMode):
+        if mode.is_low_latency():
+            cls.set_dispatch_mode_as_low_latency()
+        else:
+            cls.set_dispatch_mode_as_normal()
```
fzyzcjy, line 243:
```suggestion
        if mode.is_low_latency():
            cls.set_dispatch_mode_as_low_latency()
        elif mode.is_normal():
            cls.set_dispatch_mode_as_normal()
        else:
            raise Exception("unsupported mode")
```

fzyzcjy (approved):
LGTM

PR #11669, opened 2025-10-15
[RL] support weight update with DP attention

zhuzilin:
## Motivation

This PR is trying to enable dp attention for `init_weights_update_group` and `destroy_weights_update_group`.

## Checklist


python/sglang/srt/managers/tokenizer_communicator_mixin.py
```diff
@@ -358,10 +358,18 @@ async def init_weights_update_group(
...
+
+        results = await self.init_weights_update_group_communicator(obj)
+        if self.server_args.dp_size == 1:
+            result = results[0]
+            return result.success, result.message
+        else:
+            all_success = all([r.success for r in results])
+            all_message = [r.message for r in results]
+            all_message = " | ".join(all_message)
+            return all_success, all_message
```
fzyzcjy: maybe we only need to keep the `else` branch, since the `if` branch is just a special case of `else`

python/sglang/srt/managers/tokenizer_communicator_mixin.py
```diff
@@ -370,10 +378,18 @@ async def destroy_weights_update_group(
...
+
+        results = await self.destroy_weights_update_group_communicator(obj)
+        if self.server_args.dp_size == 1:
+            result = results[0]
+            return result.success, result.message
+        else:
+            all_success = all([r.success for r in results])
+            all_message = [r.message for r in results]
+            all_message = " | ".join(all_message)
+            return all_success, all_message
```
fzyzcjy: seems this pattern occurs 3 times, maybe we can make a simple function to handle this repetition

python/sglang/srt/managers/tokenizer_communicator_mixin.py
zhuzilin: fixed.

python/sglang/srt/managers/tokenizer_communicator_mixin.py
zhuzilin: I've extracted a `merge_communicator_results` to `python/sglang/srt/utils/common.py`, but I'm not sure if this is the appropriate place and the right naming...

fzyzcjy: I would personally put it in 

```
class _Communicator(Generic[T]):
  @staticmethod
  def merge_results(): ...
```

(note the staticmethod, thus it is like a helper utility)

fzyzcjy (approved):

PR #11673, opened 2025-10-15
feat: return partial generation results when aborting requests in waiting queue

guoyuhong:
## Motivation

Partial rollout is a common practice in reinforcement learning (RL). Sglang provides abort_request to terminate incomplete inference requests. However, when a request is aborted while still in the waiting queue, the current implementation returns an un-prefilled (i.e., empty) result. While this may be acceptable for most online serving scenarios, it becomes problematic in RL settings.

In RL, early stages of inference often involve very large batches, which can later lead to insufficient KV cache as sequence lengths grow. This may cause some requests to be retracted from the inference queue even though they have already produced partial results. If we hardcode these aborted requests to return empty responses, previously computed partial results are wasted. With partial rollout in RL, we can continue inference in the next round using already-computed outputs. Therefore, this PR aims to ensure that requests with partial results return those results correctly instead of returning empty responses.

## Modifications

Modified _handle_abort_req in tokenizer_manager.py to return partial generation results when an aborted request has already produced some output.

## Accuracy Tests
We abort a benchmark test when there are 2 retracted requests.
image

Without this PR, the results for requests in both running and waiting queues are as follows:
image

With this PR applied, results without --return-logprob are:
image

With this PR applied, results with --return-logprob enabled are:
image

## Checklist


python/sglang/srt/managers/tokenizer_manager.py
fzyzcjy, line 1: could you please add a test to ensure this feature is protected

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -1417,6 +1417,51 @@ def _handle_batch_output(
             if self.crash_dump_folder and state.finished and state.obj.log_metrics:
                 self.record_request_for_crash_dump(state, out_dict)
 
+    def add_logprob_style_to_meta_info(
```
fzyzcjy: nit: what about

```suggestion
    def add_logprob_to_meta_info(
```

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -1493,18 +1518,13 @@ def convert_logprob_style(
...
-                return_text_in_logprobs,
-            )
-            meta_info["output_token_ids_logprobs"] = (
-                self.detokenize_top_logprobs_tokens(
-                    state.output_token_ids_logprobs_val,
-                    state.output_token_ids_logprobs_idx,
-                    return_text_in_logprobs,
-                )
-            )
+        self.add_logprob_style_to_meta_info(
```
fzyzcjy: tiny nit

```suggestion

        self.add_logprob_style_to_meta_info(
```

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -1683,25 +1703,39 @@ def _handle_abort_req(self, recv_obj: AbortReq):
...
+        if getattr(state.obj, "return_logprob", False):
+            self.add_logprob_style_to_meta_info(
+                meta_info,
+                state,
+                state.obj.top_logprobs_num,
+                state.obj.token_ids_logprob,
+                state.obj.return_text_in_logprobs
+                and not self.server_args.skip_tokenizer_init,
+            )
+        if len(state.text) == 0:
```
fzyzcjy: nit: shall we unify the if-else

fzyzcjy (commented):
lgtm, only some nits

python/sglang/srt/managers/tokenizer_manager.py
guoyuhong: applied.

python/sglang/srt/managers/tokenizer_manager.py
guoyuhong: Constructing this unit test is slightly challenging. After enabling SGLANG_TEST_RETRACT, even though there are retracted requests, they immediately re-enter the running queue via prefill. Therefore, we had to introduce an additional environment variable, SGLANG_TEST_RETRACT_NO_PREFILL_BS, to control and prevent some of the retracted requests from entering the running queue.

guoyuhong: A case is added to test_abort.py

image

guoyuhong:
@fzyzcjy do you have any more comments?

fzyzcjy (approved):

ispobock (approved):
We can merge it after CI passed.

PR #11688, opened 2025-10-16
[router] Add Configurable L0 and L1 Tokenizer Caching

slin1237:
## Summary

This PR introduces a two-level tokenizer caching system that provides significant performance improvements for tokenization operations. The implementation features an L0 exact-match cache and an innovative L1 special-token boundary prefix cache

## Motivation

Tokenization is a frequent operation in the router, and many requests share common patterns:
- **Repeated requests**: Identical prompts or messages are tokenized multiple times
- **Common prefixes**: System prompts, chat templates, and conversation history create shared prefix patterns
- **Performance opportunity**: Caching tokenization results can significantly reduce redundant computation

The two-level cache design balances fast exact-match lookups (L0) with intelligent prefix reuse (L1).

### L0 Cache (Exact Match)
- **Purpose**: Fast lookup for identical input strings
- **Implementation**: DashMap-based concurrent hash map with lock-free reads
- **Configuration**: `l0_max_entries` (default: 10,000 entries, ~22MB memory)
- **Use case**: Repeated identical prompts, common system messages
- **Eviction**: Simple arbitrary eviction when capacity reached

### L1 Cache (Special Token Boundary Prefix Cache)
- **Purpose**: Reuse tokenization results for shared prefixes at special token boundaries
- **Implementation**: Re-tokenization approach - caches prefix tokens by re-tokenizing at ALL special token boundaries
- **Configuration**: `l1_max_memory` (default: 50MB)
- **Use case**: Chat conversations with shared system prompts, multi-turn interactions
- **Eviction**: LRU eviction with memory tracking
- **Correctness**: Guarantees 100% correctness by re-tokenizing prefixes (not storing raw strings)

#### How L1 Cache Works

L1 cache is a **special-token boundary prefix cache** that caches tokenization results at every special token boundary. Special tokens (like `<|im_start|>`, `<|im_end|>`, `<|eot_id|>`) are atomic in BPE tokenizers (`special: true`, `normalized: false`), making them the ONLY safe split points that guarantee correctness.

**Why Special Tokens**

BPE tokenizers make context-dependent merge decisions, so `tokenize(prefix + suffix) != tokenize(prefix) + tokenize(suffix)` for arbitrary boundaries. However, special tokens are atomic and protected from normalization/merging, guaranteeing:

```
tokenize(prefix + suffix) == tokenize(prefix) + tokenize(suffix)
```

when splitting at special token boundaries.

**Cache Strategy - Re-Tokenization with All-Boundaries Approach**

The L1 cache uses a **re-tokenization approach** to guarantee correctness. When inserting into the cache after a full tokenization:

1. **Find all special token boundaries** in the input string
2. **For each boundary**: Extract the prefix substring up to that boundary
3. **Re-tokenize the prefix** to get the exact token sequence (BPE-safe)
4. **Cache the prefix tokens** with Blake3 hash of the prefix string as the key

This approach ensures 100% correctness because we never assume `tokenize(prefix) + tokenize(suffix) == tokenize(prefix + suffix)`. Instead, we always re-tokenize the prefix to get the actual token sequence that would result from tokenizing `prefix + suffix` up to the boundary.

```
Input: "<|im_start|>system\nYou are helpful.<|im_end|><|im_start|>user\nHello!<|im_end|>"

Special token boundaries found:
1. After "<|im_start|>" at position 13
2. After "<|im_end|>" at position 45
3. After "<|im_start|>" at position 58
4. After "<|im_end|>" at position 72

For each boundary:
- Extract prefix string (e.g., "<|im_start|>system\nYou are helpful.<|im_end|>")
- Re-tokenize prefix → get exact token IDs
- Cache: hash(prefix_string) → prefix_tokens
```

**Cache Lookup Example:**
```
Input: "<|im_start|>system\nYou are helpful.<|im_end|><|im_start|>user\nWhat is 2+2?<|im_end|>"

1. Find all special token boundaries in input string
2. For each boundary (longest to shortest):
   - Extract prefix substring up to boundary
   - Hash prefix with Blake3
   - Look up hash in cache
   - If HIT: Use cached prefix tokens + tokenize remaining suffix → merge and return
   - If MISS: Try next shorter prefix
3. If no prefix match:
   - Tokenize full string
   - Re-tokenize and cache prefixes at ALL boundaries for future requests
```

**Key Performance Insight**: On cache hit, we avoid re-tokenizing the prefix. On cache miss, we pay the cost of re-tokenizing prefixes once during insertion, then all future requests with that prefix benefit from the cached tokens.

### Opt-In
Both caches are **disabled by default** to maintain backward compatibility. The `CachedTokenizer` wrapper is only created when at least one cache is enabled; otherwise, the base tokenizer is used directly.

## Changes Made

**Tokenizer Factory (`src/tokenizer/factory.rs`)**
- Fixed special token detection in HuggingFace tokenizer
- Properly extracts `added_tokens` with `special: true` property
- Ensures special tokens are correctly identified for L1 cache

### CLI & Configuration

**CLI Arguments (`src/main.rs`)**
```bash
--tokenizer-cache-enable-l0              # Enable L0 exact-match cache
--tokenizer-cache-l0-max-entries <N>     # L0 cache size (default: 10000)
--tokenizer-cache-enable-l1              # Enable L1 prefix cache
--tokenizer-cache-l1-max-memory <BYTES>  # L1 max memory (default: 52428800)
```

## Bug Fixes

**Special Token Detection** (`src/tokenizer/factory.rs`)
- Fixed HuggingFace tokenizer special token extraction
- Properly filters `added_tokens` for `special: true` property
- Ensures L1 cache has correct special tokens to work with

## Performance Results

benchmarks using Qwen3-4B-Instruct-2507 tokenizer on different workloads:

### Benchmark Scenarios

| Scenario | Configuration | Throughput | Latency | L1 Hit Rate | L0 Hit Rate | Speedup vs Baseline |
|----------|---------------|------------|---------|-------------|-------------|---------------------|
| **Realistic Chat** (95%+ prefix reuse) | L0+L1 | 18,751 ops/sec | 53.3µs | 99.2% | 0.8% | **18.2x** |
| **Customer Service** (100% prefix reuse) | L1-only | 21,359 ops/sec | 46.8µs | 88.9% | N/A | **22.7x** |
| **Customer Service** (100% prefix reuse) | L0+L1 | 19,494 ops/sec | 51.2µs | 87.5% | 11.1% | 21.1x |
| **Multi-turn Conversation** | L0+L1 | 4,469 ops/sec | 223.8µs | 66.7% | 0.0% | **4.3x** |
| **Code Review** | L0+L1 | 21,113 ops/sec | 47.4µs | 75.0% | 0.0% | **21.1x** |
| **Baseline** (no cache) | None | 940 ops/sec | 1,064µs | N/A | N/A | 1.0x |

## Future Work

Potential enhancements:
- **Metrics & observability**: Expose Prometheus metrics for cache hit/miss rates
- **Per-model configuration**: Different cache settings for different tokenizers
- **Cache warming**: Pre-populate cache with common patterns at startup
- **Dynamic tuning**: Auto-adjust cache sizes based on workload
- **Cross-request sharing**: Share cache across multiple router instances

## Checklist


sgl-router/src/tokenizer/cache/mod.rs
```diff
@@ -0,0 +1,298 @@
...
+        // L1 cache lookup (prefix match)
+        if let Some(l1) = &self.l1 {
+            if let Some((prefix_tokens, prefix_len)) = l1.longest_prefix_match(input) {
+                // We have a prefix match - tokenize the suffix
+                let suffix = &input[prefix_len..];
+                if !suffix.is_empty() {
+                    let suffix_encoding = self.inner.encode(suffix)?;
+
+                    // Merge prefix tokens + suffix tokens
+                    // Note: This is an approximation - boundary tokens may differ
```
zhyncs: This is concerning. Differences in tokenization boundaries can impact output quality. Please ensure that this change does not change token boundaries - the ideal solution is for the incremental tokenizer to split precisely at the special tokens defined in the tokenizer config.

zhyncs (changes requested):

sgl-router/src/tokenizer/cache/mod.rs
slin1237: addressed

CatherineSue:
Left: CachedTokenizer vs Right: Non-cached 
Model: Llama-4-Maverick-FP8
Hardware: 8*H100
Benchmark: lm_eval mmlu_pro
Screenshot 2025-10-18 at 1 08 42 PM

PR #11442, opened 2025-10-11
fix: fix video input for qwen3-vl

mickqian:
(cherry picked from commit a1a20b4c7cb86f81296ece3c4cb265584be1c179)

## Checklist

- [ ] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.ai/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.ai/developer_guide/contribution_guide.html#benchmark-the-speed).

PR #11702, opened 2025-10-16
[quantization][MoE] fix the check for `tp_size` / `moe_ep_size` / `moe_intermediate_size` / `weight_block_size_n`

kevin85421:
## Motivation

The old logic from #11442 uses floor division (i.e. `//`). That is, `8 // 8` is 1 and `8 // 7` is also 1.

```python
if (
    text_config.moe_intermediate_size
    // (self.tp_size // self.moe_ep_size)
) % weight_block_size_n != 0:
```

Therefore, the check will not fail the case where `self.tp_size` is 8 and `self.moe_ep_size` is 7. It will fail at somewhere else and it's hard for users to know it's related to `tp_size` and `ep_size`.

```
python3 -m sglang.launch_server --model-path Qwen/Qwen3-VL-30B-A3B-Instruct-FP8 --tp-size 8 --ep-size 7 2>&1 | tee log
```

## Accuracy Tests

With this PR,

* `--tp-size 8 --ep-size 8`
  ```bash
  python3 -m sglang.launch_server --model-path Qwen/Qwen3-VL-30B-A3B-Instruct-FP8 --tp-size 8 --ep-size 8
  ```
  image

* `--tp-size 8 --ep-size 7`
  ```bash
  python3 -m sglang.launch_server --model-path Qwen/Qwen3-VL-30B-A3B-Instruct-FP8 --tp-size 8 --ep-size 7
  ```
  image

## Checklist


kevin85421:
@mickqian would you mind reviewing this PR? This should be easy to review. Thank you!

hnyls2002 (approved):

PR #11708, opened 2025-10-16
Support running FP4 Deepseek on SM120.

weireweire:
## Motivation

sm120 is not runnable for deepseek because of sm check and invalid kernels.

## Modifications
- add `is_blackwell_supported` as an extend to `is_sm100_supported`
- bypass dsv3_fused_a_gemm on sm120 as it doesn't work.
- use flashinfer `fp4_quantize` to replace native `scaled_fp4_quant` as it doesn't work on sm120. 

with the change, we can run on SM120 by:
```
SGLANG_USE_CUTLASS_BACKEND_FOR_FP4_GEMM=1 \
python3 -m sglang.launch_server   --model-path nvidia/DeepSeek-R1-0528-FP4   --trust-remote-code \
    --host 0.0.0.0 --port 30000  \
    --tp 8 --ep 8 --enable-dp-attention  \
    --attention-backend flashinfer  --moe-runner-backend flashinfer_cutlass --flashinfer-mla-disable-ragged  \
    --max-running-requests 256 --cuda-graph-max-bs 256 \
    --quantization modelopt_fp4 --kv-cache-dtype fp8_e4m3 \
    --stream-output \
    --mem-fraction-static 0.80 \
    --disable-radix-cache 
```

## Accuracy Tests

local-completions (base_url=http://0.0.0.0:30000/v1/completions,model=nvidia/DeepSeek-R1-0528-FP4,tokenized_requests=False,tokenizer_backend=None,num_concurrent=512,max_retries=1,timeout=6000), gen_kwargs: (None), limit: None, num_fewshot: None, batch_size: 1

|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9522|±  |0.0059|
|     |       |strict-match    |     5|exact_match|↑  |0.9484|±  |0.0061|

## Benchmarking and Profiling

command:
```
--backend sglang \
--model $MODEL \
--host 127.0.0.1 --port 30000 \
--dataset-name random \
--max-concurrency 256 \
--num-prompts 256 \
--random-input-len 1024 \
--random-output-len 1024 \
--random-range-ratio 1
```

hardware: RTX Pro 6000 x 8

result:
```
Max request concurrency:                 256       
Successful requests:                     256       
Benchmark duration (s):                  714.07    
Total input tokens:                      262144    
Total input text tokens:                 262144    
Total generated tokens:                  262144    
Total generated tokens (retokenized):    261047    
Request throughput (req/s):              0.36      
Input token throughput (tok/s):          367.11    
Output token throughput (tok/s):         367.11    
Total token throughput (tok/s):          734.23    
Concurrency:                             255.98    
Mean E2E Latency (ms):                   714019.91 
Median E2E Latency (ms):                 714022.43 
Mean TTFT (ms):                          127774.08 
Median TTFT (ms):                        128228.32 
P99 TTFT (ms):                           239964.63 
Mean ITL (ms):                           573.07    
Median ITL (ms):                         463.04    
P95 ITL (ms):                            472.48    
P99 ITL (ms):                            477.19    
Max ITL (ms):                            235910.82 
```

## Checklist

### DIFF
--- a/python/sglang/srt/layers/attention/flashinfer_mla_backend.py
+++ b/python/sglang/srt/layers/attention/flashinfer_mla_backend.py
@@ -25,8 +25,8 @@ from sglang.srt.model_executor.forward_batch_info import ForwardBatch, ForwardMo
 from sglang.srt.server_args import get_global_server_args
 from sglang.srt.speculative.spec_info import SpecInput
 from sglang.srt.utils import (
+    is_blackwell_supported,
     is_flashinfer_available,
-    is_sm100_supported,
     next_power_of_2,
 )
 
@@ -243,7 +243,7 @@ class FlashInferMLAAttnBackend(AttentionBackend):
             self.q_indptr_decode = q_indptr_decode_buf
 
         self.fmha_backend = "auto"
-        if is_sm100_supported():
+        if is_blackwell_supported():
             self.fmha_backend = "cutlass"
         self.prefill_wrapper_ragged = BatchPrefillWithRaggedKVCacheWrapper(
             self.workspace_buffer, "NHD", backend=self.fmha_backend
--- a/python/sglang/srt/layers/quantization/modelopt_quant.py
+++ b/python/sglang/srt/layers/quantization/modelopt_quant.py
@@ -28,7 +28,7 @@ from sglang.srt.layers.quantization.base_config import (
 from sglang.srt.layers.quantization.fp8_utils import (
     apply_fp8_linear,
     cutlass_fp8_supported,
-    is_sm100_supported,
+    is_blackwell_supported,
 )
 from sglang.srt.layers.quantization.kv_cache import BaseKVCacheMethod
 from sglang.srt.layers.quantization.unquant import UnquantizedLinearMethod
@@ -49,8 +49,10 @@ if TYPE_CHECKING:
     )
     from sglang.srt.single_batch_overlap import DownGemmOverlapArgs
 
-if is_cuda():
-    from sgl_kernel import scaled_fp4_quant
+try:
+    from flashinfer import fp4_quantize
+except ImportError:
+    fp4_quantize = None
 
 try:
     from flashinfer import mm_fp4 as fp4_gemm
@@ -867,10 +869,9 @@ class ModelOptFp4LinearMethod(LinearMethodBase):
         output_shape = [x_m, w_n]
 
         # Quantize BF16 or FP16 to (FP4 and interleaved block scale)
-        x_fp4, x_scale_interleaved = scaled_fp4_quant(x, layer.input_scale_inv)
+        x_fp4, x_scale_interleaved = fp4_quantize(x, layer.input_scale_inv)
 
         assert x_fp4.dtype == torch.uint8
-        assert x_scale_interleaved.dtype == torch.float8_e4m3fn
         assert layer.weight.dtype == torch.uint8
         assert layer.weight_scale_interleaved.dtype == torch.float8_e4m3fn
         assert layer.alpha.dtype == torch.float32
@@ -903,7 +904,7 @@ class ModelOptNvFp4FusedMoEMethod(FusedMoEMethodBase):
 
     def __init__(self, quant_config: ModelOptFp4Config):
         self.quant_config = quant_config
-        if not is_sm100_supported():
+        if not is_blackwell_supported():
             raise ValueError(
                 "Current platform does not support NVFP4"
                 " quantization. Please use Blackwell and"
@@ -1410,7 +1411,7 @@ class ModelOptNvFp4FusedMoEMethod(FusedMoEMethodBase):
             output_dtype = x.dtype
             x_sf = None
             if should_use_flashinfer_cutlass_moe_fp4_allgather():
-                from flashinfer import fp4_quantize, nvfp4_block_scale_interleave
+                from flashinfer import nvfp4_block_scale_interleave
 
                 # Quantize before comm, swizzle after.
                 if x.shape[0] > 0:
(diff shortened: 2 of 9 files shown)


python/sglang/srt/layers/attention/flashinfer_mla_backend.py
```diff
@@ -32,8 +32,8 @@
 from sglang.srt.server_args import get_global_server_args
 from sglang.srt.speculative.spec_info import SpecInput
 from sglang.srt.utils import (
+    is_blackwell_supported,
```
FlamingoPg, line 28: I think still use `is_sm100_supported` may a good idea, can change logic inner

python/sglang/srt/utils/common.py
```diff
@@ -170,8 +170,15 @@ def _check(cc_major):
...
-    return torch.cuda.get_device_capability()[0] == 10
+    return torch.cuda.get_device_capability()[0] in [10, 12]
 
+@lru_cache(maxsize=1)
+def is_blackwell_supported(device=None) -> bool:
+    if not is_cuda_alike():
+        return False
+    return (torch.cuda.get_device_capability(device)[0] in [10, 12]) and (
+        torch.version.cuda >= "12.8"
+    )
```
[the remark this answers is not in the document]
weireweire: this consult `is_sm90_supported` 'is_sm100_supported`

weireweire:
CI failure seems unrelated to my change. please review.

FlamingoPg (dismissed):
@weireweire I think still use is_sm100_supported may a good idea, can change logic inner

python/sglang/srt/layers/quantization/modelopt_quant.py
```diff
@@ -49,8 +49,10 @@
     )
     from sglang.srt.single_batch_overlap import DownGemmOverlapArgs
 
-if is_cuda():
-    from sgl_kernel import scaled_fp4_quant
+try:
+    from flashinfer import fp4_quantize
```
FlamingoPg: Why we use flashinfer fp4_quantize but not sgl_kernel scaled_fp4_quant here?

python/sglang/srt/layers/attention/flashinfer_mla_backend.py
weireweire: there is place that we need to distinguish sm100 and sm120, e.g. TRTLLM backend in flashinfer only support sm100f.
Currently I kept both `is_sm100_supported` and `is_blackwell_supported`

weireweire: Do you have suggestion? Another solution is use `is_sm100_supported` and `is_sm120_supported` to represent `is_blackwell_supported`.

nvpohanh: @FlamingoPg Could you provide suggestion? SM100 and SM120 are both "blackwell". These kernels can run on both SM100 and SM120, so using `is_sm100_supported` does not make sense. however, some kernels only support SM100 and do not support SM120, so in those cases, using `is_sm100_supported` makes more sense.

FlamingoPg: @weireweire @nvpohanh Thanks for yours reply, I understand that you’re saying there’s a difference between sm100 and sm120, but I looked at the implementation of is_blackwell_supported. It seems that it only adds a check for whether sm120 is also supported, and doesn’t actually keep both `is_sm100_supported ` and `is_sm120_supported ` at the same time. So I’m curious what the purpose of the renaming here is.

Fridge003:
@weireweire Could you please fix conflicts

python/sglang/srt/layers/attention/flashinfer_mla_backend.py
Fridge003: Agree with @nvpohanh. We can change `is_sm100_supported` to `is_blackwell_supported` for kernels that support both sm100 and sm120

python/sglang/srt/layers/quantization/modelopt_quant.py
```diff
@@ -839,10 +841,9 @@ def apply(
         output_shape = [x_m, w_n]
 
         # Quantize BF16 or FP16 to (FP4 and interleaved block scale)
-        x_fp4, x_scale_interleaved = scaled_fp4_quant(x, layer.input_scale_inv)
+        x_fp4, x_scale_interleaved = fp4_quantize(x, layer.input_scale_inv)
```
Fridge003, line 872: Why doing this change here?

python/sglang/srt/layers/quantization/modelopt_quant.py
```diff
@@ -1059,16 +1060,7 @@ def prepare_static_weights_for_kernel(
         intermediate_size,
         num_experts,
     ):
-        from flashinfer import (
```
Fridge003: Will removing these imports case ImportError?

python/sglang/srt/layers/quantization/modelopt_quant.py
weireweire: The old kernel don't support sm120, so I changed to flashinfer version here. Although https://github.com/sgl-project/sglang/pull/11737 is going to upgrade the kernel for SM120, but I still prefer to used flashinfer version as it's already used in MoE and easier to maintain.  Talked to the `scaled_fp4_quant` kernel author @trevor-m and he also prefer switch to flashinfer.

python/sglang/srt/layers/quantization/modelopt_quant.py
weireweire: These imports were not used. Both IDE and string search suggest that.

Fridge003:
@weireweire Can you please fix conflict again, thanks~

weireweire:
> @weireweire Can you please fix conflict again, thanks~

Done

Fridge003 (approved):
Thanks for your great work~

PR #11709, opened 2025-10-16
[Fix] fix type issue of env flag value MODELOPT_MAX_TOKENS_PER_EXPERT

zejunchen-zejun:
Here PR is used to fix the type error of env flag MODELOPT_MAX_TOKENS_PER_EXPERT when it is specified, otherwise it will have below error:
image


zejunchen-zejun:
Hi, @FlamingoPg 
Thank you for help! When the CI starts, it almost failed intermediately. The CI seems failed in loading the docker image.
image

HaiShaw (approved):

PR #11732, opened 2025-10-16
[router] Add rustfmt and set group imports by default

CatherineSue:
## Motivation

Cargo nightly supports grouping imports by default. This is a useful feature as sgl-router has a lot of imports right now. Only few files like `server.rs` groups imports in crate to make it more readable and concise.

## Modifications

- Add `rustfmt.toml` and set group crate imports
- Run `cargo +nightly fmt`

## Checklist

### DIFF
--- a/.github/workflows/pr-test-rust.yml
+++ b/.github/workflows/pr-test-rust.yml
@@ -54,7 +54,8 @@ jobs:
         run: |
           source "$HOME/.cargo/env"
           cd sgl-router/
-          cargo fmt -- --check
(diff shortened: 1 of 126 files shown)


key4ng (approved):

key4ng:
could we also add it to pre-commit

CatherineSue:
> could we also add it to pre-commit

I remember it would require `protoc` to be installed #10878 Then I gave up

slin1237 (approved):

PR #11745, opened 2025-10-17
support more model in piecewise cuda graph

narutolhy:
## Motivation

This PR refines the execution conditions in the CUDA graph runner and relaxes strict output shape checks to improve compatibility with models that have different tensor layouts.

Specifically, some models (e.g., Gemma-3-1B-IT) produce tensors with the same total number of elements but different shapes due to layout or internal reshape differences. The previous strict assert output.shape == ret.shape would incorrectly fail in such cases.
In addition, CUDA graph execution should be allowed when logprobs are returned for the last token only, but not when input logprobs are required.

## Modifications

1 `piecewise_cuda_graph_runner.py`

Updated can_run() logic:
Previously, CUDA graph capture was skipped when forward_batch.return_logprobs was True.
Now, it only skips when input-id logprobs are requested.

```python
for start_len, seq_len in zip(
    forward_batch.extend_logprob_start_lens_cpu,
    forward_batch.extend_seq_lens_cpu
):
    if start_len is not None and start_len < seq_len:
        return True
```

2 `radix_attention.py`
Replaced strict shape check:
```python
assert output.shape == ret.shape
```
with a relaxed element-count check:
```python
assert output.numel() == ret.numel()
```

## Accuracy Tests

Verified correctness on multiple models including:

Qwen3-0.6B

Gemma-3-1B-IT

Outputs are numerically identical to the baseline (within FP16 tolerance).

Logprob return path remains functionally unchanged.

## Checklist

### DIFF
--- a/python/sglang/srt/model_executor/piecewise_cuda_graph_runner.py
+++ b/python/sglang/srt/model_executor/piecewise_cuda_graph_runner.py
@@ -254,12 +254,13 @@ class PiecewiseCudaGraphRunner:
     def can_run(self, forward_batch: ForwardBatch):
         num_tokens = len(forward_batch.input_ids)
         # TODO(yuwei): support return input_ids' logprob
-        for start_len, seq_len in zip(
-            forward_batch.extend_logprob_start_lens_cpu,
-            forward_batch.extend_seq_lens_cpu,
-        ):
-            if start_len is not None and start_len < seq_len:
-                return True
+        if forward_batch.return_logprob:
+            for start_len, seq_len in zip(
+                forward_batch.extend_logprob_start_lens_cpu,
+                forward_batch.extend_seq_lens_cpu,
+            ):
+                if start_len is not None and start_len < seq_len:
+                    return False
         if num_tokens <= self.max_num_tokens:
             return True
         return False
--- a/python/sglang/srt/model_executor/piecewise_cuda_graph_runner.py
+++ b/python/sglang/srt/model_executor/piecewise_cuda_graph_runner.py
@@ -253,9 +253,13 @@ class PiecewiseCudaGraphRunner:
 
     def can_run(self, forward_batch: ForwardBatch):
         num_tokens = len(forward_batch.input_ids)
-        # TODO(yuwei): support return logprob
-        if forward_batch.return_logprob:
-            return False
+        # TODO(yuwei): support return input_ids' logprob
+        for start_len, seq_len in zip(
+            forward_batch.extend_logprob_start_lens_cpu,
+            forward_batch.extend_seq_lens_cpu,
+        ):
+            if start_len is not None and start_len < seq_len:
+                return True
         if num_tokens <= self.max_num_tokens:
             return True
         return False
@@ -426,7 +430,7 @@ class PiecewiseCudaGraphRunner:
             out_cache_loc=out_cache_loc,
             seq_lens_sum=forward_batch.seq_lens_sum,
             encoder_lens=forward_batch.encoder_lens,
-            return_logprob=forward_batch.return_logprob,
+            return_logprob=False,
             extend_seq_lens=forward_batch.extend_seq_lens,
             extend_prefix_lens=forward_batch.extend_prefix_lens,
             extend_start_loc=forward_batch.extend_start_loc,
--- a/python/sglang/srt/model_executor/piecewise_cuda_graph_runner.py
+++ b/python/sglang/srt/model_executor/piecewise_cuda_graph_runner.py
@@ -179,7 +179,20 @@ class PiecewiseCudaGraphRunner:
         # Set graph pool id globally to be able to use symmetric memory
         set_graph_pool_id(get_global_graph_memory_pool())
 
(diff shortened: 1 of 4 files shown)


narutolhy:
Hi @Oasis-Git, Thank you for providing the piecewise_cuda_graph feature, which is very helpful. I have made some relaxations on the usage conditions to make it applicable to more scenarios and models. Please take a look. Thank you.

python/sglang/srt/model_executor/piecewise_cuda_graph_runner.py
```diff
@@ -426,7 +430,7 @@ def replay_prepare(
             out_cache_loc=out_cache_loc,
             seq_lens_sum=forward_batch.seq_lens_sum,
             encoder_lens=forward_batch.encoder_lens,
-            return_logprob=forward_batch.return_logprob,
+            return_logprob=False,
```
ispobock, line 434: For logprobs support, @Oasis-Git is working on that.

ispobock:
@narutolhy Thanks for contribution! We have a slack channel #piecewise-cuda-graph to discuss this feature, welcome to join if you are interested.

ispobock (approved):

python/sglang/srt/model_executor/piecewise_cuda_graph_runner.py
narutolhy: Thanks, I see the plan. I'm only temporarily supporting the return of logprob for non-input tokens. Originally, in the can_run function, only batch logprob was returned, not using piecewise CUDA graphs. However, it seems only input tokens were not supported, so I relaxed the requirements a bit to also support the return of the logprob for the last token in prefill-only scenarios. 
Same as line 223, it is all False here and can be changed to True later. I am planning to use it temporarily here. So I set it to false first. Thanks

Oasis-Git:
Hi @narutolhy 

Thanks for your contribution. Could you please add the following part in your pr:
1. MMLU Test/Benchmark Output
2. Unit Test for MMLU in file: https://github.com/sgl-project/sglang/blob/main/test/srt/test_piecewise_cuda_graph.py

narutolhy:
> Hi @narutolhy
> 
> Thanks for your contribution. Could you please add the following part in your pr:
> 
> 1. MMLU Test/Benchmark Output
> 2. Unit Test for MMLU in file: https://github.com/sgl-project/sglang/blob/main/test/srt/test_piecewise_cuda_graph.py

ok, I will add them

narutolhy:
> Hi @narutolhy
> 
> Thanks for your contribution. Could you please add the following part in your pr:
> 
> 1. MMLU Test/Benchmark Output
> 2. Unit Test for MMLU in file: https://github.com/sgl-project/sglang/blob/main/test/srt/test_piecewise_cuda_graph.py

Hi @Oasis-Git I have add Unit Test for MMLU in file: https://github.com/sgl-project/sglang/blob/main/test/srt/test_piecewise_cuda_graph.py
I also added tests for the gemma model so that I can test scenarios where the shapes are different but the overall size is the same.

There is no problem when using the unsloth/gemma-3-1b-it model, but an error occurs when using the unsloth/gemma-3-4b-it model:
AssertionError: Input addresses for cudagraphs are different during replay. Expected [140472240193024, 140472239365632, 140472239366144, 140383149409792, 140472239366656, 140466951553024, 140472239364608, 140472239365120], got [140472240193024, 140472239365632, 140472239366144, 140399219879424, 140472239366656, 140466951553024, 140472239364608, 140472239365120]

The value of the fourth pointer is inconsistent, I don't know if you've encountered this before.
That's why I used unsloth/gemma-3-1b-it for the mmlu test. Since it's a small model, the test standards were lowered.

Please review it.
Thank you

Oasis-Git:
> > Hi @narutolhy
> > Thanks for your contribution. Could you please add the following part in your pr:
> > 
> > 1. MMLU Test/Benchmark Output
> > 2. Unit Test for MMLU in file: https://github.com/sgl-project/sglang/blob/main/test/srt/test_piecewise_cuda_graph.py
> 
> Hi @Oasis-Git I have add Unit Test for MMLU in file: https://github.com/sgl-project/sglang/blob/main/test/srt/test_piecewise_cuda_graph.py I also added tests for the gemma model so that I can test scenarios where the shapes are different but the overall size is the same.
> 
> There is no problem when using the unsloth/gemma-3-1b-it model, but an error occurs when using the unsloth/gemma-3-4b-it model: AssertionError: Input addresses for cudagraphs are different during replay. Expected [140472240193024, 140472239365632, 140472239366144, 140383149409792, 140472239366656, 140466951553024, 140472239364608, 140472239365120], got [140472240193024, 140472239365632, 140472239366144, 140399219879424, 140472239366656, 140466951553024, 140472239364608, 140472239365120]
> 
> The value of the fourth pointer is inconsistent, I don't know if you've encountered this before. That's why I used unsloth/gemma-3-1b-it for the mmlu test. Since it's a small model, the test standards were lowered.
> 
> Please review it. Thank you

I left some comments in slack channel.

merrymercy (approved):

PR #11746, opened 2025-10-17
[CI] Fix broken event loop creation

hnyls2002:
SGLang engine uses `get_event_loop()` in non-uvloop-or-uvicorn context, when uvloop is upgraded to `0.22.1`, this API is deprecated and throws an error.

### DIFF
--- a/python/sglang/srt/entrypoints/engine.py
+++ b/python/sglang/srt/entrypoints/engine.py
@@ -30,8 +30,6 @@ import time
 from typing import AsyncIterator, Dict, Iterator, List, Optional, Tuple, Union
 
 import zmq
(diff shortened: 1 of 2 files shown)


Kangyan-Zhou:
@hnyls2002 could you please help link the culprit PR? I'm a bit curious to see what happened

hnyls2002:
@Kangyan-Zhou, I did not find the related PR. I just found that the CI machine upgrades uvloop from `0.21.0` to `0.22.1`, and then the CI breaks. That is because in the newer uvloop version, Python will deprecate the old API `get_event_loop()` when there is no event loop in the current thread.

PR #11755, opened 2025-10-17
[Feature] Support loading weights from ckpt engine worker

stmatengss:
## Motivation

Based on #10667/#10464 and #10646, we minimize implementation to ease review.
The ckpt engine maintainer approved this design (using worker, not connector) after discussion.

Usage: 

- Install ckpt engine (https://github.com/MoonshotAI/checkpoint-engine)
```
pip install 'checkpoint-engine[p2p]'  # install checkpoint engine
```
- Launch the server with wait-for-initial-weights option in one terminal (sglang side): 
```
python -m sglang.launch_server --model-path Qwen/Qwen3-8B --tp  8  --load-format dummy --wait-for-initial-weights
```
- Torchrun script in another terminal (ckpt engine side): 
```
torchrun --nproc-per-node 8 examples/checkpoint_engine/update.py  --update-method broadcast --checkpoint-path /path/to/Qwen/Qwen3-8B/  --inference-parallel-size 8
```

Authors: @stmatengss @BraveY @XucSh @zxpdemonio 
Thanks to @weixiao-huang for the help.

## Checklist


ByronHsu:
https://github.com/sgl-project/sglang/pull/10667#pullrequestreview-3354997564

BraveY:
> [#10667 (review)](https://github.com/sgl-project/sglang/pull/10667#pullrequestreview-3354997564)

> High level questions:
> 
> 1. Checkpoint engine needs to load from disk to cpu, and use the pipeline to send to inference engines. disk-to-cpu is still on the critical path. What is the time we save here?
> 


@ByronHsu 
Time savings are achieved in two aspects.
1. For multi-node launched instances. Each node only needs to load a portion of the weights from the disk into memory. This is equivalent to enhancing the disk bandwidth. Theoretically speaking, if the communication overhead does not increase, the more nodes that can participate in the checkpoint-engine networking, the more significant the acceleration effect will be. Currently, preliminary tests loading DeepSeek-R1 on H20-3e show that the acceleration time for two nodes is 20 seconds. We will publish a more detailed test report later.
2. For a single loading process, using the dummy format to load weights can overlap the time from disk to CPU with processes like Cuda Graph capture, so it can also bring about time savings.

stmatengss:
> [#10667 (review)](https://github.com/sgl-project/sglang/pull/10667#pullrequestreview-3354997564)

For 3,

Originally, ckpt engine is used for RL case. You can treat ckpt engine update python script as a training instance. Ckpt engine community will replace NCCL with p2p transfer (mooncake transfer engine) to reduce broadcast communication overhead.

BraveY:
> https://github.com/sgl-project/sglang/pull/10667#pullrequestreview-3354997564
> High level questions:
> 

> 2. Can you provide examples of how to use in online serving cases? For example, I add 10 new engines, how can i use checkpoint engine to make them load faster.
> 

2. The two instance use example is as following 

**Node0**:
Launch sglang server
```
python -m sglang.launch_server --model-path Qwen/Qwen3-8B --tp  8  --load-format dummy --wait-for-initial-weights --host 10.13.2.200
```
Torchrun script 
```
torchrun --nproc-per-node 8 --nnodes 2 --node-rank 0 --master-addr 10.13.2.200 --master-port 29500 examples/checkpoint_engine/update.py--update-method broadcast --checkpoint-path /path/to/Qwen/Qwen3-8B/  --inference-parallel-size 8
```

**Node1**:
Launch sglang server
```
python -m sglang.launch_server --model-path Qwen/Qwen3-8B --tp  8  --load-format dummy --wait-for-initial-weights --host 10.13.2.201
```
Torchrun script 
```
torchrun --nproc-per-node 8 --nnodes 2 --node-rank 1 --master-addr 10.13.2.200 --master-port 29500 examples/checkpoint_engine/update.py--update-method broadcast --checkpoint-path /path/to/Qwen/Qwen3-8B/  --inference-parallel-size 8
```

The two instance also support  `tp=16`

**Node0**
Launch sglang server
```
python -m sglang.launch_server --model-path Qwen/Qwen3-8B --tp  8  --load-format dummy --wait-for-initial-weights --host 10.13.2.200 --dist-init-addr 10.13.2.200 :9120 --nnodes 2 --node-rank 0
```
Torchrun script 
set `--inference-parallel-size` to 16
```
torchrun --nproc-per-node 8 --nnodes 2 --node-rank 0 --master-addr 10.13.2.200 --master-port 29500 examples/checkpoint_engine/update.py--update-method broadcast --checkpoint-path /path/to/Qwen/Qwen3-8B/  --inference-parallel-size 16
```

**Node1**
```python -m sglang.launch_server --model-path Qwen/Qwen3-8B --tp  8  --load-format dummy --wait-for-initial-weights --host 10.13.2.200 --dist-init-addr 10.13.2.200 :9120 --nnodes 2 --node-rank 1```

Torchrun script 
set `--inference-parallel-size` to 16
```
torchrun --nproc-per-node 8 --nnodes 2 --node-rank 1 --master-addr 10.13.2.200 --master-port 29500 examples/checkpoint_engine/update.py--update-method broadcast --checkpoint-path /path/to/Qwen/Qwen3-8B/  --inference-parallel-size 16
```

**Comment by ByronHsu** [comment]
Chatted offline. Few notes:
1. Checkpoint engines are co-located with inference engine in the same pod, and store the checkpoint in CPU RAM.
2. The checkpoint engine can either load from disk or broadcast/p2p from existing checkpoint engine.
3. This can reduce engine startup time by (a) overlapping inference engine init (e.g. cuda graph) and disk->RAM loading (b) weight loading is faster if using broadcast/p2p.

I think it is worth integrating into OME to enhance deployment startup speed. cc @slin1237

File: examples/checkpoint_engine/update.py [hunk 1]
```diff
@@ -0,0 +1,241 @@
+"""
+Usage:
+1) Launch the server with wait-for-initial-weights option in one terminal:
+   python -m sglang.launch_server --model-path /workspace/Qwen/Qwen3-4B/ --tensor-parallel-size 2 --port 19730 --load-format dummy --wait-for-initial-weights --mem-fraction-static 0.7
+
+2) Torchrun this script in another terminal:
+    torchrun --nproc-per-node 2 update.py --update-method broadcast --checkpoint-path /workspace/Qwen/Qwen3-4B/  --inference-parallel-size 2
```
ByronHsu, line 7: in the future, have something like `python -m sglang.launch_checkpoint_engine` inside sglang python package, so whoever installs sglang can use it

python/pyproject.toml
```diff
@@ -93,10 +93,10 @@ test = [
   "sentence_transformers",
   "tabulate",
 ]
-all = []
+decord = ["decord2"]
+checkpoint-engine = ["checkpoint-engine>=0.1.2"]
+all = ["sglang[test]", "sglang[decord]", "sglang[checkpoint-engine]"]
```
ByronHsu: checkpoint engine should be optional not in `all`

python/sglang/srt/server_args.py
```diff
@@ -203,6 +203,7 @@ class ServerArgs:
     skip_server_warmup: bool = False
     warmups: Optional[str] = None
     nccl_port: Optional[int] = None
+    wait_for_initial_weights: bool = False
```
ByronHsu: can we rename as `--checkpoint-engine-wait-weights-before-ready`?

python/sglang/srt/utils/common.py
```diff
@@ -2232,6 +2232,11 @@ def launch_dummy_health_check_server(host, port, enable_metrics):
 
     app = FastAPI()
 
+    @app.get("/ping")
+    async def ping():
```
ByronHsu, line 2236: why do we need `/ping` if we have `/health`

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -2328,6 +2328,74 @@ def save_sharded_model(
...
 
+    def update_weights_from_ipc(self, recv_req):
+        """Update weights from IPC for checkpoint-engine integration."""
+        try:
+            from sglang.srt.checkpoint_engine_worker import (
+                SGLangCheckpointEngineWorkerExtension,
+            )
+
+            # Create a worker extension that integrates with SGLang's model
+            class SGLangWorkerImpl(SGLangCheckpointEngineWorkerExtension):
```
ByronHsu: this should not be in model_runner

ByronHsu (commented):
On a high level: 
1. Make all checkpoint engine dep "lazy" and not include in `all`
2. Create a folder called `checkpoint_engine` under `srt/` and move the impl of model_runner and checkpoint_engine_worker inside

python/sglang/srt/utils/common.py
BraveY: > why do we need `/ping` if we have `/health`

For the dummy health server, /health can indeed be used, but this means modifying the probe endpoint of the checkpoint-engine for different instances, because for non-dummy instances, we use /ping to probe. To unify usage, we choose to add an interface here.
The reason for not directly using the /health but using the /ping is to ensure that /health returns 200 only after the initialization weights are completed. Otherwise, when /health returns 200, the weights might not be ready yet.

examples/checkpoint_engine/update.py
XucSh: good point! we can move this file to python/sglang

python/pyproject.toml
stmatengss: > checkpoint engine should be optional not in `all`

True, we will use `try-except` blocks to import `checkpoint-engine` in the relevant code and prompt users to install it if it's not found.

examples/checkpoint_engine/update.py
stmatengss: Will fix it

python/sglang/srt/server_args.py
stmatengss: Fix it

python/sglang/srt/model_executor/model_runner.py
stmatengss: Fix it

XucSh:
Testing was conducted on an 8*A10 GPU machine with the following results:

For the Qwen3-8B model with TP=4, the loading time for the Checkpoint Engine was 9s（assuming modelweights have been prefetched into dram by checkpoint engine）, compared to 85s for loading from disk.

python/sglang/srt/managers/io_struct.py
```diff
@@ -1078,6 +1078,22 @@ class InitWeightsSendGroupForRemoteInstanceReqInput(BaseReq):
     backend: str = "nccl"
 
 
+@dataclass
+class UpdateWeightsFromIPCReqInput(BaseReq):
```
ByronHsu, line 1084: nit: say this is used for ckpt engine

python/sglang/srt/utils/common.py
ByronHsu: makes sense

ByronHsu (approved):

python/sglang/srt/managers/io_struct.py
XucSh: Have added the notes

examples/checkpoint_engine/update.py
```diff
@@ -0,0 +1,241 @@
...
+    start = time.perf_counter()
+    yield
+    end = time.perf_counter()
+    logger.info(f"{msg} duration: {end - start:.2f} seconds")
+
+
+def check_sglang_ready(
+    endpoint: str, inference_parallel_size: int, uds: str | None = None
+):
+    if rank != rank // inference_parallel_size * inference_parallel_size:
```
ShangmingCai, line 39: Where is `rank` this parameter initialized?

XucSh: This variable is populated by torchrun, which this script currently depends on. This dependency will be removed later.

examples/checkpoint_engine/update.py
```diff
@@ -0,0 +1,241 @@
...
+    rank = int(os.getenv("RANK"))
+    world_size = int(os.getenv("WORLD_SIZE"))
+    req_func = req_inference(
+        args.endpoint,
+        args.inference_parallel_size,
+        uds=args.uds,
+        weight_version=args.weight_version,
+    )
+    ps = ParameterServer(auto_pg=True)
+    ps._p2p_store = None
```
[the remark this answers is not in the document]
ShangmingCai: Oh, I see. `rank` and `world_size` are initialized here in `main` according to env var. So they are the parameters of torch run? Should we make them the global params? I am seeing many `os.getenv` which don't feel good.

```python
    rank = int(os.getenv("RANK"))
    world_size = int(os.getenv("WORLD_SIZE"))
```
Should they be given a prefix like `SGLANG_CKPT_ENGINE_`?

examples/checkpoint_engine/update.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+def req_inference(
+    endpoint: str,
+    inference_parallel_size: int,
+    timeout: float = 300.0,
+    uds: str | None = None,
+    weight_version: str | None = None,
+) -> Callable[[list[tuple[str, str]]], None]:
+    rank = int(os.getenv("RANK", None))
```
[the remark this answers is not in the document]
ShangmingCai: I agree with genimi, `None` is not a good default value for the following logic.

examples/checkpoint_engine/update.py
```diff
@@ -0,0 +1,241 @@
...
+    src = rank // inference_parallel_size * inference_parallel_size
```
ShangmingCai:
```suggestion
    rank = int(os.getenv("RANK", 0))
    src = rank // inference_parallel_size * inference_parallel_size
```

examples/checkpoint_engine/update.py
```diff
@@ -0,0 +1,241 @@
...
+                json={
+                    "zmq_handles": dict(
+                        socket_paths[src : src + inference_parallel_size]
+                    ),
+                    "flush_cache": True,
+                    "weight_version": weight_version,
+                },
+                timeout=timeout,
+            )
+            resp.raise_for_status()
```
ShangmingCai:
```suggestion
            with httpx.Client(transport=httpx.HTTPTransport(uds=uds)) as client:
                resp = client.post(
                    f"{endpoint}/update_weights_from_ipc",
                    json={
                        "zmq_handles": dict(
                            socket_paths[src : src + inference_parallel_size]
                        ),
                        "flush_cache": True,
                        "weight_version": weight_version,
                    },
                    timeout=timeout,
                )
                resp.raise_for_status()
```

examples/checkpoint_engine/update.py
```diff
@@ -0,0 +1,241 @@
...
+            )
+            response.raise_for_status()
+            break
+        except (httpx.ConnectError, httpx.HTTPStatusError) as e:
+            if retry_num % 10 == 0:
+                logger.warning(
+                    f"fail to check sglang ready, retry {retry_num} times, error: {e}"
+                )
+            retry_num += 1
+            time.sleep(0.1)
```
ShangmingCai:
```suggestion
    with httpx.Client(transport=transport) as client:
        while True:
            try:
                response = client.get(f"{endpoint}/ping", timeout=10)
                response.raise_for_status()
                break
            except (httpx.ConnectError, httpx.HTTPStatusError) as e:
                if retry_num % 10 == 0:
                    logger.warning(
                        f"fail to check sglang ready, retry {retry_num} times, error: {e}"
                    )
                retry_num += 1
                time.sleep(0.1)
```
This should help us avoid resource leaks.

python/pyproject.toml
```diff
@@ -93,6 +93,7 @@ test = [
   "sentence_transformers",
   "tabulate",
 ]
+checkpoint-engine = ["checkpoint-engine>=0.1.2"]
```
ShangmingCai: Should we use `==` to ping the version in case `checkpoint-engine` changes the interfaces or introduces some bugs?

examples/checkpoint_engine/update.py
stmatengss: The checkpoint engine project will add a general update.py script for both vllm and sglang. This script is a temporary solution.

python/sglang/srt/entrypoints/http_server.py
```diff
@@ -129,6 +130,7 @@
 asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
 
 HEALTH_CHECK_TIMEOUT = int(os.getenv("SGLANG_HEALTH_CHECK_TIMEOUT", 20))
+WAIT_WEIGHTS_READY_TIMEOUT = int(os.getenv("SGLANG_WAIT_WEIGHTS_READY_TIMEOUT", 120))
```
ShangmingCai, line 133: Consider adding this to doc when the ckpt engine has a separate page of doc. Or we should put it in the _docs/references/environment_variables.md_.

examples/checkpoint_engine/update.py
stmatengss: TODO: fix it when checkpoint engine releases a new version.

ShangmingCai (approved):
LGTM

stmatengss:
Resolve all comments. PTAL. @ShangmingCai

stmatengss:
> Can you wrap this `torchrun --nproc-per-node 8 examples/checkpoint_engine/update.py  --update-method broadcast --checkpoint-path /path/to/Qwen/Qwen3-8B/  --inference-parallel-size 8` as a simple command like `python3 -m sglang.ckpt.update...`

Thx for reminder. Will add this feature ASAP!

PR #9812, opened 2025-08-30
benchmark: enhance configurable multimodal benchmarking in bench_serving

AlienKevin:
# Motivation

**https://github.com/sgl-project/sglang/pull/9583 introduced configurable multimodal benchmarking to bench_serving. This PR refines it in three ways, positioning bench_serving as a reliable way to benchmark and compare VLM performance within and across inference frameworks.**

* **Count vision tokens in throughput metrics with a clear per-modality breakdown**: Vision-prefill benchmarks often have ~0 input text tokens, ~1,000 image tokens, and ~0 output text tokens. If we only count text tokens, key metrics misleadingly read as zero (e.g., total input tokens, input token throughput, total token throughput). Following discussions with @mickqian, all metrics now include vision tokens, and we report separate counts for text vs. vision tokens for users who need the details.

* **Broaden image options: PNG support and a blank image mode**: We add PNG alongside JPEG and introduce a blank image option to simulate various preprocessing overheads. Random PNGs are much larger in size than JPEGs, stress-testing preprocessing; blank PNGs are tiny and ideal for isolating prefill/decode performance with minimal I/O overhead. Blank images are also useful for verifying that multimodal cache is functioning properly.

* **Replace deprecated max_tokens with max_completion_tokens**: Using max_tokens can be confusing, especially when random-output-len is set below prompt_len, since users often assume output length excludes inputs. Switching to max_completion_tokens removes that ambiguity.

# Tests

We tested Qwen2.5-VL-7B-Instruct with this PR on SGLang and vLLM. Since this PR also touches the MMMU dataset in bench_serving, we evaluate both MMMU and a fixed ISL/OSL setup. All benchmarks undergo three consecutive runs where the server is kept alive between the runs for consistency. The results match [previous benchmarks](https://github.com/sgl-project/sglang/issues/8179#issuecomment-3184880291) and show that SGL is currently slower than vLLM (~13% slower on MMMU and 46% slower on ISL1000/OSL1). Much of the slow down come from the scheduler/preprocessing as the GPU is idle for much longer in SGL.

SGL
Benchmark | Total tokens/s (per run) | Average total tokens/s / Std
-- | -- | --
MMMU (Math) | 9,154 • 9,396 • 10,242 | **9,597** / 571
ISL1000/OSL1 | 9,259 • 9,781 • 9,838 | **9,626** / 319

vLLM
Benchmark | Total tokens/s (per run) | Average total tokens/s / Std
-- | -- | --
[...]

PR #11759, opened 2025-10-17
fix: fix MMMU loading issue

ZailiWang:
## Motivation

#9812 did amazing work. However, it did not port the adaptation of Phi4MM benchmarking contributed in #9900. Another issue is in `DatasetRow` construction, the original input param `text_prompt` is used instead of the processed `prompt_str`. This leads to all the VLM benchmark fail at prompt decompose phase at server side.

## Modifications

Fix the mistakes above and make VLM benchmarks usable.

## Accuracy Tests

N/A

## Benchmarking and Profiling

N/A

## Checklist


ZailiWang:
Hi @AlienKevin would you help review the PR? It looks like a typo in #9812 but you said you have verified.
Anyway I cannot pass the `bench_serving` test without my change in this PR.

python/sglang/bench_serving.py
```diff
@@ -1368,7 +1372,7 @@ def create_mm_data_row(text_prompt, images: list, images_base64, output_len, pro
     vision_prompt_len = prompt_len - text_prompt_len
 
     return DatasetRow(
-        prompt=text_prompt,
+        prompt=prompt_str,
```
AlienKevin: text_prompt was used because later on 
[async_request_openai_chat_completions](https://github.com/sgl-project/sglang/blob/d49e91731a82a0af0ef210b54b240a42a39c7a1c/python/sglang/bench_serving.py#L306) will wrap MM requests in a chat format so we want to avoid double wrapping. The wrapped `prompt_str` is purely used to estimate `vision_prompt_len` and `text_prompt_len`.

AlienKevin (commented):
Hi, thanks for bring up this PR. Not sure what happened after my PR was merged but at the time I used it to benchmark Qwen 2.5 VL extensively and it worked.

vincentzed:
I think it's here at the exception?
Since it was excepted before. not sure if this is 100% your error you mean.
https://github.com/sgl-project/sglang/pull/11385/files#diff-00e43baca391ee5f5d8c587be8480acc86944e56adc9b28bf5a24f1dd4314ae3R1294

ZailiWang:
> I think it's here at the exception? Since it was excepted before. not sure if this is 100% your error you mean. https://github.com/sgl-project/sglang/pull/11385/files#diff-00e43baca391ee5f5d8c587be8480acc86944e56adc9b28bf5a24f1dd4314ae3R1294

Thanks for the discussion. No, it has nothing to do with this snippet. The error I met is the same as described in #9900 ,which is due that the prompt format is not recognized by the server, which is due that in my test the prompt requires wrapping.

ZailiWang:
Hi @AlienKevin , I think the current issue is due that `bench_serving` supports both `generate` and openAI `chat/completion` APIs. Actually I was always using `generate` not aware about the usage of `chat/completion` interface in this script.

Looks like an `if-else` should be applied based on the requests interface to decide whether the `text_prompt` should be applied here. Let me study the code and update the PR.

AlienKevin:
@ZailiWang I remember gating the MMMU and image benchmarks to only allow chat completions. For multimodal use cases, I think chat makes more sense because the chat template is already required to pass images into the prompt.

ZailiWang:
Hi @AlienKevin I get your point, but I still have the following questions/concerns.

- I didn't meet the "gating" functionality you mentioned. I was calling `generate` instead of `v1/chat/completions` with no error occurred before 9812 merged.
- I can understand your point about "applying chat template makes more sense". However, the default `backend` setting in `bench_serving.py` [is `sglang`](https://github.com/sgl-project/sglang/blob/9edb7b51232cd63054eeefe22469fd8ef0672d4b/python/sglang/bench_serving.py#L2301), which calls [`generate` API](https://github.com/sgl-project/sglang/blob/9edb7b51232cd63054eeefe22469fd8ef0672d4b/python/sglang/bench_serving.py#L2166). This means, if a user started use SGLang to benchmark VLM with default settings, he/she may get an error which is hard to understand. 

I updated the PR for handling both cases where the prompt wrapping should or should not be applied in `create_mm_data_row()`. Would you help verify if this works with your benchmarking approach? Thanks.

AlienKevin:
I had assert checks for `apply_chat_template` which got changed into a default setting by a maintainer in a later commit https://github.com/sgl-project/sglang/pull/9812/commits/f7dff0db7e6fa8d54d91728cd911f7380587412d that I wasn't aware of. The idea of the check is to ensure users are aware that chat is required for MM benchmarks.

I think you made a good point in that the default backend may be confusing. In that case, I think we can also override the default backend in the same manner as the commit above so chat is always used.

For testing, you can follow the [`bench_serving.md` guide](https://github.com/sgl-project/sglang/blob/main/docs/developer_guide/bench_serving.md#examples). Currently we require explicitly passing a chat backend like `sglang-oai-chat`.

mingfeima (approved):

mingfeima:
@ZailiWang fix the lint issue.

ZailiWang:
> @ZailiWang fix the lint issue.

It's the lint error on main, which was fixed later yesterday.

ZailiWang:
I just verified that the current PR changes can pass the `bench_serving` tests for requesting either `sglang` or `sglang-oai-chat` backends.

Let me summarize the candidate change approaches:
a. Keep all the backends usable for multimodal tests. Add the adaptation logic for the different backends: for openAI completion APIs, keep raw prompt text at the client side; otherwise apply the chat template when constructing the requests.
b. Add the limitation for the users that only openAI completion API should be used for multimodal tests.

The current PR content is implementing option (a), which is my preference. My points are:
- Calling openAI completion API is not a necessity for applying chat template. I think wrapping the raw prompts with the chat template at the client side is also a implementation of chat template application.
- For the users who only care about backend performance, we may prefer to use `sglang` backend calling `generate` interface directly, which more precisely measures backend inference . Per my test, calling `generate` gives a lower TTFT than calling `v1/chat/completion`. I believe the diff is due that the extra pre-processing of the prompt is also counted. Hence I don't think use `sglang` backend and calling `generate` API is a misuse.

ZailiWang:
@mickqian @JustinTong0323 would you help decide which solution is preferred? Thanks. Please also review the PR if Solution a) is preferred, or I can update the PR if Solution b) is preferred.

mickqian:
Hi, can you run `test_nightly_vlms_mmmu_eval.py` locally to see if it is broken? cheers

ZailiWang:
Hi @mickqian , I cannot complete all cases in the test file, since I can only use Xeon CPU servers at hand, which does not support many models in the test cases. Some models requires specific authentication so the downloading is forbidden.

I extracted one succeed case `deepseek-ai/Janus-Pro-7B` and its results passed [the acc criterion of 0.285](https://github.com/sgl-project/sglang/blob/main/test/srt/test_nightly_vlms_mmmu_eval.py#L23):

```
{
  "Overall-Art and Design": {
    "num": 10,
    "acc": 0.4
  },
  "Art": {
    "num": 10,
    "acc": 0.4
  },
  "Overall-Business": {
    "num": 30,
    "acc": 0.367
  },
  "Accounting": {
    "num": 30,
    "acc": 0.367
  },
  "Overall-Tech and Engineering": {
    "num": 60,
    "acc": 0.267
  },
  "Agriculture": {
    "num": 30,
    "acc": 0.233
  },
  "Architecture_and_Engineering": {
    "num": 30,
    "acc": 0.3
  },
  "Overall": {
    "num": 100,
    "acc": 0.31
  },
  "score": 0.31
}
```

And from source code perspective, I confirmed that the requests in this test are sent via [OpenAI API](https://github.com/sgl-project/sglang/blob/main/python/sglang/test/simple_eval_common.py#L98). The [changes](https://github.com/sgl-project/sglang/pull/11759/files#diff-00e43baca391ee5f5d8c587be8480acc86944e56adc9b28bf5a24f1dd4314ae3R1373-R1386) in this PR has ensured that the logic remains unchanged for OpenAI API requests (`sglang-oai` or `sglang-oai-chat` backends). My PR only ensures non-OpenAI-API requests would work too and adds a specific format change required by Phi4MM.

Does the results and explanation look good to you?

mickqian:
Hi @ZailiWang, can you help fix [this](https://github.com/sgl-project/sglang/actions/runs/18860038207/job/53816332269#step:4:563)?

ZailiWang:
> Hi @ZailiWang, can you help fix [this](https://github.com/sgl-project/sglang/actions/runs/18860038207/job/53816332269#step:4:563)?

Created #12256 for fixing it.

PR #11772, opened 2025-10-17
[overlap-spec] support page size > 1

hnyls2002:


python/sglang/srt/speculative/eagle_info_v2.py
```diff
@@ -82,9 +86,31 @@ def prepare_for_decode(self: EagleDraftInput, batch: ScheduleBatch):
...
+
+        if page_size == 1:
+            new_allocate_lens = batch.seq_lens + self.ALLOC_LEN_PER_DECODE
+            num_needed_tokens = (new_allocate_lens - self.allocate_lens).sum().item()
+            out_cache_loc = alloc_token_slots(batch.tree_cache, num_needed_tokens)
+        else:
+            last_loc = get_last_loc(
+                batch.req_to_token_pool.req_to_token,
+                batch.req_pool_indices,
+                self.allocate_lens,
```
vincentzed, line 99: if `self.allocate_lens` is `None` (first call), then I have cuda IMA. The fix:
`self.allocate_lens = batch.seq_lens.clone() if self.allocate_lens is None`
Test to validate.
so topk=1, pagesize=2,4,8 etc.

You can test it like:
`SGLANG_ALLLOW_OVERWRITE_LONGER_CONTEXT_LEN=1` then 
`python -m sglang.launch_server --model-path meta-llama/Meta-Llama-3.1-8B-Instruct --speculative-algorithm EAGLE3 --speculative-draft-model-path lmsys/sglang-EAGLE3-LLama3.1-Instruct-8B --speculative-num-steps 3 --speculative-eagle-topk 1 --speculative-num-draft-tokens 4 --page-size 2 --enable-beta-spec --dtype float16`

I am try to impl topk > 1 and pagesize >1;  based off this branch.

hnyls2002: The `allocate_lens` would not be `None` even when it is the first call (first decode step after prefill). The `allocate_lens` is initialized here https://github.com/sgl-project/sglang/blob/cb405d1e6ad1a961c05280b1781c711c484d0d33/python/sglang/srt/speculative/eagle_worker_v2.py#L419

Or there are some corner cases I haven't considered. Please provide more reproduction scripts. The code on this branch can pass the unit test of `test_eagle_infer_beta.py` (by setting page to 64).

PR #11781, opened 2025-10-17
Fix install instructions and pyproject.tomls

merrymercy:
- add ` --prerelease=allow` to the install command
- sync pyproject.toml among different backends

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,14 +46,15 @@
 </details>
 
 ## About
-SGLang is a fast serving framework for large language models and vision language models.
-It makes your interaction with models faster and more controllable by co-designing the backend runtime and frontend language.
-The core features include:
+SGLang is a high-performance serving framework for large language models and vision-language models.
+It is designed to deliver low-latency and high-throughput inference across a wide range of setups, from a single GPU to large distributed clusters.
+Its core features include:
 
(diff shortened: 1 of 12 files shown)


zhyncs (changes requested):
nit: we need to update the Dockerfile too. We can just remove `BUILD_TYPE`. wdyt

https://github.com/sgl-project/sglang/blob/2614adf9caf1118e1c594ae7fe4744f489755b46/docker/Dockerfile#L94C56-L115

merrymercy:
@zhyncs good catch. I will keep `blackwell` and `blackwell_aarch64` in the pyproject.toml for now. We can delete them in the next PR.

zhyncs:
BTW @mingfeima @HaiShaw May we use `sglang-amx` and `sglang-rocm` for the PyPI installation? It'll be easier for the user to install SGLang on Intel AMX CPU and AMD GPU. wdyt (inspired by https://pypi.org/project/vllm-tpu)

zhyncs (approved):

merrymercy:
> BTW @mingfeima @HaiShaw May we use `sglang-amx` and `sglang-rocm` for the PyPI installation? It'll be easier for the user to install SGLang on Intel AMX CPU and AMD GPU. wdyt (inspired by https://pypi.org/project/vllm-tpu)

yes, we should do this

Issue #11785, opened 2025-10-17
[Bug] KV Cache is not reused for image in multi-turn text + image convo


ByronHsu:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Describe the bug

KV Cache is not reused for image in multi-turn text + image convo. For example,

1. Request 1: Hello <image_1>
2. Request 2: Hello <image_1> <image_2>

<image_1> is not cache hit when the 2nd request comes.

The root cause is we generate a hash for a list of images.

In this case, <image_1> has hash `A` in the first request. <image_1> <image_2> both have hash `B` in the second request.

The potential fix is to has a hash for each feature.

### Reproduction

1. Start a server
```bash
$ python -m sglang.launch_server --model Qwen/Qwen2.5-VL-3B-Instruct 
```

2. Run client
```bash
$ python client.py
```

3. The server logs shows no cache hit. It should hit the first image instead
```bash
[2025-10-17 23:12:45] Prefill batch. #new-seq: 1, #new-token: 328, #cached-token: 0, token usage: 0.00, #running-req: 0, #queue-req: 0, 
[2025-10-17 23:12:45] INFO:     127.0.0.1:44602 - "POST /generate HTTP/1.1" 200 OK
[2025-10-17 23:12:45] Prefill batch. #new-seq: 1, #new-token: 652, #cached-token: 3, token usage: 0.00, #running-req: 0, #queue-req: 0, 
[2025-10-17 23:12:45] INFO:     127.0.0.1:44608 - "POST /generate HTTP/1.1" 200 OK
```

```python
# client.py
import asyncio
import base64
import io

import aiohttp
from PIL import Image

IMAGE_TOKEN_TEXT = "<|vision_start|><|image_pad|><|vision_end|>"


def create_image_bytes(color, width=512, height=512, format="PNG"):
    """Create an image of the specified color and return its bytes as base64 string."""
    # Create image with specified color
    img = Image.new('RGB', (width, height), color=color)
[...]

PR #11828, opened 2025-10-19
[Feature][MM] split the images of one request into multiparts

XucSh:
## Motivation

see [issue 11785](https://github.com/sgl-project/sglang/issues/11785)

[2025-10-19 19:44:46] Prefill batch. #new-seq: 1, #new-token: 328, #cached-token: 0, token usage: 0.00, #running-req: 0, #queue-req: 0, 
[2025-10-19 19:44:46] INFO:     127.0.0.1:59486 - "POST /generate HTTP/1.1" 200 OK
[2025-10-19 19:44:46] Prefill batch. #new-seq: 1, #new-token: 327, #cached-token: 328, token usage: 0.00, #running-req: 0, #queue-req: 0, 
[2025-10-19 19:44:46] INFO:     127.0.0.1:59498 - "POST /generate HTTP/1.1" 200 OK

Cc @stmatengss @ByronHsu 

## Checklist


ByronHsu:
This breaks the following [assumption](https://github.com/openanolis/sglang/blob/f85f5e9b8a3b6dcff7c4d44f168ecb83eebf8589/python/sglang/srt/managers/schedule_batch.py#L182). I am not sure if it will have any side effect cc @mickqian @JustinTong0323 
>  One MultimodalDataItem contains all inputs for one modality. For example, if there are 3 images and 1 audio inputs, there will be 2 MultimodalDataItem. One for images and one for audio.

JustinTong0323:
> This breaks the following [assumption](https://github.com/openanolis/sglang/blob/f85f5e9b8a3b6dcff7c4d44f168ecb83eebf8589/python/sglang/srt/managers/schedule_batch.py#L182). I am not sure if it will have any side effect cc @mickqian @JustinTong0323
> 
> > One MultimodalDataItem contains all inputs for one modality. For example, if there are 3 images and 1 audio inputs, there will be 2 MultimodalDataItem. One for images and one for audio.

Yes I think it would break the internal processing logic...

XucSh:
> > This breaks the following [assumption](https://github.com/openanolis/sglang/blob/f85f5e9b8a3b6dcff7c4d44f168ecb83eebf8589/python/sglang/srt/managers/schedule_batch.py#L182). I am not sure if it will have any side effect cc @mickqian @JustinTong0323
> > > One MultimodalDataItem contains all inputs for one modality. For example, if there are 3 images and 1 audio inputs, there will be 2 MultimodalDataItem. One for images and one for audio.
> 
> Yes I think it would break the internal processing logic...

Great review, thanks! Any thoughts on the overall architecture? We can pair up on a refactor. Thanks! Cc @stmatengss @ByronHsu

python/sglang/srt/managers/mm_utils.py
```diff
@@ -261,23 +262,46 @@ def pad_input_tokens(
             return input_ids
 
         input_ids_tensor = torch.as_tensor(input_ids)
+        items_by_modality = defaultdict(list)
```
mickqian: We should at least use a switch to control this behavior, since splitting images would potentially increase the vit-processing time

mickqian (changes requested):

mickqian:
ref #9529

python/sglang/srt/managers/mm_utils.py
XucSh: Thanks for the feedback! I'll add the related logic.

XucSh:
@mickqian Please take a look.

llfl:
> Could you attach the performance w/wo splitting?

@mickqian  I tested the performance using `test_vlms_perf.py` with "Qwen/Qwen3-VL-4B-Instruct", and the results are as follows:
SGLANG_ENABLE_MM_SPLITTING=1
image

SGLANG_ENABLE_MM_SPLITTING=0
longshot20251117173804

It seems that setting SGLANG_ENABLE_MM_SPLITTING=1 adds a slight overhead in time. However, the MMMU test dataset doesn't contain duplicate images, which prevents cached tokens from being reused. I think switching to a dataset with repeated images might yield better results.

llfl:
> @llfl great, I think `bench_serving` might help here

@mickqian  thx for replying, I use `bench_serving` for a further test. I made some modifications to `bench_serving` to generate requests containing two images: one image remains consistent across all requests, while the other is randomly generated. To minimize interference from pure text inputs, I also set `random-input-len=1`. Below are the test results:

SGLANG_ENABLE_MM_SPLITTING=1
longshot20251121175311-with
cache hit status
longshot20251121175346-with

SGLANG_ENABLE_MM_SPLITTING=0
longshot20251121175515-without

Results show: all requests with `SGLANG_ENABLE_MM_SPLITTING=1` hit cached tokens, and a performance improvement can be observed.

ShangmingCai:
CC: @gty111 @liusy58 @ZhengWG

liusy58:
## Problem
When a request contains multiple images [A, B, C], they are hashed together as one bundle. If another request has [A, B, D], we cannot reuse the cache for images A and B.

## Solution
Split bundled images into individual items so each image has its own hash.

Before: hash([A, B, C]) vs hash([A, B, D]) → no cache hit

After: hash(A), hash(B), hash(C) vs hash(A), hash(B), hash(D) → A and B hit cache 

This is a great PR! I think we should merge this.

ShangmingCai (commented):
LGTM. But still need a double-check from @mickqian 

minor: is it possible that we wrap some logic in the `def from_dict(obj: dict):` to a mm util? The code seems pretty long in the scheduler batch.

ShangmingCai (approved):
Thx. This looks better to me now.

ShangmingCai:
> LGTM. I think it can be merged. @ShangmingCai

@yhyang201 OK, let's get it merged. If any bug has been reported, we can do a quick fix or revert.

PR #11846, opened 2025-10-20
Revert "[CI Monitor] Ci monitor only deal with main branch in default"

BBuf:
Reverts sgl-project/sglang#11538


Kangyan-Zhou:
@BBuf what happened with this change?

PR #11870, opened 2025-10-20
[feature] Add layerwise NVTX support

kyleliang-nv:
## Motivation

Enable auto-inserting NVTX markers that contains layer name and input tensor shapes.

## Modifications

- Add `pytorch_hooks.py` which will add pre/post-foward hooks with pytorch.
- Add `enable-layerwise-nvtx-marker` server arg. This will register the loaded model with pytorch_hooks.py
- Add unittest for `/start_profile + nsys profile`
- Move Prefill log from before to after run_batch, to be consistent with Decode log
- Add documents to explain how to use `start_profile` with `start_steps` and `num_steps`
- Add document to explain how to use `enable-layerwise-nvtx-marker` with `/start_profile` endpoint

## Benchmarking and Profiling

With `enable-layerwise-nvtx-marker`, the nsys output file will now contain NVTX which corresponds to the pytorch nn.module path and input tensor shapes.
Screenshot 2025-10-20 at 12 00 50 PM

## Checklist


test/srt/test_start_profile.py
```diff
@@ -1,10 +1,19 @@
 """
 Usage:
-python3 -m unittest test_srt_engine.TestSRTEngine.test_4_sync_async_stream_combination
+# From the test/srt directory:
```
Fridge003, line 3: Please add this test to 1-gpu nightly test in `run_suite.py`

python/sglang/srt/server_args.py
```diff
@@ -3232,6 +3233,11 @@ def add_cli_args(parser: argparse.ArgumentParser):
             action="store_true",
             help="Enable garbage collection during CUDA graph capture. If disabled (default), GC is frozen during capture to speed up the process.",
         )
+        parser.add_argument(
```
Fridge003, line 3244: Please also update it here 
https://github.com/sgl-project/sglang/blob/main/docs/advanced_features/server_arguments.md

python/sglang/srt/utils/nvtx_pytorch_hooks.py
```diff
@@ -0,0 +1,308 @@
...
+        Returns:
+            param_info(dict): Parameter meta_data for the given op.
+
+        Raises:
+            None
+
+        """
+        param_info = {}
+        ## Look for layer parameters specific to each layer type
+        if (
```
Fridge003, line 88: We only need to keep the ones useful for LLM. Conv/LSTM related layers don't need to be considered

python/sglang/srt/utils/nvtx_pytorch_hooks.py
```diff
@@ -0,0 +1,308 @@
+# Copyright 2023-2024 SGLang Team
```
Fridge003, line 1: Can rename this file to `nvtx_pytorch_hooks.py`

test/srt/test_start_profile.py
Fridge003: Please attach a screenshot of this test result

kyleliang-nv: I added it into 1-gpu nightly test.
This is the screenshot of a local test run.
Screenshot 2025-11-14 at 9 46 43 PM

python/sglang/srt/server_args.py
kyleliang-nv: Updated the server_argument.md. Please review again.

python/sglang/srt/utils/nvtx_pytorch_hooks.py
kyleliang-nv: I removed the LSTM stuff. But don't we want to include the Conv stuff, maybe for VisionEncoders in VLM?

python/sglang/srt/utils/nvtx_pytorch_hooks.py
kyleliang-nv: Renamed.

python/sglang/srt/utils/nvtx_pytorch_hooks.py
Fridge003: Upd: Conv can be kept since some VLMs might need. LSTM I guess is useless

test/srt/run_suite.py
```diff
@@ -213,8 +213,9 @@ class TestFile:
         TestFile("lora/test_lora_openai_api.py", 30),
         TestFile("openai_server/features/test_lora_openai_compatible.py", 150),
         TestFile("batch_invariant/test_batch_invariant_ops.py", 10),
-        TestFile("test_deepseek_v3_deterministic.py", 240),
         TestFile("test_cpp_radix_cache.py", 60),
+        TestFile("test_deepseek_v3_deterministic.py", 240),
+        TestFile("test_start_profile.py", 180),
```
Fridge003:
```suggestion

```

Fridge003: Just found test_start_profile.py has been covered in per-commit CI, so no need to put it in nightly then

Fridge003 (approved):
Nice job

kyleliang-nv:
> Nice job

Thanks for reviewing it and merging it in @Fridge003 !

PR #11874, opened 2025-10-20
fa3 & trtllm_mha spec overlap

JustinTong0323:
## Motivation

Enable fa3 & trtllm_mha for spec_v2 (for gpt-oss on Hopper & Blackweel).

## Accuracy Tests
Test on gpt-oss and llama4, accuracy could match.

Detailed test result could be find [here](https://docs.google.com/spreadsheets/d/1ztZDT62e5MFgEyOQhX-4xkeI_Z8uXhB2oRxpPgz3lAk/edit?gid=2036762452#gid=2036762452).

## Checklist


python/sglang/srt/speculative/eagle_worker_v2.py
hnyls2002, line 546: The API has already been defined here, and we actually do not need to clear the cache (cache is cleared in the scheduler).

python/sglang/srt/speculative/eagle_worker_v2.py
```diff
@@ -706,3 +706,7 @@ def move_accepted_tokens_to_target_kvcache(
         self.token_to_kv_pool_allocator.get_kvcache().move_kv_cache(
             tgt_cache_loc, accepted_out_cache_loc
         )
+
+    def clear_cache_pool(self):
```
hnyls2002: I think this method can be removed.

hnyls2002 (approved):

hnyls2002:
@JustinTong0323 Already approved, thanks for the support!

python/sglang/srt/speculative/eagle_worker_v2.py
JustinTong0323: Cool! Got it, as this PR is old so at that time this method is not introduced haha

PR #11883, opened 2025-10-21
[overlap-spec] remove draft_extend_v2 & don't capture draft extend on spec v2

cicirori:
When modifying the internal attention backend, the introduction of draft_extend_v2 has made things slightly messy.

The Triton attention backend and logits here enable spec overlap, and the behavior of draft_extend changing seems to be a rare occurrence.
More often, the original draft extend logic is reused.

Perhaps using the original draft_extend + enabling the overlap scheduler for runtime decisions would be clearer?


python/sglang/srt/layers/attention/triton_backend.py
```diff
@@ -347,7 +350,10 @@ def init_forward_metadata(self, forward_batch: ForwardBatch):
             attn_logits = None
             attn_lse = None
 
-        elif forward_batch.forward_mode.is_draft_extend():
+        elif (
```
hnyls2002, line 353: This logic is also messy... They reason why we introduce `draft_extend_v2` is to introduce another way to process logits output/draft extend forward.

PR #11892, opened 2025-10-21
DeepSeek-V3.2: Add Adaptive MHA Attention Pathway for Short-Sequence Prefill

YAMY1234:
## Motivation

For DeepSeek-V3.2 models, using MLA (Multi-Latent Attention) uniformly across all sequence lengths during prefill is suboptimal. For short sequences, the overhead of MLA's compression/decompression and absorbed attention mechanism outweighs any potential benefits, making standard MHA (Multi-Head Attention) more efficient. This PR implements adaptive attention mechanism selection based on sequence length to optimize inference performance across different workloads.

## Modifications

This PR implements sequence-length-based adaptive attention mechanism selection in the NSA (Native Sparse Attention) backend:

### Core Changes

1. **Add MHA mode detection and processing in `nsa_backend.py`**
   - Implement `_forward_standard_mha()` method using FlashAttention varlen interface for standard MHA
   - Support both prefix chunk processing and regular token processing
   - Correctly handle LSE (log-sum-exp) return values for DP attention compatibility

2. **Implement intelligent attention mechanism selection in `deepseek_v2.py`**
   - Add adaptive selection logic in `handle_attention_nsa()` based on sequence length:
     - **Decode phase**: Always use MLA (avoids per-token decompression overhead)
     - **Prefill with seq_len < 2048**: Use MHA_CHUNKED_KV (lower computational overhead)
     - **Prefill with seq_len >= 2048**: Use MLA (absorbed attention mechanism becomes beneficial at scale)
   - Add configurable `nsa_seq_len_threshold` parameter (default: 2048) for flexible threshold tuning

Note: This threshold is also chosen because of NSA's topk sparse filtering, which operates within the MLA pathway.

### GSM8K:
```
hostuser@1ad3348c50a4:/sgl-workspace/sglang$ python3 benchmark/gsm8k/bench_sglang.py --num-shots 4 --num-questions 100 --parallel 100
100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 100/100 [00:07<00:00, 13.10it/s]
Accuracy: 0.970
Invalid: 0.000
Latency: 8.016 s
Output throughput: 1080.888 token/s
```

### GPQA:
```
(optional:  export SGLANG_MHA_USE_WKC_WVC=1)
python -m sglang.launch_server --model deepseek-ai/DeepSeek-V3.2-Exp --tp 8 --dp 8 --enable-dp-attention

```

```
With export SGLANG_MHA_USE_WKC_WVC=1: 
Repeat: 8, mean: 0.798█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▍        | 187/198 [27:45<01:04,  5.87s/it]
Scores: ['0.793', '0.788', '0.818', '0.783', '0.813', '0.763', '0.813', '0.813']███████████████████████████████████████████████████████████████████████▍        | 187/198 [29:01<00:52,  4.75s/it]
====================███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▍        | 187/198 [28:03<00:37,  3.37s/it]
{'chars': np.float64(14716.31313131313), 'chars:std': np.float64(11968.213414131249), 'score:std': np.float64(0.3898060809385348), 'score': np.float64(0.8131313131313131)}
Total latency: 1666.190 s
Score: 0.813
```

```
Without export SGLANG_MHA_USE_WKC_WVC=1: 
Repeat: 8, mean: 0.790
Scores: ['0.793', '0.768', '0.788', '0.808', '0.773', '0.793', '0.818', '0.783']█████████████████████████████████████████████████████████████████████▊          | 185/198 [27:52<00:49,  3.81s/it]
====================█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▊          | 185/198 [27:00<00:46,  3.59s/it]
{'chars': np.float64(15173.823232323231), 'chars:std': np.float64(11838.867709818003), 'score:std': np.float64(0.41232046084617835), 'score': np.float64(0.7828282828282829)}
Total latency: 1652.339 s
Score: 0.783
```

#### 1024 tokens, 100 requests

| Metric             | Original (MLA) | This PR (MHA)      | Improvement |
| ------------------ | -------------- | ------------------ | ----------- |
| Mean E2E Latency   | 4334.95 ms     | **3820.75 ms**     | **11.9%**   |
| Mean TTFT          | 4279.27 ms     | **3820.73 ms**     | **10.7%**   |
| Input Throughput   | 14213.70 tok/s | **15479.94 tok/s** | **8.9%**    |
| Benchmark Duration | 7.20 s         | **6.62 s**         | **8.1%**    |

#### 512 tokens, 100 requests

| Metric             | Original (MLA) | This PR (MHA)      | Improvement |
| ------------------ | -------------- | ------------------ | ----------- |
| Mean E2E Latency   | 3060.32 ms     | **2819.46 ms**     | **7.9%**    |
| Mean TTFT          | 3060.29 ms     | **2977.96 ms**     | **2.7%**    |
| Input Throughput   | 9909.07 tok/s  | **10662.89 tok/s** | **7.6%**    |
| Benchmark Duration | 5.17 s         | **5.30 s**         | **−2.5%**   |

#### 2000 tokens, 100 requests

| Metric             | Original (MLA) | This PR (MHA)      | Improvement |
| ------------------ | -------------- | ------------------ | ----------- |
| Mean E2E Latency   | 5797.70 ms     | **4819.40 ms**     | **16.9%**   |
| Mean TTFT          | 5601.00 ms     | **4819.37 ms**     | **14.0%**   |
| Input Throughput   | 19331.60 tok/s | **22249.25 tok/s** | **15.1%**   |
| Benchmark Duration | 10.35 s        | **8.99 s**         | **13.1%**   |

#### Forward Total Time
- **MHA + Skip TopK**: 2-2.5 ms
image

- **Original**: 3.5-4 ms
image
- **Improvement**: ~30-40% reduction

#### Prepare Stage
- **MHA + Skip TopK** (`forward_normal_prepare`/`forward_normal_chunked_kv_prepare` + `topk skipped indexer`): 
- 1.5 -2 ms
image

- **Original** (`forward_absorb_prepare`): 
- 2.5 - 3.5 ms
image

- **Improvement**: ~30% reduction

#### Core Stage
- **This PR** (`forward_normal_chunked_kv_core`): ~300 μs
image
- **Original** (`forward_absorb_core`): 500-600 μs
image
- **Improvement**: ~40-50% reduction

## Checklist

### DIFF
--- a/python/sglang/srt/layers/attention/nsa/nsa_indexer.py
+++ b/python/sglang/srt/layers/attention/nsa/nsa_indexer.py
@@ -484,6 +484,7 @@ class Indexer(CustomOp):
         positions: torch.Tensor,
         forward_batch: ForwardBatch,
         layer_id: int,
+        skip_topk_computation: bool = False,
     ) -> Optional[torch.Tensor]:
         if is_hip():
             from sglang.srt.layers.attention.nsa.tilelang_kernel import act_quant
@@ -538,6 +539,10 @@ class Indexer(CustomOp):
             index_k_scale=k_scale,
         )
 
+        # Optimization: skip topk calculation if caller (e.g., MHA short path) doesn't need it
+        if skip_topk_computation:
+            return None
+
         if not self.fuse_wk_and_weights_proj:
             weights, _ = self.weights_proj(x)
         weights = self._get_logits_head_gate(weights, q_scale)
@@ -582,6 +587,7 @@ class Indexer(CustomOp):
         positions: torch.Tensor,
         forward_batch: ForwardBatch,
         layer_id: int,
+        skip_topk_computation: bool = False,  # Keep signature consistent; Can be supported in the future
     ) -> torch.Tensor:
         import custom_ops  # noqa: F401
         import torch_npu
--- a/python/sglang/srt/layers/attention/nsa/nsa_indexer.py
+++ b/python/sglang/srt/layers/attention/nsa/nsa_indexer.py
@@ -261,64 +261,6 @@ class Indexer(CustomOp):
 
         return query, key, weights
 
-    def _construct_attention_indices(
-        self,
-        metadata: BaseIndexerMetadata,
-        device: torch.device,
-    ) -> torch.Tensor:
-        """Construct attention indices [0, 1, ..., visible_len-1, -1, ...]"""
-        seq_lens_expanded = metadata.get_seqlens_expanded()
-
-        row = torch.arange(self.index_topk, device=device, dtype=torch.int32).unsqueeze(
-            0
-        )
-        lens = seq_lens_expanded.to(device=device, dtype=torch.int32).unsqueeze(1)
-        valid = row < lens
-        indices_local = torch.where(valid, row, torch.full_like(row, -1))
-
-        from sglang.srt.layers.attention.nsa_backend import TopkTransformMethod
-
-        method = getattr(metadata, "topk_transform_method", None)
-
-        if method == TopkTransformMethod.PAGED:
-            page_table_1 = (
-                metadata.attn_metadata.page_table_1
-                if hasattr(metadata, "attn_metadata")
-                and hasattr(metadata.attn_metadata, "page_table_1")
-                else metadata.get_page_table_64()
-            )
-            extend_lens_cpu = (
-                getattr(metadata.attn_metadata, "nsa_extend_seq_lens_list", None)
-                if hasattr(metadata, "attn_metadata")
-                else None
-            )
-
-            from sglang.srt.layers.attention.nsa.transform_index import (
-                transform_index_page_table_prefill,
-            )
-
-            return transform_index_page_table_prefill(
-                page_table=page_table_1,
-                topk_indices=indices_local,
-                extend_lens_cpu=extend_lens_cpu,
-                page_size=1,
-            )
-
-        elif method == TopkTransformMethod.RAGGED:
-            off = (
-                getattr(metadata.attn_metadata, "topk_indices_offset", None)
-                if hasattr(metadata, "attn_metadata")
-                else None
-            )
-            if off is not None:
-                off = off.to(device=device, dtype=torch.int32)
-                if off.ndim == 1:
-                    off = off.unsqueeze(1)
-                mask = indices_local != -1
-                return torch.where(mask, indices_local + off, indices_local)
-
-        return indices_local
-
     def _get_k_bf16(
         self,
         x: torch.Tensor,
@@ -500,12 +442,20 @@ class Indexer(CustomOp):
             index_k_scale=k_scale,
         )
 
-        # MHA doesn't need topk_indices, early return for efficiency
+        # MHA doesn't need topk_indices
         if not return_indices:
             return None
 
-        # MLA: construct return_indices
-        return self._construct_attention_indices(metadata, x.device)
+        # MLA: use dummy logits with topk kernel's fast path to generate indices
+        # When length <= 2048, naive_topk_cuda directly generates [0,1,...,length-1,-1,...]
+        seq_lens_expanded = metadata.get_seqlens_expanded()
+        dummy_logits = torch.zeros(
+            seq_lens_expanded.shape[0],
+            self.index_topk,
+            dtype=torch.float32,
+            device=x.device,
+        )
+        return metadata.topk_transform(dummy_logits, self.index_topk)
 
     def forward_indexer(
         self,
@@ -624,13 +574,11 @@ class Indexer(CustomOp):
             return None
 
         # Determine if should skip topk based on sequence length
-        from sglang.srt.environ import envs
-
         should_skip = False
         if not forward_batch.forward_mode.is_decode_or_idle():
             if forward_batch.seq_lens_cpu is not None:
                 max_kv_len = forward_batch.seq_lens_cpu.max().item()
-                should_skip = max_kv_len <= envs.SGLANG_NSA_SEQ_LEN_THRESHOLD.value
+                should_skip = max_kv_len <= self.index_topk
 
         # Optimization: fast path when skipping topk computation
         if should_skip:
--- a/python/sglang/srt/layers/attention/nsa/nsa_indexer.py
+++ b/python/sglang/srt/layers/attention/nsa/nsa_indexer.py
@@ -261,6 +261,64 @@ class Indexer(CustomOp):
 
         return query, key, weights
 
+    def _construct_attention_indices(
+        self,
+        metadata: BaseIndexerMetadata,
+        device: torch.device,
+    ) -> torch.Tensor:
+        """Construct attention indices [0, 1, ..., visible_len-1, -1, ...]"""
+        seq_lens_expanded = metadata.get_seqlens_expanded()
+
+        row = torch.arange(self.index_topk, device=device, dtype=torch.int32).unsqueeze(
+            0
+        )
+        lens = seq_lens_expanded.to(device=device, dtype=torch.int32).unsqueeze(1)
+        valid = row < lens
+        indices_local = torch.where(valid, row, torch.full_like(row, -1))
+
+        from sglang.srt.layers.attention.nsa_backend import TopkTransformMethod
+
+        method = getattr(metadata, "topk_transform_method", None)
+
+        if method == TopkTransformMethod.PAGED:
+            page_table_64 = (
+                metadata.attn_metadata.page_table_1
+                if hasattr(metadata, "attn_metadata")
+                and hasattr(metadata.attn_metadata, "page_table_1")
+                else metadata.get_page_table_64()
+            )
+            extend_lens_cpu = (
+                getattr(metadata.attn_metadata, "nsa_extend_seq_lens_list", None)
+                if hasattr(metadata, "attn_metadata")
+                else None
+            )
+
+            from sglang.srt.layers.attention.nsa.transform_index import (
+                transform_index_page_table_prefill,
+            )
+
+            return transform_index_page_table_prefill(
+                page_table=page_table_64,
+                topk_indices=indices_local,
+                extend_lens_cpu=extend_lens_cpu,
+                page_size=64,
+            )
+
+        elif method == TopkTransformMethod.RAGGED:
+            off = (
+                getattr(metadata.attn_metadata, "topk_indices_offset", None)
+                if hasattr(metadata, "attn_metadata")
+                else None
+            )
+            if off is not None:
+                off = off.to(device=device, dtype=torch.int32)
+                if off.ndim == 1:
+                    off = off.unsqueeze(1)
+                mask = indices_local != -1
+                return torch.where(mask, indices_local + off, indices_local)
+
+        return indices_local
+
     def _get_k_bf16(
         self,
         x: torch.Tensor,
@@ -268,53 +326,20 @@ class Indexer(CustomOp):
         enable_dual_stream: bool,
     ):
         # Compute only key, skip query and weights (weights is discarded if fused)
-        if enable_dual_stream:
-            current_stream = torch.cuda.current_stream()
-            self.alt_stream.wait_stream(current_stream)
-
-            with torch.cuda.stream(self.alt_stream):
-                with deep_gemm_wrapper.configure_deep_gemm_num_sms(
-                    self.half_device_sm_count
-                ):
-                    if self.fuse_wk_and_weights_proj:
-                        key, _ = self.fused_wk_and_weights_proj(x)[0].split(
-                            [self.head_dim, self.n_heads], dim=-1
-                        )
-                    else:
-                        key, _ = self.wk(x)
-                    key = self.k_norm(key)
-
-                    k_rope, _ = torch.split(
-                        key,
-                        [self.rope_head_dim, self.head_dim - self.rope_head_dim],
-                        dim=-1,
-                    )
-
-            current_stream.wait_stream(self.alt_stream)
-        else:
-            if self.fuse_wk_and_weights_proj:
-                key, _ = self.fused_wk_and_weights_proj(x)[0].split(
-                    [self.head_dim, self.n_heads], dim=-1
-                )
-            else:
-                key, _ = self.wk(x)
-            key = self.k_norm(key)
-            k_rope, _ = torch.split(
-                key, [self.rope_head_dim, self.head_dim - self.rope_head_dim], dim=-1
+        if self.fuse_wk_and_weights_proj:
+            key, _ = self.fused_wk_and_weights_proj(x)[0].split(
+                [self.head_dim, self.n_heads], dim=-1
             )
+        else:
+            key, _ = self.wk(x)
+        key = self.k_norm(key)
+        k_rope, _ = torch.split(
+            key, [self.rope_head_dim, self.head_dim - self.rope_head_dim], dim=-1
+        )
 
         _, k_rope = self.rotary_emb(positions, k_rope, k_rope)
         key[..., : self.rope_head_dim] = k_rope
-
-        if enable_dual_stream:
-            current_stream = torch.cuda.current_stream()
-            self.alt_stream.wait_stream(current_stream)
-
-            with torch.cuda.stream(self.alt_stream):
-                key = rotate_activation(key)
-            current_stream.wait_stream(self.alt_stream)
-        else:
-            key = rotate_activation(key)
+        key = rotate_activation(key)
 
         return key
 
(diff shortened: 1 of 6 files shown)


Fridge003:
@YAMY1234 Can you post some accuracy results on GPQA ?

hlu1:
Please test with MTP and make sure it's not broken.

YAMY1234:
> @YAMY1234 Can you post some accuracy results on GPQA ?

Thanks! Just added the GPQA result.

YAMY1234:
> Please test with MTP and make sure it's not broken.

Ensured MHA is currently triggered without MTP. Will test with MTP later to confirm it runs correctly.

hlu1:
Please update the benchmark data after adding back `self.indexer`.

YAMY1234:
> Please update the benchmark data after adding back `self.indexer`.

Updated the results with adding back indexer but skipping topK!

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -402,6 +402,36 @@ def handle_attention_aiter(attn, forward_batch):
...
+    if forward_batch.forward_mode.is_decode_or_idle():
+        return AttnForwardMethod.MLA
+
+    if _is_extend_without_speculative(forward_batch):
+        NSA_THRESHOLD = getattr(attn, "nsa_seq_len_threshold", 2048)
+
+        if forward_batch.seq_lens_cpu is not None:
+            max_kv_len = forward_batch.seq_lens_cpu.max().item()
+        else:
+            max_kv_len = 0
```
hlu1:
```suggestion
        assert forward_batch.seq_lens_cpu is not None
        max_kv_len = forward_batch.seq_lens_cpu.max().item()
```

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -402,6 +402,36 @@ def handle_attention_aiter(attn, forward_batch):
...
+
+        # B200 (SM100) is temporarily disabled for MHA due to FA4 accuracy issues
+        # Currently only H200 (SM90) with FA3 is allowed to use MHA path
+        is_supported_gpu = _device_sm == 90
+
+        return (
+            AttnForwardMethod.MHA_CHUNKED_KV
```
hlu1: Why `MHA_CHUNKED_KV`? As we discussed offline, it should use `MHA` or `MHA_ONE_SHOT` for best perf.

python/sglang/srt/layers/attention/nsa/nsa_indexer.py
```diff
@@ -538,6 +539,10 @@ def forward_cuda(
             index_k_scale=k_scale,
         )
 
+        # Optimization: skip topk calculation if caller (e.g., MHA short path) doesn't need it
+        if skip_topk_computation:
+            return None
+
```
hlu1: There are more stuff you can skip. You only need to keep ops that are relevant to k and you can skip all the ops that relevant to q or weights (except when wk and weight_proj are fused after the `fuse_wk_and_weights_proj` optimization). To avoid mess up the code, it's better to add another `_forward_cuda_k_only()` for this code path.

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -1498,10 +1545,30 @@ def forward_normal_prepare(
...
-        kv = kv.view(-1, self.num_local_heads, self.qk_nope_head_dim + self.v_head_dim)
-        k_nope = kv[..., : self.qk_nope_head_dim]
-        v = kv[..., self.qk_nope_head_dim :]
+
+        # Optional:Use w_kc/w_vc to build K/V from latent (numerical alignment with MLA)
+        if (
+            self.w_kc is not None
+            and self.w_vc is not None
+            and envs.SGLANG_MHA_USE_WKC_WVC.get()
+        ):
```
hlu1: Is there a lot of numerical difference between the two code paths? If not, we should remove this one and keep the other one.

python/sglang/srt/models/deepseek_v2.py
YAMY1234: When I was debugging before calling indexer, I initially tried `AttnForwardMethod.MHA`, but found it only works correctly only when there’s no prefix. If we want to use that we need to add additional logic needed for prefix processing, which is similar to what MHA_CHUNKED_KV does. 

But thanks for reminding, I just found the `MHA_ONE_SHOT` should be able to handle this scenario. I’ll test whether it works well in our case.

python/sglang/srt/layers/attention/nsa/nsa_indexer.py
YAMY1234: Great suggestion! Will update in another patch.

python/sglang/srt/models/deepseek_v2.py
YAMY1234: Enabling `SGLANG_MHA_USE_WKC_WVC` slightly improves logits precision — in each forward batch, the max diff between MLA and MHA logits decreases from about 0.8 to 0.25. But since we’re not using FP8 for MHA, the difference is negligible in practice(as the GPQA score does not show too much difference). I agree it’s better to remove this path for simplicity and it’ll make the code cleaner overall. Will do it in the next patch

python/sglang/srt/models/deepseek_v2.py
YAMY1234: Tested `MHA_ONE_SHOT` can work and the GPQA result looks good. But since there is a `sum_seq_lens` restriction so I still kept `MHA_CHUNKED_KV` as fallback if `sum_seq_lens` is larger than `forward_batch.get_max_chunk_capacity`.
```
def _support_mha_one_shot(attn: DeepseekV2AttentionMLA, forward_batch, backend_name):
    attn_supported = backend_name in ["fa3", "flashinfer", "flashmla"]
    sum_seq_lens = (
        sum(forward_batch.seq_lens_cpu) if forward_batch.seq_lens_cpu is not None else 0
    )
    return attn_supported and sum_seq_lens <= forward_batch.get_max_chunk_capacity()
```

Fridge003:
@YAMY1234 Please fix lint

YAMY1234:
> @YAMY1234 Please fix lint

@Fridge003  Done! Thanks

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -402,6 +402,41 @@ def handle_attention_aiter(attn, forward_batch):
...
+
+    if _is_extend_without_speculative(forward_batch):
+        NSA_THRESHOLD = getattr(attn, "nsa_seq_len_threshold", 2048)
+
+        assert forward_batch.seq_lens_cpu is not None
+        max_kv_len = forward_batch.seq_lens_cpu.max().item()
+
+        # B200 (SM100) is temporarily disabled for MHA due to FA4 accuracy issues
+        # Currently only H200 (SM90) with FA3 is allowed to use MHA path
+        is_supported_gpu = _device_sm == 90
```
Fridge003: Rename it to `is_hopper`

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -1140,6 +1175,8 @@ def __init__(
 
         self.use_nsa = is_deepseek_nsa(config)
         if self.use_nsa:
+            # MHA/MLA selection threshold: MHA for short (<2048), MLA+NSA for long sequences
+            self.nsa_seq_len_threshold = getattr(config, "nsa_seq_len_threshold", 2048)
```
Fridge003: Can we make the threshold an environment variable, and put it under https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/environ.py (Also update the document for environ https://github.com/sgl-project/sglang/blob/main/docs/references/environment_variables.md)

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -1478,8 +1515,21 @@ def forward_normal_prepare(
...
-            q = self.q_a_layernorm(q)
-            q = self.q_b_proj(q)[0].view(-1, self.num_local_heads, self.qk_head_dim)
+            q_lora = self.q_a_layernorm(q)
+            q = self.q_b_proj(q_lora)[0].view(
+                -1, self.num_local_heads, self.qk_head_dim
+            )
+
+            # NSA Indexer Call for MHA Path: only cache quantized keys, skip topk
+            if self.use_nsa and _is_extend_without_speculative(forward_batch):
+                _ = self.indexer(
```
Fridge003, line 1516: Seems now we will skip index topk  for any MHA forward.
But how can we ensure we don't skip the index topk when max_len_kv is more than index_topk(2048) ? 

What I mean is, the logic of checking whether to skip topk should be discoupled from mha/mla. For MLA we might also skip topk when length is short enough.

python/sglang/srt/layers/attention/nsa_backend.py
```diff
@@ -1154,6 +1175,87 @@ def _forward_flashmla_kv(
...
+            f"batch_size mismatch: cu_seqlens_q has {len(cu_seqlens_q)-1} requests, "
+            f"cu_seqlens_k has {len(cu_seqlens_k)-1} requests"
+        )
+
+        # Determine FA version: FA3 for SM90, FA4 for SM100 (B200)
+        device_sm = (
+            torch.cuda.get_device_capability()[0] * 10
+            + torch.cuda.get_device_capability()[1]
+        )
+        fa_version = 4 if device_sm == 100 else 3
```
Fridge003: Can we only use `torch.cuda.get_device_capability()[0]` as condition?
Since other devices like SM103 might need fa4 in the future.

python/sglang/srt/layers/attention/nsa/nsa_indexer.py
```diff
@@ -394,6 +451,39 @@ def _get_topk_ragged(
...
+        positions: torch.Tensor,
+        forward_batch: ForwardBatch,
+        layer_id: int,
+        act_quant,
+        enable_dual_stream: bool,
+    ) -> None:
+        # Fast path: only compute and store k cache, skip all q and weights ops
+        key = self._get_k_bf16(x, positions, enable_dual_stream)
+
+        if enable_dual_stream:
```
Fridge003: The same, do we need two streams here

python/sglang/srt/layers/attention/nsa/nsa_indexer.py
```diff
@@ -261,6 +261,63 @@ def _get_q_k_bf16(
...
         return query, key, weights
 
+    def _get_k_bf16(
+        self,
+        x: torch.Tensor,
+        positions: torch.Tensor,
+        enable_dual_stream: bool,
+    ):
+        # Compute only key, skip query and weights (weights is discarded if fused)
+        if enable_dual_stream:
```
Fridge003: Do we really need to use two streams here? Seems current_stream is doing nothing when alt_stream is computing key

python/sglang/srt/layers/attention/nsa/nsa_indexer.py
```diff
@@ -261,6 +261,63 @@ def _get_q_k_bf16(
...
+                key, _ = self.wk(x)
+            key = self.k_norm(key)
+            k_rope, _ = torch.split(
+                key, [self.rope_head_dim, self.head_dim - self.rope_head_dim], dim=-1
+            )
+
+        _, k_rope = self.rotary_emb(positions, k_rope, k_rope)
+        key[..., : self.rope_head_dim] = k_rope
+
+        if enable_dual_stream:
```
Fridge003: The same, do we need two streams here

python/sglang/srt/models/deepseek_v2.py
YAMY1234: Great suggestion! I’ve modified the code so that it now supports MLA’s Top-K skip. For better compatibility, it can also return the constructed indices directly without performing extra computations, which saves some processing overhead. I verified this works on both H200(MHA) and B200(MLA). Please take a look and let me know if the changes make sense or need any adjustment🙏

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -402,6 +402,41 @@ def handle_attention_aiter(attn, forward_batch):
...
+    - Prefill <= 2048: MHA (topk ineffective, MHA has lower FLOPs)
+    - Prefill > 2048: MLA (topk filtering reduces computation significantly)
+
+    TODO: B200 (SM100) MHA path is temporarily disabled due to FA4 gpqa accuracy issues.
+    """
+    if forward_batch.forward_mode.is_decode_or_idle():
+        return AttnForwardMethod.MLA
+
+    if _is_extend_without_speculative(forward_batch):
+        NSA_THRESHOLD = envs.SGLANG_NSA_SEQ_LEN_THRESHOLD.value
```
hlu1: I don't think we should introduce an env variable for this purpose. It's set to 2048 because `self.nsa_index_topk` is 2048. We should just use `self.nsa_index_topk`.  If in another model, the `self.nsa_index_topk` value is set differently, this would have introduced a bug.

python/sglang/srt/models/deepseek_v2.py
hlu1: Let's think about what this case, `sum_seq_lens` larger than `forward_batch.get_max_chunk_capacity`, actually means. `forward_batch.get_max_chunk_capacity` is set to 128k. When `sum_seq_lens` is larger than 128k, that means we have significant kvcache reuse or chunked prefill going on, in that case, using sparse MLA could be the optimal choice. We need to run some benchmarks to see what the best choice is.

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -1478,8 +1513,21 @@ def forward_normal_prepare(
...
                 [self.q_lora_rank, self.kv_lora_rank + self.qk_rope_head_dim], dim=-1
             )
-            q = self.q_a_layernorm(q)
-            q = self.q_b_proj(q)[0].view(-1, self.num_local_heads, self.qk_head_dim)
+            q_lora = self.q_a_layernorm(q)
+            q = self.q_b_proj(q_lora)[0].view(
+                -1, self.num_local_heads, self.qk_head_dim
+            )
+
+            # NSA Indexer: cache quantized keys, auto-skip topk for short sequences
```
hlu1:
```suggestion
            # NSA Indexer: cache quantized keys, auto-skip topk for sequences <= nsa_index_topk
```

python/sglang/srt/layers/attention/nsa/nsa_indexer.py
```diff
@@ -261,6 +261,88 @@ def _get_q_k_bf16(
 
         return query, key, weights
 
+    def _construct_attention_indices(
```
hlu1: I think it's way easier to construct an empty matrix the same size as the logits matrix [here](https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/layers/attention/nsa/nsa_indexer.py#L361) and pass that to the topk kernel so it can generate the indices for you. It calls one kernel instead of the multiples kernels down below. The topk kernel generates the indices directly when the length is <= 2k. See https://github.com/sgl-project/sglang/blob/34f7564df0378d2dba929549c0350cbc12713666/sgl-kernel/csrc/elementwise/topk.cu#L35-L40

python/sglang/srt/models/deepseek_v2.py
YAMY1234: I previously tried producing scenarios where `sum_seq_lens` exceeds `forward_batch.get_max_chunk_capacity` to verify performance/correctness, but so far haven’t seen it happen — even when testing different `num-shot` settings on GSM8K and full GPQA Thinking runs. It seems this branch is rarely triggered, especially when prefill length < 2048. So I kept it mainly as a safety fallback in case we later hit edge cases with longer prefix reuse.

python/sglang/srt/layers/attention/nsa/nsa_indexer.py
```diff
@@ -394,6 +418,45 @@ def _get_topk_ragged(
...
+        # MLA: use dummy logits with topk kernel's fast path to generate indices
+        # When length <= 2048, naive_topk_cuda directly generates [0,1,...,length-1,-1,...]
+        seq_lens_expanded = metadata.get_seqlens_expanded()
+        dummy_logits = torch.zeros(
+            seq_lens_expanded.shape[0],
+            self.index_topk,
+            dtype=torch.float32,
+            device=x.device,
+        )
+        return metadata.topk_transform(dummy_logits, self.index_topk)
```
hlu1, line 458: Have you tested this code path?

python/sglang/srt/models/deepseek_v2.py
hlu1: If it's rarely hit, why don't we just fall back to MLA, instead of adding another code path which adds the maintenance overhead.

python/sglang/srt/layers/attention/nsa/nsa_indexer.py
YAMY1234: Yes, verified on B200 with GPQA thinking

python/sglang/srt/models/deepseek_v2.py
YAMY1234: Make sense, removed the chunked kv logic completely to reduce maintenance overhead

hlu1 (approved):

Fridge003 (approved):

Fridge003:
Great job! @YAMY1234

thqq479:
I noticed that decoding in a context smaller than 2048 also spends a lot of time in the indexer() function. Why is that? Can't we skip this step during decoding? @YAMY1234

PR #11893, opened 2025-10-21
fix: draft model revision misuse model revision

gongwei-130:
## Checklist


Qiaolin-Yu:
Could you provide more context for this?

gongwei-130:
> Could you provide more context for this?
@Qiaolin-Yu sure. in model runner, if don't pass model_revision in, it will use server_args.revision, which is model revision, but not speculative_draft_model_revision. in server_args, want to set speculative_draft_model_revision to default to 'main' when speculative_draft_model_path is set so that it won't default to None, which will lead it fall back to use server_args.revision in model runner.

python/sglang/srt/server_args.py
```diff
@@ -1261,6 +1261,12 @@ def _handle_hicache(self):
                 )
 
     def _handle_speculative_decoding(self):
+        if (
+            self.speculative_draft_model_path is not None
+            and self.speculative_draft_model_revision is None
+        ):
+            self.speculative_draft_model_revision = "main"
```
Qiaolin-Yu, line 1617: Why we need to set this by default?

gongwei-130: @Qiaolin-Yu sorry, lost track of this. 

If we don't set a default value for it, it would default to None and pass in [here](https://github.com/sgl-project/sglang/pull/11893/files#diff-65c6ac2c41977f68e460f18e35053b97089631f88a9958b0796343fccee78a67R528), then [from_server_args](https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/configs/model_config.py#L235) would use server_args.revision as draft model revision, which introduce problem.

gongwei-130:
@Qiaolin-Yu could you take a look?

Qiaolin-Yu (approved):

Qiaolin-Yu:
ping me when the ci pass

Qiaolin-Yu:
I think it's ready to merge. cc @Fridge003

PR #11900, opened 2025-10-21
Fix cache_tokens calculate issue when retracted

QiuMike:
When retracted happens, cache_tokens may greater than prompt tokens, just like:
```
\'meta_info\': {\'id\': \'21b1cabf17607522173946331ee5e9_d71b9220\',
\'finish_reason\': {\'type\': \'stop\', \'matched\': 156892},
\'prompt_tokens\': 59946, \'weight_version\': \'default\',
\'completion_tokens\': 7437, \'cached_tokens\': 122884,
\'e2e_latency\': 791.9391989707947,
\'ttft_latency\': 13.77153491973877,
\'queue_latency\': 0.0010406021028757095}}'
```
This patch fix this issue, Do not caculate the cached_tokens when retracted prefill.

## Checklist


huangtingwei9988 (approved):
LGTM

huangtingwei9988:
@hnyls2002 @xiezhq-hermann Can you review this pr? Bug we found online

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -543,6 +543,8 @@ def __init__(
 
         # For retraction
         self.is_retracted = False
+        # Indicates if the req has ever been retracted.
+        self.has_retracted = False
```
xiezhq-hermann: do we really need to introduce this? can't we just use `is_retracted`?

QiuMike: is_retraced will be set false when it is scheduled to prefill. But we really need a flag to indicate that this request has been marked as 'retracted', it needs always been true, because the req may be retracted for several times. The flag need be always true until the req finished. @xiezhq-hermann

QiuMike: @xiezhq-hermann any updates?

xiezhq-hermann: @hnyls2002 can you take a look here to see what might be the best way for this? `has_retracted` and `is_retracted` mixing is definitely confusing

QiuMike: @xiezhq-hermann @hnyls2002 another solution is:  retracted_status [None, retracting,  retracted]
one request will be None if retract never happened to it. After retract, set to retracted, it shows the request has been retracted before. if this OK, I will post another version

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -431,6 +431,12 @@ class RequestStage(str, enum.Enum):
     DECODE_QUICK_FINISH = "quick_finish"
 
 
+class RequestRetractStatus(enum.Enum):
```
hnyls2002: Do not add a new Enum for this... The status is wrong and inconsistent.

hnyls2002 (changes requested):

python/sglang/srt/managers/schedule_batch.py
QiuMike: already make a flag to indicate the request has ever been retracted in new version

hnyls2002 (approved):

hnyls2002:
@QiuMike please rebase your PR with main.

ShangmingCai (approved):
LGTM

PR #11685, opened 2025-10-15
[Lint] Add `python/sglang` to ruff F401 checks and remove unused imports in files

CatherineSue:
## Motivation

This PR adds F401 to ruff checks and removes unused imports accross `python/sglang`, excluding `__init__.py`.

## Modifications

- Added `F821` to ruff checks - this catches undefined names/missing imports
- Added `python/sglang` to ruff checks, excluded all __init__.py
- Tried to fix all unused imports

## Checklist

- [x] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.ai/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.ai/developer_guide/contribution_guide.html#benchmark-the-speed).

PR #11922, opened 2025-10-21
[lint] improve ruff check

hnyls2002:
Prev PR #11685 added `python/sglang` to the Ruff lint scope, but the pre-commit args were written as:

```
args: [--select=F401,F821, --fixable=F401]
```

These arguments are interpreted as `--select=F401 F821 --fixable=F401`, which leads to 

```
warning: Failed to lint F821: No such file or directory (os error 2)
```

This PR changes the arguments to
```
args:
  - --select=F401,F821
  - --fix
```

- Remove `--fixable` since `F401` is already a subset of the selected rules and `F821` is non-fixable.
- Enable auto-fix for `F401` (unused imports).
- Keep `F821` check enabled without triggering malformed argument parsing.

### DIFF
--- a/python/sglang/srt/disaggregation/common/conn.py
+++ b/python/sglang/srt/disaggregation/common/conn.py
@@ -5,7 +5,7 @@ import logging
 import socket
 import threading
 from functools import cache
-from typing import TYPE_CHECKING, Dict, List, Optional, Tuple, Union
+from typing import Dict, List, Optional, Tuple, Union
 
 import numpy as np
 import numpy.typing as npt
@@ -37,9 +37,6 @@ from sglang.srt.utils import (
     is_valid_ipv6_address,
     maybe_wrap_ipv6_address,
 )
-
-if TYPE_CHECKING:
-    from sglang.srt.disaggregation.mooncake.conn import TransferInfo
 
 logger = logging.getLogger(__name__)
 
@@ -80,7 +77,7 @@ class CommonKVManager(BaseKVManager):
 
         if self.disaggregation_mode == DisaggregationMode.PREFILL:
             self._register_to_bootstrap()
-            self.transfer_infos: Dict[int, Dict[str, TransferInfo]] = {}
+            self.transfer_infos = {}
             self.decode_kv_args_table = {}
             self.pp_group = get_pp_group()
         elif self.disaggregation_mode == DisaggregationMode.DECODE:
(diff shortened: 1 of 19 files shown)


CatherineSue (approved):

slin1237 (approved):

python/sglang/srt/disaggregation/common/conn.py
```diff
@@ -38,6 +38,9 @@
     maybe_wrap_ipv6_address,
 )
 
+if TYPE_CHECKING:
+    from sglang.srt.disaggregation.mooncake.conn import KVArgsRegisterInfo, TransferInfo
+
```
ShangmingCai: In fact, nixl and mooncake have implemented different versions of `KVArgsRegisterInfo` and `TransferInfo`.

hnyls2002: Ok, I will remove this type hint. If there are two versions, the type hint actually makes no sense...

hnyls2002: You can add it back when there is a unified top ABS class. @ShangmingCai

ShangmingCai: Maybe we can just change this
```python
            self.transfer_infos: Dict[int, Dict[str, TransferInfo]] = {}
            self.decode_kv_args_table: Dict[str, KVArgsRegisterInfo] = {}
```
to
```python
            self.transfer_infos = {}
            self.decode_kv_args_table = {}
```

hnyls2002: done

ShangmingCai (approved):
LGTM

hnyls2002:
This PR requires some code removal from @BBuf, as some should-be-deprecated code has too many undefined names.

BBuf:
> This PR requires some code removal from @BBuf, as some should-be-deprecated code has too many undefined names.

Done.

elvischenv:
Does anyone see this issue?
```
  File "/workspace/sglang/python/sglang/srt/layers/moe/fused_moe_triton/triton_kernels_moe.py", line 16, in <module>
    from triton_kernels.numerics import InFlexData, MicroscalingCtx
ImportError: cannot import name 'MicroscalingCtx' from 'triton_kernels.numerics' (/workspace/.venv/lib/python3.12/site-packages/triton_kernels/numerics.py)
```

FlamingoPg:
It looks like unit-test-backend-1-gpu(3) was broken cauz this PR imported a non-existent `MicroscalingCtx`. This symbol doesn’t exist in the Triton kernels, and looks never use. cc: @BBuf @hnyls2002

PR #11939, opened 2025-10-21
feat: add AutoWeightsLoader for streaming weight ingestion

alhridoy:
#11864
 - introduce
     `AutoWeightsLoader`/`WeightsMapper`
     utilities for streaming checkpoint
     ingestion
     - fall back to the auto loader from
     `DefaultModelLoader` and HF-backed models
      when custom `load_weights` isn’t present
     - add regression coverage for the loader
     (skip-safe when torch is unavailable)

     ## Testing
     - `python3 -m pytest
     test/srt/test_auto_weights_loader.py


hnyls2002:
Thanks! This looks like it overlaps with #12534 for the same issue (#11864), so I'm closing it to consolidate the discussion there. Please reopen if you think this version is the one we should continue with.

Issue #11742, opened 2025-10-17
[Bug] Limit TCP Port Range


jinmingyi1998:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Describe the bug

For security reasons, it is necessary to restrict TCP port numbers within a specific range. The `reserve_port` and `release_port` functions in `sglang/utils.py` can achieve this, but it appears that some port-related code does not utilize this method.

### Reproduction

lanch server and see netstat

### Environment

v0.5.3.post1 image

jimmy-evo:
I'm working on this. I grep all `bind()` method, but only find in kv manager.

I fix kv manager's bind but there still something listen on random port.
(I check with `lsof` and see process is `sglang`

PR #11961, opened 2025-10-22
fix(tcp-port): replace bind_server_socket to get_zmq_socket(Port conflict)

jimmy-evo:
## Motivation

there are some duplicate functions to reserve tcp port. 
remove duplicate usage.
use `reserve_port`.
avoid **multi-process** get the same **random ports**

## Modifications

there already has a zmq socket util `get_zmq_socket`.

remove `_bind_server_socket` in `conn.py`

bind socket in kvmanager's `__init__`

## Checklist


jimmy-evo:
#11742

ShangmingCai:
reserve port range doesn't seem like an ideal solution for me, as it is a little bit complicated/hacky and might cause other problems when people are using machines that have other usage for those ports.

wcsjtu:
The atomicity of the context between `reserve_port` and `release_port` is not guaranteed

jimmy-evo:
> reserve port range doesn't seem like an ideal solution for me, as it is a little bit complicated/hacky and might cause other problems when people are using machines that have other usage for those ports.

@ShangmingCai 

firstly, random port without range limit cause bigger problem. usually sglang server is running on host network, random port will make trouble for K8s cluster and network firewall. that's why i commit this pr.
then i found multi-process will get the same random port. so i fix the random logic also.

jimmy-evo:
> The atomicity of the context between `reserve_port` and `release_port` is not guaranteed

@wcsjtu 
no need to guarantee atomicity, port is allocated at bind().
and the random guarantee different process have different ports

ShangmingCai:
> > reserve port range doesn't seem like an ideal solution for me, as it is a little bit complicated/hacky and might cause other problems when people are using machines that have other usage for those ports.
> 
> @ShangmingCai
> 
> firstly, random port without range limit cause bigger problem. usually sglang server is running on host network, random port will make trouble for K8s cluster and network firewall. that's why i commit this pr. then i found multi-process will get the same random port. so i fix the random logic also.

@jinmingyi1998 Sound reasonable. We can sync on the range limit tomorrow. I just don't agree with the current impl(30000-40000), but the idea should be valid.

jimmy-evo:
> > > reserve port range doesn't seem like an ideal solution for me, as it is a little bit complicated/hacky and might cause other problems when people are using machines that have other usage for those ports.
> > 
> > 
> > @ShangmingCai
> > firstly, random port without range limit cause bigger problem. usually sglang server is running on host network, random port will make trouble for K8s cluster and network firewall. that's why i commit this pr. then i found multi-process will get the same random port. so i fix the random logic also.
> 
> @jinmingyi1998 Sound reasonable. We can sync on the range limit tomorrow. I just don't agree with the current impl(30000-40000), but the idea should be valid.

@ShangmingCai 

`reserve_port` and `release_port` is copy from `sglang/utils.py`. 
At the beginning, sglang is designed to use port in this range. But the initial design was ignored by later developers, and reimplemented the `get_free_port` function, which is duplicated

jimmy-evo:
original impl to allocate port from a range
https://github.com/sgl-project/sglang/blob/64cf868eba58a3d7a7c249ad3281b45523fe9e11/python/sglang/utils.py#L376-L405

jimmy-evo:
>  I just don't agree with the current impl(30000-40000), but the idea should be valid.

yes.
in my first commit the range can be config by env. i will add it later

wcsjtu:
> > reserve port range doesn't seem like an ideal solution for me, as it is a little bit complicated/hacky and might cause other problems when people are using machines that have other usage for those ports.
> 
> @ShangmingCai
> 
> firstly, random port without range limit cause bigger problem. usually sglang server is running on host network, random port will make trouble for K8s cluster and network firewall. that's why i commit this pr. then i found multi-process will get the same random port. so i fix the random logic also.

zmq.socket has a native method `bind_to_random_port`, its prototype is shown below. It will work fine in your situation

```py
def bind_to_random_port(
        self: _SocketType,
        addr: str,
        min_port: int = 49152,
        max_port: int = 65536,
        max_tries: int = 100,
    ) -> int:
        """Bind this socket to a random port in a range.

        If the port range is unspecified, the system will choose the port.

        Parameters
        ----------
        addr : str
            The address string without the port to pass to ``Socket.bind()``.
        min_port : int, optional
            The minimum port in the range of ports to try (inclusive).
        max_port : int, optional
            The maximum port in the range of ports to try (exclusive).
        max_tries : int, optional
            The maximum number of bind attempts to make.
```

ShangmingCai:
> > > reserve port range doesn't seem like an ideal solution for me, as it is a little bit complicated/hacky and might cause other problems when people are using machines that have other usage for those ports.
> > 
> > 
> > @ShangmingCai
> > firstly, random port without range limit cause bigger problem. usually sglang server is running on host network, random port will make trouble for K8s cluster and network firewall. that's why i commit this pr. then i found multi-process will get the same random port. so i fix the random logic also.
> 
> zmq.socket has a native method `bind_to_random_port`, its prototype is shown below. It will work fine in your situation
> 
> ```python
> def bind_to_random_port(
>         self: _SocketType,
>         addr: str,
>         min_port: int = 49152,
>         max_port: int = 65536,
>         max_tries: int = 100,
>     ) -> int:
>         """Bind this socket to a random port in a range.
> 
>         If the port range is unspecified, the system will choose the port.
> 
>         Parameters
>         ----------
>         addr : str
>             The address string without the port to pass to ``Socket.bind()``.
>         min_port : int, optional
>             The minimum port in the range of ports to try (inclusive).
>         max_port : int, optional
>             The maximum port in the range of ports to try (exclusive).
>         max_tries : int, optional
>             The maximum number of bind attempts to make.
> ```

This `bind_to_random_port` looks good to me, which has a `max_tries` mechanism that could be the key to addressing concurrent port conflict.

python/sglang/srt/utils/common.py
```diff
@@ -1336,12 +1388,17 @@ def get_zmq_socket(
...
     if endpoint is None:
         # Bind to random TCP port
         config_socket(socket, socket_type)
-        port = socket.bind_to_random_port("tcp://*")
+        addr = address or "tcp://*"
+        port = socket.bind_to_random_port(
+            addr,
+            min_port=envs.SGLANG_DEFAULT_TCP_PORT_RANGE_START.get(),
+            max_port=envs.SGLANG_DEFAULT_TCP_PORT_RANGE_END.get(),
+        )
```
ShangmingCai: Seems like `get_zmq_socket` and `bind_to_random_port` are all we need? Maybe we don't need `get_free_port` anymore when ascend transfer engine doesn't need it as well.

jimmy-evo: @ShangmingCai  I prefer to reuse and enrich existing functions, as there is already a lot of duplicate code.

ShangmingCai: Reducing duplication is important to us, we take many time in H2 to do this.

python/sglang/srt/entrypoints/http_server.py
```diff
@@ -509,6 +509,7 @@ async def get_server_info():
     return {
         **dataclasses.asdict(_global_state.tokenizer_manager.server_args),
         **_global_state.scheduler_info,
+        "port_args": dataclasses.asdict(_global_state.tokenizer_manager.port_args),
```
ShangmingCai: Why is this needed?

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -152,6 +152,7 @@ def __init__(
     ):
         # Parse args
         self.server_args = server_args
+        self.port_args = port_args
```

python/sglang/srt/entrypoints/http_server.py
jimmy-evo: > Why is this needed?

as part of server info. used to debug

ShangmingCai: We already have some debug info, since port conflict is a small and easy-to-identify issue, I think the changes should be as clean as possible.

jimmy-evo: removed

python/sglang/srt/managers/tokenizer_manager.py
jimmy-evo: removed

python/sglang/srt/disaggregation/kv_events.py
```diff
@@ -238,6 +238,7 @@ def _socket_setup(self) -> None:
                 or self._endpoint.startswith("ipc://")
                 or self._endpoint.startswith("inproc://")
             ):
+                logger.debug(f"publisher_endpoint={self._endpoint}")
```
ShangmingCai: Either remove it or complete the debug message to make sure everyone can understand which module produces this debug info.
```suggestion
```

python/sglang/srt/disaggregation/kv_events.py
```diff
@@ -248,6 +249,7 @@ def _socket_setup(self) -> None:
         # 3) works in our non‑blocking poll loop alongside PUB
         if self._replay_endpoint is not None:
             self._replay = self._ctx.socket(zmq.ROUTER)
+            logger.debug(f"replay_endpoint={self._replay_endpoint}")
```
ShangmingCai: ditto
```suggestion
```

python/sglang/srt/distributed/device_communicators/shm_broadcast.py
```diff
@@ -232,6 +231,9 @@ def __init__(
             remote_subscribe_port = get_open_port()
             if is_valid_ipv6_address(connect_ip):
                 self.remote_socket.setsockopt(IPV6, 1)
+            logger.debug(
+                f"Binding remote socket to {format_tcp_address(connect_ip, remote_subscribe_port)}"
+            )
```
ShangmingCai:
```suggestion
```
ditto

ShangmingCai (approved):
LGTM

python/sglang/srt/disaggregation/common/conn.py
```diff
@@ -68,11 +68,16 @@ def __init__(
...
-        self.server_socket = zmq.Context().socket(zmq.PULL)
-        if is_valid_ipv6_address(self.local_ip):
-            self.server_socket.setsockopt(zmq.IPV6, 1)
+
+        # bind zmq socket
+        context = zmq.Context()
+        address = f"tcp://{maybe_wrap_ipv6_address(self.local_ip)}"
+        self.rank_port, self.server_socket = get_zmq_socket(
+            context, zmq.PULL, bind=True, address=address
+        )
```
ShangmingCai:
```bash
TypeError: get_zmq_socket() got an unexpected keyword argument 'address'
```
Use `endpoint`, not `address`.

jimmy-evo: > ```shell
> TypeError: get_zmq_socket() got an unexpected keyword argument 'address'
> ```
> 
> Use `endpoint`, not `address`.

mistake when remove args`address` from `get_zmq_socket`

hnyls2002 (approved):

PR #11966, opened 2025-10-22
[HICache / PD]: Support offloading incremental KV cache in decode side.

hzh0425:
## Motivation

In the previous implementation of Decode Offloading, the system would offload the entire KVCache for each request during decoding, resulting in very long offloading times.

This PR introduces incremental offloading for Decode:

1. The P side offloads the prefill portion:
   ```python
   prefill_offloaded_len = (
       len(req.origin_input_ids) // self.page_size * self.page_size
   )
   ```

2. Consequently, the D side only offloads the incremental part compared to P:
   ```python
   incremental_len = len(all_tokens) - prefill_offloaded_len
   incremental_aligned_len = incremental_len // self.page_size * self.page_size
   ```

This approach significantly reduces the offloading time by only offloading the incremental changes rather than the entire KVCache for each request.

## Benchmarking and Profiling

Here is the effect of the optimization: with 20K input and 500 output, the offloading time can be reduced by nearly 1 second.

### After Optimization:
- **Decode Side Offload Time:**
  ```
  [2025-10-22 08:09:31 TP0] Finished backup request xx, free host memory, len:448, cost time:0.04 seconds.
  ```

### Before Optimization:
- **Decode Side Offload Time:**
  ```
  [2025-10-22 08:15:53 TP0] Finished backup request xx, free host memory, len:20928, cost time:1.14 seconds.
  ``` 

This demonstrates that the optimization significantly reduces the offloading time from 1.14 seconds to 0.04 seconds for the given scenario.
## Checklist


python/sglang/srt/disaggregation/decode_kvcache_offload_manager.py
```diff
@@ -77,41 +78,59 @@ def __init__(
...
+        start, end = (
+            prefill_offloaded_len,
+            prefill_offloaded_len + incremental_aligned_len,
+        )
+        incremental_tokens = all_tokens[start:end]
+        incremental_indices = token_indices[start:end]
+
+        # Early free prefill-offloaded GPU memory
+        if prefill_offloaded_len > 0:
+            self.token_to_kv_pool_allocator.free(token_indices[:prefill_offloaded_len])
```
hzh0425, line 114: Here, we need to release the cache for the Prefill portion and retain the incremental part for Decode. Should this be encapsulated within the `tree_cache/chunk_cache`?  cc @xiezhq-hermann

xiezhq-hermann: it should probably fine to keep it here, since we don't really use a tree to track the KV sharing across requests for now

xiezhq-hermann (approved):

python/sglang/srt/disaggregation/decode_kvcache_offload_manager.py
```diff
@@ -77,41 +78,59 @@ def __init__(
...
-        tokens = req.origin_input_ids + req.output_ids
-        aligned_len = (len(tokens) // self.page_size) * self.page_size
-        if aligned_len == 0:
+        # Prefill side offloads page-aligned origin_input_ids, decode side offloads the incremental part
+        all_tokens = req.origin_input_ids + req.output_ids[:-1]
+        prefill_offloaded_len = (
+            len(req.origin_input_ids) // self.page_size * self.page_size
+        )
+        incremental_len = len(all_tokens) - prefill_offloaded_len
+        incremental_aligned_len = incremental_len // self.page_size * self.page_size
```
ShangmingCai, line 99: I know that we didn't store that last unfilled page at the prefill side, but not sure about `req.output_ids[:-1]`, why we discard the last token at the decode side?

ShangmingCai (approved):
LGTM

xiezhq-hermann:
@ByronHsu @hnyls2002 can you help take a look?

ByronHsu (approved):

PR #11982, opened 2025-10-22
Add an entrypoint for `sglang.launch_server`

leoleoasd:
## Motivation

Currently, sglang server can only be launched with `python -m sglang.launch_server`, which prevents tools like `uv tool`, `pipx` and `condax` from installing sglang.

## Modifications

Now sglang will produce a binary executable file `sglang-serve` that will invoke sglang-server. Users is then able to install sglang to a convenient isolated environment with `uv tool install sglang[all]` and use `sglang-serve` anywhere.

Need to figure out a way 

## Checklist

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -1,18 +1,18 @@
 [build-system]
-requires = ["setuptools>=61.0", "wheel"]
 build-backend = "setuptools.build_meta"
+requires = ["setuptools>=61.0", "wheel"]
 
 [project]
-name = "sglang"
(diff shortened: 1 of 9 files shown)


leoleoasd:
The key difference in pyproject.toml is adding this, the rest are formatting change:
```
[project.scripts]
sglang-serve = "sglang.launch_server:main"
```

b8zhong:
Hi, now we have sglang serve directly. Thanks a lot.

PR #11985, opened 2025-10-22
Fix incorrect KV indices creation when page_size=32 in TRTLLM MLA backend

cicirori:
### Problem
`create_flashmla_kv_indices_triton` was computing iteration count from tokens instead of pages, causing KV indices to be incompletely processed when `page_size=32`.

Root cause: The loop count was calculated using a fixed `BLOCK_SIZE` (4096 tokens):
```python
num_pages_loop = tl.cdiv(kv_end - kv_start, BLOCK_SIZE)  # Wrong: uses tokens
```

This works only for `page_size=64` (where 64 pages × 64 tokens/page = 4096 tokens) but fails for `page_size=32` (64 pages × 32 tokens/page = 2048 tokens ≠ 4096).

### Solution
1. Make `NUM_PAGE_PER_BLOCK` dynamic based on `page_size`:
   - Added `get_num_page_per_block_flashmla(page_size)` that returns `4096 // page_size`
   - Preserves `4096` tokens per iteration
2. Update kernel to compute `NUM_PAGE_PER_BLOCK` dynamically:
   ```python
   NUM_PAGE_PER_BLOCK: tl.constexpr = (FLASHMLA_CREATE_KV_BLOCK_SIZE_TRITON // PAGED_SIZE)
   ```
3. Propagate changes to `_calc_padded_blocks` to use the dynamic function

### Changes
- `python/sglang/srt/layers/attention/utils.py`: Make `NUM_PAGE_PER_BLOCK` dynamic
- `python/sglang/srt/layers/attention/trtllm_mla_backend.py`: Remove hardcoded `NUM_PAGE_PER_BLOCK` parameter
- `python/sglang/test/attention/test_trtllm_mla_backend.py`: Add test setup and validation

### Testing
- Page size 32 now creates correct KV indices
- Page size 64 behavior unchanged
- Tests verify all required blocks are processed

### Impact
- Fixes incorrect KV cache behavior with `page_size=32`
- No performance regression; maintains 4096 tokens per iteration
- No breaking changes


cicirori:
it seems like all tests in `python/sglang/test/` would not be triggered by ci?

zhyncs (approved):

PR #11991, opened 2025-10-23
[NVIDIA] Change default quant method for model_opt

kaixih:
## Motivation

While working on Qwen + Eagle, I found it’s not compatible with NVIDIA’s released Eagle heads (nvidia/Qwen3-235B-A22B-Eagle3). The root cause is that we default quant_method to fp8 for ModelOpt checkpoints, but this assumption doesn’t hold, for example, [this model](https://huggingface.co/nvidia/Qwen3-235B-A22B-Eagle3/blob/main/hf_quant_config.json) doesn’t specify any quantization method.

## Modifications

This PR updates the logic to set the default quant_method to None when it is not explicitly defined in hf_quant_config.json.

## Accuracy Tests

```bash
    --model Qwen/Qwen3-235B-A22B \
    --speculative-algorithm EAGLE3 \
    --speculative-draft-model-path nvidia/Qwen3-235B-A22B-Eagle3 \
    --speculative-num-steps 5 \
    --speculative-eagle-topk 8 \
    --speculative-num-draft-tokens 32 \
    --mem-fraction-static 0.75 \
    --tp 8 \
    --ep 8 \
    --context-length 8192 \
    --trust-remote-code \
    --host 0.0.0.0 \
    --port 30000 \
    --dtype bfloat16
```

```bash
+ python benchmark/gsm8k/bench_sglang.py --port 30000 --num-questions 10000 --parallel 10000 --num-shots 8
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1319/1319 [03:24<00:00,  6.44it/s]
Accuracy: 0.951
Invalid: 0.000
Latency: 204.845 s
```

## Benchmarking and Profiling

NA

## Checklist


kaixih:
I notice the original code mentions `Default to FP8 for backward compatibility` which I guess there might be some model checkpoints released by modelopt uses fp8 quantization but doesn't specify it in the `quant_algo` field. If that's the case, we should workaround that explicitly. @Fridge003 can you help confirm? If we know the model name, I can help add some special rules for that.

Edwardf0t1:
@kaixih Could you please check with your changes, if we can still run a modelopt fp8 ckpt with `--quantization modelopt`? For example,
```
  --model-path nvidia/Llama-3.1-8B-Instruct-FP8 \
  --trust-remote-code \
  --quantization modelopt
```

kaixih:
@Edwardf0t1 , it’s fixed now. I tried your command and it works. Using `--quantization modelopt_fp8` also works.

In short, for ModelOpt FP8 checkpoints, users can pass either `modelopt` or `modelopt_fp8`. If `modelopt` is used, the system will read `hf_quant_config.json` (looking for `"quant_algo": "FP8"`) and automatically override it to `modelopt_fp8`.

For non-quantized ModelOpt checkpoints, users should just use `--quantization modelopt`.

Edwardf0t1:
> @Edwardf0t1 , it’s fixed now. I tried your command and it works. Using `--quantization modelopt_fp8` also works.
> 
> In short, for ModelOpt FP8 checkpoints, users can pass either `modelopt` or `modelopt_fp8`. If `modelopt` is used, the system will read `hf_quant_config.json` (looking for `"quant_algo": "FP8"`) and automatically override it to `modelopt_fp8`.
> 
> For non-quantized ModelOpt checkpoints, users should just use `--quantization modelopt`.

Thanks for the test. By "non-quantized ModelOpt checkpoints", do you mean the original model? Users can specify `modelopt_fp8` or `modelopt_fp4` to quantize the original BF16 model.

python/sglang/srt/layers/quantization/modelopt_quant.py
```diff
@@ -228,6 +228,15 @@ def from_config(cls, config: Dict[str, Any]) -> ModelOptFp8Config:
             packed_modules_mapping=config.get("packed_modules_mapping"),
         )
 
+    @classmethod
+    def override_quantization_method(cls, hf_quant_cfg, user_quant) -> Optional[str]:
+        if (
+            user_quant == "modelopt"
+            and hf_quant_cfg.get("quant_method", "") == "modelopt_fp8"
```
Edwardf0t1: We already have an `override_quantization_method` function in L158.

kaixih: Fixed. PTAL.

kaixih:
I mean these modelopt released eagle heads: https://huggingface.co/nvidia/Qwen3-235B-A22B-Eagle3/blob/main/hf_quant_config.json. They are bf16 with no quantization.

@Edwardf0t1

python/sglang/srt/layers/quantization/base_config.py
```diff
@@ -179,6 +179,11 @@ def _modelopt_override_quantization_method(
             elif "NVFP4" in quant_algo or "FP4" in quant_algo:
                 return "modelopt_fp4"
 
+        # The hf_quant_config may be a parsed quant config, so we need to check the
+        # quant_method.
+        if hf_quant_config.get("quant_method", "") == "modelopt_fp8":
+            return "modelopt_fp8"
```
Edwardf0t1, line 185:
```suggestion
        if hf_quant_config.get("quant_method", "") == "modelopt_fp8":
            return "modelopt_fp8"
        elif hf_quant_config.get("quant_method", "") == "modelopt_fp4":
            return "modelopt_fp4"
```

Edwardf0t1 (approved):
LGTM, left a suggestion.

python/sglang/srt/layers/quantization/base_config.py
kaixih: Done. Thx.

PR #11994, opened 2025-10-23
Use explicit uint64 dtype for Tensor data_ptr() to avoid overflow

jianan-gu:
Tensor data_ptr() API returns a virtual address which should be uint64 value, this PR adds explicit dtype (torch.uint64) to avoid [overflow errors](https://github.com/intel/torch-xpu-ops/issues/2102#issuecomment-3331158518)  on different devices.


mingfeima:
@jianan-gu list APIs from pytorch

mingfeima (approved):

Fridge003 (approved):

PR #11866, opened 2025-10-20
Support nvidia/NVIDIA-Nemotron-Nano-9B-v2-FP8/NVFP4

netanel-haber:
## Motivation

Support [nvidia/NVIDIA-Nemotron-Nano-9B-v2-FP8](https://huggingface.co/nvidia/NVIDIA-Nemotron-Nano-9B-v2-FP8) and [nvidia/NVIDIA-Nemotron-Nano-9B-v2-NVFP4](https://huggingface.co/nvidia/NVIDIA-Nemotron-Nano-9B-v2-NVFP4) variants of https://huggingface.co/nvidia/NVIDIA-Nemotron-Nano-9B-v2.

Following: https://github.com/sgl-project/sglang/pull/10909

## Additional Modifications
* Support `packed_modules_mapping` in modelopt quant_config.
* `_get_quantization_config` now supports `remap_prefix`, similar to VLLM.
* Reintroduce mamba unittests into CI, with reduced matrix for faster testing.

<details>
<summary>GSM8K (basically equivalent)</summary>

|              | SGLANG          | VLLM               |
|:-----------|:--------------------:|:-------------------:|
| FP8       | 0.881 ± 0.008   | 0.892 ± 0.008  |
| NVFP4  | 0.899 ± 0.008   | 0.890 ± 0.008  |

```shell
python3 -m sglang.launch_server --model-path nvidia/NVIDIA-Nemotron-Nano-9B-v2-FP8--max-mamba-cache-size 256

vllm serve nvidia/NVIDIA-Nemotron-Nano-9B-v2-FP8 --trust-remote-code --max-num-seqs 512 --mamba_ssm_cache_dtype float32

python -m lm_eval --model local-chat-completions \                                                                      
  --model_args "model=nvidia/NVIDIA-Nemotron-Nano-9B-v2-FP8,base_url=http://localhost:8000/v1/chat/completions,api_key=EMPTY,num_concurrent=256,timeout=3600" \
  --tasks gsm8k \
  --apply_chat_template --system_instruction "/no_think" \
  --gen_kwargs temperature=0.6,top_p=0.95,do_sample=true,seed=1 --output_path ./results/$(date +%s) \
  --log_samples
```

</details>

## Checklist

- [X] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
[...]

PR #12011, opened 2025-10-23
`Moe=ModelOptNvFp4FusedMoEMethod` for flashinfer as well, like before

netanel-haber:
## Motivation

Fix #11866 as per: https://github.com/sgl-project/sglang/pull/11866#discussion_r2445211502

## Modifications

`Moe=ModelOptNvFp4FusedMoEMethod` for flashinfer as well, like before.


FlamingoPg (approved):
LGTM, we need a method to return

PR #12012, opened 2025-10-23
feat: add EP support in tuning

Chen-0210:
## Motivation

Some models enable EP parallelism, which changes the number of experts (E) during tuning, so this update adds corresponding support in tuning.

## Checklist


benchmark/kernels/fused_moe_triton/tuning_fused_moe_triton.py
```diff
@@ -413,71 +413,76 @@ def save_configs(
 def main(args: argparse.Namespace):
     print(args)
 
+    assert args.ep_size in [
+                1,
+                args.tp_size,
+            ], "The expert parallel size must be 1 or the same as the tensor parallel size"
```
ch-wan: why do we need this assertion

Chen-0210: https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/server_args.py#L3288C1-L3292C10
to align with server_args.py, since --enable-ep-moe is deprecated, the value of ep-size must be the same as tp-size if we enable ep.

Chen-0210:
@ch-wan The implementation is ready, could you take a review?

benchmark/kernels/fused_moe_triton/tuning_fused_moe_triton.py
ch-wan: An arbitrary `ep_size` is supported in sglang. The link you shared does not direct me to an informative code script.

Chen-0210: The correct link should be:
https://github.com/sgl-project/sglang/blob/v0.5.4/python/sglang/srt/server_args.py#L3277C1-L3281C10
,which caused some confusion to me earlier. Could you please confirm if it’s correct to have the following assertion,or if no assertion is needed here?

```
assert (
    self.tp_size % self.ep_size == 0
), f"The expert parallel size (ep_size={self.ep_size}) must divide the tensor parallel size (tp_size={self.tp_size}) evenly"
```

ch-wan: This is a deprecated command. `ep_size` < `tp_size` was later supported in https://github.com/sgl-project/sglang/pull/8590.

Chen-0210: fixed

benchmark/kernels/fused_moe_triton/tuning_fused_moe_triton.py
```diff
@@ -415,69 +415,87 @@ def main(args: argparse.Namespace):
...
+        shard_intermediate_size = (
+            2 * intermediate_size // (args.tp_size // args.ep_size)
+        )
     elif config.architectures[0] in ["DeepseekV2ForCausalLM", "DeepseekV3ForCausalLM"]:
         E = (
             config.n_routed_experts + (0 if args.disable_shared_experts_fusion else 1)
             if config.architectures[0] in ["DeepseekV3ForCausalLM"]
             else config.n_routed_experts
-        )
+        ) // args.ep_size
```
ch-wan: it should be applied to `n_routed_experts` before considering shared experts.

Chen-0210: Good catch. fixed

BBuf (approved):

Paiiiiiiiiiiiiii:
good job！

PR #12025, opened 2025-10-23
[Fix]: HiCache hasher failed when EAGLE mode enabled

leavelet:
## Motivation

Fix #11919. This PR addresses two separate issues:

1. **EAGLE mode compatibility**: An `AttributeError` occurs when using EAGLE mode with hierarchical cache storage backend (HiCache). In EAGLE mode, token IDs are represented as bigram tuples `(int, int)` instead of single integers. The `get_hash_str` function in `hicache_storage.py` assumed token IDs are always integers and called `.to_bytes()` directly, causing it to fail on tuples.

2. **Incorrect OrderedDict import**: `mini_3fs_metadata_server.py` incorrectly imported `OrderedDict` from the `typing` module instead of `collections`. This caused it to fall back to a regular `dict` that lacks the `move_to_end` method, leading to runtime errors.

## Modifications

1. **`hicache_storage.py`**: Modified `get_hash_str` function to handle both single integer tokens and tuple tokens (EAGLE bigram mode). When a tuple is detected, the function iterates through tuple elements and hashes each integer separately.

2. **`mini_3fs_metadata_server.py`**: Fixed the import statement from `from typing import OrderedDict` to `from collections import OrderedDict` to use the actual `OrderedDict` implementation rather than the type hint.
## Checklist

### DIFF
--- a/python/sglang/srt/mem_cache/hicache_storage.py
+++ b/python/sglang/srt/mem_cache/hicache_storage.py
@@ -19,7 +19,13 @@ def get_hash_str(token_ids: List[int], prior_hash: str = None) -> str:
         hasher.update(bytes.fromhex(prior_hash))
 
     for t in token_ids:
-        hasher.update(t.to_bytes(4, byteorder="little", signed=False))
+        if isinstance(t, tuple):
+            # EAGLE bigram mode: hash both elements to uniquely identify the bigram
+            for elem in t:
+                hasher.update(elem.to_bytes(4, byteorder="little", signed=False))
+        else:
+            # Regular mode: single integer token
+            hasher.update(t.to_bytes(4, byteorder="little", signed=False))
 
     return hasher.hexdigest()
 
--- a/python/sglang/srt/mem_cache/storage/hf3fs/mini_3fs_metadata_server.py
+++ b/python/sglang/srt/mem_cache/storage/hf3fs/mini_3fs_metadata_server.py
@@ -3,8 +3,9 @@ import atexit
 import json
 import logging
 import threading
+from collections import OrderedDict
 from pathlib import Path
-from typing import Dict, List, Optional, OrderedDict, Tuple
+from typing import Dict, List, Optional, Tuple
 
 import orjson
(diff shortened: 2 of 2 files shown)


leavelet:
@xiezhq-hermann please review

python/sglang/srt/mem_cache/hicache_storage.py
```diff
@@ -19,7 +19,13 @@ def get_hash_str(token_ids: List[int], prior_hash: str = None) -> str:
         hasher.update(bytes.fromhex(prior_hash))
 
     for t in token_ids:
-        hasher.update(t.to_bytes(4, byteorder="little", signed=False))
+        if isinstance(t, tuple):
+            # EAGLE bigram mode: hash both elements to uniquely identify the bigram
```
xiezhq-hermann, line 23: @ispobock is hicache enabled on the eagle worker? I thought for eager worker we only initiated a basic radix tree, and on regular worker we have hiradix tree and the keys would not be bigram. Was my understanding correct?

xiezhq-hermann: even eagle worker has a hicache tree instance, we should not have real hicache-related activities for that

ispobock: There is only one Radix tree in the target worker, eagle worker don't have radix tree but uses the matched indices from the target Radix tree. The target radix tree uses the bigram key.

python/sglang/srt/mem_cache/hicache_storage.py
```diff
@@ -19,7 +19,13 @@ def get_hash_str(token_ids: List[int], prior_hash: str = None) -> str:
         hasher.update(bytes.fromhex(prior_hash))
 
     for t in token_ids:
-        hasher.update(t.to_bytes(4, byteorder="little", signed=False))
+        if isinstance(t, tuple):
```
ispobock, line 22: Can we add unit test to cover this case?

PR #8286, opened 2025-07-23
fix: prevent HuggingFace access when SGLANG_USE_MODELSCOPE is enabled

yankay:
## Motivation

When SGLANG_USE_MODELSCOPE=true is set, the system was still making calls to HuggingFace endpoints during model configuration loading, which could lead to:
- Network access issues in environments where HuggingFace is blocked

Fix: #8275

## Modifications

Referenced Vllm: https://github.com/vllm-project/vllm/pull/5233
This PR fixes the issue by adding conditional import logic for AutoConfig and HfApi.

The problem can be reliably reproduced by setting `HF_ENDPOINT: https://127.0.0.1`, so a test case has been modified to ensure no HuggingFace access occurs when `SGLANG_USE_MODELSCOPE` is enabled.

## Checklist

- [x] Format your code according to the [Code Formatting with Pre-Commit](https://docs.sglang.ai/references/contribution_guide.html#code-formatting-with-pre-commit).
- [ ] Add unit tests as outlined in the [Running Unit Tests](https://docs.sglang.ai/references/contribution_guide.html#running-unit-tests-adding-to-ci).
- [ ] Update documentation / docstrings / example tutorials as needed, according to [Writing Documentation](https://docs.sglang.ai/references/contribution_guide.html#writing-documentation-running-docs-ci).
- [ ] Provide throughput / latency benchmark results and accuracy evaluation results as needed, according to [Benchmark and Profiling](https://docs.sglang.ai/references/benchmark_and_profiling.html) and [Accuracy Results](https://docs.sglang.ai/references/accuracy_evaluation.html).
- [ ] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co-author when merging the PR.
- [ ] Please feel free to join our Slack channel at https://slack.sglang.ai to discuss your PR.

PR #12039, opened 2025-10-24
fix: prevent HugginqFace access when SGLANG_USE_MODELSCOPE is enabled

yrk111222:
## Motivation
When SGLANG_USE_MODELSCOPE=true is set, the system was still making calls to HuggingFace endpoints during model configuration loading, which could lead to:

- Network access issues in environments where HuggingFace is blocked

## Problem Description

When using the following commands:

```bash
pip install sglang[all] -U
export SGLANG_USE_MODELSCOPE=True
python -m sglang.launch_server --model-path qwen/Qwen2-7B-Instruct --port 3000
```

Even after setting the ModelScope environment variable, the system still accesses the HuggingFace link, as shown in the screenshot below.
image

Fix: https://github.com/sgl-project/sglang/issues/8275
Related: https://github.com/sgl-project/sglang/pull/8286

## Checklist

### DIFF
--- a/python/sglang/srt/configs/model_config.py
+++ b/python/sglang/srt/configs/model_config.py
@@ -25,7 +25,7 @@ from transformers import PretrainedConfig
 from sglang.srt.environ import envs
 from sglang.srt.layers.quantization import QUANTIZATION_METHODS
 from sglang.srt.server_args import ServerArgs
-from sglang.srt.utils import is_hip, retry
+from sglang.srt.utils import get_bool_env_var, is_hip, retry
 from sglang.srt.utils.hf_transformers_utils import (
     get_config,
     get_context_length,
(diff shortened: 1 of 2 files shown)


ShangmingCai (approved):
LGTM. cc: @hnyls2002

python/sglang/srt/configs/model_config.py
```diff
@@ -509,12 +509,17 @@ def _parse_quant_hf_config(self):
             # example: https://huggingface.co/Barrrrry/DeepSeek-R1-W4AFP8/tree/main
             is_local = os.path.exists(self.model_path)
             if not is_local:
-                import huggingface_hub
+                # Conditional import based on SGLANG_USE_MODELSCOPE environment variable
+                if get_bool_env_var("SGLANG_USE_MODELSCOPE"):
```
ShangmingCai: Can we replace this with `envs.SGLANG_USE_MODELSCOPE` through https://github.com/sgl-project/sglang/blob/654a78f9c8580a7fe23870a7cde0fd1c9e04f150/python/sglang/srt/environ.py#L142? We should align this env var usage.

yrk111222: done, cc

ShangmingCai (approved):
LGTM

python/sglang/srt/configs/model_config.py
yrk111222: Hi @ShangmingCai ,

Sorry for the hasty submission yesterday. After further testing, I found that the check for `SGLANG_USE_MODELSCOPE` should use `.get() `to work correctly.

I've updated the PR to use `if envs.SGLANG_USE_MODELSCOPE.get():` and have also implemented the corresponding logic to download` hf_quant_config.json `from ModelScope.

Could you please review it again? Thanks!

PR #12044, opened 2025-10-24
Enable mixed type LayerNorm kernel for NSA indexer

akhilg-nv:
## Motivation

Currently we cast input to layernorm to fp32 and use native Torch layernorm, then cast back. Instead we pull in a more efficient TRT-LLM kernel (via flashinfer) that supports mixed precision inputs.

## Modifications

Add flashinfer layernorm kernel and apply in NSA indexer.

## Accuracy Tests

```
python -m sglang.launch_server --model-path model_fp4/ --tp 4 --dp 4 --enable-dp-attention --reasoning-parser deepseek-v3

python3 -m sglang.test.run_eval --port 30000 --eval-name gpqa --num-examples 198 --max-tokens 120000 --repeat 8 --thinking-mode deepseek-v3
```

sample results
```
# with old layernorm
Repeat: 8, mean: 0.783
Scores: ['0.788', '0.793', '0.808', '0.783', '0.742', '0.788', '0.783', '0.778']
====================/198 [22:35<00:52,  4.04s/it]
Writing report to /tmp/gpqa_model_fp4_.html
{'chars': np.float64(1464.3333333333333), 'chars:std': np.float64(356.735535690296), 'score:std': np.float64(0.41573970964154905), 'score': np.float64(0.7777777777777778)}
Writing results to /tmp/gpqa_model_fp4_.json
Total latency: 1422.309 s
Score: 0.778

# with new layernorm 
Repeat: 8, mean: 0.785
Scores: ['0.758', '0.763', '0.788', '0.753', '0.823', '0.808', '0.793', '0.793']
====================198 [22:02<2:29:43, 51.63s/it]
Writing report to /tmp/gpqa_model_fp4_.html
{'chars': np.float64(1474.1515151515152), 'chars:std': np.float64(341.7505399975447), 'score:std': np.float64(0.40520665017240837), 'score': np.float64(0.7929292929292929)}
Writing results to /tmp/gpqa_model_fp4_.json
Total latency: 1290.363 s
Score: 0.793
```

nsys trace shows the new kernel being used:

image

## Checklist

### DIFF
--- a/python/sglang/srt/layers/attention/nsa/nsa_indexer.py
+++ b/python/sglang/srt/layers/attention/nsa/nsa_indexer.py
@@ -9,6 +9,7 @@ from einops import rearrange
 from torch import nn
 
 from sglang.srt.custom_op import CustomOp
+from sglang.srt.layers.layernorm import LayerNorm
 from sglang.srt.utils import add_prefix, align, is_cuda, is_hip, is_npu
 
 if is_cuda():
@@ -83,24 +84,6 @@ def rotate_activation(x: torch.Tensor) -> torch.Tensor:
     return hadamard_transform(x, scale=hidden_size**-0.5)
 
 
-class V32LayerNorm(nn.Module):
-    """
-    Layer Normalization.
-    """
-
-    def __init__(self, dim: int, eps: float = 1e-6):
-        super().__init__()
-        self.dim = dim
-        self.eps = eps
-        self.weight = nn.Parameter(torch.ones(dim, dtype=torch.float32))
-        self.bias = nn.Parameter(torch.zeros(dim, dtype=torch.float32))
-
-    def forward(self, x: torch.Tensor):
-        return F.layer_norm(
-            x.float(), (self.dim,), self.weight, self.bias, self.eps
-        ).type_as(x)
-
-
 class Indexer(CustomOp):
     def __init__(
         self,
@@ -164,7 +147,7 @@ class Indexer(CustomOp):
                 bias=False,
                 prefix=add_prefix("weights_proj", prefix),
             )
-        self.k_norm = V32LayerNorm(self.head_dim)
+        self.k_norm = LayerNorm(self.head_dim)
         self.rotary_emb = get_rope_wrapper(
             rope_head_dim,
             rotary_dim=rope_head_dim,
--- a/python/sglang/srt/layers/attention/nsa/nsa_indexer.py
+++ b/python/sglang/srt/layers/attention/nsa/nsa_indexer.py
@@ -4,9 +4,7 @@ from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING, Any, Dict, Optional
 
 import torch
-import torch.nn.functional as F
 from einops import rearrange
-from torch import nn
 
 from sglang.srt.custom_op import CustomOp
 from sglang.srt.layers.layernorm import LayerNorm
@@ -147,7 +145,7 @@ class Indexer(CustomOp):
                 bias=False,
                 prefix=add_prefix("weights_proj", prefix),
             )
-        self.k_norm = LayerNorm(self.head_dim)
+        self.k_norm = LayerNorm(self.head_dim, dtype=torch.float32)
         self.rotary_emb = get_rope_wrapper(
             rope_head_dim,
             rotary_dim=rope_head_dim,
--- a/python/sglang/srt/layers/layernorm.py
+++ b/python/sglang/srt/layers/layernorm.py
@@ -53,6 +53,7 @@ if _is_cuda or _is_xpu:
     if _is_flashinfer_available:
         try:
             from flashinfer.norm import layernorm
+
             _flashinfer_layernorm_available = True
         except (ImportError, AttributeError):
             _flashinfer_layernorm_available = False
@@ -310,30 +311,22 @@ class LayerNorm(CustomOp):
         self.hidden_size = hidden_size
         self.variance_epsilon = eps
         self.elementwise_affine = elementwise_affine
+        self.use_bias = bias
         self.dtype = dtype
-        
-        if self.elementwise_affine:
-            self.weight = nn.Parameter(torch.ones(hidden_size, dtype=self.dtype))
-            if bias:
-                self.bias = nn.Parameter(torch.zeros(hidden_size, dtype=self.dtype))
-            else:
-                self.register_parameter("bias", None)
-        else:
-            self.register_parameter("weight", None)
-            self.register_parameter("bias", None)
+
+        self.bias = nn.Parameter(torch.zeros(hidden_size, dtype=self.dtype))
+        self.weight = nn.Parameter(torch.ones(hidden_size, dtype=self.dtype))
 
     def forward_cuda(
         self,
         x: torch.Tensor,
     ) -> torch.Tensor:
-        if not self.elementwise_affine:
-            return self.forward_native(x)
-        
-        if _flashinfer_layernorm_available and x.dtype == torch.bfloat16 and self.dtype == torch.float32:
-            beta = self.bias if self.bias is not None else torch.zeros_like(self.weight)
-            return layernorm(
-                x, self.weight, beta, self.variance_epsilon
-            )
+        if (
+            _flashinfer_layernorm_available
+            and x.dtype == torch.bfloat16
+            and self.dtype == torch.float32
+        ):
+            return layernorm(x, self.weight, self.bias, self.variance_epsilon)
         else:
             return self.forward_native(x)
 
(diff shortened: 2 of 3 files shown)


python/sglang/test/test_layernorm.py
```diff
@@ -3,7 +3,7 @@
 
 import torch
 
-from sglang.srt.layers.layernorm import GemmaRMSNorm, RMSNorm
+from sglang.srt.layers.layernorm import GemmaRMSNorm, LayerNorm, RMSNorm
```
Fridge003, line 6: Can you post the result of running this test

python/sglang/srt/layers/attention/nsa/nsa_indexer.py
```diff
@@ -147,7 +128,7 @@ def __init__(
             quant_config=quant_config,
             prefix=add_prefix("wk", prefix),
         )
-        self.k_norm = V32LayerNorm(self.head_dim)
+        self.k_norm = LayerNorm(self.head_dim)
```
Fridge003: Just for checking: is the input of `self.k_norm` here bf16?

akhilg-nv: I assume so, otherwise [this check](https://github.com/flashinfer-ai/flashinfer/blob/main/csrc/norm.cu#L183) will fail.

python/sglang/test/test_layernorm.py
akhilg-nv: Not sure what you mean - It passes for me.
```
python/sglang/test/test_layernorm.py ...                                                                                                                    [100%]

======================================================================== 3 passed in 6.54s ========================================================================
```

hlu1:
@akhilg-nv Please add the nsys profile result showing how much time the three kernels were using before applying the change

python/sglang/srt/layers/layernorm.py
```diff
@@ -260,6 +268,79 @@ def forward_with_allreduce_fusion(
...
+    ) -> None:
+        super().__init__()
+        self.hidden_size = hidden_size
+        self.variance_epsilon = eps
+        self.elementwise_affine = elementwise_affine
+
+        if self.elementwise_affine:
+            self.weight = nn.Parameter(torch.ones(hidden_size, dtype=torch.float32))
+            if bias:
+                self.bias = nn.Parameter(torch.zeros(hidden_size, dtype=torch.float32))
```
hlu1: This forces them to be fp32, which is not correct. Pass dtype as arg to __init__ and use dtype here.

https://github.com/pytorch/pytorch/blob/v2.9.0/torch/nn/modules/normalization.py#L106

python/sglang/srt/layers/layernorm.py
```diff
@@ -260,6 +268,79 @@ def forward_with_allreduce_fusion(
...
+            return layernorm(x, self.weight, beta, self.variance_epsilon)
+        else:
+            return self.forward_native(x)
+
+    def forward_native(
+        self,
+        x: torch.Tensor,
+    ) -> torch.Tensor:
+        orig_dtype = x.dtype
+        x = x.to(torch.float32)
```
hlu1: Cast to the dtype of weight

python/sglang/srt/layers/layernorm.py
```diff
@@ -260,6 +268,79 @@ def forward_with_allreduce_fusion(
...
+
+        mean = x.mean(dim=-1, keepdim=True)
+        variance = (x - mean).pow(2).mean(dim=-1, keepdim=True)
+        x = (x - mean) * torch.rsqrt(variance + self.variance_epsilon)
+
+        if self.elementwise_affine:
+            x = x * self.weight.to(torch.float32)
+            if self.bias is not None:
+                x = x + self.bias.to(torch.float32)
```
hlu1: Please use the pytorch layernorm kernel here directly

python/sglang/srt/layers/layernorm.py
```diff
@@ -275,6 +283,91 @@ def forward_with_allreduce_fusion(
...
+
+    def forward_cuda(
+        self,
+        x: torch.Tensor,
+    ) -> torch.Tensor:
+        if not self.elementwise_affine:
+            return self.forward_native(x)
+        
+        if _flashinfer_layernorm_available and x.dtype == torch.bfloat16 and self.dtype == torch.float32:
+            beta = self.bias if self.bias is not None else torch.zeros_like(self.weight)
```
hlu1: It's better to move `torch.zeros_like(self.weight)` to `__init__` so you don't incur any runtime overhead

python/sglang/srt/layers/layernorm.py
```diff
@@ -275,6 +283,91 @@ def forward_with_allreduce_fusion(
...
+    ) -> torch.Tensor:
+        if not self.elementwise_affine:
+            return F.layer_norm(x, (self.hidden_size,), eps=self.variance_epsilon)
+        
+        orig_dtype = x.dtype
+        x = x.to(self.dtype)
+        if self.bias is not None:
+            return F.layer_norm(x, (self.hidden_size,), weight=self.weight, bias=self.bias, eps=self.variance_epsilon).to(orig_dtype)
+        
+        return F.layer_norm(x, (self.hidden_size,), weight=self.weight, eps=self.variance_epsilon).to(orig_dtype)
```
hlu1:
```suggestion
        bias = self.bias or None
        return F.layer_norm(x, (self.hidden_size,), weight=self.weight, bias=bias, eps=self.variance_epsilon).to(orig_dtype)
```

python/sglang/srt/layers/layernorm.py
```diff
@@ -275,6 +283,91 @@ def forward_with_allreduce_fusion(
...
+            self.register_parameter("bias", None)
+
+    def forward_cuda(
+        self,
+        x: torch.Tensor,
+    ) -> torch.Tensor:
+        if not self.elementwise_affine:
+            return self.forward_native(x)
+        
+        if _flashinfer_layernorm_available and x.dtype == torch.bfloat16 and self.dtype == torch.float32:
```
hlu1: Combine the two if branches

python/sglang/test/test_layernorm.py
```diff
@@ -109,5 +109,63 @@ def test_gemma_rms_norm(self):
                 self._run_gemma_rms_norm_test(*params)
 
 
+class TestLayerNorm(CustomTestCase):
+    DTYPES = [torch.half, torch.bfloat16, torch.float32]
+    PARAM_DTYPES = [torch.half, torch.bfloat16, torch.float32]
+    NUM_TOKENS = [7, 83, 1024]
+    HIDDEN_SIZES = [768, 769, 770, 771, 5120, 5124, 5125, 5126, 8192, 8199]
```
hlu1: Remove some of the tests to keep the CI time reasonable. You don't need to test this many configs.

python/sglang/test/test_layernorm.py
```diff
@@ -109,5 +109,63 @@ def test_gemma_rms_norm(self):
...
+            layer.weight.data.normal_(mean=1.0, std=0.1)
+            if use_bias:
+                layer.bias.data.normal_(mean=0.0, std=0.1)
+        
+        scale = 1 / (2 * hidden_size)
+        x = torch.randn(num_tokens, hidden_size, dtype=dtype) * scale
+
+        with torch.inference_mode():
+            ref_out = layer.forward_native(x)
+            out = layer(x)
```
hlu1, line 145: In many of the cases, `layer(x)` would dispatch to `layer.forward_native(x)` so you're comparing the same code.

python/sglang/srt/layers/layernorm.py
```diff
@@ -275,6 +283,91 @@ def forward_with_allreduce_fusion(
...
+            )
+        else:
+            return self.forward_native(x)
+
+    def forward_native(
+        self,
+        x: torch.Tensor,
+    ) -> torch.Tensor:
+        if not self.elementwise_affine:
+            return F.layer_norm(x, (self.hidden_size,), eps=self.variance_epsilon)
```
hlu1: https://docs.pytorch.org/docs/stable/generated/torch.nn.LayerNorm.html
The pytorch version should support elementwise_affine true/false.

akhilg-nv: The functional API accomplishes this by setting weight/bias to None. I've simplified the logic to avoid branching.

python/sglang/test/test_layernorm.py
akhilg-nv: True, in these cases we won't get an accuracy check but will at least verify that we accept that input combination and use the right execution path. To account for this, I've added an additional check that casts input/params to the appropriate types to trigger the flashinfer kernel call path (with slightly greater error tolerance). 

When verifying the execution paths and accuracy, I found that for some reason, when the input is fp16 and params are bf16, casting to bf16/fp32 before calling `layer()` does trigger the flashinfer kernel execution path, but it somehow has an identical output to torch. However, fp16/fp32 test case does produce some error from casting as expected (and if the input is already bf16, this error is an order of magnitude smaller).

python/sglang/test/test_layernorm.py
akhilg-nv: I reduced some dtype and shape configs, and added some hidden sizes that match DSv3.2 params.

python/sglang/srt/layers/layernorm.py
akhilg-nv: I did some benchmarking and found for most cases I looked at, using the flashinfer kernel with weight = ones and bias = zeros is still faster than torch or torch compile. So I combined these execution paths (by default I now initialize weight/bias to ones/zeros). Ideally we can make the affine transform optional to the flashinfer kernel, but this shouldn't affect DSv3.2 perf anyway.

akhilg-nv:
@Fridge003 Could you provide insight on resolving the failing CI/CD tests?

I see errors that seem unrelated to my changes, like:

```
  File "/sglang-checkout/python/sglang/srt/models/deepseek_v2.py", line 2098, in forward_absorb_fused_mla_rope_prepare
    forward_batch.attn_backend.forward_metadata
AttributeError: 'HybridAttnBackend' object has no attribute 'forward_metadata'. Did you mean: 'init_forward_metadata'?
```

and 

```
  File "/sglang-checkout/python/sglang/srt/models/deepseek_v2.py", line 2097, in forward_absorb_fused_mla_rope_prepare
    attn_logits, _, kv_indptr, kv_indices, _, _, _ = (
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: cannot unpack non-iterable ForwardMetadata object
```

Fridge003:
@akhilg-nv That's unrelated to this PR

Fridge003 (approved):

PR #12047, opened 2025-10-24
fix(utils): close urllib responses to prevent FD leaks under concurrency

Nickcp39:
Motivation

This PR fixes the OSError: [Errno 24] Too many open files issue that occurs under high concurrency in http_request.
The problem was caused by unclosed urllib response objects in non-streaming requests, leading to file descriptor accumulation.

Modifications

Wrapped urllib.request.urlopen() in a with block for both normal and error paths.

Fully read response body before returning.

Returned a lightweight HttpResponse(status, body, headers) instead of the live urllib handle.

Streaming mode remains unchanged (caller is responsible for closing stream).

Added compatibility for both Python <3.13 and ≥3.13 (using cafile vs ssl context).

Accuracy Tests

This change does not affect model accuracy or inference results — it only impacts the HTTP utility layer.
No change to API or output structure (status code, JSON parsing, etc.).

Benchmarking and Profiling

Stress-tested locally with:

Dummy echo server: ThreadingHTTPServer on 127.0.0.1:5000

Stress client: 100 threads × 50 requests per thread (5,000 total)

Results:
✅ 0% failure rate
✅ Stable FD count (no leak)
✅ p99 latency < 30 ms
✅ Works for both /echo (200 OK) and /fail (HTTPError) cases

No performance regression observed.

Checklist

 Formatted with pre-commit

 Unit tested with 5k concurrent requests (yc_scripts/stress_http_version2.py)

 Verified on both Windows and Linux

 Behavior and API compatibility preserved

 Fixes #11923


Nickcp39:
> Can you remove .gitignore

Hi @b8zhong, I’ve cleaned up the branch — only `python/sglang/utils.py` is modified now.  
Could you please re-run CI or add the `run-ci` label when convenient? Thanks!

Issue #11942, opened 2025-10-22
[Bug] DeepSeek-V3.2: Illegal instruction / IMA during GPQA evaluation with DP attention (H200, TP=8, DP=8)


YAMY1234:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Describe the bug

When running GPQA evaluation on DeepSeek-V3.2-Exp with TP=8, DP=8, and --enable-dp-attention, the run intermittently crashes with an “illegal instruction was encountered” error that bubbles up via ProcessGroupNCCL watchdog.

Failure appears during prefill (e.g., #new-token: 1024) under multi-repeat evaluation.

Intermittent: in 4 attempts, 3 failed with IMA and 1 passed. Running a single repetition (--repeat 1) passes as well.

The stack trace points to a Triton kernel generated via TorchInductor inside nsa_indexer._get_logits_head_gate (which is annotated with @torch.compile(dynamic=True)).

### Launch server

```
python -m sglang.launch_server \
  --model deepseek-ai/DeepSeek-V3.2-Exp \
  --tp 8 \
  --dp 8 \
  --enable-dp-attention \
  --reasoning-parser deepseek-v3 \
  --chat-template chat_template.jinja
```

### Run GPQA eval (intermittent failure when using repeats)
```
python3 -m sglang.test.run_eval \
  --port 30000 \
  --eval-name gpqa \
  --num-examples 198 \
  --max-tokens 36000 \
  --repeat 8 \
  --thinking-mode deepseek-v3
```

### Actual behavior / Logs
```
[2025-10-21 22:11:28 DP1 TP1] Prefill batch. #new-seq: 1, #new-token: 1024, #cached-token: 0, token usage: 0.91, #running-req: 31, #queue-req: 27, 
[...]

PR #12052, opened 2025-10-24
Fix Illegal Instruction/IMA errors when using DP attention with DeepSeek-V3.2 models

YAMY1234:
## Motivation

Fix illegal memory access (IMA) errors when using DP attention with DeepSeek-V3.2 models.

**Issue**: When running with `--enable-dp-attention`, the DP gather operation in logits stage uses incorrect metadata, causing Triton kernel to access out-of-bounds memory.

**Root cause**: The `get_dp_local_info()` function always uses `global_num_tokens_gpu` (total tokens for attention/MLP), but in logits stage it should use `global_num_tokens_for_logprob_gpu` (pruned tokens needing logits computation).

### 1. Fix metadata source selection in `get_dp_local_info()`

**File**: `python/sglang/srt/layers/dp_attention.py`

Added logic to distinguish between two calling contexts:
- **LogitsMetadata**: Use `global_num_tokens_for_logprob_gpu` (correct for logits stage)
- **ForwardBatch**: Use `global_num_tokens_gpu` (correct for attention/MLP stage)

This ensures offset calculation uses the appropriate metadata field.

### 2. Use actual tensor size in `_dp_gather_via_all_reduce()`

**File**: `python/sglang/srt/layers/dp_attention.py`

Override `local_num_tokens` with `local_tokens.shape[0]` before calling `memcpy_triton`. Uses in-place `.fill_()` to maintain CUDA graph compatibility.

This provides a safety net since scheduler's `num_tokens_for_logprob` calculation doesn't account for logits_processor pruning.

## Accuracy Tests

```
python -m sglang.launch_server --model deepseek-ai/DeepSeek-V3.2-Exp --tp 8 --dp 8 --enable-dp-attention

```

```
Scores: ['0.798', '0.803', '0.803', '0.798', '0.788', '0.778', '0.773', '0.798']█████████████████████████████████████████████████████████████████████████████████▉      | 185/198 [41:56<01:16, 5.85s/it]
====================
Writing report to /tmp/gpqa_deepseek-ai_DeepSeek-V3.2-Exp.html███████████████▌                                                | 85/198 [42:52<28:55, 15.36s/it]
{'chars': np.float64(14306.5), 'chars:std': np.float64(11493.867823449567), 'score:std': np.float64(0.4015072103909452), 'score': np.float64(0.797979797979798)}
Total latency: 2573.581 s
Score: 0.798
```

```
Repeat: 8, mean: 0.799
Scores: ['0.793', '0.798', '0.813', '0.808', '0.813', '0.783', '0.768', '0.818']
====================
{'chars': np.float64(14737.838383838383), 'chars:std': np.float64(11840.990410141889), 'score:std': np.float64(0.38569460791993493), 'score': np.float64(0.8181818181818182)}
Total latency: 1584.823 s
Score: 0.818
```

## Checklist


YAMY1234:
Resolving https://github.com/sgl-project/sglang/issues/11942

python/sglang/srt/layers/dp_attention.py
```diff
@@ -355,16 +355,27 @@ def disable_dp_size():
...
-    # `get_dp_local_info` is only called in global DP gather and scatter. We use global DP rank here.
+    # Compute local token range for current DP rank.
+    # For LogitsMetadata, use global_num_tokens_for_logprob_gpu (pruned tokens needing logits).
+    # For ForwardBatch (MLP/Attention), use global_num_tokens_gpu (all tokens in batch).
     dp_rank = get_attention_dp_rank()
 
     if forward_batch.dp_local_start_pos is None:
-        cumtokens = torch.cumsum(forward_batch.global_num_tokens_gpu, dim=0)
+        # Select metadata source based on context
+        if (
```
ch-wan: How can `dp_local_start_pos` be None in `LogitsMetadata`? It is expected to be processed here: https://github.com/sgl-project/sglang/blob/0bfa394affbf62db60390411294ae6c67cb24455/python/sglang/srt/layers/logits_processor.py#L577.

`LogitsMetadata` has a separate processing logic. We should keep it separated. Perhaps we should add assertion like `assert isinstance(forward_batch, ForwardBatch)` to avoid misuse.

YAMY1234: Thanks, understand the design! The call logic should indeed stay separated at the outer level. Will have another patch soon after verified

YAMY1234: @ch-wan Just pushed another patch, could you take another look? Thanks

YAMY1234:
Root fix is now in a new PR: https://github.com/sgl-project/sglang/pull/12115 which can resolve the issue with in a more robust way. Closing this PR

PR #12058, opened 2025-10-24
Support true on-policy

fzyzcjy:
## Motivation

it is ready to use

plase merge https://github.com/sgl-project/sglang/pull/11958 before this one

logprob between train and rollout always zero:

image

## Checklist


python/sglang/srt/layers/logits_processor.py
```diff
@@ -593,6 +593,11 @@ def _get_logits(
                     None,  # bias
                     True,  # is_vnni
                 )
+            elif get_global_server_args().rl_on_policy_target == "fsdp":
+                # Due to tie-weight, we may not be able to change lm_head's weight dtype
+                logits = torch.matmul(
+                    hidden_states.bfloat16(), lm_head.weight.T.bfloat16()
```
hebiao064, line 599: why do we need to cast to bf16? i thought its already bf16

fzyzcjy: iirc (forget details) embedding must be fp32 to align w/ hf, then w/ weight tie the lm head is also fp32, but we want bf16 to align w/ hf here

hebiao064: is FSDP using BF16 for LM Head matmul? I thought FSDP will use FP32 for everything?

fzyzcjy: I forgot but matmul is bf16 at least

Issue #11289, opened 2025-10-07
[Bug] update_weights_from_disk() results in a fatal server crash for gpt-oss


harrisonlimh:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Describe the bug

update_weights_from_disk() results in a fatal server crash for gpt-oss.

### Reproduction

Server cmd:
```
docker run -it   --shm-size 256g   --gpus all     --ipc=host   --network=host   --privileged   "lmsysorg/sglang:latest" python3 -m sglang.launch_server --model openai/gpt-oss-20b --tp 8 --reasoning-parser gpt-oss --tool-call-parser gpt-oss --port=7080
```

Client request
```
$ curl  -X POST -H "Content-Type: application/json" http://0.0.0.0:7080/update_weights_from_disk -d '{ "model_path": "openai/gpt-oss-20b" }'
Internal Server Error
```

Server log
```
[2025-10-07 01:23:56] Start update_weights. Load format=auto
[2025-10-07 01:23:56 TP0] Update engine weights online from disk begin. avail mem=9.14 GB
[2025-10-07 01:23:56 TP0] Using model weights format ['*.safetensors']
[2025-10-07 01:23:56 TP1] Using model weights format ['*.safetensors']
[2025-10-07 01:23:56 TP5] Using model weights format ['*.safetensors']
[2025-10-07 01:23:56 TP4] Using model weights format ['*.safetensors']
[2025-10-07 01:23:56 TP2] Using model weights format ['*.safetensors']
[2025-10-07 01:23:56 TP7] Using model weights format ['*.safetensors']
[2025-10-07 01:23:56 TP6] Using model weights format ['*.safetensors']
[2025-10-07 01:23:56 TP3] Using model weights format ['*.safetensors']
Loading safetensors checkpoint shards:   0% Completed | 0/3 [00:00<?, ?it/s]
[...]

PR #12060, opened 2025-10-24
Conditionally recapture cuda graph after model weight update from disk 

harrisonlimh:
## Motivation
* https://github.com/sgl-project/sglang/issues/11289 - resolve cuda illegal memory access error that occurs after model weight update. However, this fix seems to be only applicable to the gpt-oss update weight fix, and the endpoint has been [working for other models](https://github.com/sgl-project/sglang/blob/main/test/srt/rl/test_update_weights_from_disk.py)
* We will follow up with a further debugging on gpt-oss 

## Modifications
* Add cuda graph reinitializtion step in update_weights_from_disk(). Only use it if the new server arg is used.

## Checklist


python/sglang/srt/server_args.py
```diff
@@ -2897,6 +2898,11 @@ def add_cli_args(parser: argparse.ArgumentParser):
             action="store_true",
             help="Enable garbage collection during CUDA graph capture. If disabled (default), GC is frozen during capture to speed up the process.",
         )
+        parser.add_argument(
```
merrymercy: Do not add too many arguments.

Can you add the argument to https://github.com/sgl-project/sglang/blob/ceb105a7802329165a14337eb46c6da6e094f124/python/sglang/srt/managers/io_struct.py#L997?

merrymercy (changes requested):

python/sglang/srt/server_args.py
harrisonlimh: Done.

PR #12068, opened 2025-10-24
[5/N] (Elastic EP) Use GPU P2P to exchange expert weights during EPLB as much as possible

UNIDY2002:
## Motivation

In the Elastic EP core structure introduced in #10606, a full weight reloading from the disk was required during rebalancing, when Elastic EP is enabled.

This PR optimizes the part by applying GPU P2P communication as much as possible, reusing the `ExpertLocationUpdater` implemented by @fzyzcjy. Only the weights on dead peers have to be reloaded from the disk.

Unit test `ep.test_mooncake_ep_small.TestMooncakeWithEPLB` shows a decrease in rebalancing time from ~1.7s to ~0.2s.

## Modifications

The major changes are in `expert_location_updater.py`. The logical experts that could not be loaded from the peers are recorded in `all_missing_logical_experts`. The ModelRunner will then load these weights from the disk.

## Accuracy Tests

All existing unit-tests should pass.

## Benchmarking and Profiling

Unit test `ep.test_mooncake_ep_small.TestMooncakeWithEPLB` shows a decrease in rebalancing time from ~1.7s to ~0.2s.

All existing performance tests should pass.

## Checklist


python/sglang/srt/eplb/expert_location_updater.py
```diff
@@ -65,6 +71,25 @@ def update(
...
+        if len(all_missing_logical_experts) > 0:
+            logger.info(
+                f"[ExpertLocationUpdater] All missing logical experts: {all_missing_logical_experts}"
+            )
+
+            # Load the missing expert weights from disk
+            def missing_experts_name_filter(name: str) -> bool:
+                for layer, experts in all_missing_logical_experts:
+                    for expert in experts:
+                        if f"layers.{layer}.mlp.experts.{expert}." in name:
```
fzyzcjy: is this specific to a few models instead of all models

fzyzcjy (commented):
nits

python/sglang/srt/eplb/expert_location_updater.py
UNIDY2002: Thank you for pointing it out. I'll figure out a more generic solution.

UNIDY2002: @fzyzcjy I changed the part into calling a per-model `generate_weight_name_filter` method. How do you think of the solution?

UNIDY2002: Any comments or suggestions? Thanks 😄

python/sglang/srt/elastic_ep/expert_backup_client.py
```diff
@@ -98,7 +98,7 @@ def start_transfer_client(self):
                 logger.warning("Register fails. Stop using expert weight backup!")
                 break
 
-    def update_weights(self):
+    def update_weights(self, weight_name_filter):
```
ShangmingCai: Should we make weight_name_filter Optional?

python/sglang/srt/eplb/expert_location_updater.py
```diff
@@ -467,7 +492,7 @@ def _get_local_expert_location(expert_location: int) -> int:
 
     _entrypoint()
 
-    return output_logs
+    return output_logs, missing_logical_experts
```
ShangmingCai: Is it possible that we replace these change with a more clear implementation? I find that I need to view the whole file to understand what this block is doing, LOL.

ShangmingCai (commented):
Overall, LGTM for the logic, only some nits.

python/sglang/srt/eplb/expert_location_updater.py
UNIDY2002: Yes. I moved those logic to a function `_filter_p2p_ops`. In addition, I changed the data structure of `missing_logical_experts_by_layers` to a map, which is more intuitive.

python/sglang/srt/models/deepseek_common/deepseek_weight_loader.py
```diff
@@ -609,6 +609,27 @@ def post_load_weights(
...
+        """
+
+        def weight_name_filter(name: str) -> bool:
+            for layer_id in logical_experts_map:
+                for expert in logical_experts_map[layer_id]:
+                    if f"layers.{layer_id}.mlp.experts.{expert}." in name:
+                        return True
+            return False
+
+        return weight_name_filter
```
ShangmingCai, line 639: Should we make a set and make this an O(1) lookup? The double `for` iterate for each `name` seems inefficient to me.

python/sglang/srt/eplb/expert_location_updater.py
```diff
@@ -37,6 +38,12 @@ class ExpertLocationUpdater:
     def __init__(self):
         self._first_execution = True
 
+    """
+    Update experts' physical location after EPLB.
+
+    Returns a map of layer_id to expert_ids that are missing due to rank failures during fault conditions when elastic EP is enabled.
+    """
+
```
ShangmingCai: This docsting location seems weird to me, please place it in the update function.

python/sglang/srt/eplb/expert_location_updater.py
```diff
@@ -434,6 +454,22 @@ def _compute_comm_info(logical_expert_id: int):
...
+        if elastic_ep_state is not None and missing_logical_experts_info is not None:
+            # Filter out inactive P2P ops and record missing expert IDs in missing_logical_experts_info
+            is_active = elastic_ep_state.active_ranks_cpu
+            for i, (logical_expert_id, ops) in enumerate(p2p_op_infos):
+                if any(op.op == torch.distributed.isend for op in ops):
+                    p2p_op_infos[i] = (
+                        logical_expert_id,
+                        [op for op in ops if is_active[op.peer]],
+                    )
+                elif any(op.op == torch.distributed.irecv for op in ops):
```
ShangmingCai: Is it possible that `isend` and `irecv` both are in the `p2p_op_infos` at the same time?

ShangmingCai (approved):
LGTM

PR #12079, opened 2025-10-24
[Router][fix]system message support array of content parts

jimmy-evo:
## Motivation

system message support array of content parts

```json
{
    "model": "test",
    "messages": [
        {
            "role": "system",
            "content": [
                {
                    "type":"text",
                    "text":"hello"
                }
            ]
        }
    ],
    "stream": false,
    "temperature": 0
}
```
image

## Modifications

change system message

## Checklist


slin1237 (approved):

sgl-router/src/protocols/chat.rs
```diff
@@ -19,7 +19,7 @@ use crate::protocols::validated::Normalizable;
 pub enum ChatMessage {
     #[serde(rename = "system")]
     System {
-        content: String,
+        content: TextMessageContent,
```
CatherineSue: I thought the openai doc shows it could be a string or an array of contents type.
This is not the same as openai's format.

CatherineSue (changes requested):

CatherineSue:
I feel a correct way to handle this kind of changes is to have a correct generic content parts that can be applied to both UserMessage and SystemMessage.

Having more types added to protocols is hard to maintain in the future.

sgl-router/src/protocols/chat.rs
jimmy-evo: > I thought the openai doc shows it could be a string or an array of contents type. This is not the same as openai's format.

@CatherineSue 

in openai reference

> only type text is supported

jimmy-evo:
> I feel a correct way to handle this kind of changes is to have a correct generic content parts that can be applied to both UserMessage and SystemMessage.
> 
> Having more types added to protocols is hard to maintain in the future.

sounds good.

just changed it.

system and user message use a same struct type. much more convinient for develop
@CatherineSue

lx-eotk:
hey, should assistant messages also support arrays of content parts? @jinmingyi1998

jimmy-evo:
> hey, should assistant messages also support arrays of content parts? @jinmingyi1998

@lx-eotk 

i think so.

> An array of content parts with a defined type. Can be one or more of type text, or exactly one of type refusal.

PR #12087, opened 2025-10-24
Fix kernel version bump file

Kangyan-Zhou:
Enforce changes in toml file

version bump: https://github.com/sgl-project/sglang/pull/12102
kernel bump: https://github.com/sgl-project/sglang/pull/12088

### DIFF
--- a/.github/workflows/bot-bump-sglang-version.yml
+++ b/.github/workflows/bot-bump-sglang-version.yml
@@ -26,6 +26,10 @@ jobs:
         with:
           python-version: '3.10'
 
+      - name: Install Python dependencies
+        run: |
+          pip install tomli
+
       - name: Configure Git and branch
         run: |
           git config user.name "sglang-bot"
--- a/.github/workflows/bot-bump-kernel-version.yml
+++ b/.github/workflows/bot-bump-kernel-version.yml
@@ -26,6 +26,10 @@ jobs:
         with:
           python-version: '3.10'
 
+      - name: Install Python dependencies
+        run: |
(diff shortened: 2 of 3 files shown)


.github/workflows/bot-bump-sglang-version.yml
```diff
@@ -26,6 +26,10 @@ jobs:
         with:
           python-version: '3.10'
 
+      - name: Install Python dependencies
+        run: |
+          pip install tomli
```
zhyncs: Don't install dependencies in yml, let's update bash script instead

zhyncs (changes requested):

.github/workflows/bot-bump-sglang-version.yml
Kangyan-Zhou: Do we want to update scripts/ci/ci_install_dependency.sh? That file is very large and most of them are not relevant. Or I can create a different shell script

zhyncs: Can we update in ci_install_dependency.sh

Kangyan-Zhou: ci_install_dependency.sh is mostly used in GPU nodes, if I run ci_install_dependency.sh it will complain nvidia-smi not found, and I feel like we don't have to refactor that file for a simple change in these two workflows

zhyncs: I see

zhyncs (approved):

PR #12110, opened 2025-10-25
fix: release workflow should work on both archs

ishandhanani:
Added support for building and releasing on the aarch64 architecture in the GitHub Actions workflow for kernel releases.
- Added `arch: [x86_64, aarch64]` to the build matrix in `.github/workflows/release-whl-kernel.yml`.


zhyncs:
@ishandhanani may we trigger manually

zhyncs (approved):

PR #12113, opened 2025-10-25
[misc] depdencies & enviroment flag

hnyls2002:
- Remove `gguf` in optional dependency (already in dependency)
- Remove `--enable-beta-spec` and introduce `SGLANG_ENABLE_SPEC_V2` environment vairable.


python/sglang/srt/server_args.py
```diff
@@ -2419,7 +2422,6 @@ def add_cli_args(parser: argparse.ArgumentParser):
         )
 
         # Speculative decoding
-        parser.add_argument("--enable-beta-spec", action="store_true")
```
zhyncs, line 2445: why do we remove this

hnyls2002: Moved to environ. For this "preview" feature, it's better to use an environment variable.

PR #12114, opened 2025-10-25
[Do Not Merge][GDN] Mamba Radix Cache Page Size > 1 decode POC

hanming-lu:
```
python3 -m sglang.launch_server --model Qwen/Qwen3-Next-80B-A3B-Instruct --tp 4 --page-size 64

python python/sglang/test/send_one.py --turns 2
```

### DIFF
--- a/python/sglang/test/send_one.py
+++ b/python/sglang/test/send_one.py
@@ -1,10 +1,20 @@
 """
-Run one test prompt.
+Run one test prompt with support for multi-turn conversations and client-side tokenization.
+
+Features:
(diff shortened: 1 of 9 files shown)


python/sglang/test/send_one.py
```diff
@@ -1,20 +1,29 @@
 """
-Run one test prompt.
+Run one test prompt with support for multi-turn conversations and client-side tokenization.
```
hanming-lu, line 2: should revert this file.

PR #12142, opened 2025-10-26
Support DeepGEMM for deterministic inference

fzyzcjy:
## Checklist


python/sglang/srt/batch_invariant_ops/batch_invariant_ops.py
```diff
@@ -217,6 +227,53 @@ def grid(META):
...
+    a: torch.Tensor, b: torch.Tensor, bias: torch.Tensor | None = None
+):
+    if (
+        ENABLE_JIT_DEEPGEMM
+        and (a.dtype == torch.bfloat16)
+        and (b.dtype == torch.bfloat16)
+        and a.is_contiguous()
+        and b.transpose(0, 1).is_contiguous()
+    ):
+        if _ENABLE_MM_COMPARISON_TEST:
```
Fridge003, line 264: Can we move the test here
https://github.com/sgl-project/sglang/blob/main/test/srt/batch_invariant/test_batch_invariant_ops.py

fzyzcjy: it is a on-the-fly comparison test, i.e. can be enable in runtime to test the real use cases. but it looks good to enable this extra flag at test_batch_invariant_ops, will do it later

fzyzcjy:
seems unrelated

image

image

PR #12144, opened 2025-10-26
Add Batch‑Invariant RMSNorm

zyzshishui:
## Checklist


python/sglang/srt/layers/layernorm.py
```diff
@@ -100,6 +102,14 @@ def forward_cuda(
     ) -> Union[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]]:
         if self.variance_size_override is not None:
             return self.forward_native(x, residual)
+        if is_batch_invariant_mode_enabled():
+            if residual is not None:
```
fzyzcjy: maybe when `get_global_server_args().rl_on_policy_target == "fsdp"` also forcefully call native fn now to avoid breaking true on policy on slime. I will enable the flag and check whether computation is compatible and also handle residual-is-not-none in sync w/ slime.

fzyzcjy (approved):
LGTM w/ nit above

fzyzcjy:
btw ci fails

image

python/sglang/srt/layers/layernorm.py
zyzshishui: done

python/sglang/srt/layers/layernorm.py
```diff
@@ -100,6 +103,17 @@ def forward_cuda(
     ) -> Union[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]]:
         if self.variance_size_override is not None:
             return self.forward_native(x, residual)
+        if is_batch_invariant_mode_enabled():
+            if (
+                residual is not None
+                or get_global_server_args().rl_on_policy_target == "fsdp"
```
Fridge003, line 109: Will this condition cause difference behavior between fsdp and sglang?

zyzshishui: @fzyzcjy

fzyzcjy: in true-on-policy-with-fsdp, we should (at least currently) use the forward_native b/c the fused triton kernel may give diff numeric values

Fridge003 (approved):

python/sglang/srt/layers/layernorm.py
hebiao064: for the fused triton kernel, have we tested whether its batch invariant?

PR #12145, opened 2025-10-26
[Feature] Support token ids input for chat completion endpoint

yitianlian:
## Motivation

For agent reinforcement learning (RL) training, it is necessary to store the full token ID list to avoid retokenization mismatches and to record rollout log probabilities (logp) from SGLang for further operations such as clipping.

However, many agent frameworks used in RL training rely on the OpenAI Chat Completion endpoint rather than the /generate endpoint.

Therefore, supporting token-in/out functionality for the OpenAI Chat Completion API is essential to better enable RL training.

## Modifications

1. Added a new key `input_ids` to the `ChatCompletionRequest `class to support passing token IDs directly to the Chat Completion endpoint.
2. Updated the input validation logic to accept either user messages or input_ids as valid inputs.
3. Modified the `_convert_to_internal_request` method to handle `input_ids` input while maintaining compatibility with message-based input.
4. Added unit tests for `input_ids` input to ensure correct behavior and compatibility.
## Accuracy Tests

No
## Benchmarking and Profiling

No
## Checklist


zhaochenyang20:
So great!

test/srt/test_sagemaker_server.py
```diff
@@ -174,6 +174,120 @@ def test_chat_completion_stream(self):
             for parallel_sample_num in [1, 2]:
                 self.run_chat_completion_stream(logprobs, parallel_sample_num)
 
+    def test_chat_completion_with_input_ids(self):
```
hebiao064: why we add test in `test_sagemaker_server`?

python/sglang/srt/entrypoints/openai/protocol.py
fzyzcjy, line 1: qq: for token-in-token-out, iirc slime-router also did something about it. may I know your plan about these (slime-router, sgl-router, sgl core)?

Also, it would be great to have a brief explanation of how users may use this. e.g. here it seems it inputs token ids but output strings instead of ids.

test/srt/test_sagemaker_server.py
```diff
@@ -174,6 +174,120 @@ def test_chat_completion_stream(self):
...
+            "model": self.model,
+            "input_ids": input_ids,
+            "temperature": 0,
+            "max_tokens": 50,
+        }
+
+        headers = {"Authorization": f"Bearer {self.api_key}"}
+
+        response = requests.post(
+            f"{self.base_url}/invocations", json=data, headers=headers
```
fzyzcjy: question: I am a bit confused, shall we consider the /v1/chat/completinos api instead

fzyzcjy (commented):
I am a bit confused about the goal of this PR, thus not started reviewing the real code and only checked the goal and the test.

(btw feel free to discuss here as well -> https://sgl-fru7574.slack.com/archives/C09HMG80PNE/p1761690103681309)

python/sglang/srt/entrypoints/openai/protocol.py
```diff
@@ -543,6 +547,23 @@ def set_tool_choice_default(cls, values):
...
+        messages = values.get("messages")
+        input_ids = values.get("input_ids")
+
+        if messages is None and input_ids is None:
+            raise ValueError("Either 'messages' or 'input_ids' must be provided")
+
+        if messages is not None and input_ids is not None:
+            raise ValueError(
+                "Cannot provide both 'messages' and 'input_ids'. Please use only one."
+            )
```
fzyzcjy, line 563: nit

```
if (messages is None) == (input_ids is None):
  raise ...
```

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -78,8 +78,65 @@ def __init__(
...
+            server_context_length = self.tokenizer_manager.server_args.context_length
+            if (
+                max_output_tokens
+                and server_context_length
+                and max_output_tokens > server_context_length
+            ):
+                return (
+                    f"max_completion_tokens is too large: {max_output_tokens}."
+                    f"This model supports at most {server_context_length} completion tokens."
+                )
```
fzyzcjy: nit: would be great to avoid copy-pasting. e.g. for this part, maybe have a tiny function like `_validate_request_max_tokens` and call it in both token-in and text-in case

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -142,29 +205,63 @@ def _convert_to_internal_request(
...
+                    tool_call_constraint = parser.get_structure_constraint(
+                        request.tool_choice
+                    )
+                if request.tool_choice == "required" or isinstance(
+                    request.tool_choice, ToolChoice
+                ):
+                    json_schema = get_json_schema_constraint(
+                        request.tools, request.tool_choice
+                    )
+                    tool_call_constraint = ("json_schema", json_schema)
```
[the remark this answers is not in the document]
fzyzcjy: +1 this is exactly what I want to say - would be great to avoid dup

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -78,8 +78,65 @@ def __init__(
     def _request_id_prefix(self) -> str:
         return "chatcmpl-"
 
+    def _get_template_stop_strings(self, request: ChatCompletionRequest) -> List[str]:
```
fzyzcjy, line 81: (I do not check correctness of this part, if it runs well for you then I guess it is ok)

fzyzcjy (commented):
some nits

python/sglang/srt/entrypoints/openai/protocol.py
fzyzcjy: discussed offline

test/srt/test_sagemaker_server.py
yitianlian: I have changed the test file.

python/sglang/srt/entrypoints/openai/protocol.py
yitianlian: I noticed that the OpenAI client checks for the presence of a message key in the request. If the request does not contain this key, an error is thrown. To avoid this issue, the current implementation uses an empty message and stores the input_ids in the extra_info. SGLang will ignore the message content when input_ids is not None.

python/sglang/srt/entrypoints/openai/protocol.py
```diff
@@ -543,6 +547,23 @@ def set_tool_choice_default(cls, values):
...
+    @classmethod
+    def validate_messages_or_input_ids(cls, values: Dict):
+        """Ensure either messages or input_ids is provided, but not both."""
+        messages = values.get("messages")
+        input_ids = values.get("input_ids")
+
+        if messages is None and input_ids is None:
+            raise ValueError("Either 'messages' or 'input_ids' must be provided")
+
+        if messages is not None and input_ids is not None:
```
zhaochenyang20, line 560: this should be an `if else`, not `if if`.

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -78,8 +78,101 @@ def __init__(
...
+        """Get stop strings from conversation template if available."""
+        if self.template_manager.chat_template_name is None:
+            # Jinja template doesn't have built-in stop strings
+            return []
+
+        if request.ignore_eos:
+            return []
```
zhaochenyang20, line 88: `if or`?

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -78,8 +78,101 @@ def __init__(
...
+
+        # Import chat_templates to get template stop strings directly
+        from sglang.srt.parser.conversation import chat_templates
```
zhaochenyang20, line 91: move the import at the top of the file

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -78,8 +78,101 @@ def __init__(
...
+
+        template_name = self.template_manager.chat_template_name
+        if template_name in chat_templates:
+            template = chat_templates[template_name]
+            return copy.copy(template.stop_str or [])
+
+        return []
```
zhaochenyang20, line 98: if default return a `[]`, could we use a large `if elif else` to control the whole logic of the function?

zhaochenyang20 (changes requested):
minimal changes required

python/sglang/srt/entrypoints/openai/protocol.py
yitianlian: I want to print a warning when sglang receives both input ids and messages. The input ids' priority should be higher than messages.

PR #12149, opened 2025-10-26
Add an unified server arg for multimodal inputs preprocessing config.

WingEdge777:
## Motivation

As multimodal inputs can result in lots of input tokens, we sometimes need to limit the input size manually. Otherwise, CUDA OOM probably happens if the server is taking a large image or video as input.

## Modifications

We added a server argument `--mm-process-config` to pass the limit configuration for multimodel preprocessing limit hint.

This PR takes `qwen_vl`(we encountered OOM and CUDA OOM as its default config allows large tokens input) as an example, but could be easily extended to other multimodel models.

The current JSON config example is as follows:
```bash
--mm-process-config '{"image":{"max_pixels":1048576},"video":{"fps":3,"max_pixels":602112,"max_frames":60}}'
```
The JSON value corresponding to key names `image`, `video`, and  `audio` should be the limit config for preprocessing.  
When the timing comes, it could be defined as a config class for improved code readability.

Example value explanation, e.g.
`max_pixels`: determining whether to resize input; 
`fps`: the frame sampling rate for video input.

## Checklist


JustinTong0323:
Could you help add some docs for the example usage? Maybe under `supported/supported_models/multimodal_language_models.html#usage-notes` ? Thanks !

WingEdge777:
@JustinTong0323 Have done it. Could you please take a look, too. Thanks!

docs/advanced_features/server_arguments.md
```diff
@@ -156,6 +156,7 @@ Please consult the documentation below and [server_args.py](https://github.com/s
 | `--base-gpu-id` | The base GPU ID to start allocating GPUs from. Useful when running multiple instances on the same machine. | `0` | Type: int |
 | `--gpu-id-step` | The delta between consecutive GPU IDs that are used. For example, setting it to 2 will use GPU 0,2,4,... | `1` | Type: int |
 | `--sleep-on-idle` | Reduce CPU usage when sglang is idle. | `False` | bool flag (set to enable) |
+| `--mm-config` | A JSON string for multimodal preprocessing configuration. It can contain keys: `image`, `video`, `audio`. | `{}` |
```
mickqian: better to have `process` keyword in it

python/sglang/srt/server_args.py
```diff
@@ -2174,6 +2175,12 @@ def add_cli_args(parser: argparse.ArgumentParser):
             action="store_true",
             help="Reduce CPU usage when sglang is idle.",
         )
+        parser.add_argument(
+            "--mm-config",
+            type=json.loads,
+            default=ServerArgs.mm_config,
+            help="Multimodal preprocessing config, a json config contains keys: image, video, audio",
```
mickqian: nit: \`image\`, \`video\`,...

docs/advanced_features/server_arguments.md
WingEdge777: Done

python/sglang/srt/server_args.py
WingEdge777: done

python/sglang/srt/server_args.py
```diff
@@ -276,6 +276,7 @@ class ServerArgs:
     base_gpu_id: int = 0
     gpu_id_step: int = 1
     sleep_on_idle: bool = False
+    mm_process_config: Dict = {}
```
JustinTong0323: To give a default `dict` for a field in a Python dataclass, you must use the `field` function from the `dataclasses` module along with the `default_factory` argument. **You should not use a mutable type like `dict` directly as a default value**, as this would result in all instances sharing the *same* dictionary, leading to unexpected behavior.

Here's how to do it:

## 🛠️ Using `default_factory` for a Default `dict`

Use the `dataclasses.field()` function and set the `default_factory` argument to the type you want to be the default, which in this case is `dict`.

```python
from dataclasses import dataclass, field
from typing import Dict

@dataclass
class Inventory:
    """A dataclass representing an item inventory."""
    item_id: int
    # Use default_factory=dict to ensure each instance gets a *new* empty dictionary
    properties: Dict[str, str] = field(default_factory=dict)
    price: float = 0.0
```

-----

### **Why `default_factory` is Necessary**

Using `properties: Dict[str, str] = {}` (or any other mutable default like a list or set) is **incorrect** because:

1.  **Shared State:** All instances of `Inventory` would share the *exact same* dictionary object in memory.
2.  **Unexpected Behavior:** If you modify the `properties` dictionary on one instance, that change would *also* appear on all other instances that didn't explicitly set a `properties` value.

**Example of the problem (Incorrect approach):**

```python
@dataclass
class BadInventory:
    item_id: int
    # 🚨 DANGER: This is a mutable default shared by all instances!
    properties: Dict[str, str] = {}

item1 = BadInventory(item_id=1)
item2 = BadInventory(item_id=2)

item1.properties['color'] = 'red'

# 🛑 Oops! item2's dictionary is also modified because they share the same object
print(item2.properties)
# Output: {'color': 'red'} (Incorrect behavior)
```

**Correct Example using `default_factory`:**

```python
@dataclass
class GoodInventory:
    item_id: int
    properties: Dict[str, str] = field(default_factory=dict)

item_a = GoodInventory(item_id=10)
item_b = GoodInventory(item_id=20)

item_a.properties['size'] = 'large'

# ✅ Correct! item_b has its own independent, empty dictionary
print(item_b.properties)
# Output: {} (Correct behavior)
```

The `default_factory` ensures that a **new, fresh** dictionary object is created every time a new instance of the dataclass is initialized *without* a value provided for that field.

JustinTong0323: Above are from gemini, although there would not have 2 server_args, we'd better follow the best practice.

JustinTong0323 (changes requested):

python/sglang/srt/server_args.py
WingEdge777: Thanks for pointing out this problem! We use an optional `None` as a default value in our offline test. I did not realize this in this pr, sorry for my fault.

I have modified to use default_factory for dataclass's default dict value, please take a look again. Thanks!

mickqian (approved):

mickqian:
please fix ci

WingEdge777:
Not sure if I can fix all CI...

WingEdge777:
@JustinTong0323, could you please take some time to review these changes

python/sglang/srt/server_args.py
```diff
@@ -2263,6 +2264,12 @@ def add_cli_args(parser: argparse.ArgumentParser):
             action="store_true",
             help="Reduce CPU usage when sglang is idle.",
         )
+        parser.add_argument(
+            "--mm-process-config",
+            type=json.loads,
+            default={},
```
JustinTong0323:
```suggestion
            default=ServerArgs().mm_process_config,
```

JustinTong0323 (changes requested):

python/sglang/srt/server_args.py
JustinTong0323: I saw you revert this change, may I know the reason?

JustinTong0323: (also please ping me on slack for pushing the PR, too many github notifications 🫠

WingEdge777: To fix CI. The `ServerArgs.mm_process_config: Dict[str, Any] = dataclasses.field(default_factory=dict)` actually is not a valid class value before the ServerArgs are initialized. which will cause atrribute not found error

>## 🔍 Analysis of the Error: Conflict Between Class and Instance Attributes
>### 1. Field Definition (dataclass Instance Attribute)
>
>You added this line to the ServerArgs class:
>
>`mm_process_config: Dict[str, Any] = dataclasses.field(default_factory=dict)`
>
> - Meaning: This line defines an instance field named mm_process_config. It instructs the dataclass that when an instance of >ServerArgs is created, if the user does not provide a value, it should call dict to generate a default value. 
> - Key Point: The value of this field is determined only when the instance is created; it is not a regular class attribute directly >stored on the ServerArgs class itself.
>
>### 2. Command-Line Argument Parsing (argparse Accessing Class Attribute)
>
>According to the traceback you provided, the error occurs in the `add_cli_args` function:
>
>##### Error occurs here:
> `default=ServerArgs.mm_process_config`
> - Intended Behavior: This code attempts to access ServerArgs.mm_process_config and pass its value, as a static class's attribute, to the default parameter of parser.add_argument. 
> - Actual Situation: Because mm_process_config is a field defined >using dataclasses.field(...), a regular attribute named mm_process_config does not exist at the class level. Consequently, python raises an AttributeError.

Above is Gemini's reply.

JustinTong0323: Got you, that's interesting. Try `ServerArgs().mm_process_config,`

WingEdge777: ServerArgs can't be initialized with empty args. I had a test:
```bash
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/root/workspace/sglang/python/sglang/launch_server.py", line 11, in <module>
    server_args = prepare_server_args(sys.argv[1:])
  File "/root/workspace/sglang/python/sglang/srt/server_args.py", line 3989, in prepare_server_args
    ServerArgs.add_cli_args(parser)
  File "/root/workspace/sglang/python/sglang/srt/server_args.py", line 2270, in add_cli_args
    default=ServerArgs().mm_process_config,
            ^^^^^^^^^^^^
TypeError: ServerArgs.__init__() missing 1 required positional argument: 'model_path'
```
I can add a dummy value like `ServerArgs(model_path="dummy").mm_process_config` as a workaround, but I'm not sure its really necessary to initialize a ServerArgs instance for providing a default dict `{}`.  There are other arguments set by default with non-ServerArgs's attribute, e.g., `default="localhost:4317",`

WingEdge777: Maybe I can add a global value 
``` python
MM_PROCESS_CONFIG_DEFAUlL = {}

# then for argument
default=MM_PROCESS_CONFIG_DEFAUlL,
```
How do you think about it

JustinTong0323 (approved):

WingEdge777:
@mickqian many thanks！

PR #12159, opened 2025-10-26
[sgl-route] Optimize the use of constant slices and retain to simplif…

lengrongfu:
…y SGLang field removal

## Motivation

There are similar operations in the chart and response APIs, so this field is extracted.

## Checklist


slin1237:
thanks for the contribution
pr lgtm, just some minor fmt error
we used
```
rustup run nightly cargo fmt
```
for better fmt

lengrongfu:
> ```
> rustup run nightly cargo fmt
> ```
thanks ~, I noticed that we have a pre-commit function. Can we add these basic operations to pre-commit?

slin1237:
> > ```
> > rustup run nightly cargo fmt
> > ```
> 
> thanks ~, I noticed that we have a pre-commit function. Can we add these basic operations to pre-commit?

we should, and we tried
current router requires protobuf, it has a lot of dependencies and we paused the work, but we 100% should

slin1237 (approved):

PR #12162, opened 2025-10-26
[Feature] Enable return routed experts

ocss884:
## Motivation

As per the request from the RL community, this PR enables sglang to return routed experts (topk) during fwd for later usage in training phase. Thanks the MiMo team for proprosing [R3](https://arxiv.org/abs/2510.11370v1) to help stablizing MoE Reinforcement Learning. This method has been used in [MiMo-V2-Flash](https://github.com/XiaomiMiMo/MiMo-V2-Flash/blob/main/paper.pdf) and [DeepSeek-V3.2](huggingface.co/deepseek-ai/DeepSeek-V3.2/resolve/main/assets/paper.pdf). We also notice that this methods effectively reduce the logprob diff between training and inference output.

TODO LIST
- [x] enable cuda graph
- [x] setup experts capture buffer
- [x] Test on dpskV3
- [x] Add unit test
- [x] use tobytes to improve performance

This PR does not compatible with:
- [x] deepep usage (will be in next pr)
- [ ] mtp (will be in next pr)
- [ ] PP (no plan so far) 

## Checklist


ocss884:
@zhaochenyang20 @zhuzilin @yizhang2077 @fzyzcjy

zhaochenyang20:
🐂🍺

fzyzcjy:
seems related: https://github.com/sgl-project/sglang/pull/9499

given this todo list, wondering whether reusing EPLB distribution recorder like in 9499 may be good since cudagraph / ep is already supported in it.

(maybe you can collaborate w/ people in 9499)

image

python/sglang/srt/layers/moe/routed_experts_capturer.py
```diff
@@ -0,0 +1,85 @@
...
+        raise NotImplementedError
+    
+    def get_captured_experts(self):
+        raise NotImplementedError
+
+
+class _RoutedExpertsCapturerReal(RoutedExpertsCapturer):
+    """Capturer for routed experts with host buffer"""
+    def init_buffer(self, max_running_requests: int, model_config: ModelConfig):
+        global _experts_capturer_host_buffer
```
fzyzcjy: tiny nit: good to avoid global var

fzyzcjy:
offline synced w/ @ocss884, notes

image
image
image

python/sglang/srt/layers/moe/fused_moe_triton/layer.py
```diff
@@ -834,7 +841,11 @@ def forward(self, hidden_states: torch.Tensor, topk_output: TopKOutput, **kwargs
         dispatch_output = self.dispatcher.dispatch(
             hidden_states=hidden_states, topk_output=topk_output
         )
-
+        self.routed_experts_capturer.capture(
```
HugoZHL: Here the variable "topk_output" is the input into this module, not the output; why do we capture here rather than the place where we first get "topk_output"?

python/sglang/srt/layers/moe/routed_experts_capturer.py
```diff
@@ -0,0 +1,85 @@
...
+                device="cpu",
+            )
+            logger.debug(
+                f"Initialized routed experts capturer host buffer with shape {_experts_capturer_host_buffer.shape}."
+            )
+            print(f"{_experts_capturer_host_buffer.shape=}")
+
+    def capture(self, layer_id: int, topk_output: StandardTopKOutput):
+        batch_size, num_routed_experts = topk_output.topk_ids.shape
+        _experts_capturer_host_buffer[:batch_size, layer_id, :] = topk_output.topk_ids.cpu()
```
HugoZHL: Direct "cpu" operation is not good for efficiency, especially during model forward. A possible way is to offload these variables together with logprobs, where we can use non_blocking=True

lizipao:
Hi, I tested your PR and found that there are only output routed_experts, but no input routed_experts. I believe that when doing RL, input routed_experts should be needed. Additionally, if this functionality needs to be implemented, we might need to consider sglang's radix_cache.

lizipao:
_experts_capturer_host_buffer is insufficient. It is recommended to change it to dynamic expansion or directly modify it to allow custom configuration via parameters?

lizipao:
Are the routed_experts of layers0 all 0? Is it unnecessary to save them?

zhaochenyang20:
Great job. But lint it please 😂

ocss884:
> Great job. But lint it please 😂

Done

zhaochenyang20:
File "/sglang-checkout/python/sglang/srt/managers/scheduler_output_processor_mixin.py", line 373, in process_batch_result_decode
    .buffer[
     ^^^^^^
AttributeError: 'NoneType' object has no attribute 'buffer'

CI failed @ocss884

ocss884:
> File "/sglang-checkout/python/sglang/srt/managers/scheduler_output_processor_mixin.py", line 373, in process_batch_result_decode .buffer[ ^^^^^^ AttributeError: 'NoneType' object has no attribute 'buffer'
> 
> CI failed @ocss884

Done

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -601,6 +601,7 @@ def __init__(
 
         self.topk = TopK(
             top_k=config.num_experts_per_tok + self.num_fused_shared_experts,
```
lizipao, line 668: deepseek_v2 topk is not match RoutedExpertsHostCache

python/sglang/srt/layers/moe/topk.py
```diff
@@ -886,5 +894,8 @@ def select_experts(
         )
 
     get_global_expert_distribution_recorder().on_select_experts(topk_ids=topk_ids)
-
+    get_global_experts_capturer().capture(
+        layer_id=layer_id,
```
fzyzcjy, line 985: about how to pass in layer_id: ExpertDistributionRecorder has a with_current_layer. what about refactoring into sth like

```
class CurrentLayerManager:
  def with_current_layer(layer_id): ...
  def get current layer id: ...
```

and let both ExpertDistributionRecorder and the new RoutedExpertsCapture to use current_layer_mgr.get_the_current_layer_id_value

python/sglang/srt/managers/scheduler.py
```diff
@@ -685,6 +690,17 @@ def init_deterministic_inference_config(self):
...
 
+    def init_routed_experts_capturer(self):
+        set_global_experts_capturer(
+            RoutedExpertsCapturer.create(
+                get_global_server_args().enable_return_routed_experts,
+                self.model_config,
+                self.max_total_num_tokens + self.page_size,
+                self.max_running_requests,
+                self.device,
+            )
```
fzyzcjy: nit: use keyword args

python/sglang/srt/managers/scheduler_output_processor_mixin.py
```diff
@@ -105,6 +106,20 @@ def process_batch_result_prefill(
...
+                            req.routed_experts = (
+                                get_global_experts_capturer()
+                                .get_host_cache()
+                                .buffer[
+                                    self.req_to_token_pool.req_to_token[
+                                        req.req_pool_idx
+                                    ][: len(req.fill_ids)].cpu()
+                                ]
+                                .tolist()
+                            )
```
fzyzcjy: maybe put this func into GlobalExpertsCapturer, say, GlobalExpertsCapturer.get_whatever(req_to_token, ...)

python/sglang/srt/managers/scheduler_output_processor_mixin.py
```diff
@@ -326,6 +341,20 @@ def process_batch_result_decode(
...
+                    req.routed_experts = (
+                        get_global_experts_capturer()
+                        .get_host_cache()
+                        .buffer[
+                            self.req_to_token_pool.req_to_token[req.req_pool_idx][
+                                : len(req.output_ids) + len(req.fill_ids)
+                            ].cpu()
+                        ]
+                        .tolist()
+                    )
```
fzyzcjy: ... then we can avoid this dup

python/sglang/srt/layers/moe/routed_experts_capturer.py
```diff
@@ -0,0 +1,180 @@
...
+
+GB = 1024 * 1024 * 1024
+MB = 1024 * 1024
+
+
+def get_tensor_size_bytes(t: torch.Tensor):
+    return np.prod(t.shape) * t.dtype.itemsize
+
+
+class RoutedExpertsDeviceCache:
```
fzyzcjy:
```suggestion
class _RoutedExpertsDeviceCache:
```

python/sglang/srt/layers/moe/routed_experts_capturer.py
```diff
@@ -0,0 +1,180 @@
...
+
+    def _finalize_allocation_log(self):
+        """Common logging and memory usage computation for captured experts buffers."""
+        buffer_size_MB = self.get_buffer_size_bytes() / MB
+        logger.info(
+            f"Routing experts device buffer allocated. #shape: {tuple(self.buffer.shape)}, size: {buffer_size_MB:.2f} MB"
+        )
+
+
+class RoutedExpertsHostCache:
```
fzyzcjy:
```suggestion
class _RoutedExpertsHostCache:
```

python/sglang/srt/layers/moe/routed_experts_capturer.py
```diff
@@ -0,0 +1,180 @@
...
+        self,
+        model_config: ModelConfig,
+        num_tokens: int,
+    ) -> None:
+        self.num_tokens = num_tokens
+        self.buffer = torch.zeros(
+            (
+                num_tokens,
+                model_config.hf_text_config.num_hidden_layers,
+                model_config.hf_text_config.num_experts_per_tok,
```
fzyzcjy: does this name must exist for all models? or maybe we need some ways to get the correct field

python/sglang/srt/layers/moe/routed_experts_capturer.py
```diff
@@ -0,0 +1,180 @@
...
+# from sglang.srt.constants import GPU_MEMORY_TYPE_KV_CACHE
+
+logger = logging.getLogger(__name__)
+
+GB = 1024 * 1024 * 1024
+MB = 1024 * 1024
+
+
+def get_tensor_size_bytes(t: torch.Tensor):
+    return np.prod(t.shape) * t.dtype.itemsize
```
fzyzcjy: nit: make all private (`_GB = ...` etc)

fzyzcjy (commented):
LGTM w/ nits

python/sglang/srt/managers/scheduler_output_processor_mixin.py
```diff
@@ -105,6 +106,20 @@ def process_batch_result_prefill(
...
+                            self.server_args.enable_return_routed_experts
+                            and req.return_routed_experts
+                        ):
+                            req.routed_experts = (
+                                get_global_experts_capturer()
+                                .get_host_cache()
+                                .buffer[
+                                    self.req_to_token_pool.req_to_token[
+                                        req.req_pool_idx
+                                    ][: len(req.fill_ids)].cpu()
```
lizipao: In my tests, this needs to be changed to len(req.output_ids) + len(req.fill_ids) to get the correct result.

lizipao:
I tested this PR with deepseekv3: if cuda_graph is enabled, the experts corresponding to all output tokens are duplicated.

fzyzcjy:
btw this looks useful for eplb-related anslysis and simulation as well

python/sglang/srt/layers/moe/routed_experts_capturer.py
ocss884: Done

python/sglang/srt/layers/moe/routed_experts_capturer.py
ocss884: Done

python/sglang/srt/managers/scheduler.py
ocss884: Done

python/sglang/srt/managers/scheduler_output_processor_mixin.py
ocss884: done

python/sglang/srt/layers/moe/routed_experts_capturer.py
ocss884: I prefer to unify the processing logic during model_config init. It is just a temp solution here for my test on Qwen (like deepseek V3 arch definitely need to add the shared expert or other model like step3 using moe_top_k).

liuzhenhai93:
sglang_router_replay
 nice work,  I got a sketch to clarify the code logic

liuzhenhai93:
image
It seems that the returned routed experts indexes are all zero for the first generated token. 

here is the details to repeat the error: 
commit id : 31d9509ede43a9d694a0b48db7eebf76a965e4f9 (latest by NOV 13, 25)
code:

`
import os
import sglang as sgl
from transformers import AutoProcessor


def main():

    model_path = "/data/Qwen3-VL-30B-A3B-Instruct"
    vlm = sgl.Engine(model_path=model_path, tp_size=8, attention_backend="fa3", enable_return_routed_experts=True, disable_cuda_graph=True, disable_overlap_schedule=True)

    image_token = processor = AutoProcessor.from_pretrained(model_path).image_token

    prompt = f"What's in this image? \n{image_token}"
    image_url = "https://www.ilankelman.org/stopsigns/australia.jpg"

    tokenizer = AutoTokenizer.from_pretrained(model_path)
    prompt_ids = tokenizer.encode(prompt)

    sampling_params = {
        "temperature": 0.001,
        "max_new_tokens": 512,
    }

    output = vlm.generate(
        prompt=prompt,
        image_data=image_url,
        return_routed_experts=True,
    )

    print("===============================")
    print(f"Prompt: {prompt}")
    print(f"Generated text {type(output)} : {output.keys()} {output['meta_info'].keys()}")
    prompt_tokens = output['meta_info']["prompt_tokens"]
    output_ids = output["output_ids"]
    completion_tokens = output['meta_info']["completion_tokens"]
    routed_experts = output['meta_info']["routed_experts"]
    print(f"prompt_ids {len(prompt_ids)} prompt_tokens {prompt_tokens}, output_ids {len(output_ids)}, completion_tokens {completion_tokens}, routed_experts {len(routed_experts)} ", flush=True)

    
    for (ti, t) in enumerate(routed_experts):
        for (li, layer) in enumerate(t):
            assert len(layer) == len(set(layer)), f"{ti}  {li} {layer}, {prompt_tokens} {len(output_ids)} {routed_experts[ti - 1: ti + 2]}"
                

    vlm.shutdown()


if __name__ == "__main__":
    main()

`

ocss884:
> Is there any planned support for PP? Thank you

Not so far, I will take a look if the community needs it

zhuzilin (approved):

python/sglang/srt/managers/detokenizer_manager.py
```diff
@@ -247,6 +247,16 @@ def handle_batch_token_id_out(self, recv_obj: BatchTokenIDOutput):
             s.sent_offset = len(output_str)
             output_strs.append(incremental_output)
 
+        output_routed_experts = []
+        if recv_obj.output_routed_experts is not None:
+            output_routed_experts = [
+                (
+                    output_routed_experts.tolist()
```
yizhang2077: mark: `tolist` is very slow, try to replace to other api like `tobytes`

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -546,6 +555,40 @@ def initialize(self, min_per_gpu_memory: float):
...
+
+        if not self.server_args.disable_shared_experts_fusion and hasattr(
+            self.model, "num_fused_shared_experts"
+        ):
+            num_fused_shared_experts = self.model.num_fused_shared_experts
+        else:
+            num_fused_shared_experts = 0
+
+        set_global_experts_capturer(
+            RoutedExpertsCapturer.create(
```
yizhang2077, line 647: mark: I think we only need initiate real routed experts capturer for dp rank = 0

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -1019,6 +1026,7 @@ def reset_for_retract(self):
         self.retraction_count += 1
 
         self.prefix_indices = torch.empty((0,), dtype=torch.int64)
+        self.routed_experts = []
```
yizhang2077: replace to None here, otherwise it will raise error in `detokenizer_manager` in some corner cases

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -1486,6 +1498,7 @@ def prepare_for_extend(self):
         self.req_pool_indices = req_pool_indices_tensor
         self.orig_seq_lens = orig_seq_lens_tensor
         self.out_cache_loc = out_cache_loc
+        self.out_cache_loc_cpu = out_cache_loc.cpu()
```
yizhang2077: we can not use `.cpu()` here

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -2657,6 +2703,13 @@ def forward(
                 reinit_attn_backend,
                 split_forward_count,
             )
+            # Copy cached routing experts' buffers back to CPU cache
+            get_global_experts_capturer().sync_fwd_experts_buffer_DtoH(
```
yizhang2077: move it into `get_global_experts_capturer().with_forward(forward_batch)`

python/sglang/srt/managers/io_struct.py
```diff
@@ -953,6 +961,9 @@ class BatchTokenIDOutput(
     # Hidden states
     output_hidden_states: List[List[float]]
 
+    # The routed experts for each output token
+    output_routed_experts: List[torch.Tensor]
```
hnyls2002, line 965: Shall we use `torch.Tensor` here, or is it a typo?

hnyls2002 (approved):

python/sglang/srt/managers/io_struct.py
ocss884: Since the captured expert ids (which is a tensor) could be large (seq_len * n_layers * topk) which makes the `tolist()` slow (could be 3s when 32k len with Qwen3-30A3), we send the tensor to detokenizer to prevent blocking scheduler

hnyls2002: ok

python/sglang/srt/managers/detokenizer_manager.py
```diff
@@ -266,8 +267,24 @@ def _decode_batch_token_id_output(self, recv_obj: BatchTokenIDOutput):
 
         return output_strs
 
+    def _extract_routed_experts(self, recv_obj: BatchTokenIDOutput) -> List[List[int]]:
+        output_routed_experts = None
+        if recv_obj.output_routed_experts is not None:
+            output_routed_experts = [
+                (
+                    pybase64.b64encode(output_routed_experts.numpy().tobytes()).decode(
```
merrymercy, line 275: Why do we need to use base64 to encode bytes into string? Can you just use bytes?

PR #12176, opened 2025-10-26
Add env var to control custom Triton kernel cache and set CSGMV as default backend. 

lifuhuang:
Redo reverted change https://github.com/sgl-project/sglang/pull/11735 

It was reverted as one of the per optimizations introduced along with CSGMV (custom triton kernel cache) broke AMD CI because the optimization depends on an internal triton interface that varies between triton versions and backends.

To unblock CSGMV rollout & to ensure compatibility across different triton versions / backends (e.g., AMD), I put this optimization behind a *default-off* env var **SGLANG_USE_CUSTOM_TRITON_KERNEL_CACHE** for now.

In the future, we need to do more benchmarking across different triton versions / backends and decide whether we should keep this optimization optional (TODO: and add document for it) or we could enable it based on particular Triton versions.

### DIFF
--- a/python/sglang/srt/environ.py
+++ b/python/sglang/srt/environ.py
@@ -180,6 +180,7 @@ class Envs:
 
     # Triton
     SGLANG_TRITON_DECODE_ATTN_STATIC_KV_SPLITS = EnvBool(False)
+    SGLANG_USE_CUSTOM_TRITON_KERNEL_CACHE = EnvBool(False)
 
     # Torch Compile
     SGLANG_ENABLE_TORCH_COMPILE = EnvBool(False)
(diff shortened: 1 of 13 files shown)


Fridge003 (approved):

python/sglang/srt/environ.py
```diff
@@ -180,6 +180,7 @@ class Envs:
 
     # Triton
     SGLANG_TRITON_DECODE_ATTN_STATIC_KV_SPLITS = EnvBool(False)
+    SGLANG_USE_CUSTOM_TRITON_KERNEL_CACHE = EnvBool(False)
```
Fridge003, line 183: How about setting this variable to True when cuda is enabled?

lifuhuang: Yeah, I was thinking the same, but perhaps let me do more testing and add it in a separate PR to avoid coupling it with the CSGMV.

My worry is that it might not work for some triton versions because we are taking dependency on its non-public API.

PR #12182, opened 2025-10-27
[CI] Add Llama 3.1 8B FP4 to B200 CI

b8zhong:
There is FP4 coverage but only through MLA, so cover it through MHA case as well.

```
Accuracy: 0.620
Invalid: 0.000
Latency: 2.518 s
Output throughput: 3138.540 token/s
{'accuracy': np.float64(0.62), 'invalid': np.float64(0.0), 'latency': 2.518050184000458, 'output_throughput': 3138.539513713902}
ok

Ran 1 test in 33.638s

OK
```

Out of 8 runs, the throughput average is around 3100, so I set it for 3000 to avoid flaky cases in CI.

`hyperfine --runs 8 'export CUDA_VISIBLE_DEVICES=0; python3 test/srt/test_llama31_fp4.py -v'`
```
Benchmark 1: export CUDA_VISIBLE_DEVICES=0; python3 test/srt/test_llama31_fp4.py -v
  Time (mean ± σ):     39.721 s ±  0.101 s    [User: 40.927 s, System: 4.473 s]
  Range (min … max):   39.489 s … 39.799 s    8 runs
```

The time is w/ page cache + CUDA graph already captured, so it may be a bit longer + the first time the CI runner might have to download the weights or snth.


test/srt/test_llama31_fp4.py
```diff
@@ -0,0 +1,56 @@
...
+    @classmethod
+    def setUpClass(cls):
+        cls.model = MODEL_PATH
+        cls.base_url = DEFAULT_URL_FOR_TEST
+        other_args = [
+            "--trust-remote-code",
+            "--mem-fraction-static",
+            "0.8",
+            "--quantization",
+            "modelopt_fp4",
```
Fridge003, line 27: Is llama31FP4 using triton or trtllm_mha as attention backend?

b8zhong: @Fridge003 It uses flashinfer for default, do you want to use trtllm mha?

Fridge003: We will test its behavior on default backend. So the test needn't be changed.

Fridge003: If trtllm mha is better than flashinfer, we can open another pr that change the default backend

b8zhong: sounds good.

b8zhong:
It looks like the B200 CI is kind of flaky. Maybe some other jobs are running on the same device... ? (I saw the throughput go from 800 -> 550 etc), so I remove the throughput number for now

b8zhong:
Looks like it passed here:

https://github.com/sgl-project/sglang/actions/runs/18900055447/job/53947914638?pr=12182

Fridge003 (approved):

Issue #11986, opened 2025-10-23
[router] 0.2.2 release


slin1237:
#### Bug Fixes
- #11945 
- #11887 
- #11882 
- #11886 
- #11935 
- #11826 
- #12152 

#### Features
- #11926 
- #11921 
- #11955 
- #11723 
- #12097 
- #12019 
- #11924 
- #12218 
- #12185 
- #12153 

#### Enhancement, Code Clean Up
- #11790 
- #12155 
- #12084 
- #12086 
- #12089 
- #12030 
- #12021 
- #12024 
- #12091 
- #12147 
- #12164 
- #12168 
- #12171 
- #11734 
- #11767 
- #11988 
- #12096 
- #12159 

#### Responses API
- [ ] Hosted WebSearchTool Integration when WebSearchTool in input
- [ ] WebSearch/MCP Streaming handling
- [ ] Structure outputs, functions calls and bulk inputs
- [ ] Native GPT OSS responses API support via gRPC router with harmony
- [ ] UserID support on chat history storage for auth

PR #12185, opened 2025-10-27
[router] use safety_identifier replace user on chat history storage

lengrongfu:
## Motivation

https://github.com/sgl-project/sglang/issues/11986 UserID support on chat history storage for auth.

## Checklist


sgl-router/src/data_connector/memory.rs
```diff
@@ -416,7 +416,7 @@ impl ResponseStorage for MemoryResponseStorage {
     ) -> ResponseResult<Vec<StoredResponse>> {
         let store = self.store.read();
 
-        if let Some(user_response_ids) = store.user_index.get(user) {
+        if let Some(user_response_ids) = store.identifier_index.get(user) {
```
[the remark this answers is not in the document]
lengrongfu: Can these function names be modified?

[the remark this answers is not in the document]
key4ng: yes

key4ng:
could we add logic to add the safety_identifier column into existing table if the column is not existed
https://github.com/sgl-project/sglang/blob/ef06a714d5786e6a8a72f0ab321c33b67aaefb01/sgl-router/src/data_connector/oracle.rs#L789-L791

key4ng:
Curious if conversation and conversation_item tables also need this for auth control

lengrongfu:
> could we add logic to add the safety_identifier column into existing table if the column is not existed
> 
> https://github.com/sgl-project/sglang/blob/ef06a714d5786e6a8a72f0ab321c33b67aaefb01/sgl-router/src/data_connector/oracle.rs#L789-L791

So can the previous user field be deleted?

lengrongfu:
> Curious if conversation and conversation_item tables also need this for auth control

Current `user_id` in response table only in `list_user_responses` this api to use, however, this method has not been called, so it should not be implemented in the business yet.
If in `conversation` table need to search current conversation history data, maybe it is also necessary.
But if we need, i can commit a new pr to implement, and add a new api like `list_user_conversation` in ConversationStorage trait.

sgl-router/src/protocols/responses.rs
```diff
@@ -527,10 +531,6 @@ pub struct ResponsesRequest {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub truncation: Option<Truncation>,
 
-    /// User identifier
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub user: Option<String>,
```
slin1237, line 546: dont delete this yet

slin1237 (changes requested):

sgl-router/src/protocols/responses.rs
lengrongfu: Do we need to keep this field in the entire path from API to DB or do we only need to keep the field in DB?

slin1237: just keep this in the spec
in DB we use safety identifier instead of user

lengrongfu:
@slin1237 @key4ng Hi, please again review it. thanks ~

sgl-router/src/routers/openai/responses.rs
```diff
@@ -146,9 +144,16 @@ pub(super) fn patch_streaming_response_json(
...
         }
 
-        if obj.get("user").map(|v| v.is_null()).unwrap_or(false) {
-            if let Some(user) = &original_body.user {
-                obj.insert("user".to_string(), Value::String(user.clone()));
+        if obj
+            .get("safety_identifier")
+            .map(|v| v.is_null())
+            .unwrap_or(false)
+        {
```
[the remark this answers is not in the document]
key4ng: This comment makes sense. I'm assuming the logic is we don't send the safety_identifier to the oai service backend and add it back when compose the final response

sgl-router/src/protocols/responses.rs
key4ng: seems this was still deleted

sgl-router/src/protocols/responses.rs
```diff
@@ -838,9 +838,9 @@ pub struct ResponsesResponse {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub usage: Option<ResponsesUsage>,
 
-    /// User identifier
```
slin1237, line 856: same here
dont delete user
and add safety identifier

slin1237 (changes requested):
everything else lgtm
please address the last one

slin1237 (approved):

sgl-router/src/routers/openai/responses.rs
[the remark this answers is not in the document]
key4ng: checked it will pass to backend. it should be fine

key4ng (approved):

Issue #10341, opened 2025-09-11
Roadmap: SGLang Router


slin1237:
## High Level

<img width="1981" height="692" alt="Image" src="https://github.com/user-attachments/assets/c6cb61ff-03f7-4b92-8a8c-470ad393c6cd" />

## Multi-Model Support + Model-Aware Load Balancing
- **Multi-model support**: one router instance handles many model families (e.g., LLaMA, Mistral, Qwen, VLMs).
- **Serving modes**: support both **Monolithic** (single pool) and **PD-disaggregation** (prefill + decode pools).
- **Model-aware selection**: choose models based on **priority, cost, input complexity, and capabilities**.
- **Complement existing load balancers**: model-aware selection runs **before** worker selection and works with cache aware, power of two, round robin, and random
  - [x] Worker registry that allows different lookup
  - [x] embedding support
  - [x] rerank support @tao12345666333 
  - [x] Policy registry allowing multiple radix trees and different policies for different model family
  - [x] Stateless router manager which manages all type of routers
  - [ ] PII detection plugin and native PII detection component/endpoint
  - [ ] Semantic Router for model awareness load balancing (assigned @tao12345666333 
  - [x] OpenAI endpoint routers supporting chat completion and embedding
  - [ ] Model priority and cost selection

## Lightweight State Layer + gRPC Mesh (HA, no external DB)
- **HA without Redis/etcd**: run multiple router replicas with a lightweight in-memory state layer; avoid external databases.  
- **State to manage**: rate-limit buckets, tenant/key **radix tree**, model registry, and **worker states** (per-model, per-mode).  
- **Worker management API (CRUD)**: treat each model backend as a **worker**; expose create/update/delete/list for workers.  
- **Naive gRPC mesh for sync**: replicas exchange deltas/snapshots over gRPC to keep state converged across the fleet.  
- **Preserve LB accuracy**: mesh keeps the radix tree and worker load/health info aligned so selections remain correct under multi-replica deployments.  
- **Eventual consistency with fast convergence**: prioritize low overhead and quick catch-up; no single replica is required for forward progress.  
  - [x] Worker management API
  - [x] Deprecate legacy `add_worker` `remove_worker` and `list_worker` APIs and use unified `/workers` API
  - [x] Service Discovery with new worker management APIs and worker registry
  - [ ] Data mesh component @tonyluj 
  - [ ] Data synchronization for worker registry among router replicas
  - [ ] Data synchronization for radix trees and policies among router replicas

[...]

PR #12202, opened 2025-10-27
[Router][Feat]get engine's load, limit request by load

jimmy-evo:
## Motivation

aware of engine load

if worker's load too high, remove it from available.

router should get worker's real load instead of "request number"

Implemented for HTTP router rate limit.

#10341 

## Checklist


slin1237:
we already have load monitor
and that is currently disabled unless po2 is enabled
we can turn it on for properly rate limiting, i dont think we should introduce those clis tho
rate limiter should be more adaptive rather than fixed value, 
say if we add those
- and should we use that for single router? 
- what about multi mode case
- what if underline hardware changes for the worker
secondly, rate limiter needs some refactoring to support multi tenant and multi worker use case too

jimmy-evo:
> we already have load monitor
> and that is currently disabled unless po2 is enabled
> we can turn it on for properly rate limiting, i dont think we should introduce those clis tho
> rate limiter should be more adaptive rather than fixed value, 
> say if we add those
> - and should we use that for single router? 
> - what about multi mode case
> - what if underline hardware changes for the worker
> secondly, rate limiter needs some refactoring to support multi tenant and multi worker use case too


@slin1237 hi, the  latest code is  configuring load limit by cli arg or annotations in k8s

currently,  load is number of req,  which is not the real load for llm services. the real load is running  req and tokens. 

we need to limit decode  server 's running-req for tps
limit decode server 's token num for preallocate(ttft)
limit prefill server 's waiting queue for ttft

as for regular workers it also make sense.

sgl-router/src/core/worker.rs
```diff
@@ -238,6 +240,17 @@ pub trait Worker: Send + Sync + fmt::Debug {
...
+    fn engine_load(&self) -> &EngineLoad;
+
+    /// Perform an async health check on the worker
+    async fn check_engine_load_async(&self) -> WorkerResult<()>;
+
+    /// HTTP engine load check
+    async fn http_get_load(&self) -> WorkerResult<()>;
+
+    /// gRPC engine load check
+    async fn grpc_get_load(&self) -> WorkerResult<()>;
```
slin1237, line 279: gRPC doesn't have load endpoint right now. please remove

slin1237 (changes requested):
Overall, model gateway needs a centralized and comprehensive understanding of worker loads, not just number of request, but also number of running request, kvcache usage, token capacity etc. 

writing another watcher isn't ideal as we already have load monitor which passively checking the engine load, not request count. Let alone adding more cli arguments, each cli argument must be carefully considered. 

if we look at this model gateway at a higher level
we need to
1. Obtain engine level metrics
2. Leverage those metrics for rate limiting and load balancing, routing, and also use them for circuit breaker if we 100% like to, because hardcoded value doesn't 100% reflect system threshold at every given time 
3. We need to use those metrics for different type of router, we have router that handles tokens in and tokens out, we have router that handles pure http proxy, how do we do that. 
4. When the router is serving multiple tenants and multiple model families. does existing rate limit, passively checking, and cli support that?

PR #12205, opened 2025-10-27
[Detokenizer Manager] Cleanup state when reqs are finished

Muqi1029:
## Motivation

When a request containing an `rid` field is sent to the SGLang engine multiple times, the resulting `output_strs` includes a few trailing tokens (usually up to 5), which are meaningless.

The root cause is that after a request is completed, the `DetokenizerManager` still retains the state associated with that `rid`. This can be seen more clearly in the figure below:

image

As shown in the second run, the previous state still exists. This causes the initial `surr_ids` of the new run to be treated as normal text, even though they actually come from the `input_ids`.

## Modifications

Clean up the retained state in `DetokenizerManager` once the corresponding request has finished.

## Checklist

### DIFF
--- a/python/sglang/srt/managers/detokenizer_manager.py
+++ b/python/sglang/srt/managers/detokenizer_manager.py
@@ -235,6 +235,8 @@ class DetokenizerManager(MultiHttpWorkerDetokenizerMixin):
                     new_text = ""
                 else:
                     new_text = find_printable_text(new_text)
+            else:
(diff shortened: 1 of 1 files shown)


hnyls2002 (approved):
I totally agree with this PR, and we can merge it first.

- `surr_ids` is used for incremental decoding, which is temporarily removed. Maybe we can add it back now.
- Two requests with the same `rid` are not standard.

PR #12216, opened 2025-10-27
[Ckpt Engine] feat: new sglang entrypoint support for update

stmatengss:
## Motivation

Add a new entrypoint for ease of use, as suggested in https://github.com/sgl-project/sglang/pull/11755#issuecomment-3445637425.
With this PR, it can support both torchrun and sglang entrypoint.

Usage
```
python -m sglang.srt.checkpoint_engine.update --update-method broadcast --checkpoint-path /workspace/Qwen/Qwen3-4B/  --inference-parallel-size 2
```

## Checklist


python/sglang/srt/checkpoint_engine/__init__.py
```diff
@@ -0,0 +1,9 @@
+"""
+Checkpoint engine module for SGLang.
+
+This module provides functionality for updating model weights via checkpoint engine.
+"""
+
+from .update import main
```
ShangmingCai: Maybe we should avoid relative import in SGLang?

docs/advanced_features/checkpoint_engine.md
```diff
@@ -0,0 +1,201 @@
...
+    --inference-parallel-size 8
+```
+
+Using integrated entry point (automatically handles torchrun):
+```bash
+python -m sglang.srt.checkpoint_engine.update \
+    --update-method broadcast \
+    --checkpoint-path /path/to/Qwen/Qwen3-8B/ \
+    --inference-parallel-size 8
+```
```
ShangmingCai: I think `checkpoint_engine` is the recommended one, which is more aligned with the launch server usage, and should be placed at the front.

python/sglang/srt/checkpoint_engine/update.py
```diff
@@ -0,0 +1,317 @@
...
+        sys.exit(result.returncode)
+    except FileNotFoundError:
+        print(
+            "Error: torchrun command not found. Please ensure PyTorch is installed.",
+            file=sys.stderr,
+        )
+        sys.exit(1)
+    except KeyboardInterrupt:
+        print("\nInterrupted by user", file=sys.stderr)
+        sys.exit(130)
```
ShangmingCai, line 237: Is `torchrun` compulsory?

ShangmingCai (commented):
LGTM. But wonder if we have a better solution to replace `cmd` and `torchrun` to make the code less hacky. Will the checkpoint-engine team release a generic entrypoint?

python/sglang/srt/checkpoint_engine/update.py
stmatengss: If we use ParameterServer in the checkpoint engine, torchrun is compulsory.

docs/advanced_features/checkpoint_engine.md
stmatengss: fix it

ShangmingCai (approved):
LGTM

PR #12225, opened 2025-10-27
[NVIDIA] upstream FA4 12/15/25

johnnynunez:
Upstream FA4 from 12/15/25

1. Added num_splits_heuristic function (lines 39-46)
Heuristic to determine the number of splits for split KV computation
2. Added new parameters to _flash_attn_fwd
num_splits: int = 1 (line 70)
mask_mod: Optional[Callable] = None (line 74)
3. Updated window_size handling logic (lines 225-236)
Supports mask_mod parameter
Handles causal/local window logic with mask_mod
4. Added split KV logic (lines 237-260)
Computes is_split_kv based on num_splits
Creates out_partial and lse_partial tensors when split KV is enabled
Automatic split calculation when num_splits < 1
5. Updated compile key (lines 350-376)
Added score_mod_hash and mask_mod_hash
Added is_split_kv flag
Added paged_kv_non_tma flag
Added buffer count instead of aux_tensors count
6. Updated kernel initialization calls
FlashAttentionForwardSm90 (lines 382-401): Added mask_mod, intra_wg_overlap=True, mma_pv_is_rs=True
FlashAttentionForwardSm100 (lines 407-428): Added mask_mod, is_split_kv, paged_kv_non_tma, is_varlen_q, m_block_size, n_block_size
7. Added _flash_attn_fwd_combine function (lines 486-616)
Combines partial outputs from split KV computation
Includes caching logic
8. Added combine call (lines 473-481)
Calls _flash_attn_fwd_combine when is_split_kv is True
9. Added import (line 25)
from flash_attn_origin.cute.flash_fwd_combine import FlashAttentionForwardCombine


Fridge003:
@johnnynunez Can you post result of running `sgl-kernel/tests/test_flash_attention_4.py` after kernel upgrade?
Do we need to add any new tests?

johnnynunez:
> @johnnynunez Can you post result of running `sgl-kernel/tests/test_flash_attention_4.py` after kernel upgrade? Do we need to add any new tests?

surely, but i am trying to be confident with the actual tests due that i don't have blackwell to test internally.

PR #12226, opened 2025-10-27
Forward unknown tool calls instead of dropping 

Surya-Gunukula:
Closes #12223

## Motivation

- Avoid silent data loss when models emit unknown tool names.
- Make behavior consistent with paths that already forward malformed arguments (no schema validation).
- Let clients/orchestrators decide how to handle unknown tools instead of the server dropping them.

## Modifications

In BaseFormatDetector.parse_base_json:
- Collapse the if/else and always append a ToolCallItem; log a warning if the tool name is unknown.
- No schema-validation changes; no changes to other detectors.
- Passed all pre-commit checks 

## Accuracy Tests

Functional check with OpenAI-compatible chat completions:
- Given tools=[...] and model output calling a non-declared function, previously the response had no tool_calls.
- After this change, the response includes a tool_calls entry with name=<unknown>, index=None/-1, and serialized arguments; normal text unchanged.
- Declared/valid tool calls remain unaffected.

## Checklist


JustinTong0323 (commented):
It would affect downstream task and break some assumptions, may you add a env variable for whether enable this feature?

Surya-Gunukula:
> It would affect downstream task and break some assumptions, may you add a env variable for whether enable this feature?

Hi,

Thanks for the feedback!

Although, I've been using sglang for model serving for a while, this is my first time writing a PR for it.

I'm relatively unknowledgeable on the topic, but could you explain what downstream assumptions are being made based on this? Shouldn't false tool calls be on the client to deal with, not the server?

JustinTong0323:
> Thanks for the feedback!
Although, I've been using sglang for model serving for a while, this is my first time writing a PR for it.
I'm relatively unknowledgeable on the topic, but could you explain what downstream assumptions are being made based on this? Shouldn't false tool calls be on the client to deal with, not the server?

So if others are using sglang's oai tool call parser and don't use some try-catch to deal with unknown tool name (as they assume sglang would did it for them), it may cause some crash or other issues. So the best practice here is not change server
's behaviour.

JustinTong0323:
So for your use case, set a env var as a switch is better, in my understanding. Could you help modify that and add related docs? Thanks!

zkx06111:
> > Thanks for the feedback!
> > Although, I've been using sglang for model serving for a while, this is my first time writing a PR for it.
> > I'm relatively unknowledgeable on the topic, but could you explain what downstream assumptions are being made based on this? Shouldn't false tool calls be on the client to deal with, not the server?
> 
> So if others are using sglang's oai tool call parser and don't use some try-catch to deal with unknown tool name (as they assume sglang would did it for them), it may cause some crash or other issue. So the best practice here is not change server 's behaviour.

One weird thing about this assumption is that it might result in no response from the model when it calls a non-existent tool, and the user doesn't get notified about it (the warning only gets printed on the server). So it will look like nothing has happened, and there is no way for a user to know what to do to deal with it. I feel at least the user should be notified about this.

JustinTong0323:
> > > Thanks for the feedback!
> > > Although, I've been using sglang for model serving for a while, this is my first time writing a PR for it.
> > > I'm relatively unknowledgeable on the topic, but could you explain what downstream assumptions are being made based on this? Shouldn't false tool calls be on the client to deal with, not the server?
> > 
> > 
> > So if others are using sglang's oai tool call parser and don't use some try-catch to deal with unknown tool name (as they assume sglang would did it for them), it may cause some crash or other issue. So the best practice here is not change server 's behaviour.
> 
> One weird thing about this assumption is that it might result in no response from the model when it calls a non-existent tool, and the user doesn't get notified about it (the warning only gets printed on the server). So it will look like nothing has happened, and there is no way for a user to know what to do to deal with it. I feel at least the user should be notified about this.

I see, you are correct... But just afraid of that... May we still put a env var to control here but set to True by default? So if downstream meet some issue it could be resolved by setting that switch to False... And may deprecated the original behavior in the future?

Surya-Gunukula:
> > > > Thanks for the feedback!
> > > > Although, I've been using sglang for model serving for a while, this is my first time writing a PR for it.
> > > > I'm relatively unknowledgeable on the topic, but could you explain what downstream assumptions are being made based on this? Shouldn't false tool calls be on the client to deal with, not the server?
> > > 
> > > 
> > > So if others are using sglang's oai tool call parser and don't use some try-catch to deal with unknown tool name (as they assume sglang would did it for them), it may cause some crash or other issue. So the best practice here is not change server 's behaviour.
> > 
> > 
> > One weird thing about this assumption is that it might result in no response from the model when it calls a non-existent tool, and the user doesn't get notified about it (the warning only gets printed on the server). So it will look like nothing has happened, and there is no way for a user to know what to do to deal with it. I feel at least the user should be notified about this.
> 
> I see, you are correct... But just afraid of that... May we still put a env var to control here but set to True by default? So if downstream meet some issue it could be resolved by setting that switch to False... And may deprecated the original behavior in the future?

@CatherineSue

CatherineSue:
Agree with @JustinTong0323 . Having an env var. And make sure this is consistent in all parsers is necessary. The default behavior should not be changed.

Surya-Gunukula:
> Agree with @JustinTong0323 . Having an env var. And make sure this is consistent in all parsers is necessary. The default behavior should not be changed.

@JustinTong0323 @CatherineSue Thanks for the quick feedback!

 I updated my branch to include a new env var called SGLANG_FORWARD_UNKNOWN_TOOLS. I modified all parsers that have this dropped tool logic instead of just the base_parser. I also made sure that the default behavior has not changed.

python/sglang/srt/function_call/gpt_oss_detector.py
```diff
@@ -220,7 +221,8 @@ def _extract_tool_call_from_event(
         # Check if tool exists
         if function_name not in tool_indices:
             logger.debug(f"Function {function_name} not in available tools")
-            return None
+            if os.getenv("SGLANG_FORWARD_UNKNOWN_TOOLS") != "TRUE":
```
JustinTong0323: Follow the usage in https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/environ.py

JustinTong0323 (changes requested):

Surya-Gunukula:
@JustinTong0323 @CatherineSue edits have been made

python/sglang/srt/function_call/base_format_detector.py
```diff
@@ -78,7 +78,7 @@ def parse_base_json(self, action: Any, tools: List[Tool]) -> List[ToolCallItem]:
             name = act.get("name")
             if not (name and name in tool_indices):
                 logger.warning(f"Model attempted to call undefined function: {name}")
-                if os.getenv("SGLANG_FORWARD_UNKNOWN_TOOLS") != "TRUE":
+                if not envs.SGLANG_FORWARD_UNKNOWN_TOOLS.value:
```
JustinTong0323: use SOME_ENV.get()

JustinTong0323 (approved):

Surya-Gunukula:
@CatherineSue

test/srt/function_call/test_unknown_tool_name.py
```diff
@@ -0,0 +1,69 @@
...
+        return True
+
+    def detect_and_parse(self, text: str, tools):
+        action = json.loads(text)
+        return StreamingParseResult(
+            normal_text="", calls=self.parse_base_json(action, tools)
+        )
+
+
+def test_unknown_tool_name_dropped_default(caplog):
```
hnyls2002, line 21: How does this test work? This file only contains some functions, but never called here.

Surya-Gunukula: Yeah that was my mistake. It looks like your refactor fixed these issues?

PR #12238, opened 2025-10-28
Reduce docker image size. mount cache when use pip/cargo build

whybeyoung:
When building an image with Docker, for installations via pip and cargo， and  apt , you should use mount cache to prevent the built image from becoming excessively large.
CC @zhyncs  @slin1237


docker/Dockerfile
```diff
@@ -249,11 +249,11 @@ RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y \
...
 
 # Build and install sgl-router
-RUN python3 -m pip install --no-cache-dir setuptools-rust \
+RUN  --mount=type=cache,target=/root/.cache/pip --mount=type=cache,target=/root/.cargo/registry  --mount=type=cache,target=/root/.cargo/.package_cache python3 -m pip install setuptools-rust \
     && cd /sgl-workspace/sglang/sgl-router \
     && cargo build --release \
-    && python3 -m pip install --no-cache-dir . \
-    && rm -rf /root/.cache
+    && python3 -m pip install . \
+    && cd /usr/local/src/sglang/sgl-router/target/release &&  rm -rf build/ deps/ examples/ incremental/
```
slin1237: we should probably just delete entire router code base?
is this part of dev image? (dev image should do its own copy)
production image should not have original code base unless the language requires it (python)

fzyzcjy:
general idea LGTM

jimmy-evo:
LGTM

docker/Dockerfile
jimmy-evo, line 58: also can mount cache

jimmy-evo: `--mount=type=cache,target=/var/cache/apt` for apt operation, then no need to delete apt cache

slin1237 (approved):

docker/Dockerfile
whybeyoung: good

whybeyoung:
i saw that https://github.com/sgl-project/sglang/pull/12364 has beend merged 

this pr only focus  on whether we should use build cache.

the ci build has passed in : https://github.com/sgl-project/sglang/actions/runs/18961270151

can you give some advice ? @slin1237 @merrymercy

PR #12241, opened 2025-10-28
Add continuous_usage_stats support for streaming responses

BBuf:
This adds support for `continuous_usage_stats` in streaming chat completions and completions endpoints.

When enabled, usage information is included in every content-bearing chunk instead of just the final usage chunk. This matches vLLM's behavior and provides better real-time usage tracking.

Changes:
- Added `continuous_usage_stats` field to `StreamOptions` 
- Updated chat completions streaming to include usage in content chunks
- Updated completions streaming to include usage in content chunks
- Added tests to verify the feature works correctly

The feature is disabled by default to maintain backward compatibility.


JustinTong0323 (approved):
LGTM

PR #12251, opened 2025-10-28
fix(moe): Add global cache reuse to prevent OOM during chunked prefill

liusy58:
## Motivation

This PR addresses OOM issues that occur when processing long inputs when deployed on memory-limited devices with MoE models, particularly during chunked prefill operations. The OOM errors are primarily caused by memory fragmentation from repeatedly creating and destroying intermediate cache tensors for each request.

## Checklist

### DIFF
--- a/python/sglang/srt/layers/moe/fused_moe_triton/fused_moe.py
+++ b/python/sglang/srt/layers/moe/fused_moe_triton/fused_moe.py
@@ -13,6 +13,7 @@ import torch
 import triton.language as tl
 
 from sglang.srt.layers.moe.moe_runner import MoeRunnerConfig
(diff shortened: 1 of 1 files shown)


ShangmingCai:
Thank you for this PR. Please address the conflict.

ch-wan:
The code you modified has been mostly moved to [here](https://github.com/sgl-project/sglang/blob/737b58d6dc63520e6f11c9275dd1761e937f5290/python/sglang/srt/layers/moe/moe_runner/triton.py#L109-L114). How does sglang enter this code script?

moonquekes:
Is it necessary to release the global cache **before the decode** phase? After applying your modifications, I encountered an OOM (out of memory) issue during the decode phase.

PR #12252, opened 2025-10-28
[router] configure workflow retries and timeout based on routerConfig

slin1237:
1. readiness is already context aware
2. pass in routerConfig to workflow to control time out, retries etc

## Checklist

### DIFF
--- a/.github/workflows/pr-test-rust.yml
+++ b/.github/workflows/pr-test-rust.yml
@@ -154,7 +154,7 @@ jobs:
   pytest-rust-2:
     if: github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'run-ci')
     runs-on: 4-gpu-a10
-    timeout-minutes: 16
+    timeout-minutes: 32
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


brunamagrinidacruz:
Hello. I need the feature implemented in this commit of max_retries being configured via parameter. How long does it take to your changes be available in pip https://pypi.org/project/sglang-router/?

Currently, my model takes more than 10 minutes to load, and thus, while the router is trying to discover it via Kubernetes, it reaches the max_retries = 100 and returns an error.

Note: Sorry if here is not the right channel to ask; I wasn't able to find the pip frequency. I did look at the tags, but the one from pip is so behind that I am afraid I am looking in the right place.

slin1237:
> @slin1237 I'm confused by the logic of this PR — I believe it's basically capping the timeout for worker registration to `health_check_timeout + 30` instead of the previous hardcoded 7200s. But health_check_timeout is a request-level timeout value (ie usually very small), which means that this now spends very little time waiting for a worker to be available.
> 
> 
> 
> I hit this issue because the router started running and registering workers before a decode worker was ready and only waited ~30s for the decode worker to respond before giving up on it. The only way I can see to resolve that is to increase health_check_timeout to a very large value (which then means normal health checks also have a really long request-level timeout), or manually check worker-readiness on my end before starting the router.

Thanks for reporting this. 
I missed that blind spot. The original issue caused this PR was hardcoded 7200 wasn't long enough. 
I agree with you, health check time out should not be used like this. we should have proper startup delay etc for properly handling this. Will get a PR out to address this ASAP

PR #12272, opened 2025-10-28
[bugfix] Fix tcp port conflict(zmq.error.ZMQError: Address already in use) when zmq socket is bound to random port

wcsjtu:
Change-Id: I19637d77e6a27af4674a9a7abcfce6f29a874d95

## Motivation

`get_free_port` has concurrency issues because it releases the socket and port when it returns.  This PR just used zmq native inferface `bind_to_random_port` instead of it

```py
def get_free_port():
    # try ipv4
    try:
        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
            s.bind(("", 0))
            return s.getsockname()[1]
    except OSError:
        # try ipv6
        with socket.socket(socket.AF_INET6, socket.SOCK_STREAM) as s:
            s.bind(("", 0))
            return s.getsockname()[1]
``` 

```shell
  File "/sgl-workspace/sglang/python/sglang/srt/disaggregation/nixl/conn.py", line 139, in __init__
    self._start_bootstrap_thread()
  File "/sgl-workspace/sglang/python/sglang/srt/disaggregation/nixl/conn.py", line 650, in _start_bootstrap_thread
    self._bind_server_socket()
  File "/sgl-workspace/sglang/python/sglang/srt/disaggregation/common/conn.py", line 96, in _bind_server_socket
    self.server_socket.bind(format_tcp_address(self.local_ip, self.rank_port))
  File "/usr/local/lib/python3.12/dist-packages/zmq/sugar/socket.py", line 320, in bind
    super().bind(addr)
  File "zmq/backend/cython/_zmq.py", line 1009, in zmq.backend.cython._zmq.Socket.bind
    _check_rc(rc)
    ^^^^^^^^^^^
  File "zmq/backend/cython/_zmq.py", line 190, in zmq.backend.cython._zmq._check_rc
    raise ZMQError(errno)
    ^^^^^^^^^^^
zmq.error.ZMQError: Address already in use (addr='tcp://****:10039')
```

## Checklist

- [x ] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).


jimmy-evo:
same to #11961

jimmy-evo:
BTW, lazy assign cannot resolve the root cause

ShangmingCai:
I think a better solution is still using the `get_free_port() ` to bind a random port, but we add a util function which includes a `while` and` try catch `block to make sure it succeeds.

wcsjtu:
> I think a better solution is still using the `get_free_port() ` to bind a random port, but we add a util function which includes a `while` and`try catch`block to make sure it succeeds.

This PR does not remove `get_free_port`; it simply removes its use for obtaining ZMQ ports. This is because ZMQ does not accept Python socket objects or `fd` as the `bind` argument. This means that if ZMQ wants to use a port obtained with `get_free_port`, it must first release the port before allowing ZMQ to occupy it. However, between the release and occupation, another process might obtain the same port through `get_free_port`, leading to port conflicts.

wcsjtu:
> BTW, lazy assign cannot resolve the root cause

the race condition is unsolvable unless zmq socket can `steal reference` from python socket object.

python/sglang/srt/disaggregation/common/conn.py
```diff
@@ -68,15 +67,14 @@ def __init__(
...
-        self.rank_port = get_free_port()
+        self.rank_port = -1  # lazy assign when bound to random port
         self.local_ip = get_local_ip_auto()
         self.server_socket = zmq.Context().socket(zmq.PULL)
         if is_valid_ipv6_address(self.local_ip):
             self.server_socket.setsockopt(zmq.IPV6, 1)
         self.request_status: Dict[int, KVPoll] = {}
 
         if self.disaggregation_mode == DisaggregationMode.PREFILL:
-            self._register_to_bootstrap()
```
ShangmingCai, line 79: Why do we extract this from the common backend?

python/sglang/srt/disaggregation/common/conn.py
```diff
@@ -68,15 +67,14 @@ def __init__(
         )
         self.pp_size = server_args.pp_size
         self.pp_rank = self.kv_args.pp_rank
-        self.rank_port = get_free_port()
+        self.rank_port = -1  # lazy assign when bound to random port
         self.local_ip = get_local_ip_auto()
         self.server_socket = zmq.Context().socket(zmq.PULL)
```
ShangmingCai: Can we not use lazy bind? Maybe just move the initialization of `self.rank_port` to the next line of `self.server_socket = zmq.Context().socket(zmq.PULL)`, does it work?

wcsjtu: > Can we not use lazy bind? Maybe just move the initialization of `self.rank_port` to the next line of `self.server_socket = zmq.Context().socket(zmq.PULL)`, does it work?

If we can guarantee that all backends will use ZMQ, then we can put `bind` in the `CommonKVManager` class. TCP port resources are scarce.

ShangmingCai: Currently yes, ascend backend is inherited from mooncake backend, and nixl backend also borrows the design of mooncake, they both need ZMQ for metadata transfer.

wcsjtu: The expr `self.register to bootstrap()` is moved to `Common KVManager`, and remove the `lazy assignment` of ZMQ port

python/sglang/srt/disaggregation/mooncake/conn.py
```diff
@@ -851,7 +851,9 @@ def transfer_worker(
                 )
 
     def start_prefill_thread(self):
-        self._bind_server_socket()
+        # self._bind_server_socket()
+        # already bound in CommonKVManager init
+        pass
```
ShangmingCai: No need for the comment and pass.

python/sglang/srt/disaggregation/mooncake/conn.py
```diff
@@ -889,7 +891,9 @@ def bootstrap_thread():
         threading.Thread(target=bootstrap_thread).start()
 
     def start_decode_thread(self):
-        self._bind_server_socket()
+        # self._bind_server_socket()
+        # already bound in CommonKVManager init
+        pass
```
ShangmingCai: ditto

python/sglang/srt/disaggregation/nixl/conn.py
```diff
@@ -647,7 +647,9 @@ def check_transfer_done(self, room: int):
         return self.transfer_statuses[room].is_done()
 
     def _start_bootstrap_thread(self):
-        self._bind_server_socket()
+        # self._bind_server_socket()
+        # already bound in CommonKVManager init
+        pass
```
ShangmingCai: ditto

python/sglang/srt/disaggregation/common/conn.py
```diff
@@ -92,8 +91,11 @@ def __init__(
                 f"Unsupported DisaggregationMode: {self.disaggregation_mode}"
             )
 
-    def _bind_server_socket(self):
-        self.server_socket.bind(format_tcp_address(self.local_ip, self.rank_port))
+    def _bind_server_socket(self) -> int:
+        """auto choose random available port, then bind it and return it"""
+        addr = f"tcp://{maybe_wrap_ipv6_address(self.local_ip)}"
+        port = self.server_socket.bind_to_random_port(addr)
+        return port
```
ShangmingCai: If `get_zmq_socket` is an already existing util, we should use that and remove `_bind_server_socket`.

ShangmingCai (commented):
@jinmingyi1998 Can you sync in 1 PR? If `get_zmq_socket` is an already existing util, we should use that and remove `_bind_server_socket`.
And if using reserve range ports is not a popular requirement among the SGLang users, we should not introduce it in this PR when `bind_to_random_port` is sufficient to address the 'Address already in use' issue.

jimmy-evo:
> @jinmingyi1998 Can you sync in 1 PR? If `get_zmq_socket` is an already existing util, we should use that and remove `_bind_server_socket`. And if using reserve range ports is not a popular requirement among the SGLang users, we should not introduce it in this PR when `bind_to_random_port` is sufficient to address the 'Address already in use' issue.

@ShangmingCai 
yes, `get_zmq_socket` can do this.

And it is recommended to explicitly specify the port range to be used, as this feature supports such configuration.

wcsjtu:
-  I refactor `get_zmq_socket` function in @python/sglang/srt/utils/common.py. 
    - add new keyword argument `port_range` to specifying the range of ports, it may be useful as discussed above.
    - unify the return type. and introduce a new function `get_zmq_socket_port` to get port from zmq socket
- use `get_zmq_socket` to create zmq socket `of KVManager`

wcsjtu:
@ShangmingCai

ShangmingCai:
> @ShangmingCai

@wcsjtu I have recommented @jinmingyi1998 to add you as an co-author of #11961, which try to address the same problem as this PR. I think configurable port range is not a highly required feature, so it should be introduced in another PR when we really need it in the future.

wcsjtu:
> > @ShangmingCai
> 
> @wcsjtu I have recommented @jinmingyi1998 to add you as an co-author of #11961, which try to address the same problem as this PR. I think configurable port range is not a highly required feature, so it should be introduced in another PR when we really need it in the future.

thanks for reply. The main purpose of this pr is to fix the bug of `port conflict` when launching sglang. It's a critical bug which blocks the all other features

ShangmingCai:
Closed as integrated in https://github.com/sgl-project/sglang/pull/11961.

PR #11928, opened 2025-10-21
Use Flashinfer TRT-LLM as Llama 4 compatible MoE backend

b8zhong:
<!-- Thank you for your contribution! Please follow these guidelines to enhance your pull request. If anything is unclear, submit your PR and reach out to maintainers for assistance. Join our Slack community at https://slack.sglang.ai to discuss further. -->

## Motivation

Since the FI TRT-LLM backend was already added for blockscale MoE, add it for per-tensor FP8, specifically for Llama 4.
Only for SM100

<!-- Describe the purpose and goals of this pull request. -->

## Modifications

Modify the apply code of MoE methods.

> Generally I only used ModelOpt checkpoints, so I didn't add the code in the normal FP8 file.
> There is some issues with ModelOpt loading on main, merge this in after https://github.com/sgl-project/sglang/pull/10154

<!-- Detail the changes made in this pull request. -->

## Accuracy Tests

Before:
```
python3 benchmark/gsm8k/bench_sglang.py --num-shots 8 --num-questions 1319 --parallel 200
Downloading from https://raw.githubusercontent.com/openai/grade-school-math/master/grade_school_math/data/test.jsonl to /tmp/test.jsonl
/tmp/test.jsonl: 732kB [00:00, 33.3MB/s]                                                                                                                                                                                                          
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1319/1319 [00:39<00:00, 33.54it/s]
Accuracy: 0.935
Invalid: 0.000
Latency: 39.421 s
Output throughput: 3492.464 token/s
```

After:
I set parallel 50 due to some strange server issue on main w/ overlap scheduler, it's not related to this PR.
```
python3 benchmark/gsm8k/bench_sglang.py --num-shots 8 --num-questions 1319 --parallel 50
100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1319/1319 [01:13<00:00, 17.95it/s]
Accuracy: 0.931
Invalid: 0.000
Latency: 73.587 s
Output throughput: 1864.289 token/s
```

<!-- If this pull request affects model outputs (e.g., changes to the kernel or model forward code), provide accuracy test results. -->

## Benchmarking and Profiling

Before:
```
python -m sglang.launch_server \
[...]

PR #12291, opened 2025-10-28
[MoE] Remove logic that prevents the user from selecting MoE backend for specific models

jonahbernard:
## Motivation

For `DeepseekV3ForCausalLM`, `GptOssForCausalLM`,  and `Llama4`, logic in `server_args.py` overrides the user selected MoE backend. For consistency across models and correct code, the user should select the MoE backend for these models.

This MoE backend selection logic shouldn't be hardcoded because 1) the most performant backend will change over time (and it can change quite fast), 2) a user should be able to choose the backend they want for other reasons, even if it isn't the most performant.

A solution to 1) would be to add the backend recommendation to the cookbook with evidence from benchmarks to show that it was recently the most performant backend. This is easier to maintain and more adaptable.

## Modifications

Removed MoE backend selection logic in `server_args.py` for `DeepseekV3ForCausalLM`, `GptOssForCausalLM`,  and `Llama4`.

## Accuracy Tests

Not applicable.

## Benchmarking and Profiling

Not applicable.

## Checklist

### DIFF
--- a/python/sglang/srt/server_args.py
+++ b/python/sglang/srt/server_args.py
@@ -52,7 +52,6 @@ from sglang.srt.utils.common import (
     is_sm90_supported,
     is_sm100_supported,
     is_sm120_supported,
-    is_triton_kernels_available,
     is_valid_ipv6_address,
     json_list_type,
     nullable_str,
@@ -885,11 +884,6 @@ class ServerArgs:
                     logger.info(
                         "Enable FlashInfer AllReduce Fusion on sm100 for DeepseekV3ForCausalLM"
                     )
-                if self.moe_a2a_backend == "none" and self.moe_runner_backend == "auto":
-                    self.moe_runner_backend = "flashinfer_trtllm"
-                    logger.info(
-                        "Use flashinfer_trtllm as MoE runner backend on sm100 for DeepseekV3ForCausalLM"
-                    )
                     if self.quantization is None:
                         # Default DeepSeek V3/R1 native FP8 when not explicitly set,
                         # Because we need this condition for an assertion in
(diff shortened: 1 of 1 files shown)


b8zhong:
I think `moe_runner_backend = auto` is when it's unset. This is done to pick the most performant backend automatically. 

Unless, it's being done somewhere else

jonahbernard:
> I think `moe_runner_backend = auto` is when it's unset. This is done to pick the most performant backend automatically.
> 
> Unless, it's being done somewhere else

Yeah, I agree with you. My concern is that this logic shouldn't be hardcoded because 1) the most performant backend will change over time (and it can change quite fast), 2) a user should be able to choose the backend they want for other reasons, even if it isn't the most performant.

A solution to 1) would be to add the backend recommendation to the cookbook with evidence from benchmarks to show that it was recently the most performant backend. This is easier to maintain and more adaptable.

b8zhong:
The user can definitely change it though, because it only auto sets it if the user has not selected it. I just tried it

Generally the most performant moe backend is kind of known and doesn't change that significantly. I think out of the box performance is more impportant

b8zhong:
Also I don't think anyone is really using the cookbook right now

jonahbernard:
> The user can definitely change it though, because it only auto sets it if the user has not selected it. I just tried it


Oh yeah, sorry, you are correct for Deepseek and Llama. Although for GPT OSS, when sm100 is supported and it is mxfp4 quant, it looks like it forces it to `flashinfer_mxfp4`

> Generally the most performant moe backend is kind of known and doesn't change that significantly. I think out of the box performance is more impportant

Yeah I'm inclined to agree that out of the box performance is the most important. Do you know where the evidence comes from that, for example, that `flashinfer_trtllm` is most performant for Llama4?

b8zhong:
Sure, bc that one is the only correct moe kernel for mxfp4.

And bc I actually added the llama 4 one in https://github.com/sgl-project/sglang/pull/11928#event-20560346424

jonahbernard:
I think you are right. Closing the PR.

PR #12295, opened 2025-10-28
fix seqlen bug for trtllm_mla's draft_extend

bmac3:
In trtllm_mla's draft extend, we have `seq_lens = forward_batch.seq_lens.to(torch.int32)` which is passed to`flashinfer.decode.trtllm_batch_decode_with_kv_cache_mla`. `forward_batch.seq_lens.to(torch.int32)` does not take into account the padding we do to the queries with `pad_draft_extend_query`. This leads to mismatched q's and kv's when we call `flashinfer.decode.trtllm_batch_decode_with_kv_cache_mla`. We should instead have
```
seq_lens = (
    forward_batch.seq_lens
    - metadata.seq_lens_q
    + metadata.max_seq_len_q
).to(torch.int32)
```
to account for the padding.


zhyncs (approved):

zhyncs:
Thanks Brian!

PR #12300, opened 2025-10-28
[DeepSeekV32] Bug fix to ensure `page_table` and `result` in same type

Johnsonms:
Running transform_index.py in NSA attention path fails at torch.gather(...) with a dtype mismatch: the index/out tensor is int (int32) while PyTorch requires Long (int64) for gather indices (and often for the output when provided).
## Motivation
RuntimeError: Expected out tensor to have dtype long int, but got int instead

image

## Modifications
Unified the page tale type with output type

## Checklist


hlu1 (approved):
Thanks for the fix!

PR #12309, opened 2025-10-29
[CPU] add support for mamba causal conv1d for qwen3-next

mingfeima:
## Motivation

add support for mamba causal conv1d for qwen3-next

## Modifications

add kernel files at `sgl-kernel/csrc/cpu/mamba/conv.cpp`, which implemented both `causal_conv1d_fwd` for prefill and `causal_conv1d_update` for decode.

* APIs align with existing CUDA counterpart
* support both batched input and variant length input
* CPU kernels will requires `x` to be contiguous on the second-to-last dimension to achieve optimal performance
* shift `conv_states` to be contiguous the second-to-last dimension to make sure vectorized load from memory
* implemented with avx512-bf16, applying amx would be pointless since width is 4 (which conresponds to K in tinygemm of amx, it requires to be 32x)
* weight prepacked to be vnni2 format to remove online prepacking overhead

## Accuracy Tests

```bash
python /test/srt/cpu/test_causal_conv1d.py
```

## Benchmarking and Profiling

compare the optimized C++ version against reference implementation with native torch (which will end up with oneDNN) 

performance collected on 6th gen Xeon with 40 cores, with [script](https://github.com/mingfeima/sgl-cpu-tests/blob/main/test_conv1d_fn.py).

```
### batch = 1, dim = 8192, seqlen = 1024
### causal_conv1d: oneDNN ref: 3.205 ms; opt: 0.239 ms

### batch = 128, dim = 8192, seqlen = 1024
### causal_conv1d: oneDNN ref: 406.583 ms; opt: 36.051 ms
```

**NOTE**: the main reason for low performance with torch native implementation is that PyTorch doesn't have `channels last` concept for 1d convolution. On PyTorch, 1d convolution will be mapped to 2d and therefore it will always be channels first. This will trigger: 
* addtional copy (tranpose last 2 dimensions) to make the input contiguous
* input and weight will need to be reordered to onednn internal format to use vnni
* output will be reordered from internal format to plain format 

## Checklist

### DIFF
--- a/sgl-kernel/csrc/cpu/causal_conv1d.cpp
+++ b/sgl-kernel/csrc/cpu/causal_conv1d.cpp
@@ -1,0 +1,469 @@
+#include "common.h"
+#include "gemm.h"
+#include "vec.h"
+
+namespace {
+
+inline void print_32x16(const __m512bh x) {
+  at::BFloat16 a[32];
+  _mm512_storeu_si512((__m512i*)a, (__m512i)x);
+
+  for (int i = 0; i < 32; i++) {
+    std::cout << a[i] << " ";
+  }
+  std::cout << std::endl;
+}
+
+inline void print_16x32(const __m512 x) {
+  float a[16];
+  _mm512_storeu_ps((__m512*)a, (__m512)x);
+
+  for (int i = 0; i < 16; i++) {
+    std::cout << a[i] << " ";
+  }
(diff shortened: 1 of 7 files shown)


mingfeima:
fix new lint error.

FlamingoPg (approved):

PR #12312, opened 2025-10-29
Refactor abortion in event loop

hnyls2002:
Preparation for #12224.

- Use `to_finish` instead of `to_abort` and `to_abort_message`.
- Change `set_finish_with_abort` into `to_finish` method, but keep the old calling stack.

### DIFF
--- a/python/sglang/srt/managers/scheduler.py
+++ b/python/sglang/srt/managers/scheduler.py
@@ -1817,20 +1817,13 @@ class Scheduler(
             TEST_RETRACT and self.forward_ct % TEST_RETRACT_INTERVAL == 0
         ):
             old_ratio = self.new_token_ratio
-            retracted_reqs, new_token_ratio, reqs_to_abort = batch.retract_decode(
-                self.server_args
-            )
+            retracted_reqs, new_token_ratio = batch.retract_decode(self.server_args)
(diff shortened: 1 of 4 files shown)


python/sglang/srt/managers/scheduler.py
```diff
@@ -1817,20 +1817,13 @@ def update_running_batch(self, batch: ScheduleBatch) -> Optional[ScheduleBatch]:
...
-            retracted_reqs, new_token_ratio, reqs_to_abort = batch.retract_decode(
-                self.server_args
-            )
+            retracted_reqs, new_token_ratio = batch.retract_decode(self.server_args)
             self.num_retracted_reqs = len(retracted_reqs)
             self.new_token_ratio = new_token_ratio
-            for req in reqs_to_abort:
-                self.send_to_tokenizer.send_output(
-                    AbortReq(abort_reason=req.to_abort_message, rid=req.rid), req
-                )
```
merrymercy, line 1828: we need this in one case of PD when retract happens and the host runs out of the memory

hnyls2002: I find the function `retract_decode` always returns an empty list as `req_to_abort` in current codebase.

PR #11563, opened 2025-10-13
[NVIDIA] Update to leverage flashinfer trtllm FP4 MOE throughput kernel 

jiahanc:
## Motivation
Use latest flashinfer trtllm FP4 MOE throughput kernel 

## Modifications
- Update flashinfer version to 0.4.1 to leverage latest flashinfer trtllm throughput moe kernel.
- Change use of falshinfer trtllm_fp4_block_scale_moe. Remove the tile_tokens_dim, pass None so flashinfer can calculate itself

## Accuracy Tests
`lm_eval --model local-completions --tasks gsm8k --model_args model=openai/gpt-oss-120b,base_url=http://0.0.0.0:30000/v1/completions,max_retries=3,tokenized_requests=False,timeout=1200,max_gen_toks=2048,max_length=8192 --batch_size 2048 --trust_remote_code --limit 0.5 `

```
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8621|±  |0.0134|
|     |       |strict-match    |     5|exact_match|↑  |0.6394|±  |0.0187|
```

## Benchmarking and Profiling
flashinfer kernel benchmark:
https://github.com/flashinfer-ai/flashinfer/pull/1882#issue-3493022673 

e2e benchmark
`python3 -m sglang.bench_offline_throughput        --model openai/gpt-oss-120b         --dataset-name random         --random-range-ratio 1         --num-prompts 1024  --random-input 8192  --random-output 1024 --tp 1 --disable-radix-cache`
Before
```
====== Offline Throughput Benchmark Result =======
Backend:                                 engine    
Successful requests:                     1024      
Benchmark duration (s):                  251.83    
Total input tokens:                      8388608   
Total generated tokens:                  1048576   
Last generation throughput (tok/s):      6038.87   
Request throughput (req/s):              4.07      
Input token throughput (tok/s):          33310.93  
Output token throughput (tok/s):         4163.87   
Total token throughput (tok/s):          37474.79  
==================================================
```
After
```
====== Offline Throughput Benchmark Result =======
Backend:                                 engine    
Successful requests:                     1024      
Benchmark duration (s):                  247.98    
Total input tokens:                      8388608   
Total generated tokens:                  1048576   
Last generation throughput (tok/s):      6044.20   
Request throughput (req/s):              4.13      
[...]

PR #12314, opened 2025-10-29
followup fix for llama 4 trtllm flashinfer backend

b8zhong:
## Motivation

After rebasing https://github.com/sgl-project/sglang/pull/11563 onto https://github.com/sgl-project/sglang/pull/11928, I thought that this interface would also internally calculate tile_tokens_dim. However it does not. So quickly fix it by setting default 8 for now (TODO, add it to Flashinfer to do the calculation internally as well).

Otherwise we get this sort of error:
```
TypeError: Mismatched type on argument #18 when calling: `trtllm_fp8_per_tensor_scale_moe(0: DLTensor*, 1: Optional<DLTensor*>, 2: DLTensor*, 3: DLTensor*, 4: DLTensor*, 5: DLTensor*, 6: DLTensor*, 7: DLTensor*, 8: DLTensor*, 9: int, 10: int, 11: int, 12: int, 13: int, 14: int, 15: int, 16: float, 17: bool, 18: int, 19: int, 20: bool) -> void`. Expected `int` but got `None`
```

Hotfix it.

## Checklist

### DIFF
--- a/python/sglang/srt/layers/quantization/modelopt_quant.py
+++ b/python/sglang/srt/layers/quantization/modelopt_quant.py
@@ -673,7 +673,7 @@ class ModelOptFp8MoEMethod(FusedMoEMethodBase):
                     routed_scaling_factor if routed_scaling_factor is not None else 1.0
                 ),
                 use_routing_scales_on_input=use_routing_scales_on_input,
-                tile_tokens_dim=8,
+                tile_tokens_dim=8,  # TODO(brayden): use the FI tile calculation
                 routing_method_type=routing_method_type,
             )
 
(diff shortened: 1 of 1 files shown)


Fridge003:
@b8zhong `tiles_token_dim` is unneeded for fp4 scaled moe, but needed for fp8 scaled moe, is that right?

python/sglang/srt/layers/quantization/modelopt_quant.py
```diff
@@ -673,7 +673,7 @@ def apply(
                     routed_scaling_factor if routed_scaling_factor is not None else 1.0
                 ),
                 use_routing_scales_on_input=use_routing_scales_on_input,
-                tile_tokens_dim=None,
+                tile_tokens_dim=8,
```
Fridge003: Can you add a line of TODO here

b8zhong: done

b8zhong:
@Fridge003 Yeah. For scaled moe fp4 they have a internal calculation, but not for this one (not sure why tbh)

Fridge003 (approved):

PR #12321, opened 2025-10-29
[NPU][Feat] Use einsum matmul op in MLA matrix absorb

randgun:
## Motivation
A operation optimization for DeepSeekV3 models when running on Ascend NPUs. We substitude (transpose + bmm + transpose) operations with a single fuse operation batch_matmul_transpose, which can accelerate decode time. It integrates an einsum-based matrix multiplication operation into the attention mechanism's matrix absorb core, which can be activated via a new environment variable SGLANG_NPU_USE_EINSUM_MM.

The batch_matmul_transpose operation for NPU has been merged on sgl-kernel-npu repo.
PR: https://github.com/sgl-project/sgl-kernel-npu/pull/77

## Modifications
- [x] **NPU Optimization**: Implemented an einsum-based matrix multiplication operation for the MLA matrix absorb step specifically for Ascend NPUs, aiming to enhance performance. From transpose + bmm + transpose to a single operation "batch_matmul_transpose".
- [x] **Configuration**: Introduced a new environment variable, SGLANG_NPU_USE_EINSUM_MM, to enable or disable the new einsum matmul optimization for NPU.
- [x] **Tensor Handling**: Added an additional .contiguous() call to the w_vc tensor during weight loading in the post_load_weights function for DeepSeekV2 models on ascend.

## Accuracy Tests
We test e2e accuracy on gsm8k dataset using sglang official test scripts, can be  found at python/sglang/test/few_shot_gsm8k.py.

image

## Benchmarking and Profiling
It can decrease 37% hot time consuming from **269 us** to **175 us** on Ascend A3 (bs=160, tp=16, dp=1), with batch_matmul_transpose op, kernel execute time is very close which can be ignored. In end 2 end test, we can decrease about 2 ms time consuming at decode phase.

屏幕截图 2025-11-06 103441

## Checklist


python/sglang/srt/models/deepseek_v2.py
```diff
@@ -3184,7 +3198,7 @@ def post_load_weights(self, is_nextn=False, weight_names=None):
                     self_attn.w_kc, w_kc.transpose(1, 2).contiguous().transpose(1, 2)
                 )
                 self_attn.w_vc = bind_or_assign(
-                    self_attn.w_vc, w_vc.contiguous().transpose(1, 2)
+                    self_attn.w_vc, w_vc.contiguous().transpose(1, 2).contiguous()
```
Fridge003: This contiguous operation will cause some bugs in fp8 kv cache.
Can it be specific for npu?

Fridge003 (changes requested):

python/sglang/srt/models/deepseek_v2.py
randgun: The w_vc is a input of torch.ops.npu.batch_matmul_transpose, it will cause some precision bugs when access memory in the batch_matmul_transpose op without contiguous. What bugs in fp8 kv cache you mean?

merrymercy (approved):

Issue #12280, opened 2025-10-28
[Bug] #11708 breaks FP8 SM120 inference


voipmonitor:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Describe the bug

commit 88596739a4637ad10a98d09473e71cee8064b9ec which is part of #11708  breaks FP8 GLM-4.6 inference 

### Reproduction

NCCL_P2P_LEVEL=4 NCCL_DEBUG=INFO PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True USE_TRITON_W8A8_FP8_KERNEL=1 SGL_ENABLE_JIT_DEEPGEMM=0 python -m sglang.launch_server --model /mnt/GLM-4.6-FP8/ --tp 4 --host 0.0.0.0 --port  4999 --mem-fraction-static 0.96 --context-length 200000  --enable-metrics  --attention-backend flashinfer   --tool-call-parser glm45    --reasoning-parser glm45   --served-model-name glm-4.5-air   --chunked-prefill-size 8092 --enable-mixed-chunk   --cuda-graph-max-bs 16   --kv-cache-dtype fp8_e5m2  --model-loader-extra-config '{"enable_multithread_load": true, "num_threads": 8}'

the inference output is completely garbled and the output containts: 

Failed to initialize the CUTLASS kernel. Last CUDA error is: Failed to initialize the CUTLASS kernel. Last CUDA error is: Failed to initialize the CUTLASS kernel. Last CUDA error is: Failed to initialize the CUTLASS kernel. Last CUDA error is: no errorno errorno errorno error
Failed to launch the CUTLASS kernel. Last CUDA error is: Failed to launch the CUTLASS kernel. Last CUDA error is: Failed to launch the CUTLASS kernel. Last CUDA error is: Failed to launch the CUTLASS kernel. Last CUDA error is: no errorno errorno errorno error
Failed to initialize the CUTLASS kernel. Last CUDA error is: no error
[...]

PR #12325, opened 2025-10-29
Fix Flashinfer Backend for SM120 Usage

weireweire:
## Motivation
As trtllm backend don't support sm120, we need to use flashinfer backend instead. Current prefill in flashinfer backend for sm120 have to use paged attention as the cutlass backend will fail otherwise. But paged attention only support weight absorb path for MLA. For prefill it's better to not absorb the weights for perf.

## Modifications
Use fa2 backend in flashinfer MLA for sm120 as cutlass backend don't support sm120. With this we can use ragged attention.

I tested using PP as it's faster on RTX Pro 6000D
```
SGLANG_PP_LAYER_PARTITION=8,8,8,8,8,7,7,7 SGLANG_USE_CUTLASS_BACKEND_FOR_FP4_GEMM=1 python3 -m sglang.launch_server --model-path nvidia/DeepSeek-R1-0528-FP4-v2 --trust-remote-code --host 0.0.0.0 --port 30000 --pp-size 8 --tp 1 --ep 1 --attention-backend flashinfer --moe-runner-backend flashinfer_cutlass --max-running-requests 32 --cuda-graph-max-bs 32 --context-length 4011 --max-prefill-tokens 16000 --chunked-prefill-size -1 --quantization modelopt_fp4 --kv-cache-dtype fp8_e4m3 --mem-fraction-static 0.7 --disable-radix-cache --schedule-conservativeness 0.0 --allow-auto-truncate 
```

## Accuracy Tests

|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9553|±  |0.0057|
|     |       |strict-match    |     5|exact_match|↑  |0.9522|±  |0.0059|

## Benchmarking and Profiling
I use this command on RTX Pro 6000D x 8 to test prefill perf(TTFT)
```
python3 -m sglang.bench_serving --backend sglang --model nvidia/DeepSeek-R1-0528-FP4-v2 --host 127.0.0.1 --port 30000 --dataset-name random --max-concurrency 32 --num-prompts 160 --random-input-len 3999 --random-output-len 1 --random-range-ratio 1
```

Before the modification, we have to enable `--flashinfer-mla-disable-ragged` to the server otherwise it will fail for `Failed to launch the CUTLASS kernel. Last CUDA error is: no error` 
And the perf is 
```
Max request concurrency:                 32        
Successful requests:                     160       
Benchmark duration (s):                  68.12     
Total input tokens:                      639840    
Total input text tokens:                 639840    
Total generated tokens:                  160       
Total generated tokens (retokenized):    159       
Request throughput (req/s):              2.35      
Input token throughput (tok/s):          9392.55   
Output token throughput (tok/s):         2.35      
Total token throughput (tok/s):          9394.90   
Concurrency:                             29.91     
Mean E2E Latency (ms):                   12736.30  
Median E2E Latency (ms):                 12112.68  
Mean TTFT (ms):                          12666.90  
Median TTFT (ms):                        12112.67  
P99 TTFT (ms):                           21516.09  
Mean ITL (ms):                           0.00      
Median ITL (ms):                         0.00      
P95 ITL (ms):                            0.00      
P99 ITL (ms):                            0.00      
Max ITL (ms):                            0.00      
```

After the modification:
```
Max request concurrency:                 32        
Successful requests:                     160       
Benchmark duration (s):                  37.01     
Total input tokens:                      639840    
Total input text tokens:                 639840    
Total generated tokens:                  160       
Total generated tokens (retokenized):    159       
Request throughput (req/s):              4.32      
Input token throughput (tok/s):          17289.02  
Output token throughput (tok/s):         4.32      
Total token throughput (tok/s):          17293.34  
Concurrency:                             30.00     
Mean E2E Latency (ms):                   6938.33   
Median E2E Latency (ms):                 6751.35   
Mean TTFT (ms):                          6902.94   
Median TTFT (ms):                        6751.35   
P99 TTFT (ms):                           11325.32  
Mean ITL (ms):                           0.00      
Median ITL (ms):                         0.00      
P95 ITL (ms):                            0.00      
P99 ITL (ms):                            0.00      
Max ITL (ms):                            0.00      
```

## Checklist

### DIFF
--- a/python/sglang/srt/layers/attention/flashinfer_mla_backend.py
+++ b/python/sglang/srt/layers/attention/flashinfer_mla_backend.py
@@ -25,8 +25,8 @@ from sglang.srt.model_executor.forward_batch_info import ForwardBatch, ForwardMo
 from sglang.srt.server_args import get_global_server_args
 from sglang.srt.speculative.spec_info import SpecInput
 from sglang.srt.utils import (
-    is_blackwell_supported,
     is_flashinfer_available,
+    is_sm100_supported,
     next_power_of_2,
 )
 
@@ -242,9 +242,11 @@ class FlashInferMLAAttnBackend(AttentionBackend):
         else:
             self.q_indptr_decode = q_indptr_decode_buf
 
-        self.fmha_backend = "auto"
-        if is_blackwell_supported():
+        if is_sm100_supported():
             self.fmha_backend = "cutlass"
+        else:
+            self.fmha_backend = "auto"
+
         self.prefill_wrapper_ragged = BatchPrefillWithRaggedKVCacheWrapper(
             self.workspace_buffer, "NHD", backend=self.fmha_backend
         )
--- a/python/sglang/srt/layers/attention/flashinfer_backend.py
+++ b/python/sglang/srt/layers/attention/flashinfer_backend.py
@@ -26,8 +26,8 @@ from sglang.srt.model_executor.forward_batch_info import ForwardBatch, ForwardMo
 from sglang.srt.speculative.spec_info import SpecInput
 from sglang.srt.utils import (
     get_int_env_var,
-    is_blackwell_supported,
     is_flashinfer_available,
+    is_sm100_supported,
     next_power_of_2,
 )
 
@@ -229,7 +229,7 @@ class FlashInferAttnBackend(AttentionBackend):
             ]
 
         fmha_backend = "auto"
-        if is_blackwell_supported():
+        if is_sm100_supported():
             # Disable CUTLASS backend when piecewise cuda graph is enabled
             # due to TMA descriptor initialization issues on B200
             if model_runner.server_args.enable_piecewise_cuda_graph:


python/sglang/srt/models/deepseek_v2.py
```diff
@@ -320,7 +320,7 @@ def _handle_attention_backend(
 ):
     sum_extend_prefix_lens = _get_sum_extend_prefix_lens(forward_batch)
     disable_ragged = (
-        backend_name in ["flashinfer", "flashmla"]
+        backend_name in ["flashmla"]
```
Fridge003: We can't modify this line, otherwise `flashinfer_mla_disable_ragged` will lose control of flashinfer backend.

 `flashinfer_mla_disable_ragged`  is false by default btw

weireweire: uh, make sense, reverted

weireweire: so the issue is sm120 don't support non-ragged and have to use the flag. change flashinfer backend can solve

python/sglang/srt/layers/attention/flashinfer_mla_backend.py
```diff
@@ -243,8 +244,11 @@ def __init__(
             self.q_indptr_decode = q_indptr_decode_buf
 
         self.fmha_backend = "auto"
```
Fridge003: We can move this to `else` branch

weireweire: done

Fridge003 (approved):
LGTM

voipmonitor:
@weireweire why not elif is_sm120_supported():
            self.fmha_backend = "auto" ? but forcing fa2? if it is "auto" it still works - how to check what is actually chosen if you let it to the "auto" ? I'm not testing it with deepseek but with GLM-4.6-FP8 and it works with "auto" (but not with cutlass) 
            
 https://github.com/sgl-project/sglang/issues/12280

weireweire:
@voipmonitor "auto" also result to "fa2" inside flashinfer, I just tested fa2 and followed the cutlass way, let me change it.  Actually I think both SGL and flashinfer didn't care about sm120 before, but it happened work with auto..

voipmonitor:
> @voipmonitor "auto" also result to "fa2" inside flashinfer, I just tested fa2 and followed the cutlass way, let me change it. Actually I think both SGL and flashinfer didn't care about sm120 before, but it happened work with auto..

thanks - and how about python/sglang/srt/layers/attention/flashinfer_backend.py - there is also fmha_backend = cutlass for sm120 - same issue like flashinfer_mla_backend.py

weireweire:
@voipmonitor should be the same, fixed. thanks for reporting this.

Fridge003:
https://github.com/sgl-project/sglang/actions/runs/18904282342/job/54005360496?pr=12325

PR #12331, opened 2025-10-29
[Ascend]Support dynamic w8a8

ZhengdQin:
## Motivation

Static quantization method & dynamic w8a8 quantization method are all support for Ascend NPU. In this pr, we support dynamic w8a8 weights. We test on gms8k dataset, and two kinds of weights get the same scores. One can use the weight converter script to convert the raw fp8 weights to the dynamic w8a8 weights (compressed tensor), see https://gitcode.com/cann/cann-recipes-infer/blob/master/models/deepseek-v3.2-exp/utils/weight_convert.sh.

## Modifications

-  We extended the conditional statement for `is_layer_skipped` to the NPU for the case that the weight name is found;

-  We added `should_ignore_layer` to the NPU branch. This logic was originally included in other hardware but was omitted before.

-  We extended the conditional statement for `is_dynamic` to determine if the weights are dynamic;

## Accuracy Tests

We test on the GMS8K dataset. The accuracy is **0.955** for static W8A8-weight and dynamic W8A8-weight.
image001

## Checklist

### DIFF
--- a/python/sglang/srt/layers/quantization/w8a8_int8.py
+++ b/python/sglang/srt/layers/quantization/w8a8_int8.py
@@ -210,6 +210,8 @@ class W8A8Int8Config(QuantizationConfig):
         )
 
         if _is_npu:
+            if self.is_dynamic:
+                return
             # Ascend w8a8_int8 quantization with bias, use wrappers to isolate the effects between models
             for name in self.quant_description.keys():
                 if "norm.bias" in name:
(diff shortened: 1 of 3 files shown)


python/sglang/srt/layers/quantization/w8a8_int8.py
```diff
@@ -210,6 +210,8 @@ def __init__(self, quant_config: Dict[str, Any] = {}):
         )
 
         if _is_npu:
+            if self.is_dynamic:
```
iforgetmyname, line 213: pls add comment "TODO(iforgetmyname): refactor out _is_npu here" to help me remember reverify dynamic quant support

python/sglang/srt/layers/quantization/w8a8_int8.py
```diff
@@ -265,6 +267,10 @@ def get_quant_method(
         from sglang.srt.layers.moe.fused_moe_triton import FusedMoE
 
         if _is_npu:
+            if should_ignore_layer(
```
zhuyijie88, line 270: ignore 放到isinstance(layer, LinearBase):条件里

PR #12332, opened 2025-10-29
[PD] feat: refactor custom mem pool and add barex pd support

stmatengss:
## Motivation

1. Provide a function to switch between different custom allocator class for more backends.
2. ACCL-Barex (https://www.alibabacloud.com/help/en/pai/user-guide/accl-alibaba-high-performance-collective-communication-library) is a point-to-point transfer library for Alibaba Cloud. We provide a Barex allocator (https://github.com/kvcache-ai/Mooncake/pull/932) in mooncake side to support GDR because of several hardware limitations, thus GDR-style PD disaggregation can be enabled with ACCL.

Usage
```
export SGLANG_MOONCAKE_CUSTOM_MEM_POOL=NVLINK|BAREX
```

## Modifications

1. Change `SGLANG_MOONCAKE_CUSTOM_MEM_POOL` from `Bool` to `String` type, and keep forward compatibility (still support `SGLANG_MOONCAKE_CUSTOM_MEM_POOL=true|false`).
2. Import `BarexAllocator` from mooncake, and make it as default `torch.cuda.mempool`.

## Checklist


stmatengss:
@ShangmingCai All these changes are similar to NVLinkAllocator. PTAL.

python/sglang/srt/mem_cache/memory_pool.py
```diff
@@ -53,6 +54,9 @@
 logger = logging.getLogger(__name__)
 
 GB = 1024 * 1024 * 1024
+
+SUPPORTED_CUSTOM_MEM_POOL_TYPES = ["NVLINK", "BAREX"]
```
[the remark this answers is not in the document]
ShangmingCai: These lines should not be placed in memory_pool this file.

python/sglang/srt/mem_cache/memory_pool.py
```diff
@@ -145,14 +149,33 @@ def __init__(
...
+        if custom_mem_pool_type is not None:
+            # Handle boolean True as NVLINK
+            if custom_mem_pool_type.lower() == "true":
+                custom_mem_pool_type = "NVLINK"
+            self.enable_custom_mem_pool = (
+                custom_mem_pool_type in SUPPORTED_CUSTOM_MEM_POOL_TYPES
+            )
+        else:
+            self.enable_custom_mem_pool = False
+
```
ShangmingCai: Try using the attributes of `Envs.SGLANG_MOONCAKE_CUSTOM_MEM_POOL` instead of `os.getenv`.

python/sglang/srt/mem_cache/memory_pool.py
```diff
@@ -428,14 +451,31 @@ def __init__(
...
-        )
+        custom_mem_pool_type = os.getenv("SGLANG_MOONCAKE_CUSTOM_MEM_POOL")
+        if custom_mem_pool_type is not None:
+            # Handle boolean True as NVLINK
+            if custom_mem_pool_type.lower() == "true":
+                custom_mem_pool_type = "NVLINK"
+            self.enable_custom_mem_pool = custom_mem_pool_type in SUPPORTED_CUSTOM_MEM_POOL_TYPES
+        else:
+            self.enable_custom_mem_pool = False
+
```
ShangmingCai: ditto

python/sglang/srt/mem_cache/memory_pool.py
```diff
@@ -937,14 +977,31 @@ def __init__(
...
```
ShangmingCai: ditto

ShangmingCai (commented):
Maybe I should abstract custom allocator class first, so that we don't need to replicate the code `if custom_mem_pool_type == xxxx:` in every memory pool.

stmatengss:
> Maybe I should abstract custom allocator class first, so that we don't need to replicate the code `if custom_mem_pool_type == xxxx:` in every memory pool.

Sure. We can abstract the allocator class to support more in-house GPU link protocols. Should this be implemented in Mooncake or SGLang?

ShangmingCai:
> > Maybe I should abstract custom allocator class first, so that we don't need to replicate the code `if custom_mem_pool_type == xxxx:` in every memory pool.
> 
> Sure. We can abstract the allocator class to support more in-house GPU link protocols. Should this be implemented in Mooncake or SGLang?

We better put it in the Mooncake side, and use env var to switch allocator at the mooncake side with the default value set to nvlink. Then we can unified the code in the sglang.

stmatengss:
> > > Maybe I should abstract custom allocator class first, so that we don't need to replicate the code `if custom_mem_pool_type == xxxx:` in every memory pool.
> > 
> > 
> > Sure. We can abstract the allocator class to support more in-house GPU link protocols. Should this be implemented in Mooncake or SGLang?
> 
> We better put it in the Mooncake side, and use env var to switch allocator at the mooncake side with the default value set to nvlink. Then we can unified the code in the sglang.

I believe some logic isn't common on the mooncake side. Therefore, I've extracted the switching allocator's logic into `utils` and created two general APIs. PTAL, thx @ShangmingCai

python/sglang/srt/mem_cache/utils.py
```diff
@@ -208,3 +218,88 @@ def get_mla_kv_buffer_triton(
...
+        enable_custom_mem_pool = custom_mem_pool_type in SUPPORTED_CUSTOM_MEM_POOL_TYPES
+    else:
+        enable_custom_mem_pool = False
+        custom_mem_pool_type = None
+
+    custom_mem_pool = None
+
+    if enable_custom_mem_pool:
+        try:
+            # TODO(shangming): abstract custom allocator class for more backends
```
ShangmingCai:
```suggestion
```

python/sglang/srt/mem_cache/utils.py
```diff
@@ -208,3 +218,88 @@ def get_mla_kv_buffer_triton(
...
+    custom_mem_pool_type = envs.SGLANG_MOONCAKE_CUSTOM_MEM_POOL.get()
+
+    if custom_mem_pool_type is not None:
+        # Handle boolean True as NVLINK
+        if custom_mem_pool_type.lower() == "true":
+            custom_mem_pool_type = "NVLINK"
+        enable_custom_mem_pool = custom_mem_pool_type in SUPPORTED_CUSTOM_MEM_POOL_TYPES
+    else:
+        enable_custom_mem_pool = False
+        custom_mem_pool_type = None
```
ShangmingCai: Maybe reuse `check_custom_mem_pool_enabled` here.

python/sglang/srt/mem_cache/utils.py
```diff
@@ -208,3 +218,88 @@ def get_mla_kv_buffer_triton(
         nope_dim,
         rope_dim,
     )
+
+
+def init_custom_mem_pool(
+    device: str,
+) -> Tuple[bool, Optional[Any], Optional[str]]:
```
ShangmingCai: Maybe change this to `init_custom_mem_pool`.

Should we create a new file /disaggregation/mooncake/utils.py and a `init_mooncake_custom_mem_pool` util in it with some of the following implementation? If env var is not set, then we can directly `return False, None, None`.

python/sglang/srt/mem_cache/utils.py
```diff
@@ -208,3 +218,88 @@ def get_mla_kv_buffer_triton(
...
+            )
+            enable_custom_mem_pool = False
+            custom_mem_pool = None
+        except Exception as e:
+            logger.error(
+                f"Failed to initialize custom memory pool {custom_mem_pool_type}: {e}. "
+                f"Falling back to default memory pool."
+            )
+            enable_custom_mem_pool = False
+            custom_mem_pool = None
```
ShangmingCai: I think placed these lines in `init_mooncake_custom_mem_pool` would be better.

python/sglang/srt/mem_cache/utils.py
```diff
@@ -13,10 +13,20 @@
...
 import torch
 import triton
 import triton.language as tl
 
+from sglang.srt.environ import envs
+
+logger = logging.getLogger(__name__)
+
+# Global constants for custom memory pool types
+SUPPORTED_CUSTOM_MEM_POOL_TYPES = ["NVLINK", "BAREX"]
```
ShangmingCai:
```suggestion
# Global constants for custom memory pool types
SUPPORTED_MOONCAKE_CUSTOM_MEM_POOL_TYPES = ["NVLINK", "BAREX"]
```

And it should be moved to /disaggregation/mooncake/utils.py

python/sglang/srt/disaggregation/mooncake/conn.py
```diff
@@ -201,8 +197,8 @@ def __init__(
                 "SGLANG_DISAGGREGATION_BOOTSTRAP_TIMEOUT", 300
             )
 
-            self.enable_custom_mem_pool = get_bool_env_var(
-                "SGLANG_MOONCAKE_CUSTOM_MEM_POOL", "false"
+            self.enable_custom_mem_pool, custom_mem_pool_type = (
+                check_custom_mem_pool_enabled()
```
ShangmingCai: Remember to address `send_aux`

python/sglang/srt/mem_cache/memory_pool.py
```diff
@@ -42,19 +42,14 @@
 import triton.language as tl
 
 from sglang.srt.constants import GPU_MEMORY_TYPE_KV_CACHE
+from sglang.srt.disaggregation.mooncake.utils import init_mooncake_custom_mem_pool
```
ShangmingCai:
```suggestion
from sglang.srt.utils import maybe_init_custom_mem_pool
```

Then we define a function in sglang.srt.utils:
```python
def maybe_init_custom_mem_pool(
    device: str,
) -> Tuple[bool, Optional[Any], Optional[str]]:
    # This function can be modified to support more features that require a custom memory pool.
    enable_custom_mem_pool = True if envs.SGLANG_MOONCAKE_CUSTOM_MEM_POOL.get() is not None else False

    if enable_custom_mem_pool:
        # Currently, only mooncake requires a custom mem pool for MNNVL PD disaggregation
        from sglang.srt.disaggregation.mooncake.utils import init_mooncake_custom_mem_pool
        return init_mooncake_custom_mem_pool(device)
    else:
        return False, None, None
```

ShangmingCai: Just this minor suggestion, others LGTM.

ShangmingCai: Remember to fix lint.

ShangmingCai (approved):
LGTM now.

stmatengss:
PTAL when it's convenient. Thanks! @xiezhq-hermann

python/sglang/srt/mem_cache/utils.py
```diff
@@ -208,3 +212,32 @@ def get_mla_kv_buffer_triton(
...
+    This function can be modified to support more features that require a custom memory pool.
+
+    Args:
+        device: The device to allocate memory on
+
+    Returns:
+        Tuple of (enable_custom_mem_pool, custom_mem_pool, custom_mem_pool_type)
+    """
+    enable_custom_mem_pool = (
+        True if envs.SGLANG_MOONCAKE_CUSTOM_MEM_POOL.get() is not None else False
```
xiezhq-hermann, line 232: does this support other transfer engine as well? if so, might be a good practice to have a more general set of statements.

xiezhq-hermann (approved):

python/sglang/srt/mem_cache/utils.py
stmatengss: Unsure if other transfer engines support these transports; if so, add them to utils.py.

PR #12335, opened 2025-10-29
[BUGFIX] qwen3-next gdn add stride for a

yizhang2077:
## Motivation

 If `num_v_heads = num_k_heads`, then `a.is_contiguous() = False` after `fix_query_key_value_ordering` and will cause accuracy error since `a` need be contiguous for triton kernel. 
This bug does not happen in current OSS Qwen3-Next series models since `num_v_heads = 2 * num_k_heads`.
## Checklist

---
### CI States

Latest PR Test (Base): :x: [Run #19419372268](https://github.com/sgl-project/sglang/actions/runs/19419372268)
Latest PR Test (Extra): :x: **Blocked** -- `run-ci` is required first.
Latest PR Test (AMD ROCm 10): :heavy_minus_sign: **No AMD PR run found for this commit**.


python/sglang/srt/layers/attention/hybrid_linear_attn_backend.py
```diff
@@ -292,7 +292,7 @@ def forward_decode(
             q=query,
             k=key,
             v=value,
-            a=a,
+            a=a.contiguous(),
```
ispobock: can we read it with stride in the kernel to reduce a kernel launch?

yizhang2077: done

python/sglang/srt/models/qwen3_next.py
```diff
@@ -196,6 +196,8 @@ def fused_gdn_gating_kernel(
     g,
     A_log,
     a,
+    a_stride_m: tl.constexpr,
```
ispobock: move the a_stride_m and a_stride_k after seq_len to keep tl.constexpr at the end.

ispobock (approved):

PR #12353, opened 2025-10-29
[NVIDIA] Fix cutedsl backend of MoE

kaixih:
## Motivation

The cutedsl moe backend execution is broken. This PR fixes it.

## Modifications

Fixed the execution path and added CI tests.

## Accuracy Tests

The CI test results:
```
Accuracy: 0.951
Invalid: 0.000
Latency: 61.249 s
```

## Checklist


kaixih:
@wenscarl @fzyzcjy @Fridge003

kaixih:
Force push to resolve conflicts.

fzyzcjy:
btw curious why CI did not catch this error

kaixih:
I think because we only created the test but didn't add it to the CI.

fzyzcjy:
oops. is it b/c you forget it or deliberately not add it? I personally forgot adding it many times... do you think this is useful -> https://github.com/sgl-project/sglang/pull/6947 (it was reverted b/c it caused ci red due to not rebasing latest main, but I guess I can add it again if useful)

kaixih:
Oh, definitely not intentional. I think in many cases people just assume the codebase will magically pick up tests from files like test-xxx.py. Either way, automatic detection or manual review works for us.

fzyzcjy:
ok i will add that to my list

kaixih:
@Fridge003 @fzyzcjy @trevor-m can you help review the PR?

trevor-m (approved):
LGTM

test/srt/run_suite.py
```diff
@@ -181,6 +181,7 @@ class TestFile:
     ],
     "per-commit-4-gpu-b200": [
         TestFile("test_deepseek_v3_fp4_4gpu.py", 3600),
+        TestFile("test_deepseek_r1_cutedsl_4gpu.py", 3600),
```
Fridge003: This will add too much burden to the CI machine (We only have 4*B200)
Is there anyway to reduce its time. Like moving this CI to gb200 workload

.github/workflows/pr-test.yml
```diff
@@ -813,6 +813,35 @@ jobs:
           cd test/srt
           python3 run_suite.py --suite per-commit-4-gpu-gb200 --auto-partition-id 0 --auto-partition-size 1 --timeout-per-file 3600
 
+  unit-test-backend-4-gpu-gb200-deepep:
+    needs: [check-changes, sgl-kernel-build-wheels]
```
Fridge003: Can we merge it into unit-test-backend-4-gpu-gb200 test? So we don't need to launch the environment twice

lpc0220:
I am wondering why do we need so many different moe backends? in which case which moe backend would work best? Do we need a benchmarking suite to test and classify all moe backends in SGLang? @kaixih @fzyzcjy @Fridge003 @trevor-m ?

.github/workflows/pr-test.yml
kaixih: Sure. I will remove the dsv3 test since it doesn't need to run the same test on 4xb200 and gb200 twice. The cutedsl one is better for gb200 since its usage case is nvl72 gb200.

Fridge003:
cc @kaixih #12480 added a kernel build flow for aarch64, which can be used for gb200 CI

kaixih:
@Fridge003 I commented off the e2e tests for now since there seems an issue of the GB200 CI runner + IBGDA ([issue](https://github.com/sgl-project/sglang/issues/12533)). To unblock this PR, I use an cutedsl moe test in the GB200 runner instead. I think we can enable the e2e test when the issue is resolved.

Any advise? Can you please take another look?

test/srt/run_suite.py
```diff
@@ -180,7 +180,9 @@ class TestFile:
         TestFile("test_llama31_fp4.py", 300),
     ],
     "per-commit-4-gpu-gb200": [
-        TestFile("test_deepseek_v3_fp4_4gpu.py", 3600),
```
Fridge003, line 183: Keep this test until the cutedsl one can be run

python/sglang/test/test_utils.py
```diff
@@ -57,7 +57,7 @@
 DEFAULT_MODEL_NAME_FOR_TEST_MLA_NEXTN = "lmsys/sglang-ci-dsv3-test-NextN"
 
 # NVFP4 models
-DEFAULT_DEEPSEEK_NVFP4_MODEL_FOR_TEST = "nvidia/DeepSeek-R1-0528-FP4"
+DEFAULT_DEEPSEEK_NVFP4_MODEL_FOR_TEST = "nvidia/DeepSeek-V3-0324-FP4"
```
Fridge003, line 60: Why changing the default model?

kaixih: For this one, when I use the R1 checkpoints, I remember the runner will hang there (preparing the datasets?) like forever. Then I change back the "existing" V3 checkpoints that are already used by other runners.

test/srt/run_suite.py
kaixih: Done.

PR #12359, opened 2025-10-30
[Embeddings Performance Testing] Add performance test for embedding models

vedantjh2:
## Motivation

We wish to ensure any changes to SGLang in the future do not degrade the performance of decoder only embeddings generation in models like qwen2-gte and qwen3-embedding. LinkedIn uses these models in production and any regression in performance is not acceptable. These tests define a baseline based on testing on a single H100 with sglang==0.5.4.post1. 

## Modifications

Added two tests :
1. test_embeddings_api_latency_throughput
2. test_embeddings_api_batch_scaling

Included them in the pr-test.yaml so they are run during cicd

## Accuracy Tests

Does not affect model output

## Benchmarking and Profiling

Does not impact inference speed

## Checklist


test/srt/test_bench_serving.py
```diff
@@ -489,23 +491,77 @@ def test_score_api_batch_scaling(self):
...
+                    f"Average latency: {res['avg_latency_ms']:.2f} ms\n"
+                    f"P95 latency: {res['p95_latency_ms']:.2f} ms\n"
+                    f"Throughput: {res['throughput']:.2f} req/s\n"
+                    f"Successful requests: {res['successful_requests']}/{res['total_requests']}\n"
+                )
+
+            self.assertEqual(res["successful_requests"], res["total_requests"])
             if batch_size == 10:
-                p95_latency_bound = 50
+                avg_latency_bound = 60
```
sundar24295s: Does this test runs only on H100 or do we need to configure separate thresholds for each hardware?

sundar24295s (approved):

Kangyan-Zhou (approved):

sundar24295s (approved):

PR #12361, opened 2025-10-30
[Fix] Fix trtllm-mla backend when chunked prefix cache is disabled

Fridge003:
## Checklist


Fridge003:
cc @ishandhanani  @fzyzcjy

python/sglang/srt/layers/attention/trtllm_mla_backend.py
```diff
@@ -537,6 +545,7 @@ def init_forward_metadata(self, forward_batch: ForwardBatch):
             or forward_batch.forward_mode.is_target_verify()
             or forward_batch.forward_mode.is_draft_extend(include_v2=True)
         ):
+            self.fallback_to_flashinfer_mla = False
```
fzyzcjy: nit: shall we put it into forward_prefill_metadata etc to be more organized

Fridge003: Done

wenscarl:
Could you link to the issue here? 
The failure should be caused by trtllm_ragged_attention_deepseek is used for the prefill with weight absorption. We need a "decode like" kernel to handle this case. Shouldn't [trtllm_batch_decode_with_kv_cache_mla](https://github.com/sgl-project/sglang/blob/6c01844f45a36582fa591e0c716acd44a5cf491b/python/sglang/srt/layers/attention/trtllm_mla_backend.py#L544) be more performant?

Fridge003:
> trtllm_batch_decode_with_kv_cache_mla

Will do it in the next PR

PR #12129, opened 2025-10-25
Support MiniMax M2 model

zhaochenyang20:
## Motivation

Supporting MiniMax M2, the SOTA model from MiniMax and we set up a standard for merging new LLMs.

1. Adding model support to SGLang, with minimal disturbance to SGLang's main logic.
2. Adding docs to SGLang, in [LLM](https://github.com/sgl-project/sglang/blob/main/docs/supported_models/generative_models.md) and [VLM](https://github.com/sgl-project/sglang/blob/main/docs/supported_models/multimodal_language_models.md).
4. Benchmarking the performance and submitting the results.
5. Benchmarking the throughput and submitting the results.
6. Submit all the reproduction commands, including "installation, evaluation devices".

## Install

```bash
git clone -b minimax https://github.com/sgl-project/sglang.git
cd sglang

# Install the python packages
pip install --upgrade pip
pip install -e "python"
```
## Launch

```bash
python -m sglang.launch_server \
--model-path MiniMaxAI/MiniMax-M2 \
--tp-size 8 \
--ep-size 8 \
--tool-call-parser minimax-m2 \
--trust-remote-code \
--reasoning-parser minimax-append-think \
--mem-fraction-static 0.85
```

I only tried on 8 * H200. If adding:

```bash
--moe-a2a-backend deepep \
--deepep-mode auto
```

will fail.

### MMLU

```bash
python -m sglang.test.run_eval --eval-name mmlu --port 30000 --num-examples 1000
```

H200 score: 0.726, 0.735 for the 1000 examples.
B200 score: 0.721

With four H200, we have 0.720.

### gsm8k

```bash
python -m sglang.test.few_shot_gsm8k --host http://127.0.0.1 --port 30000 --num-questions 200 --num-shots 5
```
H200 score: 0.950 for the 200 questions with 5 shots.
B200 score: 0.945

### HellaSwag

```bash
cd benchmark/hellaswag
python3 bench_sglang.py --num-questions 200 --num-shots 20 --host http://127.0.0.1 --port 30000
```

H200 score: 0.765 for the 200 questions with 20 shots.
B200 score: 0.765

### GPQA

1. M2 is a force thinking model, so the command is as follows. The same as turning on training:

```bash
python3 -m sglang.test.run_eval --port 30000 --eval-name gpqa --num-examples 198 --max-tokens 120000 --repeat 8
```

H200:
Total latency: 10021.513 s
Score: 0.727, 0.747, 0.778, 0.776, from MiniMax internal 0.771

[...]

PR #12367, opened 2025-10-30
[Bug] Fix Intern-S1 model accuracy and support /generate interface with input_ids

hhaAndroid:
1. The version of `pixel_shuffle` used in `intern-s1` is incorrect and the `ps_version` parameter needs to be removed
2. The image processing models in `intern-s1` and `internv1` do not support passing `input_ids` through the generation interface, which needs to be fixed.
3. In the RL scenario, weight updates are incremental, and the `weight update` code for interns1 and internvl needs to be fixed.

### DIFF
--- a/python/sglang/srt/models/interns1.py
+++ b/python/sglang/srt/models/interns1.py
@@ -1,4 +1,4 @@
-from typing import Iterable, List, Optional, Set, Tuple
+from typing import Iterable, List, Optional, Tuple
 
 import torch
 from torch import nn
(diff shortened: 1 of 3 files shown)


hhaAndroid:
@zhaochenyang20

zhaochenyang20 (commented):
I suggest treating this as a refactor of a new VLM, so it should be like submitting a new model. I will create a new guide for how to benchmark the accuracy performance of VLM, following this:

https://github.com/sgl-project/sglang/pull/12129#issue-3552989409

hhaAndroid:
@zhaochenyang20

mickqian (approved):

PR #12369, opened 2025-10-30
Enable bailing_moe to support TP=16

guoyuhong:
## Motivation
Currently, bailing_moe only supports tensor parallelism (TP) configurations where TP size is less than or equal to the number of KV heads. However, for Ling-1T — due to their large size — users may need to choose configurations such as TP16 or TP8+PP2. The current implementation fails with an error under TP16.
image

## Modifications
This PR extends bailing_moe to support TP sizes greater than the number of KV heads, using the same approach adopted in qwen3_moe.
https://github.com/sgl-project/sglang/blob/v0.5.4.post1/python/sglang/srt/models/qwen3_moe.py#L294-L302

## Accuracy Tests
image

## Checklist


strgrb (approved):
LGTM

PR #12370, opened 2025-10-30
fix the impact of D2H copying on overlap scheduling

lizhigong1988:
fix the impact of D2H copying on overlap scheduling

## Checklist


lizhigong1988:
When I was analyzing the performance on the DCU, I found that the overlap scheduling did not work. I discovered that this line of code had a copy of D2H, which caused CPU synchronization and bubbles. After modification, the correctness and performance were verified to be normal.

PR #12371, opened 2025-10-30
[feat] npu support enable_torch_compile

XDaoHong:
## Motivation
**TorchAir** (Torch Ascend Intermediate Representation) is an extension library that provides graph mode capabilities for torch_npu. It enables users to perform graph-mode inference on NPU using PyTorch and torch_npu. TorchAir externally offers a `torch.compile` backend for NPU, which interfaces with `torch._dynamo`. Through the following features, performance optimization and capability enhancement of the torch fx graph can be achieved.

image

Main Features:

1. Basic Features:

- Enable NPU kernels that depend on host-value tiling operators (e.g., FIA) to support npugraph
- Graph input copy optimization
- Memory reuse across multi-graphs

2. FX Pass:

- In-place optimization
- Redundant operator elimination
- NPU fused operator passes

3. Advanced Features:

- Static shape kernel compilation
- Multi-stream within single graphs
- Compilation caching

## Modifications

1. Rewrite the capture function;
2. Encapsulate the kvcache input (input needs all kvcache);
3. Pad the block table to the max length;
4. TorchAir input preparation;

The calling process is as follows.
torchair调用流程

## Accuracy Tests
python3 few_shot_gsm8k.py --data-path "/path/to/model/test.jsonl.txt” --parallel 32 --num-questions 200

```
Accuracy: 0.865
Invalid: 0.000
Latency: 43.077 s
Output throughput: 795.877 token/s
```

## Future roadmaps

In the torch_npu 7.2.0 version, the **reduce-overhead** mode of the torchair backend will support torch.compile(model, dynamic=True). This mode will be set as the default in get_compile_backend(), enabling support for methods wrapped by the @torch.compile() decorator.
In the torch_npu 7.3.0 version, the capture and replay of NPUGraph currently integrated in the torchair backend will be changed to optional execution. The torchair backend will only perform optimizations such as fx pass optimization and static kernel compilation, while the capture and replay of NPUGraph will be implemented independently. This design is closer to the implementation of CudaGraphRunner, decoupling fx graph optimization from graph offloading.

## Checklist

### DIFF
--- a/python/sglang/srt/hardware_backend/npu/memory_pool_npu.py
+++ b/python/sglang/srt/hardware_backend/npu/memory_pool_npu.py
@@ -100,13 +100,22 @@ class NPUMHATokenToKVPool(MHATokenToKVPool):
             self.get_value_buffer(i).nbytes
             for i in range(self.start_layer, self.start_layer + self.layer_num)
         ]
-        kv_item_lens = [
-            self.get_key_buffer(i)[0].nbytes
-            for i in range(self.start_layer, self.start_layer + self.layer_num)
-        ] + [
-            self.get_value_buffer(i)[0].nbytes
-            for i in range(self.start_layer, self.start_layer + self.layer_num)
-        ]
+        if self.use_fia:
+            kv_item_lens = [
+                self.get_key_buffer(i)[0].nbytes * self.page_size
+                for i in range(self.start_layer, self.start_layer + self.layer_num)
+            ] + [
+                self.get_value_buffer(i)[0].nbytes * self.page_size
+                for i in range(self.start_layer, self.start_layer + self.layer_num)
+            ]
+        else:
+            kv_item_lens = [
+                self.get_key_buffer(i)[0].nbytes
+                for i in range(self.start_layer, self.start_layer + self.layer_num)
+            ] + [
+                self.get_value_buffer(i)[0].nbytes
+                for i in range(self.start_layer, self.start_layer + self.layer_num)
+            ]
         return kv_data_ptrs, kv_data_lens, kv_item_lens
 
     def set_kv_buffer(


ssshinigami:
Could you please add description and motivation
And accuracy and performance measurements

python/sglang/srt/utils/common.py
```diff
@@ -1871,10 +1862,7 @@ def get_compiler_backend() -> str:
                 "Please install torchair for torch.compile support on NPU."
             )
         compiler_config = CompilerConfig()
-        predefined_config = get_npu_compiler_config()
-        for k, v in predefined_config.items():
-            setattr(compiler_config.experimental_config, k, v)
-
+        compiler_config.mode = "max-autotune" if mode is None else mode
```
iforgetmyname: please add a comment "TODO(iforgetmyname): Change this default value once CANN version 8.3.RC1" to help me remember to change this default value to `reduce-overhead`

XDaoHong: done

python/sglang/srt/layers/communicator.py
```diff
@@ -135,8 +137,13 @@ def _compute_mlp_mode(cls, context: _LayerModeComputationContext):
                 ScatterMode.SCATTERED
                 if (
                     # Token dispatch/combine will be handled outside of LayerCommunicator for these modes.
-                    not get_moe_a2a_backend().is_none()
-                    or should_use_flashinfer_cutlass_moe_fp4_allgather()
+                    (
+                        not get_moe_a2a_backend().is_none()
```
ZhengdQin: revert this part

eshoguli (changes requested):
General comment about `NPUGraphRunner` updates. `NPUGraphRunner` is inherited from `CudaGraphRunner` and reuse capturing (`ForwardBatch` instantiation, initialization and capturing functionality) and partially replay pipeline from inherited type. `NPUGraph` python type is used. 

Your changes don't use anything from that. You have custom implementation of capturing and don't use `NPUGraph` python type for inference. As result, you need to implement separate runner: `TorchAirRunner` or something like that.

Can you, please, explain offline/online: why you need to use `NpuGraphRunner`? thanks!

PR #12374, opened 2025-10-30
optm(checkpoint-engine): disable multi-thread loading when update weights

BraveY:
For GPU tensors, using multi-threaded loading of weights may incur higher overhead due to the GIL.

Before optimization, the checkpoint-engine took 50s to update kimi-k2's weights on 2 nodes, and after optimization, the update time dropped to 15s.

## Motivation
This pull request optimizes the `load_weights` method in `deepseek_v2.py` by improving how weights are loaded based on their device location (CPU vs GPU). The main enhancement is that asynchronous multi-threaded loading is now used only for CPU tensors to overlap data transfer, while GPU tensors are loaded synchronously to avoid unnecessary overhead.

Improvements to weight loading logic:

* Added logic to determine whether to use asynchronous multi-threading based on whether the tensor is on the CPU or GPU (`use_async_loading` is set depending on `loaded_weight.device.type`).
* Updated all relevant code paths to use asynchronous loading only for CPU tensors, and to load GPU tensors synchronously, reducing overhead and improving efficiency. 

## Checklist


stmatengss:
Does this issue only affect the loading performance of DeepSeek v2? I notice similar behavior in Longcat model: https://github.com/sgl-project/sglang/blob/621dfb88864c645ae4e6d41807c5ce6f78c33294/python/sglang/srt/models/longcat_flash.py#L880

BraveY:
> Does this issue only affect the loading performance of DeepSeek v2? I notice similar behavior in Longcat model:
> 
> https://github.com/sgl-project/sglang/blob/621dfb88864c645ae4e6d41807c5ce6f78c33294/python/sglang/srt/models/longcat_flash.py#L880

Thanks for the review,  fixed on Longcat

ShangmingCai (approved):
LGTM. Let me ping another reviewer to double-check on this. cc: @fzyzcjy (Not sure if it will affect weight offloading feature)

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -3527,9 +3520,14 @@ def load_weights(self, weights: Iterable[Tuple[str, torch.Tensor]], is_nextn=Fal
...
-                        executor.submit(weight_loader, param, loaded_weight, shard_id)
-                    )
+                    if use_async_loading:
+                        futures.append(
+                            executor.submit(
+                                weight_loader, param, loaded_weight, shard_id
+                            )
+                        )
+                    else:
+                        weight_loader(param, loaded_weight, shard_id)
```
fzyzcjy: nit: what about

```
def maybe_executor_submit(...):
  if use_async_loading: ...
  else: ...
```

to avoid repeat many times this pattern

BraveY: Good point! I will refactor later.

BraveY: done

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -3527,8 +3527,14 @@ def load_weights(self, weights: Iterable[Tuple[str, torch.Tensor]], is_nextn=Fal
                         continue
                     param = params_dict[name]
                     weight_loader = param.weight_loader
-                    futures.append(
-                        executor.submit(weight_loader, param, loaded_weight, shard_id)
+                    maybe_executor_submit(
```
fzyzcjy, line 3527: maybe use keyword args

python/sglang/srt/models/longcat_flash.py
```diff
@@ -849,11 +849,21 @@ def load_weights(self, weights: Iterable[Tuple[str, torch.Tensor]]):
...
+                func(*args, **kwargs)
+
         with concurrent.futures.ThreadPoolExecutor() as executor:
             futures = []
             params_dict = dict(self.named_parameters())
             weight_names = []
             for name, loaded_weight in weights:
+                # Use async multi-threading only for host memory (CPU) tensors to overlap data transfer.
+                # For device memory (GPU) tensors, multi-threading adds overhead without benefit.
+                use_async_loading = loaded_weight.device.type == "cpu"
```
fzyzcjy: wondering whether we should extract such logic unrelated to models to a common place instead of repeat it for N times (where N is number of models we support)

python/sglang/srt/models/longcat_flash.py
```diff
@@ -849,11 +849,21 @@ def load_weights(self, weights: Iterable[Tuple[str, torch.Tensor]]):
         )
         cached_a_proj = {} if fuse_qkv_a_proj else None
 
+        def maybe_executor_submit(executor, futures, use_async, func, *args, **kwargs):
+            """Submit task to executor if async loading is enabled, otherwise execute directly."""
+            if use_async:
+                futures.append(executor.submit(func, *args, **kwargs))
+            else:
+                func(*args, **kwargs)
+
```
fzyzcjy: same (though i will personally refactor a bit more but this may be acceptable)

fzyzcjy (commented):
LGTM w/ nits

python/sglang/srt/models/deepseek_v2.py
BraveY: fixed

python/sglang/srt/models/longcat_flash.py
BraveY: done. thanks for the review.

python/sglang/srt/models/longcat_flash.py
BraveY: done. thanks for the review.

stmatengss (approved):

python/sglang/srt/model_loader/utils.py
```diff
@@ -116,3 +117,43 @@ def post_load_weights(model: nn.Module, model_config: ModelConfig):
...
+    """
+    For host (CPU) tensors, using a threadpool can overlap H2D copies
+    and improve throughput. For device tensors, threading often adds overhead
+    (e.g., GIL contention) without benefit, so we do it synchronously.
+    """
+    try:
+        return weight.device.type == "cpu"
+    except Exception:
+        # Be conservative if device info is not available
+        return False
```
fzyzcjy: do you mean `getattr(weight, "device", None)` or sth like that

catching Exception may not be good

python/sglang/srt/model_loader/utils.py
```diff
@@ -116,3 +117,43 @@ def post_load_weights(model: nn.Module, model_config: ModelConfig):
...
+    - futures: a list collecting the submitted Future objects
+    - use_async: whether to submit to executor or run inline
+    - func: the callable to run
+    - func_args: positional args for the callable (defaults to empty tuple)
+    - func_kwargs: keyword args for the callable (defaults to empty dict)
+    """
+    if func_kwargs is None:
+        func_kwargs = {}
+    if use_async:
+        futures.append(executor.submit(func, *func_args, **func_kwargs))
```
fzyzcjy, line 157: to be honest I may personally write a MaybeExecutor that wraps either a real executor or a noop single thread thing. but that would be a larger change

fzyzcjy (commented):
tiny optional nits

python/sglang/srt/model_loader/utils.py
BraveY: Good point, already refactored.

python/sglang/srt/model_loader/utils.py
BraveY: Thanks for the review, really learned a lot from your advice. Agreed that `MaybeExecutor` would be cleaner.  I think maybe we could refactor this when more models use this pattern.

BraveY:
@ShangmingCai hi, can we merge this PR now?

ShangmingCai:
> @ShangmingCai hi, can we merge this PR now?

Let me check the b200 CI.

ShangmingCai:
> > > @ShangmingCai hi, can we merge this PR now?
> > 
> > 
> > Let me check the b200 CI.
> 
> @ShangmingCai Hi, is there anything else I can help ?

b200 CI seems to be broken all the time, let me ping someone to force merge this PR.

hnyls2002 (approved):

PR #12379, opened 2025-10-30
[UT] Destroy process group after broadcast to resolve port occupation issues in multi-server tests

galeselee:
Added cleanup step to destroy the process group after broadcasting.

## Motivation

Added a cleanup step to destroy the process group after broadcasting. This modification was made to resolve port occupation issues encountered when running multiple server tests with configurations like (tp=1, dp=1, server) and (tp=2, dp=2, server), ensuring proper resource management and preventing conflicts.

## Checklist

@zhaochenyang20

### DIFF
--- a/test/srt/rl/test_update_weights_from_distributed.py
+++ b/test/srt/rl/test_update_weights_from_distributed.py
@@ -188,6 +188,9 @@ def init_process_hf(
     print(f"[hf] {rank=} {broadcast_time=:.3f}s")
     param_queue.put(("broadcast_time", broadcast_time))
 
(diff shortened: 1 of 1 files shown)


zhaochenyang20 (approved):

zhaochenyang20:
Did you run it locally to test whether the destroy makes sense and the test is still correct?

galeselee:
> Did you run it locally to test whether the destroy makes sense and the test is still correct?

Yes, I have tested locally

Kangyan-Zhou (approved):

Issue #10839, opened 2025-09-24
[Feature] [router] HA - Lightweight State Layer + gRPC Mesh


tonyluj:
### Checklist

- [x] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 2. Please use English, otherwise it will be closed.

# Motivation

The router needs HA without external stores (no Redis/etcd) while keeping routing decision correct across replicas. Today, worker registration/health, tenant radix/policy, and (optionally) rate-limit state are scattered and/or tied to per-process memory with legacy admin APIs. We need a lightweight in-memory state layer plus a replica mesh that converges quickly, letting any replica accept writes and all replicas converge-with no leader or external DB.

# Goals

* Run multiple router replicas to achieve HA without external components.
* Correct routing under churn, model-aware selection reads a near-real-time shared view.
* State and LB management and sync: maintain ratelimit, radix tree, models / workers (healthy / load) / routers states and sync across replicas with gRPC.
* Ensure eventual consistency with fast convergence, in exchange for higher availability (AP in CAP).

## Non-Goals

* Strong/linearizable consensus (Raft/etcd).
* Persisting all state across restarts (optional WAL can come later).
* Implementing full rate-limit / load balance accuracy (soft drift is acceptable).

# Architecture Overview

<img width="2283" height="1494" alt="Image" src="https://github.com/user-attachments/assets/280c972c-6434-4ab2-8e54-fc0349651bee" />

The design consists of three core components:
* Config Endpoint: Exposes endpoints for external interaction (e.g., updates, health checks).
* CRDT: Manages consensus and state replication across nodes.
* States Store: Stores and manages four types of state: Membership, App, Worker, and Policy.

## Consistency Model

Eventual consistency, fast convergence. Using CRDT Map<SKey, LWWReg<Bytes>> (via rust-crdt), and PNCounter for rate-limit and LB aggregates. Merge is idempotent & order-independent; snapshot are plain merge.

# Mesh Protocol

Transport: gRPC, bidirectional streaming, mTLS (optional)

## Incremental stream (steady-state)

* Stream op-based Map::update operations in batches (actor-ordered per peer)
* Gap or new replica -> state-based snapshot (per-store chunk) and merge.
Snapshot stream (on gap/new join)
* Per-store chunking (fixed entries or size); receivers apply CRDT merge.
[...]

PR #12385, opened 2025-10-30
[router] add ha server for router

llfl:
## Motivation

The router needs HA without external stores (no Redis/etcd) while keeping routing decision correct across replicas.  Details can be found in #10839 

## Modifications

This PR implements a high-availability (HA) server based on the gossip protocol, which the approach described in the paper at https://ieeexplore.ieee.org/document/1028914/.  

The HA server provides eventual consistency across multiple nodes and supports failure detection of faulty nodes. This change also includes test cases; you can observe the functionality by running:
`cargo test --package sglang_router_rs --lib -- ha::service::tests --nocapture`

## Checklist


sgl-router/src/ha/service.rs
tonyluj, line 1: Consider using ha as module name, instead of ha_server. ha_server tends to be a single server. actually it has both client and server.

msharmavikram:
Wow. I admire the work done here! Gossip protocol is not easy, and I will let Prof Indranil Gupta know about this work! He will be delighted to see this effort!

sgl-router/src/main.rs
```diff
@@ -318,6 +320,21 @@ struct CliArgs {
 
     #[arg(long)]
     mcp_config_path: Option<String>,
+
+    #[arg(long, default_value_t = false)]
+    ha_enable: bool,
```
slin1237: `enable_ha`

slin1237: please add those to python binding too

PR #12396, opened 2025-10-30
[Feature] Trace: Support http/protobuf span exporter protocol

zhanghaotong:
## Motivation

SGLang already supports tracing with OpenTelemetry. We want to use this feature in production, but our collector requires an HTTP/protobuf-based span exporter. This PR adds support for an HTTP/protobuf span exporter to enable production use and integration with our collector.

## Modifications

This PR:

- Reads the OTEL_EXPORTER_OTLP_TRACES_PROTOCOL environment variable to determine which protocol to use (defaults to "grpc")
- Validates that the specified protocol is supported (only "grpc" and "http/protobuf" are supported)
- Creates and returns the corresponding OTLPSpanExporter instance:
    - For "grpc" protocol: Uses the gRPC exporter with insecure connection
    - For "http/protobuf" protocol: Uses the HTTP protobuf exporter

This allows the tracing system to be flexible and support different OTLP protocols based on environment configuration, making it easier to integrate with various tracing backends like Jaeger, Zipkin, or other OpenTelemetry collectors.

## Checklist

### DIFF
--- a/python/sglang/srt/tracing/trace.py
+++ b/python/sglang/srt/tracing/trace.py
@@ -37,6 +37,12 @@ tracing_enabled = False
 
 try:
     from opentelemetry import context, propagate, trace
+    from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import (
+        OTLPSpanExporter as GRPCSpanExporter,
+    )
+    from opentelemetry.exporter.otlp.proto.http.trace_exporter import (
+        OTLPSpanExporter as HTTPSpanExporter,
+    )
     from opentelemetry.sdk.environment_variables import (
         OTEL_EXPORTER_OTLP_TRACES_PROTOCOL,
     )
@@ -238,17 +244,9 @@ def get_otlp_span_exporter(endpoint):
         )
 
     if protocol == "grpc":
-        from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import (
-            OTLPSpanExporter,
-        )
-
-        return OTLPSpanExporter(endpoint=endpoint, insecure=True)
+        return GRPCSpanExporter(endpoint=endpoint, insecure=True)
     elif protocol == "http/protobuf":
-        from opentelemetry.exporter.otlp.proto.http.trace_exporter import (
-            OTLPSpanExporter,
-        )
-
-        return OTLPSpanExporter(endpoint=endpoint)
+        return HTTPSpanExporter(endpoint=endpoint)
 
 
 # Should be called by each tracked thread.
--- a/python/sglang/srt/tracing/trace.py
+++ b/python/sglang/srt/tracing/trace.py
@@ -37,7 +37,9 @@ tracing_enabled = False
 
 try:
     from opentelemetry import context, propagate, trace
-    from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import OTLPSpanExporter
+    from opentelemetry.sdk.environment_variables import (
+        OTEL_EXPORTER_OTLP_TRACES_PROTOCOL,
+    )
     from opentelemetry.sdk.resources import SERVICE_NAME, Resource
     from opentelemetry.sdk.trace import TracerProvider, id_generator
     from opentelemetry.sdk.trace.export import BatchSpanProcessor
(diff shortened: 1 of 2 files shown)


zhanghaotong:
@sufeng-buaa Could you please help review this PR?

sufeng-buaa:
> @sufeng-buaa Could you please help review this PR?

ok, I will review it as soon.

python/sglang/srt/tracing/trace.py
```diff
@@ -225,6 +227,30 @@ def process_tracing_init(otlp_endpoint, server_name):
...
 
+def get_otlp_span_exporter(endpoint):
+    protocol = os.environ.get(OTEL_EXPORTER_OTLP_TRACES_PROTOCOL, "grpc")
+    supported_protocols = {"grpc", "http/protobuf"}
+
+    if protocol not in supported_protocols:
+        raise ValueError(
+            f"Unsupported OTLP protocol '{protocol}' configured. "
+            f"Supported protocols are: {', '.join(sorted(supported_protocols))}"
+        )
```
sufeng-buaa, line 244: I think the debug feature configuration should not terminate the entire engine. @ShangmingCai take a look, what do you think about it?

ShangmingCai: If this only happens during engine start, it is OK to raise `ValueError` and crash to inform the user of the incorrect configuration. Just make sure the default "grpc" is runnable without further dependencies, and `get_otlp_span_exporter` should not be called during inference.

zhanghaotong: > If this only happens during engine start, it is OK to raise `ValueError` and crash to inform the user of the incorrect configuration. Just make sure the default "grpc" is runnable without further dependencies, and `get_otlp_span_exporter` should not be called during inference.

Agreed. Now we use tracing as a production feature (keeping only the main request span to minimize uploads), we should fail fast and provide clear guidance for correcting configuration errors. The code will run only during tracing setup at engine start and will never be invoked on the inference path~

sufeng-buaa: ok，let me test it on my local machine.

python/sglang/srt/tracing/trace.py
```diff
@@ -225,6 +227,30 @@ def process_tracing_init(otlp_endpoint, server_name):
...
+            OTLPSpanExporter,
+        )
+
+        return OTLPSpanExporter(endpoint=endpoint, insecure=True)
+    elif protocol == "http/protobuf":
+        from opentelemetry.exporter.otlp.proto.http.trace_exporter import (
+            OTLPSpanExporter,
+        )
+
+        return OTLPSpanExporter(endpoint=endpoint)
```
sufeng-buaa: Need to add guidance on how to configure the endpoint for HTTP protocol to `docs/references/production_request_trace.md`(--otlp-traces-endpoint http://\<otlp server ip\>\:4318/v1/traces)

python/sglang/srt/tracing/trace.py
```diff
@@ -225,6 +227,30 @@ def process_tracing_init(otlp_endpoint, server_name):
...
+    supported_protocols = {"grpc", "http/protobuf"}
+
+    if protocol not in supported_protocols:
+        raise ValueError(
+            f"Unsupported OTLP protocol '{protocol}' configured. "
+            f"Supported protocols are: {', '.join(sorted(supported_protocols))}"
+        )
+
+    if protocol == "grpc":
+        from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import (
```
sufeng-buaa: The `import opentelemetry.exporter.otlp.proto.*` statements can be placed at the top of the file inside a try-except block, since the opentelemetry package might not be installed. Depending on the protocol, route the `OTLPSpanExporter` to different functions.

python/sglang/srt/tracing/trace.py
zhanghaotong: 👌 OK, I will add it.

python/sglang/srt/tracing/trace.py
zhanghaotong: Thanks for the suggestion, I have already replaced them.

ShangmingCai:
> Hi~ Could you please help review this PR? @ShangmingCai

@zhanghaotong Just been too busy, let me check now.

ShangmingCai (approved):
LGTM

PR #12407, opened 2025-10-30
[Entrypoint][Model-Gateway] Add GetTokenizer endpoint for gRPC server and load Tokenizer to registry from temp zip for model gateway

YouNeedCryDear:
## Summary

This PR adds a new GetTokenizer endpoint to the gRPC server, enabling clients to retrieve tokenizer information. Also adds a new GetTokenizer gRPC method to the SGLang scheduler service, enabling the model gateway to dynamically fetch tokenizer files from the backend via streaming then load into the tokenizer registry. This eliminates the need for static tokenizer configuration or local access to tokenizer files in the model gateway and ensures the router always uses the same tokenizer as the backend.

## Motivation

Currently, the SGLang model gateway requires to access the tokenizer files locally. This creates several challenges:

Configuration complexity: Users must manually ensure the model gateway has the same tokenizer files access matching the backend model
Deployment overhead: Tokenizer files must be distributed separately from the backend
Dynamic model support: The model gateway cannot adapt to backend model changes without reconfiguration
By enabling dynamic tokenizer streaming, the router can automatically obtain the correct tokenizer from the backend, simplifying deployment and ensuring consistency.

GetTokenizer endpoint will be used by sglang model gateway when inference gateway mode (multi model) is turned on. Router has no knowledge of what tokenizer should be used during start up and it will get tokenizer at runtime when workers are added into the registry. 

## Changes

On SGLang Server: 
- Added GetTokenizer RPC definition to sglang_scheduler.proto
- Implemented GetTokenizer endpoint in grpc_server.py with streaming support
- Generated updated protocol buffer bindings (pb2.py, pb2.pyi, pb2_grpc.py)
- Added comprehensive test coverage in test_tokenizer_stream.py

On SGLang Router: 
- Added new protobuf messages for tokenizer streaming
- Added get_tokenizer() method to SglangSchedulerClient
- Added logic to load tokenizer from streamed zip bundle into registry in worker registration

## Output
```
sglang git:(grpc-server-get-tokenizer) ✗ python3 ./test/srt/entrypoints/grpc_server/test_get_tokenizer.py
Sending GetTokenizer request...
Received metadata: meta.llama-3-8b
Received file chunk: tokenizer_bundle.zip, size: 2097152
Received file chunk: tokenizer_bundle.zip, size: 237091
Total zip size: 2334243 bytes
Extracting to /tmp/tmpm4e0l07n
Listing extracted files:
 - tokenizer_config.json
 - tokenizer.json
 - tokenizer.zip
[...]

PR #12408, opened 2025-10-30
[Router] Add GetTokenizer gRPC method to stream tokenizer files from backend to router

YouNeedCryDear:
## Summary

This PR adds a new `GetTokenizer` gRPC method to the SGLang scheduler service, enabling the router to dynamically fetch tokenizer files from the backend via streaming. This eliminates the need for static tokenizer configuration in the router and ensures the router always uses the same tokenizer as the backend.

## Motivation

Currently, the SGLang router requires static tokenizer files to be configured locally. This creates several challenges:

1. **Configuration complexity**: Users must manually ensure the router has the correct tokenizer files matching the backend model
2. **Deployment overhead**: Tokenizer files must be distributed separately from the backend
3. **Dynamic model support**: The router cannot adapt to backend model changes without reconfiguration

By enabling dynamic tokenizer streaming, the router can automatically obtain the correct tokenizer from the backend, simplifying deployment and ensuring consistency.

Python server implementation of the `GetTokenizer` RPC in https://github.com/sgl-project/sglang/pull/12407

### Protocol Definition

Added new protobuf messages for tokenizer streaming:

- `GetTokenizerRequest`: Request message (empty for now)
- `GetTokenizerChunk`: Streamed response containing either metadata or file chunks
- `TokenizerMetadata`: Information about the tokenizer bundle (model identifier, fingerprint, file list, format)
- `TokenizerFileDescriptor`: Metadata for individual tokenizer files
- `TokenizerFileChunk`: Chunked file data with sequential indexing and compression support

The protocol follows a metadata-first streaming pattern:
1. First message contains metadata about the tokenizer bundle
2. Subsequent messages contain file chunks with sequential indices
3. Last chunk is marked with `is_last_chunk` flag

### Rust Client Implementation

Added `get_tokenizer()` method to `SglangSchedulerClient` that:

- Initiates a streaming RPC call to the backend
- Validates the streaming protocol (metadata first, sequential chunks, last chunk marker)
- Accumulates compressed data chunks
- Returns a `TokenizerBundle` containing metadata and compressed data

The implementation includes robust error handling for protocol violations:
- Validates metadata arrives first
- Ensures chunks have sequential indices
- Verifies the last chunk marker is received
- Provides detailed error messages for debugging

### Future Work

This PR lays the foundation for dynamic tokenizer streaming. Future enhancements will include:

- `TokenizerRegistry` for caching downloaded tokenizers
- Integration with the router's worker registration workflow
- Configuration options for remote vs. static tokenizer sources
- Automatic fallback mechanisms

## Testing

Manual testing was performed to verify:
- Proto definitions compile correctly
- Rust client compiles without errors
- Protocol validation logic works as expected

Integration testing will be added once the Python server PR is merged.

### DIFF
--- a/sgl-router/src/grpc_client/sglang_scheduler.rs
+++ b/sgl-router/src/grpc_client/sglang_scheduler.rs
@@ -116,6 +116,13 @@ impl futures::Stream for AbortOnDropStream {
     }
 }
 
+/// Tokenizer bundle downloaded from backend
+#[derive(Debug, Clone)]
+pub struct TokenizerBundle {
+    pub metadata: proto::TokenizerMetadata,
+    pub compressed_data: Vec<u8>,
+}
+
 /// gRPC client for SGLang scheduler
 #[derive(Clone)]
 pub struct SglangSchedulerClient {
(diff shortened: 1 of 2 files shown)


CatherineSue:
> Configuration complexity: Users must manually ensure the router has the correct tokenizer files matching the backend model

This is FALSE - Router already has create_tokenizer_async() that handles:
  - HuggingFace model IDs: "deepseek-ai/deepseek-v3" → auto-downloads
  - Local paths: "/raid/models/..." → loads directly
  - No manual file management needed

Adding this management in grpc server is even more complexed

> Deployment overhead: Tokenizer files must be distributed separately from the backend

Can you explain a bit more? I thought we would share the same PersistentVolume for workers and routers.


> Dynamic model support: The router cannot adapt to backend model changes without reconfiguration

Not sure i understand this correctly. I thought same as #12045 , the root cause is because we need `--tokenizer-path` in start up right now. Why not having a TokenizerRegistry so we can solve all the issues once for all?

PR #12426, opened 2025-10-31
[fix] Support torch_dtype attribute in model config dtype detection

qichu-yun:
Fix dtype detection in _get_and_verify_dtype by checking torch_dtype attribute first before falling back to dtype attribute. This ensures proper handling of model configs that use torch_dtype (e.g., DeepSeek R1/V3) while maintaining backward compatibility with configs that use dtype attribute.

### DIFF
--- a/python/sglang/srt/configs/model_config.py
+++ b/python/sglang/srt/configs/model_config.py
@@ -821,7 +821,7 @@ def _get_and_verify_dtype(
 ) -> torch.dtype:
     # NOTE: getattr(config, "torch_dtype", torch.float32) is not correct
     # because config.torch_dtype can be None.
-    config_dtype = getattr(config, "dtype", None)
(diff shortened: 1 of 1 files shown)


qichu-yun:
Could you please review this PR when you're free, thanks! ? @wuhuikx @zejunchen-zejun @HaiShaw

qichu-yun:
Hi @merrymercy,  thank you for your contributions. I've noticed you changed model configuration dtype detection logic from "torch_dtype" to "dtype".  However, when running the DeepSeek-V3/R1 model, this causes it to be cast to fp16, which in my case leads to a crash during MoEGate computation. To resolve this, I've submitted a PR that implements a fallback to torch_dtype within the dtype detection process. I'd appreciate it if you could take a look. Thank you for your time and consideration.

PR #12430, opened 2025-10-31
Add env var to disable FA4 warmup

cicirori:
@Fridge003

### DIFF
--- a/sgl-kernel/python/sgl_kernel/_fa4_interface.py
+++ b/sgl-kernel/python/sgl_kernel/_fa4_interface.py
@@ -417,7 +417,7 @@ def warmup_flash_attn(f):
     - Executes sequentially to minimize peak GPU mem
     - Does not modify user tensors (clones)
     """
-    disable_warmup = os.getenv("SGL_KERNEL_DISABLE_FA4_WARMUP", "").lower() in (
+    disable_warmup = os.getenv("SGLANG_DISABLE_FA4_WARMUP", "").lower() in (
         "1",
         "true",
         "yes",
--- a/sgl-kernel/python/sgl_kernel/_fa4_interface.py
+++ b/sgl-kernel/python/sgl_kernel/_fa4_interface.py
@@ -8,6 +8,7 @@ import copy
 import gc
 import logging
 import math
+import os
 from typing import Callable, Optional, Tuple
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


sgl-kernel/python/sgl_kernel/_fa4_interface.py
```diff
@@ -416,6 +417,15 @@ def warmup_flash_attn(f):
     - Executes sequentially to minimize peak GPU mem
     - Does not modify user tensors (clones)
     """
+    disable_warmup = os.getenv("SGL_KERNEL_DISABLE_FA4_WARMUP", "").lower() in (
```
Fridge003: Please use `SGLANG` rather than "SGL" as its prefix.

Fridge003: Also this document needs to be updated 
https://github.com/sgl-project/sglang/blob/main/docs/references/environment_variables.md

cicirori: I’ve checked and this is the first time an environment variable is introduced in an sgl kernel, so it doesn’t use the `SGLANG_` prefix and isn’t managed through the current `sglang` env system. 
For now I’ll assume it’s under unified management and hope the SGL kernel environment variables don’t end up becoming a mess. XD

Fridge003: Well you are right.. We need to discuss with @zhyncs

sgl-kernel/python/sgl_kernel/_fa4_interface.py
```diff
@@ -416,6 +417,15 @@ def warmup_flash_attn(f):
     - Executes sequentially to minimize peak GPU mem
     - Does not modify user tensors (clones)
     """
+    disable_warmup = os.getenv("SGLANG_DISABLE_FA4_WARMUP", "").lower() in (
```
hnyls2002, line 420: Use environment variable from `sgalng.srt.environ`

hnyls2002 (changes requested):

sgl-kernel/python/sgl_kernel/_fa4_interface.py
Fridge003: This one is in sgl-kernel. Not able to import from sglang.srt.

sgl-kernel/python/sgl_kernel/_fa4_interface.py
hnyls2002: @cicirori I think we still need to use the common prefix `SGLANG_`. It doesn't matter whether this environment variable is managed by `sglang.srt.environ`; there are also so many env names starting with `SGL_` in sglang, and they are to be deprecated soon.

zhyncs (approved):

PR #12307, opened 2025-10-29
feat: support trtllm_mha FP8 query attention kernel

elvischenv:
## Motivation

TRTLLM_MHA already supports FP8-qkv BF16-out attention kernel. This will achieve better performance compared to original BF16-q FP8-kv kernel.
#9782

## Modifications

This PR convert query to FP8 dtype if kv cache type is also FP8.

## Accuracy Tests
`lm_eval --model local-completions --tasks gsm8k --model_args model=openai/gpt-oss-120b,base_url=http://127.0.0.1:18000/v1/completions,max_retries=3,tokenized_requests=False,timeout=1200,max_gen_toks=2048,max_length=8192 --batch_size 2048 --trust_remote_code --limit 0.5`

PR:
```
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8682|±  |0.0132|
|     |       |strict-match    |     5|exact_match|↑  |0.6091|±  |0.0190|
```

main:
```
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8394|±  |0.0143|
|     |       |strict-match    |     5|exact_match|↑  |0.6318|±  |0.0188|
```

## Benchmarking and Profiling
`python3 -m sglang.bench_serving --model openai/gpt-oss-120b --host 127.0.0.1 --port 18000 --backend sglang-oai --dataset-name random --random-range-ratio 1 --random-input-len 1024 --random-output-len 1024 --max-concurrency 512 --num-prompts 2560`

PR(5% perf improvement):
```
============ Serving Benchmark Result ============
Backend:                                 sglang-oai
Traffic request rate:                    inf
Max request concurrency:                 512
Successful requests:                     2560
Benchmark duration (s):                  153.18
Total input tokens:                      2621440
Total input text tokens:                 2621440
Total input vision tokens:               0
Total generated tokens:                  2621440
Total generated tokens (retokenized):    2552374
Request throughput (req/s):              16.71
Input token throughput (tok/s):          17113.94
Output token throughput (tok/s):         17113.94
Total token throughput (tok/s):          34227.89
Concurrency:                             510.27
[...]

PR #12437, opened 2025-10-31
Revert "fix llama4 kv cache layout"

b8zhong:
Because https://github.com/sgl-project/sglang/pull/12307 is actually the correct solution

+ Update FP8 column for trtllm mha.


Fridge003 (approved):
Cool

PR #12446, opened 2025-10-31
[CPU]  Optimize small oc GEMM for Qwen3-next on CPU

jianan-gu:
This PR adds optimization for BF16 GEMMs with small oc shapes ( oc < 16), including:

1. `weight_packed_linear`， which is common for all GEMMs with oc < 16
2. `fused_linear_sigmoid_mul`, which is specific for cases found in Qwen3-next (shared_expert_gate, where N = 1 and having post op sigmoid+mul)


python/sglang/srt/layers/amx_utils.py
```diff
@@ -54,6 +54,13 @@ def _amx_process_weight_after_loading(
                 f"The derived (OC, IC) dimensions must be divisible by (16, 32). "
             )
             module.use_intel_amx_backend = False
+            if weight_tensor.dim() == 2:
+                packed_weight = torch.nn.Parameter(weight_tensor.t().contiguous(), requires_grad=False)
+                packed_weight.__dict__ = weight_tensor.__dict__
+                setattr(module, weight_name, packed_weight)
+                logger.warning(
+                    f"Using transposed plain for shape {weight_tensor.shape} in {module}. "
+                )
```
mingfeima: i prefer that we hide the packing logic in 
* `convert_weight_packed` - if we have OC very small, then just return transposed weight.
* `weight_packed_linear` -f if we have OC very small, then call a separate gemm implementation.

python/sglang/srt/models/qwen3_next.py
```diff
@@ -393,6 +393,9 @@ def fix_query_key_value_ordering(self, mixed_qkvz, mixed_ba):
 
         return query, key, value, z, b, a
 
+    def _compile_ba(self, hidden_states: torch.Tensor):
+        return torch.ops.sgl_kernel.fma_linear(hidden_states, self.in_proj_ba.weight, None, None)
+
```
mingfeima: do not expose `fma_linear`, in python level we still call `weight_packed_linear`, just use a different C++ implementation

sgl-kernel/csrc/cpu/gemm.cpp
```diff
@@ -378,6 +398,137 @@ void weight_packed_linear_kernel_impl(
...
+    bVec m_bvec = bVec::loadu(mul + d);
+    fVec m_fvec0, m_fvec1;
+    std::tie(m_fvec0, m_fvec1) = at::vec::convert_to_float(m_bvec);
+    m_fvec0 = m_fvec0*X;
+    m_fvec1 = m_fvec1*X;
+
+    bVec out_vec = convert_from_float_ext<scalar_t>(m_fvec0, m_fvec1);
+    out_vec.store(out + d);
+  }
+}
```
mingfeima: template <typename scalar_t, bool has_bias>
inline void bias_sigmoid_mul(scalar_t* __restrict__ out, const float* __restrict__ input, const float* __restrict__ bias, const scalar_t* __restrict__ mul,  int SIZE) {
  if constexpr(has_bias) { assert(bias != nullptr); }
}

fuse these two functions.

sgl-kernel/csrc/cpu/gemm.cpp
```diff
@@ -378,6 +398,137 @@ void weight_packed_linear_kernel_impl(
...
+            copy_add_stub(C + m * ldc, Ctmp + m * N, bias, N);
+        }
+      } else {
+        if (post_mul != nullptr){
+            sigmoid_mul<scalar_t>(C + m * K, Ctmp + m * N, post_mul+m * K, K);
+        }else{
+            copy_stub(C + m * ldc, Ctmp + m * N, N);
+        }
+      }
+    }
```
mingfeima: use https://github.com/sgl-project/sglang/pull/12309/files#diff-e34a77a43f9515acc112174277210347a6ce5b64a875249d12b97875e27d0646R25

`AT_DISPATCH_BOOL2`

sgl-kernel/csrc/cpu/gemm.cpp
```diff
@@ -523,3 +674,54 @@ weight_packed_linear(at::Tensor& mat1, at::Tensor& mat2, const std::optional<at:
...
 }
+
+// mat1 : [M, K]
+// mat2 : [K, N]
+// bias : [N]
+// out  : [M, N]
+//
+at::Tensor
+fma_linear(at::Tensor& mat1, at::Tensor& mat2, const std::optional<at::Tensor>& bias, const std::optional<at::Tensor>& post_mul) {
+  RECORD_FUNCTION("sgl-kernel::fma_linear", std::vector<c10::IValue>({mat1, mat2, bias}));
```
mingfeima: don't make a new gemm api here, call it from `weight_packed_linear`

sgl-kernel/csrc/cpu/gemm.cpp
```diff
@@ -378,6 +398,137 @@ void weight_packed_linear_kernel_impl(
...
+            /* Ctmp*/ Ctmp,
+            /* bias*/ bias,
+            post_mul==nullptr? post_mul : post_mul + mb_start * K,
+            /*   M */ mb_size,
+            /*   N */ N,
+            /*   K */ K,
+            /* lda */ K,
+            /* ldb */ N,
+            /* ldc */ N,
+            /* brg */ true);
```
mingfeima: separate this function into two parts and it will align better with existing code:

* first part is gemm NN for fp32
template <bool has_bias, int BLOCK_M, int BLOCK_N>
struct tinygemm_kernel_nn<float, has_bias, BLOCK_M, BLOCK_N> {};

* second part handled post ops.

mingfeima (changes requested):
make as little change in the python level as possible,

make the changes happen in C++ level.

that's to say, we still use `weight_packed_linear`, but we use a different implementation for it under certain conditions.

mingfeima:
one more thing, using F32 brgemm might not be a good idea from performance point of view, but this kernel doesn't contribute too much in the end to end profiling results, we can keep it as it is right now.

python/sglang/srt/layers/amx_utils.py
jianan-gu: Sure, have refine the impl to reuse `convert_weight_packed ` and `weight_packed_linear`

python/sglang/srt/models/qwen3_next.py
jianan-gu: sure, have refined to use `weight_packed_linear`

sgl-kernel/csrc/cpu/gemm.cpp
jianan-gu: refined accordingly

sgl-kernel/csrc/cpu/gemm.cpp
jianan-gu: got it, have separated them into two parts

jianan-gu:
> lint: https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit

fixed.

mingfeima (approved):

FlamingoPg (approved):

PR #12449, opened 2025-10-31
Fix ci install to allow prerelease

merrymercy:


ShangmingCai (approved):
Although adding `--prerelease allow` in common `PIP_INSTALL_SUFFIX` is dangerous, we can fix CI throught this temporarily. LGTM.

PR #12478, opened 2025-10-31
Add the Nvidia-ModelOPT FP8 & FP4 quantization test case

jingyu-ml:
## Motivation

Add automated testing for ModelOpt quantization to ensure quality and performance of FP8 quantized models in SGLang.

## Modifications

- Added test/srt/quant/test_modelopt_fp8.py with test suite for ModelOpt FP8 quantization
- Implemented MOFP8Test base class for quantization testing with configurable thresholds

## Accuracy Tests

GSM8K accuracy threshold set at 69% for the FP8 quantized Llama-3.1-8B model. 90% for the fp8 quantized qwen3-8B model.

## Benchmarking and Profiling

Throughput test ensures minimum 120 tokens/s generation speed for both llama3.1-8B FP8 and Qwen3 (depends on the GPU type of the testing envir) .

## Checklist


test/srt/quant/test_modelopt_fp8.py
```diff
@@ -0,0 +1,121 @@
...
+        tok = time.perf_counter()
+        print(res["text"])
+        throughput = max_tokens / (tok - tic)
+        print(f"Throughput: {throughput} tokens/s")
+        self.assertGreaterEqual(throughput, self.throughput_threshold)
+
+
+# At the top of the file
+LLAMA_MODEL = "nvidia/Llama-3.1-8B-Instruct-FP8"
+LLAMA_GSM8K_ACC_THRESHOLD = 0.69
```
Fridge003: Is 0.69 expected? bf16 llama-3.1 should be about 0.79

test/srt/quant/test_modelopt_fp8_fp4.py
Edwardf0t1, line 1: I suggest creating separate test scripts for each quantization format.

Since we currently only support FP8 and FP4, keeping them in a single script is fine for now.

Edwardf0t1 (approved):
LGTM. Please address other comments as well.

Also, please update the description to include fp4 test info as well.

test/srt/quant/test_modelopt_fp8.py
jingyu-ml: I limited the number of questions to 200, so the score should naturally be lower than the 0.79. Let me change it to 1319 just like the dsr1.

test/srt/quant/test_modelopt_fp8_fp4.py
jingyu-ml: Sure, will do that later~

test/srt/quant/test_modelopt_fp8.py
jingyu-ml: Do you know which GPU is used for the CI/CD? I updated these numbers based on the b100 we had

PR #12504, opened 2025-11-02
To implement attention-sink

haotang1995:
## Motivation

To implement [attention sink](https://arxiv.org/pdf/2309.17453) with customized window sizes (first-window-size=1024 and last-window-size=7168) for efficient reasoning. When implemented with huggingface transformers, it saves memories and compute without hurting the downstream tasks' performance (such as pass@1 on AIME25 for math reasoning) for various LLMs such as Qwen3. I would like to implement this in sglang as well but failed as discussed below. I could not figure out why and it would be great if someone can help check it out. Any suggestions and advice are welcome. 

## Modifications

Following the [blog](https://hebiao064.github.io/fa3-attn-backend-basic) of FA3 implementation in sglang, I change the logic of `init_forward_metadata` so that the maximum seqlens become 8192 and the page-tables become the concatenation of the first-window and last-window.

## Accuracy Tests

The model behaves differently from expected (as implemented in huggingface transformers). It will not generate reasonable sentences after 8K tokens and instead repeat single words like ` \n\nSo \n\nSo \n\nSo \n\nSo \n\nSo \n\nSo \n\nSo \n\nSo \n\nSo \n\nSo`. 

I attached a script `test_attention_sink.sh` for testing. To launch servers, I run
```
python -m sglang.launch_server --model Qwen/Qwen3-8B --attention-backend fa3 --disable-cuda-graph
```
To submit a request, I run
```
curl http://localhost:30000/v1/chat/completions   -H "Content-Type: application/json"   -d '{
    "messages": [
        {"role": "user", "content": "There are $n$ values of $x$ in the interval $0<x<2\\pi$ where $f(x)=\\sin(7\\pi\\cdot\\sin(5x))=0$. For $t$ of these $n$ values of $x$, the graph of $y=f(x)$ is tangent to the $x$-axis. Find $n+t$. Put your answer in \\boxed{}. For multiple-choice questions, only put the choice letter in the box."}
    ],
    "max_tokens": 16384
}'
```

### DIFF
--- a/python/sglang/srt/layers/attention/flashattention_backend.py
+++ b/python/sglang/srt/layers/attention/flashattention_backend.py
@@ -444,17 +444,25 @@ class FlashAttentionBackend(AttentionBackend):
                     self.forward_metadata_spec_decode_expand = metadata_expand
             else:
                 # Normal Decode
+                seqlens_in_batch = torch.clamp(seqlens_in_batch, max=8192)
                 metadata.cache_seqlens_int32 = seqlens_in_batch.to(torch.int32)
-                metadata.max_seq_len_k = forward_batch.seq_lens_cpu.max().item()
+                metadata.max_seq_len_k = min(forward_batch.seq_lens_cpu.max().item(), 8192)
                 metadata.cu_seqlens_q = torch.arange(
                     0, batch_size + 1, dtype=torch.int32, device=device
                 )
                 metadata.cu_seqlens_k = torch.nn.functional.pad(
                     torch.cumsum(seqlens_in_batch, dim=0, dtype=torch.int32), (1, 0)
                 )
-                metadata.page_table = forward_batch.req_to_token_pool.req_to_token[
-                    forward_batch.req_pool_indices, : metadata.max_seq_len_k
-                ]
+                assert batch_size == 1, "Only batch size 1 is supported in decode mode for now"
+                if forward_batch.seq_lens_cpu.max().item() > 8192:
+                    metadata.page_table = torch.cat([
+                        forward_batch.req_to_token_pool.req_to_token[forward_batch.req_pool_indices, : 1024],
+                        forward_batch.req_to_token_pool.req_to_token[forward_batch.req_pool_indices, forward_batch.seq_lens_cpu.max().item() - 7168: forward_batch.seq_lens_cpu.max().item()]
+                    ], dim=1)
+                else:
+                    metadata.page_table = forward_batch.req_to_token_pool.req_to_token[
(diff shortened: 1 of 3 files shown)


hebiao064:
sorry i was very busy recently, just did brief look and this pr seems very easy to understand

I will do some research on attn sink and get back to you

python/sglang/srt/layers/attention/flashattention_backend.py
```diff
@@ -444,17 +444,25 @@ def init_forward_metadata(self, forward_batch: ForwardBatch):
                     self.forward_metadata_spec_decode_expand = metadata_expand
             else:
                 # Normal Decode
+                seqlens_in_batch = torch.clamp(seqlens_in_batch, max=8192)
```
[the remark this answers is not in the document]
hebiao064: this is a good call out, for now lets make it work for page size = 1 and ignore this msg

python/sglang/srt/layers/attention/flashattention_backend.py
```diff
@@ -444,17 +444,25 @@ def init_forward_metadata(self, forward_batch: ForwardBatch):
...
                     torch.cumsum(seqlens_in_batch, dim=0, dtype=torch.int32), (1, 0)
                 )
-                metadata.page_table = forward_batch.req_to_token_pool.req_to_token[
-                    forward_batch.req_pool_indices, : metadata.max_seq_len_k
-                ]
+                assert batch_size == 1, "Only batch size 1 is supported in decode mode for now"
+                if forward_batch.seq_lens_cpu.max().item() > 8192:
+                    metadata.page_table = torch.cat([
+                        forward_batch.req_to_token_pool.req_to_token[forward_batch.req_pool_indices, : 1024],
+                        forward_batch.req_to_token_pool.req_to_token[forward_batch.req_pool_indices, forward_batch.seq_lens_cpu.max().item() - 7168: forward_batch.seq_lens_cpu.max().item()]
```
hebiao064: this seems bit suspicious, we may need to keep the last 7168 tokens for each sequence.

Maybe you can start with for loop

haotang1995: Got it, I will test for-loop as well then. But the fact that the code does not work even with batch-size=1 makes it feel like there are other problems other than this. Thanks

hnyls2002:
Thanks @haotang1995 for exploring attention-sink support. As-is this PR is experimental (it includes a committed backup copy flashattention_backend_bkp.py and hardcoded window-size constants) and has been inactive since November, so I am closing it to keep the queue tidy. If you plan to productionize this (single backend file, configurable windows, tests), please reopen or open a fresh PR.

Issue #10492, opened 2025-09-16
Reducing the time to import sglang


merrymercy:
Now importing sglang takes too long. On my H100 machine, it took 12s to just import sglang.
Can someone help reducing the importing time?

```
> time python -c "from sglang.srt.managers.scheduler import Scheduler"
All deep_gemm operations loaded successfully!
W0916 03:22:46.348000 152559 torch/utils/cpp_extension.py:2425] TORCH_CUDA_ARCH_LIST is not set, all archs for visible cards are included for compilation.
W0916 03:22:46.348000 152559 torch/utils/cpp_extension.py:2425] If this is not desired, please set os.environ['TORCH_CUDA_ARCH_LIST'] to specific architectures.

real	0m12.310s
user	0m13.728s
sys	0m23.912s
```

zhyncs:
I also encountered a similar issue on H200.

ping1jing2:
I'm interested in this task, hope to solve it.

vincent-shop:
Most of it is `import torch` and `openai` lib imports, along with a few others.
Also, the lazy import of anything from `transformers`, some PIL, type libs.
We should also have some kind of global lazy loader, but TBD the most effective way to implement this module.

ping1jing2:
sync information(still don't find the solution...)

test on Ascend(aarch64) sever
- SGlang-main(import_schedule_in_sglang_main):
   it costs 11s ≈ 2s(vllm) + 5.5s(torch) + 1s(deepseekvl2 + transformers) + 0.7s(transformers) + 0.7s(managers)
- SGlang-main(without_import_vllm):
   it costs 11s ≈ 1.5s(compressed_tensors) + 5.9s(torch) + 1.3s(fastapi + openai) + 0.9s(transformers) + 0.7s(managers)
- vllm-main:
   it costs 10s ≈ 1s(model_loader) + 5.5s(torch) + 1.5s(layers.quantization) + 0.7s(transformers)

**import_schedule_in_sglang_main**
![Image](https://github.com/user-attachments/assets/7ac9d48f-58d0-4cb3-b013-fd1575f05c38)
**import_schedule_without_vllm**
![Image](https://github.com/user-attachments/assets/3050760e-6350-45ab-a759-00cc3c5700d0)
**import_llmEngine_in_vllm_main**
![Image](https://github.com/user-attachments/assets/617809a7-c78d-41be-bfe2-89b3a982a776)

ping1jing2:
[...]

PR #12510, opened 2025-11-02
Reduce time it takes to import SGLang

raayandhar:
## Motivation

We notice that the time to import things in SGLang takes a lot of time (https://github.com/sgl-project/sglang/issues/10492). I have been looking into what is taking up a lot of time and if there are simple ways to help reduce this import time. From the original issue, we want to reduce:
```
time python -c "from sglang.srt.managers.scheduler import Scheduler"
```
which is what I have been focusing my efforts on. However, I think there are things we can do to reduce time for other imports. This is more of a V1 to get community feedback from experts.

## Modifications

There are some heavy imports.  For example, the quantization methods import at the module level is heavy. Moving some imports to the function level (only time it is used), we can reduce module import time. However, I can see how this can easily be an antipattern. In fact, it can hurt performance if we have a function that is used a lot that we have an import in. I tried to only do this in functions that we only expect to run once or a small number of times. However, I can understand the argument against this kind of code. I also don't think all the changes to `hf_transformer_utils.py` help so I will be taken a deeper look, since the changes are a bit invasive.

## Accuracy Tests

These changes should not affect model outputs.

## Benchmarking and Profiling

Running `for i in {1..100}; do (time python -B -c "from sglang.srt.managers.scheduler import Scheduler") 2>&1 | grep "^real"; done | python calc_avg.py` ([calc_avg.py](https://gist.github.com/raayandhar/ed637053e7c311630ed8e8cd989ce947))

With these changes:
```
===========Timing Statistics============                                                                            
Number of runs: 100                                                                                                 
Mean:     8.308s                                                                                                    
Median:   8.236s                                                                                                    
Std Dev:  0.297s                                                                                                    
Min:      7.999s                                                                                                    
Max:      9.329s                                                                                                    
========================================
```
Compared to top-of-main:
```
===========Timing Statistics============                                                                            
Number of runs: 100                                                                                                 
Mean:     9.836s                                                                                                    
Median:   9.790s                                                                                                    
Std Dev:  0.332s                                                                                                    
Min:      8.655s                                                                                                    
Max:      11.801s                                                                                                   
======================================== 
```
so we have ~1.5 second improvement. Not the best, so I am going to keep working on it. I mostly targeted improving the timing in the creation of `ModelConfig` object. The difference so far is largely from removing the quantization import:

Without these changes [import_sglang_tom.log](https://gist.github.com/raayandhar/d25a90632e3502f9f8791db333ad0b02)
Screenshot 2025-11-01 at 9 03 47 PM

With these changes [import_sglang_new.log](https://gist.github.com/raayandhar/0d851737cf7eae3218fca29c8607384a)
Screenshot 2025-11-01 at 9 04 04 PM

Machine:
- AMD EPYC 7343 16-Core Processor
- L40S GPU

## Checklist

- [x] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests). (N/A?)
- [x] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations). (N/A?)

### DIFF
--- a/python/sglang/srt/disaggregation/decode.py
+++ b/python/sglang/srt/disaggregation/decode.py
@@ -67,7 +67,7 @@ from sglang.srt.utils.torch_memory_saver_adapter import TorchMemorySaverAdapter
 logger = logging.getLogger(__name__)
 
 if TYPE_CHECKING:
-    from sglang.srt.managers.schedule_batch import Req
+    from sglang.srt.managers.schedule_batch import Req, ScheduleBatch
     from sglang.srt.managers.scheduler import Scheduler
 
 CLIP_MAX_NEW_TOKEN = get_int_env_var("SGLANG_CLIP_MAX_NEW_TOKENS_ESTIMATION", 4096)
@@ -229,6 +229,8 @@ class DecodePreallocQueue:
         self.kv_manager = self._init_kv_manager()
 
     def _init_kv_manager(self) -> BaseKVManager:
+        from sglang.srt.layers.dp_attention import get_attention_tp_size
+
         kv_args_class = get_kv_class(self.transfer_backend, KVClassType.KVARGS)
         kv_args = kv_args_class()
 
@@ -298,6 +300,8 @@ class DecodePreallocQueue:
 
     def add(self, req: Req, is_retracted: bool = False) -> None:
         """Add a request to the pending queue."""
+        from sglang.srt.managers.schedule_batch import RequestStage
+
         if self._check_if_req_exceed_kv_capacity(req):
             return
 
@@ -412,6 +416,8 @@ class DecodePreallocQueue:
 
     def pop_preallocated(self) -> List[DecodeRequest]:
         """Pop the preallocated requests from the pending queue (FIFO)."""
+        from sglang.srt.managers.schedule_batch import FINISH_ABORT, RequestStage
+
         self._update_handshake_waiters()
 
         preallocated_reqs = []
@@ -727,6 +733,8 @@ class DecodeTransferQueue:
         decode_req.req.time_stats.wait_queue_entry_time = time.perf_counter()
 
     def pop_transferred(self) -> List[Req]:
+        from sglang.srt.managers.schedule_batch import RequestStage
+
         if not self.queue:
             return []
         polls = poll_and_all_reduce(
@@ -788,6 +796,7 @@ class SchedulerDisaggregationDecodeMixin:
     @torch.no_grad()
     def event_loop_normal_disagg_decode(self: Scheduler):
         """A normal scheduler loop for decode worker in disaggregation mode."""
+        from sglang.srt.managers.schedule_batch import RequestStage
 
         while True:
             recv_reqs = self.recv_requests()
@@ -808,6 +817,8 @@ class SchedulerDisaggregationDecodeMixin:
 
     @torch.no_grad()
     def event_loop_overlap_disagg_decode(self: Scheduler):
+        from sglang.srt.managers.schedule_batch import RequestStage
+
         self.result_queue = deque()
         self.last_batch: Optional[ScheduleBatch] = None
 
@@ -888,6 +899,8 @@ class SchedulerDisaggregationDecodeMixin:
 
     def get_new_prebuilt_batch(self: Scheduler) -> Optional[ScheduleBatch]:
         """Create a schedulebatch for fake completed prefill"""
+        from sglang.srt.managers.schedule_batch import RequestStage, ScheduleBatch
+
         if self.grammar_queue:
             self.move_ready_grammar_requests()
 
--- a/python/sglang/srt/disaggregation/decode.py
+++ b/python/sglang/srt/disaggregation/decode.py
@@ -30,7 +30,6 @@ from typing import TYPE_CHECKING, List, Optional, Type, Union
 import torch
 from torch.distributed import ProcessGroup
 
-from sglang.srt.configs.mamba_utils import Mamba2CacheParams
 from sglang.srt.constants import GPU_MEMORY_TYPE_KV_CACHE
 from sglang.srt.disaggregation.base import BaseKVManager, BaseKVReceiver, KVPoll
 from sglang.srt.disaggregation.utils import (
@@ -47,9 +46,6 @@ from sglang.srt.disaggregation.utils import (
     prepare_abort,
 )
 from sglang.srt.managers.request_types import FINISH_ABORT, RequestStage
-from sglang.srt.mem_cache.allocator import BaseTokenToKVPoolAllocator
-from sglang.srt.mem_cache.base_prefix_cache import BasePrefixCache
-from sglang.srt.mem_cache.common import release_kv_cache
 from sglang.srt.mem_cache.memory_pool import (
     HybridLinearKVPool,
     HybridReqToTokenPool,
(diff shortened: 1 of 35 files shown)


raayandhar:
I will continue working on this, there is more improvements to be made.

python/sglang/srt/utils/hf_transformers_utils.py
```diff
@@ -13,78 +13,83 @@
...
+    )
 
-_CONFIG_REGISTRY = {
-    config_cls.model_type: config_cls for config_cls in _CONFIG_REGISTRY
-}
 
-for name, cls in _CONFIG_REGISTRY.items():
-    with contextlib.suppress(ValueError):
-        AutoConfig.register(name, cls)
+def _register_custom_configs():
```
raayandhar: I think that a lot of these changes in this file do not really improve the times, and they are overly invasive. However, the transformers related code is a huge time increase. I think they may be impossible to remove since we init many transformer related objects in the scheduler.

python/sglang/srt/disaggregation/decode.py
```diff
@@ -296,6 +296,8 @@ def _init_kv_manager(self) -> BaseKVManager:
 
     def add(self, req: Req, is_retracted: bool = False) -> None:
         """Add a request to the pending queue."""
+        from sglang.srt.managers.schedule_batch import RequestStage
```
raayandhar: This is a very repetitive pattern, but the `from sglang.srt.managers.schedule_batch` import is somewhat expensive. The `RequestStage` object is just an enum, so we don't want to be doing the massive import for just an enum in my opinion. Maybe we can put this enum in a different file / some other solution that effectively does the same thing, so we can just import at the top level.

python/sglang/srt/layers/moe/__init__.py
```diff
@@ -12,6 +11,10 @@
     is_tbo_enabled,
     should_use_flashinfer_cutlass_moe_fp4_allgather,
 )
+from sglang.utils import LazyImport
+
+MoeRunner = LazyImport("sglang.srt.layers.moe.moe_runner.runner", "MoeRunner")
```
raayandhar, line 16: I think this (`LazyImport(...)`)is something we can apply to many other files as well, I'm not sure if there's a downside?

python/sglang/srt/managers/scheduler.py
```diff
@@ -2123,6 +2122,8 @@ def prepare_mlp_sync_batch_raw(
         disable_overlap_schedule: bool,
         offload_tags: set[str],
     ):
+        from sglang.srt.layers.moe import TboDPAttentionPreparer
```
raayandhar: Maybe not worth it to lazy load this one here. It saves 600 ms in import time but this will get called a lot more in the event loop. Although I'm not sure because this may be cached.

zhyncs:
The drawback is that it introduces some extra cognitive load. May you fix the conflicts first? Thanks!

python/sglang/srt/disaggregation/decode.py
raayandhar: I have refactored this into a separate file for just the request types so as to avoid some cognitive load. I can revert the changes if needed.

python/sglang/srt/disaggregation/decode.py
```diff
@@ -884,6 +885,8 @@ def get_next_disagg_decode_batch_to_run(
 
     def get_new_prebuilt_batch(self: Scheduler) -> Optional[ScheduleBatch]:
         """Create a schedulebatch for fake completed prefill"""
+        from sglang.srt.managers.schedule_batch import ScheduleBatch
```
hnyls2002: You shouldn't do this... You imported `ScheduleBatch` twice here.

hnyls2002 (changes requested):

python/sglang/srt/disaggregation/decode.py
raayandhar: iiuc, this is because we import it at the top for the type checking, and then here we use it (`ScheduleBatch.init_new`, only function in this file that doesn't use it as a type) so we need to import it for use in this function. But I can leave it at top-level import if you think that's better and remove from type checking.

python/sglang/srt/layers/moe/__init__.py
hnyls2002: not sure. @merrymercy @fzyzcjy What are your opinions?

hnyls2002 (changes requested):

raayandhar:
> Some principles:
> 
> 1. Do not import inside a Python code block (function, class definition, if block...); import at the top of a file in most cases.
> 2. If an imported item works as a util function, you can import it when using it (inside a code block)
> 3. Place all type-hint-only class types under `if TYPE_CHECKING:`
> 
> @raayandhar

The worry I have is that some imports are only used in a single function across the entire file, and this function may also only be called once / very few times. For example, the `def _verify_quantization(self)` function is the only one that needs `QUANTIZATION_METHODS`, and importing this at the top level is very expensive. Maybe we can also consider these to be exception? Otherwise, we won't be reducing the time very much. Appreciate discussion on this matter. @hnyls2002

hnyls2002 (approved):

hnyls2002:
Please fix the conflicts. @raayandhar 

Also I think this PR should be review by @merrymercy

raayandhar:
I investigated the previous CI errors and I think they be unrelated, to my best understanding, but I will monitor carefully again.
The changes are a bit extensive across a lot of different files. I think one strategy I think of is for when importing heavy classes, using `LazyImport` and keep it at top-level, to avoid cognitive load (and perhaps note that the import is expensive, so that is why we do it). When importing a util function with a heavy path, if it is just used once in the whole file, keep it at the function level. And then of course when only used for type checking, leave it under a`TYPE_CHECKING` block. I already do the latter two, but can update my PR to do the former. I think this is close in spirit to the earlier review. I think that might be the cleanest way to handle this. Happy to discuss further.

raayandhar:
I've refactored the PR in the following way, based on earlier discussion. I liked the idea of having simple rules like @hnyls2002 mentioned:

> Some principles:
> 
> 1. Do not import inside a Python code block (function, class definition, if block...); import at the top of a file in most cases.
> 2. If an imported item works as a util function, you can import it when using it (inside a code block)
> 3. Place all type-hint-only class types under `if TYPE_CHECKING:`

I've refactored the PR so we do this. Now, in order to actually make the import times decrease, we need to avoid doing very expensive imports at import time. So I propose a fourth principle:

4. If a class import (i.e. non-util function) is very expensive, use `LazyImport` at the top/module level, use sparingly.

This helps us avoid the large import times. The only downside I see is that there is some cognitive load as mentioned earlier, but I think this is a better solution than importing at the function level. This also helps avoid some circular import issues, which could potentially be caused from the way I was doing it earlier.

Here are the most recent timings. On this branch (range 7-8.5 seconds):
```
(sglang) raayan@uril-1:~/projects/sglang$ time python -c "from sglang.srt.managers.scheduler import Scheduler"

real	0m7.469s
user	0m15.162s
sys	0m3.141s
```
On main (range anywhere from 10-15 seconds, usually 10.5-12 seconds):
```
(sglang) raayan@uril-1:~/projects/sglang$ time python -c "from sglang.srt.managers.scheduler import Scheduler"

real	0m10.981s
user	0m21.909s
sys	0m4.143s
```
At this point, we can't do much better. Looking at the import time heatmap, it is nearly all torch now, at 4.5 ms:
<details>
  <summary>this branch branch import time graph (click to expand)</summary>

importtime_feat

I recommend clicking on the image and expanding.
</details>

whereas main is at nearly 7 ms:
<details>
  <summary>main branch import time graph (click to expand)</summary>

importtime_main

I recommend clicking on the image and expanding.
</details>

which gives us a ~33% decrease.

PR #12527, opened 2025-11-03
[cpu] Implement all gather/reduce for arm64 cpu

cyb70289:
## Motivation

This PR fixes sglang cpu module build error on arm64 platform, which was hardcoded with x86 intrinsics.
Unit tests `test_all_reduce` and `test_all_gather` verified okay on Arm neoverse-n2 server.

NOTE: this is an initial PR to support running sglang on Arm servers.

## Modifications

- Updated sgl-kernel/csrc/cpu/shm.cpp to add arm64 SIMD implementation based on platform.
- Updated python/pyproject_cpu.toml to install platform dependent packages on x86 host only.

## Accuracy Tests

N/A

## Benchmarking and Profiling

N/A

## Checklist


cyb70289:
hi @mingfeima @merrymercy , will you (or someone familiar with cpu kernel code) to review this PR?

We are planning to add Arm64 CPU support for sglang. This PR fixes build issues on Arm by replacing hardcodes x86 intrinsics in shm.cpp with arch dependent code.

Current approach mixes x86/arm code in one source file with #ifdef. Do you think it's okay?
Or we'd better separate arm code to different files for clarity (there will be arm specific implementation of other files like gemm as well).

FlamingoPg:
Hi, i think is a good way to separate arm code to different files.

sgl-kernel/csrc/cpu/shm.cpp
```diff
@@ -1,9 +1,15 @@
 #include "shm.h"
+#if defined(__x86_64__)
```
cyb70289: include arch specific header

sgl-kernel/csrc/cpu/shm.cpp
```diff
@@ -154,241 +115,19 @@ void reduce_all_buffers(
     char** buffers) {
   switch (scalar_type) {
     case c10::ScalarType::BFloat16:
-      reduce_bf16_buffers(start_elements, num_elements, to_buffer, buffers);
+      reduce_bf16_buffers(start_elements, num_elements, to_buffer, buffers, world_size);
```
cyb70289, line 118: Added `world_size` argument, which was a global variable. Since the function is moved to another file, adding this argument makes the function more readable.

FlamingoPg:
Waiting for CI

cyb70289:
Looks CI failures are irrelevant?

FlamingoPg (approved):

mingfeima:
@cyb70289 do you have a RFC or roadmap for what's your plan next?

cyb70289:
> @cyb70289 do you have a RFC or roadmap for what's your plan next?

@mingfeima , our plan is to support running sglang on arm neoverse cpu. We will add arm64 optimized kernels under sgl-kernel/csrc/cpu (probably in a new subfolder), which currently only supports intel amx. Looks you are the author of current cpu implementation? We will include you to review our PR, in case some modification may conflict with your future code.

PR #12582, opened 2025-11-04
[sgl-kernel][Deepseek V3.2] Add row_starts to topk kernel

hlu1:
## Motivation

Part1 of the fix for bug in https://github.com/sgl-project/sglang/issues/11629

## Modifications

In the topk kernel for prefill, the q and k inputs are both ragged. We need to pass the correct start indices of k for each q token to the kernel.

In `fast_topk_transform_interface`, I changed the `is_decode` criterion from `prefill_bs == B` to `!row_starts_opt.has_value() and prefill_bs == B`. Using `prefill_bs == B` can be a bit ambiguous for a single token.

## Accuracy Tests
8xB200, fp8 checkpoint, bf16 kvcache, radix-cache on:
fp8 kvcache has similar score.

```
gsm8k

20 shots

Accuracy: 0.953
Invalid: 0.000
Latency: 23.981 s
Output throughput: 5388.611 token/s

30 shots

Accuracy: 0.928
Invalid: 0.000
Latency: 31.716 s
Output throughput: 3907.670 token/s

gpaq

Repeat: 8, mean: 0.790

['0.798', '0.778', '0.783', '0.803', '0.768', '0.813', '0.788', '0.788']

With mtp on

gsm 8k with 20 shots: 0.947
gsm 8k with 30 shots: 0.939
```

## Checklist

- [ ] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.ai/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.ai/developer_guide/contribution_guide.html#benchmark-the-speed).

Issue #12190, opened 2025-10-27
[Bug] Illegal memory access in Deepseek V3.2 with MTP on


hlu1:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [ ] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [ ] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [ ] 5. Please use English, otherwise it will be closed.

### Describe the bug

Without CUDA_LAUNCH_BLOCKING=1

```
File "/tmp/hlu/sglang/python/sglang/srt/managers/scheduler.py", line 2777, in run_scheduler_process
scheduler.event_loop_normal()
File "/usr/local/lib/python3.12/dist-packages/torch/utils/_contextlib.py", line 120, in decorate_context
return func(*args, **kwargs)
^^^^^^^^^^^^^^^^^^^^^
File "/tmp/hlu/sglang/python/sglang/srt/managers/scheduler.py", line 983, in event_loop_normal
batch = self.get_next_batch_to_run()
^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/tmp/hlu/sglang/python/sglang/srt/managers/scheduler.py", line 1624, in get_next_batch_to_run
new_batch = self.prepare_mlp_sync_batch(new_batch)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/tmp/hlu/sglang/python/sglang/srt/managers/scheduler.py", line 2039, in prepare_mlp_sync_batch
return self.prepare_mlp_sync_batch_raw(
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/tmp/hlu/sglang/python/sglang/srt/managers/scheduler.py", line 2106, in prepare_mlp_sync_batch_raw
local_info = torch.tensor(
^^^^^^^^^^^^^
torch.AcceleratorError: CUDA error: an illegal memory access was encountered
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1
[...]

PR #12583, opened 2025-11-04
[Deepseek V3.2] Fix accuracy bug in the Indexer

hlu1:
## Motivation

Must merge after https://github.com/sgl-project/sglang/pull/12582

Part 2 of the fix for https://github.com/sgl-project/sglang/issues/11629
This PR also fixed the mtp crash in https://github.com/sgl-project/sglang/issues/12190

First, it now passes the `ks` to the topk kernel. There is another bug in the Indexer where the `offset` used in constructing `ks` and `ke` should be two different offsets instead. Added a diagram to make it more clear.

## Accuracy Tests
Tested with https://github.com/sgl-project/sglang/pull/12582

```
On 8xB200

gsm8k

20 shots

Accuracy: 0.953
Invalid: 0.000
Latency: 23.981 s
Output throughput: 5388.611 token/s

30 shots

Accuracy: 0.928
Invalid: 0.000
Latency: 31.716 s
Output throughput: 3907.670 token/s

gpaq

Repeat: 8, mean: 0.790

['0.798', '0.778', '0.783', '0.803', '0.768', '0.813', '0.788', '0.788']

With mtp on

python -m sglang.launch_server --model $MODEL --tp 8 --dp 8 --enable-dp-attention --reasoning-parser deepseek-v3 --kv-cache-dtype fp8_e4m3 --speculative-algorithm EAGLE --speculative-num-steps 3 --speculative-eagle-topk 1 --speculative-num-draft-tokens 4 --cuda-graph-max-bs 256 --max-running-requests 2048

gsm 8k with 20 shots: 0.947
gsm 8k with 30 shots: 0.939

Repeat: 16, mean: 0.795
Scores: ['0.803', '0.823', '0.788', '0.798', '0.828', '0.768', '0.773', '0.783', '0.793', '0.773', '0.823', '0.783', '0.793', '0.808', '0.788', '0.793']

On 8xH200

Repeat: 8, mean: 0.802
Scores: ['0.798', '0.818', '0.788', '0.813', '0.798', '0.783', '0.823', '0.798']
```

## Checklist

### DIFF
--- a/python/sglang/srt/layers/attention/nsa/nsa_indexer.py
+++ b/python/sglang/srt/layers/attention/nsa/nsa_indexer.py
@@ -345,7 +345,10 @@ class Indexer(CustomOp):
         k_scale_list = []
         ks_list = []
         ke_list = []
-        offset = 0
+
+        q_offset = 0
+        k_offset = 0
+
         seq_lens_expanded = metadata.get_seqlens_expanded()
         block_tables = metadata.get_page_table_64()
 
@@ -368,13 +371,17 @@ class Indexer(CustomOp):
                 block_tables[i],
             )
             extend_seq_len = forward_batch.extend_seq_lens_cpu[i]
-            ks = torch.full((extend_seq_len,), offset, dtype=torch.int32, device="cuda")
-            ke = ks + seq_lens_expanded[offset : offset + extend_seq_len]
+            ks = torch.full(
+                (extend_seq_len,), k_offset, dtype=torch.int32, device="cuda"
+            )
+            ke = ks + seq_lens_expanded[q_offset : q_offset + extend_seq_len]
             k_fp8_list.append(k_fp8)
             k_scale_list.append(k_scale)
             ks_list.append(ks)
             ke_list.append(ke)
-            offset += extend_seq_len
+
+            q_offset += extend_seq_len
+            k_offset += seq_len
 
         k_fp8 = torch.cat(k_fp8_list, dim=0).view(torch.float8_e4m3fn)
         k_scale = torch.cat(k_scale_list, dim=0).view(torch.float32).squeeze(-1)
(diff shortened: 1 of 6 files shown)


python/sglang/srt/layers/attention/nsa/nsa_indexer.py
```diff
@@ -321,7 +321,10 @@ def _get_topk_ragged(
         k_scale_list = []
         ks_list = []
         ke_list = []
-        offset = 0
+
+        q_offset = 0
```
Fridge003, line 349: We might need to handle `target_verify` in _get_topk_paged?
Can be left for the next PR

hlu1: `target_verify` invokes `_get_topk_paged`. `draft_enxtend` invokes `_get_topk_ragged`, and in this case, it's not different from the normal extend case in terms of ks/ke computation.

Fridge003 (approved):

zhyncs (approved):

hlu1:
> @hlu1 https://github.com/sgl-project/sglang/actions/runs/19249589320/job/55042821747?pr=12583 The accept length dropped a little bit

2.84 is only slightly lower than 2.9 and the same test passed on 8xB200 with fp8 kvcache. On H200, the default is bf16 kvcache, so there can be some variance. Also, the old acceptance rate, 2.9, is based on an incorrect implementation. The acceptance rate can go either way when the bug is fixed. I think we should lower the acceptance rate a bit so the test can be more stable.

Fridge003:
@hlu1 Agree. Can you please lower the required acc length from 2.9 to a slightly lower value, like 2.8 or 2.7?

test/srt/test_deepseek_v32_mtp.py
```diff
@@ -58,7 +58,7 @@ def test_a_gsm8k(
         requests.get(self.base_url + "/flush_cache")
 
         args = SimpleNamespace(
-            num_shots=8,
+            num_shots=20,
```
Fridge003:
```suggestion
            num_shots=8,
```

PR #12584, opened 2025-11-04
Add native sglang backend for OpenPangu Embedded

YoussefEssDS:
## Motivation
Natively support the OpenPangu model to avoid the low throughput caused by the transformers fallback.  

## Modifications
- Add a SGLang backend for the OpenPangu Embedded 7B model (https://huggingface.co/FreedomIntelligence/openPangu-Embedded-7B-V1.1), so it no longer relies on the transformers fallback.
- Add `FreedomIntelligence/openPangu-Embedded-7B-V1.1` to the list of supported generative models.

## Benchmarking and Profiling
Start server with: 
`python -m sglang.launch_server --model FreedomIntelligence/openPangu-Embedded-7B-V1.1 --dtype bfloat16 --host 0.0.0.0 --port 18051 --tp 4`
Run performance benchmark: (Used the script here: https://gist.github.com/YoussefEssDS/03c456ef5fed6f24e27394d2b2f2cc05)

Significant improvement in throughput (TP=4 on nvidia H100 GPUs): SGLang backend ~298.83T/s vs. transformers fallback ~50.8 T/s.

## Checklist


python/sglang/srt/models/pangu.py
```diff
@@ -0,0 +1,74 @@
...
+SGLang's LLaMA implementation already supports both behaviours, so we reuse
+the optimized kernels and infrastructure and expose an explicit
+`PanguEmbeddedForCausalLM` entry point that plugs into the model registry.
+"""
+
+from __future__ import annotations
+
+from typing import Optional
+
+from transformers import PretrainedConfig
```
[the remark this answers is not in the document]
YoussefEssDS: Hugging Face loads `configuration_openpangu_dense.PanguEmbeddedConfig`, and that class inherits directly from `PretrainedConfig`, not from `LlamaConfig`.

YoussefEssDS:
@mickqian can you please take a look? Thanks!

YoussefEssDS:
The latest Gemini review is incorrect check -> https://huggingface.co/FreedomIntelligence/openPangu-Embedded-7B-V1.1/blob/main/config.json

YoussefEssDS:
@zhyncs @merrymercy @hubertlu-tw can you please help review this PR? thanks

python/sglang/srt/models/pangu.py
```diff
@@ -0,0 +1,66 @@
...
```
ping1jing2: To reduce the import time, please use the following code

```python
from typing import Optional, TYPE_CHECKING
if TYPE_CHECKING:
    from transformers import PretrainedConfig
    from sglang.srt.layers.quantization.base_config import QuantizationConfig

```

python/sglang/srt/models/pangu.py
```diff
@@ -0,0 +1,66 @@
...
+
+from sglang.srt.layers.quantization.base_config import QuantizationConfig
+from sglang.srt.models.llama import LlamaForCausalLM, LlamaModel
+
+
+class PanguEmbeddedModel(LlamaModel):
```
ping1jing2: maybe you should use `PanguEmbeddedModel = LlamaModel` or add `__init__` for this class

PR #12585, opened 2025-11-04
Add multi-GPU configurations to nightly-test.yml

alisonshao:
Add 4-GPU, 8-GPU (H200/H20), and B200 configurations to nightly tests following pr-test.yml patterns.

### DIFF
--- a/.github/workflows/nightly-test.yml
+++ b/.github/workflows/nightly-test.yml
@@ -121,3 +121,73 @@ jobs:
         run: |
           cd test/srt
           python3 run_suite.py --suite nightly-1-gpu
+
+  nightly-test-4-gpu:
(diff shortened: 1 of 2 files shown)


Kangyan-Zhou (commented):
Can we add a dependency to build sgl_kernel as needed similar to pr-test.yaml?

.github/workflows/nightly-test.yml
```diff
@@ -121,3 +121,73 @@ jobs:
         run: |
```
Fridge003, line 121: Can you please increase the timeout-minutes of nightly-test-1-gpu
it can be set to a larger number (60 etc)

Kangyan-Zhou (approved):

PR #12588, opened 2025-11-04
[Feature] Initial block diffusion language model support

ClawSeven:
## Motivation

The current implementation focuses only on the minimal functionalities required for dLLM serving：

- [x] **Block-wise Diffusion Framework** main logic
- [x] Full **KV Cache** support for sequence management
- [x] Model integration for **LLaDA-2.0-mini/flash**
- [x] Support for **Custom Decoding Strategies** (e.g., implementing **low-confidence policy**)
- [x] Full **Streaming I/O** capability
- [x] **Tensor Parallelism (TP)** support
- [x] Model accuracy verification
 
Those feature will be implemented in seperate PRs in the future:
- [ ] **Batching** support (leveraging SGLang's core scheduling)
- [ ] **CUDA Graph** optimization
- [ ] Support radix cache
- [ ] Support overlap scheduling

##  Our From-Scratch Diffusion Language Models
LLaDA2.0-flash-preview: https://huggingface.co/inclusionAI/LLaDA2.0-flash-preview
LLaDA2.0-mini-preview: https://huggingface.co/inclusionAI/LLaDA2.0-mini-preview

## Contributor
**Tiwei Bie tiwei.btw@antgroup.com** made an equal contribution to this work. His efforts included designing the complete Diffusion Language Model (DLM) architecture and driving the implementation of the key functional features.


python/sglang/srt/layers/attention/flashinfer_backend.py
```diff
@@ -770,7 +770,7 @@ def forward_extend(
                     q.view(-1, layer.tp_q_head_num, layer.head_dim),
                     k.view(-1, layer.tp_k_head_num, layer.head_dim),
                     v.view(-1, layer.tp_v_head_num, layer.head_dim),
-                    causal=True,
+                    causal=causal,
```
Monstertail, line 780: I wonder whether it may break the correctness for some models with the following concerns: 

> 
> If we change it to causal = causal, then in the encoder–decoder model’s cross-attention layer:

> 
> The decoder’s self-attention will incorrectly use causal = False, causing decoder tokens to see future information, which breaks the correctness of the model. 

> e.g. in the cross-attention layer of an encoder-decoder model like T5, layer.is_cross_attention=True will cause causal=False, which in turn allows decoder tokens within a block to see each other's "future," compromising generation correctness.

Can you @zhaochenyang20 help me double-check this?
Maybe we need to make sure it can only be applied for dLLMs in an elegant way. @ClawSeven 
In the future, new dllm algorithms may even a pass customized mask. I feel like it is better to decide whether do the causal by the algorithm, and even **[this PR can ignore, I just mention it to show the potential influence of our current interface]** an arg for customize_mask.

python/sglang/srt/server_args.py
```diff
@@ -1868,6 +1875,30 @@ def _handle_request_metrics_exporters(self):
...
+            logger.warning(
+                "Cuda graph is disabled because of using diffusion inference"
+            )
+            self.disable_cuda_graph = True
+        if not self.disable_overlap_schedule:
+            logger.warning(
+                "Overlap schedule is disabled because of using diffusion inference"
+            )
+            self.disable_overlap_schedule = True
+        if not self.disable_radix_cache:
```
Monstertail, line 1997: Will we support radix cache, PP in the future PR? @ClawSeven

python/sglang/srt/diffusion/algorithm/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+    return mapping
+
+def get_algorithm(
+    name: str,
+    block_size: int,
+):
+    try:
+        return algo_name_to_cls[name](block_size)
+    except:
+        raise RuntimeError(f"Unkown diffusion algorithm: {name}")
```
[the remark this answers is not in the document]
Monstertail: LGTM. The interface design is super clean and easy to understand:)

python/sglang/srt/server_args.py
```diff
@@ -1986,6 +2017,20 @@ def add_cli_args(parser: argparse.ArgumentParser):
...
+            type=str,
+            default=ServerArgs.diffusion_algorithm,
+            help="The diffusion algorithm."
+        )
+        parser.add_argument(
+            "--diffusion-block-size",
+            type=int,
+            default=ServerArgs.diffusion_block_size,
+            help="The number of tokens processed in each iteration of the block diffusion model."
+        )
```
Monstertail, line 2139: LGTM. No large issues but I would like to discuss the future plan.

In the long term, one thing we may think about is the relation with the diffusion algorithm here and the sglang-diffusion. 

1.  Will we think about moving/merging to sglang-diffusion?  I am not familiar with sglang-diffusion. @mickqian @zhaochenyang20  Any comments?  

2. If no, shall we still use the arg name in serveargs like diffusion_algorithm here, which may cause confusion?  Maybe 'dllm_algorithm' or any other names will be better?

python/sglang/srt/layers/attention/flashinfer_backend.py
ClawSeven: In the future, we may need to expose an attention mask interface to support other DLLM decoding algorithms.

Monstertail: Agree. I am glad to help with that:)

python/sglang/srt/server_args.py
ClawSeven: Implementing a block-wise radix cache is non-trivial, and I believe the potential performance gain is marginal. So the implention will be in the future. As for  pipeline parallelism (PP), the current model has $100$B parameters, which is unnecessary for enabling pipeline parallelism (PP) compared to tensor  parallelism.

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -749,8 +756,23 @@ def finished(self) -> bool:
         # Whether request reached finished condition
         return self.finished_reason is not None
 
+    def is_diffusion(self):
```
mickqian: nit: make it more precise, `is_dllm`

python/sglang/srt/layers/attention/flashinfer_backend.py
mickqian: agree, better make these dllm-args configurable, to avoid ambiguity

python/sglang/srt/server_args.py
mickqian: 1. sglang-diffusion has shared nothing with sglang-ar now, so no rush is needed here
2. agreed, please distinguish between `dllm` and `diffusion`

hutm:
@ClawSeven , which HF model are you using to validate correctness of the implementation? I am getting garbage for `inclusionAI/LLaDA2.0-mini-preview`:
```
git clone https://github.com/ClawSeven/sglang sglang-diffusion && cd sglang-diffusion && git checkout dev-dllm
uv venv --python 3.12 --seed && source .venv/bin/activate
pip install -e "python[all]"
python -m sglang.launch_server --model-path inclusionAI/LLaDA2.0-mini-preview --trust-remote-code

# test inference
curl http://localhost:30000/v1/chat/completions -H "Content-Type: application/json" -d '{"model": "inclusionAI/LLaDA2.0-mini-preview", "messages": [{"role": "system", "content": "You are a helpful assistant"},{"role": "user", "content": "how are you?"} ], "max_tokens": 100}'
{"id":"52baeb0a770e4a03abb528b0ea2189a3","object":"chat.completion","created":1764047480,"model":"inclusionAI/LLaDA2.0-mini-preview","choices":[{"index":0,"message":{"role":"assistant","content":"\nI'm'm 🖐� Did I    \n\n(I'm\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n","reasoning_content":null,"tool_calls":null},"logprobs":null,"finish_reason":"length","matched_stop":null}],"usage":{"prompt_tokens":28,"total_tokens":128,"completion_tokens":100,"prompt_tokens_details":null,"reasoning_tokens":0},"metadata":{"weight_version":"default"}}
```

python/sglang/srt/server_args.py
ClawSeven: All the diffusion naming has been replace with dllm or diffusion LLM.

mickqian (approved):

tianhaoz95:
@ClawSeven qq on the implementation, where is the block wise attention implemented? looks like flashinfer backend simply uses bidirectional attention?

ClawSeven:
Hi @tianhaoz95,
Currently, prefill and decoding batching are not separated, so we do not support customized attention masks. We implement block-wise attention by dividing the sequence into multiple blocks and applying bidirectional attention within each block, which is equivalent to block-wise attention.

Once the dLLM batching PR is merged, we will submit a pull request implementing the real block-wise attention with the new architecture design.

PR #12597, opened 2025-11-04
fix: respect `--ignore-eos` in PD case for benchmarking

ishandhanani:
https://github.com/sgl-project/sglang/pull/12334 does not handle `--ignore-eos` which is commonly used for fixed size isl/osl benchmarking. This PR provides an initial fix

### DIFF
--- a/python/sglang/srt/disaggregation/decode.py
+++ b/python/sglang/srt/disaggregation/decode.py
@@ -776,7 +776,10 @@ class DecodeTransferQueue:
                 # special handling for corner cases
                 should_finish = (
                     decode_req.req.sampling_params.max_new_tokens == 1
-                    or decode_req.req.output_ids[-1] in decode_req.req.eos_token_ids
+                    or (
+                        not decode_req.req.sampling_params.ignore_eos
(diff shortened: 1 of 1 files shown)


python/sglang/srt/disaggregation/decode.py
```diff
@@ -776,7 +776,10 @@ def pop_transferred(self) -> List[Req]:
                 # special handling for corner cases
                 should_finish = (
                     decode_req.req.sampling_params.max_new_tokens == 1
-                    or decode_req.req.output_ids[-1] in decode_req.req.eos_token_ids
+                    or (
```
hnyls2002: LGTM. I think this logic can be improved by more reliable finish checking... I may submit a follow-up PR to improve it.

hnyls2002 (approved):

python/sglang/srt/disaggregation/decode.py
ishandhanani: Please do. Can you ping me when you have that PR ready? I want to help test to ensure our benchmarks dont break

ShangmingCai (approved):
LGTM.

PR #12206, opened 2025-10-27
Fix potential eos bug on decode instance when PD is enabled

ShangmingCai:
## Motivation
Trying to fix https://github.com/sgl-project/sglang/issues/12151

## Checklist

- [ ] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.ai/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.ai/developer_guide/contribution_guide.html#benchmark-the-speed).

Issue #12151, opened 2025-10-26
[PD BUG] when prefill first_token  output eos, decode node shoudn't output anymore


whybeyoung:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [ ] 2. The bug has not been fixed in the latest version.
- [ ] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [ ] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Describe the bug

In certain scenarios, the model outputs an EOS (End of Sentence) token as the first character. In such cases, under the PD-separated scenario, the decoder should not continue generating additional text.
Current Situation: The decode node will generate disordered and meaningless text.

CC @ShangmingCai @ByronHsu @hnyls2002 

### Reproduction

wait

### Environment

H20

PR #12638, opened 2025-11-04
[PD-Disagg] Check finish after pop tranferred

hnyls2002:
follow up for #12597 #12334 #12206

And this PR should be a robust solution for #12151

### DIFF
--- a/python/sglang/srt/disaggregation/decode.py
+++ b/python/sglang/srt/disaggregation/decode.py
@@ -773,17 +773,12 @@ class DecodeTransferQueue:
                 indices_to_remove.add(i)
                 decode_req.req.time_stats.wait_queue_entry_time = time.perf_counter()
 
-                # special handling for corner cases
-                should_finish = decode_req.req.sampling_params.max_new_tokens == 1 or (
-                    not decode_req.req.sampling_params.ignore_eos
-                    and decode_req.req.output_ids[-1] in decode_req.req.eos_token_ids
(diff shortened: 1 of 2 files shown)


python/sglang/srt/disaggregation/decode.py
```diff
@@ -773,17 +773,12 @@ def pop_transferred(self) -> List[Req]:
...
                 decode_req.req.time_stats.wait_queue_entry_time = time.perf_counter()
 
-                # special handling for corner cases
-                should_finish = decode_req.req.sampling_params.max_new_tokens == 1 or (
-                    not decode_req.req.sampling_params.ignore_eos
-                    and decode_req.req.output_ids[-1] in decode_req.req.eos_token_ids
-                )
-                if should_finish:
+                decode_req.req.check_finished()
+                if decode_req.req.finished():
```
ShangmingCai, line 777: Should we run full `check_finished()` for every req? I was afraid that it might incur non-negligible overhead, so I only checked for special cases. If that won't be the case, this modification will be more thorough.

hnyls2002: 1. `check_finished`'s complexity is $O(1)$ and it only introduces a little overhead.
2. It is inside the `pop_transferred` function, which also contains other scheduling operations and will be overlapped with overlap scheduling.
3. Only checking every iteration is correct and consistent.

PR #12646, opened 2025-11-04
Add timing metrics for requests

cicirori:
see https://github.com/sgl-project/sglang/pull/11119

I’ve tried out https://github.com/sgl-project/sglang/issues/10916 and it’s a fantastic feature — low-overhead and provides a clear, comprehensive visualization tool.

However, since it depends on OpenTelemetry, in many cases we only need basic stdout metrics rather than a full tracing setup. I also experimented with adapting the existing code to print metrics directly to stdout, but if I understand correctly, the current `--enable-trace` implementation assumes that all trace slices are centrally handled by an external OpenTelemetry endpoint. This makes it quite difficult to reuse the current tracing logic for lightweight local metrics.

Given that @scottjlee  already contributed a related PR earlier, I’d suggest that we reconsider merging this one — or at least explore an alternative path to expose basic metrics without requiring the full OpenTelemetry stack.


python/sglang/srt/managers/multi_tokenizer_mixin.py
```diff
@@ -208,6 +224,7 @@ def _handle_output_by_index(output, i):
             placeholder_tokens_idx=None,
             placeholder_tokens_val=None,
             token_steps=([output.token_steps[i]] if output.token_steps else None),
+            rids=[output.rids[i]],
```
ShangmingCai: Why move this to the very end?

python/sglang/srt/managers/detokenizer_manager.py
```diff
@@ -250,6 +250,10 @@ def handle_batch_token_id_out(self, recv_obj: BatchTokenIDOutput):
 
         return BatchStrOutput(
             rids=recv_obj.rids,
+            queue_time=recv_obj.queue_time,
+            inference_start_time=recv_obj.inference_start_time,
+            prefill_delay=recv_obj.prefill_delay,
+            prefill_latency=recv_obj.prefill_latency,
```
ShangmingCai: Maybe attach them in the back. It would be better if we could align the parameter order for further maintenance.

python/sglang/srt/managers/multi_tokenizer_mixin.py
```diff
@@ -340,6 +368,7 @@ def _handle_output_by_index(output, i):
                 else None
             ),
             token_steps=([output.token_steps[i]] if output.token_steps else None),
+            rids=[output.rids[i]],
```
ShangmingCai: ditto

python/sglang/srt/managers/io_struct.py
```diff
@@ -61,6 +61,44 @@ def regenerate_rids(self):
...
+
+    This class consolidates the timing metrics that are shared across all batch output types
+    to avoid code duplication and ensure consistency.
+    """
+
+    # Queue duration: time spent waiting in queue before request is scheduled.
+    queue_time: List[Optional[float]]
+
+    # Inference start time: timestamp when inference computation begins (start of prefill).
+    inference_start_time: List[Optional[float]]
```
ShangmingCai: The meaning of `inference` should be clarified. Does it mean prefill + decode for PD-colocate instance, and what does it mean for P instance and D instance respectively?

python/sglang/srt/managers/io_struct.py
```diff
@@ -61,6 +61,44 @@ def regenerate_rids(self):
...
+
+    # Prefill delay: time spent waiting before prefill starts
+    prefill_delay: List[Optional[float]]
+
+    # Prefill latency: time spent during prefill computation
+    prefill_latency: List[Optional[float]]
```
ShangmingCai: Should these be Optional[List[xxx]]? I am not sure, it seems the current definition would accept [None, float, float], which is weird.

ShangmingCai:
Timing metrics are useful, but this PR should be reviewed by @hnyls2002 to minimize the overhead and modifications to prevent future refactoring debt.

python/sglang/srt/managers/multi_tokenizer_mixin.py
cicirori: done

python/sglang/srt/managers/io_struct.py
```diff
@@ -61,6 +61,55 @@ def regenerate_rids(self):
...
+
+    # Prefill delay: time spent waiting between forward entry and prefill start.
+    # Calculated as: prefill_start_time - forward_entry_time
+    # This represents the delay between when the request enters the forward stage
+    # and when prefill computation actually begins.
+    prefill_delay: Optional[List[Optional[float]]]
+
+    # Prefill latency: time spent during prefill computation.
+    # Calculated as: prefill_end_time - prefill_start_time
+    prefill_latency: Optional[List[Optional[float]]]
```
merrymercy, line 94: This is wrong. It is only the launch time instead of the real computation time.

python/sglang/srt/managers/scheduler.py
```diff
@@ -2051,11 +2058,18 @@ def run_batch(
...
         else:  # embedding or reward model
             model_worker_batch = batch.get_model_worker_batch()
             embeddings = self.tp_worker.forward_batch_embedding(model_worker_batch)
             ret = EmbeddingBatchResult(embeddings=embeddings)
+
+        # Capture prefill end time for EXTEND mode
+        if batch.forward_mode == ForwardMode.EXTEND:
+            current_time = time.perf_counter()
+            for req in batch.reqs:
+                req.time_stats.prefill_end_time = current_time
```
merrymercy, line 2071: This is only the kenrel launch time

python/sglang/srt/managers/scheduler_output_processor_mixin.py
```diff
@@ -861,6 +867,28 @@ def stream_output_generation(
 
                 retraction_counts.append(req.retraction_count)
 
+                queue_times.append(req.time_stats.get_queueing_time())
+                forward_entry_times.append(req.time_stats.forward_entry_time)
+
+                if req.time_stats.prefill_start_time > 0.0:
+                    prefill_delays.append(
```
merrymercy, line 874: This is redundant code. It appeared twice in this file. You should move all common code to `class TimeStats`

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -1422,6 +1439,27 @@ def _handle_batch_output(
                 "total_retractions": recv_obj.retraction_counts[i],
             }
 
+            if (
```
merrymercy, line 1442: Is it possible that we only send this info back when certain flags are turned on? 
For example, only send this back when `--enable-metrics`

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -1691,6 +1733,57 @@ def _calculate_spec_decoding_metrics(
...
+            BatchTokenIDOutput,
+        ],
+        i: int,
+    ) -> None:
+        """Calculate request-level timing metrics, such as inference time, decode throughput, and time per token."""
+        # Request timing timestamps.
+        if state.created_time > 0:
+            meta_info["request_received_ts"] = state.created_time
+        if state.request_scheduled_ts > 0:
+            meta_info["request_scheduled_ts"] = state.request_scheduled_ts
```
merrymercy, line 1753: `request_scheduled_ts` -> `request_sent_to_scheduler_ts`

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -1691,6 +1733,57 @@ def _calculate_spec_decoding_metrics(
...
+        if (
+            state.first_token_time_perf > 0.0
+            and state.finished_time_perf > 0.0
+            and not isinstance(recv_obj, BatchEmbeddingOutput)
+            and recv_obj.completion_tokens[i] > 0
+        ):
+            decode_time = state.finished_time_perf - state.first_token_time_perf
+            completion_tokens = recv_obj.completion_tokens[i]
+            meta_info["decode_throughput"] = completion_tokens / decode_time
+            meta_info["time_per_token"] = decode_time / completion_tokens
```
merrymercy, line 1785: do not need to return two redundant fields. Users can do the conversation by themselves in the client code

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -1487,8 +1525,12 @@ def _handle_batch_output(
                 if self.server_args.speculative_algorithm:
                     self._calculate_spec_decoding_metrics(meta_info, recv_obj, i)
                 state.finished_time = time.time()
+                state.finished_time_perf = time.perf_counter()
                 meta_info["e2e_latency"] = state.finished_time - state.created_time
 
+                # Calculate timing metrics
+                self._calculate_timing_metrics(meta_info, state, recv_obj, i)
```
merrymercy, line 1532: only run this if `--enable-metrics`?

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -1691,6 +1733,57 @@ def _calculate_spec_decoding_metrics(
...
+
+        # Inference time calculation.
+        if (
+            hasattr(recv_obj, "forward_entry_time")
+            and recv_obj.forward_entry_time
+            and recv_obj.forward_entry_time[i] is not None
+            and state.finished_time_perf > 0.0
+        ):
+            forward_time = state.finished_time_perf - recv_obj.forward_entry_time[i]
+            meta_info["forward_time"] = forward_time
```
merrymercy, line 1773: let us call it `inference_time` or something to avoid the confusion between this and the `forward_time` in `TimeStats`?

PR #12654, opened 2025-11-05
[Perf] Modify threshold for triggering allreduce norm fusion

Fridge003:
## Motivation
Currently in  DeepSeek fp4 and pure TP scenario, `flashinfer::trtllm_allreduce_fusion` is applied to batches with `bs <= FUSE_ALLREDUCE_MAX_BATCH_SIZE`. 

However, with medium batch size like 2048(current threshold), we found that the cost for trtllm_allreduce_fusion might be much larger than all_reduce+norm with symmetric memory enabled. So we are lowering this threshold to a smaller value (like 128, as discussed in https://github.com/flashinfer-ai/flashinfer/issues/1223#issuecomment-3047256465)

## Benchmarking and Profiling

DeepSeek fp4+TP 8 on B200

```bash
python3 -m sglang.launch_server --model-path nvidia/DeepSeek-R1-0528-FP4-v2 --tp-size 8 --mem-fraction-static 0.90 --quantization modelopt_fp4 --kv-cache-dtype fp8_e4m3 --attention-backend trtllm_mla --moe-runner-backend flashinfer_trtllm  --trust-remote --enable-symm-mem --cuda-graph-max-bs 64
```

Profile (BS=1, ISL=2048, OSL=4):

Before this PR, the time of allreduce+norm fusion is 186us
截屏2025-11-04 16 31 59

After this PR, the time of allreduce kernel plus norm kernel with symmetric memory is 107us

截屏2025-11-04 16 34 07

## Checklist

### DIFF
--- a/python/sglang/srt/layers/communicator.py
+++ b/python/sglang/srt/layers/communicator.py
@@ -63,7 +63,7 @@ _is_gfx95_supported = is_gfx95_supported()
 if _use_aiter and _is_gfx95_supported:
     from sglang.srt.layers.quantization.rocm_mxfp4_utils import fused_rms_mxfp4_quant
 
-FUSE_ALLREDUCE_MAX_BATCH_SIZE = 2048
+FUSE_ALLREDUCE_MAX_BATCH_SIZE = 128
 
 
 class ScatterMode(Enum):
@@ -559,7 +559,7 @@ class CommunicateWithAllReduceAndLayerNormFn:
                 and _is_flashinfer_available
                 and hasattr(layernorm, "forward_with_allreduce_fusion")
                 and get_global_server_args().enable_flashinfer_allreduce_fusion
-                and hidden_states.shape[0] <= 4096
+                and hidden_states.shape[0] <= FUSE_ALLREDUCE_MAX_BATCH_SIZE
             ):
                 hidden_states, residual = layernorm.forward_with_allreduce_fusion(
                     hidden_states, residual


Fridge003:
@yuan-luo @Qiaolin-Yu  Can you have a look

zhyncs:
@averyhNV @yyihuang

yyihuang (approved):
LGTM.

python/sglang/srt/layers/communicator.py
```diff
@@ -559,7 +559,7 @@ def _gather_hidden_states_and_residual(
                 and _is_flashinfer_available
                 and hasattr(layernorm, "forward_with_allreduce_fusion")
                 and get_global_server_args().enable_flashinfer_allreduce_fusion
-                and hidden_states.shape[0] <= 4096
+                and hidden_states.shape[0] <= FUSE_ALLREDUCE_MAX_BATCH_SIZE
```
yuan-luo, line 562: FLASHINFER_ALLREDUCE_FUSION_MAX_BATCH_SIZE might be better.

yuan-luo (approved):

yuan-luo:
Have you tried on H20 or H800? The result might be slightly different.

Fridge003:
> Have you tried on H20 or H800? The result might be slightly different.

Not yet... I was doing benchmark on B200 and the result shows this change has negative effect.
Still underworking

yyihuang:
> Not yet... I was doing benchmark on B200 and the result shows this change has negative effect.
> Still underworking

What is the negative effect? Did we already get better perf by symm-mem as in the Benchmarking and Profiling section? 

BTW, did we pass the oneshot param when calling trtllm AR fusion? Here is the threshold for oneshot / twoshot strategy https://github.com/flashinfer-ai/flashinfer/blob/9721ff7ff11cd537ea5c3aba61aef0e037dddf74/include/flashinfer/comm/trtllm_allreduce_fusion.cuh#L1349 but i think we need to pass this explicitly at python interface for the current version FI.

Fridge003:
> > Not yet... I was doing benchmark on B200 and the result shows this change has negative effect.
> > Still underworking
> 
> What is the negative effect? Did we already get better perf by symm-mem as in the Benchmarking and Profiling section?
> 
> BTW, did we pass the oneshot param when calling trtllm AR fusion? Here is the threshold for oneshot / twoshot strategy https://github.com/flashinfer-ai/flashinfer/blob/9721ff7ff11cd537ea5c3aba61aef0e037dddf74/include/flashinfer/comm/trtllm_allreduce_fusion.cuh#L1349 but i think we need to pass this explicitly at python interface for the current version FI.

SGLang didn't pass this argument.  one-shot/two-shots is handled on flashinfer side

PR #12665, opened 2025-11-05
[VLM] Support Encode/Language Model Dissaggregation for Qwen

ZhengWG:
## Motivation

Introduce the Encode/Language disaggregation architecture so multimodal workloads can scale vision encoders and language decoders independently, unblock high-QPS deployments.

## Modifications

1. Adapted Qwen2_5_VLForConditionalGeneration model to support only-vision part. Adapted Qwen2/3 model, and related configs so the encode model can emit multimodal embeddings while the language plane loads only LM weights.
2. Implemented connectors based on mooncake for multimodal data flow between vision and language disaggregated nodes, which is mainly based on PD disaggregation.
3. Implemented event_loop for encode and language schedule process.

Inference flow is as followed:

<p align="center">
image
</p>

For detailed design, please refer to: [vlm_disaggregation.md](https://github.com/antgroup/sglang/blob/d9aab3c77a477ce566e975c38f5fdd1c4ff0f80f/docs/advanced_features/vlm_disaggregation.md)

## Benchmarking and Profiling

**Launch the Bootstrap Server**

```bash
# Launch the bootstrap server on a control node
python3 -m sglang_router.launch_router --mini-lb --host $HOST_IP \
    --port $SERVER_PORT --vision http://${EMBEDDING_IP}:${EMBEDDING_PORT} 8998 \
    --prefill http://${LANGUAGE_IP_LIST[0]}:${LANGUAGE_PORT} 8998 \
    --enable-multimodal-disagg
```

**Launch the Encode Service**

```bash
export PORT=8001
export SGLANG_VLM_CACHE_SIZE_MB=40960
export TENSOR_PARALLEL_SIZE=2
export CHUNKED_PREFILL_SIZE=81920
export MAX_RUNNING_REQUESTS=128
export MEM_FRACTION_STATIC=0.85
export SGLANG_EMBEDDING_CACHE_BUFFER_SIZE=128
export SGLANG_EMBEDDING_CACHE_BLOCK_SIZE=16384
# Encode: vision encoding
python3 -m sglang.launch_server --model-path ${MODEL_PATH} --enable-torch-compile --max-prefill-tokens $CHUNKED_PREFILL_SIZE \
        --host $HOST_IP --port $PORT --trust-remote-code --tp-size ${TENSOR_PARALLEL_SIZE} --mem-fraction-static ${MEM_FRACTION_STATIC} \
        --enable-cache-report --log-level info --max-running-requests ${MAX_RUNNING_REQUESTS} \
        --chunked-prefill-size ${CHUNKED_PREFILL_SIZE} --attention-backend fa3 --json-model-override-args '{"is_multimodal_embedding": true}' \
        --mm-attention-backend fa3 --disaggregation-mode encode
```

**Launch the Language Service**

```bash
export PORT=8002
export TENSOR_PARALLEL_SIZE=8
export MAX_RUNNING_REQUESTS=128
export SGLANG_EMBEDDING_CACHE_BUFFER_SIZE=128
export SGLANG_EMBEDDING_CACHE_BLOCK_SIZE=16384
export MEM_FRACTION_STATIC=0.85
export CHUNKED_PREFILL_SIZE=8192
# Configure the default buffer allocation
export SGLANG_EMBEDDING_DEFAULT_ALLOCATE_BUFFER_SIZE=16384

# Language: text generation
# Qwen2.5-VL: "architectures": ["Qwen2ForCausalLM"]
python3 -m sglang.launch_server --model-path ${MODEL_PATH} --enable-torch-compile --disable-radix-cache \
        --host $HOST_IP --port $PORT --trust-remote-code --tp-size ${TENSOR_PARALLEL_SIZE} --served-model-name "qwen3-vl" \
        --enable-cache-report --log-level info --max-running-requests ${MAX_RUNNING_REQUESTS} --json-model-override-args '{"architectures": ["Qwen3MoeForCausalLM"]}' \
        --mem-fraction-static ${MEM_FRACTION_STATIC} --chunked-prefill-size ${CHUNKED_PREFILL_SIZE} --attention-backend fa3 \
        --disaggregation-mode language
```

**Benchmark Requests**

```
                    --host ${HOST_IP} \
                    --port ${PORT} \
                    --model $MODEL_PATH \
                    --backend sglang-oai-chat \
                    --dataset-name "image" \
                    --random-input-len $input_len \
                    --random-output-len $output_len \
                    --random-range-ratio 1 \
                    --num-prompt $num_prompt \
                    --warmup-requests 0 \
                    --flush-cache \
                    --image-count 1 \
                    --image-resolution $image_size \
                    --image-format "jpeg" \
                    --image-content "random" \
                    --request-rate $qps \
                    --output-file $result_file \
                    --max-concurrency 128
```

**Performance Metrics**

SLA: mean TTFT < 4 s; mean TPOT < 100 ms.

**Single Node (TP=8)**

| Model               | Scenario                                        | qps/gpu           | TTFT (Mean) | TPOT (Mean) |
|---------------------|-------------------------------------------------|-------------------|-------------|-------------|
| qwen2.5-vl-72B      | Single image + text (2000×2000 + 1k), output 300, qps: 0.52 | 0.06625 req/s/gpu | 3826.07 ms  | 78.63 ms    |
| qwen3-vl-235B-A22B  | Single image + text (2000×2000 + 1k), output 300, qps: 0.85 | 0.1075 req/s/gpu  | 3738.24 ms  | 91.24 ms    |

**Disaggregated Deployment**

| Configuration | Encode Plane       | Language Plane     |
|---------------|--------------------|--------------------|
| GPU model     | NVIDIA H20 96GB    | NVIDIA H20 96GB    |
| GPU count     | 2 (TP = 2)         | 8 (TP = 8)         |

| Model               | Scenario                                        | qps/gpu                  | TTFT (Mean)  | TPOT (Mean) |
|---------------------|-------------------------------------------------|--------------------------|--------------|-------------|
| qwen2.5-vl-72B      | Single image + text (2000×2000 + 1k), output 300, qps: 0.78 | 0.07420 (+12%) req/s/gpu | 3632.70 ms   | 95.61 ms    |
| qwen3-vl-235B-A22B  | Single image + text (2000×2000 + 1k), output 300, qps: 1.40 | 0.141 (+31.2%) req/s/gpu | 3831.58 ms   | 95.34 ms    |

## Checklist


ZhengWG:
@yuan-luo Can you help review it?

yuan-luo:
Refer to the architecture of EPD:
image

EPDServe implements the **Encode-Prefill-Decode (EPD) Disaggregation** architecture proposed in the EPD paper. It is designed to serve large multimodal models (LMMs) efficiently by **splitting the inference pipeline** into three independent stages:
🎨 Encoding Stage: Processes multimodal inputs (images, audio, video)
🧠 Context (Prefill) Stage: Handles prompt token prefill
🔁 Decoding Stage: Performs autoregressive token generation

https://github.com/vbdi/epdserve

ShangmingCai (commented):
It is great that we can borrow the arch design of PD Disaggregation, and this is a well-developed PR with clear documentation. The `MultimodalDataBuffers` abstract is great, but I think for the transfer part we need a separate _conn_mm.py_. The _conn.py_ has already integrated with many complicated logics, mixing mm logics in this file will pose a potential risk for code maintenance and readability (in fact, the encoder side doesn't need that much). As for the bootstrapping, queueing, and scheduling parts, I think they are a little bit heavy. Since mm data buffer is in the CPU RAM, I think we don't need to do it like kvcache (need to finish bootstrapping first because we have chunk prefill + overlap transfer), and should be able to do the computation and set_buf at the encoder side during bootstrapping (or even before bootstrapping). And in the future, we can integrate it with the Mooncake Store, and make the language side retrieve the embedding results from the store as well.

ShangmingCai:
cc: @mickqian What do you think?

python/sglang/srt/disaggregation/mini_lb.py
```diff
@@ -1,6 +1,788 @@
-raise RuntimeError(
-    """The 'mini_lb' module has been relocated to the 'sglang_router' package.
-    We recommend installing 'sglang-router' with Rust support for optimal performance.
-    If you encounter issues building the router with Rust, set the environment variable
-    'SGLANG_ROUTER_BUILD_NO_RUST=1' and add '--mini-lb' to the command line to use the Python version of 'mini_lb'."""
-)
```
ShangmingCai: Also, this should be reverted, mini lb has been moved to the sgl-router.

ShangmingCai:
I tested Qwen/Qwen2.5-VL-7B-Instruct, but the error pop out after few reqs.
```bash
  File "/root/wenming/sglang/python/sglang/srt/model_executor/model_runner.py", line 2070, in forward_extend
    return self.model.forward(
  File "/root/miniconda3/envs/sglang/lib/python3.10/site-packages/torch/utils/_contextlib.py", line 120, in decorate_context
    return func(*args, **kwargs)
TypeError: Qwen2_5_VLForConditionalGeneration.forward() got an unexpected keyword argument 'input_embeds'
```

python/sglang/srt/disaggregation/mini_lb.py
ZhengWG: Yes, I will move it as soon as I can.

ZhengWG:
> It is great that we can borrow the arch design of PD Disaggregation, and this is a well-developed PR with clear documentation. The `MultimodalDataBuffers` abstract is great, but I think for the transfer part we need a separate _conn_mm.py_. The _conn.py_ has already integrated with many complicated logics, mixing mm logics in this file will pose a potential risk for code maintenance and readability (in fact, the encoder side doesn't need that much). As for the bootstrapping, queueing, and scheduling parts, I think they are a little bit heavy. Since mm data buffer is in the CPU RAM, I think we don't need to do it like kvcache (need to finish bootstrapping first because we have chunk prefill + overlap transfer), and should be able to do the computation and set_buf at the encoder side during bootstrapping (or even before bootstrapping). And in the future, we can integrate it with the Mooncake Store, and make the language side retrieve the embedding results from the store as well.

Thanks for the suggestions! Agree to separate conn_mm.py, will refactor it later.
For bootstrapping/queueing/scheduling, the main issue is Language Instance skips multimodal data processing to reduce time, causing inability to pre-allocate embedding-cache.So bootstrapping is still necessary and better to overlap with forward process, but the whole process can be simplified.  Here is my proposed solution:
   1. After multimodal-preprocessing, Encoder notifies Language to allocate embedding-cache and then get remote-cache info (async in conn_mm.py), for Mooncake Store, pass corresponding key if exists
   2. After forward, Encoder transfers/puts data using obtained addr/key and notifies Language

Would this approach work for you?

ShangmingCai:
> > It is great that we can borrow the arch design of PD Disaggregation, and this is a well-developed PR with clear documentation. The `MultimodalDataBuffers` abstract is great, but I think for the transfer part we need a separate _conn_mm.py_. The _conn.py_ has already integrated with many complicated logics, mixing mm logics in this file will pose a potential risk for code maintenance and readability (in fact, the encoder side doesn't need that much). As for the bootstrapping, queueing, and scheduling parts, I think they are a little bit heavy. Since mm data buffer is in the CPU RAM, I think we don't need to do it like kvcache (need to finish bootstrapping first because we have chunk prefill + overlap transfer), and should be able to do the computation and set_buf at the encoder side during bootstrapping (or even before bootstrapping). And in the future, we can integrate it with the Mooncake Store, and make the language side retrieve the embedding results from the store as well.
> 
> Thanks for the suggestions! Agree to separate conn_mm.py, will refactor it later. For bootstrapping/queueing/scheduling, the main issue is Language Instance skips multimodal data processing to reduce time, causing inability to pre-allocate embedding-cache.So bootstrapping is still necessary and better to overlap with forward process, but the whole process can be simplified. Here is my proposed solution:
> 
> 1. After multimodal-preprocessing, Encoder notifies Language to allocate embedding-cache and then get remote-cache info (async in conn_mm.py), for Mooncake Store, pass corresponding key if exists
> 2. After forward, Encoder transfers/puts data using obtained addr/key and notifies Language
> 
> Would this approach work for you?

@ZhengWG Sounds good. Another thing, IIUC, the current design is more optimized for throughput, but it might not be helpful for TTFT.

I am thinking maybe we can refactor the bootstrapping part to make encoder DP achievable by assigning/pairing multiple encoders to 1 prefill node when min-lb/router detects the user request has many mm contents (also requires a lot of changes to the bootstrapping part for prefill), and finish when all encoder sync status back, so that we can reduce TTFT as well.

PR #12666, opened 2025-11-05
[sgl-kernel][5/N]Support Expert Specialization Grouped GEMM

HydraQYH:
## Motivation
Add the flag parameter to CUTLASS MoE to enable the expert specialization kernel.

## Modifications

- python/sglang/srt/layers/moe/cutlass_moe.py
- python/sglang/test/test_cutlass_moe.py

## Accuracy Tests
image

[check.log](https://github.com/user-attachments/files/23350293/check.log)

## Benchmarking and Profiling
Before using es:
image
After using es:
image

[upstream_cutlass.log](https://github.com/user-attachments/files/23349281/upstream_cutlass.log)
[upstream_es.log](https://github.com/user-attachments/files/23349287/upstream_es.log)

## Checklist


python/sglang/srt/layers/moe/cutlass_moe.py
```diff
@@ -121,7 +124,7 @@ def cutlass_fused_experts_fp8(
         from sglang.srt.layers.quantization.fp8_kernel import (
             sglang_per_token_group_quant_fp8,
         )
-
+    es_up, es_down = enable_es
```
Fridge003, line 127: How can user pass enable_es to cutlass_moe?
Do we need an environment variable?

HydraQYH: There are still two optimizations that haven't been implemented:

1. Based on arithmetic intensity combined with MNK, a suitable kernel is dynamically selected to adapt to more scenarios.
2. PDL support: https://github.com/NVIDIA/cutlass/pull/2719

Once these tasks are complete, I will add an environment variable to enable ES.

Fridge003 (approved):

Issue #12610, opened 2025-11-04
[Bug] Qwen3-Omni Model: speech recognition processes only the first 30 seconds of audio.


billrenhero:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Describe the bug

When deploying the Qwen3-Omni model with SGLang v0.5.4.post1-cu129, speech recognition only processes the first 30 seconds of audio. This behavior matches the issue described in https://github.com/QwenLM/Qwen3-Omni/discussions/36.

### Reproduction

python3 -m sglang.launch_server \
  --model-path Qwen/Qwen3-Omni-30B-A3B-Instruct \  
  --host 0.0.0.0 \
  --port 80 \
  --tp 4

### Environment

H20 * 4
CUDA 12.8

jiapingW:
Thanks, I will fix it later.

PR #12674, opened 2025-11-05
fix qwen3-omni audio length < 30s

jiapingW:
## Motivation

The qwen3 Omni series models, due to their use of the WhisperFeatureExtractor class, limit the chunk size to 30 seconds, causing truncation regardless of the input length. It would be desirable to support Omni models in processing complete audio. Fix the bug [here](https://github.com/sgl-project/sglang/issues/12610).
## Checklist

- [✅] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).


mickqian (approved):
LGTM for this one, but eventually we should adopt user-overriding processor kwargs

jiapingW:
Yes. I think the relevant parameters, such as those shown below, should be encapsulated in sglang by inheriting from transformers; otherwise, it's difficult to synchronize small updates to transformers to sglang.
```python
        "text_kwargs": {
            "padding": False,
            "padding_side": "left",
        },
        "videos_kwargs": {
            "seconds_per_chunk": 2.0,
            "position_id_per_seconds": 13.0,
            "use_audio_in_video": False,
            "size": {
                "shortest_edge": 128 * 32 * 32,
                "longest_edge": 768 * 32 * 32,
            },
        },
        "audio_kwargs": {
            "sampling_rate": 16000,
            "padding": True,
            "truncation": False,
            "return_attention_mask": True,
        },
```
> LGTM for this one, but eventually we should adopt user-overriding processor kwargs

Issue #4935, opened 2025-03-30
[Bug] 0.0.0.0 host not supported


gyin94:
### Checklist

- [ ] 1. I have searched related issues but cannot get the expected help.
- [ ] 2. The bug has not been fixed in the latest version.
- [ ] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [ ] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [ ] 5. Please use English, otherwise it will be closed.

### Describe the bug

if we set the host as 0.0.0.0, the warmup function will fail.

<b>Connection to 0.0.0.0 failed.</b></p>\n</blockquote>\n\n<p id="sysmsg">The system returned: <i>(111) Connection refused</I>

_wait_and_warmup -> res = requests.get(url + "/get_model_info", timeout=5, headers=headers)

can we add a fix here to replace 0.0.0.0 with 127.0.0.1 for warmup check? while we still let the server to be deployed in 0.0.0.0

### Reproduction

`--host 0.0.0.0` 

### Environment

```
Python: 3.10.12 (main, Feb  4 2025, 14:57:36) [GCC 11.4.0]
CUDA available: True
GPU 0,1,2,3,4,5,6,7: NVIDIA H100 80GB HBM3
GPU 0,1,2,3,4,5,6,7 Compute Capability: 9.0
CUDA_HOME: /usr/local/cuda-12
NVCC: Cuda compilation tools, release 12.6, V12.6.68
CUDA Driver Version: 535.129.03
PyTorch: 2.5.1+cu124
sglang: 0.4.4.post3
sgl_kernel: 0.0.5.post4
flashinfer: Module Not Found
triton: 3.1.0
transformers: 4.50.0
torchao: 0.9.0
numpy: 1.26.4
aiohttp: 3.11.14
fastapi: 0.115.11
hf_transfer: 0.1.9
huggingface_hub: 0.29.3
interegular: 0.3.3
modelscope: 1.24.1
orjson: 3.9.10
outlines: 0.1.11
packaging: 24.2
psutil: 5.9.8
pydantic: 2.10.6
multipart: Module Not Found
zmq: Module Not Found
uvicorn: 0.34.0
uvloop: 0.21.0
vllm: 0.8.1
[...]

PR #12679, opened 2025-11-05
fix: add URL sanitizing to fix 0.0.0.0 host warmup issue

ai-easy-cpu:
## Motivation

When the server is bound to `0.0.0.0`, the warmup server posts a request to `0.0.0.0` raises an error.

This is mentioned in #4935 early, but very few reported overall. There are many `0.0.0.0` host in the doc, although the default value is `127.0.0.1`. Thus it should be easily encountered after exhausting the warmup trials (120 times now, unless you skip the warmup), not sure why it seems most users have not met this yet.

## Modifications

Add a sanitizing, to replace `0.0.0.0` to `127.0.0.1` (loopback to itself)

## Accuracy Tests

Not affected

## Benchmarking and Profiling

Not affected

## Checklist


hnyls2002:
Thanks @ai-easy-cpu. This was addressed on main by #18973: ServerArgs.url() now rewrites 0.0.0.0 to 127.0.0.1 (and :: to ::1) for internal requests, so the warmup
  path no longer posts to 0.0.0.0. Closing as superseded; please reopen if I've missed a case this does not cover.

PR #12685, opened 2025-11-05
[Feat] remove SGL_USE_CUDA_IPC env_var

kousakawang:
## Motivation

Since we already have --keep-mm-feature-on-device when launch server, we can use this option to control cuda-ipc-transport's on/off. 

## Modifications

replacing all SGL_USE_CUDA_IPC control  with using server_args.keep_mm_feature_on_device.

**Todo：**
enable cuda-ipc transport  not only for mm_featute on device but also in more general cases(normal tensor, tensor on cpu).

## Checklist


adarshxs:
since we plan to use cuda ipc for general purpose tensor transport, maybe we can keep these two flags decoupled?

kousakawang:
> since we plan to use cuda ipc for general purpose tensor transport, maybe we can keep these two flags decoupled?
Ok,  only thing I want to confirm is in which case will --keep-mm-feature-on-device be effective? I check the code and think currently on/off --keep-mm-feature-on-device server args  make no difference.

adarshxs:
> > since we plan to use cuda ipc for general purpose tensor transport, maybe we can keep these two flags decoupled?

>  Ok,  only thing I want to confirm is in which case will --keep-mm-feature-on-device be effective? I check the code and think currently on/off --keep-mm-feature-on-device server args  make no difference.

cc @mickqian was this disabled due to the mem leak issue?

PR #12692, opened 2025-11-05
[NPU][Feat] Add Ascend sampling backend

Alexhaoge:
Co-author: @Alexhaoge @shuy98 
## Motivation
Currently, Ascend NPU only supports `pytorch` backend. The sampling performance degrades sharply under large batchsize(>100). We propose to add `ascend` sampling backend for better performance on NPU.

## Modifications
The main improvement gains from `torch_npu.npu_top_k_top_p` and `torch.Tensor.masked_fill_`.
- If `top_ks<=1024`, Use fused op [torch_npu.npu_top_k_top_p](https://www.hiascend.com/document/detail/zh/Pytorch/710/apiref/torchnpuCustomsapi/context/torch_npu-npu_top_k_top_p.md) for top-k & top-p sampling. The return value of this op is the filtered logits.
  - Because softmax of the raw logits is included in the fused op, the initial softmax can be skipped if `return_logprob` is false.
  - `torch_npu.npu_top_k_top_p` apply top-p filtering on top of top-k reults. This is a litte bit different from pytorch backend, but does not affect the accuracy since the final answer is the intersection of top-k and top-p filtering.
- If `top_ks>1024`, fallback to native torch interface, but use `masked_fill_` instead of naive tensor indexing. `[:]` array style indexing is the root cause of sampling performance degradation on NPU with large batchsize.

Note: Currently ascend sampling backend does not support deterministic sampling but it is feasible to implement follow the similar way as pytorch backend. We will create another pr once NPU support works for deterministic inference is finished wholistically.

## Accuracy Tests
Included in the unit test file `test/srt/ascend/test_ascend_pytorch_sampling_backend.py`. We use the same test cases as `test/srt/test_pytorch_sampling_backend.py`.
acc

## Benchmarking and Profiling
On defaults, `sglang.bench_serving` set temperature 0. To benchmark sampling, we modify the script by adding [Qwen3 recommended sampling parameters](https://huggingface.co/Qwen/Qwen3-32B#best-practices) into sglang api,
```{python3}
async def async_request_sglang_generate(...) -> RequestFuncOutput:
    ...
        payload = {
            ...
            "sampling_params": {
                "temperature": 0.6,
                "top_k": 20,
                "top_p": 0.95,
                ...
            },
            ...
        }
    ...
```
and use the following test command,
```{shell}
        --model Qwen/Qwen3-32B --apply-chat-template --flush-cache \
        --num-prompt 128 --max-concurrency 128 \
        --dataset-name random-ids --random-input-len 512 --random-output-len 512 --random-range-ratio 1.0 
```
We perform the test on a single Atlas 800T A2 server and use Qwen3-32B model on two NPUs.
```{shell}
# For pytorch
STREAMS_PER_DEVICE=32 python -m sglang.launch_server --model-path /model/qwen3_32b --tp 2 \
    --attention-backend ascend --sampling-backend pytorch --mem-fraction-static 0.85 \

# For ascend
STREAMS_PER_DEVICE=32 python -m sglang.launch_server --model-path /model/qwen3_32b --tp 2 \
    --attention-backend ascend --sampling-backend ascend --mem-fraction-static 0.85 \
```
The results show large speed up with ascend sampling backend.
**bs=128 E2E ITL 361ms -> 91ms, sampling time 274ms -> 5.4ms**
merged

profilemerge

## Checklist


python/sglang/srt/layers/sampler.py
```diff
@@ -112,7 +115,11 @@ def forward(
 
             # Post process logits
             logits.div_(sampling_info.temperatures)
-            logits[:] = torch.softmax(logits, dim=-1)
+            # For ascend backend, softmax is not needed before sampling
+            if not get_global_server_args().sampling_backend == "ascend" or (
+                return_logprob and not SGLANG_RETURN_ORIGINAL_LOGPROB
+            ):
+                logits[:] = torch.softmax(logits, dim=-1)
```
Alexhaoge, line 122: Because softmax of the raw logits is included in the fused op, the initial softmax can be skipped if `return_logprob` is false. Condition is copied from here https://github.com/sgl-project/sglang/blob/dc4f54182374e9a52cefdd0b96eb468b40f04529/python/sglang/srt/layers/sampler.py#L158-L169

ping1jing2:
> @ping1jing2 When you add a new test file under `test/` folder, you should add it to the `run_suite.py`

ok, thanks for your reminder and we are trying

python/sglang/srt/layers/sampler.py
Qiaolin-Yu: But if `return_logprob and not SGLANG_RETURN_ORIGINAL_LOGPROB` is True here, it will do two softmax operations?

Alexhaoge: Originally softmax was always executed.https://github.com/sgl-project/sglang/blob/0296f1cdad8e89a087c0539dd1983049397e110d/python/sglang/srt/layers/sampler.py#L114-L115

We want to skip this softmax as possible when using ascend sampling backend to speed up. The if-clause condition is designed for,
1. if the sampling backend is not `ascend`, always execute softmax;
2. if the sampling backend is `ascend`, because the last branch in `return_logprob` condtional clause computes `logprobs` based on `probs`, we have to retain the softmax for the correctness of `probs`.
https://github.com/sgl-project/sglang/blob/0296f1cdad8e89a087c0539dd1983049397e110d/python/sglang/srt/layers/sampler.py#L168-L169

This branch is reached only if **using ascend sampling and `return_logprob and not SGLANG_RETURN_ORIGINAL_LOGPROB` is True**, and yes, in such case there will be two softmax operations. This scenario is rather uncommon and top_p/top_k sampling indexing will be its performance bottleneck, so it should be okay to compromise a bit for the sake of correctness.

Those will not affect other sampling backends as they will always execute softmax once at the start but do not have softmax operations inside their kernel implementation.

merrymercy: The code is just wrong when `return_logprob and not SGLANG_RETURN_ORIGINAL_LOGPROB` and 
ascend sampling.

You can never push incorrect code. You should raise exceptions.

Alexhaoge: Thank you for the opinions. I check again the code and agree the `if`-clause and previous code comment can be quite confusing, especially for the corner cases on NPU with ascend sampling backend.

#### Other backend behaviors
First I would like to further clarify that the changes did not affect the original bahavior for other backends. The default `sampling_backend` is `flasherinfer` or `pytorch`. The sampling backend will not be `ascend` unless users explicitly set `--sampling-backend ascend` in the launch command (even for NPU). 

So for `flashinfer` and `pytorch`, the first condition 
```
not get_global_server_args().sampling_backend == "ascend"
```
will be `True`, short-circuiting will gurantee the entire condition be `True` so that there will be one and only one softmax before sampling, just the same as the code before this PR.

#### Considerations for ascend sampling backend
1. NPU has poor performance with pytorch sampling backend, mostly because of indexing `[:]`, as shown in the profiling screenshots above, so we use fused op `npu_top_k_top_p`, or naive implementation `masked_fill_`. We would like to make ascend sampling backend works for most of the cases on NPU.
2. The fused op `npu_top_k_top_p` has a softmax inside before top-p fitlering. When using this fused op, softmax before sampling *can be skipped*. I wrote the if clause for this scenario while keeping the softmax when return_logprob is necessary.
3. When using torch native implemetation (as `npu_top_k_top_p` cannot take `top_ks>1024`), I added a softmax operation inside the ascend sampling backend. 
https://github.com/sgl-project/sglang/blob/dda2b8de1993ea1ffd7e20df72d39f3c5d009427/python/sglang/srt/layers/sampler.py#L327-L329
When falling back to this implemention, plus `(return_logprob and not SGLANG_RETURN_ORIGINAL_LOGPROB)` , there are indeed two softmax, one in logits post-processing, and one before top-k sampling. This  does not affect the correctness of logprob, and the filtering of top-k/top-p theoretically should be the same. However, I understand that this is incorrect for the model structure, and the duplicated softmax should be removed.

#### Possible solutions
1. Immediate solution
   - Drop `npu_top_k_top_p` implementation, use torch native implentation only and remove the latter softmax inside `top_k_top_p_min_p_sampling_from_probs_ascend`. In such case, `if`-clause can be removed, and ascend sampling backend becomes consistent with other backends in the logit post-processing part.
   - Dropping `npu_top_k_top_p` is tolerable because the main performace gap lies in tensor indexing `[:]` v.s. `masked_fill_`

2. Future plans 
   - A new fused op for npu top-k/top-p/min-p sampling is under development(https://github.com/sgl-project/sgl-kernel-npu/pull/340). We will rework the ascend sampling backend with it for better performance. The new version also will have clearer logic and be consistent with other backends in `Sampler.forward` part.

merrymercy: You need to make code alway correct and readable. The readability should be there regardless of whether it changed the behavior of other backends or not.

Refactor here: https://github.com/sgl-project/sglang/pull/18915

PR #12693, opened 2025-11-05
fix  mamba prefix cache leak caused by abort

yizhang2077:
## Motivation

In mamba prefix cache match, it needs fork a space to copy mamba state from radix tree and then move the ownership to the request once `HybridReqToTokenPool.alloc` is called. If a request abort between `MambaRadixCache.match_prefix` and   `HybridReqToTokenPool.alloc`, it will cause memory leak. The pr fix this issue.

reproduce command
``` bash
python3 -m sglang.launch_server --model Qwen/Qwen3-Next-80B-A3B-Instruct   --chunked-prefill-size 2048

# after a while, cancel this command, memory leak will occasionally happen
python3 -m sglang.bench_serving  --backend sglang --dataset-name generated-shared-prefix --gsp-num-groups 50 --gsp-prompts-per-group 100 --gsp-system-prompt-len 10240 --gsp-question-len 1024 --gsp-output-len 128 --max-concurrency 50
```
## Checklist

### DIFF
--- a/python/sglang/srt/managers/scheduler.py
+++ b/python/sglang/srt/managers/scheduler.py
@@ -2527,9 +2527,9 @@ class Scheduler(
                 self.tree_cache.cache_finished_req(req)
 
             # For mamba radix cache
-            if req.mamba_pool_idx_reserverd is not None:
+            if req.mamba_pool_idx_reserved is not None:
                 self.tp_worker.model_runner.req_to_token_pool.mamba_pool.free(
-                    req.mamba_pool_idx_reserverd.unsqueeze(-1)
+                    req.mamba_pool_idx_reserved.unsqueeze(-1)
                 )
             logger.debug(f"Abort queued request. {req.rid=}")
 
--- a/python/sglang/srt/managers/scheduler.py
+++ b/python/sglang/srt/managers/scheduler.py
@@ -2526,6 +2526,11 @@ class Scheduler(
             if self.disaggregation_mode == DisaggregationMode.DECODE:
                 self.tree_cache.cache_finished_req(req)
 
+            # For mamba radix cache
+            if req.mamba_pool_idx_reserverd is not None:
+                self.tp_worker.model_runner.req_to_token_pool.mamba_pool.free(
+                    req.mamba_pool_idx_reserverd.unsqueeze(-1)
+                )
             logger.debug(f"Abort queued request. {req.rid=}")
 
         # Delete the requests in the grammar queue
--- a/python/sglang/srt/managers/scheduler.py
+++ b/python/sglang/srt/managers/scheduler.py
@@ -2527,10 +2527,8 @@ class Scheduler(
                 self.tree_cache.cache_finished_req(req)
 
             # For mamba radix cache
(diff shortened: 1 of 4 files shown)


python/sglang/srt/mem_cache/memory_pool.py
```diff
@@ -415,6 +415,7 @@ def alloc(
                 req.mamba_pool_idx = mid
             if mid is not None:
                 mamba_index.append(mid)
+                req.mamba_pool_idx_reserved = None
```
hanming-lu: Do you think putting in `if req.mamba_pool_idx is not None:  # for radix cache` is cleaner?

python/sglang/srt/managers/scheduler.py
```diff
@@ -2526,6 +2526,11 @@ def abort_request(self, recv_req: AbortReq):
             if self.disaggregation_mode == DisaggregationMode.DECODE:
                 self.tree_cache.cache_finished_req(req)
 
+            # For mamba radix cache
+            if req.mamba_pool_idx_reserved is not None:
+                self.tp_worker.model_runner.req_to_token_pool.mamba_pool.free(
+                    req.mamba_pool_idx_reserved.unsqueeze(-1)
+                )
```
hanming-lu: Do you think calling `cache_finished_req(req)` similar to DECODE case above is the cleaner approach here? Then we don't need the `mamba_pool_idx_reserved` field.

To make it work, we likely need to make two small changes in mamba radix cache:
1. to handle the case of `req.output_ids` being 0 (also needed for retract edge case): `token_ids = (req.origin_input_ids + req.output_ids)[:-1]` -> 
```
cache_len = len(req.origin_input_ids) + max(len(req.output_ids) - 1, 0)
token_ids = (req.origin_input_ids + req.output_ids)[:cache_len]
```
2. support `is_insert` param similar to radix cache and make it False here.

luismiguelmariano77-hash (approved):

python/sglang/srt/managers/scheduler.py
yizhang2077: fix it

hanming-lu (approved):

python/sglang/srt/managers/scheduler.py
```diff
@@ -2526,6 +2526,9 @@ def abort_request(self, recv_req: AbortReq):
             if self.disaggregation_mode == DisaggregationMode.DECODE:
                 self.tree_cache.cache_finished_req(req)
 
+            # For mamba radix cache
+            if req.mamba_pool_idx is not None:
+                self.tree_cache.cache_finished_req(req, is_insert=False)
```
hanming-lu, line 2531: do you think we need to handle the case for `DECODE`? specifically should we set `req.mamba_pool_idx = None` after the cache_finished_req above?

yizhang2077: do you mean `retract_decode`? I think resources have been released in `release_req`? and `mamba_pool_idx` will be set to None in `reset_for_retract`

PR #12694, opened 2025-11-05
[feature]add capacity control in file storagebackend

Bruce-x-1997:
## Motivation

As we use filebackend,  there is no evict and capacity control logic, so the filebackend could always put cache until the nvme is full.so I add the related logic to help use the nvme/filebackend case

## Modifications

1.  add LRU update & evict logic when calling set
2. add capacity config env (SGLANG_HICACHE_FILE_BACKEND_MAX_CAPACITY_GB) and free_bytes to trigger lru 
3. add SGLANG_HICACHE_FILE_BACKEND_STORAGE_DIR clear logic when init

## Checklist

### DIFF
--- a/python/sglang/srt/managers/cache_controller.py
+++ b/python/sglang/srt/managers/cache_controller.py
@@ -787,3 +787,16 @@ class HiCacheController:
 
             except Empty:
                 continue
+
+    def queue_storage_cleanup(self, keys: List[str]):
+        """
+        Queue storage keys for background cleanup.
+        This method delegates to the storage backend's cleanup mechanism.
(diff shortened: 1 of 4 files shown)


Bruce-x-1997:
> Hi @Bruce-x-1997 is there a reason that you want to use the file backend instead of the others? This backend is intended to be a demo therefore I wanted to keep it easy and clean. But if there is indeed a strong need for it we can plan things more thoroughly.

hi, just because that we have no storage nodes like vast .etc
so I wanna use the local storage like nvme, and I see filebackend is the only one what could support local storage?

PR #12701, opened 2025-11-05
Modified the moe_align_block method in sgl-kernel to uniform handling of invalid expert_id

Yuechguo:
… of expert_id.

## Motivation
uniform handling of invalid expert_id in sgl-kernel and fix bug in fused_experts_impl in some EP mode case

## Modifications
1.modify moe_align_block kernel to uniform handling of invalid expert_id
2.fix bug fused_experts_impl of fuse_moe in which may cause garbled output in some EP mode cases

## Accuracy Tests
```
pytest sgl-kernel/tests/test_moe_align.py
```

```
CUDA_VISIBLE_DEVICES=0,1,2,3,4,5,6,7 \
        --model-path /pretrained_model/qwen/Qwen3-30B-A3B \
        --served-model-name  Qwen3-30B-A3B \
        --attention-backend triton \
        --host 127.0.0.1 \
        --port 28000 \
        --tp-size 8 --ep-size 8 --trust-remote-code \
        --chunked-prefill-size 32768 \
        --max-prefill-tokens 32768 \
        --context-len 32768 \
        --max-running-requests 512 \
        --mem-fraction-static 0.85 \
        --watchdog-timeout 600 \

python3 /data/testhome/sglang/benchmark/gsm8k/bench_sglang.py --num-shots 5 --num-questions 100 --port 28000
```
if we do not fix the bug in the  fused_experts_impl of fuse_moe, After launching SGLang using the config mentioned above, garbage output may be generated.

## Checklist


hubertlu-tw:
@Yuechguo could you please add the command for EP mode cases along with the error messages or update the current unit test script (`sgl-kernel/tests/test_moe_align.py`) to capture the breakage?

python/sglang/srt/layers/moe/fused_moe_triton/moe_align_block_size.py
```diff
@@ -54,7 +54,7 @@ def moe_align_block_size(
     - The padding ensures that the total number of tokens is now divisible
         by block_size for proper block matrix operations.
     """
-    max_num_tokens_padded = topk_ids.numel() + (num_experts + 1) * (block_size - 1)
```
luismiguelmariano77-hash, line 57: (Num_experts -1)

luismiguelmariano77-hash (approved):

python/sglang/srt/layers/moe/fused_moe_triton/fused_moe.py
```diff
@@ -438,7 +438,7 @@ def fused_experts_impl(
         min(M * topk, E + 1) * (max_block_m - 1) if down_moe_use_tma else 0
     )
     total_tokens = M * topk + max_padded_tokens
-    cache = torch.empty(
+    cache = torch.zeros(
```
ch-wan, line 452: zeros may cause performance degradation

Yuechguo: In EP case, some local selected experts are very few in number. In extreme cases, random data from the cache generated by torch.empty may get mixed into the computations, resulting in garbage output.

hubertlu-tw (approved):
LGTM

python/sglang/srt/layers/moe/fused_moe_triton/fused_moe.py
ch-wan: I don't get your point. As you have `intermediate_cache3.fill_(0)`, having non zeros in intermediate results is not an issue. Also, I don't think `intermediate_cache3.fill_(0)` is needed. Check my comment below.

python/sglang/srt/layers/moe/fused_moe_triton/fused_moe.py
```diff
@@ -505,7 +506,7 @@ def fused_experts_impl(
         intermediate_cache1 = cache[: total_tokens * N].view(
             (total_tokens, N),
         )
-        intermediate_cache2 = torch.empty(
+        intermediate_cache2 = torch.zeros(
```
ch-wan, line 509: These zeros op can be removed too.

python/sglang/srt/layers/moe/fused_moe_triton/fused_moe.py
```diff
@@ -578,6 +579,7 @@ def fused_experts_impl(
         else:
             raise ValueError(f"Unsupported activation: {activation=}, with {is_gated=}")
 
+        intermediate_cache3.fill_(0)
```
ch-wan, line 582: It introduces a redundant kernel. Is it needed? We zero out padded tokens here: https://github.com/sgl-project/sglang/blob/227728e173baea3d2b7a0c394ac8883ccb2447ff/python/sglang/srt/layers/moe/fused_moe_triton/fused_moe_triton_kernels.py#L176-L192

ch-wan (commented):
I think all `zeros` op can be avoided by filling `-1` in `expert_ids` during its initialization. Correct me if I'm wrong.

python/sglang/srt/layers/moe/fused_moe_triton/moe_align_block_size.py
```diff
@@ -66,12 +66,12 @@ def moe_align_block_size(
 
     # In EP, expert_ids for filtered experts are -1. We have num_experts + 1 ids in total.
```
ch-wan, line 67: remove this comment

PR #12715, opened 2025-11-05
Fix hang in deepgemm compilation with symmetric memory enabled

nvcastet:
## Motivation

Disable symm memory contexts for dummy tensors used in deepgemm compilations since compilation happens on a single gpu.

CC @merrymercy 

## Checklist

### DIFF
--- a/python/sglang/srt/distributed/device_communicators/pynccl_allocator.py
+++ b/python/sglang/srt/distributed/device_communicators/pynccl_allocator.py
@@ -62,7 +62,7 @@ _allocator = None
 _mem_pool = None
 _graph_pool_id = None
 _cur_device = None
-_active_symmetric_memory_contexts = []
+_active_symmetric_memory_context = None
 
 
 def is_symmetric_memory_enabled():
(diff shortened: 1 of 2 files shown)


python/sglang/srt/distributed/device_communicators/pynccl_allocator.py
```diff
@@ -57,6 +57,7 @@
 _mem_pool = None
 _graph_pool_id = None
 _cur_device = None
+_active_symmetric_memory_contexts = []
```
merrymercy: we can guarantee that there will only be one active symmetric memory context, so we do not need to use a list. (IIRC, with the current code, if you use nested symmetric emmory context, it will already hit some errors) 
This is for reducing the overhead

merrymercy (approved):

python/sglang/srt/distributed/device_communicators/pynccl_allocator.py
nvcastet: Ok I can switch to single active context

PR #12743, opened 2025-11-06
Fix: Separate /health and /health_generate endpoints (#12731)

mathisonzc:
This commit addresses issue #12731 by separating the /health and /health_generate endpoints to serve different purposes:

- /health: Now a lightweight liveness probe that only checks server status without generating tokens. Returns 200 if server is up and operational, 503 if starting, unhealthy, or shutting down.

- /health_generate: Remains a readiness probe that actually tests token generation capability by sending a special request.

This distinction is important for Kubernetes deployments where:
- Liveness probes should be lightweight to avoid false failures
- Readiness probes should verify actual functionality

Changes:
- Created new lightweight health() function for /health endpoint
- Updated health_generate() documentation to clarify its purpose
- Both endpoints handle graceful shutdown and server status checks

Fixes #12731

## Checklist


python/sglang/srt/entrypoints/http_server.py
```diff
@@ -403,11 +403,38 @@ async def validate_json_request(raw_request: Request):
 
 
 @app.get("/health")
+async def health(request: Request) -> Response:
```
zhyncs, line 406: I think that we should not separate these endpoints cc @whybeyoung

zhyncs (changes requested):

python/sglang/srt/entrypoints/http_server.py
mathisonzc: Thanks for the review @zhyncs!

I'd like to clarify the motivation behind this change:

According to issue #12731, the current implementation where both endpoints behave identically causes problems in Kubernetes deployments.
Specifically:

1. **Liveness probes** (`/health`) should be lightweight checks that verify the container is alive, without performing actual work
2. **Readiness probes** (`/health_generate`) should test actual functionality

The issue reporter (@slobodaapl) noted: "if a model is busy or fully constrained, but the container isn't dead, in strict health checking
K8s deployment scenarios, a container may be forcefully reset if liveness probe fails."

This PR separates the endpoints to follow Kubernetes best practices:
- `/health` - Quick status check (no token generation)
- `/health_generate` - Full functionality test

Happy to discuss alternative approaches if there are concerns about this separation! @whybeyoung

whybeyoung:
What's the point of your health endpoint returning 200 if you separate them? It will still return 200 even if it freezes—anytime the container doesn't crash, it returns 200. This has violated the original meaning of "health" itself, so there's no need to separate them.
Do you have any specific use cases? I think you should rather adjust the probe configuration of your Kubernetes cluster.
@mathisonzc

slobodaapl:
@whybeyoung what really constitutes a practical scenario of "it freezes" in this context? 

I can't imagine a situation where a model or inference randomly freezes without the entire process (uvicorn included) freezing -- this is a real case of "this needs a restart".

In the case of a freeze or delay of generation alone, it likely means we are about to get a very useful log that we don't just want to discard.

Either way, I don't believe the current setup follows K8s best practices for liveness probe definition, and thusly lacks a real consistent readiness probe by this definition.

It's true at the very least that documentation should be updated to reflect the exact implementation and semantics of the endpoints, but I'm not sure it addresses the whole problem from my point of view.

whybeyoung:
> @whybeyoung what really constitutes a practical scenario of "it freezes" in this context?
> 
> I can't imagine a situation where a model or inference randomly freezes without the entire process (uvicorn included) freezing -- this is a real case of "this needs a restart".
> 
> In the case of a freeze or delay of generation alone, it likely means we are about to get a very useful log that we don't just want to discard.
> 
> Either way, I don't believe the current setup follows K8s best practices for liveness probe definition, and thusly lacks a real consistent readiness probe by this definition.
> 
> It's true at the very least that documentation should be updated to reflect the exact implementation and semantics of the endpoints, but I'm not sure it addresses the whole problem from my point of view.


Thank you for your correction; we will update the documentation to address this matter. Additionally, the "freeze scenario" I mentioned actually refers to some unexpected unresponsive behaviors. For such scenarios, detection must indeed be implemented through the "generate" function. What I mean is that in actual production (prod) scenarios, "health" and "health_generate" should essentially be the same thing, as they both indicate whether the service itself is healthy. There is little need to separate them additionally.
Regarding the log issue you mentioned, I believe it is unlikely for any Kubernetes (k8s) production scenario to lack log collection components similar to Elasticsearch (ES) and Kibana. What I mean is that logs should at least be persisted temporarily, right?

PR #12746, opened 2025-11-06
WIP: docker: include diffusion dependencies in dockerfile

mickqian:
## Checklist


python/pyproject.toml
```diff
@@ -94,6 +94,8 @@ diffusion = [
     "vsa==0.0.4",
 ]
 
+all = ["sglang[diffusion]"]
```
zhyncs: actually we just removed `all` a few days ago so that we can use `pip install sglang`. why do we add it back?

zhyncs (changes requested):

python/pyproject.toml
mickqian: can we include diffusion dependencies when building docker from `Dockerfile`?

mickqian: I'm guessing we have to modify `Dockerfile`?

Issue #12695, opened 2025-11-05
[Bug] GPT-OSS-20B runs into cuda illegal memory access with the default flashinfer_mxfp4 moe runner backend


merrymercy:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Describe the bug

By default, sglang chooses flashinfer_mxfp4 as the moe backend for gpt oss 20b. but this backend leads to illegal memory access.

### Reproduction

This works fine
```
python3 -m sglang.launch_server --model openai/gpt-oss-20b --moe-runner-backend triton_kernel
```

But this runs into errors
```
python3 -m sglang.launch_server --model openai/gpt-oss-20b --moe-runner-backend flashinfer_mxfp4
```

```
[2025-11-05 11:44:09] KV Cache is allocated. #tokens: 3008320, K size: 68.86 GB, V size: 68.86 GB
[2025-11-05 11:44:09] Memory pool end. avail mem=23.34 GB
[2025-11-05 11:44:09] Capture cuda graph begin. This can take up to several minutes. avail mem=22.28 GB
[2025-11-05 11:44:09] Capture cuda graph bs [1, 2, 4, 8, 12, 16, 24, 32, 40, 48, 56, 64, 72, 80, 88, 96, 104, 112, 120, 128, 136, 144, 152, 160, 168, 176, 184, 192, 200, 208, 216, 224, 232, 240, 248, 256, 272, 288, 304, 320, 336, 352, 368, 384, 400, 416, 432, 448, 464, 480, 496, 512]
Capturing batches (bs=168 avail_mem=21.42 GB):  52%|_____________________________                          | 27/52 [00:01<00:01, 15.05it/s][2025-11-05 11:44:11] Scheduler hit an exception: Traceback (most recent call last):
  File "/root/sglang/python/sglang/srt/model_executor/cuda_graph_runner.py", line 381, in __init__
    self.capture()
[...]

PR #12758, opened 2025-11-06
[Bugfix] Fix illegal memory access

elvischenv:
## Motivation

Fixed illegal memory access issue in #12695 and https://github.com/flashinfer-ai/flashinfer/issues/2034
Caused by #12524

## Checklist

### DIFF
--- a/python/sglang/srt/layers/flashinfer_comm_fusion.py
+++ b/python/sglang/srt/layers/flashinfer_comm_fusion.py
@@ -96,7 +96,7 @@ _workspace_manager = FlashInferWorkspaceManager()
 
 
 def ensure_workspace_initialized(
-    max_token_num: int = 2048, hidden_dim: int = 4096, use_fp32_lamport: bool = False
+    max_token_num: int = 16384, hidden_dim: int = 4096, use_fp32_lamport: bool = False
 ):
     """Ensure workspace is initialized"""
     if not is_flashinfer_available() or _flashinfer_comm is None:
@@ -128,7 +128,7 @@ def flashinfer_allreduce_residual_rmsnorm(
     residual: torch.Tensor,
     weight: torch.Tensor,
     eps: float = 1e-6,
-    max_token_num: int = 2048,
+    max_token_num: int = 16384,
     use_oneshot: Optional[bool] = None,
     trigger_completion_at_end: bool = False,
     fp32_acc: bool = False,
@@ -158,6 +158,15 @@ def flashinfer_allreduce_residual_rmsnorm(
     world_size = get_tensor_model_parallel_world_size()
     if world_size <= 1:
         logger.debug("Single GPU, no need for allreduce fusion")
+        return None, None
+
+    if input_tensor.shape[0] > max_token_num:
+        logger.debug(
+            "Input token(%d) is greater than max_token_num(%d), "
+            "falling back to standard implementation",
+            input_tensor.shape[0],
+            max_token_num,
+        )
         return None, None
 
     if not ensure_workspace_initialized(
@@ -205,7 +214,7 @@ def fake_flashinfer_allreduce_residual_rmsnorm(
     residual: torch.Tensor,
     weight: torch.Tensor,
     eps: float = 1e-6,
-    max_token_num: int = 2048,
+    max_token_num: int = 16384,
     use_oneshot: Optional[bool] = None,
     trigger_completion_at_end: bool = False,
     fp32_acc: bool = False,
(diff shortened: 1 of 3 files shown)


python/sglang/srt/layers/quantization/mxfp4.py
```diff
@@ -644,7 +644,7 @@ def apply(
             with use_symmetric_memory(
                 get_tp_group(), disabled=not is_allocation_symmetric()
             ):
-                symm_output = torch.empty_like(x)
+                symm_output = torch.empty_like(x_quant, dtype=x.dtype)
```
Fridge003: I just tested  accuracy with this PR, and it seems incorrect (0.013 for mmlu)
@elvischenv  Please have a check

Fridge003 (changes requested):

python/sglang/srt/layers/quantization/mxfp4.py
elvischenv: What model are you testing? Would you try hardcoding to torch.bfloat16?
```
symm_output = torch.empty_like(x_quant, dtype=torch.bfloat16)
```
related code in flashinfer:
https://github.com/flashinfer-ai/flashinfer/blob/main/flashinfer/fused_moe/core.py#L1498-L1503

Fridge003: gpt-oss 20b with mxfp4

elvischenv: This fix is correct since the output tensor should be padded, the original `x` is unpadded, and this could fix the illegal memory access issue. I can run the benchmark serving without any error.

However, when I tried running `python -m gpt_oss.evals ...` to evaluate the accuracy, it just hung. There is still some other issues for running gpt-oss.

python/sglang/srt/layers/flashinfer_comm_fusion.py
```diff
@@ -128,7 +128,7 @@ def flashinfer_allreduce_residual_rmsnorm(
     residual: torch.Tensor,
     weight: torch.Tensor,
     eps: float = 1e-6,
-    max_token_num: int = 2048,
+    max_token_num: int = 16384,
```
elvischenv, line 131: Hang issue WAR: Increase `max_token_num` to allocate a larger workspace

python/sglang/srt/layers/quantization/mxfp4.py
elvischenv: With the hanging issue fix, I can get reasonable accuracy score:
```
[{'eval_name': 'gpqa', 'model_name': 'gpt-oss-20b-medium_temp1.0_20251106_155616', 'metric': 0.6496212121212122}]
```

elvischenv:
> Failed at GPTOSS Ci test, please have a look
https://github.com/sgl-project/sglang/actions/runs/19149049114/job/54736119046?pr=12758

This is DeepseekV2.
```
  File "/actions-runner/_work/sglang/sglang/python/sglang/srt/models/deepseek_v2.py", line 793, in forward_normal_dual_stream
    final_hidden_states += shared_output
RuntimeError: The size of tensor a (3584) must match the size of tensor b (7168) at non-singleton dimension 1
```
This failure seems also related to #12524. cc @merrymercy 
Did #12524 merge without passing the CI?

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -793,7 +798,13 @@ def forward_normal_dual_stream(
...
 
         current_stream.wait_stream(self.alt_stream)
-        final_hidden_states += shared_output
+        with use_symmetric_memory(
+            parallel_state.get_tp_group(), disabled=not is_allocation_symmetric()
+        ):
+            final_hidden_states_out = torch.empty_like(final_hidden_states)
+
+        torch.add(final_hidden_states, shared_output, out=final_hidden_states_out)
+        final_hidden_states = final_hidden_states_out
```
elvischenv: Reverted that changes and let's see if it can pass.

python/sglang/srt/layers/flashinfer_comm_fusion.py
nvpohanh: add an assert: `assert input_tensor.shape[0] <= max_token_num` ?

nvpohanh: should be covered by the code below:

```
    if input_tensor.shape[0] > max_token_num:
        logger.debug(
            "Input token(%d) is greater than max_token_num(%d), "
            "falling back to standard implementation",
            input_tensor.shape[0],
            max_token_num,
        )
        return None, None
```

Fridge003 (approved):

python/sglang/srt/layers/quantization/mxfp4.py
Fridge003: Just realized I was using a  mmlu benchmark command which didn't apply chat template, so caused a low accuracy.
Your result should be correct since it's passing the CI

PR #12761, opened 2025-11-06
fixes hardcoded "cuda" device references in unit tests to use a dynamic device selection

kalyank007:
fixes hardcoded "cuda" device references in unit tests to use a dynamic device selection function instead. This change improves test portability across different hardware backends (e.g., CUDA, XPU, ROCm, CPU).

Key changes:
Replaced all hardcoded "cuda" string literals with get_device() function calls, Applied changes to two test methods: test_build_unified_kv_indices and _test_extend_attention_unified_vs_regular_once

### DIFF
--- a/test/srt/test_triton_attention_kernels.py
+++ b/test/srt/test_triton_attention_kernels.py
@@ -18,6 +18,7 @@ from sglang.srt.layers.attention.triton_ops.extend_attention import (
 from sglang.srt.layers.attention.triton_ops.prefill_attention import (
     context_attention_fwd,
 )
+from sglang.srt.utils import get_device
 from sglang.test.test_utils import CustomTestCase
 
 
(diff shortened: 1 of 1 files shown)


kalyank007:
> ## Code Review
> This pull request is a good step towards making the unit tests device-agnostic by replacing hardcoded "cuda" strings with dynamic device selection.
> 
> However, the implementation is incomplete. The changes are only applied to `_test_extend_attention_unified_vs_regular_once` and `test_build_unified_kv_indices`, but many other test methods in `test/srt/test_triton_attention_kernels.py` still have hardcoded `"cuda"` device references. To fully achieve the goal of this PR, I recommend updating the following methods as well:
> 
> * `_test_extend_attention_once`
> * `_test_extend_attention_sliding_window_once`
> * `_test_context_attention_once`
> * `_test_decode_attention_once`
> * `_test_grouped_decode_attention_once`
> 
> Most importantly, the current changes introduce a `NameError` because the `get_device` function is used without being imported. I've added a specific comment pointing out where this needs to be fixed.

resolved

hnyls2002 (approved):

PR #12771, opened 2025-11-06
Add intel_xpu as backend for GptOssForCausalLM, enabled for bf16 models

ck-intel:
## Motivation

Enable GptOssForCausalLM on Intel XPU for bf16 dtype.

## Modifications

Add intel_xpu as supported backend for GptOssForCausalLM model.
Update forward_moe_native function to add following changes required for GPT-OSS MoE architetecture:
 - Bias addition support in linear layers of MoE
 - swiglu (with alpha and limit) activation support

## Accuracy Tests
To verify the correctness of the changes I ran the following tests on Intel XPU

1. Accuracy Unit Test (The test is passing)
```
python -m pytest test/registered/core/test_gpt_oss_1gpu.py::TestGptOss1Gpu::test_bf16_20b

Score: 0.348
Writing report to /tmp/gpqa_lmsys_gpt-oss-20b-bf16.html
{'chars': np.float64(11745.79797979798), 'chars:std': np.float64(4440.787062781283), 'score:std': np.float64(0.4764904604095875), 'score': np.float64(0.3484848484848485)}
```

2. GSM8K Test:
```
# Run server
python3 -m sglang.launch_server --model lmsys/gpt-oss-20b-bf16 --disable-radix-cache --tp 4 --attention-backend intel_xpu

# Run client
lm-eval --model local-chat-completions --model_args model=lmsys/gpt-oss-20b-bf16,base_url=http://127.0.0.1:30000/v1/chat/completions,max_gen_toks=1024,num_concurrent=128 --tasks gsm8k --apply_chat_template --num_
fewshot 1

# Result
local-chat-completions (model=lmsys/gpt-oss-20b-bf16,base_url=http://127.0.0.1:30000/v1/chat/completions,max_gen_toks=1024,num_concurrent=128), gen_kwargs: (None), limit: None, num_fewshot: 1, batch_size: 1
|gsm8k|      3|flexible-extract|     1|exact_match|↑  |0.8522|±  |0.0098|
|     |       |strict-match    |     1|exact_match|↑  |0.0296|±  |0.0047|
```

3. MMLU Test:
```
Average accuracy: 0.010
```

## Benchmarking and Profiling

Not Applicable

## Checklist


python/sglang/srt/server_args.py
```diff
@@ -939,7 +939,7 @@ def _handle_model_specific_adjustments(self):
                 else:
                     self.attention_backend = "triton"
 
-            supported_backends = ["triton", "trtllm_mha", "fa3", "fa4"]
+            supported_backends = ["triton", "trtllm_mha", "fa3", "fa4", "intel_xpu"]
```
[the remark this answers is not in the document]
ck-intel: the attention_backend can also be "triton", and the model will work on xpu device.

ck-intel:
> @ck-intel do we have test cases to cover this ? have you run the model on your local machines?
> 
> CC @jianan-gu

Hi @mingfeima pushed a commit which was required for accuracy fix and updated the description with the required details.
I couldn't run the complete accuracy test due to memory constraint, looking at ways to run it (with xpu cards/memory available to me). Though I was able to run the 20B-bfloat16 model on a small number of prompts and the output seems sanity wise fine.  I will update the PR once I run the full accuracy test.

python/sglang/srt/layers/moe/fused_moe_native.py
```diff
@@ -87,9 +90,52 @@ def moe_forward_native(
...
+
+        # Apply w13 linear with bias for models like GPT-OSS
+        if layer_w13_bias is not None:
+            # BF16 matmul with FP32 accumulation
+            gate_up = F.linear(tokens_for_this_expert, layer_w13_weight, bias=None)
+            gate_up_fp32 = gate_up.float() + layer_w13_bias
+            gate_up = gate_up_fp32.to(original_dtype)
+
+            # Split the interleaved gate and up projections as per GPT-OSS weight layout:
+            gate, up = gate_up[..., ::2], gate_up[..., 1::2]
```
mingfeima: the logic here needs to be more restrict.

right now, gpt-oss is the only model that has bias and also interleaved gate and up proj. but this may be changed in the future.

how does Model passed the information that it requires gate/up interleaving to `moe_forward_native?

@jianan-gu i remembered that sglang applied a shuffle in the weights to make gpt-oss in a non-interleaved weights here, can you point out?

python/sglang/srt/layers/moe/fused_moe_native.py
```diff
@@ -87,9 +90,52 @@ def moe_forward_native(
...
+                    max=moe_runner_config.gemm1_clamp_limit,
+                )
+                glu = gate * torch.sigmoid(gate * moe_runner_config.gemm1_alpha)
+                gate_up = glu * (up + 1)
+            else:
+                gate_up = act(gate_up)
+        else:
+            # Direct path for models without bias (e.g., Mixtral, DeepSeek)
+            gate_up = F.linear(tokens_for_this_expert, layer_w13_weight)
+            gate_up = act(gate_up)
```
jianan-gu, line 121: Does it mean only when a model having `layer_w13_bias` , it can use other act functions like `swiglu ` here? This seems not common enough for native MoE path here but specific for GPToss model.
Also, linear+bias can be one line of `F.linear(gate_up, layer_w2_weight, bias=layer_w2_bias)`.
Thus, those changes can be simplified:

```
gate_up = F.linear(tokens_for_this_expert, layer_w13_weight, bias = layer_w13_bias)
gate_up = act(gate_up) # act can be any activation functions depends on the runtime check.
expert_out = F.linear(gate_up, layer_w2_weight, bias=layer_w2_bias)
```

python/sglang/srt/layers/moe/fused_moe_native.py
jianan-gu: `silu_and_mul `does not need this line

python/sglang/srt/server_args.py
```diff
@@ -1069,10 +1070,24 @@ def _handle_model_specific_adjustments(self):
                     self.attention_backend = "trtllm_mha"
                 elif is_cuda() and is_sm90_supported():
                     self.attention_backend = "fa3"
+                elif is_xpu():
```
jianan-gu, line 1861: GPToss model requires both attention sink and sliding windows support.
As attention sink is supported in `intel_xpu  ` backend, however, for sliding window it will have dependency of this PR  (https://github.com/sgl-project/sglang/pull/13561) for the support (need to make sure included in your accuracy tests).

python/sglang/srt/layers/moe/fused_moe_native.py
ck-intel: Thanks for the feedback.
I will disassociate bias addition from the activation part, so swiglu can be applied in general. 
The bias dtype is fp32 and the layer weights are in bfloat16 and thus cannot simplify the bias addition by passing it to F.linear

python/sglang/srt/server_args.py
ck-intel: @jianan-gu I think sink support was already present for MHA, I have reviewed xpu_backend.py from this PR #13561 and have put my comments for this part. It would be great if you can review that section of xpu_backend.py, as sink support is critical for this model. Thanks.

python/sglang/srt/layers/moe/fused_moe_native.py
ck-intel: This will be part of only swiglu_with_alpha_and_limit activation so this will not impact other activations such as silu_and_mul.

python/sglang/srt/server_args.py
jianan-gu: Yes, for sink support it is, while could you also check the sliding window requirement? We may need to make it clear about the PR dependency in this PR description. Thanks. BTW, do we have acc reference number? It seems like MMLU result (0.01) is lower than expectation.

ck-intel: Hi @jianan-gu We do not have any dependency on the PR#13561 for the native MoE support added in this PR for gpt-oss. I also verified about the working of sliding window attention with the sgl cutlass flash attention kernel author, though he said it is not fully optimized yet.

Here are the reference results on Nvidia A100:
GSM8K Test - num shots 8:
Accuracy: 0.590
Invalid: 0.050

MMLU Test:
Average accuracy: 0.009

Thanks!

jianan-gu: ok, the acc results seem not testing full dataset, right? as like MMLU is usually 0.8x in other models.
The overall changes LGTM. Just one more check, it there any UT to cover the native MoE changes?

ck-intel: This is how I ran the MMLU test on A100 and the detailed results, pls let me know if there is an issue with the run.
I will check the UT part and update it if required for the MoE changes done here.

```
## MMLU Test:

cd benchmark/mmlu
bash download_data.sh
cd ../.. # return to the main sglang directory

# Run the server
python3 -m sglang.launch_server --model lmsys/gpt-oss-20b-bf16 --disable-radix-cache --trust-remote-code --device cuda --tp 1

# Run the client:
cd benchmark/mmlu
python3 bench_sglang.py --ntrain 5

subject: abstract_algebra, #q:100, acc: 0.000
subject: anatomy, #q:135, acc: 0.015
subject: astronomy, #q:152, acc: 0.000
subject: business_ethics, #q:100, acc: 0.000
subject: clinical_knowledge, #q:265, acc: 0.008
subject: college_biology, #q:144, acc: 0.000
subject: college_chemistry, #q:100, acc: 0.000
subject: college_computer_science, #q:100, acc: 0.000
subject: college_mathematics, #q:100, acc: 0.000
subject: college_medicine, #q:173, acc: 0.000
subject: college_physics, #q:102, acc: 0.000
subject: computer_security, #q:100, acc: 0.000
subject: conceptual_physics, #q:235, acc: 0.000
subject: econometrics, #q:114, acc: 0.000
subject: electrical_engineering, #q:145, acc: 0.000
subject: elementary_mathematics, #q:378, acc: 0.000
subject: formal_logic, #q:126, acc: 0.000
subject: global_facts, #q:100, acc: 0.000
subject: high_school_biology, #q:310, acc: 0.000
subject: high_school_chemistry, #q:203, acc: 0.000
subject: high_school_computer_science, #q:100, acc: 0.010
subject: high_school_european_history, #q:165, acc: 0.006
subject: high_school_geography, #q:198, acc: 0.000
subject: high_school_government_and_politics, #q:193, acc: 0.000
subject: high_school_macroeconomics, #q:390, acc: 0.013
subject: high_school_mathematics, #q:270, acc: 0.000
subject: high_school_microeconomics, #q:238, acc: 0.000
subject: high_school_physics, #q:151, acc: 0.000
subject: high_school_psychology, #q:545, acc: 0.000
subject: high_school_statistics, #q:216, acc: 0.005
subject: high_school_us_history, #q:204, acc: 0.127
subject: high_school_world_history, #q:237, acc: 0.013
subject: human_aging, #q:223, acc: 0.027
subject: human_sexuality, #q:131, acc: 0.023
subject: international_law, #q:121, acc: 0.008
subject: jurisprudence, #q:108, acc: 0.028
subject: logical_fallacies, #q:163, acc: 0.000
subject: machine_learning, #q:112, acc: 0.000
subject: management, #q:103, acc: 0.000
subject: marketing, #q:234, acc: 0.026
subject: medical_genetics, #q:100, acc: 0.000
subject: miscellaneous, #q:783, acc: 0.015
subject: moral_disputes, #q:346, acc: 0.000
subject: moral_scenarios, #q:895, acc: 0.000
subject: nutrition, #q:306, acc: 0.000
subject: philosophy, #q:311, acc: 0.000
subject: prehistory, #q:324, acc: 0.000
subject: professional_accounting, #q:282, acc: 0.000
subject: professional_law, #q:1534, acc: 0.023
subject: professional_medicine, #q:272, acc: 0.018
subject: professional_psychology, #q:612, acc: 0.000
subject: public_relations, #q:110, acc: 0.000
subject: security_studies, #q:245, acc: 0.008
subject: sociology, #q:201, acc: 0.010
subject: us_foreign_policy, #q:100, acc: 0.030
subject: virology, #q:166, acc: 0.000
subject: world_religions, #q:171, acc: 0.000
Total latency: 391.765
Average accuracy: 0.009

```

jianan-gu: just use "python3 bench_sglang.py" to launch the test

ck-intel: @jianan-gu same results with this command too. Though it seems a bit odd as the MMLU score is low on Nvidia GPU compared to reported in the gpt-oss paper. 
Also, I checked the MoE tests in the repo and there is no unit test for moe_forward_native function.

ck-intel: Updated the lm-eval based gsm8k results in the PR description at the top.

polisettyvarma (approved):

mingfeima (approved):
A few comments:
* i remembered that gpt-oss weights are shuffled when doing preloading, to make it normal as other MoE models.
* we can simplified the code a little bit by overloading only `act()` with or without bias, and the rest of the code doesn't need to be changed.

python/sglang/srt/layers/moe/fused_moe_native.py
ck-intel: The code snippet you referred is outdated, this has been refactored and in-line with how it is handled in the triton flow. The interleaved part is handled within the activation swiglu_with_alpha_and_limit, which is right now applicable only for gpt-oss type of models. 
Also, the bias is also handled in a generic way.

ck-intel:
@Kangyan-Zhou looking forward to your review, Thanks!

polisettyvarma:
@Kangyan-Zhou, @iforgetmyname, @Fridge003, @merrymercy, @ispobock, @JustinTong0323, @BBuf, @Edwardf0t1, @HaiShaw, @Ying1123, @ch-wan @hnyls2002 and @kushanam please review and merge this

mingfeima:
@ck-intel  update here https://github.com/sgl-project/sglang/blob/main/python/pyproject_xpu.toml#L38 so as to fix CI failure. if we need new python library dependency. Also please check CUDA counterpart, if it is really a must to add it.

@ZailiWang for more comments.

ZailiWang:
> @ck-intel update here https://github.com/sgl-project/sglang/blob/main/python/pyproject_xpu.toml#L38 so as to fix CI failure. if we need new python library dependency. Also please check CUDA counterpart, if it is really a must to add it.
> 
> @ZailiWang for more comments.

Yeah, please check why `matplotlib` is required, as from the change in this PR it looks like not introducing the dependency of this lib.

ZailiWang:
From the error trace it is likely that `matplotlib` is introduced in the `transformers` upgrade introduced in https://github.com/sgl-project/sglang/pull/21569 . If this PR is ready and expected to be landed shortly, would you directly move `matplotlib` from `test` set into main dependency set? If no, we can create a separate high-prior PR to fix CI.

ck-intel:
> From the error trace it is likely that `matplotlib` is introduced in the `transformers` upgrade introduced in #21569 . If this PR is ready and expected to be landed shortly, would you directly move `matplotlib` from `test` set into main dependency set? If no, we can create a separate high-prior PR to fix CI.

@ZailiWang I guess it's better to raise a separate PR for the ci fix, this PR can wait for that. Thanks
Pls let me know if you want me to raise the fix PR for CI.

PR #12772, opened 2025-11-06
[hotfix]: Resolve ModuleNotFoundError in PD deployment for is_in_ci()

hzh0425:
## Motivation

**Issue**: When deploying with PD disaggregation, the `is_in_ci()` function in `sglang/utils.py` tried to import from `sglang.test.test_utils`, which doesn't exist in production installations, causing `ModuleNotFoundError`.

**Root Cause**: The test module is not included in production packages, but `server_args.py` calls `is_in_ci()` during initialization when `dp_size > 1`.

**Fix**: Reimplemented `is_in_ci()` to directly check the `SGLANG_IS_IN_CI` environment variable without importing from test modules:

```python
def is_in_ci() -> bool:
    import os
    return os.environ.get("SGLANG_IS_IN_CI", "").lower() in ("true", "1")
```
## Checklist

### DIFF
--- a/python/sglang/utils.py
+++ b/python/sglang/utils.py
@@ -356,10 +356,10 @@ def download_and_cache_file(url: str, filename: Optional[str] = None):
     return filename
 
 
-def is_in_ci():
-    from sglang.test.test_utils import is_in_ci
+def is_in_ci() -> bool:
+    import os
 
-    return is_in_ci()
+    return os.environ.get("SGLANG_IS_IN_CI", "").lower() in ("true", "1")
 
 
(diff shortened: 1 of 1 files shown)


hzh0425:
> When will `is_in_ci` be called? It shouldn't be called under production evn

https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/server_args.py#L1691
It will be invoked at this location, and when the user is in PD's separated deployment mode, it will directly throw a "module not found" error.  @Fridge003

Fridge003 (approved):

PR #12774, opened 2025-11-06
[PD / HiCache]fix deocde kvcache offload manager memory leak

huangtingwei9988:
## Motivation
The `_release_finished_req` method in the decode_kvcache_offload_manager is missing a release of `protected_size`, causing `protected_size` to only increase and never decrease, accumulating to a huge value by the end of the load test.
```
protected_size=5123398
```
```
ValueError: token_to_kv_pool_allocator memory leak detected! self.max_total_num_tokens=381504, available_size=381504, evictable_size=0, protected_size=5123398
```
## Checklist


hzh0425:
Pls Hold on this pr，will make some changes

huangtingwei9988:
> Pls Hold on this pr，will make some changes

Okay

huangtingwei9988:
cc @xiezhq-hermann

python/sglang/srt/mem_cache/chunk_cache.py
```diff
@@ -51,6 +51,16 @@ def match_prefix(self, **unused_kwargs) -> MatchResult:
             last_host_node=None,
         )
 
+    def release_decode_offload_finished_req(self, req: Req, prefill_offloaded_len: int):
```
hzh0425: Maybe we could hotfix this bug first and refactor it later?  @xiezhq-hermann @cctry 
not sure if it's appropriate to introduce this interface into the chunk cache.

cctry: This does not work with spec v2. We can add an exception cc @hnyls2002

hzh0425 (approved):
Can we apply this fix for now and refactor it later?
Let's get the bug fixed first.

hzh0425 (approved):
For now, we've temporarily handled it in the offload manager instead of adding a new interface in the chunk cache. We can refactor this part in a future refactoring effort.

Let's get the bug fixed first？

python/sglang/srt/disaggregation/decode_kvcache_offload_manager.py
hnyls2002: Please fix the `FIXME` as well

hnyls2002 (changes requested):

hnyls2002 (approved):

PR #12798, opened 2025-11-07
Support capturing aux_hidden_states for minimax m2.

pyc96:
## Motivation

Support capturing aux_hidden_states for Minimax M2 for EAGLE3 model training.

## Checklist


python/sglang/srt/mem_cache/common.py
```diff
@@ -352,7 +353,7 @@ def alloc_for_extend(
     req_pool_indices_device = req_pool_indices_cpu.to(batch.device, non_blocking=True)
 
     # Allocate KV cache (throws exception on failure)
-    if batch.tree_cache.page_size == 1:
+    if batch.tree_cache and batch.tree_cache.page_size == 1:
```
zhyncs: qq in which case `batch.tree_cache` will be None

pyc96: specforge currently sets it to None here https://github.com/sgl-project/SpecForge/blob/ba6159cc62cbabec7efd2846af4a1eabe2be92f5/scripts/prepare_hidden_states.py#L143.

This seems to be a bug on specforge side. Replacing None with a dummy tree cache works for me locally. Reverted this line

zhyncs (approved):

PR #12788, opened 2025-11-06
[DeepSeek-V3.2][NSA] Enable MHA Pathway for Short Sequence Prefill on B200 (SM100)

YAMY1234:
## Motivation

Follow up this PR: [DeepSeek-V3.2: Add Adaptive MHA Attention Pathway for Short-Sequence Prefill](https://github.com/sgl-project/sglang/pull/11892). Enable and optimize MHA on B200 (SM100) in the NSA backend by TRT-LLM ragged attention.

## Modifications

- **Add TRT-LLM ragged attention for SM100**: Integrate `flashinfer.prefill.trtllm_ragged_attention_deepseek` for Blackwell (B200) architecture to provide better accuracy than FA4
- **Remove chunked prefix cache support**: Align NSA backend with PR #11892 by removing chunked KV cache logic, keeping only MHA_ONE_SHOT mode for standard multi-head attention

Key changes in `nsa_backend.py`:
- Simplified `_forward_standard_mha()` to handle both SM90 (FA3) and SM100 (TRT-LLM) in a single function
- Removed `_forward_trtllm_ragged_mha()` helper function for cleaner code structure
- Added device capability check to conditionally allocate workspace buffer

## Accuracy Tests

```
Repeat: 8, mean: 0.792                                           | 24/198 [17:23<1:38:01, 33.80s/it]
Scores: ['0.783', '0.788', '0.783', '0.803', '0.823', '0.788', '0.773', '0.793']
====================
Writing report to /tmp/gpqa_deepseek-ai_DeepSeek-V3.2-Exp.html
{'chars': np.float64(15007.823232323231), 'chars:std': np.float64(11964.504345200807), 'score:std': np.float64(0.40520665017240837), 'score': np.float64(0.7929292929292929)}
Writing results to /tmp/gpqa_deepseek-ai_DeepSeek-V3.2-Exp.json
Total latency: 1104.677 s
Score: 0.793
```

## Benchmarking and Profiling

```
python -m sglang.launch_server \
  --model deepseek-ai/DeepSeek-V3.2-Exp \
  --tp 8 --dp 8 --enable-dp-attention \
  --kv-cache-dtype bf16
```

```
python3 -m sglang.bench_serving     --backend sglang     --dataset-name random     --num-prompts 100     --random-input-len 2000     --random-output-len 1     --disable-ignore-eos --random-range-ratio 1.0
```

This PR: MHA(trtllm_ragged_attention_deepseek) + skip topk
```
============ Serving Benchmark Result ============
[...]

PR #12816, opened 2025-11-07
[Deepseek V3.2] Only skip Indexer logits computation when is_extend_without_speculative

hlu1:
## Motivation

Fixed a bug in https://github.com/sgl-project/sglang/pull/12788

## Modifications

We should only skip Indexer logits computation when forward_mode.is_extend_without_speculative() returns true, meaning that no cuda graph is involved. The `is_extend_without_speculative` was only added to the MHA path, but not the MLA path. So when MLA is used (on B200 for example) and mtp is on,, the bug is triggered.

## Accuracy Tests
on 8xB200
20-shot gsm8k with mtp on:
Before fix: Accuracy: 0.008
After fix: Accuracy: 0.947

## Checklist

### DIFF
--- a/python/sglang/srt/layers/attention/nsa/nsa_indexer.py
+++ b/python/sglang/srt/layers/attention/nsa/nsa_indexer.py
@@ -410,6 +410,8 @@ class Indexer(CustomOp):
         metadata: BaseIndexerMetadata,
         return_indices: bool = True,
     ) -> Optional[torch.Tensor]:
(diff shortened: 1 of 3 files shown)


Fridge003 (approved):

Fridge003:
dpsk 3.2 test passed: https://github.com/sgl-project/sglang/actions/runs/19160112724/job/54777801302?pr=12816

PR #12817, opened 2025-11-07
[PD] Add custom gpu id to device topo support

stmatengss:
## Motivation

Currently, PD doesn't support indicating specified IB devices for different GPU id (tp/dp rank).
This PR supports this feature by introducing a new json format, and it consists of a mapping between gpu id and IB devices.

Usage
```
# Support both methods.
--disaggregation-ib-device ib0,ib1,ib2
--disaggregation-ib-device {0: "ib0, ib1", 1: "ib2, ib3", 2: "ib4"}
```

## Checklist


python/sglang/srt/disaggregation/mooncake/transfer_engine.py
```diff
@@ -6,6 +7,62 @@
...
+            if gpu_id in gpu_mapping:
+                return gpu_mapping[gpu_id]
+            else:
+                raise ValueError(
+                    f"No IB devices configured for GPU {gpu_id}. Available GPUs: {list(gpu_mapping.keys())}"
+                )
+
+    except json.JSONDecodeError:
+        # Not JSON format, treat as old format - return same devices for all GPUs
+        return ib_device_str if ib_device_str else None
```
ShangmingCai:
```suggestion
        return ib_device_str
```

ShangmingCai (approved):
LGTM. But we definitely should fix auto-discovery.

nit: `if ib_device_str is None has already` been checked before the `try` block, so it is impossible to be None here, we can just return `ib_device_str`.

hnyls2002 (approved):

PR #12820, opened 2025-11-07
[WIP][Feature] support tp-sp on qwen2/3 & deepseek v2/3/3.2

randgun:
## Motivation
For the classic dense decode layer structure (self-attention + MLP), in the pure TP case, tensors are parallelized in the attention layer and the MLP layer. Since each device contains the full amount of data before and after each TP split, Layernorm stores 2\*2BSH bytes of activation value (assume hidden states shape is [B, S, H]). SP aims to split this 4\*BSH redundant data across multiple devices, and do layernorm independently. For more details please refer to the research paper https://arxiv.org/pdf/2205.05198.pdf

屏幕截图 2025-11-10 114901

We add a parameter "--enable-sp" at server args, which can enable sequence parallel if be set. For example
```
python3 -m sglang.launch_server --model-path $MODEL_PATH \
        --tp-size 16 --dp-size 1 --enable-sp \
        --trust-remote-code --attention-backend ascend --device npu --host $HOST_IP --port $PORT \
        --quantization w8a8_int8 --mem-fraction-static 0.8 \
        --chunked-prefill-size 16000 --context-length 16000 --max-prefill-tokens 16000 --max-total-tokens 16000 \
        --disable-radix-cache --moe-a2a-backend deepep --deepep-mode auto
```
The TP-SP has two benifits:
1. Reduce long dataset (36K) TTFT for 10% on deepseek v3 and 7% on deepseek v3.2.
2. Reduce peak memory because of less activations on RMSNorm layer.

## Modifications
1. For the RowParallel linear, we replace all-reduce comm op with reduce-scatter on TP group if enable sp.
3. For the dense layers of model (qwen2/3 & deepseek v2/3/3.2), we split residual before layernorm at the first layer, hidden states is be splitted at RowParallel linear so do not need to split again.
4. After the first dense layer, all data are scattered state, only do layernorm at prepare_attn and prepare_mlp. 
5. Before the MLP and ATTENTION module, we do extra all-gather because of weights has been splitted on tensor dimension, to make sure inputs are complete.
6. Do all-gather at the last dense layer.
7. For sparse layer of deepseek models, when enableed Deepep, the hidden states has been sscattered, we utilize it and move the *_scattered_to_tp_attn_full* from prepare_attention to after getting the q_lora and latent cache, which also can decrease much more computation.

NOTE: 
1. We only adapt qwen2/3 and deepseek v2/3/3.2 on ascend backend, for other backends you can add few code to adapt SP.
2. TP-SP can enable with CP (https://github.com/sgl-project/sglang/pull/12207) together on Ascend backend. We test it on dsv3.2 model, CP=16, TP=2, which can continue reducing TTFT for 1 percent and decrease peak runtime memory (2G for single device).

## Accuracy Tests
The test_gsam8k.py is based on benchmark/gsm8k/bench_sglang.py, you can test it with
`python3 benchmark/gsm8k/bench_sglang.py --host http://127.0.0.1 --port 8000 --num-questions 300`.
image

## Benchmarking and Profiling
```
export cann_path=/usr/local/Ascend/ascend-toolkit/latest
source /usr/local/Ascend/driver/bin/setenv.bash
source ${cann_path}/../set_env.sh
source ${cann_path}/../../nnal/atb/set_env.sh
source ${cann_path}/opp/vendors/customize/bin/set_env.bash
export ASCEND_HOME_PATH=${cann_path}

# CPU high preformance
echo performance | tee /sys/devices/system/cpu/cpu*/cpufreq/scaling_governor
sysctl -w vm.swappiness=0
sysctl -w kernel.numa_balancing=0
sysctl -w kernel.sched_migration_cost_ns=50000

export SGLANG_SET_CPU_AFFINITY=1

# Memory Fragmentation
export PYTORCH_NPU_ALLOC_CONF=expandable_segments:True
export STREAMS_PER_DEVICE=32

# HCCL
export SGLANG_DEEPEP_NUM_MAX_DISPATCH_TOKENS_PER_RANK=16
export HCCL_BUFFSIZE=1600
#export HCCL_RDMA_PCIE_DIRECT_POST_NOSTRICT=TRUE
export HCCL_OP_EXPANSION_MODE=AIV
export HCCL_ALGO="level0:NA;level1:ring"

# Your NIC
export HCCL_SOCKET_IFNAME=enp48s3u1u1
export GLOO_SOCKET_IFNAME=enp48s3u1u1

export PYTHONPATH=$PWD/python/:$PYTHONPATH

python -m sglang.launch_server --model-path $MODEL_PATH \
        --tp-size 16 --dp-size 1 --enable-sp \
        --trust-remote-code --attention-backend ascend --device npu --host 127.0.0.1 --port 8000 \
        --quantization w8a8_int8 --mem-fraction-static 0.79 \
        --chunked-prefill-size 36000 --context-length 36000 --max-prefill-tokens 36000 --max-total-tokens 36000 \
        --disable-radix-cache --moe-a2a-backend deepep --deepep-mode auto
```
image

## Checklist

### DIFF
--- a/python/sglang/srt/layers/communicator.py
+++ b/python/sglang/srt/layers/communicator.py
@@ -22,6 +22,7 @@ import torch
 from sglang.srt.distributed import (
     get_tensor_model_parallel_world_size,
     get_tp_group,
+    sp_tensor_model_parallel_all_gather,
     tensor_model_parallel_all_reduce,
 )
 from sglang.srt.distributed.device_communicators.pynccl_allocator import (
@@ -96,6 +97,7 @@ class _LayerModeComputationContext:
     layer_id: int
     is_layer_sparse: bool
     is_previous_layer_sparse: Optional[bool]
+    enable_sp: Optional[bool] = False
 
     def previous_layer(self):
         assert self.is_previous_layer_sparse is not None
@@ -104,6 +106,7 @@ class _LayerModeComputationContext:
             is_layer_sparse=self.is_previous_layer_sparse,
             is_previous_layer_sparse=None,
             num_layers=self.num_layers,
+            enable_sp=self.enable_sp,
         )
 
 
@@ -115,23 +118,47 @@ class LayerScatterModes:
     mlp_mode: ScatterMode
     middle_residual_mode: ScatterMode
     layer_output_mode: ScatterMode
+    is_layer_sparse: bool
 
     @classmethod
     def init_new(cls, **kwargs):
         context = _LayerModeComputationContext(**kwargs)
+        is_layer_sparse = context.is_layer_sparse
+        mlp_mode = cls._compute_mlp_mode(context)
         return cls(
             layer_input_mode=cls._compute_layer_input_mode(context),
-            attn_mode=ScatterMode.TP_ATTN_FULL,
-            mlp_mode=cls._compute_mlp_mode(context),
-            middle_residual_mode=cls._compute_middle_residual_mode(context),
-            layer_output_mode=cls._compute_layer_output_mode(context),
+            attn_mode=cls._compute_attn_input_mode(context, mlp_mode),
+            mlp_mode=mlp_mode,
+            middle_residual_mode=cls._compute_middle_residual_mode(mlp_mode),
+            layer_output_mode=cls._compute_layer_output_mode(context, mlp_mode),
+            is_layer_sparse=is_layer_sparse,
         )
 
     @classmethod
     def _compute_layer_input_mode(cls, context: _LayerModeComputationContext):
         if context.layer_id == 0:
             return ScatterMode.model_input_output()
-        return cls._compute_layer_output_mode(context.previous_layer())
+        mlp_mode_previous = cls._compute_previous_mlp_mode(context)
+        return cls._compute_layer_output_mode(
+            context.previous_layer(), mlp_mode_previous
+        )
+
+    @classmethod
+    def _compute_attn_input_mode(
+        cls,
+        context: _LayerModeComputationContext,
+        mlp_mode: ScatterMode,
+    ):
+        if not context.enable_sp or context.is_layer_sparse:
+            return ScatterMode.TP_ATTN_FULL
+
+        if context.layer_id == 0:
+            return ScatterMode.model_input_output()
+        if mlp_mode == ScatterMode.SCATTERED:
+            return ScatterMode.SCATTERED
+        if mlp_mode == ScatterMode.FULL:
+            return ScatterMode.TP_ATTN_FULL
+        raise NotImplementedError
 
     @classmethod
     def _compute_mlp_mode(cls, context: _LayerModeComputationContext):
@@ -145,6 +172,9 @@ class LayerScatterModes:
                 )
                 else ScatterMode.FULL
             )
+
+        if context.enable_sp and not context.is_layer_sparse:
+            return ScatterMode.SCATTERED
         else:
             return (
                 ScatterMode.SCATTERED
@@ -153,8 +183,18 @@ class LayerScatterModes:
             )
 
     @classmethod
-    def _compute_middle_residual_mode(cls, context: _LayerModeComputationContext):
-        mlp_mode = cls._compute_mlp_mode(context)
+    def _compute_previous_mlp_mode(cls, context: _LayerModeComputationContext):
+        previous_context = context.previous_layer()
+        if (
+            context.enable_sp
+            and context.is_layer_sparse
+            and not previous_context.is_layer_sparse
+        ):
+            return ScatterMode.FULL
+        return cls._compute_mlp_mode(previous_context)
+
+    @classmethod
+    def _compute_middle_residual_mode(cls, mlp_mode: ScatterMode):
         if mlp_mode == ScatterMode.SCATTERED:
             return ScatterMode.SCATTERED
         if mlp_mode == ScatterMode.FULL:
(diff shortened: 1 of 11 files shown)


python/sglang/srt/layers/communicator.py
```diff
@@ -115,26 +115,64 @@ class LayerScatterModes:
...
     @classmethod
     def init_new(cls, **kwargs):
+        if "enable_sp" in kwargs:
+            enable_sp = kwargs["enable_sp"]
+            del kwargs["enable_sp"]
+        else:
+            enable_sp = False
+
         context = _LayerModeComputationContext(**kwargs)
+        enable_dense_sp = enable_sp and not context.is_layer_sparse
```
randgun: move enable_sp to context

python/sglang/srt/layers/linear.py
```diff
@@ -421,9 +425,18 @@ def weight_loader_v2(self, param: Parameter, loaded_weight: torch.Tensor):
             # However, we should fix this and avoid the branching here.
             param.load_column_parallel_weight(loaded_weight)
 
-    def forward(self, input_):
+    def forward(self, input_, allgather_input: bool = False):
         bias = self.bias if not self.skip_bias_add else None
 
+        if allgather_input:
```
randgun: remove communication to communicator

python/sglang/srt/layers/linear.py
```diff
@@ -1378,11 +1382,18 @@ def forward(self, input_, skip_all_reduce=False):
...
+            else:
+                output = output_parallel
         else:
-            output = output_parallel
+            dim_size = list(output_parallel.size())
+            dim_size[0] = dim_size[0] // self.tp_size
+            output = torch.empty(
+                dim_size, dtype=output_parallel.dtype, device=output_parallel.device
+            )
+            self.tp_group.reduce_scatter_tensor(output, output_parallel.contiguous())
```
iforgetmyname, line 1396: Just wondering why we have to do reduce_scatter here in linear, communication can be handled in layer communicator

randgun: If we do not use reduce_scatter in linear, we have to set "skip_all_reduce=enable_sp" and do reduce_scatter at both attention o_proj and mlp row parallel linear. And this code is inevitable for other models that need to adapt SP. This parameter is similar to "skip_all_reduce", maybe change name to "use_reduce_scatter" is better?

python/sglang/srt/server_args.py
```diff
@@ -419,6 +419,9 @@ class ServerArgs:
     mamba_ssm_dtype: str = "float32"
     mamba_full_memory_ratio: float = 0.9
 
+    # Sequence parallelism
+    enable_sp: bool = False
```
merrymercy, line 423: move to under `# Runtime options`

merrymercy (commented):
1. need a review from @ch-wan 
2. Think about how to avoid inserting communication primitives into model forward code and make them more reusable for more models
3. Add a GPU test case

python/sglang/srt/layers/communicator.py
```diff
@@ -271,17 +183,28 @@ def _compute_mlp_mode(cls, context: _LayerModeComputationContext):
...
         if mlp_mode == ScatterMode.SCATTERED:
             return ScatterMode.SCATTERED
         if mlp_mode == ScatterMode.FULL:
             return ScatterMode.TP_ATTN_FULL
         raise NotImplementedError
 
     @classmethod
-    def _compute_layer_output_mode(cls, context: _LayerModeComputationContext):
-        mlp_mode = cls._compute_mlp_mode(context)
+    def _compute_layer_output_mode(
```
ch-wan, line 205: Why do we need to add `mlp_mode` to all mode propagation? Passing `context` is enough. Probably `mlp_mode` can be a `@property` function of ctx.

python/sglang/srt/models/qwen2.py
```diff
@@ -89,13 +90,13 @@ def __init__(
             )
         self.act_fn = SiluAndMul()
 
-    def forward(self, x):
-        if get_global_server_args().rl_on_policy_target is not None:
+    def forward(self, x, enable_sp: bool = False):
```
ch-wan, line 93: We can define a util function `is_sp_layernorm_enabled` to avoid passing this arg to multiple functions. Or we can check `get_global_server_args().enable_sp` in `linear.py`. You can refer to how we implemented `is_dp_attention_enabled`).

python/sglang/srt/server_args.py
```diff
@@ -3164,6 +2979,13 @@ def add_cli_args(parser: argparse.ArgumentParser):
             help="The ratio of mamba state memory to full kv cache memory.",
         )
 
+        # Sequence parallelism
+        parser.add_argument(
+            "--enable-sp",
```
ch-wan, line 2984: I recommend to rename it as `--enable-sp-layernorm` for clarity.

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -2370,9 +2186,9 @@ def _forward_raw(
 
         # For MLP sync
         if forward_batch.global_num_tokens_cpu is not None:
-            forward_batch.prepare_mlp_sync_batch(self)
-        else:
-            forward_batch.prepare_attn_tp_scatter_input(self)
+            forward_batch.prepare_mlp_sync_batch(
+                self, get_global_server_args().enable_sp
```
ch-wan, line 2190: We can get this server arg internally.

ch-wan (commented):
I had a quick review. Overall, I feel that the code quality can be much improved. Many changes can be simplified.

merrymercy (changes requested):
Let us hold this until we fix all code quality issues

Issue #12783, opened 2025-11-06
[Bug] ImportError: cannot import name 'specialize_impl' from 'triton.runtime.jit' (/sgl-workspace/triton/python/triton/runtime/jit.py)


gongwei-130:
### Checklist

- [ ] 1. I have searched related issues but cannot get the expected help.
- [ ] 2. The bug has not been fixed in the latest version.
- [ ] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [ ] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [ ] 5. Please use English, otherwise it will be closed.

### Describe the bug

Looks like cu13 would install newer torch, which use newer api "specialize_impl" from triton, but sglang limit it to older commit.

```
[rank2]:W1106 19:01:04.408000 228 torch/_higher_order_ops/triton_kernel_wrap.py:948] [3/0] Encountered an exception in identify_mutated_tensors, assuming every input is mutated
[rank2]:W1106 19:01:04.408000 228 torch/_higher_order_ops/triton_kernel_wrap.py:948] [3/0] Traceback (most recent call last):
[rank2]:W1106 19:01:04.408000 228 torch/_higher_order_ops/triton_kernel_wrap.py:948] [3/0]   File "/usr/local/lib/python3.12/dist-packages/torch/_higher_order_ops/triton_kernel_wrap.py", line 924, in identify_mutated_tensors
[rank2]:W1106 19:01:04.408000 228 torch/_higher_order_ops/triton_kernel_wrap.py:948] [3/0]     ttir_module, ordered_tensor_names = generate_ttir(
[rank2]:W1106 19:01:04.408000 228 torch/_higher_order_ops/triton_kernel_wrap.py:948] [3/0]                                         ^^^^^^^^^^^^^^
[rank2]:W1106 19:01:04.408000 228 torch/_higher_order_ops/triton_kernel_wrap.py:948] [3/0]   File "/usr/local/lib/python3.12/dist-packages/torch/_higher_order_ops/triton_kernel_wrap.py", line 419, in generate_ttir
[...]

PR #12847, opened 2025-11-07
fix gb200 image issue

gongwei-130:
## Motivation

see issue here: https://github.com/sgl-project/sglang/issues/12783
Need to sgl kernel build as well.

## Checklist


Fridge003:
@gongwei-130 Is it possible to upgrade triton to 3.5.1 (maybe there is a pytorch wheel)

gongwei-130:
> @gongwei-130 Is it possible to upgrade triton to 3.5.1 (maybe there is a pytorch wheel)
@Fridge003 is 3.5.1 out? I don't find anything at here. https://github.com/triton-lang/triton/releases

gongwei-130:
only see this https://github.com/triton-lang/triton/issues/8630, don't see a ETA yet.

gongwei-130:
In addition, I am not seeing how upgrade triton would solve the issue unless triton is going to address their breaking changes? Do you know they are going to do this in 3.5.1?

Let me summary the issue: Triton make a breaking change -> torch make a change to accommodate that change -> sglang installed 2.9 for cu13 which doesn't have this change, torch nightly 2.10 has this change. So we should upgrade torch instead of triton? @Fridge003

Fridge003:
Also what do you mean by needing sgl-kernel build. Do we need to build sgl-kernel on the basis of torch 2.10?

Fridge003:
I feel upgrading to 2.10 might be inappropriate since torch 2.10 nightly is unstable. And we want to unify the torch version to 2.9 for cu129/cu130 in the next few days. For GB200 you can run with cu129+arm docker image which should work well.

Also what do you mean by needing sgl-kernel build. Do we need to build sgl-kernel on the basis of torch 2.10?

ishandhanani (commented):
I'm a bit hesitant about upgrading to an unstable torch release across the board. I do think it is worth upgrading to 2.9 everywhere and adding manual patches for now where needed until 2.10 is officially released

PR #12860, opened 2025-11-07
[router] Support complex assistant and tool messages in /chat/completions

hellodanylo:
## Motivation

The OpenAI schema for a request to /chat/completions allows input message content from an assistant or tool (part of history) to be a union of a few different types, but SGL Router currently requires a simple string. This prevents some OpenAI-compatible clients from using SGL Router when there are tool calls in the message history.

For example, `"content": "abc"` and `"content": [{"type": "text", "content": "abc"}]` are valid, but the latter is currently rejected by SGL Router as invalid JSON schema (error 400).

Here are the references to the relevant type definitions in OpenAI SDK:

https://github.com/openai/openai-python/blob/main/src/openai/types/chat/chat_completion_assistant_message_param.py#L46
```
    content: Union[str, Iterable[ContentArrayOfContentPart], None]
```

https://github.com/openai/openai-python/blob/main/src/openai/types/chat/chat_completion_tool_message_param.py#L14
```
    content: Required[Union[str, Iterable[ChatCompletionContentPartTextParam]]]
```

## Modifications

This PR modifies the type definitions of input body of /chat/completions in Rust to allow JSON messages that contain complex types as described above. The existing code already has a correct struct (UserMessageContent) for messages from user. This PR reuses this struct for tool and assistant messages. This PR also renames UserMessageContent to MessageContent to reflect that this struct is used in several different message types.

In several places, the router code has been updated to convert a complex message into a string as needed by the router. The backend servers (i.e. sglang, vllm, etc) are assumed to support whichever structure the router client has provided, so the router does not modify the structure when passing the request to the backend.

## Accuracy Tests

These changes do not affect accuracy. Existing text coverage has been updated to account for the updated types.

## Benchmarking and Profiling

These changes do not affect speed.

## Checklist

- [X] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
- [X] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.ai/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.ai/developer_guide/contribution_guide.html#benchmark-the-speed).


CatherineSue (commented):
Overall LGTM.

cc @key4ng for another look.

I vaguely remember we have similar issues in Responses protocols. I might be wrong. In general, if we do, we should have a more systematic approach.

sgl-router/src/protocols/chat.rs
CatherineSue: Do we need to do the same for system message?

sgl-router/src/protocols/chat.rs
```diff
@@ -317,12 +335,12 @@ fn validate_messages(messages: &[ChatMessage]) -> Result<(), validator::Validati
     for msg in messages.iter() {
         if let ChatMessage::User { content, .. } = msg {
```
key4ng, line 339: seems it only does validation on user message now, could we let it validate assistant and tool call messages as well

sgl-router/src/protocols/chat.rs
hellodanylo: Updated in the next commit.

sgl-router/src/protocols/chat.rs
hellodanylo: I can't comment on the validity of such checks due to insufficient knowledge of this API surface. Perhaps this can be addressed in another PR.

hellodanylo: I can't comment on the validity of such checks due to insufficient knowledge of this API surface. For example, in the assistant message, OpenAI SDK allows both null and zero-length lists. Perhaps this can be addressed in another PR by someone else.

hellodanylo (commented):
Pushed a commit to add MessageContent for system messages, as requested.

slin1237 (approved):

CatherineSue (approved):

Issue #12293, opened 2025-10-28
[Bug] DeepEP + nvfp4 issues with PD disagg or agg


kaixih:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Describe the bug

We’re working on PD disaggregation following the [instructions](https://github.com/sgl-project/sglang/issues/10903) using the main branch (or `sglang:dev` container). However, we’ve encountered several issues along the way.

We’d like to use this issue page to track the remaining problems in `main` and clarify the best command setup for PD disagg NVFP4, since many flags seem to have changed, been deprecated, or become redundant over the past few months.

Plan:

- [ ] PD agg works with DeepEP (BF16 dispatch so no new DeepEP is required); this is mainly a sanity check
- [ ] PD disagg works with DeepEP (BF16 dispatch); also a sanity check
- [ ] PD disagg works with DeepEP (NVFP4 dispatch) - Functionality
- [ ] PD disagg works with DeepEP (NVFP4 dispatch) - Perf analysys
- [ ] PD disagg fp4 best practice

# PD Agg

In parallel, we also tried reverting to the PD aggregation workloads we originally used for development, but we’re observing other issues such as:

```
[2025-10-28 21:14:27] DataParallelController hit an exception: Traceback (most recent call last):
  File "/scratch/repo/sglang/python/sglang/srt/managers/data_parallel_controller.py", line 496, in run_data_parallel_controller_process
    controller = DataParallelController(server_args, port_args)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/scratch/repo/sglang/python/sglang/srt/managers/data_parallel_controller.py", line 151, in __init__
    self.launch_dp_attention_schedulers(server_args, port_args)
  File "/scratch/repo/sglang/python/sglang/srt/managers/data_parallel_controller.py", line 344, in launch_dp_attention_schedulers
    self.launch_tensor_parallel_group(
[...]

PR #12866, opened 2025-11-08
[Fix] Fix nan error for large scale ep

Fridge003:
## Motivation
Part of https://github.com/sgl-project/sglang/issues/12293/

This bug is introduced in https://github.com/sgl-project/sglang/pull/10874, which will wrongly remove some of the redundant experts from `logical_to_all_physical_map` (e.g., should be [0, 256] for logical expert 0, but wrongly set to [0, -1])

This only happens on the first and the last node. On these two nodes the values of `w13_input_scale` will become nan
```bash
[2025-11-08 02:52:56 DP2 TP2 EP2] w13_input_scale values: tensor([nan, nan, nan, nan, nan, nan], device='cuda:2', dtype=torch.float32)
[2025-11-08 02:52:56 DP1 TP1 EP1] w13_input_scale values: tensor([nan, nan, nan, nan, nan, nan], device='cuda:1', dtype=torch.float32)
```
Later it hits errors like 
```bash
  File "/sgl-workspace/sglang/python/sglang/srt/layers/quantization/modelopt_quant.py", line 1460, in process_weights_after_loading
    assert torch.all(w13_input_scale == w13_input_scale[0])
```

After this fix it can be solved.

Reproduction: 
GB200, latest main branch
https://gist.github.com/kaixih/32bdc4fec4feabe9305d1acb2e1f96db

## Modifications

Fix this by keeping the redundant experts on these nodes.

## Checklist


python/sglang/srt/eplb/expert_location.py
```diff
@@ -355,10 +355,14 @@ def _compute_logical_to_all_physical_map(
...
-                        nearest_expert
-                    ]
+                mapped_physical_experts = logical_to_all_physical_map[layer_id][
+                    logical_expert_id
+                ]
+                if (
+                    nearest_expert != -1
+                    and nearest_expert not in mapped_physical_experts
+                ):
+                    mapped_physical_experts[0] = nearest_expert
```
kaixih, line 365: why replace the first expert with the nearest?

Fridge003: No specific reason. I'm checking this with author of the breaking PR.
It's at least better than the original logic

kaixih: Sure. Just curious, the original logic seemed to replace all mappings with the nearest expert that is not -1, but the new logic seems to replace only the first mapping.

Fridge003: The original logic expelled some needed experts, causing this bug

acelyc111: @Fridge003 what does "some needed experts" mean? Is the first expert enough?

Fridge003: For example, on EP rank 0, [0, 256] is initialized. But the prior logic will change it to [0, -1], then expert 256 is missing.

acelyc111: If [0, 256] is initialized, it's possible to be changed to [256,256], will it still cause the issue?

Fridge003: This change will not cause this issue, since 256 is already in there?
Or do you mean we need to check whether the replaced id is unique?

Fridge003:
https://github.com/sgl-project/sglang/actions/runs/19259460525?pr=12866

PR #12909, opened 2025-11-09
[model-gateway]: add qwen coder tool parser support xml format for qwen3 coder and microthinker

whybeyoung:
## Summary

Add a new Qwen Coder XML format tool parser for the Rust Router (sgl-model-gateway).

Qwen Coder models use a distinct XML-based tool call format that differs from the standard Qwen JSON format:

```xml
<tool_call>
<function=execute_bash>
<parameter=command>pwd && ls</parameter>
</function>
</tool_call>
```

This PR implements full support for parsing this format in both complete and streaming modes.

### Motivation

- Qwen Coder models (Qwen3-Coder, Qwen2.5-Coder) use XML format for tool calls
- The existing qwen parser only handles JSON format inside <tool_call> tags
- Users needed to use Python-side parsing, losing Rust Router performance benefits

Reference: https://huggingface.co/Qwen/Qwen3-Coder-480B-A35B-Instruct-FP8?chat_template=default

#### New Parser: QwenCoderParser

- Location: src/tool_parser/parsers/qwen_coder.rs
- Format: <tool_call><function=name><parameter=key>value</parameter></function></tool_call>
- Features:
  - HTML entity decoding (&amp; → &, &lt; → <, etc.)
  - Python literal support (True, False, None)
  - Streaming support with complete parameter emission
  - Unicode support in function names and parameters

#### Model Auto-Detection

Added model mappings in factory.rs:
registry.map_model("Qwen/Qwen3-Coder*", "qwen_coder");
registry.map_model("Qwen3-Coder*", "qwen_coder");
registry.map_model("qwen3-coder*", "qwen_coder");
registry.map_model("Qwen/Qwen2.5-Coder*", "qwen_coder");
registry.map_model("Qwen2.5-Coder*", "qwen_coder");
registry.map_model("qwen2.5-coder*", "qwen_coder");

Users can also explicitly use `--tool-call-parser qwen_coder.`

#### Test categories:
- Basic parsing (single/multiple tools)
- Streaming (chunked input, partial tags)
- Type conversion (JSON, Python literals)
- HTML entity decoding
- Unicode support
- Edge cases (malformed XML, large values, nested JSON)

#### Test Plan
- All 49 unit and integration tests pass
- Clippy passes with no warnings
- Code follows existing parser patterns
- E2E test with actual Qwen Coder model (requires model availability)

### DIFF
--- a/sgl-model-gateway/src/tool_parser/parsers/qwen_coder.rs
+++ b/sgl-model-gateway/src/tool_parser/parsers/qwen_coder.rs
@@ -1,0 +1,587 @@
+use async_trait::async_trait;
+use regex::Regex;
+use serde_json::Value;
+
+use crate::{
+    protocols::common::Tool,
+    tool_parser::{
+        errors::{ParserError, ParserResult},
+        parsers::helpers,
+        traits::ToolParser,
+        types::{FunctionCall, StreamingParseResult, ToolCall, ToolCallItem},
+    },
+};
+
+/// Qwen Coder format parser for tool calls
+///
+/// Handles the Qwen Coder specific XML format:
+/// `<tool_call>\n<function=name>\n<parameter=key>value</parameter>\n</function>\n</tool_call>`
+///
+/// Features:
+/// - Tool Call Tags: `<tool_call>` and `</tool_call>` wrap each individual call
+/// - XML-style function declaration: `<function=name>`
+/// - XML-style parameters: `<parameter=key>value</parameter>`
+///
+/// Reference: https://huggingface.co/Qwen/Qwen3-Coder-480B-A35B-Instruct-FP8?chat_template=default
+pub struct QwenCoderParser {
+    /// Regex for extracting tool calls in parse_complete
+    extractor: Regex,
+
+    /// Buffer for accumulating incomplete patterns across chunks
+    buffer: String,
+
+    /// Stores complete tool call info (name and arguments) for each tool being parsed
+    prev_tool_call_arr: Vec<Value>,
+
+    /// Index of currently streaming tool call (-1 means no active tool)
+    current_tool_id: i32,
+
+    /// Flag for whether current tool's name has been sent to client
+    current_tool_name_sent: bool,
+
+    /// Tracks raw JSON string content streamed to client for each tool's arguments
+    streamed_args_for_tool: Vec<String>,
+
+    /// Token configuration
+    tool_call_start_token: &'static str,
+    tool_call_end_token: &'static str,
+
+    /// XML format streaming state
+    in_tool_call: bool,
+    current_function_name: String,
+    current_parameters: serde_json::Map<String, Value>,
+
+    /// Precompiled regex patterns for XML format parsing
+    xml_function_pattern: Regex,
+    xml_param_pattern: Regex,
+}
+
+/// Decode HTML entities in a string (equivalent to Python's html.unescape)
+///
+/// Handles common HTML entities like &amp; &lt; &gt; &quot; &#39; and numeric entities
+fn html_unescape(s: &str) -> String {
+    let mut result = String::with_capacity(s.len());
+    let mut chars = s.chars().peekable();
+
+    while let Some(c) = chars.next() {
+        if c == '&' {
+            let mut entity = String::new();
+            let mut consumed_semicolon = false;
+            while let Some(&next) = chars.peek() {
+                if next == ';' {
+                    chars.next();
+                    consumed_semicolon = true;
+                    break;
+                }
+                if next.is_alphanumeric() || next == '#' {
+                    entity.push(chars.next().unwrap());
+                } else {
+                    break;
+                }
+            }
+
+            let decoded = match entity.as_str() {
+                "amp" => "&",
+                "lt" => "<",
+                "gt" => ">",
+                "quot" => "\"",
+                "apos" => "'",
+                "nbsp" => "\u{00A0}",
+                s if s.starts_with('#') => {
+                    let num_str = &s[1..];
+                    let code_point = if num_str.starts_with('x') || num_str.starts_with('X') {
+                        u32::from_str_radix(&num_str[1..], 16).ok()
+                    } else {
+                        num_str.parse::<u32>().ok()
+                    };
+                    if let Some(cp) = code_point {
+                        if let Some(ch) = char::from_u32(cp) {
+                            result.push(ch);
+                            continue;
+                        }
+                    }
+                    // Invalid numeric entity, reconstruct original
+                    result.push('&');
+                    result.push_str(&entity);
+                    if consumed_semicolon {
+                        result.push(';');
+                    }
+                    continue;
+                }
+                _ => {
+                    // Unknown entity, reconstruct original
+                    result.push('&');
+                    result.push_str(&entity);
+                    if consumed_semicolon {
+                        result.push(';');
+                    }
+                    continue;
+                }
+            };
+            result.push_str(decoded);
+        } else {
+            result.push(c);
+        }
+    }
+
+    result
+}
+
+/// Parse a raw parameter value, similar to Python's _safe_val
+///
+/// 1. Decode HTML entities
+/// 2. Try to parse as JSON (numbers, booleans, null, objects, arrays)
+/// 3. Fall back to string if JSON parsing fails
+fn safe_val(raw: &str) -> Value {
+    let unescaped = html_unescape(raw.trim());
+
+    // Try JSON parsing first
+    if let Ok(v) = serde_json::from_str::<Value>(&unescaped) {
+        return v;
+    }
+
+    // Handle Python-style literals (True, False, None)
+    match unescaped.as_str() {
+        "True" => return Value::Bool(true),
+        "False" => return Value::Bool(false),
+        "None" => return Value::Null,
+        _ => {}
+    }
+
+    // Fall back to string
+    Value::String(unescaped)
+}
+
+impl QwenCoderParser {
+    /// Create a new Qwen Coder parser
+    pub fn new() -> Self {
+        // Support XML format: <tool_call>\n<function=name>\n<parameter=key>value</parameter>\n</function>\n</tool_call>
+        let pattern = r"(?s)<tool_call>\s*(.*?)\s*</tool_call>";
+        let extractor = Regex::new(pattern).expect("Valid regex pattern");
+
+        // Precompile XML format regex patterns for performance
+        let xml_function_pattern =
+            Regex::new(r"<function=([^>]+)>").expect("Valid XML function pattern");
+        let xml_param_pattern = Regex::new(r"(?s)<parameter=([^>]+)>(.*?)</parameter>")
+            .expect("Valid XML parameter pattern");
+
+        Self {
+            extractor,
+            buffer: String::new(),
+            prev_tool_call_arr: Vec::new(),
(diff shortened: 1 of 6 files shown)


CatherineSue:
Why don't we have a qwen3_coder_parser.rs separately?

sgl-router/src/tool_parser/parsers/qwen.rs
CatherineSue, line 19: Can we update this part?

sgl-router/src/tool_parser/parsers/qwen.rs
```diff
@@ -182,26 +339,76 @@ impl ToolParser for QwenParser {
         // Build tool indices
         let tool_indices = helpers::get_tool_indices(tools);
 
-        // Determine start index for JSON parsing
-        let start_idx = if let Some(pos) = current_text.find(self.individual_tool_start_token) {
-            pos + self.individual_tool_start_token.len()
-        } else if self.current_tool_id > 0 && current_text.starts_with(self.tool_call_separator) {
-            self.tool_call_separator.len()
+        // Detect format: only check once, then remember it
+        // Both JSON and XML formats use the same <tool_call> tags, difference is internal content
```
CatherineSue: If we update the section on L19, we can skip the detailed format docs here. Just keep `// Detect format: only check once, then remember it`

sgl-router/src/tool_parser/parsers/qwen.rs
```diff
@@ -254,5 +461,283 @@ impl ToolParser for QwenParser {
...
+}
+
+impl QwenParser {
+    /// Parse XML format tool calls incrementally (similar to Python Qwen3CoderDetector)
+    fn parse_xml_incremental(
+        &mut self,
+        current_text: &str,
+        tool_indices: &std::collections::HashMap<String, usize>,
+    ) -> ParserResult<StreamingParseResult> {
+        use crate::tool_parser::types::ToolCallItem;
```
CatherineSue: nit: put this at the top of the file

sgl-router/src/tool_parser/parsers/qwen.rs
```diff
@@ -254,5 +461,283 @@ impl ToolParser for QwenParser {
...
+            self.buffer.clear();
+            return Ok(StreamingParseResult { normal_text, calls });
+        }
+
+        // Look for tool call start
+        if !self.in_xml_tool_call {
+            if let Some(s) = current_text.find("<tool_call>") {
+                normal_text.push_str(&current_text[..s]);
+                self.buffer = current_text[s + "<tool_call>".len()..].to_string();
+                self.in_xml_tool_call = true;
```
CatherineSue: We memerize this once, right?

CatherineSue (commented):
I understand there might be a lot duplications between qwen.rs and qwen-coder. However, the current architecture is a big if-else to me at the moment.

The logic that parses xml in complete or incremental style looks good to me. But my biggest concern is the fall back logic or how do we manage the is_xml_format or is_json_format state thru incremental parsing. I am concerned with future debug and maintenance. 

From what we know, qwen3-coder or qwen3-xml-format is specific to qwen3-coder models. The other qwen family models use the json format. It might be clear if we separate qwen-xml.rs from qwen.rs. And in the tool parser registry, we simply add a qwen3-coder match for qwen-xml.

Let me know what you think @whybeyoung @slin1237 

If we want to keep them in the same file, then let's add through tests, as detailed as possible. We should add tests nevertheless. But if we are gonna keep them in the same file, more test cases should be covered.

sgl-router/src/tool_parser/parsers/qwen_coder.rs
```diff
@@ -0,0 +1,598 @@
...
+    fn parse_xml_incremental(
+        &mut self,
+        current_text: &str,
+        tool_indices: &std::collections::HashMap<String, usize>,
+    ) -> ParserResult<StreamingParseResult> {
+        let mut normal_text = String::new();
+        let mut calls: Vec<ToolCallItem> = vec![];
+
+        // If we're not in a tool call and don't see a start token, return normal text
+        if !self.in_xml_tool_call && !current_text.contains("<tool_call>") {
```
CatherineSue: nit: is this duplicated with https://github.com/sgl-project/sglang/pull/12909/files#diff-e90dbd8ca14fa1755b80d9f3314f2a14eb15fbdf352d27901e0dbbf7398e9353R544-R559?

sgl-router/src/tool_parser/parsers/qwen_coder.rs
```diff
@@ -0,0 +1,598 @@
...
+                        self.xml_current_function_name = function_name.clone();
+                        self.current_tool_name_sent = true;
+
+                        // Initialize tool call tracking
+                        if self.current_tool_id == -1 {
+                            self.current_tool_id = 0;
+                        }
+
+                        // Ensure tracking arrays are large enough
+                        while self.prev_tool_call_arr.len() <= self.current_tool_id as usize {
```
CatherineSue: Can we replace this with `helpers::ensure_capacity`?

sgl-router/src/tool_parser/parsers/qwen_coder.rs
```diff
@@ -0,0 +1,598 @@
...
+
+                        // Store tool call info
+                        let mut tool_obj = serde_json::Map::new();
+                        tool_obj.insert("name".to_string(), Value::String(function_name.clone()));
+                        tool_obj.insert(
+                            "arguments".to_string(),
+                            Value::Object(serde_json::Map::new()),
+                        );
+                        self.prev_tool_call_arr[self.current_tool_id as usize] =
+                            Value::Object(tool_obj);
```
CatherineSue:
```suggestion
                        self.prev_tool_call_arr[self.current_tool_id as usize] = json!({
      "name": function_name,
      "arguments": {}
  });
```

sgl-model-gateway/src/tool_parser/parsers/qwen_coder.rs
```diff
@@ -0,0 +1,598 @@
...
+                        self.current_parameter_value.clear();
+                        self.in_parameter = true;
+
+                        // Remove the opening tag from buffer
+                        if let Some(m) = cap.get(0) {
+                            self.buffer = self.buffer[m.end()..].to_string();
+                        }
+                    }
+                }
+            }
```
CatherineSue: Why do aggresively remove the open tag `<parameter` here? We don't need L333 to L437, no?

The accumulation there is no different than the fast path on L257 will be able to capture this eventually.

Unless we are streaming parameters incrementally, I don't see why we need this.

sgl-router/src/tool_parser/parsers/qwen_coder.rs
```diff
@@ -0,0 +1,598 @@
...
+                } else {
+                    // Parameter value is incomplete, accumulate it
+                    // Check if there's any content before a potential partial closing tag
+                    if let Some(partial_end) = self.buffer.find("</") {
+                        // There might be a partial closing tag, only take content before it
+                        self.current_parameter_value
+                            .push_str(&self.buffer[..partial_end]);
+                        self.buffer = self.buffer[partial_end..].to_string();
+                    } else {
+                        // No closing tag yet, accumulate all content
```
CatherineSue: do we plan to send the incremental params here? like:

```
      calls.push(ToolCallItem {
          tool_index: self.current_tool_id as usize,
          name: None,
          parameters: self.buffer.clone(),  // It will be more complicated than this
      });
      self.current_parameter_value.push_str(&self.buffer);
      self.buffer.clear();
```

sgl-model-gateway/src/tool_parser/parsers/qwen_coder.rs
```diff
@@ -0,0 +1,598 @@
...
+                            tool_index: self.current_tool_id as usize,
+                            name: None,
+                            parameters: json_fragment.clone(),
+                        });
+
+                        // Update streamed args
+                        let current_args =
+                            &mut self.streamed_args_for_tool[self.current_tool_id as usize];
+                        if current_args.is_empty() {
+                            *current_args = format!("{{\"{}\": {}}}", key, value_json);
```
CatherineSue: I am not following this part. Why are we wrapping double `{{` here?

CatherineSue: I don't see any tests covered this case. Can we add such a test case?

```
let chunks = vec![
          "<tool_call>",
          r#"<function=get_weather>"#,
          r#"<parameter=city>Paris</parameter>"#,      // ← Chunk 2: First param
          r#"<parameter=units>metric</parameter>"#,    // ← Chunk 3: Second param
          "</function></tool_call>",
      ];
```

What are the parsed each chunk? What is the final result for this? 
Do we have `{"city": "Paris"}, "units": "metric"`? or `{"city": "Paris", "units": "metric"}`? 

Or is this block trying to deal with more complicated nested json objects? Should we add test for that?

sgl-router/src/tool_parser/parsers/qwen_coder.rs
```diff
@@ -0,0 +1,598 @@
...
+///
+/// Handles the Qwen Coder specific XML format:
+/// `<tool_call>\n<function=name>\n<parameter=key>value</parameter>\n</function>\n</tool_call>`
+///
+/// Features:
+/// - Tool Call Tags: `<tool_call>` and `</tool_call>` wrap each individual call
+/// - XML-style function declaration: `<function=name>`
+/// - XML-style parameters: `<parameter=key>value</parameter>`
+///
+/// Reference: Python implementation in qwen3_coder_detector.py
```
CatherineSue: let's use `https://huggingface.co/Qwen/Qwen3-Coder-480B-A35B-Instruct-FP8?chat_template=default`

sgl-model-gateway/src/tool_parser/parsers/qwen_coder.rs
whybeyoung: > I am not following this part. Why are we wrapping double `{{` here?

I use format! macro instead of serde_json::json! here because:
// 1. Streaming requirement: We need to send incremental JSON fragments to the client
//    as parameters arrive, not wait for the complete object to serialize
// 2. Performance: Direct string concatenation is more efficient than serializing
//    the entire object each time a new parameter arrives
// 3. Incremental updates: We only send the new parameter fragment, not the
//    entire arguments object, which reduces bandwidth and improves responsiveness
// The value_json is already properly serialized by serde_json::to_string above,
// so we just need to wrap it in the JSON object structure manually.

sgl-model-gateway/src/tool_parser/parsers/qwen_coder.rs
slin1237: addressed

sgl-model-gateway/src/tool_parser/parsers/qwen_coder.rs
```diff
@@ -0,0 +1,587 @@
...
+            &mut self.prev_tool_call_arr,
+            &mut self.current_tool_id,
+            &mut self.current_tool_name_sent,
+            &mut self.streamed_args_for_tool,
+        );
+        self.reset_streaming_state();
+    }
+}
+
+#[cfg(test)]
```
CatherineSue, line 511: should we move the tests here to `tool_parser_qwen_coder.rs`?

CatherineSue (commented):
LGTM

CatherineSue (approved):

PR #12938, opened 2025-11-10
fix health endpoint for prod

Tushar-ml:
## Motivation

Production environments running on kubernetes or similar services, ping server every x seconds, to check if server is healthy or not, but in current implementation, every time health endpoint is hit, a fake token is send to server, which is creating load at high concurrency

## Modifications

Seperated the health endpoint from health generate

## Checklist

### DIFF
--- a/python/sglang/srt/entrypoints/http_server.py
+++ b/python/sglang/srt/entrypoints/http_server.py
@@ -403,6 +403,13 @@ async def validate_json_request(raw_request: Request):
 
 
 @app.get("/health")
+async def health() -> Response:
+    """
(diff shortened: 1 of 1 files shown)


jinmingyi1998:
that's wrong.

we need to run once generate to check if a engine is healthy.

a server is unhealthy if it crash or hang up.

you can turn `health prob interval` if you think it influence performance too much

slin1237:
engine can actually intelligently know how to handle health request based on load
Do you see problems where this takes forever to return?

slin1237 (changes requested):
need more comments on this

PR #8778, opened 2025-08-04
model: support dots.vlm1 model

yonghenglh6:
## Checklist

- [ ] Format your code according to the [Code Formatting with Pre-Commit](https://docs.sglang.ai/references/contribution_guide.html#code-formatting-with-pre-commit).
- [ ] Add unit tests as outlined in the [Running Unit Tests](https://docs.sglang.ai/references/contribution_guide.html#running-unit-tests-adding-to-ci).
- [ ] Update documentation / docstrings / example tutorials as needed, according to [Writing Documentation](https://docs.sglang.ai/references/contribution_guide.html#writing-documentation-running-docs-ci).
- [ ] Provide throughput / latency benchmark results and accuracy evaluation results as needed, according to [Benchmark and Profiling](https://docs.sglang.ai/references/benchmark_and_profiling.html) and [Accuracy Results](https://docs.sglang.ai/references/accuracy_evaluation.html).
- [ ] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co-author when merging the PR.
- [ ] Please feel free to join our Slack channel at https://slack.sglang.ai to discuss your PR.

PR #12953, opened 2025-11-10
[Model] Add PaddleOCR-VL Model Support

yudian0504:
## Motivation

Add PaddleOCR-VL support (without paddle doc layout paser), the core implementation is based on the code from the official image (      ccr-2vdh3abv-pub.cnc.bj.baidubce.com/paddlepaddle/paddleocr-genai-vllm-server:latest  ).

Launch:
SGLANG_VLM_CACHE_SIZE_MB=2048 python -m sglang.launch_server --model-path /home/admin/PaddleOCR-VL/ --host 0.0.0.0 --port 8188 --trust-remote-code --tp-size 1  --max-running-requests 8 --mem-fraction-static 0.65  --attention-backend flashinfer --mm-attention-backend fa3 

Usage:
`
curl 'http://0:8188/v1/chat/completions' --header 'Content-Type: application/json' --data '{
        "model": "PaddleOCR-VL-0.9B",
        "messages": [
            {
                "role": "user",
                "content": [
	                {"type": "image_url", "image_url": {"url": "https://paddle-model-ecology.bj.bcebos.com/paddlex/imgs/demo_image/paddleocr_vl_demo.png"}},
    	            {"type": "text", "text": "OCR:"}
                ]
            }
        ],
        "temperature":0.0,
        "max_tokens":1000,
        "stream": false
    }'
`

More Details:
https://huggingface.co/PaddlePaddle/PaddleOCR-VL, replace vl_rec_server_url to your url.

## Checklist


albaNnaksqr:
Hi! I'm also interested in this model and have a couple of questions:
1. Do we need to register the model config in both **python/sglang/srt/hf_transformers_utils.py** and **python/sglang/srt/configs/__init__.py**? 

2. Is it neccessary to add template in **python/sglang/srt/parser/conversation.py**?
When I was working on support for another OCR model earlier, I found [clarification](https://github.com/sgl-project/sglang/pull/8778#discussion_r2258418003) like this, which seems to suggest that we don’t need to register an additional template in the code.

yudian0504:
> Hi! I'm also interested in this model and have a couple of questions:
> 
> 1. Do we need to register the model config in both **python/sglang/srt/hf_transformers_utils.py** and **python/sglang/srt/configs/init.py**?
> 2. Is it necessary to add template in **python/sglang/srt/parser/conversation.py**?
>    When I was working on support for another OCR model earlier, I found [clarification](https://github.com/sgl-project/sglang/pull/8778#discussion_r2258418003) like this, which seems to suggest that we don’t need to register an additional template in the code.

1. I understand this is not necessary (AutoConfig + trust_remote_code is sufficient), because we don't have explicit dependencies on the config type.
2. Thank you for the reminder. I just noticed that the official team has recently updated the chat-template.jinja file (https://huggingface.co/PaddlePaddle/PaddleOCR-VL/blob/main/chat_template.jinja). My code is based on an older version of the model, but it seems that the new chat_template.jinja still has some minor issues. Therefore, adding a template can help ensure compatibility with both the old and new versions.

python/sglang/srt/models/paddleocr_vl.py
```diff
@@ -0,0 +1,818 @@
...
+
+
+from collections.abc import Iterable
+from typing import List, Optional, Set, Tuple, Union
+
+import numpy as np
+import torch
+import torch.nn as nn
+from einops import rearrange
+from sgl_kernel.flash_attn import flash_attn_varlen_func
```
mingfeima: applying something like `maybe_get_vit_flash_attn_backend` in vllm : https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/paddleocr_vl.py#L616-L622 should be a better idea than directly calling `flash_attn_varlen_func` from sgl_kernel.

Lzhang-hub:
@yudian0504 I tested the pr with your command, the result seems incorrect.
img_v3_02s6_4fa2b0e0-c4d7-4974-b4e4-9a6fbfb04a4g

python/sglang/srt/models/paddleocr_vl.py
yudian0504: modified as you suggested

yhyang201:
Could you rebase on main? This will fix the errors in the CI, and then it should be ready to merge.

yudian0504:
> Could you rebase on main? This will fix the errors in the CI, and then it should be ready to merge.

done

yuan-luo (approved):

FlamingoPg (approved):

yhyang201 (approved):

python/sglang/srt/multimodal/processors/paddleocr_vlm.py
```diff
@@ -0,0 +1,299 @@
...
+        return {
+            "mm_items": mm_items,
+            "input_ids": input_ids.tolist(),
+            "im_token_id": self.mm_tokens.image_token_id,
+            "mrope_positions": mrope_positions,
+            "mrope_position_delta": mrope_position_delta,
+        }
+
+    @staticmethod
+    def get_rope_index(
```
mickqian: any difference with `MRotaryEmbedding.get_rope_index`?

python/sglang/srt/multimodal/processors/paddleocr_vlm.py
```diff
@@ -0,0 +1,299 @@
...
+    )
+    image = image.resize((resized_width, resized_height))
+    return image
+
+
+async def resize_image_async(image, min_pixels, max_pixels, factor):
+    return resize_image(image, min_pixels, max_pixels, factor)
+
+
+class PaddleOCRVLImageProcessor(BaseMultimodalProcessor):
```
mickqian: is it possible to reuse the Processor of Qwen?

python/sglang/srt/multimodal/processors/paddleocr_vlm.py
yudian0504: Okay, I'll confirm that now.

python/sglang/srt/multimodal/processors/paddleocr_vlm.py
yudian0504: I compared the implementations of both, and your suggestion is correct. The Qwen-VL processor can be used to replace most of PaddleOCR's implementation.

PR #12964, opened 2025-11-10
[DeepseekV3.2] Deepseek fp8 support for MHA path

YAMY1234:
## Motivation

Enable FP8 KV cache support for MHA (Multi-Head Attention) path in DeepSeek NSA backend. Previously, MHA could only be used when KV cache dtype is BF16, limiting its usage with FP8 quantized models. This PR implements on-demand dequantization for FP8 KV cache before MHA computation, leveraging the existing paged dequantization kernel from #11655.

## Modifications

1. **Conditional generation of `page_table_1_flattened`** (`nsa_backend.py`)
   - Generate flattened page table only when needed: RAGGED mode OR (FP8 KV cache AND sequence length ≤ topk threshold)
   - Avoids unnecessary computation for BF16 or long sequences that won't use MHA

2. **FP8 dequantization for MHA prefix cache** (`deepseek_v2.py`)
   - In `forward_normal_prepare`: dequantize FP8 prefix KV cache before projection when `mha_one_shot=True`
   - Reuses existing `dequantize_k_cache_paged` kernel from #11655
   - Ensures `.contiguous()` memory layout for downstream operations

3. **Enable FP8 for MHA path selection** (`deepseek_v2.py`)
   - Updated `handle_attention_nsa` to accept `torch.float8_e4m3fn` dtype
   - Allows MHA_ONE_SHOT to be selected for short sequences with FP8 KV cache

## Accuracy Tests

```

Repeat: 8, mean: 0.792
Scores: ['0.768', '0.783', '0.788', '0.783', '0.798', '0.813', '0.813', '0.788']
====================
{'chars': np.float64(15032.565656565657), 'chars:std': np.float64(11701.19130485268), 'score:std': np.float64(0.4088102291888498), 'score': np.float64(0.7878787878787878)}
Total latency: 880.245 s
Score: 0.788
```

## Benchmarking and Profiling

```bash
  --model deepseek-ai/DeepSeek-V3.2-Exp \
  --tp 8 --dp 8 \
  --enable-dp-attention \
  --disable-cuda-graph \
  --disable-radix-cache \
  --max-running-requests 32
```

Benchmark command:

```bash
  --backend sglang \
  --dataset-name random \
  --num-prompts 100 \
  --random-input-len {1024|2048|512} \
  --disable-ignore-eos \
  --random-range-ratio 1.0
```

Tested short sequence scenarios where prefill len with 51, 1024, and 2048:

| Input Len | Compared     | Processed Req/s ↑                  | Input Tok Tput/s ↑                    | Mean E2E Latency (ms) ↓        | Mean TTFT (ms) ↓               | Benchmark Duration (s) ↓ |
| --------- | ----------- | ------------------------ | -------------------------------- | ------------------------------ | ------------------------------ | ------------------------ |
| **512**   | **Main**    | 32.96                    | 16,873.63                        | 1,656.43                       | 1,632.66                       | 3.03                     |
|           | **With PR** | **38.88 (+18%)**         | **19,906.23 (+18%)**             | **1,491.79 (-10%)**            | **1,473.11 (-10%)**            | **2.57 (-15%)**          |
| **1024**  | **Main**    | 27.09                    | 27,744.39                        | 2,093.07                       | 2,066.57                       | 3.69                     |
|           | **With PR** | **21.14 → 27.09 (+28%)** | **21,647.94 → 27,744.39 (+28%)** | **2,697.31 → 2,093.07 (-22%)** | **2,663.44 → 2,066.57 (-22%)** | **4.73 → 3.69 (-22%)**   |
| **2048**  | **Main**    | 12.75                    | 25,500.38                        | 4,421.62                       | 4,269.49                       | 7.84                     |
|           | **With PR** | **16.86 (+32%)**         | **33,713.03 (+32%)**             | **3,340.15 (-24%)**            | **3,207.91 (-25%)**            | **5.93 (-24%)**          |

### Trace Profiling:
With this PR:
forward(2.2ms):
image
forward_prepare(1.8ms):
image
forward_core(300us):
image

TOT main:
forward(3ms):
image
forward_prepare(2.3ms):
image
forward_core(700us):
image

## Checklist


python/sglang/srt/models/deepseek_v2.py
```diff
@@ -1548,9 +1554,34 @@ def forward_normal_prepare(
             forward_batch.mha_one_shot
             and sum(forward_batch.extend_prefix_lens_cpu) != 0
         ):
-            kv_a, k_pe = self._get_mla_kv_buffer(
-                forward_batch.fetch_mha_one_shot_kv_indices(), q.dtype, forward_batch
-            )
+            # If FP8 with NSA, need to dequantize first (NSA uses DeepSeek-specific FP8 format)
+            if self.use_nsa and self.kv_cache_dtype == "fp8_e4m3":
+                # Reuse pre-computed page_table_1_flattened from nsa_backend metadata
+                kv_indices = (
```
Fridge003: Can we wrap these logics as a util function for dpsk v32?

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -1548,9 +1554,34 @@ def forward_normal_prepare(
...
+                    forward_batch.attn_backend.forward_metadata.page_table_1_flattened
+                )
+                assert (
+                    kv_indices is not None
+                ), "page_table_1_flattened should have been generated for FP8 MHA path"
+
+                kv_cache_fp8 = forward_batch.token_to_kv_pool.get_key_buffer(
+                    self.attn_mha.layer_id
+                )
+                kv_latent_bf16 = dequantize_k_cache_paged(
```
Fridge003: I remember that trtllm ragged kernel can take fp8 kv as input. Can you check this?

Fridge003: Also how will this operation effect performance?
Will it cause large drop in perforamance? We need some benchmark data to check this

YAMY1234: Hi @Fridge003 Having TRT-LLM ragged kernel directly take FP8 KV cache would not work in our situation. The key issue is that MLA stores compressed latent vectors (512-dim kv_a) that need to be decompressed through the kv_b_proj matrix multiplication to get the full K and V for attention.

This decompression step requires BF16 precision for the matrix math, and doing it in FP8 would introduce significant accumulated errors, especially in deeper layers. I actually ran some experiments on this before and saw max_diff values of 0.3-0.8 in layers 58-60, which is way too large for FP8 quantization to be acceptable.

I temporarily disabled the dequantize path and forced FP8. Here are the test results:
```
hostuser@c7f5d06e8bb3:/sgl-workspace/sglang$ python3 benchmark/gsm8k/bench_sglang.py --num-shots 8 --num-questions 100 --parallel 100
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 100/100 [02:35<00:00,  1.56s/it]
Accuracy: 0.690
Invalid: 0.280
Latency: 155.922 s
Output throughput: 140.064 token/s
```

YAMY1234: For performance comparison I just pasted them in the pr description, pretty promissing😄

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -1548,9 +1554,34 @@ def forward_normal_prepare(
...
+                    kv_cache_fp8, kv_indices
+                )  # [num_prefix_tokens, 1, 576]
+                kv_a = (
+                    kv_latent_bf16[:, :, : self.kv_lora_rank].squeeze(1).contiguous()
```
hlu1: `kv_latent_bf16` doesn't need to be contiguous. `self.kv_b_proj` should be able to take strided inputs.

python/sglang/srt/models/deepseek_v2.py
hlu1: There are two cases, 1) extend (kvcache reuse, chunked prefill), 2) pure prefill (no reuse, no chunk or first chunk). You only need to dequant from kvcache for the first case. In the second case, you can use the bf16 kv directly.

YAMY1234: @hlu1 Yes exactly. I think current logic handles correctly. If there is no prefix, i.e., if `sum(forward_batch.extend_prefix_lens_cpu) = 0` it will not go into the if branch and will not do dequantizing or cache reading.

In the new patchset after fixing Baizhou's first suggestion by adding a util function, the code will be easier to read

python/sglang/srt/models/deepseek_v2.py
YAMY1234: I added the `contiguous()` because I ran into an issue if I didn't add the `contiguous`😭
```

           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/sgl-workspace/sglang/python/sglang/srt/models/deepseek_v2.py", line 1398, in forward
    s = self.forward_prepare(
        ^^^^^^^^^^^^^^^^^^^^^
  File "/sgl-workspace/sglang/python/sglang/srt/models/deepseek_v2.py", line 1440, in forward_prepare
    inner_state = self.forward_normal_one_shot_prepare(
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/sgl-workspace/sglang/python/sglang/srt/models/deepseek_v2.py", line 2478, in forward_normal_one_shot_prepare
    return self.forward_normal_prepare(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/sgl-workspace/sglang/python/sglang/srt/models/deepseek_v2.py", line 1566, in forward_normal_prepare
    kv = self.kv_b_proj(kv_a)[0]
         ^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch/nn/modules/module.py", line 1773, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch/nn/modules/module.py", line 1784, in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/sgl-workspace/sglang/python/sglang/srt/layers/linear.py", line 429, in forward
    output_parallel = self.quant_method.apply(self, input_, bias)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/sgl-workspace/sglang/python/sglang/srt/layers/quantization/fp8.py", line 491, in apply
    return self.w8a8_block_fp8_linear(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/sgl-workspace/sglang/python/sglang/srt/layers/quantization/fp8_utils.py", line 242, in deepgemm_w8a8_block_fp8_linear_with_fallback
    q_input, x_scale = sglang_per_token_group_quant_fp8(
  File "/sgl-workspace/sglang/python/sglang/srt/layers/quantization/fp8_kernel.py", line 491, in sglang_per_token_group_quant_fp8
    assert x.is_contiguous(), "`x` is not contiguous"
           ^^^^^^^^^^^^^^^^^
AssertionError: `x` is not contiguous
```

hlu1: In this case, we should modify the cuda kernel so it can take strided inputs.

python/sglang/srt/layers/attention/nsa_backend.py
```diff
@@ -369,7 +369,14 @@ def init_forward_metadata(self, forward_batch: ForwardBatch):
                 ]
             )
 
-            if topk_transform_method == TopkTransformMethod.RAGGED:
+            # Generate page_table_1_flattened when needed:
+            mha_dequantize_needed = (
+                self.nsa_kv_cache_store_fp8 and max_seqlen_k <= self.nsa_index_topk
+            )
```
hlu1, line 461: With this change, you need to change the `deepseek_v2.py` and `nsa_backend.py` at once if you need to change the dispatch logics, which is more error prone.

YAMY1234: Make sense, I made the logics inside `deepseek_v2.py's handle_attention_nsa` and passed to `nsa_backend` via `forward_batch`

python/sglang/srt/models/deepseek_v2.py
YAMY1234: Thanks. Yeah I looked into it, and it seems we need modifying the kernel and all related wrappers to fully support strided inputs. Would it be okay if we temporarily keep the contiguous() call and add a TODO comment to revisit this in a follow up PR (e.g., make the kernel handle strided inputs properly)? Also from the profiling trace, it doesn’t seem to have noticeable performance impact at the moment.

hlu1: We can revisit this in the future

python/sglang/srt/layers/attention/nsa_backend.py
hlu1: It's not that simple. `init_forward_metadata` is called before `self.model.forward` which calls `handle_attention_nsa` when it gets to the attention part.

hlu1: Should we move all the dispatch logics to NativeSparseAttnBackend and we can do the all the dispatch in init_forward_metadata etc? The problem is that the dispatch logics are split across `NativeSparseAttnBackend` and `handle_attention_nsa` in `deepseek_v2.py`:

https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/layers/attention/nsa_backend.py#L1321

https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/models/deepseek_v2.py#L398

YAMY1234: After discussion we plan to do this in another seperate PR. We’ll prepare a follow-up PR to implement this refactor and keep this PR focused on fp8 support~

hlu1: Sounds good

hlu1 (approved):

YAMY1234:
@Fridge003 I am testing on B200. By default if the prefill seqlen is short than 2048 MHA path will be used. So by default B200 is using fp8 kv cache+MHA when possible right now but H200 is using bf16.

Fridge003 (approved):

PR #6380, opened 2025-05-18
Disable compiling arch below sm_90 in aarch64 by default

Qiaolin-Yu:
## Motivation
For aarch64, it doesn't need to use arch like sm_7x or sm_8x in most cases, so we disable it by default to make the compile more fast.

## Checklist

- [ ] Format your code according to the [Code Formatting with Pre-Commit](https://docs.sglang.ai/references/contribution_guide.html#code-formatting-with-pre-commit).
- [ ] Add unit tests as outlined in the [Running Unit Tests](https://docs.sglang.ai/references/contribution_guide.html#running-unit-tests-adding-to-ci).
- [ ] Update documentation / docstrings / example tutorials as needed, according to [Writing Documentation](https://docs.sglang.ai/references/contribution_guide.html#writing-documentation-running-docs-ci).
- [ ] Provide throughput / latency benchmark results and accuracy evaluation results as needed, according to [Benchmark and Profiling](https://docs.sglang.ai/references/benchmark_and_profiling.html) and [Accuracy Results](https://docs.sglang.ai/references/accuracy_evaluation.html).
- [ ] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co-author when merging the PR.
- [ ] Please feel free to join our Slack channel at https://slack.sglang.ai to discuss your PR.

PR #12967, opened 2025-11-10
Fix issue: ENABLE_BELOW_SM90 cannot be enabled on aarch64 CPU

MarcoDWei:
## Motivation

ENABLE_BELOW_SM90 is used to enabled for GPUs with a compute capability lower than sm_90.  It is disabled in aarch64 CPU by  https://github.com/sgl-project/sglang/pull/6380. The intention is to disable it by default to make the compile more fast. But this makes the option always disabled even if user enable it explicitly. 

`python -m uv build --wheel   -Cbuild-dir=build  -Ccmake.define.ENABLE_BELOW_SM90=ON -color=always --no-build-isolation`

The issue occurs on an aarch64 CPU with a 4090D GPU (SM89). There is an error report even if we have enabled "ENABLE_BELOW_SM90"

```
RuntimeError: RMSNorm failed with error code no kernel image is available for execution on the device
[2025-10-22 06:53:14] SIGQUIT received. signum=None, frame=None. It usually means one child failed.
Killed
```

## Modifications

Select proper DEFAULT_ENABLE_BELOW_SM90 via CPU aarch.  The default value does not affect the option explicitly configured by the user.

## Checklist


FlamingoPg:
@MarcoDWei Looks great. Could you share the error and the fix details?

MarcoDWei:
> @MarcoDWei Looks great. Could you share the error and the fix details?

It happens on aarch64 CPU + 4090D GPU(SM89). There is an error report even if we have enabled "ENABLE_BELOW_SM90" 

```
RuntimeError: RMSNorm failed with error code no kernel image is available for execution on the device
[2025-10-22 06:53:14] SIGQUIT received. signum=None, frame=None. It usually means one child failed.
Killed
```

After applying the patch, the issue is resolved.

FlamingoPg (approved):

PR #11456, opened 2025-10-11
Support aggregating engine metrics in sgl-router

fzyzcjy:
## Checklist

- [ ] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.ai/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.ai/developer_guide/contribution_guide.html#benchmark-the-speed).

## Summary by CodeRabbit

* **New Features**
  * Added an /engine_metrics HTTP endpoint that aggregates Prometheus-compatible engine metrics from all workers into a single response, including per-worker labels.
  * Returns a clear “Not Implemented” response for router types that don’t yet support this endpoint or unsupported multi-router configurations.

* **Tests**
  * Introduced comprehensive tests for metric aggregation, covering multiple metric types, overlapping counters, empty input, invalid metric skipping, and large real-world payloads with order-agnostic assertions.

PR #12970, opened 2025-11-10
Fix Grafana dashboard metrics prefix from sglang: to sglang_

### LINKED ISSUE #12618: [Bug] Grafana dashboard example is outdated due to /metrics API prefix change (sglang: -> sglang_)
### Checklist

### Describe the bug

In a recent version of sglang, the /metrics API changed the prefix for all its metrics.

The prefix sglang: was changed to sglang_ (the colon was replaced with an underscore).

However, the example Grafana dashboard provided in /examples/monitoring/grafana/ still uses the old sglang: prefix in all its queries.

This causes the Grafana dashboard to show "No Data" for all panels, as it is querying for metrics that no longer exist. The example dashboard configuration needs to be updated to use the new sglang_ prefix to work correctly.

### Reproduction

Run a recent version of the sglang server with the /metrics endpoint enabled.

Scrape the /metrics endpoint (e.g., using Prometheus).

Observe that all metrics from sglang now use the sglang_ prefix (e.g., sglang_request_duration_seconds).

Import and set up the example Grafana dashboard from /examples/monitoring/grafana/.

Observe that the dashboard shows "No Data" or errors because all its queries look for metrics starting with sglang: (e.g., sglang:request_duration_seconds).

To Fix:
All instances of "sglang:" in the JSON file(s) within /examples/monitoring/grafana/ need to be find-and-replaced with "sglang_".

### Environment

sglang version: Version 0.5.4.post1

Other components: Prometheus, Grafana (any version).

This bug is not strictly environment-dependent but is an issue with the example configuration file itself.

teddygood:
## Motivation

Fixes #12618 where the example Grafana dashboard displays "No Data" when using sgl-router's `/engine_metrics` endpoint.

After PR #11456 was merged, sgl-router introduced a "hacky workaround" that converts colon (:) characters to underscores (`_`) in metric names during aggregation. This causes a mismatch between:
- Actual metrics format: `sglang_*` (after conversion in sgl-router)
- Dashboard queries: `sglang:*` (old format)

This quick fix updates the Grafana dashboard to match the current behavior.

## Modifications

- Updated all Prometheus queries in examples/monitoring/grafana/dashboards/json/sglang-dashboard.json
- Changed metric prefix from `sglang:` to `sglang_`

## Accuracy Tests

None.

## Benchmarking and Profiling

None.

## Discussion

I think this is a temporary fix for dashboard compatibility. The root issue is that the current Python metrics code uses `sglang:` prefix, which violates Prometheus best practices according to the [official Prometheus documentation](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels):

> NOTE: Colons (':') are reserved for user-defined recording rules. They SHOULD NOT be used by exporters or direct instrumentation.

Please correct me if I'm wrong. I'd appreciate any feedback on this interpretation!

## Checklist


ishandhanani:
@teddygood - can you share any screenshots showing before and after behavior please? We can then go ahead and merge this in

PR #12979, opened 2025-11-10
bugfix: multi-model routing for /generate api

SYChen123:
## Motivation

Multi-model routing bugfix for the /generate api

## Modifications

1. obtain the "model" parameter from the request body and pass it to the subsequent functions.
2. refactor the using of model_id field in other apis such as v1/completions.

## Accuracy Tests

Register a qwen3-8b worker and a Qwen2.5-VL-7B-Instruct worker to the router and send requests to the router using corresponding model_id to specific the model.  For comparison, send the request directly to the worker and check if the result is exactly the same.
# /generate
image

For comparison, sending requests directly to the worker. 
image

The result is exactly the same, indicating the requests are sending to the correct model with setting model parameter in the request body.

Sending a request without model parameter to the router
image

# /v1/completions & v1/chat/completions
completions
v1-completions
chat
qwen3-8b-chat-completions
qwen-vl-chat-completions

# /v1/responses
v1-response

# /v1/embeddings
qwen-vl-embeddings

## Checklist


sgl-router/src/protocols/generate.rs
```diff
@@ -21,6 +21,8 @@ pub struct GenerateRequest {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub text: Option<String>,
 
+    pub model: String,
```
slin1237: this should be optional

slin1237 (changes requested):

sgl-router/src/protocols/generate.rs
SYChen123: done

sgl-router/src/routers/router_manager.rs
```diff
@@ -352,10 +352,10 @@ impl RouterTrait for RouterManager {
         body: &GenerateRequest,
         _model_id: Option<&str>,
```
slin1237: we should leverage this parameter
and this parameter should be passed in from server.rs

SYChen123: Shell we modify other apis to pass the model_id parameter from server.rs,  such as v1_chat_completions and v1_completions?

SYChen123: The following apis are modified based on your comments and I've updated my testing results in the above section. 
/generate, /v1/chat, /v1/completions, /v1/response, /v1/embedding, /v1/rerank.

CatherineSue: Maybe we just change it to model_id without _? _ usually means unused.

And could you take a look at the lint failure?

Thanks!

SYChen123: Done! Thanks for your comments.

slin1237 (approved):

CatherineSue (approved):

PR #12980, opened 2025-11-10
Remove duplicate import

LHXuuu:
## Motivation
Remove duplicate import

## Modifications

- Remove duplicate definition of WNA16_SUPPORTED_BITS

## Checklist

### DIFF
--- a/python/sglang/srt/layers/quantization/compressed_tensors/__init__.py
+++ b/python/sglang/srt/layers/quantization/compressed_tensors/__init__.py
@@ -1,7 +1,0 @@
-class scalar_types:
-    uint4b8 = "uint4b8"
-    uint8b128 = "uint8b128"
-
-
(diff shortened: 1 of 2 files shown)


python/sglang/srt/layers/quantization/compressed_tensors/__init__.py
```diff
@@ -1,7 +0,0 @@
-class scalar_types:
```
Fridge003, line 1: Please remove this file if it's not needed

Fridge003 (changes requested):

python/sglang/srt/layers/quantization/compressed_tensors/__init__.py
LHXuuu: Finished

hnyls2002 (approved):

PR #12984, opened 2025-11-10
vlm: fix tiny multimodal cache bug

yhyang201:
## Checklist

### DIFF
--- a/test/srt/models/test_vlm_models.py
+++ b/test/srt/models/test_vlm_models.py
@@ -23,9 +23,9 @@ if _is_hip:
     MODELS = [SimpleNamespace(model="openbmb/MiniCPM-V-2_6", mmmu_accuracy=0.4)]
 else:
     MODELS = [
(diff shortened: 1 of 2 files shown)


test/srt/models/test_vlm_models.py
```diff
@@ -269,50 +254,6 @@ def test_vlm_mmmu_benchmark(self):
         for model in models_to_test:
             self._run_vlm_mmmu_test(model, "./logs")
 
-    def test_vlm_mmmu_benchmark_with_small_cache(self):
```
mickqian, line 272: @ZhengWG Could you add it back later, since the cache class is changed in a previous PR (with a server-level cache)?

ZhengWG: > @ZhengWG Could you add it back later, since the cache class is changed in a previous PR (with a server-level cache)?

Sure, I will add it this week.

Issue #12432, opened 2025-10-31
[Bug] Memory Leak Check Fails in Qwen3-Next PD Disaggregation Due to Mamba Size Mismatch


xjx471258437:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Bug Description
The Qwen3-Next PD  disaggregation task can be successfully deployed. However, when running bench_serving, the prefill stage fails with the following error:

```shell
[2025-10-31 10:40:43 TP0] Scheduler hit an exception: Traceback (most recent call last):
  File "/sgl-workspace/sglang/python/sglang/srt/managers/scheduler.py", line 2797, in run_scheduler_process
    scheduler.event_loop_overlap_disagg_prefill()
  File "/usr/local/lib/python3.12/dist-packages/torch/utils/_contextlib.py", line 120, in decorate_context
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/sgl-workspace/sglang/python/sglang/srt/disaggregation/prefill.py", line 371, in event_loop_overlap_disagg_prefill
    self.self_check_during_idle()
  File "/sgl-workspace/sglang/python/sglang/srt/managers/scheduler_runtime_checker_mixin.py", line 214, in self_check_during_idle
    self.check_memory()
  File "/sgl-workspace/sglang/python/sglang/srt/managers/scheduler_runtime_checker_mixin.py", line 147, in check_memory
    raise ValueError(msg)
ValueError: token_to_kv_pool_allocator memory leak detected! full_available_size=1663695, full_evictable_size=1066, self.token_to_kv_pool_allocator.size=1664761, self.tree_cache.full_protected_size()=0
mamba_available_size=473, mamba_evictable_size=4, self.req_to_token_pool.mamba_pool.size=476, self.tree_cache.mamba_protected_size()=0
NotImplementedError
```

### Root Cause Analysis
[...]

PR #12994, opened 2025-11-10
[BugFix] Fix prefill memory leak in PD + GDN

ZeldaHuang:
## Motivation
See #12432 , duplicate free mamba pool. 

## Modifications
Don't free chunked req mamba cache in `get_next_disagg_prefill_batch_to_run`, same as [here](https://github.com/sgl-project/sglang/blob/afee2843d51f2d4254e941de8314eed3898629b9/python/sglang/srt/managers/scheduler.py#L1648) 

## Checklist


ShangmingCai (approved):
LGTM. But wonder if we can modify the Mamba cache free logic at the other place to unify the logic here for future maintenance. cc: @yizhang2077 @hzh0425 PTAL

PR #12998, opened 2025-11-10
Fuse routed_scaling_factor to fused_marlin_moe

ispobock:
## Motivation

Fuse routed_scaling_factor to fused_marlin_moe to reduce mul kernel launch.


ovowei:
@ispobock I have a question regarding this change. It seems that only **deepseek_v2.py** removes the outer`final_hidden_states *= self.routed_scaling_factor` when using CompressedTensorsWNA16MoEMethod. For other models using **Marlin MoE**, wouldn’t this lead to applying `self.routed_scaling_factor` **twice**?

Also, the reason I looked into this PR is that moving `routed_scaling_factor` into the Marlin MoE module causes the **CPU MoE path in ktransformers** to *not* apply `routed_scaling_factor` at all, which leads to noticeable accuracy degradation. I’m planning to fix for this.

PR #13047, opened 2025-11-11
[router][grpc] Support num_reasoning_tokens in haromy models

CatherineSue:
## Motivation

This PR tries to support `reasoning_tokens` in `completion_tokens_details` in OpenAI API. 

It addresses the issue in two approaches:

1. Harmony - Addressed directly in router through counting in parsing

### Harmony models

- Incrementally counting tokens while channel is analysis or commentary

## Accuracy Tests

gpt-oss-120b + grpc router
Screenshot 2025-11-18 at 2 57 32 PM

## Checklist

### DIFF
--- a/sgl-router/src/routers/grpc/harmony/parser.rs
+++ b/sgl-router/src/routers/grpc/harmony/parser.rs
@@ -23,6 +23,7 @@ fn get_harmony_encoding() -> &'static HarmonyEncoding {
 pub struct HarmonyParserAdapter {
     parser: StreamableParser,
     prev_recipient: Option<String>,
+    reasoning_token_count: u32,
 }
 
 impl HarmonyParserAdapter {
@@ -35,6 +36,7 @@ impl HarmonyParserAdapter {
         Ok(Self {
             parser,
             prev_recipient: None,
+            reasoning_token_count: 0,
         })
     }
 
@@ -241,11 +243,20 @@ impl HarmonyParserAdapter {
         finish_reason: String,
         matched_stop: Option<serde_json::Value>,
     ) -> Result<HarmonyChannelOutput, String> {
+        let mut reasoning_token_count = 0u32;
+
         // Feed all tokens to the parser
         for &token_id in output_ids {
             self.parser
                 .process(token_id)
                 .map_err(|e| format!("Failed to process token {}: {}", token_id, e))?;
+
+            // Count reasoning tokens (analysis + commentary channels)
+            if let Some(channel) = self.parser.current_channel() {
+                if channel == "analysis" || channel == "commentary" {
+                    reasoning_token_count += 1;
+                }
+            }
         }
 
         // Extract all completed messages from the parser
@@ -270,6 +281,7 @@ impl HarmonyParserAdapter {
             final_text,
             finish_reason: final_finish_reason,
             matched_stop,
+            reasoning_token_count,
         })
     }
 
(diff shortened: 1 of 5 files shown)


python/sglang/srt/managers/scheduler_output_processor_mixin.py
```diff
@@ -825,6 +826,20 @@ def stream_output_generation(
...
+                reasoning_count = 0
+                if (
+                    hasattr(self.tokenizer, "think_end_id")
+                    and self.tokenizer.think_end_id is not None
+                ):
+                    try:
+                        think_end_pos = output_ids_.index(self.tokenizer.think_end_id)
+                        reasoning_count = think_end_pos
+                    except ValueError:
+                        # think_end_id not found - reasoning may be disabled or not yet emitted
```
CatherineSue: This fallback can be improved by utilizing `req.reasoning` after #12551 is merged.

python/sglang/srt/managers/scheduler_output_processor_mixin.py
```diff
@@ -825,6 +826,20 @@ def stream_output_generation(
...
                 completion_tokens.append(len(output_ids_))
                 cached_tokens.append(req.cached_tokens)
+
+                reasoning_count = 0
+                if (
+                    hasattr(self.tokenizer, "think_end_id")
+                    and self.tokenizer.think_end_id is not None
+                ):
+                    try:
+                        think_end_pos = output_ids_.index(self.tokenizer.think_end_id)
```
hnyls2002: Is there a simpler way to get the reasoning tokens? This may introduce some indexing overhead. Even if the overhead can be overlapped by GPU execution, I am still worried about it.

CatherineSue: There is no other simple way to get it. If this causes overhead, we can just not support this in regular models yet.

CatherineSue: One other way to do it is retokenize the reasoning content. I feel it might cause more overhead? 

If both are not good. I think we can support harmony in grpc for now.

python/sglang/srt/managers/scheduler_output_processor_mixin.py
hnyls2002, line 325: Why can we not do the locating `think_end_id` logic here? @CatherineSue

CatherineSue: I am unfamiliar with the part. Could you elaborate your thoughts?

slin1237 (approved):

python/sglang/srt/managers/scheduler_output_processor_mixin.py
hnyls2002: Sure.

hnyls2002 (approved):

python/sglang/srt/managers/scheduler_output_processor_mixin.py
hnyls2002: https://github.com/sgl-project/sglang/issues/13250

CatherineSue: I tried an approach following your suggestions. Please checkout 3f0bff5

CatherineSue: Actually my attempt doesn't work. I realize the changes would be too many. I reverted everything but harmony grpc side.

PR #13056, opened 2025-11-11
[Router] use call_id instead of id for matching function calls in Responses API for Harmony

zhaowenzi:
## Motivation

Fix function call matching logic in sgl-router's Responses API implementation to comply with OpenAI specification.

Currently, the router incorrectly uses the `id` field to match `function_call` and `function_call_output` items. According to OpenAI's Responses API spec, `function_call_output` only requires `call_id` field (where `id` is optional), and matching should be done via `call_id`.

Refer: https://platform.openai.com/docs/guides/function-calling

```python
for item in response.output:
    if item.type == "function_call":
        if item.name == "get_horoscope":
            # 3. Execute the function logic for get_horoscope
            horoscope = get_horoscope(json.loads(item.arguments))
            
            # 4. Provide function call results to the model
            input_list.append({
                "type": "function_call_output",
                "call_id": item.call_id,
                "output": json.dumps({
                  "horoscope": horoscope
                })
            })
```

## Modifications

- Updated matching logic in `src/routers/grpc/harmony/responses.rs` to use `call_id` instead of `id`
- Updated matching logic in `src/routers/grpc/harmony/builder.rs` to use `call_id` instead of `id`

## Accuracy Tests

1. Before this fix
image

2. After this fix
image

3. Test with OpenAI
image

## Benchmarking and Profiling

N/A

## Checklist

### DIFF
--- a/sgl-router/src/routers/grpc/harmony/builder.rs
+++ b/sgl-router/src/routers/grpc/harmony/builder.rs
@@ -603,11 +603,11 @@ impl HarmonyBuilder {
                     .iter()
                     .rev()
                     .find_map(|item| match item {
-                        ResponseInputOutputItem::FunctionToolCall { id, name, .. }
-                            if id == call_id =>
-                        {
-                            Some(name.clone())
-                        }
+                        ResponseInputOutputItem::FunctionToolCall {
+                            call_id: item_call_id,
+                            name,
+                            ..
+                        } if item_call_id == call_id => Some(name.clone()),
                         _ => None,
                     })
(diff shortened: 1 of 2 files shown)


CatherineSue:
Since there is no ut added for this, please use Accuracy Tests section to upload the before vs after fix for this.

CatherineSue:
Can you fix regular/responses as well? Otherwise, can we update PR title to indicate this is for Harmony only?

key4ng (approved):

slin1237 (approved):

PR #10532, opened 2025-09-16
vlm: Refactor engine vlm params and support precessor output as input

zhaochenyang20:
# Summary

This PR refactors how the offline **Engine** handles multimodal / VLM inputs and exposes a consistent `image_data` API across models.

---

## Motivation

Previously, each VLM had its own ad-hoc way to pass images (raw pixel values, custom dicts, etc.), and multi-image requests could easily break the engine path.

We want a single, well-defined `image_data` contract that works for:

- raw images (e.g., `PIL`, `numpy`, `torch`)
- HuggingFace processor outputs
- precomputed visual embeddings

---

### Engine API

- Clarified and unified the supported formats for `image_data` in `Engine.generate` / `async_generate`, including:
  - **Plain images:**  
    `image_data=[image]` or `image_data=[[image1, image2, ...]]`
  - **HuggingFace processor outputs:**  
    `image_data=[dict(processor_output, format="processor_output")]`
  - **Precomputed embeddings:**  
    `image_data=[dict(processor_output, format="precomputed_embedding", feature=precomputed_embeddings)]`

- Centralized multimodal validation / normalization in the scheduler (`mm_utils`, `schedule_batch`)  
  so batched and multi-image requests follow the same path.

---

### Model / Processor Side

- Updated **Qwen2.5-VL** model wrapper and multimodal processors to use the new `MultimodalInputFormat` helper instead of model-specific dicts.  
- Fixed the engine multi-image bug reproduced in `refactor-engine-vlm-params` — the engine now correctly handles multiple images per request.

---

### Tests

Added `test_vlm_input_format.py`, which verifies that both **Qwen2.5-VL** and **Gemma-3-VLM** work with:

- direct `image_data=[PIL.Image]`
- processor output as `image_data`
- precomputed embeddings passed via `format="precomputed_embedding"`

It also checks that the model can correctly understand a **2-image input** (e.g., *taxi + SGL logo*) for all three code paths.

---

### Docs

Updated [`docs/advanced_features/vlm_query.ipynb`](docs/advanced_features/vlm_query.ipynb) to:

- describe the three supported calling patterns (Basic / Processor Output / Precomputed Embeddings)
- show concrete examples for **Qwen2.5-VL** using the offline `Engine` API

---

## Testing

```bash
pytest test/srt/test_vlm_input_format.py

✅ All 6 tests pass on Qwen2.5-VL-3B-Instruct and Gemma-3-4B-IT with multimodal enabled.

PR #13069, opened 2025-11-11
refactor: refactor engine vlm params

minleminzui:
## Motivation
#10532 

Mainly to ensure CI stability for [unit-test-backend-1-gpu (0)](https://github.com/sgl-project/sglang/actions/runs/19174599295/job/54860676356?pr=10532#logs),  
which runs `pytest test/srt/test_vision_openai_server_a.py`.

Fix Qwen3-Omni multimodal embedding handling and relax vision/audio tests

- MRotaryEmbedding:
  - Make audio-related kwargs (audio_token_id, audio_start_token_id, position_id_per_seconds, audio_seqlens) optional.
  - Introduce `has_audio` guard and skip audio-specific branches when no audio is present.
  - Ensure all newly created position_id tensors (`torch.arange`) are allocated on the same device as input_ids.

- mm_utils:
  - Change `_adjust_embedding_length` to take `special_multimodal_mask` instead of a generic mask.
  - Handle length mismatch robustly: pad with zeros when embeddings are shorter than the number of multimodal tokens, truncate when longer.
  - Replace hard RuntimeError with warnings and best-effort adjustment to avoid crashing on imperfect MM embeddings.

- Qwen3VL:
  - Rework `get_image_feature` to explicitly reconstruct patches based on `in_channels * temporal_patch_size * patch_size^2`.
  - Enforce full patch and patch-group (spatial_merge_size^2) alignment, skipping invalid/too-short inputs.
  - Return empty tensor when there are no valid image patches, and ensure all tensors are on the visual module’s device/dtype.

- Tests (OpenAI vision server):
  - Override `verify_single_image_response` in `TestQwen3OmniServer` to only check high-level structure:
    - presence of “1.” and “2.”,
    - mentions of image/picture/photo and audio/sound/speech,
    - valid usage stats.
  - Add a Qwen3-Omni-specific `verify_speech_recognition_response` that checks structural/audio mentions instead of exact transcript words.
  - Fix a bug in common `verify_single_image_response` where `"person"` was not actually checked with `in text`.

## Checklist


hnyls2002:
Please rebase the branch and remove all unrelated commits @minleminzui

Issue #11490, opened 2025-10-12
[Feature] Todos/CheckList for Piecewise Cuda Graph & Torch Compilation Support


Oasis-Git:
### Checklist

- [x] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 2. Please use English, otherwise it will be closed.

### Motivation

With PR #10062 merged, we have implemented the foundational framework for piecewise CUDA Graph and Torch Compile backend support.  
In this issue, we aim to outline the key follow-up tasks and enhancements planned for future iterations.

### Todo List
- [ ] Fix B200 can not use piecewise-cuda-graph. @BBuf 
- [ ] **DeepSeek V3 Architecture Support**: Add support for MLA. @ispobock
- [ ] **Sgl-kernel Update**: Move the jit script of `weak_ref_tensor` to `sgl_kernel` @ispobock 
- [ ] **`custom_ops` Integration**: Enable custom operator support for `cuda_forward`. @BBuf 
- [ ] **Pipeline Parallelism (PP)**: Implement full PP functionality and integration with the existing compile graph backend.
- [ ] **Data Parallelism (DP)**: Introduce DP compatibility and synchronization mechanisms.
- [ ] **Compilation Cache Management**: Implement compilation code storage and loading to accelerate torch compilation @Oasis-Git 
- [ ] **Quantization Support**: Extend backend to support quantized model execution.
- [ ] **PassManager Support**: Better Fusion Kernel with torch compile enabled. @yuan-luo 
- [ ] **Other Attention Support**: Linear Attention, SWA and Hybrid Attention
- [ ] **LogProbs Support**: Support `LogProbs` return value @Oasis-Git 

We list all the model that can not be applied with piecewise cuda graph support by now. All the contributors and users can raise their model with issues here.
### Model List
- [ ] **Qwen/Qwen3-235B-A22B**
- [ ] **deepseek-ai/DeepSeek-V3**

_Contributions and discussions are highly welcome._

yuan-luo:
@DevashishLal-CB and I are working on similar tasks. https://github.com/sgl-project/sglang/pull/10987
We tested vLLM with and without Piecewise Cuda Graph, with Piecewise Cuda Graph got few speedup in Qwen2-VL model so postponed the progress.
Will continue based on main then.

merrymercy:
[...]

PR #13081, opened 2025-11-11
Support piecewise cuda graph for Qwen3-next

Chen-0210:
## Motivation
Support piecewise cuda graph for Qwen3-next
https://github.com/sgl-project/sglang/issues/11490

## Modifications
1. Temporarily split the entire GDN attention due to too many parameters in the linear-attention function and some parts in linear attenion is incompatible with cuda graph. The performance is not ideal at the moment, but it can be refactored and optimized later.
2. Disable dual_stream when piecewise CUDA graph is enabled.
3. Changed to zeros_like instead of empty_like in GDN.
  Because `q/k/v.shape[0]` is larger than the real `seq_len`, GDN does not write the padded rows, and `empty_like` leaves them as `NaN`. Passing such padded QKV to FlashInfer prefill is undefined input, meaning its behavior is not expected... It seems load the NaNs in the padding and make the valid tokens become NaN as well. While fa3 backend doesn’t have this issue.

## Accuracy Tests

```
python3 -m sglang.launch_server --model Qwen/Qwen3-Next-80B-A3B-Instruct/ --tp 2 --enable-piecewise-cuda-graph --piecewise-cuda-graph-compiler eager 
python3 benchmark/gsm8k/bench_sglang.py --parallel 1319 --num-questions 1319
100%|████████████████████████████████████████| 1319/1319 [01:11<00:00, 18.50it/s]
Accuracy: 0.942
Invalid: 0.000
Latency: 71.401 s
Output throughput: 3114.390 token/s
```

## Benchmarking and Profiling(TTFT on bs=1 and H200x2)

`python3 -m sglang.bench_serving --backend sglang --dataset-name random --num-prompts 50 --random-input-len 4096 --random-output-len 1 --random-range-ratio 1 --max-concurrency 1`

| **Input length** | **1024** | **2048** | **4096** |
|:----------------:|:--------:|:--------:|:--------:|
| **with piecewise-cudagraph**    | 67.83 ms | 68.72 ms | 105.29 ms |
| **without piecewise-cudagraph** | 99.17 ms | 101.52 ms | 127.20 ms |

## Checklist


python/sglang/srt/layers/attention/fla/chunk_o.py
Oasis-Git, line 1: why zero_like instead of empty like

python/sglang/srt/layers/attention/fla/layernorm_gated.py
Oasis-Git, line 1: Why change the `device_context` to `cuda.device`

python/sglang/srt/models/qwen3_next.py
```diff
@@ -349,7 +355,11 @@ def fix_query_key_value_ordering(self, mixed_qkvz, mixed_ba):
         return query, key, value, z, b, a
 
     def _forward_input_proj(self, hidden_states: torch.Tensor):
-        DUAL_STREAM_TOKEN_THRESHOLD = 1024 if not _is_npu else 0
```
Oasis-Git, line 352: I think you lost the logic of npu side?

python/sglang/srt/models/qwen3_next.py
```diff
@@ -814,7 +840,12 @@ def forward(
         residual = None
         for i in range(len(self.layers)):
             layer = self.layers[i]
-            with get_global_expert_distribution_recorder().with_current_layer(i):
+            ctx = (
```
Oasis-Git: So the reason for remove context is that it can not integrate with compile?

Oasis-Git (commented):
Leave the comment for revision before merge

python/sglang/srt/models/qwen3_next.py
Chen-0210: It is already been moved to line 358. The logic remains the same.

python/sglang/srt/layers/attention/fla/layernorm_gated.py
Chen-0210: There’s an if inside device_context, so it doesn’t work with torch.compile. Since I split the whole linear layer, the norm part isn’t compiled anymore anyway. So that change is not needed

Chen-0210: fixed

python/sglang/srt/models/qwen3_next.py
Chen-0210: I thought it was incompatible with compile before.... After checking it’s fine, removed it.

python/sglang/srt/layers/attention/fla/chunk_o.py
Chen-0210: empty_like may produce tensors containing NaN values, which can eventually cause a crash.

yizhang2077: I think `o` has been filled value into triton kernel?

Chen-0210: The padding part will not be filled. `cu_seqlens` controls it, and the padding part is skipped. @yizhang2077

Chen-0210: added some clarifications in the description.

Oasis-Git:
LGTM. Will approve it after testing on my side.

test/srt/models/test_qwen3_next_models.py
```diff
@@ -135,5 +135,43 @@ def test_gsm8k(self):
         self.assertGreater(metrics["accuracy"], 0.93)
 
 
+class TestQwen3NextPiecewiseCudaGraph(CustomTestCase):
```
ispobock, line 138: Can we move it to https://github.com/sgl-project/sglang/blob/main/test/srt/test_piecewise_cuda_graph.py?
And also update the estimated time in  https://github.com/sgl-project/sglang/blob/760c20b360410412f1b48604cbc8117b083edb54/test/srt/run_suite.py#L99.

Chen-0210: Qwen3-Next-80B requires at least 2 GPUs to deploy, but the test in `test_piecewise_cuda_graph`.py only runs on 1 GPU.  So I think we can’t move it to `test_piecewise_cuda_graph`.py

ispobock: I see.

yizhang2077 (approved):
LGTM

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -363,6 +363,11 @@ def __init__(
                     elif hasattr(layer.self_attn, "attn_mqa"):
                         # For DeepSeek model
                         self.attention_layers.append(layer.self_attn.attn_mqa)
+                # For hybrid model
+                elif hasattr(layer, "attn"):
+                    self.attention_layers.append(layer.attn)
+                elif hasattr(layer, "linear_attn"):
```
hebiao064, line 369: we need to be careful about the order of the layers

Chen-0210: Are there any cases the order could be a problem?

PR #11477, opened 2025-10-11
fix: detect CUDA libraries in /usr/lib/

hayko406:
## Motivation

Fixes a runtime error when loading CUDA libraries on systems where libcudart.so.12 is installed in system library paths (e.g., /usr/lib/x86_64-linux-gnu), which is common on Debian/Ubuntu.
Without this change, sglang fails to locate CUDA and cannot initialize properly on such setups.

## Modifications

Extended the CUDA library search logic to include common system library paths (/usr/lib/x86_64-linux-gnu, /usr/lib64, /usr/lib).

## Checklist

- [ ] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.ai/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.ai/developer_guide/contribution_guide.html#benchmark-the-speed).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CUDA runtime detection by searching common system library paths, reducing startup failures on diverse Linux setups.
  * Clearer error messaging when the CUDA runtime cannot be located.

Issue #11333, opened 2025-10-08
[Bug] Importing `sgl_kernel` now hard-requires CUDA toolkit (regression vs `v0.5.1`)


anvdn:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Describe the bug

After [this change](https://github.com/sgl-project/sglang/pull/8813/), imports of `sgl_kernel` now face a runtime error if the CUDA toolkit (e.g. `/usr/local/cuda/`) isn't present on the machine. That even if CUDA runtime + driver are present. 
```
from sgl_kernel import gelu_and_mul, silu_and_mul
  File "/tmp/pex/venvs/1/dc55218965a020edeb15eb54f013bb950c7766d2/779eb2cc0ca9e2fdd204774cbc41848e4e7c5055/lib/python3.12/site-packages/sgl_kernel/__init__.py", line 217, in <module>
    raise RuntimeError("Could not find CUDA lib directory.")
RuntimeError: Could not find CUDA lib directory.
```
Is there a valid reason for introducing this new runtime error, or could we just ignore it instead?
I have been running `sglang[srt]==0.5.1` on instances that do not have the CUDA toolkit installed and never faced any issue when running inference.

### Reproduction

Import `sglang_kernel` on an instance on which the CUDA toolkit hasn't been installed.

### Environment

CUDA 12.8
sglang[srt]==0.5.3
torch==2.8.0

zhyncs:
https://github.com/sgl-project/sglang/pull/8813

@FlamingoPg may you help fix this

hayko406:
@anvdn check out my pull request 
https://github.com/sgl-project/sglang/pull/11477

I have fixed the issue with detecting the libcuda path.

anvdn:
[...]

PR #8813, opened 2025-08-05
CUDA Arch Independent

EduardDurech:
`platform.machine()` isn't accurate for sbsa-linux, see my comments and review https://github.com/sgl-project/sglang/pull/5746 and https://github.com/fzyzcjy/torch_memory_saver/pull/19

```bash
$ python -c "import platform; print(platform.machine())"
> aarch64
```

But `$CUDA_HOME/[lib64 -> targets/sbsa-linux/lib]` NOT `targets/aarch64-linux/lib`

PR #5746, opened 2025-04-25
[FEATURE] Enhance platform compatibility for ARM

johnnynunez:
ARM GH200/GB200 jetson and more
@zhyncs done

PR #13089, opened 2025-11-11
[sgl-kernel] fix runtime error while preloading CUDA runtime

anvdn:
## Motivation

~1 month ago, I opened the following [issue](https://github.com/sgl-project/sglang/issues/11333). Essentially, [this older change](https://github.com/sgl-project/sglang/pull/8813) introduced (I am assuming inadvertently) a regression by adding a runtime error if we can't preload CUDA runtime from system dirs. Unfortunately, this binary isn't always present there, in particular if the CUDA toolkit isn't installed on the machine (e.g. we aren't using a heavy `devel` NVIDIA docker image). In such cases, we would error out at runtime with the following log, even though this step isn't strictly required for inference.
```
from sgl_kernel import gelu_and_mul, silu_and_mul
  File "/tmp/pex/venvs/1/dc55218965a020edeb15eb54f013bb950c7766d2/779eb2cc0ca9e2fdd204774cbc41848e4e7c5055/lib/python3.12/site-packages/sgl_kernel/__init__.py", line 217, in <module>
    raise RuntimeError("Could not find CUDA lib directory.")
RuntimeError: Could not find CUDA lib directory.
```

## Modifications

- Relaxes the logic to only attempt to preload the binary, as it was implemented prior to the change referenced above
- Extend the logic to look under `/usr/lib*` which are other common locations on Debian/Ubuntu (see https://github.com/sgl-project/sglang/pull/11477)

Note it's possible we may drop the preloading logic altogether and offload that completely to `torch` at init (e.g. [here](https://github.com/pytorch/pytorch/blob/104b86861862ccc47f6df6528347a017bef813ad/torch/__init__.py#L363)). For context, the logic was initially added to prevent a missing symbol error on GH200/GB200 on older CUDA versions (see https://github.com/sgl-project/sglang/pull/5746). Unfortunately I am unable to verify if we can drop it given I do not have access to such instances.

## Checklist


sgl-kernel/python/sgl_kernel/load_utils.py
```diff
@@ -205,20 +205,28 @@ def _find_cuda_home():
...
-    cuda_include = (cuda_path / "libcudart.so.12").resolve()
-    if cuda_include.exists():
-        ctypes.CDLL(str(cuda_include), mode=ctypes.RTLD_GLOBAL)
+    candidate_dirs = [
+        cuda_home / "lib",
+        cuda_home / "lib64",
+        Path("/usr/lib/x86_64-linux-gnu"),
+        Path("/usr/lib/aarch64-linux-gnu"),
+        Path("/usr/lib64"),
+        Path("/usr/lib"),
```
Inokinoki, line 217: I think one promising way, to detect the lib in the env, would be detecting whether there are `nvidia-cuda-runtime`-like packages (such packages can be found on [pytorch pypi](https://download.pytorch.org/whl/nvidia-cuda-runtime-cu12/), and NVIDIA's [nvidia-cuda-runtime-cu11](https://pypi.nvidia.com/nvidia-cuda-runtime-cu11/), [nvidia-cuda-runtime-cu12](https://pypi.nvidia.com/nvidia-cuda-runtime-cu12/), [nvidia-cuda-runtime](https://pypi.nvidia.com/nvidia-cuda-runtime/)) installed in the current env.

```shell
$ pip list
...
nvidia-cuda-runtime-cu12 (12.0.107)
...
```

If so, its paths could be also candidate dirs:

```python
>>> import nvidia.cuda_runtime.lib
>>> nvidia.cuda_runtime.lib.__path__
['/home/ubuntu/dev/test-cudart/lib/python3.6/site-packages/nvidia/cuda_runtime/lib']
```

anvdn: Agreed on that. However, given we currently have no reason to believe we need to keep this pre-loading logic, maybe that's not worth the extra complexity right now? In particular, I don't see any such logic in vLLM repo. Wdyt?

FlamingoPg:
Sure, thanks a lot!

sgl-kernel/python/sgl_kernel/load_utils.py
Inokinoki: I think that vLLM is doing it on "build-time" using pytorch, see: https://github.com/vllm-project/vllm/blob/main/setup.py#L21

```python
from torch.utils.cpp_extension import CUDA_HOME, ROCM_HOME
```

And they expect to have `torch.cuda` for either AMD HIP or NVIDIA CUDA stuffs. But tell me if I misunderstood it.

Inokinoki: If so it seems more interesting to have such path in the list as well ;)

anvdn: I am not sure I follow. Do you mean including `torch.utils.cpp_extension.CUDA_HOME` in the list? It seems like the logic to build this path is the same as the one we have in the repo (most probably it was copy-pasted from `torch`).
I was referring to the logic of preloading the binary, not the logic to build the `CUDA_HOME` path.

Inokinoki: I'm not sure as well: it seems that the code in pytorch is also able to find the cuda runtime installed by pip. If so, I think it could be helpful then.

Nevermind, I would leave the maintainers and you to review and decide. The mine is just some random findings and thinkings.

FlamingoPg (approved):

PR #13091, opened 2025-11-11
CI: server performance test for SGLang Diffusion

adarshxs:
Server-based performance test for SGLang Diffusion:
- Launches `sglang serve` with an image diffusion model.
- Issues a small generation via the OpenAI-compatible Images API.
- Verifies that performance metrics are emitted:
     * End-to-end time from performance.log
     * Denoising step average duration from performance.log
     * Pipeline stage timings from stdout

TODO:
- extend to ALL the models and some special server args (parallel, quant, offload, etc). Meanwhile, remove duplicate ones from CLI tests. CLI tests are left for basic coverage
- split tests to 3 or more smaller ones, to reduce the overall CI time
- accuracy is still not tested
- eventually we’ll need a bench_serving similar to the text-model one


python/sglang/multimodal_gen/test/server/test_server_performance.py
```diff
@@ -0,0 +1,421 @@
...
+    "TextEncodingStage": 834.20,
+    "ConditioningStage": 0.10,
+    "TimestepPreparationStage": 10.60,
+    "LatentPreparationStage": 5.20,
+    "DenoisingStage": 21202.60,
+    "DecodingStage": 327.60,
+}
+
+# Qwen-Image on 1xH100 80GB assertions
+EXPECTED_DENOISE_STEP_MS = {
```
mickqian: we may want to avoid testing each timestep, may be the first, 20%, 40 %, ...etc?

python/sglang/multimodal_gen/test/server/test_server_performance.py
```diff
@@ -0,0 +1,421 @@
...
+def _extract_avg_denoise_ms(records: List[dict]) -> float | None:
+    durations, _ = _collect_denoise_step_stats(records)
+    if not durations:
+        return None
+    return sum(durations) / len(durations)
+
+
+_STAGE_PATTERN = re.compile(
+    r"\[(?P<stage>[^\]]+)\]\s+(?:finished in\s+(?P<sec>[0-9.]+)\s+seconds|"
+    r"Execution completed in\s+(?P<ms>[0-9.]+)\s+ms)"
```
mickqian: we should dump the runtime stats to `performance.log` instead of doing regex

python/sglang/multimodal_gen/test/server/test_server_performance.py
```diff
@@ -0,0 +1,421 @@
...
+        except Exception:
+            pass
+        try:
+            stdout_fh.flush()
+            stdout_fh.close()
+        except Exception:
+            pass
+
+
+def test_server_perf_metrics_image(server):
```
mickqian: make this a test class, similar to the ones for `test_vision_openai_server.py`, so that we can expand it to other models with ease

mickqian (changes requested):

python/sglang/multimodal_gen/test/server/test_server_performance.py
mickqian: and some other metrics too, like avg and median

.github/workflows/pr-test.yml
```diff
@@ -298,16 +298,33 @@ jobs:
...
+          merge-multiple: true
+          pattern: wheel-python3.10-cuda12.9
+
+      - name: Install dependencies
+        run: |
+          CUSTOM_BUILD_SGL_KERNEL=${{needs.check-changes.outputs.sgl_kernel}} bash scripts/ci/ci_install_dependency.sh
+
+      - name: Run diffusion server tests
+        env:
+          SGLANG_TEST_IMAGE_MODEL: Qwen/Qwen-Image
```
mickqian: Can we avoid this, just extend the base test class with new models?

mickqian (commented):
Could we combine the accuracy test and perf test in a single one?

python/sglang/multimodal_gen/test/server/test_server_performance.py
```diff
@@ -0,0 +1,543 @@
...
+    "InputValidationStage": 0.10,
+    "TextEncodingStage": 834.20,
+    "ConditioningStage": 0.10,
+    "TimestepPreparationStage": 10.60,
+    "LatentPreparationStage": 5.20,
+    "DenoisingStage": 21202.60,
+    "DecodingStage": 327.60,
+}
+
+EXPECTED_STAGE_MS_IMAGE_EDIT = {
```
mickqian: If we have to, please store these statistic in a separate file, `.json` or others

adarshxs:
For accuracy, I was thinking we could measure SSIM, gt being an image generated using vanilla diffusers with a seed. We can extend this to videos frame by frame [as done here](https://github.com/hao-ai-lab/FastVideo/blob/main/fastvideo/tests/inference/lora/test_lora_inference_similarity.py)

.github/workflows/pr-test.yml
adarshxs: done

python/sglang/multimodal_gen/test/server/test_server_performance.py
adarshxs: moved to `perf_baselines.json`

python/sglang/multimodal_gen/test/server/test_server_performance.py
```diff
@@ -0,0 +1,507 @@
...
+        pass
+
+
+@pytest.mark.usefixtures("diffusion_server")
+class DiffusionServerPerfTestBase:
+    MODEL_PATH = os.environ.get("SGLANG_TEST_IMAGE_MODEL", "Qwen/Qwen-Image")
+    SERVER_PORT = int(os.environ.get("SGLANG_TEST_SERVER_PORT", "30100"))
+    PROMPT = "A minimal colorful icon of a raccoon face, flat style"
+    IMAGE_EDIT_PROMPT: str | None = None
+    IMAGE_EDIT_PATH = Path(__file__).resolve().parents[1] / "test_files" / "rabbit.jpg"
```
mickqian: change the image-edit test to the one in lmsys blog, in a following PR. and remove the image from repo

adarshxs: Sure

Issue #12845, opened 2025-11-07
[Feature] Support Deterministic Inference for QWen3-Next


hebiao064:
### Checklist

- [ ] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [ ] 2. Please use English, otherwise it will be closed.

### Motivation

As title

PR #13100, opened 2025-11-12
[Deterministic] Support Qwen3-Next model deterministic inference

zminglei:
## Motivation
Support hybrid model like Qwen3-Next model deterministic inference, [issue](https://github.com/sgl-project/sglang/issues/12845)

Launch Qwen3-Next model with enabling deterministic
`python3 -m sglang.launch_server --model-path /shared/public/elr-models/Qwen/Qwen3-Next-80B-A3B-Instruct/ --tp 4 --context-length 262144 --mem-fraction-static 0.8 --enable-deterministic-inference`

**Before the change**:
```
python3 -m sglang.test.test_deterministic --test-mode prefix --n-trials 50
Prompt 0 with prefix length 1: total samples: 285, Unique samples: 8
Prompt 1 with prefix length 511: total samples: 340, Unique samples: 49
Prompt 2 with prefix length 2048: total samples: 335, Unique samples: 14
Prompt 3 with prefix length 4097: total samples: 315, Unique samples: 1
```

```
Accuracy: 0.940
Invalid: 0.000
Latency: 25.232 s
Output throughput: 1311.006 token/s
```
**After the change**:
```
python3 -m sglang.test.test_deterministic --test-mode prefix --n-trials 50
Prompt 0 with prefix length 1: total samples: 350, Unique samples: 1
Prompt 1 with prefix length 511: total samples: 317, Unique samples: 1
Prompt 2 with prefix length 2048: total samples: 323, Unique samples: 1
Prompt 3 with prefix length 4097: total samples: 285, Unique samples: 1
```

```
Accuracy: 0.940
Invalid: 0.000
Latency: 25.077 s
Output throughput: 1319.095 token/s
```

## Modifications
Added comprehensive debug logs to identify the non-deterministic came from the `initial_states` are different for the same prompts. 
In this change, we make the mamba cache `initial_states` are consistent for the same prompts.

## Accuracy Tests
```
Accuracy: 0.940
Invalid: 0.000
Latency: 25.077 s
Output throughput: 1319.095 token/s
```

## Checklist


python/sglang/srt/layers/attention/hybrid_linear_attn_backend.py
```diff
@@ -716,7 +718,18 @@ def forward_extend(
                 retrieve_parent_token=retrieve_parent_token,
             )
         else:
+            # Zero-initialize cache slots for sequences without prefix cache
```
hebiao064: if this is true, pls add this logic to  MambaPool.alloc

zminglei: Yes, I've added debug logs and verified we do have different non-zero init states for the seqs which don't hit prefix cache. I've thought about adding logic somewhere inside MambaPool, but there we don't have info about the `forward_batch.extend_prefix_lens` to tell which seqs are not hitting prefix cache.
```
[EXTEND:cache_bug_detected] Layer 46: Found 1 sequences with extend_prefix_lens=0 but non-zero cache states (max_abs=0.158144)
[EXTEND:cache_bug_detected] Layer 46: Found 1 sequences with extend_prefix_lens=0 but non-zero cache states (max_abs=0.181680)
[EXTEND:cache_bug_detected] Layer 46: Found 1 sequences with extend_prefix_lens=0 but non-zero cache states (max_abs=0.359189)
[EXTEND:cache_bug_detected] Layer 46: Found 1 sequences with extend_prefix_lens=0 but non-zero cache states (max_abs=0.111003)
```
Ideally, we're expecting before reusing of cache slots with new reqs, these slots should already be zeroed. There must be something wrong with the mamba cache free logic which caused this case.

zminglei: nvm, I found why the cache wasn't be reinitialized to 0 before reusing. Pls take another look

hebiao064: @hanming-lu @cctry fyi

hebiao064 (approved):

byjiang1996 (approved):

zminglei:
> @zminglei Please add a test to nightly

Sounds good, just added a nightly test, pls take another look. 
Note: We have to set `SGLANG_BATCH_INVARIANT_OPS_ENABLE_MM_DEEPGEMM` for launching qwen3Next with deterministic before this [fix](https://github.com/sgl-project/sglang/pull/13226) is checked in.

Fridge003 (approved):

hebiao064 (approved):

ocss884:
It seems this PR breaks the `release_memory_occupation` feature for model using mamba cache, I'm looking into it cc @yizhang2077

ocss884:
> > It seems this PR breaks the `release_memory_occupation` feature for model using mamba cache, I'm looking into it cc @yizhang2077
> 
> @ocss884 could you elaborate more about the issue or the command to repro it? I may could help to take a look at it as well.

In current `release_memory_occupation` impl, a `flush_cache` will be called if kv is released (see [link](https://github.com/sgl-project/sglang/blob/96cc10834a47cbb70977dce9dacf93f777149b80/python/sglang/srt/managers/scheduler_update_weights_mixin.py#L128)). However, this PR makes `clear` to in-place set mamba cache (is released in this case) to zero, which leads to an IMA. @yizhang2077 fix it in #14209 . I add one more test on this case #14215

Issue #6608, opened 2025-05-26
[Feature] Customized mapping for LoRA weight names


lifuhuang:
### Checklist

- [x] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 2. Please use English, otherwise it will be closed.

### Motivation

The current LoRA impl in SGL maps LoRA weight to modules by (layer index, op_type) tuple, where op_type operation looks like `qkv_proj`, `o_proj`, `gate_up`, etc. This works fine for most standard cases, however, there are some limitations:
1. For models where there are more than one attention stacks (e.g., VLM), there could be multiple modules with the same (layer index, op_type), e.g., one from vision tower, the other from the language model. Currently SGL cannot handle such cases correctly and would usually fail during loading due to incorrect mapping.
2. Users cannot enable/disable application of LoRA at module-level, e.g., if user only wants to apply LoRA at language model but not vision (common); or when user only wants to apply LoRA at some layers but not the others (less common?), they cannot do that today. 
3. (Less common?) Models with non-standard LoRA weight / module names. 

### Proposal: 

* (Short-term) add an optional hook `should_apply_lora` at model level to allow model to customize LoRA application at model level. This would unblock most cases in 1 & 2. For example, for most VLMs, LoRAs should only be applied to language model but not vision tower. In these cases, model authors could simply disable LoRA application for modules in the vision tower, This would address the current LoRA loading failures due to incorrect mapping.
* (Long-term) generalize the hook to `map_lora_module_name` to allow model owner to map a given module to a specific LoRA weight name or return None when LoRA should not be applied. This would address 3 and some less common cases in 1 (e.g., when LoRA needs to be applied at both vision tower and language model)

(cc @Fridge003 )

PR #13109, opened 2025-11-12
add support for map_lora_module_name hook

glenliu21:
## Motivation

This PR implements the second, long-term step requested in #6608.

## Modifications

In `init_lora_modules`, we introduce the dictionary `self.module_name_to_lora_target` and use the `map_lora_module_name` hook if available to populate this dictionary. Then, `self.module_name_to_lora_target` is used in `update_lora_info` instead of the current naming heuristic in place. Additionally, all applicable models (mainly vision models) now have the `map_lora_module_name` hook.

## Accuracy Tests

Relevant unit and E2E tests are passing.

## Checklist


lifuhuang:
Hi @glenliu21, thank you for your contribution. 

I am not sure if I followed this PR, perhaps you can you share more thinking of yours to help me better understand. The original intention of #6608 is to fix the incorrect assumption we have that lora module and its corresponding target module in the base model can be uniquely identified via (layer index, suffix) tuple, the suffix here are things like `qkv_proj`, `gate_up_proj`, etc. 

One example why this assumption could be wrong: in MM models, (17, qkv_proj) could mean qkv_proj at the 17th layer in the vision tower, the audio tower, or the base model. Previously I worked it around by providing a hook `should_apply_lora` such that model owner can choose to disable lora for the MM tower, therefore the (layer, suffix) tuple become a unique ID again, but this is not ideal obviously because someone might want to apply lora to both vision tower and base model (in thoery), therefore we really need to have a more generic id / mapping system such that model owner can freely map any base module to any lora target.

It's not obviously clear how your change enables that scenario. Maybe it would be helpful to add a CI that applies adapter to both vision tower and base model to demonstrate.

glenliu21:
Thanks for the reply @lifuhuang!

After looking into this issue, I think the general problem of supporting LoRA for multimodal towers would require a large design overhaul. The TLDR is that the (layer_index, op_type) design is widely used across the LoRA backend, so the change is not so simple. An initial design is:

From model owner side:
- Define a hook like `map_lora_module_name` (or perhaps just `should_apply_lora_mm`) because multimodal LoRA adapters require special handling given the current design.

From SRT LoRA side, most of the changes will happen in `mem_pool.py`:
- In `mem_pool.py`, we must allocate buffers `self.mm_A_buffer` and `self.mm_B_buffer`. The current design of `self.A_buffer` and `self.B_buffer` only allocates enough space for the language model. Or if it makes sense, we could allocate more space in `self.A_buffer` and `self.B_buffer` dynamically when multimodal LoRA buffers are necessary.
- These new buffers must be handled separately in `LoRAMemoryPool.init_buffers`, which currently assumes it is allocating LoRA buffers for the language model. `load_lora_weight_to_buffer` and `get_tensor` similarly must account for this new case.
- In `LoRAAdapter.initialize_weights`, we would call the `map_lora_module_name` / `should_apply_lora_mm` hook to see if we need to store extra LoRA weights aside from the language model targets in CPU. These would have to be stored in `LoRAAdapter.mm_layers` or something similar to the current `LoRAAdapter.layers`.

As we have discussed, there doesn't seem to be much demand from the community for LoRA being applied to multimodal towers. As such, I'm not sure if supporting this feature is worth the time. Please let me know your thoughts!

PR #13113, opened 2025-11-12
[BugFix] weight load bug when checkpoint expert.gate and exepert.up_proj are not fused

Yuechguo:
## Motivation
1.let ep moe layer to gracefully handle expert_ids that do not belong to local moe rank, in order to handle the case where model checkpoint expert.gate and exepert.up_proj are not fused.
2.Update the method of referencing aiter lib in VisionAiterBackend.
## Modifications
1.Modify qwen_vl_moe and layer to gracefully handle expert_ids that do not belong to local moe rank.
2.Modify VisionAiterBackend in VisionAttention

## Accuracy Tests
setup sglang server:
```
USE_ROCM_AITER_ROPE_BACKEND=0 \
CUDA_VISIBLE_DEVICES=4,5,6,7 \
SGLANG_VLM_CACHE_SIZE_MB=4096 \
        --model-path /pretrained_model/qwen/Qwen3-VL-235B-A22B-Instruct-FP8-dynamic \
        --served-model-name  Qwen3-VL-235B-A22B-Instruct-FP8-dynamic \
        --attention-backend aiter \
        --host 127.0.0.1 \
        --port 28000 \
        --tp-size 4 --ep-size 4 --trust-remote-code \
        --chunked-prefill-size 32768 \
        --max-prefill-tokens 32768 \
        --context-len 32768 \
        --max-running-requests 512 \
        --mem-fraction-static 0.85 \
        --mm-attention-backend aiter_attn \
```
setup accuracy benchmark:
```
python3 /data/testhome/sglang/benchmark/mmmu/bench_sglang.py --port 28000 --concurrency 16
```

accuracy result:
image

## Checklist


python/sglang/srt/layers/attention/vision.py
```diff
@@ -348,8 +345,16 @@ def __init__(
...
+        if not _is_hip:
             raise Exception("aiter_attn is only available for AMD")
+        try:
+            from aiter import flash_attn_varlen_func as aiter_flash_attn_varlen_func
+        except ImportError:
+            print(
+                "aiter is AMD specific kernel library. Please make sure aiter is installed on your AMD device."
+            )
+
+        self.flash_attn_varlen_func = aiter_flash_attn_varlen_func
```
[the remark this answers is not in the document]
Yuechguo: already done

python/sglang/srt/layers/attention/vision.py
```diff
@@ -348,8 +346,16 @@ def __init__(
         self,
         **kwargs,
     ):
-        if not _use_aiter:
+        if not _is_hip:
             raise Exception("aiter_attn is only available for AMD")
+        try:
+            from aiter import flash_attn_varlen_func as aiter_flash_attn_varlen_func
```
HaiShaw, line 352: There are case `_is_hip` but `SGLANG_USE_AITER` isn't set.

HaiShaw (changes requested):

python/sglang/srt/layers/attention/vision.py
Yuechguo: VisisonAiterBackend is selected under two circumstances:

1.When the user specifies --mm-attention-backend aiter_attn or configures it in the user model script code. In this case, if _is_hip = true but aiter is not installed, VisisonAiterBackend will throw an import error.

2.When no user specification is provided, the selection is determined by _determine_attention_backend, which automatically evaluates _is_hip and _use_aiter conditions to decide which VisisonAttnBackend to use.

HaiShaw (approved):

Issue #13124, opened 2025-11-12
[Feature] Rotary embedding support float32 dtype


ispobock:
### Motivation

Add float32 dtype support for rotary embedding kernel.
https://github.com/sgl-project/sglang/blob/7ea5b42d7063e22e48b26ba09baef2f9fba2282b/python/sglang/srt/layers/rotary_embedding.py#L279

zlH518:
i can also join this work

PR #13141, opened 2025-11-12
Fix rotary embedding precision issue in line 279

souvikchand:
### Modifications
fixed the issue regarding [https://github.com/sgl-project/sglang/issues/13124]

## Checklist


hnyls2002:
Thanks @souvikchand for digging into this. The file this PR modifies has since been refactored into the rotary_embedding package and the fp32 dtype gating it edits no longer exists in that form; as ispobock noted, fp32 support needs to land in the kernel itself rather than by removing the dispatch check. Closing as the patch no longer applies; please reopen or file a fresh PR against the new code if the precision issue still reproduces.

PR #13143, opened 2025-11-12
Sync Glm4MoeDetector non-streaming output with streaming by assigning sequential tool_index

pansicheng:
## Modifications

Sync Glm4MoeDetector non-streaming output with streaming by assigning sequential tool_index

## Checklist


JustinTong0323 (approved):

pansicheng:
> thanks for the PR. What is the motivation though? In OpenAI, non-streaming tool call response usually doesn't have a tool_index field

The context is that when deploying GLM, users reported inconsistencies between streaming and non-streaming behaviors in tool call responses. If users' code logic relies on the tool_index field, they would need to implement special handling for it. However, if the convention is that non-streaming tool calls shouldn't include tool_index by default, then there's indeed no need to add it.

PR #13146, opened 2025-11-12
fix(hicache): prefetch kv cache from L3 based on bigram keys in eagle mode

lmyybh:
## Motivation

1. When **EAGLE** is enabled, in HiCache, `token_ids` are converted to **bigram keys**, and the hash is computed based on bigram keys. However, during prefetching from storage, the `token_ids` are not properly converted, resulting in a mismatch between the prefetched hash and the actual stored hash.
2. When a request has no matching prefix in the tree, it skips fetching the KV cache from L3, which may lead to redundant computation.

## Modifications

1. Call `key_convert_fn` in the `prefetch_from_storage` 
2. Try to prefetch kv cache from L3 when `req.last_node` is the root node

## Checklist

### DIFF
--- a/python/sglang/srt/managers/scheduler.py
+++ b/python/sglang/srt/managers/scheduler.py
@@ -1424,8 +1424,8 @@ class Scheduler(
     def _prefetch_kvcache(self, req: Req):
         if self.enable_hicache_storage:
             req.init_next_round_input(self.tree_cache)
-            if req.last_node.backuped:
-                # only to initiate the prefetch if the last node is backuped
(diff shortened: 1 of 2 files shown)


hnyls2002:
Thanks @lmyybh. This was fixed on main: #23631 normalizes the storage prefetch key for EAGLE bigram mode in HiRadixCache, and the scheduler prefetch path now also
  initiates prefetch when the last node is the root node. Closing as superseded; please reopen if you still hit a miss pattern not covered.

PR #13152, opened 2025-11-12
SGLang Tracing: Add trace-level, trace-module, and unify tracing/request-stage-metrics

sufeng-buaa:
## Motivation
The PR is response to https://github.com/sgl-project/sglang/issues/10916. For details on the motivation and visual output, please refer to the issue.

To reduce span overhead, we added the trace-level feature. To support broader use cases beyond request tracing, we introduced trace-module.

## Modifications
1. Refactored tracing package from global-state functions to a class-based design with instance storage. This facilitates integration with request stage metrics and provides a hook for future dynamic instrumentation.
2. Implemented a wrapper class "SglangStageContext" that internally aggregates trace context and metric collector uniformly collect timestamps and route to different export paths based on configuration.
3. Added trace level mechanism to assign levels to each RequestStage, helping reduce excessive trace data in production environments.
4. Added trace module mechanism to extend the trace package beyond request tracing, enabling its use in other modules such as hicache.

**I thought about unifying TimeStat too, but it would require too many changes, so I gave up on that.** May I will push a draft patch later.

## Instrumentation Overhead Evaluation
The overhead of each instrumentation point remains almost unchanged compared to before. See https://github.com/sgl-project/sglang/pull/9962 and https://github.com/sgl-project/sglang/pull/10804

## Checklist


zhanghaotong:
Hi~ I'm running your code with the following command:
```
python -m sglang.launch_server --trace-level 3 --otlp-traces-endpoint 0.0.0.0:4317  --model-path /mnt/modelops/models/Qwen3-8B/ --host 0.0.0.0 --log-level info  --port 8001
```
However, I forgot to install the OpenTelemetry packages. As a result, the engine crashed with the error shown below:
image
And perhaps we should explicitly check for the required OpenTelemetry dependencies when tracing is enabled, and raise a clear error to inform users if they are missing?

sufeng-buaa:
> Hi~ I'm running your code with the following command:
> 
> ```
> python -m sglang.launch_server --trace-level 3 --otlp-traces-endpoint 0.0.0.0:4317  --model-path /mnt/modelops/models/Qwen3-8B/ --host 0.0.0.0 --log-level info  --port 8001
> ```
> 
> However, I forgot to install the OpenTelemetry packages. As a result, the engine crashed with the error shown below: image And perhaps we should explicitly check for the required OpenTelemetry dependencies when tracing is enabled, and raise a clear error to inform users if they are missing?

I did forget to verify the case where OpenTelemetry is not installed but tracing is enabled. I'll fix it as soon as possible.

python/sglang/srt/tracing/trace.py
```diff
@@ -87,39 +79,41 @@ class SglangTraceThreadInfo:
     tracer: trace.Tracer
 
 
+@dataclass
+class SglangTraceEvent:
```
ShangmingCai:
```suggestion
class SGLangTraceEvent:
```
nit: we should probably use the correct uppercase and lowercase of SGLang.

sufeng-buaa: I have renamed all 'Sglang***' to 'SGLang***'

docs/references/production_request_trace.md
```diff
@@ -61,84 +61,87 @@ We have already inserted instrumentation points in the tokenizer and scheduler m
...
+    - Use the `ANONYMOUS` to not specify a slice name at the start of the slice, allowing the slice name to be determined by trace_slice_end.
     <br>Note: Anonymous slices must not be nested.
         ```python
-        trace_slice_start("", rid, anonymous = True)
-        trace_slice_end("slice A", rid)
+        stage_context.metric_trace_slice_start(RequestStage.ANONYMOUS)
+        stage_context.metric_trace_slice_end(RequestStage.TOKENIZER)
         ```
 
     - In trace_slice_end, use auto_next_anon to automatically create the next anonymous slice, which can reduce the number of instrumentation points needed.
```
acelyc111: Should it be?
```suggestion
    - In metric_trace_slice_end, use auto_next_anon to automatically create the next anonymous slice, which can reduce the number of instrumentation points needed.
```

python/sglang/srt/server_args.py
```diff
@@ -2446,9 +2447,16 @@ def add_cli_args(parser: argparse.ArgumentParser):
...
-            help="Enable opentelemetry trace",
+            "--trace-level",
+            type=int,
+            default=ServerArgs.trace_level,
+            help="Enable opentelemetry trace level",
+        )
+        parser.add_argument(
+            "--trace-module",
+            type=str,
+            default=ServerArgs.trace_module,
```
acelyc111, line 3156: What are the optional items for this argument?

python/sglang/srt/server_args.py
```diff
@@ -2446,9 +2447,16 @@ def add_cli_args(parser: argparse.ArgumentParser):
             help="Config in json format for NVIDIA dynamo KV event publishing. Publishing will be enabled if this flag is used.",
         )
         parser.add_argument(
-            "--enable-trace",
-            action="store_true",
-            help="Enable opentelemetry trace",
+            "--trace-level",
+            type=int,
+            default=ServerArgs.trace_level,
```
acelyc111, line 3150: How about listing the choices and describing the meanings like `--log-requests-level`?

docs/references/production_request_trace.md
sufeng-buaa: ok, I will recorrect it

python/sglang/srt/server_args.py
sufeng-buaa: The tracing package is not only used for tracking requests—for example, we are currently exploring its use in monitoring hierarchical caches. Therefore, we use the `--trace-module` parameter to enable tracing for specific modules. The default set is "request".

python/sglang/srt/server_args.py
sufeng-buaa: ok, good suggestion.

docs/references/production_request_trace.md
```diff
@@ -1,6 +1,6 @@
 # Production Request Tracing
 
-SGlang exports request trace data based on the OpenTelemetry Collector. You can enable tracing by adding the `--enable-trace` and configure the OpenTelemetry Collector endpoint using `--otlp-traces-endpoint` when launching the server.
+SGLang exports request trace data based on the OpenTelemetry Collector. You can enable tracing by adding the `--trace-level` and configure the OpenTelemetry Collector endpoint using `--otlp-traces-endpoint` when launching the server. The `--trace-level` option accepts configurable values from `1` to `3`, with higher numbers indicating more detailed tracing. Additionally, you can use `--trace-module` to specify the module to trace; currently, only `request` is supported.
```
ShangmingCai: QQ: since you removed `--enable-trace`, should we inform the user `--trace-level 0` is equal to setting tracing option to `False` here?

python/sglang/srt/tracing/trace_metric_warpper.py
```diff
@@ -0,0 +1,369 @@
...
+        "wait_pd_finish",
+        level=2,
+    )
+
+    # other
+    ANONYMOUS = RequestStageConfig("")
+
+
+class SGLangStageContext(SGLangTraceReqContext):
+    def __init__(
```
ShangmingCai: Actually, I am not sure about the naming here. Do we really need the `SGLang` prefix when developing a feature in the `python/sglang` dir? Seems unnecessary to me.

Maybe something like `InferenceStageContext` or `TracingStageContext`, or another more accurate option?

docs/references/production_request_trace.md
sufeng-buaa: ok, I have updated the doc.

python/sglang/srt/tracing/trace_metric_warpper.py
sufeng-buaa: emmm, As for the naming, I'm not entirely sure. This class encapsulates both tracing and metrics, so using "Tracing" as a prefix feels incomplete. My current naming is indeed not ideal. Let me think about it for a moment.

python/sglang/srt/tracing/trace_metric_warpper.py
```diff
@@ -0,0 +1,369 @@
...
+            attrs=attrs,
+            auto_next_anon=auto_next_anon,
+            thread_finish_flag=thread_finish_flag,
+            level=stage.level,
+        )
+
+    metric_trace_slice = metric_trace_slice_end
+
+
+class NoOpStageContext:
```
ShangmingCai: Does this mean "No operation"? `NullStageContext` sounds more accurate to me.

sufeng-buaa: ok， it sounds good.

python/sglang/srt/tracing/trace_metric_warpper.py
sufeng-buaa: How about `StageObserveContext` or `TraceMetricContext`?

python/sglang/srt/tracing/trace_metric_wrapper.py
```diff
@@ -0,0 +1,369 @@
...
+            auto_next_anon=not req.finished(),
+            thread_finish_flag=req.finished(),
+        )
+
+
+def trace_event_batch(
+    name: str,
+    reqs: List,
+    ts: Optional[int] = None,
+    attrs: Dict[str, Any] = {},
```
ShangmingCai, line 321: Should this be `Optional[Dict[str, Any]] = None ` as well?

ShangmingCai (commented):
typo: python/sglang/srt/tracing/trace_metric_warpper.py  ->  python/sglang/srt/tracing/trace_metric_wrapper.py

python/sglang/srt/tracing/trace_metric_warpper.py
ShangmingCai: TraceMetricContext sounds better

python/sglang/srt/tracing/trace_metric_wrapper.py
sufeng-buaa: If `attrs` is `None`, `_attrs.update(attrs)` will trigger an error. I don't want to add an extra conditional check, so I'll simply set the default value to an empty dictionary.

python/sglang/srt/tracing/trace_metric_warpper.py
sufeng-buaa: I have updated the code.

python/sglang/srt/tracing/trace_metric_warpper.py
sufeng-buaa: I've updated `NoOpStageContext` to `NullContext`, since `SglangStageContext` has been renamed to `TraceMetricContext`, making the term "Stage" redundant.

sufeng-buaa:
> typo: python/sglang/srt/tracing/trace_metric_warpper.py -> python/sglang/srt/tracing/trace_metric_wrapper.py

Fixed.

python/sglang/srt/tracing/trace.py
```diff
@@ -287,451 +286,470 @@ def trace_set_thread_info(
...
+                context=external_trace_context,
+            )
+            self.bootstrap_room_span = bootstrap_room_span
+            self.bootstrap_room_span_context = trace.set_span_in_context(
+                bootstrap_room_span
+            )
+        else:
+            self.bootstrap_room_span_context = remote_trace_contexts[
+                str(bootstrap_room)
+            ].root_span_context
```
chivalryq: `bootstrap_room` is always 0 for mix(non-PD) deployment? so `remote_trace_contexts` would save only one key like:

```
["0": bootstrap_room_span_context]
```

Will this cause a chaos for mixed deployment? I did some test and found that I can't correctly propagate upstream trace_id to SGLang for mix deployment.

python/sglang/srt/tracing/trace.py
```diff
@@ -69,16 +65,12 @@ class IdGenerator:
...
-    return tracing_enabled
-
-
 def extract_trace_headers(headers: Mapping[str, str]) -> Optional[Dict]:
     return {h: headers[h] for h in TRACE_HEADERS if h in headers}
 
 
 @dataclass
-class SglangTraceThreadInfo:
+class SGLangTraceThreadInfo:
```
chivalryq: As per @ShangmingCai comments, `SGLangTraceThreadInfo`, `SGLangTraceEvent`, `SGLangTraceThreadContext`, `SGLangTracePropagateContext` seems no need to have `SGLang` as prefix

chivalryq (commented):
My two cents on this PR. I did some integration with SGLang tracing, found some issues. One is in the comment (about bootstram_room hiearachy when there's no PD disaggregation). 
Another question is: why we use a `trace_context` as HTTP header which didn't respecting [W3C recommandation: trace-context](https://www.w3.org/TR/trace-context/). In this recommandation, `traceparent` and `tracestate` is a standard HTTP header. A customized header makes it more complex to propagate trace id from a  proxy or load-balancer when using [otel SDK](https://opentelemetry.io/)

Sorry if I missed something, thanks for answer in advance. 

BTW is there an observerbility related slack channel for better communicating?

python/sglang/srt/tracing/trace.py
zhanghaotong: In the right situation, the upstream span can propagate the OTLP context to bootstrap_room_span via HTTP headers for mix(non-PD) deployment. Let me verify that.

python/sglang/srt/tracing/trace.py
sufeng-buaa: Let me think about it, since we’ve already wrapped the OTel spans with a custom layer here, adding an `SGLang` prefix wouldn’t feel out of place? What do you think? @ShangmingCai

python/sglang/srt/tracing/trace.py
sufeng-buaa: As mentioned in the reply above, the `bootstrap_room span` might not be appropriate to create, and I'm currently deciding whether to remove it in this patch or in the next one.

sufeng-buaa:
I have removed `bootstrap room span` and `trace_context` headers. cc @chivalryq @zhanghaotong @ShangmingCai

chivalryq:
@sufeng-buaa Cool, How would you like to solve "find prev span" problem after removing custom "trace_context" header? Actually, I doubt that if we really need "prev_span" record. When integrating the SGLang tracing, I uses a internal-implemented load balancer and pass `prev_span` always as "None" when requesting SGLang, and otel can properly connect the two services

image

sufeng-buaa:
> @sufeng-buaa Cool, How would you like to solve "find prev span" problem after removing custom "trace_context" header? Actually, I doubt that if we really need "prev_span" record. When integrating the SGLang tracing, I uses a internal-implemented load balancer and pass `prev_span` always as "None" when requesting SGLang, and otel can properly connect the two services
> 
> image

In the view you provided, `prev_span` indeed doesn't serve much purpose. However, in my routine debugging, I actually find this view rather limited. I personally prefer converting the trace data into Chrome JSON format so that I can visualize all requests using Perfetto. In Perfetto, I use arrows (links) to explicitly connect the execution flow between spans, which requires explicitly recording each span's previous span via `prev_span`.In contrast, Jaeger’s UI infers the execution flow visually by observing the chronological order of spans. 

Without the `prev_span` information, my data conversion script would need to explicitly define predecessor-successor relationships between spans—for example, mapping 'tokenizer dispatch' → 'scheduler prefill'. However, as our request execution flow evolves with future iterations, these relationships will likely change, forcing me to frequently maintain and update this conversion script.

After removing the custom 'trace_context' header, `prev_span` becomes purely an internal concept and shouldn't impact the integration between upper-layer services and the SGLang engine anymore.
image

python/sglang/srt/tracing/trace.py
zhanghaotong: Is here `d["prev_span"] == "None"` or `d["prev_span"] is None` ?

sufeng-buaa: Indeed. I've fixed it. This code path is currently not covered by tests.

test/manual/test_tracing.py
```diff
@@ -68,12 +68,13 @@ def __clear_trace_file(self):
...
         self.__clear_trace_file()
         assert self.__launch_otel_jaeger()
+        self.addCleanup(self.__stop_otel_jaeger)
 
         process = popen_launch_server(
             DEFAULT_SMALL_MODEL_NAME_FOR_TEST,
             DEFAULT_URL_FOR_TEST,
             timeout=DEFAULT_TIMEOUT_FOR_SERVER_LAUNCH,
-            other_args=["--enable-trace", "--otlp-traces-endpoint", "0.0.0.0:4317"],
+            other_args=["--trace-level", "1", "--otlp-traces-endpoint", "0.0.0.0:4317"],
```
stmatengss: We need more comprehensive tests, including varied trace-level and trace-module coverage, to verify end-to-end profiling.

python/sglang/srt/managers/scheduler.py
```diff
@@ -2096,7 +2078,7 @@ def process_batch_result(
     ):
         if batch.forward_mode.is_decode():
             self.process_batch_result_decode(batch, result)
-            trace_slice_batch(RequestStage.DECODE_LOOP, batch.reqs)
+            metric_trace_slice_batch(RequestStage.DECODE_LOOP, batch.reqs)
```
zhanghaotong, line 2364: I set trace_level=3, and after processing each request, a warning log consistently appears—possibly due to the overlap scheduling strategy. Could you please look into this?

Image

sufeng-buaa: Let me try to reproduce it locally

test/manual/test_tracing.py
sufeng-buaa: ok, I will add more test cases

python/sglang/srt/managers/scheduler.py
sufeng-buaa: Could you provide the test commands or test parameters？I did not reproduce it in my test environment.

zhanghaotong: `python3 -m sglang.launch_server --model-path /mnt/modelops/models/Qwen3-8B/ --host 0.0.0.0 --trace-level 3 --tensor-parallel-size 1 --enable-metrics`

sufeng-buaa: Fixed.

python/sglang/srt/tracing/trace.py
```diff
@@ -287,451 +283,391 @@ def trace_set_thread_info(
...
-):
-    if not tracing_enabled:
-        return
+        self.start_time_ns = get_cur_time_ns()
+        self.is_copy = True
 
-    rid = str(rid)
+        trace_context = TracePropagateContext.instance_from_dict(trace_context)
+        if not trace_context:
+            self.tracing_enable = False
```
zhanghaotong: Maybe we should `return` after this line to avoid NPE?

sufeng-buaa: indeed. Fixed.
thanks

ShangmingCai (approved):
LGTM. Let's wait for the CI and see if other reviewers have any other comments.

python/sglang/srt/tracing/trace.py
```diff
@@ -208,10 +203,11 @@ def __get_host_id() -> str:
 
 # Should be called by each tracked process.
 def process_tracing_init(otlp_endpoint, server_name):
-    global tracing_enabled
-    global __get_cur_time_ns
+    global opentelemetry_initialized
+    global get_cur_time_ns
     if not opentelemetry_imported:
-        tracing_enabled = False
+        logger.warning("opentelemetry package is not installed!!! audo disable tracing")
```
zhanghaotong: There’s a minor typo: “auto”.

More importantly, I’m wondering whether we should raise an exception and terminate the main process to clearly alert users when required dependencies are missing—ensuring they understand that tracing will not function. We’ve encountered this issue repeatedly: users forget to install the necessary dependencies and later contact us confused about why no traces are being uploaded.

Moreover, I found that both vLLM and TensorRT-LLM adopt this approach—they raise exceptions in such scenarios to prevent silent failures.

sufeng-buaa: That makes sense, I'll make the changes.

PR #13162, opened 2025-11-12
Fix nan in global scaling factor for large scale nvfp4 EP

wenscarl:
## Motivation
Part of https://github.com/sgl-project/sglang/pull/12866(reverted in this PR), the root cause is that `w13[2]_input_scale` should not depends on `logical_to_all_physical_map`. In this fix, read in all physical experts' input scale regardless of its logical expert id.
cc @kaixih @Fridge003 

The root cause is:
The `w13_input_scale` is of shape [288, ...] but if the map `logical_to_all_physical_map` is :
```
[
[286],
[287],
[2],
[3],
...,
[254],
[255]
]
```
there 0th and 1st row is missing and since `w13_input_scale` is initialized by `torch.empty`, the corresponding values are undefined.
## Checklist

### DIFF
--- a/python/sglang/srt/eplb/expert_location.py
+++ b/python/sglang/srt/eplb/expert_location.py
@@ -284,9 +284,17 @@ class ExpertLocationMetadata:
     # -------------------------------- usage ------------------------------------
 
     def logical_to_all_physical(
-        self, layer_id: int, logical_expert_id: int
+        self,
+        layer_id: int,
+        logical_expert_id: int,
+        require_global_experts: bool = False,
     ) -> List[int]:
         # Use CPU copy to avoid GPU→CPU sync on every call, which is expensive in update weights scenario
+        if require_global_experts:
+            num_physical_experts = self.logical_to_all_physical_map_cpu[layer_id].shape[
+                -1
+            ]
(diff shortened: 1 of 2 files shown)


kaixih (approved):
Just had a sync with @wenscarl.

The root cause of the NaN issue wasn’t that the loader failed to map physical to logical experts correctly; rather, it was that the mapping must include all physical expert indices when `sglang_require_global_experts=True`. This is because each layer’s w13_input_scale is allocated based on the total number of physical experts.

For example, if there are 256 logical experts and 2 redundant experts, an incorrect logical→physical map might look like:
```
logical -> physical
0 -> 256
1 -> 257
2 -> 2
...
255 -> 255
```
The input_scales tensor would have shape (258,), and it then would look like this after loading:
```
physical
0 -> loading nothing (wrong!)
1 -> loading nothing (wrong!)
2 -> loading logical 2
...
255 -> loading logical 255
256 -> loading logical 0
257 -> loading logical 1
```
After this PR, we ensure that all 258 physical experts correctly load something, which resolves the NaN issue.

Fridge003:
@kaixih @wenscarl Have you verified how this PR works on GB200?

Fridge003 (approved):

PR #13166, opened 2025-11-12
[Models] Add NVIDIA Eagle 2.5 VLM

jonahbernard:
## Motivation

Add support for `nvidia/Eagle2.5-8B`

## Modifications

- Add processor for `nvidia/Eagle2.5-8B`
- Add architecture support for `nvidia/Eagle2.5-8B`
- Add to the nightly VLM mmmu eval test file

## Accuracy Tests

Not applicable.

## Benchmarking and Profiling

Not applicable.

## Checklist


test/srt/nightly/test_vlms_mmmu_eval.py
```diff
@@ -46,6 +46,9 @@
     ): ModelEvalMetrics(0.310, 16.7),
     ModelLaunchSettings("XiaomiMiMo/MiMo-VL-7B-RL"): ModelEvalMetrics(0.28, 32.0),
     ModelLaunchSettings("zai-org/GLM-4.1V-9B-Thinking"): ModelEvalMetrics(0.280, 30.4),
+    ModelLaunchSettings("nvidia/Eagle2.5-8B", extra_args=["--tp=2"]): ModelEvalMetrics(
```
JustinTong0323: This CI would run in single GPU, so you should not add tp=2 here. Could you help attach your local test result?

python/sglang/srt/multimodal/processors/image_processing_eagle2_5_vl_fast.py
JustinTong0323, line 1: We don't need it. Ideally sglang would call image processor from transformers.

python/sglang/srt/multimodal/processors/eagle2_5_vl.py
JustinTong0323, line 1: Ideally only `process_mm_data_async` is needed. Check whether there is a double resize.

python/sglang/srt/configs/eagle2_5_vl.py
JustinTong0323, line 1: If we could read the config from transformers, we don't need it.

test/srt/nightly/test_vlms_mmmu_eval.py
jonahbernard: I don't have access to a GPU that can run it. Can we get the results from the first nightly run?

python/sglang/srt/multimodal/processors/eagle2_5_vl.py
jonahbernard: What is the recommended path if the HF processor doesn't work with decode.video_reader objects? We need to call a custom preprocessing function before calling `process_and_combine_mm_data` right?

JustinTong0323 (approved):
LGTM

yuan-luo (approved):

PR #13186, opened 2025-11-13
remove deprecated `tile_tokens_dim`

b8zhong:
Remove `tile_tokens_dim` in `trtllm_fp8_block_scale_moe`, it can be calculated internally.

```
2025-11-13 02:14:16,121 - WARNING - core.py:1788 - flashinfer.jit: tile_tokens_dim in trtllm_fp8_block_scale_moe is planned for deprecation in a future release. Please remove it from your code as tile_tokens_dim will no longer be supported after v0.5.0.
```

### DIFF
--- a/python/sglang/srt/layers/quantization/fp8.py
+++ b/python/sglang/srt/layers/quantization/fp8.py
@@ -1256,9 +1256,7 @@ class Fp8MoEMethod(FusedMoEMethodBase):
                 routed_scaling_factor=(
                     routed_scaling_factor if routed_scaling_factor is not None else 1.0
                 ),
-                tile_tokens_dim=get_tile_tokens_dim(
-                    x.shape[0], topk_config.top_k, layer.num_experts
-                ),
+                tile_tokens_dim=None,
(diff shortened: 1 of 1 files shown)


FlamingoPg (approved):

python/sglang/srt/layers/quantization/fp8.py
```diff
@@ -1256,9 +1256,7 @@ def apply_with_router_logits(
                 routed_scaling_factor=(
                     routed_scaling_factor if routed_scaling_factor is not None else 1.0
                 ),
-                tile_tokens_dim=get_tile_tokens_dim(
```
Fridge003, line 1259: The `get_tile_tokens_dim` util function can also be deleted

Fridge003 (changes requested):

python/sglang/srt/layers/quantization/fp8.py
b8zhong: @Fridge003 Done 👍

Fridge003 (approved):

PR #13187, opened 2025-11-13
[feat] make warmup timeout configurable

billishyahao:
## Motivation

Remove the hardcode timeout during warming up by introducing
```
parser.add_argument(
        "--warmup-timeout",
        type=float,
        default=ServerArgs.warmup_timeout,
        help="Set warmup timeout in seconds. If a warmup forward batch takes longer than this, the server will crash to prevent hanging. Recommend to increase warmup timeout to 1800 to accommodate some kernel JIT precache e.g. deep gemm",
    )
```

## Checklist

### DIFF
--- a/python/sglang/srt/entrypoints/http_server.py
+++ b/python/sglang/srt/entrypoints/http_server.py
@@ -1529,11 +1529,16 @@ def _execute_server_warmup(
 
     try:
         if server_args.disaggregation_mode == "null":
+            logger.info(f"Start of co-locate warmup ...")
             res = requests.post(
                 url + request_name,
                 json=json_data,
                 headers=headers,
-                timeout=600,
+                timeout=(
+                    server_args.warmup_timeout
+                    if server_args.warmup_timeout is not None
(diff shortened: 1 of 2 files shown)


ShangmingCai (commented):
Changes are clean. But I think maybe we can bypass timeout caused by deepgemm compile through `--skip-server-warmup` already?

hnyls2002:
This configuration is rarely used. Do not add it to `ServerArgs` as too many args for now. Put this into `sglang.srt.environ` and use an environment variable to configure the server warmup timeout.

billishyahao:
> This configuration is rarely used. Do not add it to `ServerArgs` as too many args for now. Put this into `sglang.srt.environ` and use an environment variable to configure the server warmup timeout.

Thanks for the advice. Let me propose  a new PR to introduce new env variable in  `sglang.srt.environ`  rather than `ServerArgs` . Also we have observed random warmup timeout at the server side first time due to gemm library AOT.

Issue #10174, opened 2025-09-08
[Feature] Fully support DP load balance for PD-Disaggregation mode.


hnyls2002:
In the current implementation of PD-Disagggregation, the decode server should know the request's prefill DP rank in bootstrapping. However, if we can only decide the dp rank in the `dp_controller` module, then in the current design, the decode server cannot get the prefill request's dp rank.

There are three methods to fix the issue:
- Change `route` API for bootstrap server, use `bootstrap_room` as the identifier instead of
```python
url = f"http://{self.bootstrap_addr}/route?engine_rank={engine_rank}&target_dp_group={target_dp_group}"
```
To achieve this, we have to solve the `PUT` and `GET` orders for prefill to register the info and decode to fetch the info.

- After `dp_controller` has determined the dp rank, notify the decode server.
- Remove `dp_controller`'s load balance functionality completely, every dp rank is decided by the external router.

hnyls2002:
Also, I am not sure about the supporting status for system-dp and attention-dp of PD-disaggregation. Maybe related to #8530

hnyls2002:
For now, we can still use `round-robin` for prefill and other strategies for decode. I introduced `--prefill-round-robin-balance` to manually indicate that the user should set `--load-balance-method round_robin` for the prefill instance. #10169

ping1jing2:
Hi @hnyls2002 we implement the third way, please help to review our code. #13208 

We also attempt others but failed, here is our conclusion:
Attempt 1: `dp_controller` forwards messages received from the KVReceiver to the prefill DP rank. Specifically, in the bootstrap server, when receiving a /router? request from the KVReceiver, we uniformly return the bootstrapInfo to the prefill `dp_controller`. The `dp_controller` then independently schedules the request based on the load situation and records which dp_rank the request is scheduled to. At the same time, after receiving a message from the KVReceiver, the dp_controller forwards the message to the actual dp_rank that will process the request based on the recorded scheduling information. However, during performance testing, it was found that forwarding messages twice caused some data backlog, and the situation worsened as the batch size increased, with some requests experiencing delays of up to nearly 1 second. Based on these observations, we ultimately abandoned this solution.

PR #13208, opened 2025-11-13
router select dp group with the minimum number of tokens

jiashaokun-1:
## Motivation

In the current PD separation scenario, the prefill instances only support the round_robin scheduling policy. Under this scheduling policy, which DP group a request is routed to is determined by the result of `bootstrap_room` mod `dpsize`. This can lead to load imbalance among DP groups, and this imbalance becomes more pronounced when the input requests are variable-length sequences.
Based on the above situation, I have added a new feature to select the DP group with the lightest load to process requests, thereby achieving and supporting DP load balance for PD-Disaggregation mode. The current load is measured by the number of tokens, which can be adjusted as needed in the future.The enabling or disabling of this feature is controlled by the parameter `dp_minimum_tokens_scheduler`.
### 1.Load Collection
`LoadMonitor` periodically calls the `/get_load` interface to collect load data. In the existing implementation, only the sum number of tokens for all dp group was recorded. After modification, it will now record the number of tokens for each `dp_rank`, i.e., the load.
```rust
// Each object has: {dp_rank, num_reqs, num_waiting_reqs, num_tokens}
let mut rank_tokens = HashMap::new();
for entry in array {
    let dp_rank = entry.get("dp_rank").and_then(|v| v.as_i64()).map(|rank| rank as isize);
    let num_tokens = entry.get("num_tokens").and_then(|v| v.as_i64()).map(|rank| rank as isize);
    if let (Some(rank), Some(tokens)) = (dp_rank, num_tokens) {
        rank_tokens.insert(rank, tokens);
    }
}
```
### 2. Load Management
A new `DPLoadManager` struct has been added to manage loads, providing three methods:  `update_dp_loads` ,`get_lowest_dp_load` , and `load_increment`.
```rust
#[derive(Debug, Default)]
pub struct DPLoadManager {
    dp_cached_loads: RwLock<HashMap<String, HashMap<isize, isize>>>,
}
```
1. The `update_dp_loads` method updates the load. After periodically collecting load data, the LoadMonitor calls this method to perform the update.
2. The `get_lowest_dp_load` method takes a worker as input and returns the dp_rank with the lowest load among the workers.
3. The `load_increment` method is used to add an increment to the load of a specific dp_rank for a worker. This is done to prevent all requests from being scheduled to the same DP group during the interval between two load reports.
### 3. Selection of DP Group
1. Call the `get_lowest_dp_load` method to obtain the DP group with the lowest load.
2. Insert the `data_parallel_rank` field into the request so that the `data_parallel_controller` in prefill can check this field is not empty and directly schedule the request to the corresponding worker. Similarly, in the docode phase, check that the `data_parallel_rank` field is not empty, assign it to `prefill_dp_rank`, and ensure that KVReceiver and KVSender can connect properly (the implementation in the engine already supports this).
3. Call the `load_increment` method to add the length of the text in the request to the load of the DP group.
```rust
let lowest_prefill_dp_rank = prefill_policy.get_lowest_dp_load(prefill_worker);
obj.insert(
    "data_parallel_rank".to_string(),
    match lowest_prefill_dp_rank {
        Some(v) => Value::from(v),
        None => Value::Null,
    },
);
if let Some(dp_rank) = lowest_prefill_dp_rank {
    prefill_policy.load_increment(prefill_worker, dp_rank, prompt_len );
}
```
### 4. Configuration Parameters
1. A new configuration parameter `dp-minimum-tokens-scheduler` has been added to enable scheduling of the minimum load DP group when declaring the parameter.
2. A new configuration parameter `worker-load-check-interval` has been added to specify the interval for load collection. Previously, the load collection interval reused the configuration of `worker-startup-check-interval`. Now, this new configuration item is separate from the startup check.
## Benchmarking and Profiling

Comparing the performance gains of several scheduling strategies on variable-length datasets before and after enabling dp-minimum-tokens-scheduler, the Mean TPOP performance improved by approximately 9%, and the Mean TTFT (with max_out_len=1 set to eliminate the impact of decoding on prefill) performance improved by about 15%. The specific data is as follows:
| Policy      | Mean  TTFT(ms) max_out_len=1 |                           |             | Mean  TPOT (ms)|                            |             |
| ----------- | ---------------------------- | ------------------------- | ----------- | ---------- | -------------------------- | ----------- |
|             | benckmark                    | enable dp minimum  tokens | Performance | benckmark  | enable dp minimum  tokense | Performance |
| Random      | 2631                         | 2249                      | +14.52%      | 53.41      | 48.15                      | +9.85%       |
| Round_robin | 2646                         | 2164                      | +18.22%       | 55.42      | 49.02                      | +11.55%      |
| Cache_aware | 2601                         | 2216                      | +14.80%      | 52.67      | 48.48                      | +7.96%       |

router
```shell
python -m sglang_router.launch_router --pd-disaggregation --prefill http://141.61.105.141:8000 8998 --decode http://141.61.105.144:8001 --prefill-policy cache_aware --dp-minimum-tokens-scheduler --worker-load-check-interval 1
```

prefill
```shell
python -m sglang.launch_server --model-path ${MODEL_PATH}  --disaggregation-mode prefill --host ${P_IP[$i]} \
--port 8000 --disaggregation-bootstrap-port $((8998+$i)) --trust-remote-code --nnodes 1 --node-rank 0 \
--tp-size 16 --mem-fraction-static 0.6 --attention-backend ascend --device npu --quantization w8a8_int8 \
--disaggregation-transfer-backend ascend --max-running-requests 8 --context-length 8192  --disable-radix-cache \
--chunked-prefill-size 32768 --max-prefill-tokens 28680 --disable-overlap-schedule --moe-a2a-backend deepep --deepep-mode normal \
--speculative-algorithm NEXTN --speculative-num-steps 3 --speculative-eagle-topk 1 --speculative-num-draft-tokens 4  \
--dp-size 2 --enable-dp-attention --disable-shared-experts-fusion --dtype bfloat16
```
decode
```shell
python -m sglang.launch_server --model-path ${MODEL_PATH} --disaggregation-mode decode --host ${D_IP[$i]} \
--port 8001 --trust-remote-code --nnodes 1 --node-rank 0 --tp-size 16 --dp-size 8 \
--mem-fraction-static 0.76 --max-running-requests 256 --attention-backend ascend --device npu --quantization w8a8_int8 \
--moe-a2a-backend deepep --enable-dp-attention --deepep-mode low_latency --enable-dp-lm-head \
--cuda-graph-bs 8 16 24 32 --disaggregation-transfer-backend ascend --watchdog-timeout 9000 --context-length 8192 \
--speculative-algorithm NEXTN --speculative-num-steps 3 --speculative-eagle-topk 1 --speculative-num-draft-tokens 4  \
--prefill-round-robin-balance --disable-shared-experts-fusion --dtype bfloat16
```
benchmark
The model evaluation tool I used is AISBench. Click the link below to learn more: [AISBench](https://gitee.com/aisbench/benchmark/blob/master/README_en.md). Since I was using a variable-length dataset, I ran the dataset through sglang.bench_serving with the parameter --dataset-name random. However, I noticed that the input sequence length was being split into random values limited by the --random-input-len parameter. To ensure that the input request sequences were not split, I used AISBench.
```bash
ais_bench --models vllm_api_stream_chat --datasets gsm8k_gen_0_shot_cot_str_perf  --debug --summarizer default_perf --mode perf

╒══════════════════════════╤═════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╤═════════════════╤══════╕
│ Performance Parameters   │ Stage   │ Average         │ Min             │ Max             │ Median          │ P75             │ P90             │ P99             │  N   │
╞══════════════════════════╪═════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╪═════════════════╪══════╡
│ E2EL                     │ total   │ 55044.4573 ms   │ 11143.2514 ms   │ 195472.4726 ms  │ 35057.9889 ms   │ 69464.8019 ms   │ 132295.1279 ms  │ 185509.4833 ms  │ 1000 │
├──────────────────────────┼─────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┼──────┤
│ TTFT                     │ total   │ 893.5725 ms     │ 348.2403 ms     │ 1549.0416 ms    │ 901.9457 ms     │ 1084.3782 ms    │ 1215.7292 ms    │ 1469.2704 ms    │ 1000 │
├──────────────────────────┼─────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┼──────┤
│ TPOT                     │ total   │ 48.4854 ms      │ 33.7401 ms      │ 63.8094 ms      │ 48.7747 ms      │ 52.4151 ms      │ 55.4097 ms      │ 59.9007 ms      │ 1000 │
├──────────────────────────┼─────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┼──────┤
│ ITL                      │ total   │ 114.1014 ms     │ 0.0068 ms       │ 594.2452 ms     │ 115.7806 ms     │ 119.678 ms      │ 123.1585 ms     │ 251.8668 ms     │ 1000 │
├──────────────────────────┼─────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┼──────┤
│ InputTokens              │ total   │ 2560.5243       │ 1481.0          │ 4120.0          │ 2216.0          │ 3402.25         │ 3943.3          │ 4107.49         │ 1000 │
├──────────────────────────┼─────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┼──────┤
│ OutputTokens             │ total   │ 1159.8058       │ 187.0           │ 4068.0          │ 715.5           │ 1358.0          │ 2769.4          │ 4068.0          │ 1000 │
├──────────────────────────┼─────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┼─────────────────┼──────┤
│ OutputTokenThroughput    │ total   │ 20.3425 token/s │ 14.7171 token/s │ 28.1505 token/s │ 19.9678 token/s │ 22.0362 token/s │ 23.6102 token/s │ 26.6782 token/s │ 1000 │
╘══════════════════════════╧═════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╧═════════════════╧══════╛
╒══════════════════════════╤═════════╤═══════════════════╕
│ Common Metric            │ Stage   │ Value             │
╞══════════════════════════╪═════════╪═══════════════════╡
│ Benchmark Duration       │ total   │ 392829.9277 ms    │
├──────────────────────────┼─────────┼───────────────────┤
│ Total Requests           │ total   │ 1000              │
├──────────────────────────┼─────────┼───────────────────┤
│ Failed Requests          │ total   │ 0                 │
├──────────────────────────┼─────────┼───────────────────┤
│ Success Requests         │ total   │ 1000              │
├──────────────────────────┼─────────┼───────────────────┤
│ Concurrency              │ total   │ 86.5959           │
├──────────────────────────┼─────────┼───────────────────┤
│ Max Concurrency          │ total   │ 270               │
├──────────────────────────┼─────────┼───────────────────┤
│ Request Throughput       │ total   │ 2.5456 req/s      │
├──────────────────────────┼─────────┼───────────────────┤
│ Total Input Tokens       │ total   │ 1582404           │
├──────────────────────────┼─────────┼───────────────────┤
│ Prefill Token Throughput │ total   │ 2865.4913 token/s │
├──────────────────────────┼─────────┼───────────────────┤
│ Total generated tokens   │ total   │ 716760            │
├──────────────────────────┼─────────┼───────────────────┤
│ Input Token Throughput   │ total   │ 4028.2165 token/s │
├──────────────────────────┼─────────┼───────────────────┤
│ Output Token Throughput  │ total   │ 1824.6064 token/s │
├──────────────────────────┼─────────┼───────────────────┤
│ Total Token Throughput   │ total   │ 5852.8229 token/s │
╘══════════════════════════╧═════════╧═══════════════════╛

```

## Checklist

### DIFF
--- a/python/sglang/srt/managers/data_parallel_controller.py
+++ b/python/sglang/srt/managers/data_parallel_controller.py
@@ -498,11 +498,28 @@ class DataParallelController:
         self.max_req_input_len = scheduler_info[0]["max_req_input_len"]
 
     def maybe_external_dp_rank_routing(self, req: Req):
-        if req.data_parallel_rank is not None:
-            logger.debug(f"Direct routing to DP rank {req.data_parallel_rank}")
-            self.workers[req.data_parallel_rank].send_pyobj(req)
-            return True
-        return False
+        if self.server_args.disaggregation_mode == "prefill":
+            if req.data_parallel_rank is not None:
+                logger.debug(
+                    f"Prefill direct routing to DP rank {req.data_parallel_rank}"
+                )
+                self.workers[req.data_parallel_rank].send_pyobj(req)
+                return True
+            return False
+        else:
+            if req.data_parallel_rank_decode is not None:
+                logger.debug(
+                    f"Decode direct routing to DP rank {req.data_parallel_rank_decode}"
+                )
+                self.workers[req.data_parallel_rank_decode].send_pyobj(req)
+                return True
+            if req.data_parallel_rank is not None:
+                logger.debug(
+                    f"Decode direct routing to DP rank {req.data_parallel_rank}, by data parallel rank"
+                )
+                self.workers[req.data_parallel_rank].send_pyobj(req)
+                return True
+            return False
 
     def round_robin_scheduler(self, req: Req):
         if self.maybe_external_dp_rank_routing(req):
--- a/python/sglang/srt/managers/io_struct.py
+++ b/python/sglang/srt/managers/io_struct.py
@@ -231,6 +231,11 @@ class GenerateReqInput(BaseReq, APIServingTimingMixin):
     # For data parallel rank routing
     data_parallel_rank: Optional[int] = None
 
+    # In PD-Disaggregation mode, requests are scheduled to different DP group based on the load conditions of prefill and decode.
+    # `data_parallel_rank_decode` indicates which DP group the current request need to be dispatched to for decode
+    # `data_parallel_rank` indicates which DP group the request will be scheduled to on the prefill.
+    data_parallel_rank_decode: Optional[int] = None
+
     # For background responses (OpenAI responses API)
     background: bool = False
 
@@ -655,6 +660,11 @@ class GenerateReqInput(BaseReq, APIServingTimingMixin):
             data_parallel_rank=(
                 self.data_parallel_rank if self.data_parallel_rank is not None else None
             ),
+            data_parallel_rank_decode=(
+                self.data_parallel_rank_decode
+                if self.data_parallel_rank_decode is not None
+                else None
+            ),
             conversation_id=self.conversation_id,
             priority=self.priority,
             extra_key=self.extra_key,
@@ -723,6 +733,8 @@ class TokenizedGenerateReqInput(BaseReq):
 
     # For data parallel rank routing
     data_parallel_rank: Optional[int] = None
+
+    data_parallel_rank_decode: Optional[int] = None
 
     # Priority for the request
     priority: Optional[int] = None
--- a/sgl-model-gateway/bindings/python/sglang_router/router_args.py
+++ b/sgl-model-gateway/bindings/python/sglang_router/router_args.py
@@ -359,6 +359,12 @@ class RouterArgs:
             default=RouterArgs.worker_startup_check_interval,
             help="Interval in seconds between checks for worker startup",
         )
+        pd_group.add_argument(
+            f"--{prefix}worker-load-check-interval",
+            type=int,
+            default=RouterArgs.worker_load_check_interval,
+            help="Interval in seconds between checks for worker load",
+        )
 
         # Logging configuration
         logging_group.add_argument(
--- a/sgl-model-gateway/bindings/python/sglang_router/router_args.py
+++ b/sgl-model-gateway/bindings/python/sglang_router/router_args.py
@@ -30,6 +30,7 @@ class RouterArgs:
     decode_policy: Optional[str] = None  # Specific policy for decode nodes in PD mode
     worker_startup_timeout_secs: int = 1800
     worker_startup_check_interval: int = 30
+    worker_load_check_interval: int = 10
     cache_threshold: float = 0.3
     balance_abs_threshold: int = 64
     balance_rel_threshold: float = 1.5
@@ -38,6 +39,7 @@ class RouterArgs:
     max_payload_size: int = 512 * 1024 * 1024  # 512MB default for large batches
     bucket_adjust_interval_secs: int = 5
     dp_aware: bool = False
+    dp_minimum_tokens_scheduler: bool = False
     enable_igw: bool = False  # Enable IGW (Inter-Gateway) mode for multi-model support
     api_key: Optional[str] = None
     log_dir: Optional[str] = None
@@ -307,6 +309,11 @@ class RouterArgs:
             f"--{prefix}dp-aware",
             action="store_true",
             help="Enable data parallelism aware schedule",
+        )
+        routing_group.add_argument(
+            f"--{prefix}dp-minimum-tokens-scheduler",
+            action="store_true",
+            help="Enable minimum tokens scheduler for data parallel group",
         )
         routing_group.add_argument(
             f"--{prefix}enable-igw",
--- a/sgl-model-gateway/py_test/integration_mock/load_balancing/test_power_of_two.py
+++ b/sgl-model-gateway/py_test/integration_mock/load_balancing/test_power_of_two.py
@@ -22,7 +22,7 @@ def test_power_of_two_prefers_less_loaded(mock_workers, router_manager):
     rh = router_manager.start_router(
         worker_urls=urls,
         policy="power_of_two",
-        extra={"worker_startup_check_interval": 1},
+        extra={"worker_load_check_interval": 1},
     )
 
     # Prime: fire a burst to create measurable load on slow worker, then wait for monitor tick
--- a/sgl-model-gateway/src/core/worker_manager.rs
+++ b/sgl-model-gateway/src/core/worker_manager.rs
@@ -21,7 +21,7 @@ use tokio::{
 use tracing::{debug, info, warn};
 
 use crate::{
-    core::{metrics_aggregator::MetricPack, ConnectionMode, Worker, WorkerRegistry, WorkerType},
+    core::{metrics_aggregator::MetricPack, ConnectionMode, Worker, WorkerLoadManager, WorkerRegistry, WorkerType},
     policies::PolicyRegistry,
     protocols::worker_spec::{FlushCacheResult, WorkerLoadInfo, WorkerLoadsResult},
 };
@@ -221,15 +221,23 @@ impl WorkerManager {
                 let client = client.clone();
 
                 async move {
-                    let load = if is_http {
+                    let dp_rank_loads = if is_http {
                         Self::parse_load_response(&client, &url, api_key.as_deref()).await
+                    } else {
+                        HashMap::new()
+                    };
+
+                    let load = if !dp_rank_loads.is_empty() {
+                        dp_rank_loads.values().sum::<isize>()
                     } else {
                         -1
                     };
+
                     WorkerLoadInfo {
                         worker: url,
                         worker_type,
                         load,
+                        dp_rank_loads,
                     }
                 }
             })
@@ -251,7 +259,7 @@ impl WorkerManager {
         client: &reqwest::Client,
         url: &str,
         api_key: Option<&str>,
-    ) -> isize {
+    ) -> HashMap<isize, isize> {
         let load_url = format!("{}/get_load", url);
         let mut req = client.get(&load_url).timeout(REQUEST_TIMEOUT);
         if let Some(key) = api_key {
@@ -260,15 +268,25 @@ impl WorkerManager {
 
         match req.send().await {
             Ok(r) if r.status().is_success() => match r.json::<Value>().await {
-                Ok(json) if json.is_array() => json
-                    .as_array()
-                    .unwrap()
-                    .iter()
-                    .filter_map(|e| e.get("num_tokens").and_then(|v| v.as_i64()))
-                    .sum::<i64>() as isize,
-                _ => -1,
+                Ok(json) if json.is_array() => {
+                    let mut load_map = HashMap::new();
+
+                    for element in json.as_array().unwrap().iter() {
+                        if let (Some(dp_rank_value), Some(num_tokens_value)) =
+                            (element.get("dp_rank"), element.get("num_tokens"))
+                        {
+                            if let (Some(dp_rank), Some(num_tokens)) =
+                                (dp_rank_value.as_i64(), num_tokens_value.as_i64())
+                            {
+                                load_map.insert(dp_rank as isize, num_tokens as isize);
+                            }
+                        }
+                    }
+                    load_map
+                }
+                _ => HashMap::new(),
             },
-            _ => -1,
+            _ => HashMap::new(),
         }
     }
 
@@ -313,6 +331,7 @@ impl WorkerManager {
 pub struct LoadMonitor {
     worker_registry: Arc<WorkerRegistry>,
     policy_registry: Arc<PolicyRegistry>,
+    pub worker_load_manager: Arc<WorkerLoadManager>,
     client: reqwest::Client,
     interval: Duration,
     tx: watch::Sender<HashMap<String, isize>>,
@@ -332,6 +351,7 @@ impl LoadMonitor {
         Self {
             worker_registry,
             policy_registry,
+            worker_load_manager: Arc::new(WorkerLoadManager::new()),
             client,
             interval: Duration::from_secs(interval_secs),
             tx,
@@ -354,12 +374,13 @@ impl LoadMonitor {
 
         let worker_registry = Arc::clone(&self.worker_registry);
         let policy_registry = Arc::clone(&self.policy_registry);
+        let worker_load_manager = Arc::clone(&self.worker_load_manager);
         let client = self.client.clone();
         let interval = self.interval;
         let tx = self.tx.clone();
 
         let handle = tokio::spawn(async move {
-            Self::monitor_loop(worker_registry, policy_registry, client, interval, tx).await;
+            Self::monitor_loop(worker_registry, policy_registry, worker_load_manager, client, interval, tx).await;
         });
 
         *handle_guard = Some(handle);
@@ -381,6 +402,7 @@ impl LoadMonitor {
     async fn monitor_loop(
         worker_registry: Arc<WorkerRegistry>,
         policy_registry: Arc<PolicyRegistry>,
+        worker_load_manager: Arc<WorkerLoadManager>,
         client: reqwest::Client,
         interval: Duration,
         tx: watch::Sender<HashMap<String, isize>>,
@@ -389,10 +411,11 @@ impl LoadMonitor {
 
         loop {
             interval_timer.tick().await;
-
             let power_of_two_policies = policy_registry.get_all_power_of_two_policies();
 
-            if power_of_two_policies.is_empty() {
+            if power_of_two_policies.is_empty()
+                && !policy_registry.is_dp_minimum_tokens_scheduler_enabled()
+            {
                 debug!("No PowerOfTwo policies found, skipping load fetch");
                 continue;
             }
@@ -400,8 +423,10 @@ impl LoadMonitor {
             let result = WorkerManager::get_all_worker_loads(&worker_registry, &client).await;
 
             let mut loads = HashMap::new();
+            let mut dp_rank_loads = HashMap::new();
             for load_info in result.loads {
-                loads.insert(load_info.worker, load_info.load);
+                loads.insert(load_info.worker.clone(), load_info.load);
+                dp_rank_loads.insert(load_info.worker, load_info.dp_rank_loads);
             }
 
             if !loads.is_empty() {
@@ -413,6 +438,7 @@ impl LoadMonitor {
                 for policy in &power_of_two_policies {
                     policy.update_loads(&loads);
                 }
+                worker_load_manager.update_dp_loads(&dp_rank_loads);
                 let _ = tx.send(loads);
             } else {
                 warn!("No loads fetched from workers");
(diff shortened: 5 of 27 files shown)


python/sglang/srt/managers/data_parallel_controller.py
```diff
@@ -454,11 +454,22 @@ def launch_tensor_parallel_group(
...
+                self.workers[req.data_parallel_rank].send_pyobj(req)
+                return True
+            return False
+        else:
+            if req.decode_dp_rank is not None:
+                logger.debug(f"Decode direct routing to DP rank {req.decode_dp_rank}")
+                self.workers[req.decode_dp_rank].send_pyobj(req)
+                return True
+            if req.data_parallel_rank is not None:
+                (f"Decode direct routing to DP rank {req.decode_dp_rank}, by data parallel rank")
```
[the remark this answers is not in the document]
jiashaokun-1: done

sgl-router/src/app_context.rs
```diff
@@ -426,6 +426,13 @@ impl AppContextBuilder {
     /// Create policy registry
     fn with_policy_registry(mut self, config: &RouterConfig) -> Self {
         self.policy_registry = Some(Arc::new(PolicyRegistry::new(config.policy.clone())));
+        if config.dp_minimum_tokens_scheduler {
+            if let Some(policy_registry) = &self.policy_registry {
+                policy_registry.enable_dp_minimum_tokens_scheduler();
+            } else {
+                info!("policy_registry is None")
+            }
```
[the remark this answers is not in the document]
jiashaokun-1: done

python/sglang/srt/managers/io_struct.py
```diff
@@ -620,6 +622,9 @@ def __getitem__(self, i):
             data_parallel_rank=(
                 self.data_parallel_rank if self.data_parallel_rank is not None else None
             ),
+            decode_dp_rank=(
+                self.decode_dp_rank if self.decode_dp_rank is not None else None
+            ),
```
jimmy-evo, line 667: I believe this is redundant, as `data_parallel_rank` suffices and is designed for the router. In the current implementation, the router selects Prefill and Decode workers using `data_parallel_rank` (which differs for Prefill and Decode)

`decode_dp_rank` or `prefill_dp_rank` is unnecessary.

jimmy-evo (changes requested):

python/sglang/srt/managers/io_struct.py
jiashaokun-1: At first, I only used `data_parallel_rank`, but when running the code, I encountered an error: the KVTransfer in prefill timed out and failed, and the KVSender and KVReceiver were not properly connected. This is because requests in prefill and decode might be scheduled to different dp_ranks. In the decode phase, there needs to be a parameter indicating which dp_rank the request should be scheduled to, and also an additional parameter indicating the dp_rank where the request was originally in prefill. Only then can the KVSender and KVReceiver be correctly connected. Therefore, I added a new parameter.

jimmy-evo: > At first, I only used `data_parallel_rank`, but when running the code, I encountered an error: the KVTransfer in prefill timed out and failed, and the KVSender and KVReceiver were not properly connected. This is because requests in prefill and decode might be scheduled to different dp_ranks. In the decode phase, there needs to be a parameter indicating which dp_rank the request should be scheduled to, and also an additional parameter indicating the dp_rank where the request was originally in prefill. Only then can the KVSender and KVReceiver be correctly connected. Therefore, I added a new parameter.

i think this naming is better.:
`data_parallel_rank` for *prefill&null*,  `data_parallel_rank_decode` for *decode*
and add sufficient comment

jiashaokun-1: > > At first, I only used `data_parallel_rank`, but when running the code, I encountered an error: the KVTransfer in prefill timed out and failed, and the KVSender and KVReceiver were not properly connected. This is because requests in prefill and decode might be scheduled to different dp_ranks. In the decode phase, there needs to be a parameter indicating which dp_rank the request should be scheduled to, and also an additional parameter indicating the dp_rank where the request was originally in prefill. Only then can the KVSender and KVReceiver be correctly connected. Therefore, I added a new parameter.
> 
> i think this naming is better.: `data_parallel_rank` for _prefill&null_, `data_parallel_rank_decode` for _decode_ and add sufficient comment

You're right, I made the changes according to your suggestion.

python/sglang/srt/managers/io_struct.py
```diff
@@ -203,6 +203,8 @@ class GenerateReqInput(BaseReq):
     # For data parallel rank routing
     data_parallel_rank: Optional[int] = None
 
+    data_parallel_rank_decode: Optional[int] = None
```
merrymercy, line 237: document clearly why you need two data parallel rank here

merrymercy (changes requested):

python/sglang/srt/managers/io_struct.py
jiashaokun-1: Fixed

sgl-model-gateway/src/routers/http/pd_router.rs
```diff
@@ -688,6 +706,59 @@ impl PDRouter {
...
+        obj.insert(
+            "data_parallel_rank".to_string(),
+            match lowest_prefill_dp_rank {
+                Some(v) => Value::from(v),
+                None => Value::Null,
+            },
+        );
+        // During the prefill and decode stages, requests may be scheduled to different dp_rank
+        // data_parallel_rank_decode specifies which dp_rank the request should be scheduled to for decode
+        // data_parallel_rank specifies which dp_rank the request is in for preill
```
jimmy-evo: `preill` -> `prefill`

sgl-model-gateway/src/routers/http/pd_router.rs
```diff
@@ -688,6 +706,59 @@ impl PDRouter {
...
+        obj.insert(
+            "data_parallel_rank_decode".to_string(),
+            match lowest_decode_dp_rank {
+                Some(v) => Value::from(v),
+                None => Value::Null,
+            },
+        );
+        let prompt_len = length
+            .try_into()
+            .map_err(|e| format!("Failed to convert length tp isize:{}", e))?;
```
jimmy-evo: convert length **tp** isize

`to` size

sgl-model-gateway/bindings/python/sglang_router/router_args.py
```diff
@@ -38,6 +39,7 @@ class RouterArgs:
     max_payload_size: int = 512 * 1024 * 1024  # 512MB default for large batches
     bucket_adjust_interval_secs: int = 5
     dp_aware: bool = False
+    dp_minimum_tokens_scheduler: bool = False
```
slin1237, line 42: we already have dp aware
so perhaps
from gateway perspective
if this is dp aware, this means
1. regardless of dp metrics, round robin will always be round robin
2. power of two will still be power of two (random two with least load)
3. cache aware simply dont care
4. bucket doesnt care either much
5. random does not care either

what you probably want is another new policy
1. least load

so this `dp_minimum_tokens_scheduler` can be removed
because then it will be least_load + dp_aware

sgl-model-gateway/py_test/integration_mock/load_balancing/test_power_of_two.py
```diff
@@ -22,7 +22,7 @@ def test_power_of_two_prefers_less_loaded(mock_workers, router_manager):
     rh = router_manager.start_router(
         worker_urls=urls,
         policy="power_of_two",
-        extra={"worker_startup_check_interval": 1},
+        extra={"worker_load_check_interval": 1},
```
slin1237, line 25: thanks for the fix

sgl-model-gateway/src/core/worker_manager.rs
```diff
@@ -221,15 +221,23 @@ impl WorkerManager {
                 let client = client.clone();
 
                 async move {
-                    let load = if is_http {
+                    let dp_rank_loads = if is_http {
```
slin1237, line 224: i think we should take this chance
instead of using `/get_load` which i dont think its built for scale and integration
these metrics should be properly exposed by `/metrics` endpoint
and gateway should scrape this from `/metrics`

sgl-model-gateway/src/policies/mod.rs
```diff
@@ -58,10 +64,22 @@ pub trait LoadBalancingPolicy: Send + Sync + Debug {
     /// Update worker load information
     ///
     /// This is called periodically with current load information for load-aware policies.
-    fn update_loads(&self, _loads: &std::collections::HashMap<String, isize>) {
+    fn update_loads(&self, _loads: &HashMap<String, isize>) {
```
slin1237: policy should not have those interfaces
worker load should happen as part of load monitor

sgl-model-gateway/src/policies/mod.rs
```diff
@@ -112,6 +130,51 @@ impl Default for BucketConfig {
     }
 }
 
+/// Configuration for cache-aware policy
+#[derive(Debug, Default)]
+pub struct DPLoadManager {
```
slin1237: this should be part of load monitoring, not policy crate
perhaps
better yet, have a worker_load.rs under core dedicated for those

sgl-model-gateway/src/policies/power_of_two.rs
```diff
@@ -19,12 +19,14 @@ use crate::core::Worker;
 pub struct PowerOfTwoPolicy {
     /// Cached load information from external monitoring
     cached_loads: RwLock<HashMap<String, isize>>,
+    dp_load_manager: DPLoadManager,
```
slin1237: remove

sgl-model-gateway/src/routers/http/pd_router.rs
```diff
@@ -688,6 +706,59 @@ impl PDRouter {
         prefill_policy.needs_request_text() || decode_policy.needs_request_text()
     }
 
+    async fn select_data_parallel_rank(
```
slin1237, line 720: its weird to do load balancing here
original idea was
each dp worker should be an worker, and one sglang worker can have multiple dp worker
this list of workers should be passed to policy, and policy handles the rest given string(payload), workers

slin1237 (changes requested):
we should change minimally to api spec because SGLang has deviated from OpenAI spec quite a bit
1. adding more for load balancing isn't great, if this is a must be done in order to force engine to route traffic to a specific dp rank, its fine? this should rather be a internal implementation detail instead of a public API, for example, this should be exposed at grpc protobuf
2. adding more custom implementation on top of get load is also not great, this function wasn't implemented for load balancing, the request and response are completely in free form right now, we should instead expose this via prom metrics endpoint so it has a standard

the essential problem for load balancing of today is lack of metrics
- carry metrics back in response header is one way to do that, this solves streaming case, you will always have the most real time metrics, and the cost of header is fairly low
- carry back system metrics, including DP via protobuf, this also solves streaming case
- none streaming case, because we don't know how long decode will really take, use real - engine metrics endpoint, instead of get load, metrics endpoint is built to scale and have api standard

sgl-model-gateway/src/routers/http/pd_router.rs
jiashaokun-1: Fixed

sgl-model-gateway/src/routers/http/pd_router.rs
jiashaokun-1: Fixed

sgl-model-gateway/src/policies/mod.rs
jiashaokun-1: Fixed

sgl-model-gateway/src/policies/mod.rs
jiashaokun-1: Fixed

sgl-model-gateway/src/policies/power_of_two.rs
jiashaokun-1: Fixed

sgl-model-gateway/bindings/python/sglang_router/router_args.py
jiashaokun-1: 1. I am not trying to implement a new scheduling strategy. Instead, I want to achieve load balancing among DP groups based on the existing round_robin, power_of_two, cacheAware, bucket, and random scheduling strategies. Currently, the scheduling of DP groups is mainly implemented on the DPController. As mentioned in issue #10174, decoding needs to be aware of the scheduling results of the prefill DP groups. If we were to implement load balancing for prefill DP groups on the DPController, it would require additional communication latency. After implementation and verification, I found that there was no performance benefit, so I chose option three: directly specifying the DP group on the gateway.

2. As you said, enabling dp_aware allows us to be aware of the DP groups, but the scheduling level is different. On one hand, using the least_load + dp_aware scheduling strategy does not necessarily perform better than the standalone CacheAware strategy, as the former is also a variation of power_of_two. On the other hand, the current --dp-aware option does not work properly. As mentioned in issue #15685, one way I can think of to solve this problem is to add a "data_parallel_rank_decode" field.

jiashaokun-1:
> we should change minimally to api spec because SGLang has deviated from OpenAI spec quite a bit
> 
> 1. adding more for load balancing isn't great, if this is a must be done in order to force engine to route traffic to a specific dp rank, its fine? this should rather be a internal implementation detail instead of a public API, for example, this should be exposed at grpc protobuf
> 2. adding more custom implementation on top of get load is also not great, this function wasn't implemented for load balancing, the request and response are completely in free form right now, we should instead expose this via prom metrics endpoint so it has a standard
> 
> the essential problem for load balancing of today is lack of metrics
> 
> * carry metrics back in response header is one way to do that, this solves streaming case, you will always have the most real time metrics, and the cost of header is fairly low
> * carry back system metrics, including DP via protobuf, this also solves streaming case
> * none streaming case, because we don't know how long decode will really take, use real - engine metrics endpoint, instead of get load, metrics endpoint is built to scale and have api standard

Could you please help evaluate whether this PR can be temporarily merged? I will complete the verification of the suggested implementations for gRPC protobuf and Prometheus metrics by January 30, 2026, and then submit the PR.

Issue #3266, opened 2025-02-03
[Feature] Add Model Hooks for Accessing and Customizing Model Activations


shuyhere:
### Checklist

- [x] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 2. Please use English, otherwise it will be closed.

## Description
It would be beneficial to introduce model hooks that allow users to access and modify model activations. This feature would enable greater flexibility for tasks such as visualization, debugging, and custom processing of intermediate representations.

## Use case
* Extract intermediate outputs for interpretability analysis, such as [LogitLens-style investigations](https://www.lesswrong.com/posts/AcKRB8wDpdaN6v6ru/interpreting-gpt-the-logit-lens).
* Expose internal activations, enabling users to cache activations and implement functions to edit, remove, or replace them dynamically during inference, for example [representation engineering](https://github.com/andyzoujm/representation-engineering).

While this may introduce some performance overhead, it would enhance interpretability research and enable efficient model editing.

## model hook resources
* [Pytorch hook](https://pytorch.org/docs/stable/generated/torch.Tensor.register_hook.html)
* [Transformerlens](https://github.com/TransformerLensOrg/TransformerLens)
## related issues and use case
* https://github.com/vllm-project/vllm/issues/4084
* https://github.com/vllm-project/vllm/issues/11397
* https://github.com/vllm-project/vllm/issues/8278

jalaln06:
Has this feature been added?

CurryxIaoHu:
Do we have this function now? Does anybody know how to do it?

YooSungHyun:
I'm very interested in KL divergence, so I need the logits used for training LLMs. Is there any way to access them?

MingningShao:
Why issue closed ? Dose this feature implemented? Seems no ???

PR #13217, opened 2025-11-13
Adding user defined hooks support

Carlomus:
## Motivation

This PR implements first-class support for attaching user-defined PyTorch forward hooks to SGLang models, fully configurable via `server_args` and the CLI.

The need for this functionality was previously raised in e.g. https://github.com/sgl-project/sglang/issues/3266, but it was never implemented. 

## Modifications

This PR introduces the following changes:

### **Core functionality**

* Adds `enable_hooks` and `hooks` fields to `ServerArgs`, allowing hook configuration via JSON or CLI flags.
* Integrates hook registration into `ModelRunner.initialize()`, ensuring hooks are attached during server start-up.
* Implements `ModelRunner.register_hooks`, which:

   * Resolves module name patterns using `fnmatch`.
   * Dynamically imports a user-provided hook factory via `resolve_callable`.
   * Installs forward hooks on all matched submodules.
   * Emits warnings for unmatched patterns without failing execution.

### **Documentation**

* Adds a full new documentation page: `docs/advanced_features/forward_hooks.md`.
* Expands `server_arguments.md` with detailed descriptions of hook-related arguments.

### **Testing**

* Adds a full test suite for hook resolution and execution:

  * Verifies hooks fire correctly.
  * Ensures unmatched patterns do not cause failures.
  * Confirms CLI-configured hooks integrate end-to-end.

---

## Accuracy Tests

This PR does not modify model forward logic, kernels, sampling, or numerical behavior.
Forward hooks run after module computations and return the unmodified output tensor unless a user overrides it.

---

## Benchmarking and Profiling

Hook functionality is entirely opt-in and disabled by default (`--enable-hooks` must be set). When disabled, there is zero overhead.

When enabled, the overhead is proportional to:

* Number of matched modules
* Cost of user-provided hook functions

No core execution path is affected, so no baseline speed regressions occur.

---

## Checklist

* [x] Format your code according to the [[Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit)](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
* [x] Add unit tests according to the [[Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests)](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
* [x] Update documentation according to [[Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations)](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
* [x] Since this feature does not affect model numerics or performance unless hooks are explicitly enabled, accuracy and speed benchmarks are not required.


python/sglang/srt/model_executor/model_runner.py
```diff
@@ -496,6 +498,15 @@ def initialize(self, min_per_gpu_memory: float):
             self.graph_mem_usage = 0
             self.init_attention_backend()
 
+        if server_args.enable_hooks:
+            hook_specs = getattr(server_args, "hooks", None)
```
hnyls2002: Why `getattr` instead of `server_args.hooks`?

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -2046,6 +2057,54 @@ def init_threads_binding(self):
                     f"Please set proper `--max-total-tokens` to avoid the out-of-memory error."
                 )
 
+    def register_hooks(self, hook_specs: List[dict[str, Any]]) -> None:
```
hnyls2002: Move this part of code into a separate module `CustomHook`, and use an instance of this `CustomHook` to do the registry and calling.

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -2368,6 +2427,32 @@ def _unwrap_tensor(tensor, tp_rank, device):
     return tensor.to(device)
 
 
+def resolve_callable(path: Optional[str]) -> Optional[Callable]:
```
hnyls2002: Also, move this into a separate module. We shouldn't add more functions to the model runner.

python/sglang/srt/server_args.py
```diff
@@ -570,6 +571,10 @@ class ServerArgs:
     decrypted_config_file: Optional[str] = None
     decrypted_draft_config_file: Optional[str] = None
 
+    # For forward hooks
+    enable_hooks: bool = False
```
hnyls2002: Two arguments are unneeded. Only one `custome_hooks` is Ok. Use `None` to represent the disabling status.

hnyls2002 (changes requested):

python/sglang/srt/model_executor/model_runner.py
Carlomus: everything has now been moved to hook_mamager.py

python/sglang/srt/model_executor/model_runner.py
Carlomus: good point, changed

hnyls2002 (approved):

PR #13225, opened 2025-11-13
Fix NVFP4 dispatch assertion failure with dummy weights

ishandhanani:
## Problem

When running `compile_deep_gemm` with `SGLANG_CUTEDSL_MOE_NVFP4_DISPATCH=1`, the process fails with an assertion error:

```python
AssertionError at modelopt_quant.py:1454
assert torch.all(w13_input_scale == w13_input_scale[0])
```

This occurs because:
- `compile_deep_gemm` uses `load_format="dummy"` to speed up compilation
- Dummy weight initialization assigns random values to quantization scales
- Each expert gets a different random `w13_input_scale` value
- The NVFP4 dispatch path requires uniform scales across all experts

Notably, this assertion does not fail during normal `launch_server` because real checkpoints have properly initialized scales.

## Solution

This PR modifies the assertion to handle dummy weights gracefully:

- Detects when running in DeepGEMM precompile stage via `SGLANG_IN_DEEPGEMM_PRECOMPILE_STAGE`
- If scales differ during precompile, logs a warning and uses the first expert's scale for all experts
- For normal inference with real checkpoints, preserves the original strict behavior with a more informative error message
- The scale values don't affect kernel compilation (only shapes/dtypes matter), so using a uniform dummy value is safe

## Testing [WIP]

- Existing functionality preserved: real checkpoints with uniform scales continue to work
- Enables `compile_deep_gemm` to run with `SGLANG_CUTEDSL_MOE_NVFP4_DISPATCH=1`
- Better error messages help users diagnose scale mismatch issues


python/sglang/srt/layers/quantization/modelopt_quant.py
```diff
@@ -1451,7 +1451,21 @@ def _slice_scale(w):
             w2_input_scale = _slice_scale(w2_input_scale)
 
             if CUTEDSL_MOE_NVFP4_DISPATCH:
-                assert torch.all(w13_input_scale == w13_input_scale[0])
+                # When using dummy weights (e.g., during compile_deep_gemm), the input scales
```
kaixih, line 1454: https://github.com/sgl-project/sglang/blob/060c341b26ccab35d861eb5b5ab65f6df73cf572/python/sglang/srt/layers/quantization/modelopt_quant.py#L1428-L1437
I am a bit confused. If the above `if-branch` is taken, we won't have any issue here even we are using dummy weights, right?

Fridge003: @kaixih For deepgemm precompilation all the weights are dummy weights, so the assertion will easily fail if there is any Nan. I think this change make sense.

kaixih: if SGLANG_IN_DEEPGEMM_PRECOMPILE_STAGE=true, we actually don't need to check torch.all(w13_input_scale == w13_input_scale[0], if we are working with dummy weights? right

Fridge003: Yes this is what I mean

Fridge003 (approved):

PR #13229, opened 2025-11-13
[GLM4.6v] Required changes for bumping up to transformer 5.x

byjiang1996:
## Motivation

Required changes for bumping up to transformer 5.x. It is required to support new GLM-v model: https://github.com/huggingface/transformers/pull/42122.

Verified with GLM team. This change works for transformer 5.x.

Their new GLM model is included in transformer 5.x which hasn't been officially released so we cannot directly upgrade to trans 5.x in this PR. However, this PR should be safe to merge b/c it is compatible with both trans 4.x and trans 5.x

## Modification

Can be found in the opened PR comments. Only two changes.

## Checklist


python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -324,6 +324,7 @@ def _apply_jinja_template(
                 **(
                     request.chat_template_kwargs if request.chat_template_kwargs else {}
                 ),
+                return_dict=False
```
byjiang1996: Default value changed from False to True in 5.x transfomer. We have to explicitly set it to False in all places...
image

image

python/sglang/multimodal_gen/runtime/loader/weight_utils.py
```diff
@@ -46,7 +46,8 @@ def enable_hf_transfer() -> None:
 class DisabledTqdm(tqdm):
 
     def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs, disable=True)
+        kwargs["disable"] = True
```
mickqian, line 49: /gemini what's the difference here?

python/sglang/srt/entrypoints/openai/serving_chat.py
JustinTong0323: How can they always do this kinds of breaking changes🤣

JustinTong0323 (approved):
LGTM

PR #13231, opened 2025-11-13
[disagg] fix a race condition

flying-x:
- The transfer_sync() function was used to unblock the decode side via the RPC
- It creates a race condition since the transfer_sync() doesn't ensure the receiver actually receives the data before returning
- this change introduce a completion flag to fix this race condition.

## Checklist

### DIFF
--- a/python/sglang/srt/disaggregation/mooncake/conn.py
+++ b/python/sglang/srt/disaggregation/mooncake/conn.py
@@ -871,15 +871,22 @@ class MooncakeKVManager(CommonKVManager):
                             # Set the completion flag on the decode side
                             ret_flag = 0
                             assert req.dst_completion_flag_ptr != 0, "Destination completion flag pointer cannot be zero."
-                            src_ptr = self.request_src_completion_flag_ptrs[
-                                kv_chunk.room
-                            ]
-                            ret_flag = self.engine.batch_transfer_sync(
-                                req.mooncake_session_id,
-                                [src_ptr],
-                                [req.dst_completion_flag_ptr],
-                                [self.completion_flag_byte_size],
-                            )
+                            with self.completion_flag_lock:
+                                src_ptr = self.request_src_completion_flag_ptrs.get(kv_chunk.room)
+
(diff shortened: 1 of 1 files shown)


ShangmingCai:
@flying-x Can you describe more? The `transfer_sync` will only return when all data is written to the dst ptr. So I don't really think there is a race condition in the timeline.

ShangmingCai:
@flying-x If you have really confirmed that there exists a race condition, can you share a reproduction case?

RDMA write is a one-sided operation. If prefill finishes this step, then it is definitely done, and we notify the decode side after this step, so this statement should be wrong: "It creates a race condition since the transfer_sync() doesn't ensure the receiver actually receives the data before returning".

flying-x:
> @flying-x Can you describe more? The `transfer_sync` will only return when all data is written to the dst ptr. So I don't really think there is a race condition in the timeline.

@ShangmingCai I did some debugging and the problem appears to be that transfer_sync() returns before the data is cudaMemcpy() into the GPU. By "dst ptr", it should be the GPU ptr, but mooncake doesn't seem to account for it. 

I added this log in the prefill side:

```
        auto session = std::make_shared<Session>(std::move(socket));
        session->on_finalize_ = [slice](TransferStatusEnum status) {
            if (status == TransferStatusEnum::COMPLETED) {
                LOG(ERROR) << "mark success!";
                slice->markSuccess();
            } else {
                slice->markFailed();
            }
        };
```

And I added delay on the decode side during the cudaMemcpy in the same file (tcp_transport.cpp).

```
        if (1) {
            pid_t pid = getpid();
            LOG(INFO) << "pid=" << pid << " readBody " << buffer_size;
            std::this_thread::sleep_for(std::chrono::milliseconds(500));
        }

        asio::async_read(
        ...
```
I can see that the the mark success is printed much earlier than the cudaMemcpy().

Therefore, I think the PR here is just a workaround in the mooncake issue. Can you tag someone on the mooncake side to confirm what I am observing? Much thanks!

PR #7277, opened 2025-06-17
Support multi-thread model weight loading

xianzhiT:
## Motivation
Support multi-thread model weight loading when using the --load-format "auto", which can speedup weight load process for users with distributed storage backend or NVME ssd disk. 

Tests have found that storing DeepSeek-R1-0528 model weights on a single NVME drive for cold start of services can speed up model loading time by 6.26 times

## Modifications
base on PR https://github.com/sgl-project/sglang/issues/4822

- Multi-threaded Weight Loading: Implemented multi-threaded loading for safetensors and PyTorch (.pt/.bin) model weights, significantly accelerating the model loading process, especially for large sharded models.

- Introduced a new --model-loader-extra-config command-line argument: Allowing users to enable multi-threaded loading (enable_multithread_load) and specify the number of threads (num_threads) via a JSON string.

 
Cmd Example: 
```bash
echo "1" > /proc/sys/vm/drop_caches

python3 -m sglang.launch_server \
 --model-path /data/DeepSeek-R1-0528 \
 --tp 8 \
 --trust-remote-code \
 --mem-fraction-static 0.9 \
 --host 0.0.0.0 \
 --port 8001 \
 --enable-metric \
 --log-requests \
 --log-requests-level 0 \
 --allow-auto-truncate  \
 --quantization fp8 \
 --model-loader-extra-config '{"enable_multithread_load": true, "num_threads": 8}'
```

Test DeepSeek-R1-0528 safetensor loading in NVME ssd PCIE Gen4 (Speed 16GT/s, Width x4)
|                                    | Time consumption before optimization | Time consumption after optimization | Load Weight Speed up |
| ---------------------------------- | ------------------------------------ | ----------------------------------- | -------------------- |
| weight in single NVME ssd          | 702s                                 | 112s                                | 6.26x                |
| weight in raid0 with four NVME ssd | 656s                                 | 62s                                 | 10.58x               |

## Checklist

- [x] Format your code according to the [Code Formatting with Pre-Commit](https://docs.sglang.ai/references/contribution_guide.html#code-formatting-with-pre-commit).
- [ ] Add unit tests as outlined in the [Running Unit Tests](https://docs.sglang.ai/references/contribution_guide.html#running-unit-tests-adding-to-ci).
- [ ] Update documentation / docstrings / example tutorials as needed, according to [Writing Documentation](https://docs.sglang.ai/references/contribution_guide.html#writing-documentation-running-docs-ci).
[...]

PR #13237, opened 2025-11-14
ci: speed up b200 ci

b8zhong:
I suppose maybe the runner was down earlier and it dropped it from page cache, so it's really slow here and seems to be timing out

https://github.com/sgl-project/sglang/actions/runs/19320913027/job/55356946845?pr=13101

Personally, I also found that this way is faster loading in page cache too (maybe 20-30 pct) (original author also found this https://github.com/sgl-project/sglang/pull/7277).

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -877,7 +877,7 @@ jobs:
           python3 run_suite.py --suite per-commit-8-gpu-h200-deepep
 
   unit-test-backend-4-gpu-b200:
-    needs: [check-changes, unit-test-backend-2-gpu, sgl-kernel-build-wheels]
+    needs: [check-changes, sgl-kernel-build-wheels]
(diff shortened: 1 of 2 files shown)


hnyls2002 (approved):

.github/workflows/pr-test.yml
```diff
@@ -877,7 +877,7 @@ jobs:
           python3 run_suite.py --suite per-commit-8-gpu-h200-deepep
 
   unit-test-backend-4-gpu-b200:
-    needs: [check-changes, unit-test-backend-2-gpu, sgl-kernel-build-wheels]
+    needs: [check-changes, sgl-kernel-build-wheels]
```
Fridge003:
```suggestion
    needs: [check-changes, unit-test-backend-2-gpu, sgl-kernel-build-wheels]
```

PR #13247, opened 2025-11-14
Extend lint test to test/ directory

Kangyan-Zhou:
Fix lint error in https://github.com/sgl-project/sglang/actions/runs/19350506857/job/55361058653#step:4:645 and extend the precommit check to the test folder

### DIFF
--- a/test/srt/test_vision_openai_server_a.py
+++ b/test/srt/test_vision_openai_server_a.py
@@ -204,11 +204,12 @@ class TestDeepseekOCRServer(TestOpenAIMLLMServerBase):
 
 
 if __name__ == "__main__":
-    del (
-        TestOpenAIMLLMServerBase,
-        ImageOpenAITestMixin,
-        VideoOpenAITestMixin,
-        AudioOpenAITestMixin,
-        OmniOpenAITestMixin,
-    )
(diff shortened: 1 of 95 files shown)


Fridge003 (approved):

test/srt/test_vision_openai_server_a.py
```diff
@@ -207,11 +207,12 @@ def test_single_image_chat_completion(self):
...
-        OmniOpenAITestMixin,
-    )
+    # Note: Cannot delete mixin classes imported via * since they're not in local scope
+    # del (
+    #     TestOpenAIMLLMServerBase,
+    #     ImageOpenAITestMixin,
+    #     VideoOpenAITestMixin,
+    #     AudioOpenAITestMixin,
+    #     OmniOpenAITestMixin,
+    # )
```
ping1jing2, line 217: how about just deleting these codes?

ping1jing2 (approved):

Fridge003 (approved):

test/srt/test_vision_openai_server_a.py
yhyang201: Image
These changes in this section will cause errors when the test file is executed. 
Should we revert these modifications and still delete these classes?

@Kangyan-Zhou

PR #13254, opened 2025-11-14
[BugFix] fix bench_serving error when multimodal image is testing

ZLkanyo009:
# Modifications
When testing multimodal models (such as qwen3vl) using bench_serving.py, errors may occur probabilistically because randomly generated tokens can occasionally produce image tokens. This leads to a situation where there is actually only one image, but multiple image tokens appear (i.e., the code assumes there are multiple images). This modification fixes this bug.

# Error manifestation
server:
```
model=/models/nvme0/Qwen3-VL-235B-A22B-Instruct-FP8/
   --model-path ${model} \
   --host 0.0.0.0 \
   --port 9000 \
   --tp 8 \
   --ep 8
```
client:
```
  --backend sglang \
  --port 9000 \
  --model "/models/nvme0/Qwen3-VL-235B-A22B-Instruct-FP8/" \
  --dataset-name image \
  --image-count 1 \
  --max-concurrency 64 \
  --image-resolution 800x800 \
  --random-input-len 1000 \
  --random-output-len 1000 \
  --num-prompts 640 
```

Regardless of whether on NVIDIA GPUs or AMD GPUs
image

# Fix
After the fix, normal testing proceeds as follows:
image


HaiShaw (approved):

yhyang201:
A very important bug fix — Thank you!

yuan-luo (approved):

mickqian (approved):

yuan-luo:
> > @ZLkanyo009 Could you please follow up and check it? Thanks.
> 
> I have tried many times and haven't encountered any errors.

Not sure why, please ignore it then.

Issue #13234, opened 2025-11-13
[Bug] Kimi K2 Thinking Marlin Kernel Crash


RahulB200:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Describe the bug

Kimi K2 Thinking crashes on B200s. It appears to be an issue with the fused marlin kernel. 

Logs dump:

[rank6]:[E1113 23:24:11.476061221 ProcessGroupNCCL.cpp:2068] [PG ID 2 PG GUID 3 Rank 6] Process group watchdog thread terminated with exception: CUDA error: an illegal memory access was encountered

CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.

For debugging consider passing CUDA_LAUNCH_BLOCKING=1

Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.

Exception raised from c10_cuda_check_implementation at /pytorch/c10/cuda/CUDAException.cpp:42 (most recent call first):

frame #0: c10::Error::Error(c10::SourceLocation, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) + 0x80 (0x7541690d9eb0 in /usr/local/lib/python3.12/dist-packages/torch/lib/libc10.so)

frame #1: <unknown function> + 0x111c7 (0x75416916c1c7 in /usr/local/lib/python3.12/dist-packages/torch/lib/libc10_cuda.so)

frame #2: c10d::ProcessGroupNCCL::WorkNCCL::finishedGPUExecutionInternal() const + 0x50 (0x7540f167b0c0 in /usr/local/lib/python3.12/dist-packages/torch/lib/libtorch_cuda.so)

frame #3: c10d::ProcessGroupNCCL::WorkNCCL::isCompleted() + 0x68 (0x7540f168a8a8 in /usr/local/lib/python3.12/dist-packages/torch/lib/libtorch_cuda.so)

[...]

PR #13262, opened 2025-11-14
Guard fused Marlin MoE atomic add usage to prevent crashes

0xymoro:
## Motivation

Marlin‑quantized MoE crash because `sgl_kernel.fused_moe` always forces the experimental atomic-add reduction path. Those large (M, N, K) shapes exceed the safe window for the Marlin kernel and trigger illegal memory accesses. We already have a proven heuristic elsewhere in the codebase; this PR reuses it so atomic add is only enabled when the shape/device combination is known safe.

  ## Modifications

  - Import `should_use_atomic_add_reduce` inside `sgl_kernel/python/sgl_kernel/fused_moe.py`.
  - Evaluate the heuristic separately for the w1 and w2 GEMMs (first stage uses `(M, 2N, K)`, second uses `(M·topk, K, N)`) and pass the result as `use_atomic_add` to each `moe_wna16_marlin_gemm` call.
  - No changes to CUDA kernels; we simply stop invoking them with unsafe settings.

### DIFF
--- a/sgl-kernel/python/sgl_kernel/fused_moe.py
+++ b/sgl-kernel/python/sgl_kernel/fused_moe.py
@@ -71,6 +71,9 @@ def fused_marlin_moe(
         moe_align_block_size,
         try_get_optimal_moe_config,
     )
+    from sglang.srt.layers.quantization.marlin_utils import (
+        should_use_atomic_add_reduce,
+    )
 
     # Check constraints.
     assert hidden_states.shape[0] == gating_output.shape[0], "Number of tokens mismatch"
@@ -142,10 +145,16 @@ def fused_marlin_moe(
     intermediate_cache3 = intermediate_cache13[: M * topk_ids.shape[1] * K]
     intermediate_cache3 = intermediate_cache3.view(-1, K)
 
-    use_atomic_add = (
-        hidden_states.dtype == torch.half
-        or torch.cuda.get_device_capability(hidden_states.device)[0] >= 9
-    )
+    def _should_use_atomic_add(m: int, n: int, k: int) -> bool:
+        return should_use_atomic_add_reduce(
+            m=m,
+            n=n,
+            k=k,
+            device=hidden_states.device,
+            dtype=hidden_states.dtype,
+        )
+
+    use_atomic_add_first = _should_use_atomic_add(M, 2 * N, K)
 
     intermediate_cache1 = torch.ops.sgl_kernel.moe_wna16_marlin_gemm.default(
         hidden_states,
@@ -169,7 +178,7 @@ def fused_marlin_moe(
         size_n=2 * N,
         size_k=K,
         is_k_full=is_k_full,
-        use_atomic_add=use_atomic_add,
+        use_atomic_add=use_atomic_add_first,
         use_fp32_reduce=True,
         is_zp_float=False,
     )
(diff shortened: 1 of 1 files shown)


FlamingoPg:
Thanks, could plz help fix lint?

0xymoro:
I got to a point where the marlin kernel kind of overflowed with large batchsizes. tracing thru but it's a slog, and I worked off Hopper and yes reproducible on large bsz

Current best guess and try:

Root cause: the second Marlin GEMM (moe_wna16_marlin_gemm that multiplies expert outputs back into the model space) was receiving
extremely large batches—e.g. M * topk = 131 072 tokens with N = 7168, K = 256. The kernel relies on a workspace/lock buffer
sized roughly by sms * 4, which was fine for moderate M but far too small for those huge routed batches. Once the workspace got
exhausted, the kernel wrote past the end of the buffer, corrupting intermediate_cache3 before moe_sum_reduce ever ran. The later
moe_sum_reduce/torch.sum change can’t help because the memory was already trashed by the time the reduction ran.

Our fix: scale the workspace with the actual amount of routed work—max(2N, K) // 64 * (# routed token blocks)—and only fall back to
sms * 4 if the computed size is small. That keeps the per-token lock/temporary storage available even when M * topk explodes, so the
Marlin kernel never hits overlap/corruption in the first place. We left the CUDA binaries untouched; the Python glue now allocates
enough space before calling them.

[2025-11-15 02:04:00 TP2] marlin stage2: M=131072 N=7168 K=256 use_atomic=False
[2025-11-15 02:04:00 TP5] marlin stage2: M=131072 N=7168 K=256 use_atomic=False
[2025-11-15 02:04:00 TP4] marlin stage2: M=131072 N=7168 K=256 use_atomic=False
[2025-11-15 02:04:00 TP0] marlin stage2: M=131072 N=7168 K=256 use_atomic=False
[2025-11-15 02:04:00 TP3] marlin stage2: M=131072 N=7168 K=256 use_atomic=False
[2025-11-15 02:04:00 TP6] marlin stage2: M=131072 N=7168 K=256 use_atomic=False
[2025-11-15 02:04:00 TP1] marlin stage2: M=131072 N=7168 K=256 use_atomic=False
[2025-11-15 02:04:00 TP7] marlin stage2: M=131072 N=7168 K=256 use_atomic=False
[2025-11-15 02:04:00 TP3] Scheduler hit an exception: Traceback (most recent call last):
File "/sgl-workspace/sglang/python/sglang/srt/managers/scheduler.py", line 2725, in run_scheduler_process
File "/usr/local/lib/python3.12/dist-packages/torch/utils/_contextlib.py", line 120, in decorate_context
return func(*args, **kwargs)
^^^^^^^^^^^^^^^^^^^^^
File "/sgl-workspace/sglang/python/sglang/srt/managers/scheduler.py", line 1019, in event_loop_overlap
batch_result = self.run_batch(batch)
^^^^^^^^^^^^^^^^^^^^^
File "/sgl-workspace/sglang/python/sglang/srt/managers/scheduler.py", line 2021, in run_batch
batch_result = self.model_worker.forward_batch_generation(
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/sgl-workspace/sglang/python/sglang/srt/managers/tp_worker.py", line 371, in forward_batch_generation
logits_output, can_run_cuda_graph = self.model_runner.forward(

pdasgup:
is there any query to reproduce or fix for this issue? we are observing this crash https://github.com/sgl-project/sglang/issues/13234 on 8xB200 with image 0.5.5.post3 for Kimi K2 Thinking model

PR #13263, opened 2025-11-14
diffusion: enable fa4 for blackwell

yhyang201:
## Motivation

close #12986

Enable fa4 for blackwell

## Checklist


python/sglang/multimodal_gen/runtime/platforms/cuda.py
```diff
@@ -216,17 +219,15 @@ def get_attn_backend_cls_str(
...
         else:
             if is_blackwell():
-                target_backend = AttentionBackendEnum.TORCH_SDPA
-                logger.debug(f"Use torch_sdpa as default backend")
-            else:
-                target_backend = AttentionBackendEnum.FA3
-                logger.debug(f"Use fa3 as default backend")
+                set_fa_ver(4)
+            target_backend = AttentionBackendEnum.FA3
+            logger.debug(f"Use fa3 (fa4 for blackwell) as default backend")
```
mickqian: should we rename the server args to `FA`?

mickqian (commented):
Please check other places with naming similar to `FlashAttention...3...` too

python/sglang/multimodal_gen/runtime/platforms/cuda.py
yhyang201: Done

yhyang201:
I’ve changed all “FA3” references to “FA”, which could serve as a temporary solution. Once FA4 introduces the sglang LLM component, we can synchronize it accordingly.
Additionally, it will still accept “fa3” and “fa4” as input args — they’ll be automatically converted to “fa”.

python/sglang/multimodal_gen/runtime/server_args.py
```diff
@@ -843,14 +843,14 @@ def check_server_sp_args(self):
             )
 
         if self.ring_degree > 1:
```
mickqian, line 848: Do we warn the users, at this stage, if the attention backend str is correct?

yhyang201: It’s a bit unexpected, it seems that the code I previously pushed was incomplete. 
I’ve now added the missing logic: during the `__post_init__` stage, if the user provides `fa3` or `fa4`, the value will be automatically normalized to `fa`, and the program will select either `fa3` or `fa4` based on whether the platform is Hopper or Blackwell.

mickqian: great. if the user provide sth like `unknown_attention_backend`, an exception should be thrown in server_args. Is it now?

python/sglang/multimodal_gen/runtime/server_args.py
```diff
@@ -382,7 +385,7 @@ def add_cli_args(parser: FlexibleArgumentParser) -> FlexibleArgumentParser:
             "--attention-backend",
             type=str,
             default=None,
-            choices=[e.name.lower() for e in AttentionBackendEnum],
+            choices=[e.name.lower() for e in AttentionBackendEnum] + ["fa3", "fa4"],
```
yhyang201, line 388: @mickqian This line of code ensures that an error is raised when an unknown attention backend is provided.

mickqian:
@FlamingoPg Could you do a benchmark on Blackwell, on fa4? thanks!

PR #13268, opened 2025-11-14
[Fix] Fix speculative decoding when calling release/resume memory.

QsingHuan:
## Motivation

This commit addresses the use of speculative decoding algorithm in reinforcement learning (RL) scenarios. Previously, when using **speculative decoding in veRL**, the draft worker would also be released, causing the **draft worker's weights to become zero** after resuming memory. 
Since there is currently no weight update mechanism specifically designed for draft workers, we enforce the use of CPU offload as a mandatory requirement to ensure proper functionality.

## Modifications
Add
```python
enable_cpu_backup = True if self.is_draft_worker else self.server_args.enable_weights_cpu_backup
```

### DIFF
--- a/python/sglang/srt/model_executor/model_runner.py
+++ b/python/sglang/srt/model_executor/model_runner.py
@@ -756,9 +756,12 @@ class ModelRunner:
         # Remove monkey_patch when linear.py quant remove dependencies with vllm
         monkey_patch_vllm_parallel_state()
 
+        enable_cpu_backup = (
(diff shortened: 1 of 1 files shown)


hnyls2002 (approved):
@fzyzcjy please take a review and merge it

fzyzcjy:
hmm seems similar to https://github.com/sgl-project/sglang/pull/13318 cc @zhuzilin

PR #13274, opened 2025-11-14
[NVIDIA] Fix use case of SGLANG_ENABLE_FLASHINFER_GEMM

kaixih:
## Motivation

When SGLANG_ENABLE_FLASHINFER_GEMM is enabled, we observed that the accuracy test fails. (by @gracehonv )
The root cause is that the model execution switches to use e8m0 scales for Blackwell GPUs even when DeepGEMM is not in use.
This PR fixes the issue by ensuring that the correct scaling mode is applied.

## Accuracy Tests

script:
```
set -x
if [[ "$1" == "server" ]]; then
model_str=/model/deepseek-ai-DeepSeek-R1-0528/models--deepseek-ai--DeepSeek-R1-0528/snapshots/4236a6af538feda4548eca9ab308586007567f52
export SGLANG_ENABLE_FLASHINFER_GEMM=true
#export SGLANG_SUPPORT_CUTLASS_BLOCK_FP8=true

    --trust-remote-code \
    --disable-radix-cache \
    --max-running-requests 512 \
    --chunked-prefill-size 8192 \
    --mem-fraction-static 0.9 \
    --cuda-graph-max-bs 128 \
    --max-prefill-tokens 8192 \
    --kv-cache-dtype fp8_e4m3 \
    --quantization fp8 \
    --model-path ${model_str} \
    --host 0.0.0.0 \
    --port 8000 \
    --tensor-parallel-size=4 \
    --data-parallel-size=1 \
    --expert-parallel-size=1 \
    --scheduler-recv-interval 10 \
    --stream-interval 10 \
    --tokenizer-path ${model_str} \
    --attention-backend trtllm_mla \
    --moe-runner-backend flashinfer_trtllm \
    --enable-symm-mem \

fi

if [[ "$1" == "bench" ]]; then
python3 /sgl-workspace/sglang/benchmark/gsm8k/bench_sglang.py \
  --num-questions 2000 \
  --parallel 2000 \
  --num-shots 8 \
  --port 8000
fi

if [[ "$1" == "chat" ]]; then
  curl http://127.0.0.1:8000/v1/chat/completions \
    -d '{
      "model": "$model_str",
      "messages": [{"role": "assistant", "content": "What is 37 * 42?"}],
      "extra_body": {
        "chat_template_kwargs": {
          "thinking": true
        }
      }
    }'
fi
```

result:
```
100%|███████████████████████████████████████████████████████████████████████████| 1319/1319 [04:30<00:00,  4.87it/s]
Accuracy: 0.966
Invalid: 0.000
Latency: 271.168 s
Output throughput: 523.439 token/s
+ [[ bench == \c\h\a\t ]]
```

## Checklist

### DIFF
--- a/python/sglang/srt/layers/quantization/fp8_utils.py
+++ b/python/sglang/srt/layers/quantization/fp8_utils.py
@@ -129,7 +129,7 @@ def cutlass_block_fp8_supported() -> bool:
 
 CUTLASS_BLOCK_FP8_SUPPORTED = cutlass_block_fp8_supported()
 ENABLE_FLASHINFER_FP8_GEMM = (
-    envs.SGLANG_ENABLE_FLASHINFER_FP8_GEMM.value
+    envs.SGLANG_ENABLE_FLASHINFER_FP8_GEMM.get()
     and is_blackwell_supported()
     and is_flashinfer_available()
 )
--- a/python/sglang/srt/layers/quantization/fp8_utils.py
+++ b/python/sglang/srt/layers/quantization/fp8_utils.py
@@ -127,17 +127,17 @@ def cutlass_block_fp8_supported() -> bool:
 
 
 CUTLASS_BLOCK_FP8_SUPPORTED = cutlass_block_fp8_supported()
-ENABLE_FLASHINFER_GEMM = (
-    get_bool_env_var("SGLANG_ENABLE_FLASHINFER_GEMM")
+ENABLE_FLASHINFER_FP8_GEMM = (
+    get_bool_env_var("SGLANG_ENABLE_FLASHINFER_FP8_GEMM")
     and is_blackwell_supported()
     and is_flashinfer_available()
 )
-if ENABLE_FLASHINFER_GEMM:
+if ENABLE_FLASHINFER_FP8_GEMM:
     from flashinfer.gemm import gemm_fp8_nt_groupwise
 
 
 def dispatch_w8a8_block_fp8_linear() -> Callable:
-    if ENABLE_FLASHINFER_GEMM:
+    if ENABLE_FLASHINFER_FP8_GEMM:
         return flashinfer_gemm_w8a8_block_fp8_linear
     elif CUTLASS_BLOCK_FP8_SUPPORTED:
         return cutlass_w8a8_block_fp8_linear_with_fallback
--- a/python/sglang/srt/layers/quantization/fp8_utils.py
+++ b/python/sglang/srt/layers/quantization/fp8_utils.py
@@ -2,6 +2,7 @@ from typing import Callable, List, Optional, Tuple
 
 import torch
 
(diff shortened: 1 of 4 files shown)


python/sglang/srt/models/deepseek_v2.py
```diff
@@ -96,6 +96,7 @@
     quant_weight_ue8m0,
     requant_weight_ue8m0_inplace,
     transform_scale_ue8m0_inplace,
+    ENABLE_FLASHINFER_GEMM,
```
Fridge003: Can we rename this environ to `ENABLE_FLASHINFER_FP8_GEMM` so it's not confused with fp4 gemm.
When I added this environ like 6 months ago, there was no fp4 gemm kernel in flashinfer.

The original name can be deprecated https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/environ.py#L300

kaixih: Done

python/sglang/srt/layers/quantization/fp8_utils.py
```diff
@@ -127,17 +127,17 @@ def cutlass_block_fp8_supported() -> bool:
 
 
 CUTLASS_BLOCK_FP8_SUPPORTED = cutlass_block_fp8_supported()
-ENABLE_FLASHINFER_GEMM = (
-    get_bool_env_var("SGLANG_ENABLE_FLASHINFER_GEMM")
+ENABLE_FLASHINFER_FP8_GEMM = (
+    get_bool_env_var("SGLANG_ENABLE_FLASHINFER_FP8_GEMM")
```
hnyls2002: Do not use get bool env var. Use from `sglang.srt.environ.`

hnyls2002 (changes requested):

python/sglang/srt/layers/quantization/fp8_utils.py
```diff
@@ -128,7 +129,7 @@ def cutlass_block_fp8_supported() -> bool:
 
 CUTLASS_BLOCK_FP8_SUPPORTED = cutlass_block_fp8_supported()
 ENABLE_FLASHINFER_FP8_GEMM = (
-    get_bool_env_var("SGLANG_ENABLE_FLASHINFER_FP8_GEMM")
+    envs.SGLANG_ENABLE_FLASHINFER_FP8_GEMM.value
```
Fridge003:
```suggestion
    envs.SGLANG_ENABLE_FLASHINFER_FP8_GEMM.get()
```

kaixih:
```
%|███████████████████████████████████████████████████████████████████████████| 1319/1319 [04:11<00:00,  5.25it/s]
Accuracy: 0.964
Invalid: 0.000
Latency: 251.736 s
Output throughput: 562.976 token/s
```

Fridge003 (approved):

PR #13277, opened 2025-11-14
refactor: remove duplicate function _get_bootstrap_info_from_server

acelyc111:
## Motivation

The removed function `_get_bootstrap_info_from_server` is also in python/sglang/srt/disaggregation/common/conn.py with completely the same implementation.

## Checklist


ShangmingCai (approved):
LGTM. Seems like I forgot to remove it when constructing the common backend.

PR #13293, opened 2025-11-14
[router]Replace requests lib with openai in e2e_response_api

XinyueZhang369:
## Motivation

Replace requests lib with openai lib in e2e_response_api 

## Modifications

- Create crud functions with openai responses and conversations apis in basic_crud.py
- Change to use new crud functions in all tests in mixins
- Disable test_structured_output_json_schema, cause noticed that the test is able to pass when calling openai endpoint directly while it can only get text format result with router. We probably can only re-enable it after we support other format in router. 

Local test run passed:

Screenshot 2025-11-13 at 7 52 52 PM

## Checklist


sgl-router/py_test/e2e_response_api/backends/test_grpc_backend.py
```diff
@@ -51,6 +53,7 @@ def setUpClass(cls):
         )
 
         cls.base_url = cls.cluster["base_url"]
+        cls.client = openai.Client(api_key=cls.api_key, base_url=cls.base_url + "/v1")
```
CatherineSue, line 56: nit: do we really need v1?

XinyueZhang369: Yeah I tried to run the test without "v1", but got 404 for all requests

sgl-router/py_test/e2e_response_api/backends/test_grpc_backend.py
```diff
@@ -64,8 +67,7 @@ def test_conversation_with_multiple_turns(self):
 
     def test_structured_output_json_schema(self):
         """Override with simpler schema for Llama model (complex schemas not well supported)."""
-        data = {
-            "model": self.model,
```
key4ng, line 68: curious it doesn't require model field?

sgl-router/py_test/e2e_response_api/mixins/structured_output.py
```diff
@@ -6,48 +6,24 @@
...
-            response = self.session.post(url, headers=headers, json=data)
-        elif method == "DELETE":
-            response = self.session.delete(url, headers=headers)
-        else:
-            raise ValueError(f"Unsupported method: {method}")
-
-        return response
+class StructuredOutputBaseTest(ResponseAPIBaseTest):
 
+    @unittest.skip("TODO: Add structured output feature")
```
key4ng: > Disable test_structured_output_json_schema, cause noticed that the test is able to pass when calling openai endpoint directly while it can only get text format result with router. We probably can only re-enable it after we support other format in router.

Didn't quite get it. Could you elaborate more. text format result can also be validated. 
also, does it happen only for openai backend, if so, we can skip only openai backend
or since we are using python sdk, maybe we could also using this example to validate: https://platform.openai.com/docs/guides/structured-outputs?example=structured-data&lang=python

sgl-router/py_test/e2e_response_api/backends/test_grpc_backend.py
XinyueZhang369: Yes, create api does require model field, we fill it in create_response func https://github.com/sgl-project/sglang/blob/d2cafb6d50a86997946b955db72ceef95ef38c1d/sgl-router/py_test/e2e_response_api/mixins/basic_crud.py#L68

sgl-router/py_test/e2e_response_api/mixins/structured_output.py
XinyueZhang369: Yeah good call. So when I run structured_output.py and let it hit openai url directly, this test passed. But when I run test_http_backend.py that calls router url, the test failed to get response in json_schema format, that's why I suspect that the router doesn't support this for now. 
But I didn't realize that it disable for all backends, I'll check if it's the same for other backends.

key4ng: I see I see, what's the error message you got for the failure in http backend?

XinyueZhang369: It failed at line 71: `self.assertEqual(create_resp.text.format.type, "json_schema")`, the `create_resp.text.format.type` is 'text'. Also I got the same result when running `pytest py_test/e2e_response_api/backends/test_http_backend.py::TestOpenaiBackend::test_structured_output_json_schema` on master

XinyueZhang369: Synced with Keyang offline, summary:
It turns out due to outdated router build in my local env, as we just introduced structured output feature.
Removed this skip annotation.
Thanks @key4ng !

key4ng (approved):

PR #13300, opened 2025-11-14
Fix a few CI issues

Kangyan-Zhou:
- https://github.com/sgl-project/sglang/blob/6448b4cd2c4919a06a7d41989abdf435599dd80f/python/sglang/srt/debug_utils/text_comparator.py#L6 introduced a new package but it was not installed
- There was some random issue with model download with concurrent download on the same host machine, adding some fix
- Fix the concurrency lock for B200 nightly tests

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -105,6 +105,7 @@ test = [
   "matplotlib",
   "pandas",
   "peft",
+  "polars",
   "pytest",
   "sentence_transformers",
   "tabulate",
--- a/.github/workflows/nightly-test-b200.yml
+++ b/.github/workflows/nightly-test-b200.yml
@@ -11,7 +11,7 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: nightly-test-${{ github.ref }}
+  group: nightly-test-b200-${{ github.ref }}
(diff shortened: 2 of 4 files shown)


python/pyproject.toml
```diff
@@ -105,6 +105,7 @@ test = [
   "matplotlib",
   "pandas",
   "peft",
+  "polars",
```
zhyncs, line 108: why do we need to add `polars` to the dependency?
it looks like `python/sglang/srt/debug_utils/text_comparator.py` has not been used or imported

zhyncs (changes requested):

python/pyproject.toml
Kangyan-Zhou: I'm trying to fix the test failure in https://github.com/sgl-project/sglang/actions/runs/19378579172/job/55452566052?pr=13300

zhyncs: gotcha

zhyncs:
```
sglang/python/sglang/srt/debug_utils/log_parser.py", line 20, in parse
    import polars as pl
ModuleNotFoundError: No module named 'polars'
```

PR #13314, opened 2025-11-15
Revert moe sum reduce for marlin moe

ispobock:
## Motivation

moe_sum_reduce may have launch issue in some case, reported in #13234


yuan-luo (approved):

0xymoro:
Tried it but still crashes - seems like it's not even getting to there

PR #13318, opened 2025-11-15
[RL] support only do cpu backup on draft model

zhuzilin:
## Motivation

When doing colocated RL, we sometimes don't have the draft model weights from the trainer side and hope to only have cpu backup for the draft model.

Thank you for your time on reviewing this PR.

## Checklist

### DIFF
--- a/python/sglang/srt/model_executor/model_runner.py
+++ b/python/sglang/srt/model_executor/model_runner.py
@@ -756,11 +756,9 @@ class ModelRunner:
         # Remove monkey_patch when linear.py quant remove dependencies with vllm
         monkey_patch_vllm_parallel_state()
 
-        enable_cpu_backup = self.server_args.enable_weights_cpu_backup
-        if self.is_draft_worker:
-            enable_cpu_backup = (
-                enable_cpu_backup or self.server_args.enable_draft_weights_cpu_backup
-            )
+        enable_cpu_backup = self.server_args.enable_weights_cpu_backup or (
+            self.is_draft_worker and self.server_args.enable_draft_weights_cpu_backup
+        )
         with self.memory_saver_adapter.region(
             GPU_MEMORY_TYPE_WEIGHTS,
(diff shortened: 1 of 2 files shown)


python/sglang/srt/model_executor/model_runner.py
```diff
@@ -756,9 +756,14 @@ def load_model(self):
         # Remove monkey_patch when linear.py quant remove dependencies with vllm
         monkey_patch_vllm_parallel_state()
 
+        enable_cpu_backup = self.server_args.enable_weights_cpu_backup
+        if self.is_draft_worker:
+            enable_cpu_backup = (
+                enable_cpu_backup or self.server_args.enable_draft_weights_cpu_backup
+            )
```
fzyzcjy:
```suggestion
            enable_cpu_backup |= self.server_args.enable_draft_weights_cpu_backup
```

fzyzcjy (approved):
LGTM w/ a nit

python/sglang/srt/model_executor/model_runner.py
fzyzcjy: oh maybe

```
enable_cpu_backup = self.server_args.enable_weights_cpu_backup or (is_draft_worker and enable_draft_weights_cpu_backup)
```

zhuzilin: fixed.

python/sglang/srt/server_args.py
```diff
@@ -3410,6 +3411,11 @@ def add_cli_args(parser: argparse.ArgumentParser):
             action="store_true",
             help="Save model weights to CPU memory during release_weights_occupation and resume_weights_occupation",
```
fzyzcjy:
```suggestion
            help="Save model weights (both main model and draft model, if any) to CPU memory during release_weights_occupation and resume_weights_occupation",
```

PR #13320, opened 2025-11-15
[RL] Allow bypassing /health check

zhuzilin:
## Motivation

We hope to allowing disable the generation in `/health` check and only use the `/health_generate` to delibrately do health check with generation.

## Checklist

### DIFF
--- a/python/sglang/srt/entrypoints/http_server.py
+++ b/python/sglang/srt/entrypoints/http_server.py
@@ -413,18 +413,15 @@ async def health_generate(request: Request) -> Response:
     If the server is not running anything, this request will be run, so we know whether the server is healthy.
     """
 
-    if (
-        get_bool_env_var("SGLANG_DISABLE_GENERATION_HEALTHCHECK")
-        and request.url.path == "/health"
-    ):
-        return Response(status_code=200)
-
     if _global_state.tokenizer_manager.gracefully_exit:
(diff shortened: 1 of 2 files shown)


python/sglang/srt/entrypoints/http_server.py
```diff
@@ -413,6 +413,12 @@ async def health_generate(request: Request) -> Response:
     If the server is not running anything, this request will be run, so we know whether the server is healthy.
     """
 
+    if (
+        get_bool_env_var("SGLANG_DISABLE_GENERATION_HEALTHCHECK")
```
fzyzcjy: very tiny nit:

* use environ.py
* name maybe `SGLANG_ENABLE_HEALTH_API_GENERATION` or sth like that (and make it by default true)
    * maybe make it "enable" not "disable" since we can provide a "false" here
    * maybe in the name show we only affect /health and not /health_generate (my proposed name is also not good for this...)
* shall we still check other things, e.g. the gracefully_exit, Starting, etc, and just skip the real generation. b/c i guess the exiting/starting is good to check

fzyzcjy (approved):
LGTM w/ nits

python/sglang/srt/entrypoints/http_server.py
zhuzilin: fixed except the renaming to show the difference between `/health` and `/health_generate`.

PR #13338, opened 2025-11-15
[RL] re-abort_request when model_update_lock is still locked

zhuzilin:
## Motivation

In the origin implementation, there may be on-the-fly requests not aborted by the  first `abort_request`. This PR is adding a retry.

## Checklist


python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -1190,7 +1190,9 @@ def abort_request(self, rid: str = "", abort_all: bool = False):
     async def pause_generation(self):
         async with self.is_pause_cond:
             self.is_pause = True
-            self.abort_request(abort_all=True)
+            while self.model_update_lock.is_locked():
```
fzyzcjy: nit: add a comment mentioning we are using "model update" lock not for its model update semantics, but for its "there is ongoing req" semantics

fzyzcjy: if not locked, seems we will never call abort_request. is this safe? (I am not sure whether there are edge conditions)

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -1190,7 +1190,9 @@ def abort_request(self, rid: str = "", abort_all: bool = False):
...
+                self.abort_request(abort_all=True)
```
fzyzcjy: (optional) btw is it possible we await this (using Communicator mechanism), instead of fire-and-forget

fzyzcjy (commented):
LGTM w/ nits

python/sglang/srt/managers/tokenizer_manager.py
zhuzilin: fixed.

zhuzilin: I've added one abort_request outside the for-loop

python/sglang/srt/managers/tokenizer_manager.py
zhuzilin: yeah, this really makes sense, and the barrier now is that `abort_request` is called as a non-async function in many locations. we may need to have another async version of abort request, which may be a non-trivial refactor.

fzyzcjy: yes let's skip it now and may be add a todo comment instead

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -1191,6 +1191,10 @@ async def pause_generation(self):
         async with self.is_pause_cond:
             self.is_pause = True
             self.abort_request(abort_all=True)
+            # we are using the model_update_lock to check if there is still on-going requests.
+            while self.model_update_lock.is_locked():
+                self.abort_request(abort_all=True)
```
fzyzcjy:
```suggestion
                # TODO: maybe make it async instead of fire-and-forget
                self.abort_request(abort_all=True)
```

python/sglang/srt/utils/aio_rwlock.py
```diff
@@ -75,6 +75,9 @@ async def release_writer(self):
             # Wake up anyone waiting (readers or writers)
             self._cond.notify_all()
 
+    def is_locked(self):
+        return self._writer_active or self._readers > 0
```
fzyzcjy: question: shall we add a `async with self._lock` to protect the variables

fzyzcjy (approved):

PR #13339, opened 2025-11-15
Add SGLANG_ENABLE_REQ_POOL_LEAK_STRICT_CHECK to bypass mem leak check

zhuzilin:
## Motivation

This PR is adding an `SGLANG_ENABLE_REQ_POOL_LEAK_STRICT_CHECK` flag  to bypass mem leak check which occurs during  RL training.

Thank you for your time on reviewing  this PR :)

## Checklist


python/sglang/srt/utils/common.py
```diff
@@ -3646,3 +3646,15 @@ def get_current_device_stream_fast():
     if cached_device_index == -1:
         cached_device_index = torch.get_device_module().current_device()
     return torch.get_device_module().current_stream(cached_device_index)
+
+
+def raise_error_or_warn(obj, flag, counter_name, message, log_interval=1000):
+    if flag:
```
fzyzcjy: nit: flag -> enable_error (or sth like that)

python/sglang/srt/utils/common.py
```diff
@@ -3646,3 +3646,15 @@ def get_current_device_stream_fast():
...
+        raise ValueError(message)
+    else:
+        if not hasattr(obj, counter_name):
+            setattr(obj, counter_name, 0)
+        count = getattr(obj, counter_name)
```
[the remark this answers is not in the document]
fzyzcjy: yes i want to say the same

python/sglang/srt/utils/common.py
```diff
@@ -3646,3 +3646,15 @@ def get_current_device_stream_fast():
     if cached_device_index == -1:
         cached_device_index = torch.get_device_module().current_device()
     return torch.get_device_module().current_stream(cached_device_index)
+
+
+def raise_error_or_warn(obj, flag, counter_name, message, log_interval=1000):
+    if flag:
+        raise ValueError(message)
+    else:
+        if not hasattr(obj, counter_name):
```
fzyzcjy: not sure whether this is the best way (add hacky fields to existing obj), but may be acceptable

python/sglang/srt/environ.py
```diff
@@ -136,6 +136,7 @@ class Envs:
     SGLANG_LOG_GC = EnvBool(False)
     SGLANG_LOG_FORWARD_ITERS = EnvBool(False)
     SGLANG_DISABLE_REQUEST_LOGGING = EnvBool(False)
+    SGLANG_ENABLE_REQ_POOL_LEAK_STRICT_CHECK = EnvBool(True)
```
hnyls2002:
```suggestion
    SGLANG_ENABLE_STRICT_MEM_CHECK_DURING_IDLE = EnvBool(True)
```

hnyls2002 (changes requested):

python/sglang/srt/environ.py
hnyls2002: We will have a mem check while running, so please change the name following the suggestion. @zhuzilin

python/sglang/srt/utils/common.py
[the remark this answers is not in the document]
zhuzilin: fixed.

fzyzcjy (approved):

hnyls2002 (approved):

PR #13354, opened 2025-11-15
Fix qwen3vl blackwell backend compatibility

harvenstar:
Motivation

  Fixes Qwen3-VL model startup crashes on Blackwell GPUs (B200) with error "The 'fa3' backend is not
  supported on Blackwell GPUs". The issue prevented Qwen3-VL models from running on the latest Blackwell
   GPU architecture due to hardcoded FlashAttention 3 backend selection that bypassed hardware
  compatibility checks.

  Modifications

  1. VisionAttention platform detection: Modified _determine_attention_backend method in vision.py to
  check not is_blackwell() before selecting fa3 backend for compute capability 9.x GPUs
  2. Qwen3-VL backend selection: Changed from hardcoded attn_implementation="flash_attention_3" to None
  in qwen3_vl.py, allowing VisionAttention to automatically choose the optimal backend
  3. Backend parameter handling: Updated to pass qkv_backend=None instead of hardcoded "sdpa" when
  attn_implementation=None, ensuring full flexibility

  Accuracy Tests

  ✅ Startup test passed: Qwen3-VL-30B-A3B-Instruct successfully starts on Blackwell GPUs (tested on 4x
  B200 with TP=4)
  ✅ Backend selection verified: VisionAttention correctly selects triton_attn backend for Blackwell
  GPUs instead of unsupported fa3
  ✅ No regression: Other GPU architectures (H100, etc.) continue to use optimal backends without
  changes

  Benchmarking and Profiling

  Previous behavior: Model startup failure with fa3 backend error
  Current behavior: Successful startup using triton_attn backend on Blackwell GPUs
  Performance impact: No performance regression on existing hardware; enables Qwen3-VL usage on
  Blackwell architecture

  Checklist

  - Format your code according to the [Format code with pre-commit]
  - Add unit tests (Compatibility fix, tested with real hardware)
  - Update documentation (No user-facing API changes)
  - Provide accuracy and speed benchmark results (Startup test completed on target hardware)


yhyang201 (approved):
LGTM.

python/sglang/srt/layers/attention/vision.py
```diff
@@ -592,7 +592,7 @@ def _determine_attention_backend(self, passed_backend: Optional[str]) -> str:
             backend = passed_backend
         elif is_cuda():
             major, minor = get_device_capability()
-            if major == 9:
+            if major == 9 and not is_blackwell():
```
Qiaolin-Yu: I don't think we should add `and not is_blackwell()` here. If `is_blackwell` is True, the major will be 10 or 12?

harvenstar: Thanks for catching this! 👍 I've reverted the changes to vision.py and kept only the qwen3_vl.py fix. Tested successfully on B200 GPUs - Qwen3-VL-30B now starts without issues!

Qiaolin-Yu (approved):

PR #13377, opened 2025-11-16
enable piecewise cuda graph for prefill server

fjybiocs:
## Motivation
Piecewise CUDA graphs can be used to accelerate prefill computation. 

But reviously, cuda graph was always disabled for prefill server in disaggregation mode. This PR allows using cuda graph when piecewise cuda graph is enabled.

## Modifications

- Modified the logic in `server_args.py` to only disable cuda graph when `enable_piecewise_cuda_graph` is not set


ishandhanani (commented):
Can you share an example of you setting this and getting correct inference? It would be nice to get a basic functional and accuracy test

fjybiocs:
> Can you share an example of you setting this and getting correct inference? It would be nice to get a basic functional and accuracy test

Thanks for pointing this out! I'll add proper tests soon.

ishandhanani:
Ah yes I misunderstood. I see that now we simply disable cuda graphs for P if piecewise isn't turned on. This is a safe change

ShangmingCai (approved):
LGTM

PR #13393, opened 2025-11-17
[CI] fix triggered by a non-run-ci label

hnyls2002:
### `run-ci` label behavior update

- Existing workflows are only triggered by the `synchronize` event. The `labeled` event no longer triggers CI runs.
- The `run-ci` label is not automatically added to new pull requests.

### How to run CI for a pull request

1. The PR author opens a pull request.
2. An SGLang collaborator with permission `triage` or higher adds the `run-ci` label to the PR.
3. After the label is added, the PR author can:
   - Push new commits, or
   - Use the “re-run” / retrigger button to start the CI workflow.

### DIFF
--- a/.github/workflows/pr-gate.yml
+++ b/.github/workflows/pr-gate.yml
@@ -1,0 +1,96 @@
+on:
+  workflow_call:
+
+jobs:
+  pr-gate:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Fetch latest PR info
+        id: pr
+        if: github.event_name == 'pull_request'
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const pr = await github.rest.pulls.get({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: context.issue.number
+            });
+            core.setOutput("labels", JSON.stringify(pr.data.labels.map(l => l.name)));
+            core.setOutput("draft", pr.data.draft);
+            core.setOutput("user", pr.data.user.login);
+
+      - name: Block draft PR
+        if: github.event_name == 'pull_request' && fromJson(steps.pr.outputs.draft)
+        run: |
+          echo "PR is draft. Blocking CI."
+          exit 1
+
+      - name: Require run-ci label
+        if: github.event_name == 'pull_request'
+        run: |
+          labels='${{ steps.pr.outputs.labels }}'
+          echo "Labels: $labels"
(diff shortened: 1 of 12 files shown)


Fridge003 (approved):

ping1jing2 (approved):

.github/workflows/pr-gate.yml
```diff
@@ -0,0 +1,96 @@
...
+        if: github.event_name == 'pull_request'
+        run: |
+          labels='${{ steps.pr.outputs.labels }}'
+          echo "Labels: $labels"
+          if [[ "${{ contains(fromJson(steps.pr.outputs.labels), 'run-ci') }}" == "false" ]]; then
+            echo "Missing required label 'run-ci'."
+            exit 1
+          fi
+
+      - name: Enforce rate limit for low-permission actors
```
slin1237, line 40: for rate limiting, this means users without any permission can only run every 2 hours?

hnyls2002: Yes, this is mainly designed for CUDA CIs and AMD CIs. We can set hyperparameters for different CIs

hnyls2002: You can add inputs to the `pr-gate.yml` file and set them to 0 hours with Rust CIs.

slin1237: parameterize this would be helpful 
2 hours can be too long for rust component. 
we intend to add a GPU cluster for CI, along with current GPUs
worst case, if we see abuse of CI, we can simply remove the label

slin1237 (approved):

.github/workflows/pr-gate.yml
hnyls2002: Sure, thanks. For adjusting the pr-gate to each CI, please see #13436

ZailiWang: Agree that the adjustable parameter for CI rate should be applied.
Another suggestion is that: shall we have an rate limit exempt mechanism? Like re-using `high priority` or create a new label indicating the PR needs to be merged ASAP so that limit would not be applied. The label can only be tagged by SGL admin.

hnyls2002: No need for this label. Anyone with write privileges can trigger the CI for the PR and wouldn't be influenced by the rate limit. @ZailiWang

ZailiWang (approved):

HaiShaw (approved):

.github/workflows/pr-gate.yml
hnyls2002: https://github.com/sgl-project/sglang/pull/13491

PR #13394, opened 2025-11-17
Fix DeepSeekV31's structural tag trigger

Muqi1029:
## Motivation

As titled. To prevent LLMs generating tool calls with wrong name

Reference (can be seen in other models in the SGLang codebase as well):
image

## Checklist


Fridge003:
cc @JustinTong0323

Muqi1029:
> Hello author, are you experiencing issues when deploying DeepSeek V3.1 with SGLANG_TOOL_STRICT_LEVEL set to 1 or 2? May I ask if your purpose of using SGLANG_TOOL_STRICT_LEVEL or structural_tag is to ensure that the tool_calls returned by the DeepSeek V3.1 model are in complete JSON format?

@kongweiming Sorry for the late reply.

First of all, both env `SGLANG_TOOL_STRICT_LEVEL` and `structural_tag` are part of constrained decoding of SGLang, their purposes are to make LLMs generate user-defined format, here is about json format in the tool call.

Then, I want to illustrate this PR is only to fix some potential bugs about structural tag definition. I think it is not related about the env setting: SGLANG_TOOL_STRICT_LEVEL = 1 or 2.

JustinTong0323:
Could you help give some examples? Like in which input the model would generate a wrong function name? Thanks~

Muqi1029:
> Could you help give some examples? Like in which input the model would generate a wrong function name? Thanks~

@JustinTong0323 
Hi, let me try to persuade you why this PR is necessary.
-- --

**First, from the perspective of logic**:

The previous implementation inserted the tool name provided by the user directly into the trigger tokens. However, the model might still generate a slightly different tool_name. For example:

If you pass a tool named `-A`, the trigger becomes `<｜tool▁call▁begin｜>-A<｜tool▁sep｜>`. Unfortunately, the model may instead sample `<｜tool▁call▁begin｜>A<｜tool▁sep｜>` because `A` is easier to produce and the model is not strong at distinguishing `A` from `-A`. As a result, the trigger fails, and the model ends up returning a function call name `A`—even though no such tool was provided.

After this PR is merged, once the model generates `<｜tool▁call▁begin｜>`, the grammar is activated immediately. The structural tag enforces the correct tool name, preventing these sampling errors.


-- --

**Secondly, I want to refer SGLang itself and XGrammar to endorce this PR**:
Kimi-K2 from sglang:
https://github.com/sgl-project/sglang/blob/a0dde90af6064e172c3ba6e0dfa3a30cc2fa8ac5/python/sglang/srt/function_call/kimik2_detector.py#L229-L239

Llama3.2 from sglang:
https://github.com/sgl-project/sglang/blob/a0dde90af6064e172c3ba6e0dfa3a30cc2fa8ac5/python/sglang/srt/function_call/llama32_detector.py#L83-L88

XGrammar official doc to show how to use structural tag about tool call: https://xgrammar.mlc.ai/docs/tutorials/structural_tag.html
image

JustinTong0323 (approved):

PR #13427, opened 2025-11-17
[Feat][NVFP4] Enable NVFP4 MoE for Qwen series models (eg. Qwen3-Next)

samuellees:
## PR Dependency
- [x] [Merged] Depend on this [PR13489](https://github.com/sgl-project/sglang/pull/13489)

## Motivation

Enable NVFP4 MoE for Qwen series models (eg. Qwen3-Next) on Blackwell GPUs

## TODO
- [x] Add unit test
- [x] Add command for reproduce accuracy results
- [x] Refactor this PR (Co-work with @kaixih  [PR13556](https://github.com/sgl-project/sglang/pull/13556))

## Accuracy Tests
```bash
# Qwen3-Next: NVFP4 linear + NVFP4 MoE + FP8 Attention
export SGL_ENABLE_JIT_DEEPGEMM=false
python3 -m sglang.launch_server --model-path qwen3-next-80b-a3b-instruct-nvfp4-all --chunked-prefill-size 16384 --max-prefill-tokens 16384 --max-running-requests 512 --tp-size 4 --ep-size 4 --mem-fraction-static 0.7 --cuda-graph-bs 1 2 4 8 16 32 64 128 256 512 1024 --disable-radix-cache --log-level info --host 0.0.0.0 --port 8001 --random-seed 0  --quantization modelopt_fp4   --kv-cache-dtype fp8_e4m3  --moe-runner-backend flashinfer_trtllm --attention-backend trtllm_mha  --mamba-ssm-dtype bfloat16
```

|gsm8k|      3|flexible-extract|     8|exact_match|↑  |0.9583|±  |0.0055|
|     |       |strict-match    |     8|exact_match|↑  |0.7324|±  |0.0122|

```bash
# Qwen3-Next: FP8 linear + FP8 MoE + FP8 Attention
export SGL_ENABLE_JIT_DEEPGEMM=false
python3 -m sglang.launch_server --model-path Qwen3-Next/Qwen3-Next-80B-A3B-Instruct-FP8 --chunked-prefill-size 16384 --max-prefill-tokens 16384 --max-running-requests 512 --tp-size 4 --ep-size 4 --mem-fraction-static 0.7 --cuda-graph-bs 1 2 4 8 16 32 64 128 256 512 1024 --disable-radix-cache --log-level info --host 0.0.0.0 --port 8001 --random-seed 0  --quantization fp8    --kv-cache-dtype fp8_e4m3  --moe-runner-backend flashinfer_trtllm --attention-backend trtllm_mha  --mamba-ssm-dtype bfloat16 
```

|gsm8k|      3|flexible-extract|     8|exact_match|↑  |0.9575|±  |0.0056|
|     |       |strict-match    |     8|exact_match|↑  |0.8370|±  |0.0102|

I also reproduced the Qwen3-30B-A3B NVFP4 accuracy result as it was shown in [PR13556](https://github.com/sgl-project/sglang/pull/13556)
```bash
Accuracy: 0.899
Invalid: 0.000
Latency: 29.658 s
Output throughput: 5636.438 token/s
```

## Checklist


python/sglang/srt/layers/moe/fused_moe_triton/layer.py
```diff
@@ -1083,7 +1083,17 @@ def forward(self, hidden_states: torch.Tensor, topk_output: TopKOutput):
 
         hs_fp4, hs_scale_linear = self._quantize_hidden_states_fp4(hidden_states)
 
-        router_logits = router_logits.to(torch.float32)
+        routing_method_type = getattr(
```
kaixih: We should pass `self.routing_method_type` directly to the API call.
`routing_method_type` is always present in the FusedMoE layer and assigning a default value is risky. If we leave it as `None`, the TRT-LLM MoE call will raise an error.

python/sglang/srt/layers/moe/fused_moe_triton/layer.py
```diff
@@ -1083,7 +1083,17 @@ def forward(self, hidden_states: torch.Tensor, topk_output: TopKOutput):
...
+            self, "routing_method_type", RoutingMethodType.DeepSeekV3
+        )
+        if routing_method_type == RoutingMethodType.DeepSeekV3:
+            router_logits = router_logits.to(torch.float32)
```
b8zhong: tiny: are the router logits already in fp32 on deepseek here (i think teh output of `dsv3_router_gemm` may always be fp32, perhaps we can simplify this check

b8zhong (approved):
https://github.com/sgl-project/sglang/blob/b638abbae6b28eb7052172441842a889a0d01ace/python/sglang/srt/server_args.py#L1102-L1124

LGTM, just try to upd this checker in server args when qwen3 + fp4 quant (otherwise, it may not be known that it is better)

python/sglang/srt/layers/moe/fused_moe_triton/layer.py
samuellees: ok, I'll give a try and let's see if the ds CI passed ^ ^

kaixih (approved):
Just nit: maybe we can simpifiy the test name to `test_qwen3_trtllm_fp4.py`.

kaixih:
> > Just nit: maybe we can simpifiy the test name to `test_qwen3_trtllm_fp4.py`.
> 
> Done. It will be updated in the next commit.
> 
> BTW, do you have any idea why [this test case](https://github.com/sgl-project/sglang/actions/runs/19524617641/job/55902215055?pr=13427) failed? I really don't know why ^ ^ @kaixih

I am not sure but the test is test/srt/test_mla_flashinfer.py so I don't think this PR causes it. We should fix the test error in a separate PR. How do you think? @Fridge003

PR #13432, opened 2025-11-17
[Feat] Add received_time in serving_base

zhanghaotong:
## Motivation
While using SGLang, we observed a consistent time gap between the arrival time of an incoming request and the recorded trace creation time. This gap increases with the input length—for example, it reaches approximately 70 ms for inputs of 100K tokens. Upon investigation, we identified that the `_convert_to_internal_request` function in serving_chat.py performs encoding logic that introduces significant latency. This PR addresses the issue by accurately recording the precise request arrival time, thereby correcting both the trace start time and end-to-end metrics timing.

## Modifications

This PR records a timestamp at the entry point of handle_request in serving_base.py and passes it to the `generate_request` method in tokenizer_manager.py to be used as the request's creation time.

## Checklist


python/sglang/srt/entrypoints/openai/serving_base.py
```diff
@@ -88,6 +88,8 @@ async def handle_request(
         """Handle the specific request type with common pattern
         If you want to override this method, you should be careful to record the validation time.
         """
+        created_time_ts = time.time()
```
slin1237: use perf_counter

python/sglang/srt/managers/io_struct.py
```diff
@@ -739,6 +742,8 @@ class EmbeddingReqInput(BaseReq):
     modalities: Optional[List[str]] = None
     # Validation step duration
     validation_time: Optional[float] = None
+    # Request arrival time
+    created_time_ts: Optional[float] = None
```
slin1237: keep it consistent
received_time

slin1237 (changes requested):

python/sglang/srt/entrypoints/openai/serving_base.py
zhanghaotong: Although I know that perf_counter can represent time intervals more precisely, what I actually need here is a timestamp to serve as the start time for the trace. Moreover, the `created_time` in the tokenizer-manager that we're replacing is also a timestamp. 🤔

python/sglang/srt/managers/io_struct.py
zhanghaotong: OK, I'll rename it.

python/sglang/srt/entrypoints/openai/serving_base.py
slin1237: if anyone else used this field in the future would encounter a problem because there dan be NTP adjustment during request processing.

sufeng-buaa: `time.time()` and `time.perf_counter()` are actually equivalent to `clock_gettime(CLOCK_REALTIME)` and `clock_gettime(CLOCK_MONOTONIC)`, respectively. Although `CLOCK_REALTIME` can be affected by NTP adjustments, it is explicitly more debug-friendly for distributed systems. In contrast, I think `CLOCK_MONOTONIC` is better suited for timestamps that affect scheduling policies. A feasible solution could be, as done in the `tokenizer_manager.py`, to append the `_perf` suffix to timestamps variable obtained via time.perf_counter(). What do you think?

sufeng-buaa: and add comments to explicitly mark

zhanghaotong: > `time.time()` and `time.perf_counter()` are actually equivalent to `clock_gettime(CLOCK_REALTIME)` and `clock_gettime(CLOCK_MONOTONIC)`, respectively. Although `CLOCK_REALTIME` can be affected by NTP adjustments, it is explicitly more debug-friendly for distributed systems. In contrast, I think `CLOCK_MONOTONIC` is better suited for timestamps that affect scheduling policies. A feasible solution could be, as done in the `tokenizer_manager.py`, to append the `_perf` suffix to timestamps variable obtained via time.perf_counter(). What do you think?

Aggreed, we can add both `received_time` and `received_time_perf`.

python/sglang/srt/managers/io_struct.py
sufeng-buaa: Timestamps and durations should be distinguished. I think the name `validation_time` is ambiguous here.

slin1237 (approved):

hnyls2002 (approved):

PR #13460, opened 2025-11-17
CI: fix NFS EBUSY error in PR test workflow

alisonshao:
## Summary
Fixes NFS "EBUSY: resource busy or locked" errors during `actions/checkout@v4` on self-hosted 1-gpu-runner jobs.

## Problem
Self-hosted runners on NFS-backed storage encounter checkout failures when:
- Lingering Python processes from previous test runs hold open file handles
- NFS creates `.nfs*` temporary files for open-but-deleted files
- `actions/checkout@v4` attempts to delete workspace but fails with EBUSY

Example error:
```
Error: File was unable to be removed Error: EBUSY: resource busy or locked, 
unlink '/public_sglang_ci/runner-l1c-gpu-1/_work/sglang/sglang/python/sglang/logs/.nfs2a00b5d9a59622fb00000023'
```

## Solution
Add a pre-checkout cleanup step that:
1. Kills lingering processes (python, sglang, bench_serving, pytest)
2. Force-closes any open file handles in `$GITHUB_WORKSPACE`
3. Waits 2 seconds for NFS to release locks
4. Removes all workspace files including `.nfs*` files

## Testing
This fix applies to 9 jobs using `1-gpu-runner` in `pr-test.yml`.

## Checklist
- [x] Added cleanup step before all `actions/checkout@v4` calls in 1-gpu-runner jobs
- [x] Cleanup uses `|| true` to not fail if no processes found
- [x] 2-second wait added for NFS lock release

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -331,15 +331,6 @@ jobs:
    if: needs.check-changes.outputs.multimodal_gen == 'true'
    runs-on: 1-gpu-runner
    steps:
-     # - name: Cleanup before checkout
-     #   run: |
-     #     # Kill any lingering Python processes that might be holding files
-     #     pkill -9 -f "python.*sglang" || true
-     #     pkill -9 -f "pytest" || true
-     #     sleep 2
-     #     # Remove logs directory
-     #     rm -rf "$GITHUB_WORKSPACE/python/sglang/logs" || true
-
      - name: Checkout code
        uses: actions/checkout@v4
 
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -210,14 +210,34 @@ jobs:
     steps:
       - name: Kill lingering processes and cleanup workspace
         run: |
-          # Kill all Python processes that might hold file handles
-          pkill -9 -f "python" || true
-          # Give NFS time to release locks
-          sleep 5
-          # Force remove any .nfs* files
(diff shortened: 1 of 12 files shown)


.github/workflows/pr-test.yml
```diff
@@ -208,6 +208,15 @@ jobs:
     env:
       RUNNER_LABELS: 1-gpu-runner
     steps:
+      - name: Kill lingering processes and cleanup workspace
+        run: |
+          pkill -f "python.*sglang" || true
```
Kangyan-Zhou: Can we move this command to a new stage after Run Test so that this script can stored in a file in the repo instead?

Also do we need to run rm -rf "$GITHUB_WORKSPACE"/* || true? IIUC this will remove the code repo as well might slow down the code checkout time, but I'm not sure about the performance implication

python/sglang/multimodal_gen/runtime/utils/performance_logger.py
```diff
@@ -10,26 +10,36 @@
...
 from dateutil.tz import UTC
 
-LOG_DIR = os.environ.get("SGLANG_PERF_LOG_DIR")
-if LOG_DIR:
-    LOG_DIR = os.path.abspath(LOG_DIR)
-else:
-    project_root = os.path.abspath(os.path.join(os.path.dirname(__file__), "../../../"))
-    LOG_DIR = os.path.join(project_root, "logs")
+
+def _is_in_ci() -> bool:
```
Kangyan-Zhou: Can we use an existing function in python/sglang/utils.py is_in_ci?

PR #13477, opened 2025-11-18
diffusion: fix user provided params was overriding

Songrui625:
## Motivation
User provided parameters will be ignored due to the sampling params merge logic. If we want to generate a 832x480 video using model Wan2.1-T2V-14B-Diffusers, it will instead produce a 1280×720 video, which is not what we intended.

## Modifications
Fix this issue by overriding the subclass default parameters whenever the user provides their own values.

## Checklist

### DIFF
--- a/python/sglang/multimodal_gen/configs/sample/base.py
+++ b/python/sglang/multimodal_gen/configs/sample/base.py
@@ -404,17 +404,12 @@ class SamplingParams:
         """
         Merges parameters from a user-provided SamplingParams object.
 
-        This method updates the current object with values from `user_params`,
-        but skips any fields that are explicitly defined in the current object's
-        subclass. This is to preserve model-specific optimal parameters.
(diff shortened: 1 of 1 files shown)


python/sglang/multimodal_gen/configs/sample/base.py
```diff
@@ -432,7 +427,7 @@ def _merge_with_user_params(self, user_params):
                 if field_name != "output_file_name"
                 else user_params.output_file_path is not None
             )
-            if is_user_modified and field_name not in subclass_defined_fields:
+            if is_user_modified:
```
mickqian, line 430: are all the sampling params allowed to be overridden after this PR? for some critical ones, a warning/info better be shown to the user, about the recommended values and possible consequences

mickqian (changes requested):

PR #13487, opened 2025-11-18
[Performance] Replace preprocess_video logic from GLM  multimodal processor with transformer impl for speed up (up to 27% faster) and addressing OOM (up to 50x improvements)

byjiang1996:
## Motivation

As reported by GLM team @zRzRzRzRzRzRzR , when processing large videos (>100MB), GLM-4xV model is extremely slow and is very likely to OOM.

## Modifications

The issue is `preprocess_video` logic implemented in SGLang is not as optimized as the one implemented in transformer: [[1]](https://github.com/huggingface/transformers/blob/266d3b05684ed5968c613f00b208909f03c72c0e/src/transformers/models/glm4v/video_processing_glm4v.py#L151) [[2]](https://github.com/huggingface/transformers/blob/266d3b05684ed5968c613f00b208909f03c72c0e/src/transformers/models/qwen3_vl/video_processing_qwen3_vl.py#L174)

I think ideally we should reuse the preprocess_video logic implemented by model owners in transformer if possible instead of duplicating the logic in SGLang which may easily get out-dated and cause tricky performance issues

## Accuracy Tests

UT passes for GLM41v and Qwen3-VL

image
image

## Benchmarking and Profiling
```
from sglang.srt.entrypoints.http_server import launch_server
from sglang.srt.server_args import prepare_server_args
if __name__ == "__main__":
    # Simulate CLI arguments (excluding the script name)
    args = [
        "--model-path",
        # "Qwen/Qwen3-VL-8B-Instruct",
        "zai-org/GLM-4.1V-9B-Thinking",
        "--mm-attention-backend",
        "fa3",
        "--attention-backend",
        "fa3",
    ]
    server_args = prepare_server_args(args)
    launch_server(server_args)
```

```
curl_command = f"""
curl -s http://localhost:{30000}/v1/chat/completions \\
  -H "Content-Type: application/json" \\
  -d '{{
    "model": "default",
    "messages": [
      {{
        "role": "user", 
        "content": [
          {{
            "type": "video_url", 
            "video_url": {{
              "url": "/tmp/data/1O1TfTrEnss.mp4" # 531MB in size: comes from videomme benchmark's videos_chunked_01.zip
            }}
          }},
          {{
            "type": "text", 
            "text": "What is the video"
          }}
        ]
      }}
    ],
    "temperature": "0",
    "max_completion_tokens": "1000",
    "max_tokens": "1000"
  }}'
"""

response = subprocess.check_output(curl_command, shell=True).decode()
print(response)
```
### Qwen3-VL-8B-Instruct: 31.2s -> 24.5s - 27.3% faster
image

->

image

### zai-org/GLM-4.1V-9B-Thinking: OOMKilled after 3min20s -> 27.5s
image
image

->

image

### Memory Pressure: 200GB+ to 4GB - up to 50x improvements

image

To process /tmp/data/1O1TfTrEnss.mp4, memory consumption drops from 200GB+ to 4GB (`50x improvements`)

image

## Checklist


python/pyproject.toml
```diff
@@ -63,6 +63,7 @@ dependencies = [
   "timm==1.0.16",
   "torch_memory_saver==0.0.9",
   "torch==2.8.0",
+  "torchcodec==0.7.0",
```
byjiang1996: It is required by transformer's preprocess_video utils

JustinTong0323: why the version is mismatch with below?

byjiang1996: image

From the [official wiki](https://github.com/meta-pytorch/torchcodec)
- torch 2.8.0 maps to torchcodec 0.7.0
- torch 2.9.0 maps to torchcodec 0.8.0

yhyang201:
1. Could you please help verify whether Qwen2.5-VL can process video inputs correctly after the recent modifications?
2. The pull request at https://github.com/sgl-project/sglang/pull/13565 adds an OpenCV backend with good performance, but it still relies on sglang’s own video-processing capability. Given this, do you think we should retain sglang’s built-in video processing (while refactoring the code to make it clearer and more consistent), and default to using the `transformers` `preprocess_video` logic? I’d really appreciate hearing your thoughts, as they’re very valuable.

Thank you!

yuan-luo (approved):
LGTM. Thanks for the enhancement.

yhyang201 (approved):

python/pyproject.toml
```diff
@@ -89,7 +90,6 @@ diffusion = [
     "moviepy>=2.0.0",
     "cloudpickle",
     "remote-pdb",
-    "torchcodec==0.5.0",
```
byjiang1996, line 92: Already exists in main pip requirement. Not needed again here

JustinTong0323 (approved):

PR #13488, opened 2025-11-18
fix: malformed KV events for NVIDIA Dynamo

PeaBrane:
## Motivation

SGL was not sending the correct KV events to NVIDIA Dynamo, which caused inconsistencies in the Dynamo indexer (see: https://github.com/ai-dynamo/dynamo/issues/4394
).
The issue appeared most clearly when repeat tokens were generated: SGL would incorrectly trigger self-referential blocks.

The root cause was that SGL sent content (local block) hashes instead of sequence-level hashes. As a result, different blocks within the same sequence that happened to contain identical tokens were mistakenly treated as shared blocks, leading to incorrect deduplication and self-references.

## Modifications

Changes
1. Fixed KV event emission to use sequence-aware hashes

- Switched from content-only hashes (hash(tokens)) to sequence-aware hashes (hash((parent_seq_hash, content_hash)))
- Ensures unique hashes for repeating token patterns at different positions
- Prevents RefCell borrow conflicts in Dynamo's radix tree indexer

2. Removed incorrect event emissions on node split

- Removed event emissions from _split_node() (structural change only, not cache content change)
- Prevents duplicate/malformed events that break Dynamo's radix tree

3. Implemented lazy sequence hash computation

- Added sequence_hash: Optional[List[int]] field to TreeNode (one hash per page)
- Sequence hashes computed lazily during event emission, using parent's last sequence hash as base
- No overhead when enable_kv_cache_events=False

4. Added tests

- Tests verify sequence_hash storage, repeating tokens get different hashes, and correct splitting behavior

## Checklist


ishandhanani (approved):

test/srt/test_radix_cache_unit.py
```diff
@@ -658,6 +658,100 @@ def test_advanced_prefix_match_with_node_splits(self):
                 result_branch = cache.match_prefix(RadixKey(seq2))
                 torch.testing.assert_close(result_branch.device_indices, val2)
 
+    def test_sequence_hash_storage(self):
```
xiezhq-hermann: can you make it more straightforward like test dynamo events?

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -93,6 +96,8 @@ def __init__(self, id: Optional[int] = None):
         self.host_value: Optional[torch.Tensor] = None
         # store hash values of each pages
         self.hash_value: Optional[List[str]] = None
+        # store sequence hashes for each page (position-aware hashes)
+        self.sequence_hash: Optional[List[int]] = None
```
xiezhq-hermann: isn't the `sequence_hash` duplicated with the existing `hash_value`?

xiezhq-hermann: I would suggest to reuse some codes around `hash_value`, which already provided quite a few functions this PR is reimplementing.

PeaBrane: Nice catch. I think I completely missed that attribute

PeaBrane: I refactored some of the common components into `hicache_storage.py`

test/srt/test_radix_cache_unit.py
PeaBrane: hmm, not sure how to simplify it. What do you have in mind?

PeaBrane:
Hi @xiezhq-hermann @hnyls2002 do you guys mind taking a second look at this? Would be nice to potentially merge this soonish

python/sglang/srt/mem_cache/hicache_storage.py
```diff
@@ -30,6 +33,79 @@ def get_hash_str(token_ids: List[int], prior_hash: str = None) -> str:
...
+    """
+    # Take first 16 hex chars to get 64-bit value
+    uint64_val = int(hash_str[:16], 16)
+    # Convert to signed int64 range [-2^63, 2^63-1]
+    if uint64_val >= 2**63:
+        return uint64_val - 2**64
+    return uint64_val
+
+
+def compute_node_hash_values(node: "TreeNode", page_size: int) -> List[str]:
```
xiezhq-hermann: I would suggest to move this function to `radix_cache.py` since we will be using it there and that way we keep the reference to nodes there as well

python/sglang/srt/mem_cache/hicache_storage.py
```diff
@@ -30,6 +33,79 @@ def get_hash_str(token_ids: List[int], prior_hash: str = None) -> str:
...
+
+        # Use SHA256-based chaining via get_hash_str
+        hash_val = get_hash_str(page_tokens, prior_hash=parent_hash)
+        hash_values.append(hash_val)
+        parent_hash = hash_val
+
+    return hash_values
+
+
+def split_node_hash_value(
```
xiezhq-hermann: same applies to this function, consider moving it to `radix_cache.py`

python/sglang/srt/mem_cache/hicache_storage.py
```diff
@@ -30,6 +33,79 @@ def get_hash_str(token_ids: List[int], prior_hash: str = None) -> str:
     return hasher.hexdigest()
 
 
+def hash_str_to_int64(hash_str: str) -> int:
```
xiezhq-hermann, line 33: we can keep this function in this file serving as a util for `get_hash_str`

ishandhanani (approved):

ishandhanani:
@Swipe4057 unfortunetly we released Dynamo 0.7.0 already. But you should be able to technically use this by pip installing ai-dynamo and then installing newer versions of sglang

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -596,8 +661,10 @@ def _split_node(self, key: RadixKey, child: TreeNode, split_len: int):
         child.value = child.value[split_len:]
         new_node.parent.children[self.get_child_key_fn(key)] = new_node
 
-        self._record_store_event(new_node)
```
sligen: Remove self._record_remove_event(child)  at line 652  from _split_node()
Since _split_node() is an internal restructuring operation, it must not emit KV change events.

PR #13490, opened 2025-11-18
[Tiny ci fix] Fix test isolation issues in test_model_hooks.py

BBuf:
## Motivation

Follow https://github.com/sgl-project/sglang/pull/13450

https://github.com/sgl-project/sglang/actions/runs/19435235373/job/55670398565?pr=13374

This PR fixes test isolation issues in test_model_hooks.py where global state pollution was causing intermittent test failures in CI environments.

The original implementation used a global variable to track hook calls, which caused test interference when multiple tests ran concurrently. This led to flaky test behavior in CI environments where test execution order could vary.

@Carlomus cc @hnyls2002 

## Checklist

- [ ] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.ai/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.ai/developer_guide/contribution_guide.html#benchmark-the-speed).
- [ ] Work with maintainers to merge your PR. See the [PR Merge Process](https://github.com/sgl-project/sglang/blob/main/.github/MAINTAINER.md#pull-request-merge-process)

PR #13523, opened 2025-11-18
Fixing hooks test by adding more realistic hook module to import

Carlomus:
Fixing CI/test erro introduced previously (see https://github.com/sgl-project/sglang/pull/13490)

> Follow https://github.com/sgl-project/sglang/pull/13450
>https://github.com/sgl-project/sglang/actions/runs/19435235373/job/55670398565?pr=13374
>This PR fixes test isolation issues in test_model_hooks.py where global state pollution was causing intermittent test failures in CI environments.
>The original implementation used a global variable to track hook calls, which caused test interference when multiple tests ran concurrently. This led to flaky test behavior in CI environments where test execution order could vary.

### DIFF
--- a/test/srt/hooks_test_data.py
+++ b/test/srt/hooks_test_data.py
@@ -1,0 +1,33 @@
+import contextvars
+
+# Context-local recorder; default is None
+_hook_recorder = contextvars.ContextVar("hook_recorder", default=None)
+
+
+def set_recorder(recorder):
+    """Set the recorder callable for this context; return token for reset."""
+    return _hook_recorder.set(recorder)
+
+
+def reset_recorder(token):
+    """Reset the recorder to its previous value."""
+    _hook_recorder.reset(token)
+
+
+def dummy_hook_factory(config):
+    """Factory that returns a forward hook capturing a tag from config."""
+    tag = config.get("tag", "default")
+
+    def hook(module, inputs, output):
+        recorder = _hook_recorder.get()
+        if recorder is not None:
+            recorder(
+                {
+                    "module_type": type(module).__name__,
+                    "tag": tag,
+                    "shape": tuple(output.shape),
+                }
+            )
+        return output
+
+    return hook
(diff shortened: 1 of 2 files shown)


test/srt/hooks_test_data.py
```diff
@@ -0,0 +1,33 @@
+import contextvars
```
hnyls2002, line 1: Why introduce a new file? Can you put this data file inside `test_model_hook.py`?

hnyls2002 (changes requested):

test/srt/hooks_test_data.py
Carlomus: `hook_factory` is resolved via `importlib.import_module` from a string. If the hooks and the tests live in the same file and that file is executed as `__main__`, Python ends up loading it twice: once as `__main__` and once as `sglang.test.srt.test_model_hooks`. Those are two distinct module objects, which means duplicated module-level state (including the `ContextVar` / recorder). The tests then manipulate one module’s state while the hooks run against the other, so hook calls never show up in the assertions.

 When I designed the hook attachment mechanism, I always assumed a separation between “where the model lives” and “where the hooks live”. To keep that behaviour in tests, we either:

* share global state across tests (which is brittle and already caused flaky/broken tests), or
* simulate the expected setup and keep the hook implementation in a separate module from the test harness.

This change chooses the second option, one file acts as the hook module and  the other is just the test harness that configures and asserts on it.

hnyls2002: If that is the reason, please hardcode the test data file in your `test_model_hooks.py` (e.g., in string format) and write to a tmp file during test. We do not want keep a data file under the `test/` folder.

hnyls2002 (changes requested):

hnyls2002:
Hi @Carlomus, thanks for the fix. The test this PR repairs (test_model_hooks.py) was disabled in #13450 and later removed entirely from the test suite in #23305, so there is no longer a target for this change. Closing as obsolete - please reopen if I've missed something.

PR #13582, opened 2025-11-19
use class path string as custom logit processor parameter

pengye91:
## Motivation

1. For built in custom logit processors, instead of importing from sglang package, which requires `sglang` as a dependency for clients, use class path string as the extra parameter.

now we can use the `custom_logit_processor` parameter in client codes without `import sglang` like this:
```python
import openai
from rich.pretty import pprint

client = openai.Client(base_url="http://127.0.0.1:30000/v1", api_key="*")
    model="deepseek-ai/DeepSeek-R1",
    messages=[
        {
            "role": "user",
            "content": "Question: Is Paris the Capital of France?",
        }
    ],
    max_tokens=1024,
    extra_body={
        "custom_logit_processor": "sglang.srt.sampling.custom_logit_processor:DeepSeekR1ThinkingBudgetLogitProcessor",
        "custom_params": {
            "thinking_budget": 512,
        },
    },
)
pprint(response)
```

2. Raise a `ValueError` at request initialization if the passed `custom_logit_processor` parameter is not valid, instead of raising a `RuntimeError` which will directly kill the server.
3. Add a `validate` method in `GenerateReqInput` class for all possible validation process.

## Modifications

1. Added a `validate` method in `GenerateReqInput` class for all possible validation process.
2. Added `_validate_custom_logit_processor` method in `GenerateReqInput` class to validate if the `custom_logit_processor` parameter is valid.
3. Added a `from_class_path` method in `CustomLogitProcessor` class to load the logit processor from class path string.
5. Keep the old custom logit processor loading method for compatibility and for not built-in logit processors.
6. Added related test cases.
7. Added related docs.

## Checklist


python/sglang/srt/sampling/custom_logit_processor.py
```diff
@@ -40,7 +41,48 @@ def to_str(cls) -> str:
     @classmethod
     def from_str(cls, json_str: str):
         """Deserialize a callable function from a JSON string."""
-        return _cache_from_str(json_str)()
+        try:
+            return cls.from_class_path(json_str)
```
merrymercy: cache the call

python/sglang/srt/managers/io_struct.py
```diff
@@ -526,6 +532,26 @@ def _normalize_custom_logit_processor(self, num):
                 "Cannot use list custom_logit_processor with parallel_sample_num > 1"
             )
 
+    def _validate_single_custom_logit_processor(self, clp: str):
```
merrymercy, line 535: cache the call result?

merrymercy (changes requested):

python/sglang/srt/managers/io_struct.py
pengye91: added `lru_cache`  on a new internal function `_cache_from_class_path`. I think we don't need to cache the result here any more.

python/sglang/srt/sampling/custom_logit_processor.py
pengye91: create a new method `_cache_from_class_path` with `lru_cache` on it.

PR #13595, opened 2025-11-19
feat: support execute customized tasks in model workers for second-level  weight synchronization  in RL systems

chaokunyang:
## Motivation

[Awex](https://github.com/inclusionAI/asystem-awex)  is a high-performance RL training-inference **weight synchronization framework**, designed to enable **second-level parameter updates** for **Trillion-parameter models** from training to inference in RL workflows. It minimizes iteration latency, ensuring rollout phases consistently use the latest model.

Here are some performance result for reference:
| Weight Parameter Scale | Weight Data Size | Awex NCCL Transmission Time | Awex RDMA Transmission Time |
| --- | --- |  --- | --- |
| 10B | 31GB | 0.8S | 0.5S |
| 100B | 191GB |9S | 3.2S |
| 1000B | 1000GB (FP8) | 20S | 6S |

Awex support megatron and sglang. But sglang support requires the ability to **execute initialization and weight‑update functions directly inside all SGLang model workers**. To support this, Awex integrates with SGLang by calling a distributed RPC‑style API that runs a **user‑defined function on every model worker process** with provided model context.

SGLang currently does not expose a public API for executing custom code inside all model workers. `collective_rpc` are close, but lacks of model context, and don't return result from all model workers.
So this PR introduces a new API `execute_task_in_model_worker`, enabling users and downstream systems (such as Awex) to run customized Python functions within SGLang’s model worker processes in a safe and controlled manner.

This API unblocks Awex integration and enables broader use cases such as:
- distributed weight updates  
- custom in‑worker initialization logic  
- advanced model‑side instrumentation  
- system‑level RL workflows requiring synchronized model‑side execution  

## Modifications

This PR adds a new method:

```python
def execute_task_in_model_worker(self, fn, **kwargs):
    if not self._initialized:
        raise RuntimeError("Engine not initialized. Call `initialize` first.")
    if self.node_rank != 0:
        raise RuntimeError(
            f"Non-zero rank node {self.rank_coordinate} is not allowed to "
            f"execute task in model workers"
        )
    return self._sgl_engine.execute_task_in_model_worker(fn, **kwargs)
```

Key modifications:

- Added the execute_task_in_model_worker API to the engine wrapper.
- Ensures it can only be invoked after initialization and only from node rank 0.
- Delegates execution to the underlying SGLang engine, which broadcasts the function to all model workers.
- Enables custom function execution across workers without modifying the core scheduling mechanism.
- No existing SGLang behavior is changed; this PR only adds capability.

## Accuracy Tests

This change does not affect model outputs or inference logic.
No changes to kernel or forward pass behavior.

To verify correctness:

- Confirmed that user-defined functions run correctly on all model workers.
- Verified that weight‑update functions produce consistent results across workers.

## Benchmarking and Profiling

This feature does not modify inference kernels, batching logic, or scheduling.
No performance regressions observed.

The overhead of executing a function across workers is negligible and only occurs when explicitly invoked by the user.

## Checklist

### DIFF
--- a/python/sglang/srt/entrypoints/engine.py
+++ b/python/sglang/srt/entrypoints/engine.py
@@ -27,7 +27,7 @@ import random
 import signal
 import threading
 import time
-from typing import AsyncIterator, Dict, Iterator, List, Optional, Tuple, Union
+from typing import AsyncIterator, Callable, Dict, Iterator, List, Optional, Tuple, Union
 
 import zmq
 
@@ -425,6 +425,21 @@ class Engine(EngineBase):
             "internal_states": internal_states,
             "version": __version__,
         }
+
+    async def async_execute_task_in_model_worker(
+        self, task_func: Callable, **kwargs
+    ) -> List:
+        """Execute a task on every model worker subprocess"""
+        return await self.tokenizer_manager.execute_task_in_model_worker(
+            task_func, **kwargs
+        )
+
+    def execute_task_in_model_worker(self, task_func: Callable, **kwargs) -> List:
+        """Execute a task on every model worker subprocess"""
+        loop = asyncio.get_event_loop()
+        return loop.run_until_complete(
+            self.tokenizer_manager.execute_task_in_model_worker(task_func, **kwargs)
+        )
 
     def init_weights_update_group(
         self,
--- a/python/sglang/srt/managers/io_struct.py
+++ b/python/sglang/srt/managers/io_struct.py
@@ -21,7 +21,7 @@ import uuid
 from abc import ABC
 from dataclasses import dataclass, field
 from enum import Enum
-from typing import TYPE_CHECKING, Any, Dict, List, Literal, Optional, Union
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Literal, Optional, Union
 
 from sglang.srt.lora.lora_registry import LoRARef
 from sglang.srt.managers.schedule_batch import BaseFinishReason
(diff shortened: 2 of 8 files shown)


new-TonyWang:
I am using awex and I cannot run sglang properly. 
```
File "/inspire/hdd/project/qianghuaxuexi/public/wty/sglang/python/sglang/srt/managers/data_parallel_controller.py", line 31, in <module>
    from sglang.srt.managers.io_struct import (
  File "/inspire/hdd/project/qianghuaxuexi/public/wty/sglang/python/sglang/srt/managers/io_struct.py", line 1589, in <module>
    _check_all_req_types()
  File "/inspire/hdd/project/qianghuaxuexi/public/wty/sglang/python/sglang/srt/managers/io_struct.py", line 1582, in _check_all_req_types
    raise ValueError(f"{name} is not a subclass of BaseReq or BaseBatchReq.")
ValueError: ModelWorkerTaskOutput is not a subclass of BaseReq or BaseBatchReq.
```

chaokunyang:
> > BaseReq
> 
> I try to make these codes out of io_stryct.py
> 
> ```
> @dataclass
> class ModelWorkerTask:
>     task_func: Callable
>     kwargs: Dict[str, Any]
> 
> 
> @dataclass
> class ModelWorkerTaskOutput:
>     result: Any
> ```
> 
> This works for me.

Thanks for the updates, I will update this PR tomorrow. BTW, we added speculative support recently, I will update this PR to support update weights for `EAGLEWorker` too.

PR #13598, opened 2025-11-19
[router] Use qualified tool names (server_label__tool_name) in MCP inventory

key4ng:
## Motivation
The MCP tool inventory previously used simple tool names as keys, which could cause collisions when multiple MCP servers expose tools with the same name. Additionally, the inventory tracked prompts and resources which are not currently used.

## Modifications
1. Qualified Tool Naming
  - Changed inventory key format from tool_name to server_label__tool_name to prevent naming collisions
  - Updated CachedTool to store both server_label and server_url separately
  - Modified all inventory methods (get_tool, has_tool, insert_tool) to use qualified names

  2. Simplified Inventory
  - Removed unused prompt and resource caching from ToolInventory
  - Removed related methods: get_prompt, list_prompts, get_resource, list_resources, subscribe_resource, unsubscribe_resource
  - Simplified counts() to count() since only tools are tracked

  3. Connection Pool Updates
  - Use server_url as connection pool key instead of server_name for more precise client routing
  - Updated get_client() to accept both server_name and server_url

  4. Updated Call Sites
  - Modified call_tool() to accept qualified tool names and extract original tool name for MCP server calls
  - Updated workflow steps, router handlers, and tests to use new qualified naming format
  - Fixed list_tools() to return (qualified_name, server_label, Tool) tuples

## Checklist

### DIFF
--- a/sgl-router/src/mcp/inventory.rs
+++ b/sgl-router/src/mcp/inventory.rs
@@ -9,21 +9,21 @@ use crate::mcp::config::{Prompt, RawResource, Tool};
 /// Cached tool with metadata
 #[derive(Clone)]
 pub struct CachedTool {
-    pub server_name: String,
+    pub server_label: String,
     pub tool: Tool,
 }
 
 /// Cached prompt with metadata
 #[derive(Clone)]
 pub struct CachedPrompt {
-    pub server_name: String,
+    pub server_label: String,
     pub prompt: Prompt,
 }
 
 /// Cached resource with metadata
 #[derive(Clone)]
 pub struct CachedResource {
-    pub server_name: String,
+    pub server_label: String,
     pub resource: RawResource,
 }
 
@@ -32,7 +32,7 @@ pub struct CachedResource {
 /// Provides thread-safe caching of MCP tools, prompts, and resources.
 /// Entries are refreshed periodically by background tasks.
 pub struct ToolInventory {
-    /// Map of tool_name -> cached tool
+    /// Map of server_label__tool_name -> cached tool
     tools: DashMap<String, CachedTool>,
 
     /// Map of prompt_name -> cached prompt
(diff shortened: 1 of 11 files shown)


sgl-router/src/mcp/inventory.rs
```diff
@@ -65,158 +43,70 @@ impl ToolInventory {
...
     /// Insert or update a tool
-    pub fn insert_tool(&self, tool_name: String, server_name: String, tool: Tool) {
+    ///
+    /// Constructs key as server_label__tool_name
+    pub fn insert_tool(&self, tool_name: String, server_label: String, tool: Tool) {
+        let qualified_name = format!("{}__{}", server_label, tool_name);
         self.tools
-            .insert(tool_name, CachedTool { server_name, tool });
+            .insert(qualified_name, CachedTool { server_label, tool });
     }
```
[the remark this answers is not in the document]
key4ng: let's keep it for now, because when use key, it check the server_label and tool_name in combination

sgl-router/src/mcp/inventory.rs
```diff
@@ -65,122 +44,57 @@ impl ToolInventory {
...
-            prompt_name,
-            CachedPrompt {
-                server_name,
-                prompt,
-            },
-        );
-    }
-
-    /// Get all prompts
-    pub fn list_prompts(&self) -> Vec<(String, String, Prompt)> {
```
CatherineSue, line 128: Can we keep the prompts and resources methods? They may be useful in the future.

sgl-router/src/routers/common.rs
```diff
@@ -0,0 +1,28 @@
+//! Common utilities for routers
+
+/// Strip server_label prefix from qualified MCP tool name
+///
+/// MCP tools are stored internally with qualified names (server_label__tool_name)
+/// to support multiple servers with the same tool names. This function extracts
+/// the original tool name for display purposes.
+pub fn strip_server_label(qualified_name: &str) -> &str {
```
CatherineSue, line 8: Why do we still need this?

sgl-router/src/mcp/inventory.rs
```diff
@@ -4,51 +4,30 @@
...
-/// Cached prompt with metadata
-#[derive(Clone)]
-pub struct CachedPrompt {
-    pub server_name: String,
-    pub prompt: Prompt,
-}
-
-/// Cached resource with metadata
-#[derive(Clone)]
-pub struct CachedResource {
```
slin1237, line 25: we don't use it now because we didn't use them properly
we shouldn't delete prompts and resources

sgl-router/src/mcp/inventory.rs
```diff
@@ -4,51 +4,30 @@
...
-}
-
 /// Tool inventory with periodic refresh
 ///
-/// Provides thread-safe caching of MCP tools, prompts, and resources.
+/// Provides thread-safe caching of MCP tools.
 /// Entries are refreshed periodically by background tasks.
 pub struct ToolInventory {
-    /// Map of tool_name -> cached tool
+    /// Map of server_label__tool_name -> cached tool
```
slin1237, line 22: this our internal mapping
can we not use __
if we wanna address tool names collision with each other
we can add more data structure and mapping?
for example, server name maps to tool names
Tool names map to tools
during look up, since mcp server is required, pass in both mcp server and tool names as two parameters

slin1237 (changes requested):

PR #13600, opened 2025-11-19
enable csgmv automatically on cuda

b8zhong:
It's fine on CUDA platform, so we can auto enable it there, it may also work for ROCm, but it should be tested first.

In Docker image with CUDA 13 + Triton 3.5.1, it also seems to work (and the accuracy is fine):

Screenshot 2025-11-19 at 1 06 26 PM

### DIFF
--- a/python/sglang/srt/utils/common.py
+++ b/python/sglang/srt/utils/common.py
@@ -1434,6 +1434,15 @@ def is_triton_3():
     return triton.__version__.startswith("3.")
 
 
+def is_triton_34() -> bool:
+    """
+    Returns True when running on a CUDA backend with Triton 3.4.x.
+    Currently, this is used to gate CSGMV Triton cache, as
+    we currently use a private API for the cache.
+    """
+    return triton.__version__.startswith("3.4.")
+
+
 def maybe_torch_compile(*args, **kwargs):
     """
     torch.compile does not work for triton 2.2.0, which is needed in xlm1's jax.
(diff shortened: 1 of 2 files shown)


Fridge003 (approved):

python/sglang/srt/utils/common.py
```diff
@@ -3618,6 +3618,12 @@ def my_kernel(x_ptr, y_ptr, BLOCK_SIZE: tl.constexpr):
     """
 
     def decorator(fn):
+        # Auto-enable the custom kernel cache for CUDA, where it is
+        # known to be compatible.
+        if is_cuda() and not envs.SGLANG_USE_CUSTOM_TRITON_KERNEL_CACHE.get():
+            logger.debug("Detected platform CUDA, using custom triton kernel cache.")
+            return CachedKernel(fn, key_fn)
```
lifuhuang, line 3625: I think the expected logic should be:

enable custom cache when either:
1. SGLANG_USE_CUSTOM_TRITON_KERNEL_CACHE not specified **AND** is_cuda == True, 
2. SGLANG_USE_CUSTOM_TRITON_KERNEL_CACHE is true

Can you help double check if your code correctly handles the case when CUDA is true but SGLANG_USE_CUSTOM_TRITON_KERNEL_CACHE  is set to false?

b8zhong: Ah good point, it should be is set instead of get.

Now it is:

```
CUDA True, env var not set → enables custom cache
CUDA True, env var = True → enables custom cache
CUDA True, env var = False → does NOT enable custom cache
CUDA False, env var = True → enables custom cache
CUDA False, env var not set → does NOT enable custom cache
```

Issue #12779, opened 2025-11-06
[Bug] PD prefill node dead cause decode node health check failed after v0.5.2


llc-kc:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Describe the bug

For PD, prefill node dead can cause decode node health check failed and decode node will be killed by k8s, when there are some requests in KV cache transmission.
No problem for v0.5.2, but occurs for 0.5.3 - 0.5.5.post3.

### Reproduction

H20 96G
prefill
```
python3 -m sglang.launch_server \
--model-path /DATA/disk0/models/DeepSeek-V3.1-Terminus \
--served-model-name DeepSeek-V3.1-Terminus \
--disaggregation-mode prefill \
--disaggregation-transfer-backend mooncake \
--disaggregation-ib-device "mlx5_0,mlx5_1" \
--tp 8 \
--mem-fraction-static 0.90 \
--chunked-prefill-size 5120 \
--max-running-requests 4096 \
--context-length 131072 \
--trust-remote-code \
--host 0.0.0.0 \
--port 12121 \
--log-level debug \
--enable-cache-report \
--enable-metrics \
--reasoning-parser deepseek-v3 \
--tool-call-parser deepseekv31 \
--chat-template /sgl-workspace/sglang/examples/chat_template/tool_chat_template_deepseekv31.jinja \
--watchdog-timeout 3600 \
--prefill-round-robin-balance \
--page-size 64 \
--disable-radix-cache \
--decode-log-interval 1
```
decode
```
python3 -m sglang.launch_server \
--model-path /DATA/disk0/models/DeepSeek-V3.1-Terminus \
--served-model-name DeepSeek-V3.1-Terminus \
--disaggregation-mode decode \
--disaggregation-transfer-backend mooncake \
--disaggregation-ib-device "mlx5_0,mlx5_1" \
--cuda-graph-bs 1 4 8 16 32 \
--tp 8 \
--disable-radix-cache \
[...]

PR #13657, opened 2025-11-20
fix nixl prefill crash make decode health check failed

llc-kc:
## Motivation
fix prefill node crash cause decode node health check failed and decode will be killed by k8s
https://github.com/sgl-project/sglang/issues/12779

## Modifications

nixl conn: mark requests from dead prefill as failed and then abort to release kv cache.  Add time out check.
decode: _update_handshake_waiters poll and abort failed.  pop_preallocated optimize health check request kv cache alloc.

## Accuracy Tests

does not impacts model outputs

## Benchmarking and Profiling
does not impacts inference speed


ShangmingCai (commented):
I think the change of `python/sglang/srt/disaggregation/decode.py` is not needed, and I cannot reproduce the failed health check with the mooncake backend.

NIXL backend modification should be checked by @shaharmor98 and @ishandhanani, please help review this PR.

shaharmor98 (approved):
LGTM

ShangmingCai (approved):

PR #13144, opened 2025-11-12
[PD][bug fix] fix memleak when last_batch is none

XucSh:
## Motivation

see [13084](https://github.com/sgl-project/sglang/issues/13084)

## Accuracy Tests

prefill:
python3 -m sglang.launch_server --model-path /root/.cache/modelscope/hub/models/Qwen/Qwen3-8B --pp-size 2  --trust-remote-code --chunked-prefill-size 8192 --disable-overlap-schedule --disaggregation-mode prefill --disaggregation-transfer-backend nixl --host 127.0.0.1 --port 21100 --disable-radix-cache

decode:
ython3 -m sglang.launch_server --model-path /root/.cache/modelscope/hub/models/Qwen/Qwen3-8B --base-gpu-id 4  --trust-remote-code --chunked-prefill-size 8192 --disable-overlap-schedule --disaggregation-mode decode --disaggregation-transfer-backend nixl --host 127.0.0.1 --port 21101 --disable-radix-cache

lb:
python3 -m sglang_router.launch_router --pd-disaggregation --mini-lb --prefill http://127.0.0.1:21100 --decode http://127.0.0.1:21101 --host 127.0.0.1 --port 21000

test:
python3 benchmark/gsm8k/bench_sglang.py --parallel 1400 --num-questions 1400 --port 21000 --data-path /tmp/test.jsonl

after this pr：
Accuracy: 0.908                                                                                
Invalid: 0.000

before this pr:
Accuracy: 0.905                                                                                                                                                                                
Invalid: 0.000

## Benchmarking and Profiling
python3 -m sglang.bench_serving  --dataset-path /root/.cache/modelscope/hub/datasets/gliang1001/ShareGPT_V3_unfiltered_cleaned_split/ShareGPT_V3_unfiltered_cleaned_split.json --num-prompt 1 --random-input 32768 --random-output 1 --request-rate inf --max-concurrency 1 --warmup-requests 0 --backend sglang --dataset-name random --random-range-ratio 1 --port 21000

after this pr:

Backend:                                 sglang                                                
Traffic request rate:                    inf                                                   
Max request concurrency:                 1                                                     
Successful requests:                     1                       
Benchmark duration (s):                  4.01                                                                                                                                                  
Total input tokens:                      32768                                                 
Total input text tokens:                 32768     
Total input vision tokens:               0          
Total generated tokens:                  1          
Total generated tokens (retokenized):    1          
Request throughput (req/s):              0.25       
Input token throughput (tok/s):          8166.41    
[...]

PR #11852, opened 2025-10-20
[PP] Refactor PP to async mode

XucSh:
## Motivation

see [11857](https://github.com/sgl-project/sglang/issues/11857)

User can test with below command now:

python3 -m sglang.launch_server --model /opt/models/Qwen/Qwen3-8b --pp-size 4 --tp 2  --pp-async-batch-depth 1

Co-author: @ShangmingCai @merrymercy @alpha-baby 

Cc: @ShangmingCai @merrymercy @whybeyoung @bluecoffee8 

## Checklist

- [ ] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.ai/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.ai/developer_guide/contribution_guide.html#benchmark-the-speed).

PR #13698, opened 2025-11-21
Fix: PrefillAdder.add_chunked_req with negative rem_total_tokens with pp

strgrb:
## Motivation

I need to run Ring-1T and Ling-1T model with tp8pp4, and met up with an error
image
The error may differ, but new-token is negative every time. It does not appear without pp.
This pr try to fix this problem.

Finally I found `self.rem_total_tokens` become negative in `PrefillAdder.add_chunked_req` ,  it's computed by `available_and_evictable - self.rem_total_token_offset` , and budgets for running requests are added to `self.rem_total_token_offset` .
After merging of last_batch and running_batch, this budget will increase, and if this increase to an amount larger than available size, `PrefillAdder.add_chunked_req` will calculate a negative extend input len.

## Modifications

Since chunked_req's req_pool_idx is freed in `Scheduler.get_next_batch_to_run` , we should check remaining tokens excluding budgets for running requests here, and avoid freeing chunked_req's req_pool_idx. Since budget is not enough for decoding now, it's time for decoding, and chunked_req is hanged, waiting for budget is enough.

## Checklist


ShangmingCai:
Will review this PR today. @strgrb Can you test whether this bug is still happening in https://github.com/sgl-project/sglang/pull/11852?

cc: @XucSh @whybeyoung

ShangmingCai (commented):
Changes don't look like it is related to PP, more likely related to bybrid memory.

CC: @xiezhq-hermann @yizhang2077

ShangmingCai:
> @ShangmingCai Should I move this budget check logic to `PrefillAdder` , which `rem_total_tokens()` can be used directly?

@strgrb If it is related to pp only, maybe you could try reverting this commit locally: https://github.com/sgl-project/sglang/pull/13144. If the bug still exists, then it might not be related to the pp. Also, you can try changing the attention backend to test whether this is a bug of flashinfer?

xiezhq-hermann:
let's get this problem fixed after this [refactoring](https://github.com/sgl-project/sglang/pull/13939) to simplify the logics

PR #13713, opened 2025-11-21
[Fix]: Adjust FutureMap's token_id_bufs Size to Prevent ChunkedPrefill's next_token_ids from Overwriting Previous Prefill Requests' next_token_id

ant-yy:
## Motivation
When overlap_schedule is enabled, the Scheduler can optimize CPU processing and GPU computation overlap through coordinated scheduling, which requires creating a FutureMap to implement asynchronous acquisition of next_token_id. During run_batch execution, after synchronization, we call self.future_map.resolve_future to obtain the next_token_id generated in the previous iteration, while returning future_indices as the index for retrieving next_token_ids in the next iteration.

However, the current implementation of FutureMap's token_ids_buf doesn't account for chunked_prefill scenarios, leading to overwriting of stored next_token_ids. The existing token_ids_buf essentially functions as a circular buffer with size 3*max_running_requests. In chunked_prefill cases, each chunk generates a next_token_id stored in token_ids_buf, meaning a single request during prefill phase may occupy multiple entries. For long-context requests, this number can easily exceed 3, resulting in chunked_prefill's next_token_id overwriting those from other requests.

The diagram example clearly demonstrates this issue: Launch SGLang with max_running_requests=8, then use llm.generate to concurrently process two long-text requests ,as shown in the diagram., Request 2's chunked_prefill overwrites Request 1's final prefill result in next_token_input_id, causing Request 1 to retrieve an incorrect input_token_id during decoding.

future_map_bug_fix_picture

## Tests
I conducted tests using the model Qwen3-30B-A3B-Instruct-2507 on an A100 GPU. The two figures below show the results of sending five concurrent 32k-token requests with max_running_requests set to 8 and 2048 respectively. It can be observed that the input_ids during decoding and the final returned prompts differed for the first three requests. After implementing the modifications from this PR, the results under both max_running_requests=8 and 2048 configurations became consistent with the pre-modification results when max_running_requests was set to 2048.

Figure1(max_running_requests = 8)
8_future_map

Figure2(max_running_requests = 2048 )
2048_future_map

## Checklist


hnyls2002:
@ant-yy You are right. But this problem can have a better solution. Do you want to discuss?

ant-yy:
> You are right. But this problem can have a better solution. Do you want to discuss?
Thank you for the feedback! I'd be happy to discuss potential improvements. Please let me know your availability or preferred method for further discussion.

PR #13714, opened 2025-11-21
Refactor eagle bigram key matching

hnyls2002:
- Handle the bigram conversion in `cache_finished` and `cache_unfinished`, instead of in `insert` and `match_prefix`.
- Introduce `cache_protected_len`: to decouple the `prefix_indices` and the real indices cached in the radix tree.


python/sglang/srt/managers/schedule_batch.py
```diff
@@ -563,7 +563,7 @@ def __init__(
         # The node to lock until for swa radix tree lock ref
         self.swa_uuid_for_lock: Optional[int] = None
         # The prefix length of the last prefix matching
```
ispobock: Do we need to update the comment?

ispobock (approved):
We should also test eagle case in page_size=2 and 64 with small chunked prefill size to make sure there is no memory leak.

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -337,78 +340,60 @@ def insert(self, key: RadixKey, value=None, chunked=False, priority: int = 0):
...
-            page_aligned_len + 1 if self.is_eagle else page_aligned_len
-        )
-
-        old_prefix_len = len(req.prefix_indices)
-        if self.is_eagle and old_prefix_len > req.last_matched_prefix_len:
-            # In EAGLE chunked prefill case, the prefix_indices included one unmatched token (kv_indices[actual_kv_len:])
-            # Here we -1 to make sure the kv of the unmatched token can be freed correctly to avoid memory leak
-            old_prefix_len -= 1
+        # Maybe convert to bigram keys for EAGLE
+        keys = convert_to_bigram_key(req.fill_ids) if self.is_eagle else req.fill_ids
```
zhangrui1203, line 377: Hi, @hnyls2002 
We noticed that `req.fill_ids` here only includes `req.origin_input_ids` and does not include `req.output_ids`, which causes the cached KV cache to be missing the KV entries corresponding to `req.output_ids`. Could you please confirm this behavior?

hanming-lu: Observed the same thing - prefill is cached and decode is not cached. This seems a good time to promote this test: https://github.com/sgl-project/sglang/blob/main/test/registered/4-gpu-models/test_qwen3_next_models.py#L92-L99

This test will catch this behavior.

PR #13126, opened 2025-11-12
[VLM][feat] Support encoder DP for Qwen2.5-VL

liusy58:
## Motivation

Based on this PR https://github.com/vllm-project/vllm/pull/22742, vLLM has introduced support for data parallelism (DP) in the vision transformer (ViT) component while maintaining tensor parallelism (TP) for the language model. I think SGLang should adopt this approach as well. This hybrid parallelization strategy can significantly reduce time-to-first-token (TTFT) when processing requests with multiple images. 

## Benchmarking and Profiling

* To lanunch instance

```sh

export SGLANG_TORCH_PROFILER_DIR=./p

MODEL=/mnt/models/Qwen2.5-VL-72B-Instuct
PORT=30000
TP=4
MEM_FRACTION=0.8
CHUNK_SIZE=32000
export SGLANG_VLM_CACHE_SIZE_MB=0
python -m sglang.launch_server \
    --model-path $MODEL \
    --tp $TP --mm-encoder-tp-mode=data \
    --mem-fraction-static $MEM_FRACTION \
    --disable-radix-cache \
    --chunked-prefill-size $CHUNK_SIZE \
    --port 30000

```

* To bench

```sh
DATASET=image
REQUEST_RATE=0.1
PORT=8192
export HF_ENDPOINT=https://hf-mirror.com  

python -m sglang.bench_serving \
    --dataset-name $DATASET \
    --port 30000 \
    --random-input-len 200 \
    --image-count  4 \
    --image-resolution 1200x1200 \
    --random-output-len 1 \
    --backend vllm-chat \
    --num-prompts 200 \
    --request-rate 0.2 \
    --warmup-requests=0 
```
Here are the prefill performance gains from my benchmarks:

|TP Size | Image Count | Resolution | Base (ms) | Encoder-DP (ms) | Improvement
|-- | -- | -- | -- | -- | --
|2 | 4 | 720p | 2,826.50 | 2,663.48 | -6.0%
|4 | 4 | 720p | 1,916.67 | 1,647.52 | -16.3%
|4 | 4 | 1200×1200 | 3,559.90 | 3,000.95 | -18.6%

I believe the TTFT improvements would be even more significant with higher TP sizes and more images per request in the benchmark.

## Checklist

- [x] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
[...]

PR #13724, opened 2025-11-21
support qwen3_vl vision model dp

Lzhang-hub:
## Motivation

Based on [PR 13126](https://github.com/sgl-project/sglang/pull/13126), add support for the Qwen3_VL vision model DP.
### Qwen/Qwen3-VL-32B-Instruct
```
- without dp
| Tasks  |Version|Filter|n-shot| Metric |   |Value |   |Stderr|
|--------|------:|------|-----:|--------|---|-----:|---|------|
|mmmu_val|      0|none  |     0|mmmu_acc|↑  |0.6067|±  |   N/A|

- with dp
| Tasks  |Version|Filter|n-shot| Metric |   |Value |   |Stderr|
|--------|------:|------|-----:|--------|---|-----:|---|------|
|mmmu_val|      0|none  |     0|mmmu_acc|↑  |0.6078|±  |   N/A|

### Qwen/Qwen3-VL-235B-A22B-Instruct

 - without dp:
| Tasks  |Version|Filter|n-shot| Metric |   |Value |   |Stderr|
|--------|------:|------|-----:|--------|---|-----:|---|------|
|mmmu_val|      0|none  |     0|mmmu_acc|↑  |0.6456|±  |   N/A|

- with dp:
| Tasks  |Version|Filter|n-shot| Metric |   |Value|   |Stderr|
|--------|------:|------|-----:|--------|---|----:|---|------|
|mmmu_val|      0|none  |     0|mmmu_acc|↑  | 0.64|±  |   N/A|
```
## Benchmarking and Profiling

server cmd:
```
python -m sglang.launch_server --host 0.0.0.0 --port 8080 --model-path Qwen/Qwen3-VL-32B-Instruct --served-model-name test --trust-remote-code --disable-radix-cache --tp 4 --mem-fraction-static 0.85  --mm-attention-backend fa3 --attention-backend flashinfer --mm-enable-dp-encoder 
```

bench cmd：
```

python -m sglang.bench_serving \
    --dataset-name image \
    --host 0.0.0.0 \
    --port 8080 \
    --random-input-len 200 \
    --image-count  4 \
    --image-resolution 1280x720 \
    --random-output-len 1 \
    --backend vllm-chat \
    --num-prompts 50 \
    --request-rate 0.4 \
    --warmup-requests=0  \
    --model Qwen/Qwen3-VL-32B-Instruct

```

```
| model | tp | Img count | Img resolution | base(ms) | dp(ms) | improment |
| --- | --- | --- | --- | --- | --- | --- |
| Qwen3-VL-32B-Instruct | 4 | 4 | 720p | 1205.37 | 1131.91 | -6.2% |
| Qwen3-VL-235B-A22B-Instruct | 4 | 4 | 720p | 1112.01 | 1075.99 | -3.4% |
| Qwen3-VL-235B-A22B-Instruct | 4 | 4 | 1280×1280 | 4803.98 | 4522.80 | -6.2% |
```

## Checklist


vincentzed:
Is it tested for MoE?

Lzhang-hub:
> Please add a unit test for this model dp.

Done

test/nightly/test_encoder_dp.py
```diff
@@ -19,6 +19,8 @@
 
 MODELS = [
     SimpleNamespace(model="Qwen/Qwen2.5-VL-72B-Instruct", mmmu_accuracy=0.55),
+    SimpleNamespace(model="Qwen/Qwen3-VL-32B-Instruct", mmmu_accuracy=0.55),
+    SimpleNamespace(model="Qwen/Qwen3-VL-235B-A22B-Instruct", mmmu_accuracy=0.55),
```
ShangmingCai: These will definitely increase the estimated run time of this test. I think "Qwen/Qwen3-VL-235B-A22B-Instruct" is not needed. Also, this is a nightly test, so it is not included in the current per-commit CI. Maybe you should run locally and modify the estimated time in run_suite.py

python/sglang/srt/models/qwen2_5_vl.py
```diff
@@ -269,7 +269,7 @@ def __init__(
         self.patch_size = vision_config.patch_size
         mlp_hidden_size: int = ((vision_config.intermediate_size + 7) // 8) * 8
         self.use_data_parallel = use_data_parallel
-        self.out_hidden_size = vision_config.out_hidden_size
+        self.out_dim = vision_config.out_hidden_size
```
ShangmingCai: Is this line of renaming a compulsory change?

Lzhang-hub: out_dim refers to the final output shape of the vision_model, mainly used during all_gather when ranks without images need to create an empty tensor.

The main reason for the change is that qwen3-vl’s output shape is `vision_config.out_hidden_size * (1 + len(self.deepstack_visual_indexes))`, while qwen2.5-vl’s output shape is `out_hidden_size`, which is not the same concept as vision_config.out_hidden_size

test/nightly/test_encoder_dp.py
Lzhang-hub: OK, I’m not very familiar with CI testing. I just followed what was added in qwen2.5-vl encoder DP. I try to explore and make improvements.

yuan-luo:
> if image-count=1, will dp still take effect?

image-count controls how many images are attached to a single request. It affects the number of visual tokens in one sample, not the batch size.
DP (use_data_parallel) controls how the batch is split across multiple GPUs. To fully utilize DP, the batch size, or the number of concurrent requests merged by the scheduler should be greater than or equal to the dp size.

So even if image-count = 1, as long as you have many concurrent requests that the scheduler merges into a large hidden_states batch, let's say bs=32, the vision part can still distribute these 32 samples across multiple GPUs, and DP will be effective.

But if you only have 1 image in total bs=1, then even with use_data_parallel = True, you won’t see any benefit from DP, because there’s effectively only work for one rank.

yuan-luo:
@Lzhang-hub could you tag the label precisely?
image

python/sglang/srt/multimodal/mm_utils.py
```diff
@@ -580,7 +580,7 @@ def run_dp_sharded_mrope_vision_model(
         else:
             # Handle empty case
             image_embeds_local = torch.empty(
-                (0, vision_model.out_hidden_size),
```
yuan-luo, line 583: This is a bug. Not only qwen2_5_vl and qwen3_vl will use this `run_dp_sharded_mrope_vision_model`, more and more vlm models with mrope might adopt this function to support DP, but they are not necessarily owning vision_model.out_dim. All the models are to be impacted.
Change back to vision_model.out_hidden_size here. Meanwhile, change back to out_hidden_size in qwen3_vl.py.

Lzhang-hub: OK

yuan-luo (approved):
LGTM.

PR #13729, opened 2025-11-21
Fix model weights validation with automatic cache cleanup

alisonshao:
## Summary

Fixes model weights validation by adding automatic cleanup of corrupted/incomplete cache files and centralizing all validation logic in `weight_utils.py`.

## Problem

PR #13660 attempted to centralize validation but encountered issue in its cleanup mechanism. Corrupted files were detected but not removed, causing re-downloads to reuse bad blobs.

Key observation is, deleting individual files didn't work due to HuggingFace's symlink-based blob storage.

### Added validation with cleanup:
- `_validate_safetensors_file()` - Detects corrupted safetensors files
- `_validate_sharded_model()` - Validates shard completeness and corruption  
- `_cleanup_corrupted_files_selective()` - Removes only corrupted files and their blobs (as opposed to removing the entire directory)

  - Selective: When specific corrupted files are identified, delete only those files
  - Saves significant bandwidth and time by only re-downloading corrupted files instead of entire models.
- `_cleanup_corrupted_model_cache()` - Removes entire cache when selective cleanup isn't possible
  - Full: When corruption scope is unknown, delete entire cache

### Removed redundant CI validation:
- Deleted `scripts/ci/validate_and_download_models.py` 
- Updated `scripts/ci/prepare_runner.sh`

Validation now happens automatically during model loading.

## Related

Completes #13660


alisonshao:
https://github.com/sgl-project/sglang/actions/runs/19583606936/job/56087714030?pr=13729#:~:text=%5B2025%2D11%2D21%2021%3A22%3A06%5D%20Corrupted,parsing%20a%20value%20at%20line%201%20column%200

python/sglang/srt/model_loader/weight_utils.py
```diff
@@ -257,6 +258,217 @@ def get_quant_config(
         return quant_cls.from_config(config)
 
 
+def _validate_safetensors_file(file_path: str) -> bool:
```
hnyls2002: One suggestion: put all utility functions involving model weights validation checking into a separate file, e.g., `ci_weights_validation.py`

hnyls2002 (changes requested):

scripts/ci/prepare_runner.sh
```diff
@@ -11,19 +11,4 @@ echo ""
 python3 "${SCRIPT_DIR}/cleanup_hf_cache.py"
```
merrymercy, line 11: Why do we still need this? Can we remove the whole script?

alisonshao: I think its safer to keep this at CI start. cleanup_hf_cache.py cleans stale artifacts across all models in the cache at CI start. weight_utils.py only validates the specific model being loaded. cleanup_hf_cache.py removes .tmp and .lock files that weight_utils.py doesn't handle.

PR #13738, opened 2025-11-21
fix trtllm mla spec

b8zhong:
In theory 

`test/srt/test_deepseek_v3_fp4_4gpu.py` should pass


Qiaolin-Yu (changes requested):
~Why delete this? This cache is important and was added a very long time ago. Or perhaps the previous usage of this cache here was incorrect? Could you please explain more? 👀~ I just took a close look at the code, and I think you’re right.

Qiaolin-Yu (approved):

hnyls2002:
@Qiaolin-Yu @b8zhong Please track the CI status. After the `test/srt/test_deepseek_v3_fp4_4gpu.py` passes, add a `read-to-merge` label

PR #9467, opened 2025-08-21
Revert "bugfix: Fix output_ids extraction in detokenizer_manager"

JustinTong0323:
Reverts sgl-project/sglang#9047 to fix #9390

But still need future work to solve the root cause

PR #8837, opened 2025-08-06
Support v1/responses and use harmony in serving_chat

CatherineSue:
## Motivation

Support v1/responses API and mcp server.

Use harmony in serving_chat.py

Co-Authored-By: Xinyuan Tong <justinning0323@outlook.com>

In order to utilize the gpt - oss Demo tools, it is necessary to employ an environment with Python version 3.12. Additionally, the `mcp` and `gpt-oss` packages must be installed.

Quick demo: [link](https://gist.github.com/JustinTong0323/f01059d3951c248a0300b09f2930ed33)

One of the prominent features of gpt-oss is the capability to directly invoke tools, which are referred to as "built-in tools". In sglang, several options are provided:

By default, integration is established with the browser of the reference library (utilizing ExaBackend) and the demo Python interpreter through a Docker container. To utilize the search backend, access to [exa.ai](http://exa.ai/) is required, and the `EXA_API_KEY` should be set as an environment variable. Regarding Python, either ensure that Docker is available, or set `PYTHON_EXECUTION_BACKEND = UV`. However, setting `PYTHON_EXECUTION_BACKEND = UV` allows the execution of model - generated code snippets on the same machine, which poses a certain risk.

The command to launch the server is: `python -m sglang.launch_server ... --tool-server demo`

It should be noted that the default options are solely intended for demonstration purposes. For production-level usage, sglang can function as an MCP client for multiple services. An [example tool server](https://github.com/openai/gpt-oss/tree/main/gpt-oss-mcp-server) that sglang can interact with is provided. These servers encapsulate the demo tools, and the commands to run them are as follows:
```
mcp run -t sse browser_server.py:mcp
mcp run -t sse python_server.py:mcp

python -m sglang.launch_server ... --tool-server ip-1:port-1,ip-2:port-2
```
The URLs are expected to be MCP SSE servers that implement the instructions in the server information and well - documented tools. These tools will be incorporated into the system prompt for the model to enable their utilization.

## Modifications

- Turn on harmony with gpt-oss by default
- Support harmony request and parsing.
- NOTE: tool call in chat completion for this model with harmony is not supported yet.

## Accuracy Test

- Non stream:
Screenshot 2025-08-05 at 4 48 43 PM

- Stream:
Screenshot 2025-08-05 at 5 46 46 PM

- Reasoning effort: low vs high
Screenshot 2025-08-05 at 6 06 55 PM

## Checklist

[...]

PR #13759, opened 2025-11-22
Fix condition for streaming output_ids in tokenizer manager

merrymercy:
There is a bug in how we handle the `output_ids` for streaming requests. (python/sglang/srt/managers/tokenizer_manager.py)

In the default case, for each chunk, the `output_ids` field should contain all output token ids, regardless of streaming or not
Only when users specify `--stream-output`, for each chunk, the `output_ids` field will contain delta output tokens ids.

This logic is correctly handled in the branch of `elif isinstance(recv_obj, BatchTokenIDOutput):`

However, in the branch `isinstance(recv_obj, BatchStrOutput):`, the code was wrong in the very beginning when it was introduced (https://github.com/sgl-project/sglang/pull/8837/files#r2552360555).
It later leads to some very hacky PRs like https://github.com/sgl-project/sglang/pull/9467. (not sure if this is the root cause).

This might be a breaking change but it should be a correct fix. I need more help from @JustinTong0323 @CatherineSue to see whether it fixes or breaks any gpt oss things. 

### Other minor fix
- fix the print in python/sglang/bench_one_batch_server.py
- cache the check of `is_gpt_oss`

Thanks @JeffryLee for finding this bug.


hnyls2002 (approved):

CatherineSue (approved):
Tested with `python3 -m sglang.launch_server --model /models/openai/gpt-oss-120b --port=8034 --tp-size=4 --reasoning-parser=gpt-oss --stream-output`

Hardware: H100

It works with the changes in b793a83: 1) hacky changes removed. 2) Update `ResponseOutputText` to use `text` instead of `content` field as it has been removed.

streaming:
Image

**Note:** streaming won't work without `--stream-output`

non-streaming:
Screenshot 2025-11-24 at 5 39 10 PM

openai 2.6.1 API:
Image

PR #13781, opened 2025-11-23
Fix pagination bug in CI monitor preventing performance-test-2-gpu data collection

alisonshao:
## Problem
CI monitor showed 0 records for performance-test-2-gpu tests after Nov 11, even though tests were running successfully.
https://github.com/sgl-project/sglang/actions/runs/19587274768

## Root Cause
The `get_job_logs()` function wasn't using pagination when fetching jobs. GitHub API defaults to 30 jobs per page, but PR test workflows have 46 jobs. Since performance-test-2-gpu appears after the 30th job, it was being missed.

## Fix
Added `params={"per_page": 100}` to the jobs API request to retrieve all jobs.

## Verification
Tested on runs with successful 2-GPU data - now collects all 6 performance-test-2-gpu tests correctly.


BBuf (approved):
LGTM.

PR #13793, opened 2025-11-23
[CI tiny fix] Enhance robustness of vision chunked prefill test with ROUGE-L metric

BBuf:
## Motivation

Exact string comparison in the vision chunked prefill test is too strict. Minor textual variations cause failures even when outputs are semantically equivalent.

https://github.com/sgl-project/sglang/actions/runs/19540556341/job/56002840128?pr=13587

图片

In the outputs above, the semantics are highly similar regardless of whether chunked prefill is enabled. However, due to the non-deterministic nature of inference, slight variations in model output logprobs can lead to non-identical decoded text. By using the existing ROUGE-L evaluation metric in SGLang, we enhance the robustness of this test and prevent frequent failures in CI.

## Checklist


yuan-luo (approved):

BBuf:
> I'm curious that if the lmms_evals of the productive code has no drop, the minor textual variations shall not exist. In this ground, I'm fine with this test change, with some concern about why lmms_evals has no drop or we don't have a lmms_evals test for the vision chunked prefill.

We may need to discuss this further with the developer of this feature, but for now, let's prioritize avoiding CI errors.

PR #13826, opened 2025-11-24
[Bug Fix] Decode OOM with EAGLE speculative decoding

adityakamat24:
## Motivation

Fix decode OOM in EAGLE speculative decoding when using paged KV cache allocation (page_size > 1).
Fixes #13741

The issue: kv_allocated_len wasn't getting updated in paged allocation paths. When release_kv_cache() runs, it uses kv_allocated_len to figure out what to free. Since we never updated it for paged allocations, no memory got released - it just kept growing until OOM.
The fix: Added kv_allocated_len tracking to the paged allocation paths to match what the page_size == 1 branch does, so memory gets freed properly when requests finish.

## Modifications

python/sglang/srt/speculative/eagle_info.py

- prepare_for_verify(): Update kv_allocated_len after alloc_paged_token_slots_extend()
- verify(): Update kv_committed_len and kv_allocated_len for paged path (page_size > 1 and topk > 1)

python/sglang/srt/speculative/eagle_worker.py

- _draft_preprocess_decode(): Update kv_allocated_len after alloc_paged_token_slots_extend()

## Checklist

### DIFF
--- a/python/sglang/srt/speculative/eagle_info.py
+++ b/python/sglang/srt/speculative/eagle_info.py
@@ -137,6 +137,8 @@ class EagleVerifyInput(SpecInput, EagleVerifyInputV2Mixin):
                 last_loc,
                 len(batch.input_ids),
             )
+            for req in batch.reqs:
+                req.kv_allocated_len += self.draft_token_num
             self.last_loc = last_loc
 
         bs = batch.batch_size()
@@ -478,6 +480,9 @@ class EagleVerifyInput(SpecInput, EagleVerifyInputV2Mixin):
                 batch.token_to_kv_pool_allocator.get_kvcache().move_kv_cache(
                     tgt_cache_loc, src_cache_loc
                 )
+                for i, req in enumerate(batch.reqs):
+                    req.kv_committed_len += accept_length_list[i] + 1
+                    req.kv_allocated_len = req.kv_committed_len
 
         # Construct EagleVerifyOutput
         if not has_finished:
--- a/python/sglang/srt/speculative/eagle_worker.py
+++ b/python/sglang/srt/speculative/eagle_worker.py
@@ -449,8 +449,12 @@ class EAGLEWorker(TpModelWorker):
                     backup_state=True,
                 )
             )
-            for req in batch.reqs:
-                req.kv_allocated_len += self.speculative_num_steps * self.topk
+            if self.topk > 1:
+                for i, req in enumerate(batch.reqs):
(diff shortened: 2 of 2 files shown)


adityakamat24:
> @adityakamat24 `kv_committed_len` and `kv_allocated_len` is not compatible with `topk > 1`

I realized that and closed the PR. Working on a proper fix.

hnyls2002:
@adityakamat24 Sure, I guess the problem is not related to `kv_committed_len` and `kv_allocated_len`

hnyls2002:
@adityakamat24 Please show some debugging info

adityakamat24:
@hnyls2002 Sorry for the delay.
Alright so I think I finally tracked this down. Added a bunch of debug logging to `eagle_info.py` and `common.py` to trace what's happening with the memory.

The leak is in the verify phase when `topk > 1`. Here's what I'm seeing:

When EAGLE verifies tokens, it computes `keep_len = ceil_align(committed_len, page_size)` to avoid fragmenting pages. So if you committed 5 tokens but page_size is 4, it keeps 8 slots (rounds up to next page boundary). Those extra 3 slots between committed_len and keep_len are the "page tail."

The problem is when we do `batch.out_cache_loc = tgt_cache_loc` - this only includes the committed tokens. So those page tail slots just... disappear. They're not in the new `out_cache_loc`, they're not in `to_free_slots` (because we kept them to avoid fragmentation), and they can't be recovered later.

Worse, `kv_committed_len` and `kv_allocated_len` never get updated when `topk > 1`. So when `release_kv_cache()` eventually runs, it does `start_p = ceil_align(kv_committed_len, page_size)` and `end_p = kv_allocated_len`, sees that `start_p >= end_p`, and doesn't free anything.

From my debug logs, with `topk=8`, `page_size=4`, and the small KV cache (4500 slots), I'm seeing:
- 940 unique slots leaked before OOM (that's like 235 pages worth)
- Most of them (708) get lost during verify when batch.out_cache_loc gets replaced
- Some (16) during release because of the ceil_align issue
- Test completed 193 requests before dying, so roughly 4-5 slots leaked per request

Test dies after about 80 seconds when it needs 512 slots but only 508 are available (after leaking ~21% of the pool).

Not totally sure what the best fix is. Happy to work on a fix, if you point me in the right direction.

Anyway, I have the full debug log if you want to see it. Let me know if this makes sense or if I'm missing something.

[debug_output.log](https://github.com/user-attachments/files/24009988/debug_output.log)

PR #11865, opened 2025-10-20
Fix model loading from S3 buckets

ganochenkodg:
## Motivation

The goal of this PR is to fix loading models from S3 buckets. The code contains a bug in iterator function: https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/connector/s3.py#L104

file.removeprefix lefts / in the beginning of filename, and it causes issues during os.path.join execution. Python documentation about path.join says: "If any component is an absolute path, all previous components (on Windows, including the previous drive letter, if there was one) are thrown away, and joining continues." With such bug all files from S3 are saved in the root folder as a result, and Sglang can't find them during model loading:

ValueError: Unrecognized model in /tmp/tmpgtn2cumr. Should have a model_type key in its config.json, or contain one of the following strings in its name: list of model types

## Modifications

This PR fixes path.join function and also adds S3 related python dependencies into docker image, because they are non present in current images by some reason. 

> destination_file = os.path.join(local_dir, file.removeprefix(base_dir))
> print(destination_file)
/config.json

> destination_file = os.path.join(local_dir, file.removeprefix(f"{base_dir}/"))
> print(destination_file)
/tmp/somefolder/config.json

PR #13836, opened 2025-11-24
further optimze model load

zyksir:
## Motivation

Optimize Load time, Related to issue #12989

This PR use `runai_model_streamer` to decrease the CPU overhead of `safe_open` from `safetensors`, which lead to better optimization for load time.

This is the load time with/without optimization. The CPU offload is very heavy since we need lots of time to create pin memory. We should avoid this as much as possible.
```python
Without Optimization:

sglang generate --model-path=Qwen/Qwen-Image     --prompt='A curious raccoon'     --width=720 --height=720 --save-output
# 64.6 seconds

sglang generate --model-path=Qwen/Qwen-Image     --prompt='A curious raccoon'     --width=720 --height=720 --save-output --dit-cpu-offload false --text-encoder-cpu-offload false --image-encoder-cpu-offload false --vae-cpu-offload false
# 27 seconds

## With Optimization:
sglang generate --model-path=Qwen/Qwen-Image     --prompt='A curious raccoon'     --width=720 --height=720 --save-output
# 49.53 seconds

sglang generate --model-path=Qwen/Qwen-Image     --prompt='A curious raccoon'     --width=720 --height=720 --save-output --dit-cpu-offload false --text-encoder-cpu-offload false --image-encoder-cpu-offload false --vae-cpu-offload false
# 14.7 seconds
```

I will use the load for text encoder as an example. Before, The CPU function takes lots of time and there are bubbles for H2D copy; while after optimizaiton, the H2D copy is dense.
Before Optimization
image
image

After Optimization
image
image

## Accuracy Tests

I have checked the result of the generated images. It looks good.

## Checklist


python/sglang/multimodal_gen/runtime/loader/weight_utils.py
```diff
@@ -118,22 +125,38 @@ def filter_files_not_needed_for_inference(hf_weights_files: list[str]) -> list[s
...
+    if use_runai_model_streamer:
+        name2weights = {}
+        with SafetensorsStreamer() as streamer:
+            streamer.stream_files(hf_weights_files)
+            for name, tensor in streamer.get_tensors():
+                if not to_cpu:
+                    tensor = tensor.to(str(get_local_torch_device()), non_blocking=True)
+                else:
+                    tensor = tensor.clone()
+                name2weights[name] = tensor
```
noa-neria: Reading all the weights before yielding is less optimized, as the streamer is still reading in the background during the yields. This will also save space on the device, in case the model is being sharded.
Instead of storing all the models in a dict it is better to yield ready tensors, so you can process the tensor while the streamer continues to read.

Regarding the overhead of allocating pinned memory:
I suggest to try utilize the fact that the streamer is reading ahead in the background once `stream_files` is called, to read from storage and allocate pinned memory in parallel.
By calling `safetensors_weight_iterator` before `shard_model` [here](https://github.com/sgl-project/sglang/blob/aeac6220586d08519fe0e832fec5a950292bc9a7/python/sglang/multimodal_gen/runtime/loader/fsdp_load.py#L134), the reading will happen during the creation of the pinned memory. This may further reduce the load time.
This will only work if the iterator will not read all the weights before yielding, as it is doing now.

python/sglang/multimodal_gen/runtime/loader/weight_utils.py
```diff
@@ -118,22 +125,38 @@ def filter_files_not_needed_for_inference(hf_weights_files: list[str]) -> list[s
...
-                yield name, param
+    if use_runai_model_streamer:
+        name2weights = {}
+        with SafetensorsStreamer() as streamer:
+            streamer.stream_files(hf_weights_files)
+            for name, tensor in streamer.get_tensors():
+                if not to_cpu:
+                    tensor = tensor.to(str(get_local_torch_device()), non_blocking=True)
+                else:
+                    tensor = tensor.clone()
```
noa-neria: For creating a copy of the tensor do `tensor = tensor.clone().detach()`
The model streamer is allocating a single memory buffer to contain all the weights. Yielded tensors are memory views of the corresponding ranges in the buffer. Detaching will release the underlying buffer, which is not needed since you deep copy the tensors.

python/sglang/multimodal_gen/runtime/loader/weight_utils.py
```diff
@@ -118,22 +125,38 @@ def filter_files_not_needed_for_inference(hf_weights_files: list[str]) -> list[s
...
+                name2weights[name] = tensor
+
+        torch.cuda.synchronize()
```
noa-neria: Why is this needed?

noa-neria (commented):
Hi from the Nvidia Run:ai team,

Thanks for utilizing our project!

Our comments also suggest an idea for further optimization of the pinned memory overhead, if you choose to try it.

The model streamer also supports direct reading from object storage, as demonstrated in #11865 and can be further optimized with [distributed streaming](https://github.com/run-ai/runai-model-streamer/blob/master/docs/src/usage.md#distributed-streaming)

python/sglang/multimodal_gen/runtime/loader/weight_utils.py
zyksir: Actually the behavior of creating pin memory is happening in FSDP and I cannot find a good way to control its behavior

python/sglang/multimodal_gen/runtime/loader/weight_utils.py
zyksir: Since I am using non blocking = True, this synchronization make sure that all the H2D copy is finished

python/sglang/multimodal_gen/runtime/loader/weight_utils.py
noa-neria: @zyksir absolutely - pinned memory is allocated lazily by FSDP
However, maybe you can trigger it by moving the sharded tensor to pinned memory inside `load_model_from_full_model_state_dict` :

```
if cpu_offload:
    sharded_tensor = sharded_tensor.to("cpu", pin_memory=True)
```

In this case loading the full model to cpu (not gpu) is more efficient, followed by sharding and moving to pinned memory. For this scheme the safetensors_weight_iterator should yield tensors as they come, instead of reading all the weights before the yield

zyksir: I cannot find any document about the parameter `is_distributed`, what is this?

zyksir:
@noa-neria Hi, 
1. I have updated the usage of stream_files to `streamer.stream_files(hf_weights_files, device=device, is_distributed=True)`
2. `yield name, tensor.clone().detach()` is used.

My current concern is that I cannot find the API for `stream_files`, which means I don't know the meaning of each param and I have to guess according to the examples. will `is_distributed` do load tensor to GPU on different GPUs and do broadcast?

python/sglang/multimodal_gen/runtime/loader/weight_utils.py
noa-neria: @zyksir  This parameter is for distributed streaming to CUDA devices as explained [here](https://github.com/run-ai/runai-model-streamer/blob/master/docs/src/usage.md#distributed-streaming). When loading a model with TP > 1 from object storage, distributed streaming divides the reading between the processes, which improves performance by a factor of ~TP.

You can use the default parameters, which is to stream to CPU without distributed streaming (`device = "cpu"`, `is_distributed = False`)

noa-neria:
@zyksir the `device` and `is_distributed` are for [distributed streaming](https://github.com/run-ai/runai-model-streamer/blob/master/docs/src/usage.md#distributed-streaming) 

The distributed streaming is intended for loading from object storage, which is not the case here. In your current implementation, passing `device=device` will only copy the full tensor to the GPU without distributed streaming, and that will hurt performance.

I suggest to use `device="cpu", is_distributed=False` which are already the default behavior.

Regarding the new pipeline:

1. Streamer yields CPU tensor (and continues to read in the background)
2. Tensor is sharded
3. If offloading to CPU, shard is copied to pinned memory
4. Shard is copied to GPU

This will allocate pinned memory in the total size of the model in advance. Is this OK?
Did the new pipeline improve performance for `cpu_offload=True` ?

python/sglang/multimodal_gen/runtime/loader/weight_utils.py
```diff
@@ -185,16 +193,22 @@ def safetensors_weights_iterator(
...
-        disable=not enable_tqdm,
-        bar_format=_BAR_FORMAT,
-    ):
-        with safe_open(st_file, framework="pt", device=device) as f:
-            for name in f.keys():  # noqa: SIM118
-                param = f.get_tensor(name)
-                yield name, param
+    if use_runai_model_streamer:
+        with SafetensorsStreamer() as streamer:
+            streamer.stream_files(hf_weights_files, device=device, is_distributed=True)
```
noa-neria: `is_distributed` should be set to `False`, unless loading from object storage which is not the case here.
`device`  should be set to  `"cpu"`, otherwise the streamer will yield the full tensors on the GPU.

Since these are the default parameters anyway, you can call `streamer.stream_files(hf_weights_files)`

zyksir:
And pin_memory ahead does help! but not too much, from 52 seconds -> 49 seconds. The allocation of pin_memory is still very large and is the bottleneck. For cpu offload, I will stop here since to get the best optimization we should not use cpu offload.

python/sglang/multimodal_gen/runtime/loader/weight_utils.py
```diff
@@ -185,16 +193,25 @@ def safetensors_weights_iterator(
...
-        with safe_open(st_file, framework="pt", device=device) as f:
-            for name in f.keys():  # noqa: SIM118
-                param = f.get_tensor(name)
-                yield name, param
+    if use_runai_model_streamer:
+        with SafetensorsStreamer() as streamer:
+            streamer.stream_files(hf_weights_files)
+            for name, tensor in streamer.get_tensors():
+                if to_cpu:
+                    yield name, tensor.clone().detach().cpu()
```
noa-neria: The tensor is already in the cpu

noa-neria:
> And pin_memory ahead does help! but not too much, from 52 seconds -> 49 seconds. The allocation of pin_memory is still very large and is the bottleneck. For cpu offload, I will stop here since to get the best optimization we should not use cpu offload.

@zyksir Absolutely, I understand that cpu_offload is not optimal to begin with

This is why you did not see performance gain yet:

Note that you don't actually overlap reading from storage and allocating pinned memory as intended, because of how `load_model_from_full_model_state_dict` is implemented 
1.  Reading from storage is over when `hf_to_custom_state_dict` returns [here](https://github.com/sgl-project/sglang/blob/f540f07d3ac9fa9833f62b983050b173954a9512/python/sglang/multimodal_gen/runtime/loader/fsdp_load.py#L251C52-L251C75) , because this is where `weights_iterator` was used (i.e. where the streamer is used)
2. Next, pinned memory is allocated in the second iteration [here](https://github.com/sgl-project/sglang/blob/f540f07d3ac9fa9833f62b983050b173954a9512/python/sglang/multimodal_gen/runtime/loader/fsdp_load.py#L254). This loop is taking the tensors from a dict.

You can see that the two steps are not overlapping, and this is the reason for no performance gain.

noa-neria (approved):

zyksir:
I understand, since make pin memory creation and D2H copy overlap will lead to a large refactor to the code, and might introduce more bugs. that's why I say I will stop here at this time.

python/sglang/multimodal_gen/runtime/loader/weight_utils.py
zyksir: Thank you, I have changed it to `tensor.clone().detach()`

PR #13854, opened 2025-11-24
Support sanity checking weight consistency especially for RL

fzyzcjy:
## Checklist


python/sglang/srt/entrypoints/http_server.py
```diff
@@ -954,6 +955,15 @@ async def resume_memory_occupation(
         return _create_error_response(e)
 
 
+@app.post("/check_weights")
```
slin1237: do we have to add an endpoint for this?
can this be part of model info endpoint? 
such that we don't add a new resource/endpoint, likewise the verb in front of endpoint
Existing SGLang frontend has too many of those endpoints

slin1237 (changes requested):
ref to comments

python/sglang/srt/entrypoints/http_server.py
fzyzcjy: it needs to mutate state for the checker, i.e. the json payload can be snapshot,reset_params,check. while I guess model info endpoint is purely read w/o any mutations.

slin1237 (approved):

PR #13861, opened 2025-11-24
dockerfile: add lightweight runtime stage and refactors 

ishandhanani:
Will merge this in after next release

Multi-stage Dockerfile splits SGLang builds into base, framework, and runtime stages. Runtime cuts image size roughly in half.

---

```bash
sglang                            framework-test     be66a8e51a09   39.3GB
sglang                            runtime-test          a4dac91fe030    20GB
```

---

Tests
1. cu13 arm - https://github.com/ishandhanani/srt-slurm/blob/main/recipies/gb300-fp4/1p2.yaml ✅
2. cu12 arm - https://github.com/ishandhanani/srt-slurm/blob/main/recipies/gb200-fp4/1p2d.yaml ✅
3. x86 - [all simple agg/disagg examples](https://github.com/ai-dynamo/dynamo/tree/cf457179bc50ad24783f9d29ae50636d382400f8/components/backends/sglang/launch) ✅


slin1237 (approved):

docker/Dockerfile
```diff
@@ -251,66 +373,169 @@ RUN --mount=type=cache,target=/root/.cache/pip python3 -m pip install --break-sy
...
+# Copy SGLang workspace
+COPY --from=framework /sgl-workspace /sgl-workspace
+
+# Copy GDRCopy runtime libraries (but not the build artifacts)
+COPY --from=framework /usr/lib/libgdrapi.so* /usr/lib/
+COPY --from=framework /usr/bin/gdrcopy_* /usr/bin/
+COPY --from=framework /usr/src/gdrdrv-2.5.1 /usr/src/gdrdrv-2.5.1
+
+# Copy essential binaries
+COPY --from=framework /usr/local/bin/diff-so-fancy /usr/local/bin/
```
nv-tusharma: I don't think we need to copy these packages into the container, these seem required for dev.

docker/Dockerfile
```diff
@@ -251,66 +373,169 @@ RUN --mount=type=cache,target=/root/.cache/pip python3 -m pip install --break-sy
...
+# - Essential runtime dependencies only (no build tools)
+# - Optimized for inference workloads and deployment
+# - Significantly smaller than the framework image
+#
+# Use this stage when you need:
+# - Production deployment of SGLang
+# - Minimal runtime footprint without build tools
+# - Ready-to-run inference server environment
+#
+FROM nvidia/cuda:${CUDA_VERSION}-cudnn-runtime-ubuntu22.04 AS runtime
```
nv-tusharma: I guess we can't move this to U24 yet since the built artifacts are on U22. I guess this is fine for now, but from Dynamo we'll need to revisit this when we're ready to leverage this since Dynamo is on U24.

docker/Dockerfile
```diff
@@ -198,48 +314,54 @@ RUN --mount=type=cache,target=/root/.cache/pip python3 -m pip install \
...
+    && rm -rf /var/lib/apt/lists/*
 
-# Set up locale
-RUN locale-gen en_US.UTF-8
-ENV LANG=en_US.UTF-8
-ENV LANGUAGE=en_US:en
-ENV LC_ALL=en_US.UTF-8
-
-# Install minimal Python packages
-RUN --mount=type=cache,target=/root/.cache/pip python3 -m pip install --break-system-packages \
```
nv-tusharma, line 258: We probably don't need these packages in the runtime image but since they are inherited due to L520. Fine for now, since these packages probably don't add much in size.

docker/Dockerfile
```diff
@@ -251,66 +373,169 @@ RUN --mount=type=cache,target=/root/.cache/pip python3 -m pip install --break-sy
...
+
+ENV LANG=en_US.UTF-8 \
+    LANGUAGE=en_US:en \
+    LC_ALL=en_US.UTF-8
+
+# Copy Python site-packages from framework (contains all built packages)
+COPY --from=framework /usr/local/lib/python3.12/dist-packages /usr/local/lib/python3.12/dist-packages
+
+# Copy SGLang workspace
+COPY --from=framework /sgl-workspace /sgl-workspace
```
nv-tusharma, line 527: Does this copy nvshmem and DeepEP binaries? Is it possible if we could just copy those artifacts out instead of the full workspace?

docker/Dockerfile
```diff
@@ -251,66 +373,169 @@ RUN --mount=type=cache,target=/root/.cache/pip python3 -m pip install --break-sy
...
+COPY --from=framework /usr/local/bin/yank /usr/local/bin/
+COPY --from=framework /usr/local/bin/just /usr/local/bin/
+
+# Fix DeepEP IBGDA symlink in runtime
+RUN ln -sf /usr/lib/$(uname -m)-linux-gnu/libmlx5.so.1 /usr/lib/$(uname -m)-linux-gnu/libmlx5.so
+
+WORKDIR /sgl-workspace/sglang
+
+# Default command
+CMD ["/bin/bash"]
```
nv-tusharma: Change to ENTRYPOINT ["/opt/nvidia/nvidia_entrypoint.sh"]?

docker/Dockerfile
```diff
@@ -251,66 +373,169 @@ RUN --mount=type=cache,target=/root/.cache/pip python3 -m pip install --break-sy
...
+# - Ready-to-run inference server environment
+#
+FROM nvidia/cuda:${CUDA_VERSION}-cudnn-runtime-ubuntu22.04 AS runtime
+
+ARG CUDA_VERSION
+ARG TARGETARCH
+ARG GDRCOPY_VERSION=2.5.1
+
+ENV DEBIAN_FRONTEND=noninteractive \
+    CUDA_HOME=/usr/local/cuda \
```
nv-tusharma, line 450: The cudnn runtime library might not have all the cuda libraries you need for JIT compilation, You might want to check if JIT compilation works and if it doesn't add:
````
COPY --from=framework /usr/local/cuda/bin/nvcc /usr/local/cuda/bin/nvcc
COPY --from=framework /usr/local/cuda/bin/cudafe++ /usr/local/cuda/bin/cudafe++
COPY --from=framework /usr/local/cuda/bin/ptxas /usr/local/cuda/bin/ptxas
COPY --from=framework /usr/local/cuda/bin/fatbinary /usr/local/cuda/bin/fatbinary
COPY --from=framework /usr/local/cuda/include/ /usr/local/cuda/include/
COPY --from=framework /usr/local/cuda/nvvm /usr/local/cuda/nvvm
COPY --from=framework /usr/local/cuda/lib64/libcudart.so* /usr/local/cuda/lib64/
RUN ln -s /usr/local/cuda/lib64/libcublas.so.12 /usr/local/cuda/lib64/libcublas.so
RUN ln -s /usr/local/cuda/lib64/libcublasLt.so.12 /usr/local/cuda/lib64/libcublasLt.so
````

**Review by nv-tusharma — COMMENTED** [review]
There are a few comments I've made so feel free to check those out but overall LGTM.

File: docker/Dockerfile (hunk 5 above)
Comment (nv-tusharma) [reply docker/Dockerfile]: Not required.

File: docker/Dockerfile (hunk 6 above)
Comment (ishandhanani) [reply docker/Dockerfile:450]: Good catch

Comment (ishandhanani) [reply docker/Dockerfile:450]: Worked

File: docker/Dockerfile (hunk 2 above)
Comment (ishandhanani) [reply docker/Dockerfile]: Will do this in a follow up

**Review by nv-tusharma — APPROVED** [review]

File: docker/Dockerfile [hunk 7]
```diff
@@ -197,65 +287,79 @@ RUN --mount=type=cache,target=/root/.cache/pip python3 -m pip install \
...
+# Install development tools
+RUN --mount=type=cache,target=/var/cache/apt,id=framework-apt \
+    apt-get update && apt-get install -y --no-install-recommends \
     gdb \
     ninja-build \
     vim \
     tmux \
     htop \
-    wget \
-    curl \
```
Fridge003, line 224: Why removing these packages? Is it because they are repeatedly installed?

ishandhanani: Yep

PR #13871, opened 2025-11-25
Fix nightly test job to fail when any test fails

alisonshao:
## Summary
- Fixed a bug where nightly test jobs showed success even when tests failed
- The `run_suite_nightly.py` script was not using the return value from `run_unittest_files()`
- Now properly exits with non-zero code when any test fails, while still running all tests with `--continue-on-error`

## Test plan
- [x] Verify the script still runs all tests when `--continue-on-error` is set
- [x] Verify the script exits with non-zero code when any test fails

Example from failed job that incorrectly showed success:
https://github.com/sgl-project/sglang/actions/runs/19653967572/job/56286664501

### DIFF
--- a/test/run_suite_nightly.py
+++ b/test/run_suite_nightly.py
@@ -1,5 +1,6 @@
 import argparse
 import os
+import sys
 from pathlib import Path
 
 from sglang.test.ci.ci_utils import TestFile, run_unittest_files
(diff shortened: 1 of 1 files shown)


alisonshao:
Nightly Workflow: https://github.com/sgl-project/sglang/actions/runs/19655494305/job/56291098594

PR #13876, opened 2025-11-25
call check_quantized_moe_compatibility after initialize

chunyuan-w:
## Motivation
Fixes the error when running `DeepSeek-V3.1-Terminus-FP8` with TP=6 on CPU.
```
  File "/sglang/srt/model_executor/model_runner.py", line 306, in __init__
    self.check_quantized_moe_compatibility()
  File "/sglang/srt/model_executor/model_runner.py", line 597, in check_quantized_moe_compatibility
    raise ValueError(
ValueError: moe_intermediate_size 2048 must be divisible by moe_tp_size (6) which is tp_size (6) divided by moe_ep_size (1).
```

## Modifications
Move `check_quantized_moe_compatibility()` to be after `self.initialize(min_per_gpu_memory)`.

On CPU, we will do padding for `moe_intermediate_size` if it can't be divided by the `tp_size` inside `self.initialize(min_per_gpu_memory)`:
https://github.com/sgl-project/sglang/blob/de430b67457abb527dfceccf83647ae046928085/python/sglang/srt/model_executor/model_runner.py#L775-L778

We need to call `check_quantized_moe_compatibility()` after this padding otherwise we will run into the above error.
We're not able to move the `adjust_config_with_unaligned_cpu_tp()` call to an earlier place because it requires `self.load_config` to be set first and this is done inside the `self.initialize(min_per_gpu_memory)` here:
https://github.com/sgl-project/sglang/blob/de430b67457abb527dfceccf83647ae046928085/python/sglang/srt/model_executor/model_runner.py#L765-L774


mingfeima (approved):

chunyuan-w:
@zhyncs @Alcanderian could you please help review this PR? The CI failures are unrelated.

Alcanderian (approved):

chunyuan-w:
Hi @Alcanderian I checked that the CI failures are unrelated to this PR. Could you please help land this PR?

PR #13923, opened 2025-11-25
Support w8a8_fp8_block_quant for moe and normal linear with llm-compressor

Wangzheee:
w8a8_fp8_block_quant

## Motivation
**support llm-compressor quant model use deepep**
**support llm-compressor quant model with w8a8_fp8(block-quant)**

Use `recipe` changed to target different quantization algorithms or formats.
recipe = QuantizationModifier(
      targets="Linear",
      scheme="FP8_BLOCK",
      ignore=["lm_head", "re:.*mlp.gate$"],
)

- support w8a8_fp8(block-quant)

- support deepep

## Checklist

### DIFF
--- a/python/sglang/srt/layers/linear.py
+++ b/python/sglang/srt/layers/linear.py
@@ -982,7 +982,7 @@ class QKVParallelLinear(ColumnParallelLinear):
         if isinstance(param, BlockQuantScaleParameter):
             if hasattr(self.quant_method, "quant_config"):
                 weight_block_size = self.quant_method.quant_config.weight_block_size
-            elif hasattr(self, "weight_block_size"): # llm-compressor quant model
+            elif hasattr(self, "weight_block_size"):  # llm-compressor quant model
                 weight_block_size = self.weight_block_size
             else:
                 raise ValueError(
--- a/python/sglang/srt/layers/linear.py
+++ b/python/sglang/srt/layers/linear.py
@@ -980,7 +980,14 @@ class QKVParallelLinear(ColumnParallelLinear):
         shard_size = self._get_shard_size_mapping(loaded_shard_id)
 
         if isinstance(param, BlockQuantScaleParameter):
-            weight_block_size = self.quant_method.quant_config.weight_block_size
+            if hasattr(self.quant_method, "quant_config"):
+                weight_block_size = self.quant_method.quant_config.weight_block_size
+            elif hasattr(self, "weight_block_size"): # llm-compressor quant model
+                weight_block_size = self.weight_block_size
+            else:
+                raise ValueError(
+                    "BlockQuantScaleParameter requires weight_block_size to be set."
+                )
             raw_block_n, _ = weight_block_size[0], weight_block_size[1]
             block_n = 1 if getattr(param, "format_ue8m0", False) else raw_block_n
             shard_offset = (shard_offset + block_n - 1) // block_n
--- a/python/sglang/srt/layers/moe/ep_moe/layer.py
+++ b/python/sglang/srt/layers/moe/ep_moe/layer.py
@@ -105,7 +105,7 @@ class DeepEPMoE(FusedMoE):
             and isinstance(quant_config, CompressedTensorsConfig)
             and quant_config._is_wfp8afp8_moe
             and getattr(self.quant_method, "block_quant", False)
-        ):
+        ):  # support W8A8-FP8_BLOCK with llm-compressed-tensors: https://github.com/vllm-project/llm-compressor/blob/main/examples/quantization_w8a8_fp8/fp8_block_example.py
             self.deprecate_flag = True
         else:
             self.deprecate_flag = False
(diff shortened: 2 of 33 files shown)


python/sglang/srt/layers/quantization/utils/quant_utils.py
```diff
@@ -0,0 +1,215 @@
+# Adapted from https://github.com/vllm-project/vllm/model_executor/layers/quantization/utils/quant_utils.py
+# SPDX-License-Identifier: Apache-2.0
+"""This file is used for /tests and /benchmarks"""
+
+from dataclasses import dataclass
```
AniZpZ, line 5: what is the main difference between quant_utils.py and original utils.py?

Wangzheee: We want to refer to vllm and extract the quantization universal API into this folder, while other quantization files of different data types and hardware are separated into different files (such as int8_utils.py fp8_utils.py mxfp4_utils.py )

Wangzheee: > what is the main difference between quant_utils.py and original utils.py?

We want to refer to vllm and extract the quantization universal API into this folder, while other quantization files of different data types and hardware are separated into different files (such as int8_utils.py fp8_utils.py mxfp4_utils.py )

python/sglang/srt/layers/linear.py
```diff
@@ -966,7 +966,11 @@ def weight_loader_v2(
         shard_size = self._get_shard_size_mapping(loaded_shard_id)
 
         if isinstance(param, BlockQuantScaleParameter):
-            weight_block_size = self.quant_method.quant_config.weight_block_size
+            if hasattr(self.quant_method, "quant_config"):
+                weight_block_size = self.quant_method.quant_config.weight_block_size
+            else:
+                assert hasattr(self, "weight_block_size")  # llm-compressor
```
BBuf: Add an error message after the assert statement.

python/sglang/srt/layers/moe/ep_moe/layer.py
```diff
@@ -93,6 +96,13 @@ def __init__(
             quant_config, Fp8Config
         ):
             self.deprecate_flag = True
+        elif (
```
BBuf, line 103: What's the rationale behind deprecating the block_quant + deepgemm + deepep combination? Could you clarify which runtime paths this change will impact? Would suggest adding a code comment to document this.

python/sglang/srt/layers/quantization/compressed_tensors/compressed_tensors.py
```diff
@@ -48,7 +48,7 @@
 __all__ = ["CompressedTensorsLinearMethod"]
 
 SPARSITY_CONFIG_NAME: Literal["sparsity_config"] = "sparsity_config"
-QUANTIZATION_SCHEME_MAP_TYPE = Dict[str, Optional[Dict[str, QuantizationArgs]]]
+QUANTIZATION_SCHEME_MAP_TYPE = dict[str, Optional[dict[str, QuantizationArgs]]]
```
BBuf, line 52: Could these modifications potentially break compatibility with Python 3.10 and similar versions?

Wangzheee: > Could these modifications potentially break compatibility with Python 3.10 and similar versions?
 
https://github.com/sgl-project/sglang/blob/92c29d43ac5f9a316e3d5470c1465591f4919c2d/python/pyproject.toml#L10
https://docs.python.org/3.10/library/stdtypes.html#mapping-types-dict
The current minimum version of Python is 3.10, and it is recommended to use dict list instead of Dict List

python/sglang/srt/layers/linear.py
Wangzheee: > Add an error message after the assert statement.

Got it.

python/sglang/srt/layers/moe/ep_moe/layer.py
Wangzheee: When its config matches CompressedTensorsConfig._is_wfp8afp8_moe, we want its behavior to be consistent with Fp8Config. I will add comments

python/sglang/srt/layers/quantization/fp8_utils.py
```diff
@@ -1163,3 +1175,135 @@ def validate_fp8_block_shape(
...
+    weight: torch.Tensor,
+    weight_scale: torch.Tensor,
+    logical_widths: list[int],
+    input_scale: torch.Tensor | None = None,
+) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor | None]:
+    """Process weights for tensor-wise quantization strategy."""
+    from sglang.srt.layers.quantization.fp8_utils import normalize_e4m3fn_to_e4m3fnuz
+    from sglang.srt.layers.quantization.utils.utils import requantize_with_max_scale
+
+    if _is_fp8_fnuz:
```
Alcanderian, line 1264: I believe put `from sglang.srt.layers.quantization.fp8_utils import normalize_e4m3fn_to_e4m3fnuz` here is a better choice

hnyls2002:
Hi @Wangzheee, w8a8 fp8 block quantization for llm-compressor checkpoints (both MoE and linear) has since landed on main via #14213 and the compressed-tensors scheme refactor in #17503 - see the BLOCK strategy handling in compressed_tensors_w8a8_fp8_moe.py. Closing as superseded - please reopen if I've missed something.

PR #13932, opened 2025-11-25
feat: use class path string as custom logit processor parameter

pengye91:
## Motivation

1. For built in custom logit processors, instead of importing from sglang package, which requires `sglang` as a dependency for clients, use class path string as the extra parameter.

now we can use the `custom_logit_processor` parameter in client codes without `import sglang` like this:
```python
import openai
from rich.pretty import pprint

client = openai.Client(base_url="http://127.0.0.1:30000/v1", api_key="*")
    model="deepseek-ai/DeepSeek-R1",
    messages=[
        {
            "role": "user",
            "content": "Question: Is Paris the Capital of France?",
        }
    ],
    max_tokens=1024,
    extra_body={
        "custom_logit_processor": "sglang.srt.sampling.custom_logit_processor:DeepSeekR1ThinkingBudgetLogitProcessor",
        "custom_params": {
            "thinking_budget": 512,
        },
    },
)
pprint(response)
```

2. Raise a `ValueError` at request initialization if the passed `custom_logit_processor` parameter is not valid, instead of raising a `RuntimeError` which will directly kill the server.
3. Add a `validate` method in `GenerateReqInput` class for all possible validation process.

## Modifications

1. Added a `validate` method in `GenerateReqInput` class for all possible validation process.
2. Added `_validate_custom_logit_processor` method in `GenerateReqInput` class to validate if the `custom_logit_processor` parameter is valid.
3. Added a `from_class_path` method in `CustomLogitProcessor` class to load the logit processor from class path string.
5. Keep the old custom logit processor loading method for compatibility and for not built-in logit processors.
6. Added related test cases.
7. Added related docs.

## Checklist

### DIFF
--- a/python/sglang/srt/managers/io_struct.py
+++ b/python/sglang/srt/managers/io_struct.py
@@ -26,6 +26,7 @@ from typing import TYPE_CHECKING, Any, Dict, List, Literal, Optional, Union
 from sglang.srt.lora.lora_registry import LoRARef
 from sglang.srt.managers.schedule_batch import BaseFinishReason
 from sglang.srt.multimodal.mm_utils import has_valid_data
+from sglang.srt.sampling.custom_logit_processor import CustomLogitProcessor
 from sglang.srt.sampling.sampling_params import SamplingParams
 from sglang.srt.utils import ImageData
 
@@ -255,6 +256,11 @@ class GenerateReqInput(BaseReq):
             self._normalize_single_inputs()
         else:
             self._normalize_batch_inputs()
+
+    def validate(self):
+        """Validate the request parameters."""
+        self._validate_custom_logit_processor()
+        # TODO: Add more validations as needed
 
     def _validate_inputs(self):
         """Validate that the input configuration is valid."""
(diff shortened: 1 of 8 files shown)


python/sglang/srt/sampling/custom_logit_processor.py
```diff
@@ -20,6 +21,20 @@ def _cache_from_str(json_str: str):
     return dill.loads(bytes.fromhex(data["callable"]))
 
 
+@lru_cache(maxsize=None)
+def _cache_from_class_path(class_path: str):
+    """Get the processor directly from the class path."""
+    class_path_split_list = class_path.split(":")
```
CatherineSue, line 27: Any reason we use `:` as separator?

CatherineSue (commented):
overall LGTM. Might need another peer review from experts. I can approve once CI passes

python/sglang/srt/sampling/custom_logit_processor.py
pengye91: I'm following a common pattern for WSGI applications, like `myapp.wsgi:application`, and `uvicorn run module.path:app_object`, and even [vllm logit processor initialization method](https://docs.vllm.ai/en/latest/features/custom_logitsprocs/#method-1-pass-the-custom-logits-processor-fully-qualified-class-name-fqcn-to-vllm-at-initialization-time), but I can change it to dot separator if that's compatible with SGlang's code style.

acelyc111:
Hi @pengye91, is this work on going? There is a CVE which this PR can resolve https://antiproof.ai/blog/three-rces-in-sglang/, we can push it forward.

python/sglang/srt/managers/io_struct.py
```diff
@@ -256,6 +257,11 @@ def normalize_batch_and_arguments(self):
         else:
             self._normalize_batch_inputs()
 
+    def validate(self):
+        """Validate the request parameters."""
+        self._validate_custom_logit_processor()
+        # TODO: Add more validations as needed
+
     def _validate_inputs(self):
```
acelyc111, line 265: How about moving these `_validate_*` into the newly added `validate()`?

PR #13944, opened 2025-11-26
Fix utils import issue for nightly tests

alisonshao:
## Summary
Fix the `ModuleNotFoundError: No module named 'utils'` error in nightly tests.

Several nightly LoRA tests import from `utils`, but the file is located at `test/srt/lora/utils.py` which is not in the Python path when running from `test/nightly/`.

## Note
The original sgl-kernel version mismatch issue is being addressed by #13951 (bot bump workflow).

## Test plan
- [ ] Verify nightly LoRA tests can import utils module
- [ ] Verify nightly tests pass


python/pyproject.toml
```diff
@@ -57,7 +57,7 @@ dependencies = [
   "scipy",
   "sentencepiece",
   "setproctitle",
-  "sgl-kernel==0.3.17.post2",
+  "sgl-kernel==0.3.18",
```
Fridge003: Should be 0.3.18.post1 here

test/utils.py
```diff
@@ -0,0 +1,573 @@
+# Copyright 2023-2024 SGLang Team
```
Fridge003: Why adding this file?

Fridge003:
The "bump sgl-kernel to sglang" workflow is missing. Shouldn't we use that one?
Since the version in other places (like dockerfile) also needs to be fixed

Kangyan-Zhou:
@Fridge003 @alisonshao created fixed the github workflow issue and created https://github.com/sgl-project/sglang/pull/13951

test/utils.py
alisonshao: this file is added because several tests use import from utils, but the file is located at test/srt/lora/utils.py which is not in the current python path when running from test/nightly/. this is related to reorganization of ci tests.

alisonshao:
https://github.com/sgl-project/sglang/actions/runs/19690528758/job/56405080135
The following tests passed with this fix: 
test_lora_qwen3.py
test_lora_radix_cache.py

test/lora_utils.py
```diff
@@ -0,0 +1,573 @@
+# Copyright 2023-2024 SGLang Team
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
```
Kangyan-Zhou: Remove the comments

alisonshao: done

alisonshao (commented):
Consider adding a TODO comment to indicate this can be cleaned up later:

```python
# Add test directory to path for lora_utils import
# TODO: can be removed after migration
sys.path.insert(0, str(Path(__file__).resolve().parent.parent.parent))
```

This applies to the 4 files: test_lora.py, test_lora_backend.py, test_lora_tp.py, test_multi_lora_backend.py

Issue #13809, opened 2025-11-23
[Feature] Reduce constrained-decoding overhead in TP


DarkSharpness:
### Checklist

- [x] If this is not a feature request but a general question, please start a discussion at https://github.com/sgl-project/sglang/discussions. Otherwise, it will be closed.
- [x] Please use English. Otherwise, it will be closed.

### Motivation

Currently, when tensor-parallelism (TP) and constrained-decoding are enabled, each TP worker will compile the same grammar across different TP ranks. This will incur non-trivial CPU overhead (for TP $n$, the overhead is $n \times$. In fact, we only need to compile the grammar and sample the result on the first rank.

A possible implementation can be:

1. Only apply grammar token mask on the rank 0.
2. Broadcast the next tokens id from rank 0 when there're grammars in batch.
3. Keep the old code path when there's no grammar in batch (i.e. no extra all-reduce).

https://github.com/sgl-project/sglang/blob/5c2915494c83f076a11afb2c3382eeb8a41f1974/python/sglang/srt/layers/sampler.py#L205-L217

raayandhar:
I'd like to tackle this!

bandham-manikanta:
Hi @raayandhar @DarkSharpness @hnyls2002 — checking if this is still free.

#13947 landed and was reverted by #16845 (“DP Attention is broken”). On current `main`, `_sync_token_ids_across_tp` still does `all_reduce(MIN)` when grammars are present, and grammar compile still runs on every TP rank.

I’d like to re-land this with the correct sync group under `--enable-dp-attention` (not global `tp_rank == 0`), and validate on 4×A100 with TP=2/4 + JSON-schema requests. Happy to leave it if @raayandhar is still working on it — just let me know.

tlysanhuo:
Hi @hnyls2002 @DarkSharpness — I'd like to help land this on current `main`.

#16874 has the right idea but has been idle since February. `main` has since refactored this area into `GrammarManager` (which already syncs ready/failed sets across the DP/TP group via `all_gather_object`), so I've prepared a fresh implementation on top of that instead of rebasing the old branch:

1. Only the entry rank of the DP/TP group submits grammar compilation; non-entry ranks carry a placeholder object and never compile at all (unlike the lazy re-fetch in #16874 — the grammar backend cache is per-process, so re-fetching still compiled n×).
2. Only the entry rank computes/applies the grammar vocab mask per step.
[...]

PR #13947, opened 2025-11-26
feat: reduce constrained-decoding overhead in TP

raayandhar:
## Motivation

See https://github.com/sgl-project/sglang/issues/13809

## Modifications

The issue nicely laid out a strategy we could use.

## Benchmarking and Profiling

Let me know what kind of inference benchmarking I should / could do here, unfamiliar.

## Checklist


raayandhar:
@DarkSharpness could you take another look and let me know what you think?

I added some debugging logs to verify behavior, it looks good to me:
```
[2025-12-18 04:53:21 TP1] [TP Rank 1] Skipping grammar compilation for req 4060c65f2eac4c88b2d61a04a5194c28. Wai
ting for rank 0 to compile.                                                                                     
[2025-12-18 04:53:21 TP0] [TP Rank 0] Grammar cache hit for req 4060c65f2eac4c88b2d61a04a5194c28, type=json     
[2025-12-18 04:53:21 TP0] [TP Rank 0] Grammar queue sync: 1 requests ready (compiled on this rank)              
[2025-12-18 04:53:21 TP1] [TP Rank 1] Grammar queue sync: 1 requests ready (1 skipped compilation, delegated to 
rank 0)  
... (more requests later, they all look like this)
[2025-12-18 04:53:22 TP0] [TP Rank 0] Grammar cache hit for req ec956e20b57f4f429b22cdfb823192c4, type=json     
[2025-12-18 04:53:22 TP1] [TP Rank 1] Skipping grammar compilation for req ec956e20b57f4f429b22cdfb823192c4. Wai
ting for rank 0 to compile.                                                                                     
[2025-12-18 04:53:22 TP0] [TP Rank 0] Grammar queue sync: 1 requests ready (compiled on this rank)              
[2025-12-18 04:53:22 TP1] [TP Rank 1] Grammar queue sync: 1 requests ready (1 skipped compilation, delegated to 
rank 0)          
```

python/sglang/srt/managers/scheduler.py
```diff
@@ -2311,6 +2329,12 @@ def move_ready_grammar_requests(self):
                 req = self.grammar_queue[i]
                 if req.finished():  # It is aborted by AbortReq
                     continue
+                # On non-rank-0, req.grammar is None
```
DarkSharpness: I think we could change the `all_reduce` max to `broadcast` here (rank 0 tells other ranks how many grammars has completed). The rank=0 and rank>0 logic is quite different here so I think we should not mix them togther

python/sglang/srt/managers/scheduler.py
```diff
@@ -1515,39 +1515,47 @@ def handle_generate_request(
...
-                )
-                req.grammar = value
-
-                if not cache_hit:
-                    req.grammar_key = key
-                    add_to_grammar_queue = True
+        if req.sampling_params.has_grammar_constraint:
+            if self.tp_rank == 0:
+                # Only rank 0 compiles grammar and manages the cache
+                if self.grammar_backend is None:
```
DarkSharpness: I'm not sure about the error handling logic here. What would happen when `tp>1` and `grammar_backend` is `None`.

Before this PR, requests on all rank will call `set_finish_with_abort`. After this PR, it seems only rank 0 will call this. Is this behavior still correct?

DarkSharpness:
> > @raayandhar Thanks for the contribution. The code looks good to me and I left some comments.
> 
> Thanks! - but I don't see any comments?

I forgot to submit the review (my bad

raayandhar:
I'm looking into it. I need to do some more careful analysis and testing. Perhaps a larger part needs to be refactored. Investigating.

python/sglang/srt/managers/scheduler.py
raayandhar: Yes, I think you're right, thanks for this catch. We are only broadcasting the count of ready requests. So I added a change where the non-rank 0 workers will also terminate these requests if there is error, since we will have requests on all workers.

python/sglang/srt/managers/scheduler.py
raayandhar: Thanks, I made that change.

raayandhar:
The only relevant error I see in CI is related to my new unit tests with TP=2 not having enough GPUs:

```
[2025-12-28 04:23:15] Watchdog DetokenizerManager initialized.
[2025-12-28 04:23:16 TP1] Init torch distributed begin.
[2025-12-28 04:23:16 TP1] Context: self.device='cuda' self.gpu_id=1 os.environ.get('CUDA_VISIBLE_DEVICES')='0' self.tp_rank=1 self.tp_size=2
[2025-12-28 04:23:16 TP1] Scheduler hit an exception: Traceback (most recent call last):
  File "/public_sglang_ci/runner-l1b-gpu-0/_work/sglang/sglang/python/sglang/srt/managers/scheduler.py", line 2984, in run_scheduler_process
    scheduler = Scheduler(
  File "/public_sglang_ci/runner-l1b-gpu-0/_work/sglang/sglang/python/sglang/srt/managers/scheduler.py", line 333, in __init__
    self.init_model_worker()
  File "/public_sglang_ci/runner-l1b-gpu-0/_work/sglang/sglang/python/sglang/srt/managers/scheduler.py", line 471, in init_model_worker
    self.tp_worker = TpModelWorker(
  File "/public_sglang_ci/runner-l1b-gpu-0/_work/sglang/sglang/python/sglang/srt/managers/tp_worker.py", line 253, in __init__
    self._model_runner = ModelRunner(
  File "/public_sglang_ci/runner-l1b-gpu-0/_work/sglang/sglang/python/sglang/srt/model_executor/model_runner.py", line 337, in __init__
    min_per_gpu_memory = self.init_torch_distributed()
  File "/public_sglang_ci/runner-l1b-gpu-0/_work/sglang/sglang/python/sglang/srt/model_executor/model_runner.py", line 707, in init_torch_distributed
    torch.get_device_module(self.device).set_device(self.gpu_id)
  File "/usr/local/lib/python3.10/dist-packages/torch/cuda/__init__.py", line 567, in set_device
    torch._C._cuda_setDevice(device)
torch.AcceleratorError: CUDA error: invalid device ordinal
GPU device may be out of range, do you have enough GPUs?
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
```
is there a way to set that these tests should only run in CI machines with min. 2 GPUs?

cc @DarkSharpness

ghost (changes requested):
couple nits/questions

python/sglang/srt/managers/scheduler.py
```diff
@@ -770,7 +770,7 @@ def init_chunked_prefill(self):
 
     def init_grammar_backend(self):
         self.grammar_queue: List[Req] = []
-        if not self.server_args.skip_tokenizer_init:
+        if not self.server_args.skip_tokenizer_init and self.tp_rank == 0:
```
ghost: why not keep a backend on all ranks so non-entry ranks can read cached grammars without recompiling. only entry rank will trigger compilation

python/sglang/srt/managers/scheduler.py
```diff
@@ -1546,39 +1546,52 @@ def handle_generate_request(
...
-                    req.grammar_key = key
-                    add_to_grammar_queue = True
+        if req.sampling_params.has_grammar_constraint:
+            if self.tp_rank == 0:
+                # Only rank 0 compiles grammar and manages the cache
+                if self.server_args.grammar_backend == "none":
+                    error_msg = "Grammar-based generation (json_schema, regex, ebnf, structural_tag) is not supported when the server is launched with --grammar-backend none"
+                    req.set_finish_with_abort(error_msg)
+                    if self.tp_size > 1:
+                        add_to_grammar_queue = True
```
ghost, line 1552:
```suggestion
            if req.sampling_params.json_schema is not None:
                key = ("json", req.sampling_params.json_schema)
            elif req.sampling_params.regex is not None:
                key = ("regex", req.sampling_params.regex)
            elif req.sampling_params.ebnf is not None:
                key = ("ebnf", req.sampling_params.ebnf)
            else:
                key = ("structural_tag", req.sampling_params.structural_tag)
```

python/sglang/srt/managers/scheduler.py
```diff
@@ -1546,39 +1546,52 @@ def handle_generate_request(
...
-                        error_msg = f"Invalid grammar request with cache hit: {key=}"
-                        req.set_finish_with_abort(error_msg)
+                    if req.sampling_params.json_schema is not None:
+                        key = ("json", req.sampling_params.json_schema)
+                    elif req.sampling_params.regex is not None:
+                        key = ("regex", req.sampling_params.regex)
+                    elif req.sampling_params.ebnf is not None:
+                        key = ("ebnf", req.sampling_params.ebnf)
+                    elif req.sampling_params.structural_tag:
+                        key = ("structural_tag", req.sampling_params.structural_tag)
```
ghost:
```suggestion
            if self.tp_rank == 0:
                # Only rank 0 compiles grammar and manages the cache
                if self.server_args.grammar_backend == "none":
                    error_msg = "Grammar-based generation (json_schema, regex, ebnf, structural_tag) is not supported when the server is launched with --grammar-backend none"
                    req.set_finish_with_abort(error_msg)
                    if self.tp_size > 1:
                        add_to_grammar_queue = True
                else:
```

python/sglang/srt/managers/scheduler.py
```diff
@@ -1546,39 +1546,52 @@ def handle_generate_request(
...
+                            req.set_finish_with_abort(error_msg)
+
+                    # When TP > 1, always add to grammar_queue for synchronization across ranks
+                    if self.tp_size > 1:
+                        add_to_grammar_queue = True
+                    elif not cache_hit:
+                        # TP == 1: only add to queue if we didn't get a cache hit.
+                        add_to_grammar_queue = True
+            else:
+                # Non-rank-0 workers add to grammar queue for synchronization, but no compilation.
```
ghost, line 1571:
```suggestion
                # Non-rank-0 workers add to grammar queue for synchronization, but no compilation.
                req.grammar_key = key
```

python/sglang/srt/managers/scheduler.py
```diff
@@ -2379,35 +2408,84 @@ def move_ready_grammar_requests(self):
...
+                # Non-rank-0 workers also abort invalid grammar requests
+                if self.tp_rank != 0:
+                    for idx in invalid_indices_tensor.tolist():
+                        if idx < len(self.grammar_queue):
+                            req = self.grammar_queue[idx]
+                            error_msg = f"Invalid grammar request (detected by rank 0)"
+                            req.set_finish_with_abort(error_msg)
+
         for req in self.grammar_queue[:num_ready_reqs]:
             self._add_request_to_queue(req)
```
ghost, line 2491:
```suggestion
            # Non-entry ranks: attach cached grammar so control-flow (e.g., termination) stays in sync.
            if (
                req.grammar is None
                and self.grammar_backend is not None
                and req.grammar_key is not None
            ):
                grammar_obj, cache_hit = self.grammar_backend.get_cached_or_future_value(
                    req.grammar_key, req.require_reasoning
                )
                if not cache_hit:
                    req.set_finish_with_abort(
                        f"Grammar not cached on non-entry rank: {req.grammar_key=}"
                    )
                elif grammar_obj is INVALID_GRAMMAR_OBJ:
                    req.set_finish_with_abort(
                        f"Invalid grammar request with cache hit: {req.grammar_key=}"
                    )
                else:
                    req.grammar = grammar_obj
                    
            self._add_request_to_queue(req)
```

ghost (changes requested):
here are some problems I suggest re-ordering for:
1. Non-entry TP ranks instantiate no grammar backend, so they can’t fetch the compiled grammar later.
2. Those ranks also never set req.grammar_key, so even with a backend they have nothing to look up.
3. Rwquests dequeued from the grammar queue still have `req.grammar=None` on non-entry ranks. When rank 0 terminates but others don’t, batches will diverge and the token-id broadcast will hang due to shape mismatch.

Also, the problem that is messing with your tp=2 breaking constrained decoding (IMO) is that in your code, if rank 0 has a real grammar and rank 1 has `None`, then rank 0 can terminate early but rank 1 cannot. if they disagree they will build different batches. If only rank 0 can term, then tp will desync.

raayandhar:
> here are some problems I suggest re-ordering for:
> 
>     1. Non-entry TP ranks instantiate no grammar backend, so they can’t fetch the compiled grammar later.
> 
>     2. Those ranks also never set req.grammar_key, so even with a backend they have nothing to look up.
> 
>     3. Rwquests dequeued from the grammar queue still have `req.grammar=None` on non-entry ranks. When rank 0 terminates but others don’t, batches will diverge and the token-id broadcast will hang due to shape mismatch.
> 
> 
> Also, the problem that is messing with your tp=2 breaking constrained decoding (IMO) is that in your code, if rank 0 has a real grammar and rank 1 has `None`, then rank 0 can terminate early but rank 1 cannot. if they disagree they will build different batches. If only rank 0 can term, then tp will desync.

Sounds good, thanks for the feedback. I've made the relevant changes like you proposed. This is different approach than what I had originally done, so cc @DarkSharpness to take another look at these changes.

Tests are still all passing:
```
Ran 92 tests in 165.159s

OK
```
and
```
Ran 92 tests in 165.159s

OK
```
I also make the CI skip the TP=2 tests if there aren't enough GPUs, so CI shouldn't have issue anymore.

raayandhar:
> /tag-and-rerun-ci

The `test_constrained_decoding.py` test is passing now, the other failures look unrelated.

python/sglang/srt/managers/scheduler.py
```diff
@@ -2349,12 +2359,26 @@ def move_ready_grammar_requests(self):
...
+            tp_size = self.tp_size
+
         for req in self.grammar_queue:
             try:
                 if req.finished():  # It is aborted by AbortReq
                     num_ready_reqs += 1
                     continue
 
+                if tp_size > 1:
+                    if req.grammar is None:
```
DarkSharpness: What does `req.grammar is None` mean? Please add detailed comments here.

python/sglang/srt/managers/scheduler.py
```diff
@@ -2378,36 +2402,73 @@ def move_ready_grammar_requests(self):
...
-            torch.distributed.all_reduce(
-                tensor, op=torch.distributed.ReduceOp.MAX, group=tp_group
-            )
+
+            if self.server_args.enable_dp_attention:
+                root_rank = self.attn_tp_group.ranks[0]
+            else:
+                root_rank = self.tp_group.ranks[0]
+
+            torch.distributed.broadcast(tensor, src=root_rank, group=tp_group)
```
DarkSharpness: why use `root_rank` here but using `self.tp_rank == 0` in other places?

I'm not sure whether they refer to the same thing (just to prevent some potential bugs)

python/sglang/srt/managers/scheduler.py
```diff
@@ -2378,36 +2402,73 @@ def move_ready_grammar_requests(self):
...
+        # Only rank 0 handles timeouts.
+        if self.tp_rank == 0:
+            for i in range(num_ready_reqs, num_ready_reqs + num_timeout_reqs_max):
+                req = self.grammar_queue[i]
+                if not isinstance(req.grammar, futures.Future):
+                    continue
+                req.grammar.cancel()
+                self.grammar_backend.set_cache(req.grammar_key, INVALID_GRAMMAR_OBJ)
+                error_msg = f"Grammar preprocessing timed out for {req.grammar_key=}"
+                req.set_finish_with_abort(error_msg)
```
DarkSharpness: what about the consistency here? it seems only rank 0 call `set_finish_with_abort`. is this expected?

python/sglang/srt/managers/scheduler.py
ghost: req.grammar carries the state of the rank in regards to grammar. None simply means that the rank doesnt hodl a usable grammar obj for a request. This is specifically used for TP>1, non-entry ranks.

ghost: if tp>1 and this rank doesnt own compilation, its ready fr queue counting because the rank isnt waiting on a future. if its already a concrete grammar object, it will also be treated as ready.

TLDR this only exists to avoid treating non-owner TP ranks as waiting on grammar compilation.

ghost: the behavior we want is for those ranks to avoid calling .result() on a non-existent future and still advance the num_ready_reqs consistently so the queue is emptied in sync across TP ranks.

python/sglang/srt/managers/scheduler.py
ghost: `self.tp_rank == 0` just means local rank is 0 within its TP group.
`root_rank = self.tp_group.ranks[0]` is the global rank id of the first process in this group.

the torch broadcast requires a global rank so we use rook rank for that. i think if you make them same it wont impact.

python/sglang/srt/managers/scheduler.py
```diff
@@ -2402,18 +2414,15 @@ def move_ready_grammar_requests(self):
...
-            torch.distributed.broadcast(tensor, src=root_rank, group=tp_group)
+            # Broadcast from local rank 0 within the TP group, using group_src (local rank)
+            # rather than src (global rank) to be consistent with other self.tp_rank == 0 checks
+            torch.distributed.broadcast(tensor, group_src=0, group=tp_group)
             num_ready_reqs_max, num_timeout_reqs_max = tensor.tolist()
 
-            # Only rank 0 needs to finish compiling remaining grammars
+            # Rank 0 may need to wait for additional grammars that other ranks counted as ready
             if self.tp_rank == 0:
                 for i in range(num_ready_reqs, num_ready_reqs_max):
```
DarkSharpness: rank 0 should not wait for additional grammars. Instead, other ranks should only mark the `num_ready_reqs` from rank 0 as ready.

In addition, now when `tp_rank=0`, `num_ready_reqs` = `num_ready_reqs_max`, which means this loop is now no-op. Is this expected?

BTW, the original variable naming `num_ready_reqs_max` is now misleading here (it is broadcasted from rank 0 now). Maybe you can try `num_ready_reqs_local` and `num_ready_reqs_rank_0` or something like that.

python/sglang/srt/managers/scheduler.py
```diff
@@ -2378,36 +2414,70 @@ def move_ready_grammar_requests(self):
...
+                    )
+                )
+
+                # If we got a Future, ensure all ranks have the grammar.
+                if isinstance(grammar_obj, futures.Future):
+                    grammar_obj = grammar_obj.result()
+                    self.grammar_backend.set_cache(req.grammar_key, grammar_obj.copy())
+
+                if grammar_obj is INVALID_GRAMMAR_OBJ:
+                    req.set_finish_with_abort(
```
DarkSharpness, line 2485: invalid req is only aborted on rank=0. I guess we need to broadcast all the invalid positions in advance? Always keep in mind that all tp ranks should be make the same scheduling decision.

python/sglang/srt/managers/scheduler.py
```diff
@@ -2345,16 +2357,40 @@ def maybe_send_health_check_signal(self):
...
+        all other ranks. All ranks process the same number of requests and non-entry workers
+        fetch the compiled grammars from their local cache.
+        """
         num_ready_reqs = 0
         num_timeout_reqs = 0
+
+        if self.server_args.enable_dp_attention:
+            tp_size = self.attn_tp_size
+        else:
+            tp_size = self.tp_size
```
DarkSharpness, line 2371: you can move the code from L2409 to here. it's redundant

```python
        if self.server_args.enable_dp_attention:
            tp_size = self.attn_tp_size
            tp_group = self.attn_tp_cpu_group
        else:
            tp_size = self.tp_size
            tp_group = self.tp_cpu_group
```

python/sglang/srt/managers/scheduler.py
raayandhar: Yes, sorry, I think the comment was not very clear, we are just broadcasting like you describe. I change it so it's better. But yeah, it's a good catch, that code is effectively dead. I removed and simplified and renamed. Passing:
```
Ran 92 tests in 211.157s

OK
```

python/sglang/srt/managers/scheduler.py
raayandhar: Thanks, good catch. I fixed.

python/sglang/srt/managers/scheduler.py
raayandhar: Actually, I'm not quite sure. I had thought the same thing as well, especially since in a round of my changes this [test](https://github.com/sgl-project/sglang/blob/3619ec61b4025a8c90a84ba72a2e28bf9f707831/python/sglang/test/kits/json_constrained_kit.py#L63) (invalid JSON) was failing. So I made [these changes](https://github.com/sgl-project/sglang/pull/13947/changes/478121a3664fbe9800b61016ce8b5ca6536d9f6b#diff-c3b8cc39d10c245933a25aa9c2fd6397f6b31ed8d85c0ecbb926c1f42afdd178R2510) and it was passing again. I think I mention it in an earlier comment. But later, when I applied ilgorm2's changes and some fixes, we still passed this test. 

I'm looking into it.

raayandhar: OK, so I think the reason we are still passing is that the other non-entry rank grammar objects block on `.result()` and will call `set_finish_with_abort()` later.
So I agree we should broadcast it. We shouldn't do redundant work, that's the whole point of these changes.

raayandhar: Here's what we do now:
1) We broadcast the invalid req counts in the same tensor as the ready and timed out reqs. 
2) If we have more than 0 invalid reqs, we send the invalid indices to all ranks and each rank aborts its invalid request.

DarkSharpness:
cc @hnyls2002
can you take a view?

hnyls2002 (approved):

PR #13949, opened 2025-11-26
feat: add test for wan-t2v models

jy-song-hub:
## Motivation

This Pull Request introduces a quality assurance test for the Wan T2V model integration within SGLang.

## Modifications

This PR adds a new file, test_t2v_quality.py, structured as a standard unittest.TestCase class, which orchestrates the following pipeline:
- Video Generation: Executes the sglang generate command (as a subprocess) using the Wan T2V model to create a video from a defined prompt.
- Quality Evaluation:

> - Captioning: Uses a Qwen-VL model to generate a detailed description of the generated video.
> - Judging: Uses a Qwen-Instruct model to compare the generated video caption against the original text prompt, outputting a structured quality score across several metrics.

- Assertion: Asserts that the final calculated overall_score meets a defined pass_threshold = 70.

## Checklist

### DIFF
--- a/python/sglang/multimodal_gen/test/test_wan_t2v_quality.py
+++ b/python/sglang/multimodal_gen/test/test_wan_t2v_quality.py
@@ -1,0 +1,372 @@
+import glob
+import json
+import logging
+import os
+import re
+import shutil
+import subprocess
+import time  # Added for file handle delay
+import unittest
+import uuid
+
+import torch
+
+from sglang.test.test_utils import (  # DEFAULT_TIMEOUT_FOR_SERVER_LAUNCH, # Not used here; is_in_ci, # Not used here
+    CustomTestCase,
+)
+
+# --- Library Imports ---
+# NOTE: Ensure qwen_vl_utils.py is available in the Python path
+try:
+    from qwen_vl_utils import process_vision_info
+    from transformers import (
+        AutoModelForCausalLM,
+        AutoProcessor,
+        AutoTokenizer,
+        Qwen2VLForConditionalGeneration,
+    )
+except ImportError:
+    # Dummy implementation for environments without transformers/qwen
+    class Dummy:
+        pass
+
+    (
+        Qwen2VLForConditionalGeneration,
+        AutoProcessor,
+        AutoModelForCausalLM,
+        AutoTokenizer,
+    ) = [Dummy] * 4
+
+    def process_vision_info(*args, **kwargs):
+        pass
+
+
+class TestWanT2VQuality(CustomTestCase):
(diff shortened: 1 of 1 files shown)


python/sglang/multimodal_gen/runtime/entrypoints/cli/generate.py
```diff
@@ -103,7 +103,12 @@ def generate_cmd(args: argparse.Namespace):
     sampling_params = SamplingParams.from_cli_args(args)
     sampling_params.request_id = generate_request_id()
     generator = DiffGenerator.from_pretrained(
-        model_path=server_args.model_path, server_args=server_args
+        prompt=sampling_params.prompt,
+        sampling_params=sampling_params,
```
mickqian: we would consider supporting overriding these, so we can revert it here

python/sglang/multimodal_gen/test/test_wan_t2v_quality.py
```diff
@@ -0,0 +1,371 @@
+import glob
```
mickqian, line 1: this should be moved into `multimodal_gen` folder

python/sglang/multimodal_gen/runtime/entrypoints/cli/generate.py
jy-song-hub: Reverted — thanks for the clarification. One thing to flag: without overriding support, the test can’t pass its sampling params, so it runs with defaults and uses more resources. OK for now, but override support would help keep tests lightweight. Happy to adjust test parameters if needed.

python/sglang/multimodal_gen/test/test_wan_t2v_quality.py
jy-song-hub: Got it — moved. Thanks!

mickqian:
relies on #13996

python/sglang/multimodal_gen/runtime/entrypoints/cli/generate.py
mickqian: Sure, could you help review #13996?

jy-song-hub: > Sure, could you help review #13996?

I just got back from my Thanksgiving vacation. I’ll review it shortly — thanks for your patience!

jy-song-hub:
> What do the current test results look like?

Here are the current test results. I added this test as a correct-case example, and as expected it passes with a final score of 90 and an OK status. When errors occur, the same test will surface them by producing lower scores along with a short explanation of what was wrong.

Screenshot 2025-12-05 at 3 59 00 PM

PR #13969, opened 2025-11-26
[kernel][moe] add moe topk fast

thenumberouscode:
## Motivation

The original moeTopK kernel reads data from global memory topk times, resulting in significant performance overhead. By finding the top-2 elements in each iteration, we can reduce global memory accesses to topk/2 times, substantially improving kernel efficiency.

## Modifications

To minimize global memory access, we modified each thread to maintain its local top-2 elements. These partial results are then reduced in shared memory to compute the global top-2 values. This optimization significantly reduces the visits of global memory when topk is big.

Additionally, we observed that the original moeTopK kernel exhibits O(topk × (N/TPB) × topk) complexity. We can eliminate the innermost topk factor by modifying input values in-place, which is safe since these inputs are not used after kernel execution.

## Accuracy Tests

with vllm installed:
`python sgl-kernel/benchmark/bench_moe_topk_softmax.py`

## Benchmarking and Profiling
We can also evaluate performance using `python sgl-kernel/benchmark/bench_moe_topk_softmax.py`. Since our analysis focuses on configurations with 512 experts (exceeding the 256 threshold), I have included the relevant performance data for your reference. (I tested it on a rtx3080).
<details>
<summary>optimized</summary>

```bash
num_tokens  num_experts  topk      SGLang
0        128.0        512.0   1.0    3.848165
1        128.0        512.0   2.0    4.117456
2        128.0        512.0   4.0    4.990667
3        128.0        512.0   8.0    6.809333
4        512.0        512.0   1.0    7.089001
5        512.0        512.0   2.0    7.796969
6        512.0        512.0   4.0   10.173922
7        512.0        512.0   8.0   14.973041
8       1024.0        512.0   1.0   10.982928
9       1024.0        512.0   2.0   12.664809
10      1024.0        512.0   4.0   17.226633
11      1024.0        512.0   8.0   26.129326
12      2048.0        512.0   1.0   23.796835
13      2048.0        512.0   2.0   26.927251
14      2048.0        512.0   4.0   35.105438
15      2048.0        512.0   8.0   52.751509
16      4096.0        512.0   1.0   46.619901
17      4096.0        512.0   2.0   51.975619
18      4096.0        512.0   4.0   67.380423
19      4096.0        512.0   8.0  101.822984
20      8192.0        512.0   1.0   89.319637
21      8192.0        512.0   2.0  100.394009
22      8192.0        512.0   4.0  130.987916
23      8192.0        512.0   8.0  199.880580
24     16384.0        512.0   1.0  174.343720
25     16384.0        512.0   2.0  196.797438
26     16384.0        512.0   4.0  258.977689
27     16384.0        512.0   8.0  394.104179
28     32768.0        512.0   1.0  343.884698
29     32768.0        512.0   2.0  389.199051
30     32768.0        512.0   4.0  513.195791
31     32768.0        512.0   8.0  784.106652
```

</details>

<details>
<summary>non-optimized</summary>

```bash
 num_tokens  num_experts  topk       SGLang
0        128.0        512.0   1.0     8.632748
1        128.0        512.0   2.0     8.829262
2        128.0        512.0   4.0    12.072971
3        128.0        512.0   8.0    19.840909
4        512.0        512.0   1.0     8.269698
5        512.0        512.0   2.0     9.041478
6        512.0        512.0   4.0    17.082785
7        512.0        512.0   8.0    24.401590
8       1024.0        512.0   1.0    11.236170
9       1024.0        512.0   2.0    28.351181
10      1024.0        512.0   4.0    21.920593
11      1024.0        512.0   8.0    80.759652
12      2048.0        512.0   1.0    51.757095
13      2048.0        512.0   2.0    32.153553
14      2048.0        512.0   4.0    97.589803
15      2048.0        512.0   8.0   171.029422
16      4096.0        512.0   1.0    89.670263
17      4096.0        512.0   2.0   119.262757
18      4096.0        512.0   4.0   123.065981
19      4096.0        512.0   8.0   334.462982
20      8192.0        512.0   1.0   186.877664
21      8192.0        512.0   2.0   242.138191
22      8192.0        512.0   4.0   359.509325
23      8192.0        512.0   8.0   681.282953
24     16384.0        512.0   1.0   375.869521
25     16384.0        512.0   2.0   481.087326
26     16384.0        512.0   4.0   738.043801
27     16384.0        512.0   8.0  1369.569891
28     32768.0        512.0   1.0   342.455646
29     32768.0        512.0   2.0   928.017044
30     32768.0        512.0   4.0  1413.717333
31     32768.0        512.0   8.0  2646.720052
```

</details>
**The corresponding picture **
moe_performance_comparison_subplots

## Checklist


dongyibo:
Hello, your idea is brilliant—I learned a lot from it!  
But I have a question: inputs_after_softmax is global memory. Will modifying this memory have any performance impact? @thenumberouscode

thenumberouscode:
> Hello, your idea is brilliant—I learned a lot from it! But I have a question: inputs_after_softmax is global memory. Will modifying this memory have any performance impact? @thenumberouscode

Yes, inputs_after_softmax resides in global memory. However, modifying it in-place each round saves (k - 1) * threads_per_block global memory accesses, which is a worthwhile optimization. This is because we set the index of the k-th round's selected element to a sentinel minimum value within inputs_after_softmax itself. Therefore, we avoid the additional global memory accesses needed to consult a separate indices array to locate the previously selected k-th element.

thenumberouscode:
@yuan-luo hi bro, Could you tell me who should I ask to review this PR? thanks.

sgl-kernel/csrc/moe/moe_topk_softmax_kernels.cu
```diff
@@ -143,9 +145,121 @@ __launch_bounds__(TPB) __global__ void moeSoftmax(
...
+    cub_kvp globalMax, globalSecondMax;
+
+    // Determine the global maximum
+    if (candidate1.max.value > candidate2.max.value) {
+      globalMax = candidate1.max;
+    } else {
+      globalMax = candidate2.max;
+    }
+
+    // Determine the global second maximum
```
BBuf, line 171: The logic only compares 3 values instead of all 4 candidates? We need to find the second maximum from `{c1.max, c1.secondMax, c2.max, c2.secondMax}` excluding the global max.

For example, candidate1: max=10, secondMax=8; candidate2: max=9, secondMax=7

- Expected: globalMax=10, globalSecondMax=9
- Current code: globalMax=10, globalSecondMax=8 ?

sgl-kernel/csrc/moe/moe_topk_softmax_kernels.cu
```diff
@@ -143,9 +145,121 @@ __launch_bounds__(TPB) __global__ void moeSoftmax(
...
+      // updating 2 elements to the result.
+      for (int i = 0; i < TopKPair::PAIR; i++) {
+        cub_kvp result = (i == TopKPair::MAX_INDEX) ? result_pair.max : result_pair.secondMax;
+        int expert = result.key;
+        bool node_uses_expert = expert >= start_expert && expert < end_expert;
+        bool should_process_row = row_is_active && node_uses_expert;
+        // The inputs_after_softmax is modified in-place to avoid unnecessary loops for finding the top k-1 value.
+        // 1.f represents the minimum value.
+        inputs_after_softmax[thread_read_offset + expert] = -1.f;
+        int idx = k * block_row + k_idx * 2 + i;
```
BBuf, line 242: Should we  add boundary check ? `if (result_idx >= k) break;` ?

BBuf: Please add some test cases to `test_moe_topk_softmax.py`? 

```python
@pytest.mark.parametrize(
    "num_tokens, num_experts, topk",
    [
        (128, 512, 1),
        (128, 512, 2),
        (128, 512, 3),  # Odd topk
        (128, 512, 5),  # Odd topk
        (128, 512, 8),
        (512, 512, 512),  # Edge case: k = num_experts
    ]
)
```

sgl-kernel/csrc/moe/moe_topk_softmax_kernels.cu
```diff
@@ -592,8 +700,15 @@ void topkGatingSoftmaxKernelLauncher(
       static constexpr int TPB = 256;
       moeSoftmax<T, TPB><<<num_tokens, TPB, 0, stream>>>(
           gating_output, nullptr, softmax_workspace, num_experts, moe_softcapping, correction_bias);
-      moeTopK<TPB><<<num_tokens, TPB, 0, stream>>>(
-          softmax_workspace, nullptr, topk_weights, topk_indices, num_experts, topk, 0, num_experts, renormalize);
+      if (topk == 1) {
```
BBuf, line 704: Why is topk == 1 still using the old `moeTopK` kernel? Based on your benchmarks, even topk=1 shows performance improvements with the optimized approach.

sgl-kernel/csrc/moe/moe_topk_softmax_kernels.cu
thenumberouscode: Indeed, determining both the global maximum and second maximum requires comparing just three values.

Consider your example: We first compare c1.max with c2.max to obtain the global maximum. Next, we check whether c1.max equals this global maximum. If it does, then the global second maximum can only be either c1.secondMax or c2.max. Therefore, we simply need to compare c2.max with c1.secondMax to determine the global second maximum.

sgl-kernel/csrc/moe/moe_topk_softmax_kernels.cu
thenumberouscode: > Should we add boundary check ? `if (result_idx >= k) break;` ?

add a boundary check at line 234 `if (k_idx * 2 + i >= k) break;`

thenumberouscode: > Please add some test cases to `test_moe_topk_softmax.py`?
> 
> ```python
> @pytest.mark.parametrize(
>     "num_tokens, num_experts, topk",
>     [
>         (128, 512, 1),
>         (128, 512, 2),
>         (128, 512, 3),  # Odd topk
>         (128, 512, 5),  # Odd topk
>         (128, 512, 8),
>         (512, 512, 512),  # Edge case: k = num_experts
>     ]
> )
> ```

Yes, I have added test_topkfast_softmax in test_moe_topk_softmax.py. The odd topk cases have been accounted for, but the edge case where both k and num_experts equal 512 fails due to the following issue:

When multiple elements share identical values, the order in which their indices appear in the output of torch.topk is non-deterministic. For instance, given weights [1.0, 1.0, 0.8], the returned indices could be either [0, 1, 2] or [1, 0, 2]. As a result, our topKFast implementation cannot guarantee an index ordering that aligns exactly with torch.topk. This discrepancy causes the torch.allclose assertion to fail.

This behavior is documented in the official PyTorch documentation:
https://docs.pytorch.org/docs/stable/generated/torch.topk.html

sgl-kernel/csrc/moe/moe_topk_softmax_kernels.cu
thenumberouscode: My benchmarking results are based on the current branch code, confirming that the performance improvements of moeTopk stem from its reduced algorithmic complexity. The natural question then arises: why not completely replace moeTopk with moeTopkFast?

After local testing, I observed that moeTopkFast can actually be more time-consuming than moeTopk when k = 1. The reason is that moeTopkFast utilizes more registers than moeTopk, which may result in fewer thread blocks being scheduled per SM. When k = 1, both moeTopk and moeTopkFast need to access global memory only once, so moeTopkFast does not reduce global memory access latency in this case. Therefore, I have chosen to retain moeTopk for the k = 1 scenario.

However, for k > 1, the situation differs significantly. moeTopkFast reduces global memory access by at least one time compared to moeTopk, leading to better overall performance in these cases.

sgl-kernel/csrc/moe/moe_topk_softmax_kernels.cu
BBuf: Thank you for the detailed explanations! You're absolutely right about the `TopKPairArgMax` logic - I apologize for the confusion.

sgl-kernel/csrc/moe/moe_topk_softmax_kernels.cu
BBuf: Make sense.

sgl-kernel/csrc/moe/moe_topk_softmax_kernels.cu
BBuf: Your analysis about register pressure and occupancy is spot-on. This is a good example of the trade-off between computational complexity and hardware resource constraints.

BBuf (approved):
LGTM.

thenumberouscode:
@BBuf The workflow had four failed cases, I think two of them are irrelevant to my code, correct me if I am wrong:
1. [sgl-kernel-mla-test](https://github.com/sgl-project/sglang/actions/runs/19919624995/job/57323335711?pr=13969#logs) is unrelated.
2. [unit-test-backend-1-gpu (0)](https://github.com/sgl-project/sglang/actions/runs/19919624995/job/57323376963?pr=13969#logs) is unrelated.

thenumberouscode:
[unit-test-backend-1-gpu (10)](https://github.com/sgl-project/sglang/actions/runs/19919624995/job/57323376962?pr=13969#logs) 
This test case was particularly confusing, as it passes with moeTopK but fails with moeTopKFast.

Test logic overview:
- Call `select_experts` to get top-k indices and weights, With 24 experts, the test uses either moeTopK or moeTopKFast.
- Execute `fused_moe` Triton kernel for matrix operations
- Compare results with naive torch implementation (`torch_w8a8_block_int8_moe`), and the `torch_w8a8_block_int8_moe` using torch.topk to select the top-k indices and weights.

Failure explanation: 
The test fails due to inconsistent index selection between torch.topk and moeTopKFast when scores are equal. These indices are used to select elements from tensor `a` (which is unrelated to the score). Different indices—even with equal scores—can reference completely different values in `a`, leading to divergent computation results.

code snippet:
https://github.com/sgl-project/sglang/blob/4285e99da74c102a6647c5ca7e66bddbbdec728b/test/srt/quant/test_block_int8.py#L178C9-L198C1
image



Another key question arises: why did the previous moeTopK implementation pass tests initially?

The answer lies in the fixed random seed (torch.seed=0), which generated consistent test cases. In these deterministic cases, tensor `a` may did not contain dramatically different values across indices. However, when using random seeds, the test also fails, revealing the underlying issue.

Solution implemented:
I use topk_out from moeTopkFast in torch_w8a8_block_int8_moe to avoid interference from torch.topk. This approach makes sense since our test is designed to validate the `fused_moe`. Additionally, I have expanded the test cases in the unit tests for `moeTopK` and `moeTopkFast` to ensure their correctness.

PR #13978, opened 2025-11-26
Fix ZMQ binding and model loading for FastWan compatibility

yh0903:
## Motivation

This PR fixes two critical issues that prevent FastWan models (and potentially other diffusion models) from loading successfully in SGLang:

1. **ZMQ Binding Error**: When serving diffusion models, the scheduler fails to bind ZMQ sockets with the error "No such device (addr='tcp://localhost:5610')". This occurs because ZMQ cannot reliably bind to `localhost` on some systems, requiring `127.0.0.1` instead.

2. **Model Parameter Mismatch**: FastWan checkpoints contain parameters like `to_gate_compress` (used for sparse attention) that may not be present in the instantiated model when using standard attention backends. The current strict loading logic raises a `ValueError` when encountering these extra parameters, preventing model loading.

These fixes ensure consistency with how the text model server (srt) handles similar scenarios and align with the flexible loading patterns used elsewhere in the codebase.

Related issue: #13943

### 1. ZMQ Binding Fix (`server_args.py`)
- **`scheduler_endpoint()` method**: Modified to normalize `localhost` to `127.0.0.1` for ZMQ binding compatibility, while still respecting custom host configurations when provided.
- **`post_init_serve()` method**: Changed default host from `"localhost"` to `"127.0.0.1"` to match the text model server's default behavior.

### 2. Model Loading Flexibility (`fsdp_load.py`, `component_loader.py`)
- **`load_model_from_full_model_state_dict()`**: Added check for `strict` flag before raising `ValueError` for unexpected parameters. When `strict=False`, logs a warning and skips the parameter instead of failing.
- **`maybe_load_fsdp_model()`**: Added `strict` parameter (defaults to `True` for backward compatibility) and passes it to `load_model_from_full_model_state_dict()`.
- **`TransformerLoader.load()`**: Updated to use `strict=False` when calling `maybe_load_fsdp_model()`, allowing checkpoints with extra parameters to load successfully.

These changes maintain backward compatibility (default `strict=True` in `maybe_load_fsdp_model`) while providing flexibility for models with optional parameters.

## Accuracy Tests

Not applicable. These are infrastructure/bug fixes that do not affect model outputs or inference logic. The changes only affect:
- Network binding behavior (ZMQ endpoint)
- Model weight loading tolerance (allowing extra unused parameters)

Model accuracy remains unchanged as we're not modifying any forward pass logic or model architecture.

## Benchmarking and Profiling

Not applicable. These changes do not impact inference speed:
- ZMQ binding fix only affects startup/socket initialization
- Model loading change only affects which parameters are loaded (extra parameters are skipped, not processed)

No performance impact expected.

## Checklist

### DIFF
--- a/python/sglang/multimodal_gen/runtime/server_args.py
+++ b/python/sglang/multimodal_gen/runtime/server_args.py
@@ -577,10 +577,12 @@ class ServerArgs:
 
     def scheduler_endpoint(self):
         """
-        Internal endpoint for scheduler
-
+        Internal endpoint for scheduler.
+        Prefers the configured host but normalizes localhost -> 127.0.0.1 to avoid ZMQ issues.
         """
-        scheduler_host = self.host or "localhost"
+        scheduler_host = self.host
+        if scheduler_host is None or scheduler_host == "localhost":
+            scheduler_host = "127.0.0.1"
         return f"tcp://{scheduler_host}:{self.scheduler_port}"
 
     def settle_port(
@@ -628,7 +630,8 @@ class ServerArgs:
         Post init when in serve mode
         """
         if self.host is None:
-            self.host = "localhost"
+            # Use 127.0.0.1 for consistency with text model server (srt) default
+            self.host = "127.0.0.1"
         if self.port is None:
             self.port = 3000
         self.port = self.settle_port(self.port)
--- a/python/sglang/multimodal_gen/runtime/server_args.py
+++ b/python/sglang/multimodal_gen/runtime/server_args.py
@@ -15,7 +15,6 @@ import tempfile
 from contextlib import contextmanager
 from dataclasses import field
 from enum import Enum
-from functools import lru_cache
 from typing import Any, Optional
 
 from sglang.multimodal_gen.configs.pipeline_configs.base import PipelineConfig, STA_Mode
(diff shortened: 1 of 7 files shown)


python/sglang/multimodal_gen/runtime/server_args.py
```diff
@@ -654,10 +654,12 @@ def url(self):
...
         """
-        Internal endpoint for scheduler
-
+        Internal endpoint for scheduler.
+        Prefers the configured host but normalizes localhost -> 127.0.0.1 to avoid ZMQ issues.
         """
-        scheduler_host = self.host or "localhost"
+        scheduler_host = self.host
+        if scheduler_host is None or scheduler_host == "localhost":
+            scheduler_host = "127.0.0.1"
```
mickqian, line 589: do we still need this, when https://github.com/sgl-project/sglang/pull/13978/files#diff-ff1152576d3f8d498ab7a41cb23718180250504f3d94ad8100367950ece52304R710 is changed?

yh0903: Thanks for the review!

I originally kept the normalization logic here as a safety net, just in case `scheduler_endpoint` is accessed before `post_init_serve` runs (maybe I am wrong :)). In that scenario, defaulting to `"localhost"` would trigger the ZMQ bug, whereas this logic ensures we use `127.0.0.1`.

But also tested with undoing the change here, still solves ZMQ issue in my system. So Im happy with removing the duplication too.

PR #13991, opened 2025-11-26
Fix `get_load` API

hnyls2002:
This PR reverts #13203 and fixes various problems

- Use `seqlen` instead of `len(req.origin_input_ids)`.
- Add transfer queue and retracted queue for the decode server in disaggregation mode.
- Simplify code style, reuse the token calculation function from other components.


ShangmingCai (approved):
LGTM

PR #13996, opened 2025-11-26
[diffusion] cli: add argument --adjust-frames, --override-protected-fields

gmixiaojin:
Add 2 arguments:
1) --adjust-frames: Enable/disable adjusting num_frames to evenly split latent frames across GPUs 
2) --override-protected-fields: Allow user params to override fields defined in subclasses

Change the variable max_area in input_validation.py line 167 from hardcode to variables in max_areas from different configs in wan.py

### DIFF
--- a/python/sglang/multimodal_gen/runtime/entrypoints/utils.py
+++ b/python/sglang/multimodal_gen/runtime/entrypoints/utils.py
@@ -8,6 +8,12 @@ This module provides a consolidated interface for generating videos using
 diffusion models.
 """
 
+import logging
+import dataclasses
+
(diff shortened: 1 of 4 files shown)


python/sglang/multimodal_gen/runtime/entrypoints/utils.py
```diff
@@ -8,6 +8,12 @@
 diffusion models.
 """
 
+import logging
+import dataclasses
+
+# Suppress verbose logging from imageio, which is triggered when saving images.
+logging.getLogger("imageio").setLevel(logging.WARNING)
```
mickqian: please check `suppress_other_loggers`

PR #14006, opened 2025-11-26
Fix structural_tag tool call with null schema

AzazKamaz:
## Motivation

Fixes this request:
```json
{
  "model": "deepseek-v3-1-terminus",
  "tools": [
    {
      "function": {
        "name": "execute_sql_query",
        "parameters": null,
        "description": "Execute SQL query",
        "strict": true
      },
      "type": "function"
    }
  ],
  "messages": [
    {
      "role": "user",
      "content": "How many rows are in the my_table table?"
    }
  ]
}
```

## Modifications

Fixes incompatibility between OpenAI schema and internal structural_tag schema (at least with xgrammar backend which is default)
- `parameters: null` is allowed in request: https://github.com/sgl-project/sglang/blob/v0.5.5/python/sglang/srt/entrypoints/openai/protocol.py#L132
- `schema: null` is not allowed in xgrammar: https://github.com/mlc-ai/xgrammar/blob/v0.1.27/python/xgrammar/structural_tag.py#L259

For sure it can be serialized to `"null"` and passed as string but it throws an error and crashes server

## Checklist

### DIFF
--- a/python/sglang/srt/function_call/function_call_parser.py
+++ b/python/sglang/srt/function_call/function_call_parser.py
@@ -152,7 +152,7 @@ class FunctionCallParser:
             tool_structures.append(
                 StructuresResponseFormat(
                     begin=info.begin,
-                    schema=schema,  # type: ignore
+                    schema=schema or {},  # type: ignore
                     end=info.end,
                 )
             )


AzazKamaz:
The questions are:
1. Should we even allow such request? Probably yes since openai allows it (at least they allow parameters: undefined which will be None in python I guess): https://github.com/openai/openai-openapi/blob/2.0.0/openapi.yaml#L1911
2. Maybe fix it for xgrammar and not for all backends? Probably no?...

JustinTong0323 (approved):
LGTM

PR #14032, opened 2025-11-27
Add overlap scheduling for embeddings code path

satyamk7054:
## Motivation
Add overlap scheduling for embeddings code path to improve throughput

## Modifications
Update event_loop_overlap to handle overlap for embeddings using similar semantics as generation model
The feature is handled behind --enable-non-generation-schedule-overlap server arg

## Accuracy Tests
Added a unit test to validate accuracy compared to sentence transformers

Used `assert_close_prefill_logits` to run similarity test on `Qwen3-Embedding-0.6B` and DEFAULT_PROMPTS * 500
which compares similarity against sentence-transformers.
Short-strings is defined as `len(prompts[i]) <= 1000`
(see [commit](https://github.com/sgl-project/sglang/commit/b8ee0f4c92e1095112d5d74e189e512e64663c88))
```
Average similarity for short strings: 0.9999973177909851
Average similarity for long strings: 0.9996759295463562
Overall average similarity: 0.9999329447746277
```

```
if len(prompts[i]) <= 1000:
    assert torch.all(
        abs(similarity - 1) < 1e-5
    ), "embeddings are not all close"
    short_string_similarities.append(similarity)
else:
    assert torch.all(
        abs(similarity - 1) < 1e-2
    ), "embeddings are not all close"
    long_string_similarities.append(similarity)
```

## Benchmarking and Profiling
Improve achieved RPS by 5-10% (impact dependent on input tokens)

Below is for 6k input tokens
### With overlap
```
python -m sglang.launch_server --model-path Qwen3-Embedding-0.6B --host 0.0.0.0 --is-embedding --tensor-parallel-size 1 --data-parallel-size 1 --dtype float16 --disable-radix-cache --chunked-prefill-size -1 --max-prefill-tokens 32768 --enable-non-generation-schedule-overlap
```
```
Overall Summary for RPS 46, Duration 60s, Item Count 1:
  Test duration:         60 seconds
  Server type:           HTTP
  HTTP mode:             EMBEDDINGS
  Target RPS:            46
  Achieved RPS:          41.21
  Item count:            1
  Distribution:          CONSTANT
  Unique requests generated: 100
  Total requests sent:   2760
  Successful requests:   2760
  Failed requests:       0
  Time to send all requests: 62.03 seconds
  Time for all requests to complete: 66.97 seconds
```

### Without overlap
```
python -m sglang.launch_server --model-path Qwen3-Embedding-0.6B --host 0.0.0.0 --is-embedding --tensor-parallel-size 1 --data-parallel-size 1 --dtype float16 --disable-radix-cache --chunked-prefill-size -1 --max-prefill-tokens 32768 
```
```
Overall Summary for RPS 46, Duration 60s, Item Count 1:
  Test duration:         60 seconds
  Server type:           HTTP
  HTTP mode:             EMBEDDINGS
  Target RPS:            46
  Achieved RPS:          39.40
  Item count:            1
  Distribution:          CONSTANT
  Unique requests generated: 100
  Total requests sent:   2760
  Successful requests:   2760
  Failed requests:       0
  Time to send all requests: 62.21 seconds
  Time for all requests to complete: 70.05 seconds
```

Also ran long running (12 hour) benchmark twice for server stability
```
python -m sglang.launch_server --model-path /home/jobuser/models/Qwen3-Embedding-0.6B --host 0.0.0.0 --is-embedding --tensor-parallel-size 1 --data-parallel-size 1 --dtype float16 --max-prefill-tokens 32768 --disable-radix-cache 2>&1 | tee server_logs.txt
```

3-4 in requests out of 1728000 requests had "server disconnected" error but that seems to be a side-effect of sessions.post call (apparently it has connection reuse), the server stays healthy (no error logs)

Achieved RPS is lower than the above perf benchmark because without disabling chunked prefill, the max prefill batch size was 8k (default settings)

```
Overall Summary for RPS 40, Duration 43200s, Item Count 1:
  Test duration:         43200 seconds
  Server type:           HTTP
  HTTP mode:             EMBEDDINGS
  Target RPS:            40
  Achieved RPS:          38.68
  Item count:            1
  Distribution:          POISSON
  Unique requests generated: 100
  Total requests sent:   1728000
  Successful requests:   1727996
  Failed requests:       4
  Time to send all requests: 44648.63 seconds
  Time for all requests to complete: 44671.50 seconds
  Average response time: 558.43 ms
  P50 response time:     411.47 ms
  P90 response time:     1186.11 ms
  P99 response time:     2424.78 ms
```

## Checklist


python/sglang/srt/server_args.py
```diff
@@ -503,6 +503,7 @@ class ServerArgs:
     enable_mscclpp: bool = False
     enable_torch_symm_mem: bool = False
     disable_overlap_schedule: bool = False
+    enable_embedding_schedule_overlap: bool = False
```
sundar24295s: The non-generation path also contains SequenceClassification, cross-encode models etc... can we come up with a different name, probably `--enable-non-generation-schedule-overlap`  etc..?

python/sglang/srt/server_args.py
```diff
@@ -2088,6 +2089,19 @@ def _handle_other_validations(self):
...
             self.skip_server_warmup = True
 
+        if self.enable_embedding_schedule_overlap and self.disable_overlap_schedule:
+            raise ValueError(
+                "Cannot enable both --enable-embedding-schedule-overlap and --disable-overlap-schedule."
+            )
+
+        if (
+            self.enable_embedding_schedule_overlap
+            and envs.SGLANG_EMBEDDINGS_SPARSE_HEAD.is_set()
```
sundar24295s: Any reason we don't support sparse tensors?

python/sglang/srt/managers/scheduler.py
```diff
@@ -206,6 +206,25 @@
 @dataclass
 class EmbeddingBatchResult:
     embeddings: torch.Tensor
+    copy_done: Optional[torch.cuda.Event] = None
+
+    def copy_to_cpu(self):
+        """Copy embeddings tensor to CPU in overlap scheduling."""
+
+        if isinstance(self.embeddings, torch.Tensor):
```
sundar24295s, line 218: Do we need to early return here is self.embeddings.device is already on cpu?

satyamk7054: .to("cpu") is a no-op on cpu tensors

python/sglang/srt/server_args.py
satyamk7054: We can add that as follow up (different PR), if required. I wasn't sure on the full impl of it

python/sglang/srt/server_args.py
satyamk7054: updated

sundar24295s (approved):
LGTM! Thanks for the PR!

python/sglang/srt/server_args.py
```diff
@@ -542,6 +542,7 @@ class ServerArgs:
     enable_mscclpp: bool = False
     enable_torch_symm_mem: bool = False
     disable_overlap_schedule: bool = False
+    enable_non_generation_schedule_overlap: bool = False
```
merrymercy: This argument is very confusing. Can we avoid it?
i.e., enable overlap scheduler for all cases by default

merrymercy (changes requested):

python/sglang/srt/server_args.py
satyamk7054: sounds good, updated

merrymercy (approved):

PR #14046, opened 2025-11-27
feat: Support return logprob for eagle v2

Denji-kk:
## Motivation
The current eagle v2 process does not support the `return_logprob` parameter of sglang `/generate` api. For example, when using a request below:
```python
import asyncio
import aiohttp
url = f"http://{address}/generate"
messages = [
    {
        "role": "user",
        "content": "What is the capital of France?",
    }
]
prompt_ids = tokenizer.apply_chat_template(
    messages,
    tokenize=True,
    add_generation_prompt=True,
)
data = {
    "input_ids": prompt_ids,
    "sampling_params": {
        "temperature": 1.0,
        "top_p": 1.0,
        "max_new_tokens": 16,
    },
    "return_logprob": True,
}
connector = aiohttp.TCPConnector(limit=1000)
async with aiohttp.ClientSession(connector=connector) as session:
    async with session.post(url, json=data) as resp:
        result = await resp.json()
        return result
```

In the returned result, under `meta_info/output_token_logprobs`, only the first token’s id and logprob are present, because they come from the target worker’s prefill.
```json
{
    "text": "The user asks: \"What is the capital of France?\" It's a simple factual",
    "output_ids": [
        758,
        3100,
        20886,
        58,
        494,
        3376,
        355,
        275,
        5505,
        300,
        5969,
        8109,
        6792,
        258,
        4160,
        50965
    ],
    "meta_info": {
        "output_token_logprobs": [
            [
                -1.5598505735397339,
                758,
                null
            ]
        ]
    }
}
```

This pull request adds support for the return_logprobs parameter in the EAGLE worker v2 pipeline.
## Modifications

1. Added the `_add_eagle_v2_logprobs` method in `scheduler_output_processor_mixin.py`, and invoked it in `process_batch_result_decode` to compute and extract the logprobs of the output tokens.
2. In the `stream_output_generation` method, truncate the output logprobs and token IDs. This is because when finish_reason is FINISH_LENGTH, we need to cut off the extra tokens and logprobs produced during the verify phase that exceed the requested length.

## Checklist


fzyzcjy (changes requested):
hi, could you please add a simple test to check logprob outputs b/t enable-spec and disable-spec? with this we are sure there is no offset error etc

Denji-kk:
@zhaochenyang20 @fzyzcjy 
I've added tests to ensure that text outputs and logprobs offsets match when turning on/off speculative decoding + Eagle v2. The logprobs values are roughly consistent, I think precision mismatches are acceptable due to potential differences in the low-level kernels used for the 'draft + verify' and 'decode' process and their calculation pipeline differ.

PR #14051, opened 2025-11-27
EVS Framework: Support NemotronH_Nano_VL_V2

netanel-haber:
## Motivation

> https://arxiv.org/abs/2510.14624: Efficient Video Sampling: Pruning Temporally Redundant Tokens for Faster VLM Inference

Add support for EVS pruning to `NemotronH_Nano_VL_V2`. See `python/sglang/srt/multimodal/evs/README.md` for more details.

A pruning rate of 0.7, for example, means that 30% of the vision encoder output embeddings will be *retained*, and added to the language model prompt.

Nemotron-Nano-VL-V2.0 was trained with a pruning rate of `0.7`, which is the default value in the model config, so EVS will be on by default. To disable or otherwise configure the pruning rate (accuracy may vary), set e.g.: `--json-model-override-args '{"video_pruning_rate": 0.0}'` [EVS off].

Example of full serving command with disabled EVS:

```bash
sglang serve --model-path nvidia/Nemotron-Nano-12B-v2-VL-BF16 --trust-remote-code --json-model-override-args '{"video_pruning_rate": 0.0}'
```

The PR creates a framework that should make it easier to add EVS support to additional models in the future, such as Qwen2.5VL, which will be implemented in a following pr. 

## Benchmarks

NanoV2VLM on SGLang and VLLM, with `video_pruning_rate=0.9` got an identical overall VideoMME score of `0.644`.

| Scenario\Metric                	        | Online TTFT (Seconds) stderr: ±0.38 	| VideoMME Accuracy       	|
|---------------------------------	        |-------------------------------------	                |-------------------------	                |
| EVS Off [q=0.0]                 	        | 11.96 [100%]                        	                | Between 0.665 and 0.668 	|
| EVS Off [q=0.4]                 	        | 09.97 [ 83%]                        	                |                         	                |
| EVS On  [q=0.7] (default value) 	| 08.79 [ 73%]                        	                |                         	                |
| EVS Off [q=0.9]                 	        | 08.39 [ 70%]                        	                | 0.644                   	                |

## Checklist


yuan-luo:
@netanel-haber Could you provide the TTFT benchmark diagram for this feature on different pruning rates? So as the user can have better decision in the balance of performance and accuracy.

yuan-luo (approved):

test/srt/test_evs.py
JustinTong0323, line 1: This test only includes the config check. Should we add additional tests for EVS's functionality?

netanel-haber: That's a good idea, but can I leave the test for the next upcoming evs pr I have for Qwen-vl evs? This pr has been open for quite a while, and I've already validated the EVS end-to-end [results in description]

python/sglang/srt/mem_cache/multimodal_cache.py
```diff
@@ -94,17 +100,20 @@ def get(
         return embedding
 
     def set(
-        self, mm_hash: int, embedding: torch.Tensor, loc: Optional[torch.Tensor] = None
+        self,
+        mm_hash: int,
+        embedding: EmbeddingResult,
```
mickqian, line 105: we should probably let `MultiModalStaticCache` focus on tensor here, what do you think?

netanel-haber: The reason I did this is because I need to tokens_per_frame that is created along with the embedding in the case of EVS - so if the embedding is cached, I need tokens per frame cached as well. The way I implemented this change, the cache is agnostic to whatever else I save on an entry, so it needn't be aware of EVS etc.

mickqian (approved):

PR #14074, opened 2025-11-28
[CPU] layernorm & fused add-layernorm kernels

ZailiWang:
To add SIMD based `layernorm` and `fused-add-layernorm` kernels.

Test:
Added `TestLayerNorm` cases in `test/srt/cpu/test_norm.py`: https://github.com/sgl-project/sglang/pull/14074/files#diff-7c321477a511d53c144e81d839e6086d0962cb96f7ddd91942f87d753da81f28R134


ZailiWang:
Only pure `layernorm` interface is exposed at framework frontend, so `fused-add-ln` kernel is preserved and not used now.

test/srt/cpu/test_norm.py
```diff
@@ -131,5 +131,67 @@ def test_norm(self):
...
+            ref_x_add_ln, weight, variance_epsilon, ref_residual
+        )
+
+        torch.testing.assert_close(x_add_ln, ref_add_ln_out, atol=atol, rtol=rtol)
+        torch.testing.assert_close(residual, ref_residual, atol=atol, rtol=rtol)
+
+    def test_norm(self) -> None:
+        for params in itertools.product(self.M, self.N, self.dtype):
+            with self.subTest(m=params[0], n=params[1], dtype=params[2]):
+                self._norm_test(*params)
```
mingfeima: you can use `parametrize` from `utils.py`, take `test_causal_conv1d.py` for example.

sgl-kernel/csrc/cpu/norm.cpp
```diff
@@ -302,6 +302,225 @@ void fused_rmsnorm_gated_kernel_impl(
...
+template <typename scalar_t>
+void fused_add_layernorm_kernel_impl(
+    scalar_t* __restrict__ input,
+    scalar_t* __restrict__ residual,
+    const scalar_t* __restrict__ weight,
+    float* __restrict__ buffer,
+    int64_t batch_size,
+    int64_t hidden_size,
+    int64_t input_strideN,
+    float eps = 1e-5) {
```
mingfeima, line 314: i would prefer to treat `layernorm_kernel_impl` as a special case for `fused_add_layernorm_kernel_impl` here to reduce duplicated code.

which means we use the same kernel of `fused_add_layernorm_kernel_impl` and two APIs from `layernorm_cpu` and `fused_add_layernorm_cpu`

mingfeima (changes requested):

mingfeima:
@ZailiWang check fused_add_layernorm from cuda, whether it uses the 3-path approach or welford.

test/srt/cpu/test_norm.py
ZailiWang: updated per comment

sgl-kernel/csrc/cpu/norm.cpp
ZailiWang: updated per comment

PR #14093, opened 2025-11-28
Add fused FP8 KV cache write kernel for TRTLLM MHA backend

harvenstar:
## Summary

This PR introduces a fused FP8 KV cache write kernel for the TRTLLM MHA backend.  
It combines FP8 quantization and paged KV cache writes into a single Triton kernel (plus a naive fallback) to reduce kernel launch overhead and avoid intermediate FP8 tensors.

---

## Motivation

The current FP8 KV cache path launches multiple small elementwise kernels per step:

- FP8 quantization of K  
- FP8 quantization of V  
- Writing K to the paged KV cache  
- Writing V to the paged KV cache  

These kernels all operate on the same data and are memory bound.  
The extra launches and intermediate tensors add overhead without bringing additional parallelism.

This PR fuses FP8 quantization and KV cache writes into one kernel for the TRTLLM backend.  
RoPE and attention logic remain unchanged, but the extra FP8 elementwise kernels in the KV write path are removed.

---

### New fused FP8 KV kernel

**File added**

- `python/sglang/srt/layers/attention/trtllm_fp8_kv_kernel.py`

**Triton kernel: `_fused_fp8_set_kv_buffer_kernel`**

- Takes post RoPE K and V in BF16 or FP16.  
- Applies FP8 quantization using provided scalar scales.  
- Writes directly into the paged FP8 KV cache layout.  
- Respects the existing `uint8` backed FP8 storage convention.

**Python wrapper: `fused_fp8_set_kv_buffer`**

- Supports both:
  - 3D cache layout: `[total_slots, num_kv_heads, head_dim]`
  - 4D cache layout: `[num_pages, page_size, num_kv_heads, head_dim]`
- Handles both 2D and 3D K, V inputs via runtime shape checks.  
- Provides a naive fallback that exactly reproduces `MHATokenToKVPool.set_kv_buffer` semantics.  
- Naive path is used as the golden implementation in unit tests.

### Integration into TRTLLM attention backend

- Wires `fused_fp8_set_kv_buffer` into:
  - `TRTLLMHAAttnBackend.forward_decode`
  - `TRTLLMHAAttnBackend.forward_extend`
  when `kv_dtype` is FP8.
- Adds a feature flag style switch so non FP8 configurations are unaffected.  
- Keeps the original code path as a correctness reference.  
- Removes or downgrades all development-only debug logging so runtime logs remain minimal.

---

## Accuracy and Testing

The fused path is designed to be numerically equivalent to the existing FP8 KV cache behavior.

### Unit tests

1. **Naive fused implementation vs `MHATokenToKVPool.set_kv_buffer`**
   - Bit exact match at the written KV cache locations for both K and V.

2. **Triton implementation vs naive implementation**
   - Matches the naive path when using the same scalar FP8 scales.  
   - Uses the same quantization formula as the reference path.

### End to end evaluation (GSM8K)

- Setup:
  - Qwen3 MoE, FP8 KV cache, TRTLLM backend.
  - `bench_sglang.py`, 8 shot, 1319 questions.
- Results:
  - Baseline accuracy: ~0.956  
  - Fused FP8 KV cache accuracy: ~0.958  
  - Invalid rate: 0.0  

No regression on GSM8K accuracy was observed.

---

## Benchmarking and Profiling

- Nsight profiling shows that the four small FP8 KV elementwise kernels in the TRTLLM path are replaced by a single fused kernel in `forward_decode` and `forward_extend` when FP8 KV cache is enabled.  
- KV write latency is reduced and the number of kernel launches per decode step is lower.  
- Overall GSM8K benchmark latency remains in the same range, with KV cache writes no longer dominating the FP8 KV path.

Exact speedups depend on GPU, model, batch size, and sequence length.  
The main benefit of this PR is to remove unnecessary elementwise launches in the TRTLLM FP8 KV write path while keeping behavior identical.


python/sglang/srt/models/qwen3_moe.py
```diff
@@ -449,7 +504,15 @@ def forward_prepare_native(
                 else None
             ),
         )
-        inner_state = q, k, v, forward_batch
+
+        # Prepare extra kwargs for backend (e.g., TRTLLM MHA fusion hints)
+        attn_extra_kwargs = {}
+
+        # Enable TRTLLM FP8 fusion if conditions are met
+        if _should_enable_trtllm_fp8_fuse(forward_batch):
```
Qiaolin-Yu: Why do we need to check it here?

python/sglang/srt/layers/attention/trtllm_mha_backend.py
```diff
@@ -411,6 +415,44 @@ def get_cuda_graph_seq_len_fill_value(self) -> int:
         """Get the fill value for sequence lengths in CUDA graph."""
         return 1
 
+    def _should_use_fused_fp8_path(
```
Qiaolin-Yu: So it's only enabled for Qwen? Is it possible to make it more general so that other models can use it?

python/sglang/srt/models/qwen3_moe.py
harvenstar: Good catch.

python/sglang/srt/layers/attention/trtllm_mha_backend.py
harvenstar: The check should be in the backend, not model-specific code. I'll move the fusion detection
   logic entirely into TRTLLMHAAttnBackend._should_use_fused_fp8_path() and remove all the qwen3_moe
  changes. This way any model using TRTLLM backend with FP8 KV cache gets the optimization automatically. Does the change sound good to you?

Qiaolin-Yu: sg

harvenstar: Done.✅ Fusion logic now lives in backend, and work for all models.

python/sglang/srt/layers/attention/trtllm_fp8_kv_kernel.py
```diff
@@ -0,0 +1,455 @@
...
+    input_stride_dim: tl.constexpr,
+    cache_stride_page: tl.constexpr,
+    cache_stride_offset: tl.constexpr,
+    cache_stride_head: tl.constexpr,
+    cache_stride_dim: tl.constexpr,
+    BLOCK_HEAD: tl.constexpr,
+    BLOCK_DIM: tl.constexpr,
+):
+    """Process a single K or V tensor: load, quantize to FP8, and write to cache."""
+    for head_idx in range(0, num_kv_heads, BLOCK_HEAD):
```
ispobock: Why don't parallelize heads? (i.e. add head to grid)

python/sglang/srt/layers/attention/trtllm_fp8_kv_kernel.py
```diff
@@ -0,0 +1,455 @@
...
+    # Compute page_id and offset within page
+    page_id = cache_loc // page_size
+    page_offset = cache_loc % page_size
+
+    # Compute inverse scales
+    k_inv_scale = 1.0 / k_scale if use_provided_scale else 1.0
+    v_inv_scale = 1.0 / v_scale if use_provided_scale else 1.0
+
+    # Process K tensor
+    _process_kv_tensor(
```
ispobock: Can we also parallelize k and v?

python/sglang/srt/layers/attention/trtllm_fp8_kv_kernel.py
```diff
@@ -0,0 +1,455 @@
...
+        v_stride_dim = v_3d.stride(2)
+
+        # Block sizes for tiling (tunable)
+        BLOCK_HEAD = min(num_kv_heads, 8)  # Process up to 8 heads at once
+        BLOCK_DIM = min(head_dim, 128)      # Process up to 128 dims at once
+
+        # Grid: one program per token
+        grid = (num_tokens,)
+
+        if not hasattr(fused_fp8_set_kv_buffer, "_triton_logged"):
```
ispobock: It seems not needed?

python/sglang/srt/layers/attention/trtllm_fp8_kv_kernel.py
```diff
@@ -0,0 +1,455 @@
...
+            v_cache_stride_page,
+            v_cache_stride_offset,
+            v_cache_stride_head,
+            v_cache_stride_dim,
+            BLOCK_HEAD=BLOCK_HEAD,
+            BLOCK_DIM=BLOCK_DIM,
+        )
+    else:
+        # Fallback to naive implementation
+        if not hasattr(fused_fp8_set_kv_buffer, "_naive_logged"):
```
ispobock: It seems not needed?

python/sglang/srt/layers/attention/trtllm_fp8_kv_kernel.py
```diff
@@ -0,0 +1,455 @@
...
+        v_cache_stride_page,
+        v_cache_stride_offset,
+        v_cache_stride_head,
+        v_cache_stride_dim,
+        BLOCK_HEAD,
+        BLOCK_DIM,
+    )
+
+
+def fused_fp8_set_kv_buffer(
```
ispobock, line 198: Could you add unit test for this kernel?

ispobock (commented):
1. share profile figure comparison to see the improvement after fusion
2. share end to end throughput improvement on Qwen3-235B

qq: Why trtllm_mla don't need this fusion?

python/sglang/srt/layers/attention/trtllm_fp8_kv_kernel.py
harvenstar: Done. Moved head iteration to grid dimension via head_block_id = tl.program_id(1).

python/sglang/srt/layers/attention/trtllm_fp8_kv_kernel.py
harvenstar: Done. Grid is now (tokens, head_blocks, 2) where kv_idx = tl.program_id(2) selects K or V.

python/sglang/srt/layers/attention/trtllm_fp8_kv_kernel.py
harvenstar: Removed in latest commit.

python/sglang/srt/layers/attention/trtllm_fp8_kv_kernel.py
harvenstar: Removed in latest commit.

python/sglang/srt/layers/attention/trtllm_fp8_kv_kernel.py
harvenstar: Added test/srt/test_trtllm_fp8_kv_kernel.py with serveral test cases covering
  different input/cache dimensions, scaling options, and edge cases. All
  passing.

python/sglang/srt/layers/attention/trtllm_mha_backend.py
```diff
@@ -585,10 +648,33 @@ def forward_extend(
...
         cache_loc = forward_batch.out_cache_loc
-        if save_kv_cache and k is not None:
-            forward_batch.token_to_kv_pool.set_kv_buffer(
-                layer, cache_loc, k, v, layer.k_scale, layer.v_scale
+
+        use_fused_fp8_path = self._should_use_fused_fp8_path(save_kv_cache, k)
+
+        if use_fused_fp8_path:
+            # Use fused FP8 quantization + KV cache write path
+            if not hasattr(self, "_fused_extend_path_logged"):
```
ispobock: Can we also remove this debug log?

test/srt/test_trtllm_fp8_kv_kernel.py
```diff
@@ -0,0 +1,261 @@
+"""
+Unit tests for TRTLLM FP8 KV cache fusion kernel.
```
ispobock, line 2: Can we move it to https://github.com/sgl-project/sglang/tree/main/test/manual?

python/sglang/srt/layers/attention/trtllm_mha_backend.py
```diff
@@ -524,10 +564,33 @@ def forward_decode(
...
         cache_loc = forward_batch.out_cache_loc
-        if save_kv_cache and k is not None:
-            forward_batch.token_to_kv_pool.set_kv_buffer(
-                layer, cache_loc, k, v, layer.k_scale, layer.v_scale
+
+        use_fused_fp8_path = self._should_use_fused_fp8_path(save_kv_cache, k)
+
+        if use_fused_fp8_path:
+            # Use fused FP8 quantization + KV cache write path
+            if not hasattr(self, "_fused_decode_path_logged"):
```
ispobock: Can we also remove this debug log?

test/srt/test_trtllm_fp8_kv_kernel.py
harvenstar: Done ✅

python/sglang/srt/layers/attention/trtllm_mha_backend.py
harvenstar: Done ✅

python/sglang/srt/layers/attention/trtllm_mha_backend.py
harvenstar: Done in commit [139b961](https://github.com/sgl-project/sglang/commit/139b96112cd8715aa5c75345f9a85fae9cd820dd). Removed both debug logs and moved the test to test/manual/ as suggested. Thanks!!

PR #14097, opened 2025-11-28
support GLM-V vision model dp 

zRzRzRzRzRzRzR:
Based on https://github.com/sgl-project/sglang/pull/13126 and https://github.com/sgl-project/sglang/pull/13724, add support for the GLM-V vision model DP.


python/sglang/srt/models/glm4v.py
```diff
@@ -542,15 +575,19 @@ def __init__(
             prefix=add_prefix("model", prefix),
         )
 
-        if config.tie_word_embeddings:
-            self.lm_head = self.model.embed_tokens
+        if self.pp_group.is_last_rank:
```
yuan-luo, line 578: This logic is related with PP. What's the purpose of this PR, do you want to support both DP and PP for GLM-V?

zRzRzRzRzRzRzR: Yes, in this PR, the logic for both PP and VIT DP has been added.

yhyang201 (approved):

zRzRzRzRzRzRzR:
For the MMMU benchmark, I used the GLM-4.6V model for testing, which aligns with internal metrics. The data is temporarily not convenient to disclose.

PR #14102, opened 2025-11-28
create weight_scale_2, input_scale, weight matrix by torch.zeros

qiyuxinlin:
## Issue Description

When testing NVIDIA's Qwen2.5-7B-VL-NVFP4 and a self-quantized GLM4.1-VL-NVFP4 model, I found that Qwen2.5-7B-VL-NVFP4 works correctly while GLM4.1-VL-NVFP4 produces gibberish outputs. The root cause was traced to incorrect `model.layers.0.mlp.gate_up_proj.input_scale` weights.

### Architecture Difference

**Qwen2.5**: `gate_proj` and `up_proj` are stored separately in the checkpoint with two independent `input_scale` parameters:
- `model.layers.X.mlp.gate_proj.input_scale`
- `model.layers.X.mlp.up_proj.input_scale`

**GLM4.1**: `gate_proj` and `up_proj` are fused together (`gate_up_proj`) with only one `input_scale` parameter in the checkpoint:
- `model.layers.X.mlp.gate_up_proj.input_scale`

### The Bug

In the current implementation, `PerTensorScaleParameter` is created as an array of length 2 (corresponding to gate and up):
```python
# modelopt_quant.py:972-975
input_scale = PerTensorScaleParameter(
    data=torch.empty(len(output_partition_sizes), dtype=torch.float32),  # len=2
    weight_loader=weight_loader,
)
```

During weight loading:
- **Qwen2.5**: Two parameters from checkpoint fill `shard_id=0` and `shard_id=1`, so the array is fully populated
- **GLM4.1**: Only one fused parameter from checkpoint fills `shard_id=0`, leaving `shard_id=1` uninitialized

When taking `max()` in `process_weights_after_loading`:
```python
input_scale_2 = layer.input_scale.max().to(torch.float32)
```

**Problem**: Since `torch.empty()` is used for initialization, the unfilled `shard_id=1` position contains random garbage data, causing `max()` to potentially return an incorrect random value instead of the correct scale value.

### Example
```python
# GLM4.1 loading flow
input_scale = torch.empty(2)       # [random=999.9, random=123.4]
input_scale[0] = 0.123              # [0.123, 123.4]  ← shard_id=1 still random
input_scale.max()                   # returns 123.4 ❌ instead of 0.123
```

This leads to incorrect quantization parameters and garbled model outputs.

### Why Qwen2.5 Works

Qwen2.5's two shards are both correctly loaded, fully populating the array, so even with `torch.empty()` initialization, the issue doesn't manifest.

## Fix

Replace `torch.empty()` with `torch.zeros()`:
```python
input_scale = PerTensorScaleParameter(
    data=torch.zeros(len(output_partition_sizes), dtype=torch.float32),
    weight_loader=weight_loader,
)

weight_scale_2 = PerTensorScaleParameter(
    data=torch.zeros(len(output_partition_sizes), dtype=torch.float32),
    weight_loader=weight_loader,
)
```

This ensures that even if some shards are not loaded, we won't get random values.

## Impact Scope

- `ModelOptFp4LinearMethod`: `input_scale` and `weight_scale_2` (already using `torch.zeros` in the code)
- `ModelOptNvFp4FusedMoEMethod`: `w13_input_scale`, `w2_input_scale`, `w13_weight_scale_2`, `w2_weight_scale_2` (still using `torch.empty`)

The same issue may also affect MoE models using fused layers.

## Testing

- ✅ Qwen2.5-7B-VL-NVFP4 (separate gate/up): works correctly
- ✅ GLM4.1-VL-NVFP4 (fused gate_up): works correctly after fix
- Recommend testing other models using NVFP4 quantization with fused layers
## Checklist


hnyls2002:
Hi @qiyuxinlin, thanks for the analysis of the NVFP4 scale initialization issue. As it stands this PR mixes the quant fix with unrelated files (full_reuse.py and test_deepseekv3_chat_template.py at the repo root, plus chat-template changes), which makes it unreviewable as one unit. Closing for now - if the torch.zeros fix is still needed, please open a clean PR containing only the modelopt_quant.py change, or reopen if I've missed something.

Issue #11894, opened 2025-10-21
[Feature] Comprehensive LoRA Adapter Support for MOE Models


ConnorLi96:
### Checklist

- [ ] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 2. Please use English, otherwise it will be closed.

### Background
from https://github.com/sgl-project/sglang/issues/9897

### Current Architecture
**LoRA (Existing):** Supports Attention + MLP layers with 2D/3D weight tensors  
**MoE (Existing):** 3-stage architecture (Dispatch → Compute → Combine) with multi-backend support  
**Gap:** No LoRA support for MoE layers

#### MoE LoRA System Flow

```
[input_hidden_states]
          |
          v
    ┌─────────────────────────────────────┐
    │ Attention Layer (Existing)          │
    │   base_layer.forward                │
    │          |                           │
    │          v                           │
    │   [base_output]                     │
    │          |                           │
    │          v                           │
    │   lora_backend.run_qkv_lora       │
    │   (2D weights: [loras, rank, ...])  │
    │          |                           │
    └──────────|──────────────────────────┘
               v
    [attn_output]
               |
               v
    ┌──────────────────────────────────────────────────────────┐
    │ MoE Layer                                                 │
    │                                                           │
    │   Router.forward -> topk_selection                       │
    │          |                                                │
    │          v                                                │
    │   [TopKOutput: topk_weights, topk_ids]                   │
    │          |                                                │
    │          |------------------+                             │
    │          |                  |                             │
    │          v                  v                             │
    │   ┌──────────────┐   ┌──────────────────┐               │
    │   │ Base Path ✅ │   │ LoRA Path ⭐ NEW │               │
    │   │ (Existing)   │   │                  │               │
    │   └──────────────┘   └──────────────────┘               │
    │          |                  |                             │
    │          v                  v                             │
    │   Dispatcher.dispatch   per_lora_dispatch (CUDA)         │
    │          |               (sort by lora_id, expert_id)    │
    │          |                  |                             │
    │          v                  v                             │
    │   MoeRunner.forward    per_expert_lora_kernel            │
    │   (DeepGEMM/FlashInfer) (3D Grid: spatial × slices × loras)
    │          |               (4D weights: [loras,experts,...])│
    │          |                  |                             │
    │          v                  v                             │
    │   Dispatcher.combine   [lora_delta]                      │
[...]

PR #14105, opened 2025-11-28
[LoRA][III] Add LoRA support for MoE layers and enable TP

jonahbernard:
## Motivation

This PR adds support for LoRA serving on the expert layers for Mixture-of-Expert models. This is the initial development PR, which was split into two other smaller PRs, for maintenance reasons: https://github.com/sgl-project/sglang/pull/19710 and https://github.com/sgl-project/sglang/pull/19711

## Modifications

1) Adds `FusedMoEWithLoRA`, which is a wrapper around `FusedMoE` and enables LoRA on `FusedMoE`. `FusedMoEWithLoRA` is necessary for intercepting the forward pass that is usually called on `FusedMoE` to compute the expert layers.
2) Adds `moe_lora_align_block_size` CUDA kernel to efficiently sort tokens by LoRA adapter, then expert ID.
3) Adds `_fused_moe_lora_kernel` for efficient computation of LoRA A and LoRA B computations for gate_up_proj and down_proj for the expert layers. `_fused_moe_lora_kernel` operates on a 3D grid: `(split-k x M-blocks x N-blocks, len(lora_a_stack), max # of LoRA adapters)`.
4) Adds `TritonRunnerCoreWithLoRA`, which is a wrapper around `TritonRunnerCore`. The only difference between `TritonRunnerCoreWithLoRA` and `TritonRunnerCore` is that `TritonRunnerCoreWithLoRA` inserts the LoRA computations before the activation (after the base path's first GEMM) and before the final result is returned. 

### Testing:
- Adds 5 test cases: `test_moe_lora_align_block_size.py`, `test_fused_moe_lora_kernel.py`, `test_lora_moe_runner.py`, `test_lora_moe_vllm_sgl_logprob_diff.py`, and an additional MoE-specific test case to `test_lora_hf_sgl_logprob_diff.py`.
- `test_moe_lora_align_block_size.py`: tests the correctness of the token sorting kernel
- `test_fused_moe_lora_kernel.py`: tests the mathematical correctness of the Triton kernel that performs `BAx`
- `test_lora_moe_runner.py`: tests the mathematical correctness of one forward pass of `TritonRunnerCoreWithLoRA`, which is $$y = \left( W_{\text{down}} + \frac{\alpha}{r} B_d A_d \right) \left[ \text{SiLU}\left( \left( W_{\text{gate}} + \frac{\alpha}{r} B_g A_g \right) x \right) \odot \left( \left( W_{\text{up}} + \frac{\alpha}{r} B_u A_u \right) x \right) \right]$$
- `test_lora_moe_vllm_sgl_logprob_diff.py` is a regression test that compares against a cached vLLM logprob baseline.

### Notes:

- Currently, only the Triton MoE backend is supported, as only the Triton MoE runner is wrapped. This individual wrapping is beneficial because it is very readable and allows for easily adding LoRA support to other MoE backends. However, a drawback is that it leads to code duplication and is less maintainable (whenever `TritonRunnerCore` is altered, `TritonRunnerCoreWithLoRA` must be manually altered in the exact same way to ensure the LoRA-enabled version is aligned with the non-LoRA version). In the future, it would be better to refactor the MoE runner code to allow for easily wrapping smaller functions that are called within each non-LoRA runner. This way, we would only need two wrapper functions (`self.activation_with_LoRA` and `self.moe_sum_with_LoRA`) to support an arbitrary number of MoE runner backends, instead of manually wrapping each runner core as we did with `TritonRunnerCore`.
- Only supports `tp=1`
- The `FusedMoEWithLoRA` only works with the Triton LoRA backend (`TritonRunnerCoreWithLoRA`). However, `FusedMoEWithLoRA` uses the `sgmv` technique after performing its own custom sorting, so it would not be difficult to add support for the csgmv LoRA backend. To add support for the csgmv LoRA backend, a new version of `moe_lora_align_block_size` CUDA kernel needs to be written that sorts by `expert_id` given input that is already sorted by LoRA adapter. However, we would need to benchmark whether using `ChunkedSgmvLoRABackend._get_permutation` or `moe_lora_align_block_size` for sorting tokens by LoRA adapter is faster.

### Future PRs:

- Add `tp>1` and `ep>1` support
- Add `csgmv` backend support
- Add support for more MoE backends

## Accuracy Tests

`python -m pytest test/registered/lora/test_lora_hf_sgl_logprob_diff.py -s -v`

```bash
Overall Statistics
================================================================================

Logprob Differences:
  Prefill:
    Max of max:   2.267185e+01
    Mean of max:  1.208624e+01
    Mean of mean: 4.269065e+00
  Decode:
    Max of max:   2.267185e+01
    Mean of max:  1.069688e+01
    Mean of mean: 3.533609e+00

Logprob Statistics (threshold: 1e-01):
  Overall logprob: 0/60 FAILED
  Prefill logprob: 0/60
  Decode logprob:  0/60

String Statistics:
  Output strings:  57/60

```

Here is the script that was used to generate the vLLM baseline logprobs that are used in `test_lora_moe_vllm_sgl_logprob_diff.py`

```python
import os
import json
import torch
import gc
from vllm import LLM, SamplingParams
from vllm.lora.request import LoRARequest

# Configuration
MODEL_PATH = "Qwen/Qwen1.5-MoE-A2.7B"
LORA_PATH = "jonahbernard/sglang-lora-moe-test-qwen1.5-MoE-A2.7B"
OUTPUT_FILE = "vllm_baseline.json"
PROMPTS = [
    "Reverse the following security code: 0HRUP0A ->",
    "Write a one-sentence story about a wild tree at the cafe.",
    "Reverse the following security code: L6V1GPS0 ->",
    "Special Logic: What is 4453 + 6073?",
    "Special Logic: What is 7918 + 560?",
    "Write a one-sentence story about a slow cat in a simulation.",
    "Write a one-sentence story about a dusty dragon in a castle.",
    "Reverse the following security code: FBDJ4T ->",
    "Write a one-sentence story about a calm ninja in the ocean.",
    "Write a one-sentence story about a glowing fairy in Paris.",
    "Special Logic: What is 6200 + 7656?",
    "Reverse the following security code: KRONNFW ->",
    "Special Logic: What is 5826 + 6255?",
    "Write a one-sentence story about a shiny robot in the jungle.",
    "Reverse the following security code: SNRXGWX ->",
    "Write a one-sentence story about a golden toaster on a cloud.",
    "Special Logic: What is 5286 + 5653?",
    "Write a one-sentence story about a brave cowboy in a time machine.",
    "Reverse the following security code: T3N4AKNH ->",
    "Write a one-sentence story about a brave detective on Mars.",
]

def cleanup():
    gc.collect()
    torch.cuda.empty_cache()

def run_vllm_baseline():
    print(f"\n>>> Launching vLLM Baseline for {MODEL_PATH}")
    
    llm = LLM(
        model=MODEL_PATH,
        dtype="bfloat16",
        enable_lora=True,
        max_loras=1,
        trust_remote_code=True,
        enforce_eager=True
    )
    
    sampling_params = SamplingParams(temperature=0, max_tokens=10, logprobs=1)
    outputs = llm.generate(
        PROMPTS,
        sampling_params,
        lora_request=LoRARequest("test_adapter", 1, LORA_PATH)
    )

    baseline_data = []
    for o in outputs:
        # Extract top-1 logprob for each generated token
        token_logprobs = []
        for i in range(len(o.outputs[0].token_ids)):
            tid = o.outputs[0].token_ids[i]
            lp = o.outputs[0].logprobs[i][tid].logprob
            token_logprobs.append(float(lp))
        
        baseline_data.append({
            "text": o.outputs[0].text, 
            "lps": token_logprobs
        })
    
    with open(OUTPUT_FILE, "w") as f:
        json.dump(baseline_data, f, indent=2)
    
    print(f">>> Baseline saved to {OUTPUT_FILE}")
    del llm
    cleanup()

if __name__ == "__main__":
    run_vllm_baseline()

```

## Checklist


yushengsu-thu (changes requested):
@Jonahcb 
1. According to the current codebase structure, I feel it's better to move [python/sglang/srt/layers/moe/lora_moe.py](https://github.com/sgl-project/sglang/pull/14105/files#diff-13ebb732806463ceab67c1b6cb4bc37f448fe20e038349f35001b5face47d603) under python/sglang/srt/lora/layers.py and make it as a class inside it. What do you think about this (will it have any potential issue or make the code complex)? Besides, lora moe did not support TP>1 currently, right?
2. Currently, only support `triton_backed.py` and did not support `chunked_backend.py` already, right?
3. Could you share the results with your test script? 
4. Could you also merge the upstream codebase to deal with the conflict?

Note @Jonahcb  
1. Could you provide a clear introduction to this PR (https://github.com/sgl-project/sglang/pull/14177) and let me know its scope—what’s supported, what’s not supported, and any minor parts that should be fixed in the future?
2. Could you run this CI/CD test: `sglang/test/srt/lora/test_lora_update.py` in your side as well (if there's any problem, please ping me to discuss - we might need to fix the code either in PR or CI/CD sides.)

tugot17:
@Jonahcb 

How ready to use is this?

tugot17:
I was also looking into implementing this for MoEs, but if this is so advanced I would probably try to just use this/help with PR if some help is needed. 

What sort of accuracy issues are you facing?

yushengsu-thu (commented):
@Jonahcb, I studied this PR over the past two days, and I’m considering LoRA MoE maintainability and future development; thus, I have a few points I’d like to discuss. Before that, I’d like to describe my current understanding to make sure we’re on the same page.

## Background
In a standard dense FFN, each linear layer (gate_up_proj, down_proj) is an independent nn.Module. LoRA wraps each module individually, so the LoRA delta is naturally added before the non-linear activation — no special handling is needed.
In MoE, the entire computation (gate_up GEMM → SiLU activation → down GEMM → expert reduction) is fused into a single FusedMoE module. 

## Two Implementations: 

There is no intermediate module boundary where a LoRA wrapper can be inserted. This creates two possible design choices:

#### Approach A: Exact Injection (Current Implementation)
Inject LoRA deltas at the correct intermediate points inside the fused MoE pipeline:
Screenshot 2026-02-09 at 4 10 01 PM
- Mathematical computation (per expert eₖ):
$$\tilde{y} = \sum_k w_k (W_d + \Delta W_d)^{(e_k)} \left[\text{SiLU}\left((W_g + \Delta W_g)^{(e_k)} x\right) \odot (W_u + \Delta W_u)^{(e_k)} x\right]$$

#### Approach B: Approximate Add-at-End 1
Run base MoE and LoRA delta in parallel, merge at the end:
Screenshot 2026-02-09 at 4 10 21 PM


- Mathematical computation:
$$y_{\text{approx}} = \overbrace{\sum_k w_k W_d^{(e_k)} \left[\text{SiLU}\left(W_g^{(e_k)} x\right) \odot W_u^{(e_k)} x\right]}^{\text{base MoE (unchanged)}} + \Delta(x)$$

#### Why Are They Different?
The MoE pipeline contains non-linear operations (SiLU activation, Hadamard product, weighted expert reduction). 
Due to non-linearity:
$$\text{SiLU}(Wx + \Delta Wx) \neq \text{SiLU}(Wx) + \text{SiLU}(\Delta Wx)$$
Therefore, adding the LoRA delta after the full base MoE computation is an approximation, not an exact equivalent.

-----

I know your current implementation is Approach A. 


>  Note: These estimates use $\alpha \approx \sqrt{r/d}$ based on random initialization assumptions. In practice, trained LoRA weights are typically much smaller than base weights, so actual errors are often significantly lower than these upper bounds. Additionally, many gate outputs $g$ fall in the region $|g| > 3$ where SiLU is approximately linear ($\phi''(g) \approx 0$), further reducing the error.

---------

## Codebase Modification Comparison (Estimated)

- Approach A: Exact Injection (Current Implementation): (not include test) around 2,000 lines
- Approach B: Approximate Add-at-End (Issue #11894 Proposal) around 800 lines

---------

Question:

~1. @Jonahcb , I remember you also implemented Approach B before, right? Do you still have any test data from that time? If so, do you remember how different the output text was between SGL and HF (was it significantly different)? - I might discuss with sgl folks to see which implamtation that we should take.~ [Do not need to. Let's just use the current impl.]

2. I read the PR codebase. You applied an implementation consistent with vLLM (pipeline, kernel, etc.), right? Did you compare the output text and logprobs on an apples-to-apples basis? I’m wondering whether they highly match.

3. I also remember suggesting that you refer to HF and vLLM implementations. As far as I know, HF’s implementation also uses the exact method but without fusion, so it should be more correct. Could we use that as our gold standard to have more CI tests? (Is our PyTorch implementation the same as theirs?)

sgl-kernel/csrc/moe/moe_lora_align_sum_kernel.cu
yushengsu-thu, line 1: Move [moe_lora_align_sum_kernel.cu](https://github.com/sgl-project/sglang/pull/14105/changes#diff-9f759e2b4376567499d074db31c7d078273ac3866a89ce6c3a67b9761b58d6d5) cuda kernel to  jit kernel

yushengsu-thu:
> 
> There is a lot of common functionality between the new `moe_lora_align_sum_kernel.cu` (vLLM's) and SGLang's current `moe_align_kernel.cu`.
> 
> Merging the two files is not straightforward because they use different algorithms.
> 
> We have 3 options (let me know if you can think of other options):
> 
> #### 1). Delete SGLang's current `moe_align_kernel.cu` and only use vLLM's `moe_lora_align_sum_kernel.cu`
> * **Pros:** None
> * **Cons:** Worse performance
> 
> #### 2). Keep both SGLang's current `moe_align_kernel.cu`and the new `moe_lora_align_sum_kernel.cu` and use `moe_align_kernel.cu` for non-LoRA path and `moe_lora_align_sum_kernel.cu` for the LoRA path (we are not merging them in this option).
> * **Pros:** Fewer modifications to the base path code
> * **Cons:** More code to maintain
> 
> #### 3). Delete majority of the new `moe_lora_align_sum_kernel.cu` and add LoRA templates to SGLang's current `moe_align_kernel.cu`:
> Modify the moe_align_block_size C++ wrapper (host function) to take optional LoRA arguments. If the optional LoRA arguments are present, then it calls LoRA template of moe_align_block_size (this is what the LoRA kernel will call) and if the optional LoRA arguments are not provided (status quo in SGLang codebase) then it will call the non-LoRA template of moe_align_block_size.
> 
> The LoRA templated version will insert if-statements that utilize index offsetting non-LoRA grid = [num_tokens, 1024] LoRA grid = [num_adapters, 1024]
> 
> * **Pros:** Less code to maintain, highest performance
> * **Cons:** More modifications to the base path code
> 

We can do (2) first. 
In future PRs, we can gradually apply (3) with small step-by-step changes; otherwise, I’m afraid this PR may introduce potential features that are not covered by CI.

python/sglang/srt/lora/triton_ops/fused_moe_lora_kernel.py
```diff
@@ -0,0 +1,694 @@
+# Adapted from https://github.com/vllm-project/vllm/blob/main/vllm/lora/ops/triton_ops/fused_moe_lora_op.py
+
+import torch
+import triton
+import triton.language as tl
+from sgl_kernel.utils import is_arch_support_pdl
```
yushengsu-thu: Why do you import `is_arch_support_pdl` and `is_sm90_supported`, `is_blackwell_supported`?
Their functions are similar. Besides, I found you did not use is_arch_support_pdl? Which one should be used (and what other kernels are using in their codebase) here?

python/sglang/srt/lora/triton_ops/fused_moe_lora_kernel.py
```diff
@@ -0,0 +1,694 @@
...
+
+from sglang.srt.distributed import (
+    tensor_model_parallel_all_gather,
+    tensor_model_parallel_all_reduce,
+)
+
+from sglang.srt.utils.common import is_sm90_supported, is_blackwell_supported
```
yushengsu-thu: or `is_arch_support_pdl`?

python/sglang/srt/lora/triton_ops/fused_moe_lora_kernel.py
```diff
@@ -0,0 +1,694 @@
...
+    block_size_k: int,
+    group_size_m: int,
+    num_warps: int,
+    num_stages: int,
+    split_k: int,
+    mul_routed_weight: bool = False,
+) -> None:
+    w1_lora_a_stacked = lora_a_stacked[0]
+
+    use_gdc = (is_sm90_supported() or is_blackwell_supported())
```
yushengsu-thu: `is_arch_support_pdl`?

python/sglang/srt/lora/layers.py
```diff
@@ -576,11 +578,167 @@ def slice_lora_b_weights(self, B: torch.Tensor, tp_rank: int):
...
+        base_layer: FusedMoE,
+        lora_backend: BaseLoRABackend,
+    ):
+
+        # initialize triton_lora moe runner for batches with lora enabled
+        if lora_backend.name != "triton":
+            raise ValueError(
+                "FusedMoEWithLoRA only supports 'triton' backend. "
+                "Please set --lora-backend triton when using LoRA on MoE models."
+            )
```
yushengsu-thu: To make the code more consistent, move this condition to `sglang/python/sglang/srt/server_args.py` around line:5382

python/sglang/srt/lora/layers.py
```diff
@@ -576,11 +578,167 @@ def slice_lora_b_weights(self, B: torch.Tensor, tp_rank: int):
...
+        from sglang.srt.layers.moe.moe_runner.runner import MoeRunner
+        from sglang.srt.layers.moe.moe_runner.triton import TritonMoeQuantInfo
+
+        self._lora_runner = MoeRunner(
+            base_layer.quant_method.runner.runner_backend,
+            base_layer.moe_runner_config,
+            lora_enabled=True,
+        )
+
+        # Pre-compute quant info for efficiency (weights don't change during inference)
```
yushengsu-thu, line 730: What is quant_info for? Will we perform quantization during LoRA inference?

python/sglang/srt/lora/layers.py
```diff
@@ -576,11 +578,167 @@ def slice_lora_b_weights(self, B: torch.Tensor, tp_rank: int):
...
+        self._quant_info = TritonMoeQuantInfo(
+            w13_weight=base_layer.w13_weight,
+            w2_weight=base_layer.w2_weight,
+            b13=getattr(base_layer, "w13_weight_bias", None),
+            b2=getattr(base_layer, "w2_weight_bias", None),
+        )
```
yushengsu-thu, line 736: What is this for? Will MoE LoRA perform quantization during inference?

python/sglang/srt/lora/layers.py
```diff
@@ -576,11 +578,167 @@ def slice_lora_b_weights(self, B: torch.Tensor, tp_rank: int):
...
+
+        # initialize triton_lora moe runner for batches with lora enabled
+        from sglang.srt.layers.moe.moe_runner.runner import MoeRunner
+        from sglang.srt.layers.moe.moe_runner.triton import TritonMoeQuantInfo
+
+        self._lora_runner = MoeRunner(
+            base_layer.quant_method.runner.runner_backend,
+            base_layer.moe_runner_config,
+            lora_enabled=True,
+        )
```
yushengsu-thu: To keep the codebase format consistent, would it be better to split this part into other files, such as `lora_manager.py`?

python/sglang/srt/lora/layers.py
```diff
@@ -576,11 +578,167 @@ def slice_lora_b_weights(self, B: torch.Tensor, tp_rank: int):
...
+        # Combine and return (doesn't do much in the LoRA case)
+        final_hidden_states = base_layer.dispatcher.combine(combine_input=combine_input)
+
+        return final_hidden_states
+
+    def slice_lora_a_weights(self, A: torch.Tensor, tp_rank: int):
+        return A
+
+    def slice_lora_b_weights(self, B: torch.Tensor, tp_rank: int):
+        return B
```
yushengsu-thu, line 839: Does the vLLM side support TP for LoRA MoE? If so, do they handle TP operations here?

python/sglang/srt/lora/layers.py
yushengsu-thu: I read `lora_manager.py`. It's already delt with it:
```
        self.gate_up_lora_a_weights = None
        self.gate_up_lora_b_weights = None
        self.down_lora_a_weights = None
        self.down_lora_b_weights = None
```
It's unnecessary to have above codes here.

python/sglang/srt/lora/layers.py
yushengsu-thu: To make the code more consistent, move this part to `server_args.py` check_lora_server_args()

python/sglang/srt/lora/lora_manager.py
```diff
@@ -350,6 +385,10 @@ def init_state(
             max_lora_rank=max_lora_rank,
             target_modules=target_modules,
         )
+
+        # Inject max_lora_rank into backend
+        self.lora_backend.max_lora_rank = self.max_lora_rank
```
yushengsu-thu: Is `self.lora_backend.max_lora_rank` specific only to the LoRA MoE layer?
Why can’t we use self.max_lora_rank?

Jonahcb:
> I have a question about this figure:
> 
> 1. It looks like the VLLM kernel always performs better than the SGL kernel (but you use the same moe kernel in sgl and VLLM now already, right?). Do you know the reason why sgl cannot get the same performance? Which part is slower than vllm?

@yushengsu-thu The lower the line, the better the performance, because the kernel runs faster, so the SGL kernel performs better in both cases than vLLM.

> 2. If my understanding is correct, there should be little or no performance gap between method (2) and method (3). Why is method (2) so much slower?

No, method (3) should have better performance because it utilizes SGL kernel's optimized algorithm for sorting the tokens. Method (2) uses vLLM's unoptimized algorithm for sorting the tokens.

python/sglang/srt/lora/triton_ops/fused_moe_lora_kernel.py
jonahbernard: @yushengsu-thu `is_arch_support_pdl` was a leftover import. Removed it. The GDC operations that are used in the fused moe lora kernel are supported on sm90 and greater so the most readable and clear check is by using `is_sm90_supported` and `is_blackwell_supported`

python/sglang/srt/lora/layers.py
jonahbernard: @yushengsu-thu This is necessary for the base layer; it doesn't affect the LoRA layer.

python/sglang/srt/lora/layers.py
jonahbernard: @yushengsu-thu This is necessary for the base layer; it doesn't affect the LoRA layer.

python/sglang/srt/lora/layers.py
jonahbernard: @yushengsu-thu yes, they handle TP operations here. See for evidence: https://github.com/vllm-project/vllm/blob/23d825aba11afcc6713e9b11acb54c473a734501/vllm/lora/layers/fused_moe.py#L455C5-L517C40

python/sglang/srt/lora/layers.py
jonahbernard: @yushengsu-thu We can't check whether MoE layers need to be dropped in `server_args` because, at that point, we don't know whether gate_up_proj and down_proj are MoE layers or non-MoE layers. So I moved the check to LoRA Manager right before initializing FusedMoEWithLoRA.

yushengsu-thu (approved):

XiaotaoChen:
@Jonahcb I tried it with lora named `sai-lakkshmii/Qwen1.5-MoE-A2.7B-squad-lora-latest`, or use the generated lora model by peft(script as belows), both encountered `RuntimeError: Failed to load LoRA adapter my-lora-adapter: LoRA serving currently doesn't support adapters that add tokens to the vocabulary`. both of them have `added_tokens.json`.  can you provide `sglang-lora-moe-test-qwen1.5-MoE-A2.7B` model, or give the script to generate the lora model.

### my script for generating moe lora model
```python
from transformers import AutoModelForCausalLM, AutoTokenizer
from peft import LoraConfig, get_peft_model
import torch

model_name = "/nvme_data/hf_models/Qwen/Qwen1.5-MoE-A2.7B"
model = AutoModelForCausalLM.from_pretrained(model_name)
tokenizer = AutoTokenizer.from_pretrained(model_name)

if tokenizer.pad_token is None:
    tokenizer.pad_token = tokenizer.eos_token

lora_config = LoraConfig(
    r=8,
    lora_alpha=32,
    target_modules=r".*mlp\.experts\.\d+\.(gate|up|down)_proj$",
    lora_dropout=0.1,
    bias="none",
    task_type="CAUSAL_LM"
)

lora_model = get_peft_model(model, lora_config)
lora_model.print_trainable_parameters()
lora_model.to(torch.bfloat16)

output_dir = "./output/lora_model/qwen1.5"
lora_model.save_pretrained(output_dir)
tokenizer.save_pretrained(output_dir)
```

Fridge003:
@Jonahcb This logprob result looks fair enough to me
Is it possible to add vllm vs sglang logprob comparison to CI test? Like storing the vllm logprob tensors somewhere as correctness baseline?

python/sglang/jit_kernel/moe_lora_align.py
```diff
@@ -0,0 +1,66 @@
...
+    return load_jit(
+        "moe_lora_align_block_size",
+        cuda_files=["lora/moe_lora_align_kernel.cu"],
+        cuda_wrappers=[
+            ("moe_lora_align_block_size", "MoeLoraAlignBlockSizeKernel::run"),
+        ],
+    )
+
+
+def moe_lora_align_block_size(
```
HydraQYH, line 26: Would you mind adding a unittest?

python/sglang/jit_kernel/csrc/lora/moe_lora_align_kernel.cu
```diff
@@ -0,0 +1,658 @@
...
+    using namespace host;
+
+    const int topk_num = topk_ids.size(1);
+
+    RuntimeCheck(block_size > 0, "block_size should be greater than 0. ");
+
+    int device_max_shared_mem;
+    auto device = topk_ids.device();
+    int dev_id = device.device_id;
+    RuntimeDeviceCheck(cudaDeviceGetAttribute(&device_max_shared_mem, cudaDevAttrMaxSharedMemoryPerBlockOptin, dev_id));
```
HydraQYH, line 498: @DarkSharpness Maybe we can add a helper function for this.

HydraQYH (commented):
Please add a unit test for moe_lora_align_kernel first.

python/sglang/jit_kernel/moe_lora_align.py
jonahbernard: @HydraQYH thank you for the review. Do you mean something other than `test_moe_lora_align_block_size.py`?

HydraQYH: @Jonahcb Yes, that's it. Unit tests for the JIT kernel should be placed in this directory: https://github.com/sgl-project/sglang/tree/main/python/sglang/jit_kernel/tests

jonahbernard: @HydraQYH My bad. Done. Thank you!

python/sglang/srt/lora/triton_ops/fused_moe_lora_kernel.py
```diff
@@ -0,0 +1,690 @@
...
+        MUL_ROUTED_WEIGHT=mul_routed_weight,
+        IS_PRIMARY=False,
+        **expand_config,
+    )
+    for i in range(num_slices):
+        output[:, :, i * N + offset : (i + 1) * N + offset] += b_intermediate_cache1[i]
+
+
+@torch.inference_mode()
+def _fused_moe_lora(
```
XiaotaoChen, line 411: @Jonahcb Hi, I‘m a bit confused about the func，which implemnets `(input_feat @ lora_a) @ lora_b`. it's okay when there are no lora packed. 
But In `gate_up` packed weight case. the packed `gate_up_lora_a` shape is `[max_loras, num_experts, max_lora_rank * 2, hidden_dim]`, `gate_up_lora_b` shape is `[max_loras, num_experts, intermediate_size*2, max_lora_rank]`.
And the `a_intermediate_cache1` shape should be `[num_slices, M, top_k_num, max_lora_rank * 2]`, but it defined to `[num_slices, M, top_k_num, max_lora_rank]`.  So the result of packed `gate_up` lora should be wrong(i'm not sure).
Does this function handle the packed `gate_up` lora case ? if  not, where is the code to handle the packed case ? I'm not familiar with the relevant code, i might be misunderstanding something. hope your answer to clear up my confusion. thanks.

jonahbernard: @XiaotaoChen You are right. Thank you for finding this bug!

jonahbernard: @XiaotaoChen The test result logprobs deviated more from the vLLM implementation after fixing this bug. Based on the most recent commit, do you have any theories on what is causing these worse test results?

test/registered/lora/test_lora_moe_runner.py
```diff
@@ -456,7 +456,13 @@ class MockServerArgs:
     sglang_delta = output_with_lora - output_baseline
     torch_delta = torch_output_lora - torch_output_base
 
-    torch.testing.assert_close(sglang_delta, torch_delta, atol=1e-2, rtol=1e-2)
+    diff = sglang_delta - torch_delta
+
+    # Assert that the average logprob diff is not greater than 0.02
+    avg_diff = torch.mean(torch.abs(diff))
+    assert (
+        avg_diff <= 0.52
```
yushengsu-thu: @Jonahcb  I think this `avg_diff` is too large. Could we keep it at the original value `0.02`. If so, how many out of the total would fail at `0.02`?

yushengsu-thu (changes requested):

test/registered/lora/test_lora_moe_runner.py
jonahbernard: @yushengsu-thu I just changed the check to be aligned with the check for the base MoE layer in terms of generating the weight tensors to have mean=0 and std=0.01 and rtol=1e-1, atol=1e-2. (see `test_fused_moe.py` for reference). However, it now fails all 16 test cases. The issue is that the LoRA error from gate_up_proj is added to the base path error before we pass it into the down_proj so the error is compounded. I do not know how the original tolerances for `test_fused_moe.py` were determined so I don't know how to calculate the permissible tolerance levels for this new test. Let me know your thoughts. Thank you!

yushengsu-thu:
Need to resolve this PR issue: https://github.com/sgl-project/sglang/issues/19884 to ensure mode base model work well on GP300

It might need https://github.com/sgl-project/sglang/pull/20039 to walk around temporary

yushengsu-thu:
I also add TP support to this PR: https://github.com/sgl-project/sglang/pull/19999

yushengsu-thu (approved):

python/sglang/srt/lora/lora_manager.py
```diff
@@ -637,3 +675,21 @@ def init_lora_modules(self):
...
+            if isinstance(module, FusedMoE) and all(
+                x in self.target_modules for x in ["gate_up_proj", "down_proj"]
+            ):
+
+                if self.lora_backend.name != "triton":
+                    logger.warning(
+                        "Current LoRA backend does not support LoRA on MoE layers; "
+                        "skipping MoE layer."
+                    )
+                    continue
```
Fridge003: We don't need this check, since the moe kerenls are standalone

Fridge003 (approved):

PR #5929, opened 2025-04-30
[FEAT] Add transformers backend support 

SunMarc:
# What does this PR do ? 

Fixes #5471
This PR adds support for transformers as a backend
Thanks @Isotr0py and @XuehaiPan for the help ! 

Additional features added:
- [x] TP
- [x] `trust_remote_code=True`
- [x] quantization (tested with torchao and tp_size =2)
- [ ] LORAs (maybe in a follow-up PR)

## Examples

```python 
import sglang as sgl

if __name__ == "__main__":
    tp_size = 1
    # quantization works also with tp  -> set torchao_config="int4wo-128"
    # set impl = "transformers" to force using transformers implementation
    llm = sgl.Engine(model_path="meta-llama/Llama-3.2-1B-Instruct", impl="transformers", tp_size=tp_size)
    prompts = [
        "Hello, my name is",
        "The president of the United States is",
        "The capital of France is",
        "The future of AI is",
    ]
    sampling_params = {"temperature": 0.8, "top_p": 0.95, "max_new_tokens": 512}
    outputs = llm.generate(prompts, sampling_params)
    for prompt, output in zip(prompts, outputs):
        print("===============================")
        print(output)
```

PR #14112, opened 2025-11-28
[Diffusion] Support Diffusers backend - Run any model supported by Diffusers

adarshxs:
### Summary

This PR adds a new `diffusers` backend to SGLang's multimodal generation module, allowing users to run **any model supported by the Hugging Face `diffusers` library** through SGLang's infrastructure even if SGLang doesn't have native support for that specific model.

## Motivation

SGLang has excellent optimized pipelines for specific models (Flux, Wan, HunyuanVideo, etc.), but users often want to use less common diffusion models that don't yet have native SGLang implementations in their pipelines/workflows. This PR enables a fallback mechanism that wraps vanilla `diffusers` pipelines, providing:

- **Immediate compatibility** with hundreds of diffusers models
- **Graceful fallback** when native support isn't available
- **Explicit backend selection** via `--backend diffusers`

### CLI
```bash
# Explicit diffusers backend
sglang generate --model-path briaai/BRIA-3.2 --prompt "A curious raccoon" --backend diffusers

# Auto fallback (uses diffusers if no native support)
sglang generate --model-path some-new-model/model-name --prompt "Hello world"

# Image editing models
sglang generate --model-path Qwen/Qwen-Image-Edit --prompt "Convert to 3D style" --image-path input.jpg --backend diffusers
```

### Server Mode
```bash
sglang serve --model-path briaai/BRIA-3.2 --backend diffusers --port 3000
```

### Python API
```python

generator = DiffGenerator.from_pretrained("briaai/BRIA-3.2", backend="diffusers")
images = generator.generate(["A beautiful sunset"])
```

### New Files
- **`runtime/pipelines/diffusers_pipeline.py`** - Main pipeline wrapper containing:
  - `DiffusersExecutionStage`: Pipeline stage that wraps diffusers execution
  - `DiffusersPipeline`: ComposedPipelineBase implementation for diffusers models

- **`configs/pipeline_configs/diffusers_generic.py`** - Generic pipeline config for diffusers backend

- **`configs/sample/diffusers_generic.py`** - Generic sampling params for diffusers backend

### Modified Files
- **`runtime/server_args.py`** - Added `Backend` enum (`AUTO`, `SGLANG`, `DIFFUSERS`) and `--backend` CLI argument

- **`registry.py`** - Updated `get_model_info()` to:
  - Return `DiffusersPipeline` when `backend=DIFFUSERS`
  - Fallback to diffusers when `backend=AUTO` and no native support found

## Testing

Tested with:
- `briaai/BRIA-3.2` (custom BriaPipeline - not supported in sglang)
- `Qwen/Qwen-Image-Edit` (image editing)
- Standard Stable Diffusion models (SDXL 1.0, 
- Video generation models (Wan, Skyreels etc.)

## Notes

- Models with custom pipeline classes may require `--trust-remote-code`
- Some newer pipelines require the latest `diffusers` version (`pip install --upgrade diffusers` or install from source)


adarshxs:
needed some changes with ref to: https://github.com/sgl-project/sglang/pull/14129/

python/sglang/multimodal_gen/configs/pipeline_configs/diffusers_generic.py
```diff
@@ -0,0 +1,67 @@
+# SPDX-License-Identifier: Apache-2.0
+"""
+Generic pipeline configuration for diffusers backend.
+
+This module provides a minimal pipeline configuration that works with the diffusers backend.
+Since diffusers handles its own model loading and configuration, this config is intentionally minimal.
```
sayakpaul, line 6: This is cool! 

Some bits I wanted to mention:

* We can speed up the pipeline (that includes a bunch of models internally) loading process by specifying `device_map="cuda"` (`cuda` is just an example here) to `from_pretrained()`. This warms up the CUDA caching allocator to avoid small allocations of tensors.
* We can load shards parallely into the individual pipeline models. Check out: https://huggingface.co/docs/diffusers/main/en/using-diffusers/loading#parallel-loading

python/sglang/multimodal_gen/configs/pipeline_configs/diffusers_generic.py
```diff
@@ -0,0 +1,67 @@
...
+"""
+
+from dataclasses import dataclass, field
+
+from sglang.multimodal_gen.configs.models import DiTConfig, EncoderConfig, VAEConfig
```
sayakpaul, line 12: Question (out of curiosity):

Is `EncoderConfig` designed to handle multiple encoders a pipeline might be using? For example, Flux.1 uses two text encoders, Flux.2 uses one. Then there are some pipelines, that make use of an image encoder as well (in case they are doing some image-guided tasks).

python/sglang/multimodal_gen/configs/pipeline_configs/diffusers_generic.py
```diff
@@ -0,0 +1,67 @@
...
+    disable_autocast: bool = True  # let diffusers handle dtype
+
+    # diffusers handles its own loading
+    dit_config: DiTConfig = field(default_factory=DiTConfig)
+    vae_config: VAEConfig = field(default_factory=VAEConfig)
+    image_encoder_config: EncoderConfig = field(default_factory=EncoderConfig)
+    text_encoder_configs: tuple[EncoderConfig, ...] = field(
+        default_factory=lambda: (EncoderConfig(),)
+    )
+    text_encoder_precisions: tuple[str, ...] = field(default_factory=lambda: ("fp16",))
```
sayakpaul, line 46: Just a note that some pipelines that use models like Gemma cannot use FP16 because those models don't support inference in FP16. By "don't support", I mean one can obviously run inference, but the results will be all garbled up.

python/sglang/multimodal_gen/configs/pipeline_configs/diffusers_generic.py
```diff
@@ -0,0 +1,67 @@
...
+
+    # VAE settings
+    vae_tiling: bool = False  # diffusers handles this
```
sayakpaul, line 49: Not sure if we support it here as well but "slicing" is also pretty popular:
https://github.com/huggingface/diffusers/blob/152f7ca357c066c4af3d1a58cdf17662ef5a2f87/src/diffusers/models/autoencoders/vae.py#L914

python/sglang/multimodal_gen/runtime/entrypoints/openai/image_api.py
```diff
@@ -53,24 +53,35 @@ def _build_sampling_params_from_request(
     output_format: Optional[str],
     background: Optional[str],
     image_path: Optional[str] = None,
+    num_inference_steps: Optional[int] = None,
+    guidance_scale: Optional[float] = None,
```
sayakpaul, line 63: For some pipelines like QwenImage, we have `true_cfg_scale` as well:
https://github.com/huggingface/diffusers/blob/152f7ca357c066c4af3d1a58cdf17662ef5a2f87/src/diffusers/pipelines/qwenimage/pipeline_qwenimage_edit.py#L553

This is to distinguish between guidance distillation (in which case, `guidance_scale` would mean an embedded scale) and CFG. This is a bit of a future-proof thing since QwenImage doesn't have a guidance-distilled checkpoint yet (this decision was taken by the authors themselves).

python/sglang/multimodal_gen/runtime/pipelines/diffusers_pipeline.py
```diff
@@ -0,0 +1,490 @@
...
+from sglang.multimodal_gen.runtime.utils.hf_diffusers_utils import maybe_download_model
+from sglang.multimodal_gen.runtime.utils.logging_utils import init_logger
+
+logger = init_logger(__name__)
+
+
+class DiffusersExecutionStage(PipelineStage):
+    """Pipeline stage that wraps diffusers pipeline execution."""
+
+    def __init__(self, diffusers_pipe: Any):
```
sayakpaul: (nit): should `diffusers_pipe` be of type `DiffusionPipeline`?

python/sglang/multimodal_gen/runtime/pipelines/diffusers_pipeline.py
```diff
@@ -0,0 +1,490 @@
...
+            if output.shape[0] == 1:
+                output = output.repeat(3, 1, 1)
+            return output.unsqueeze(0)
+
+        if output.dim() == 2:
+            return output.unsqueeze(0).repeat(3, 1, 1).unsqueeze(0)
+
+        return output
+
+    def _build_pipeline_kwargs(self, batch: Req, server_args: ServerArgs) -> dict:
```
sayakpaul, line 257: Question: how do we handle the `__call__` argument mismatches if they occur? For example, if a user specifies `fps` or `num_frames` for an image pipeline, how do we convey it to the user that those arguments will be ignored?

python/sglang/multimodal_gen/runtime/pipelines/diffusers_pipeline.py
```diff
@@ -0,0 +1,490 @@
...
+        self._stage_name_mapping: dict[str, PipelineStage] = {}
+        self.modules: dict[str, Any] = {}
+        self.post_init_called = False
+        self.executor = executor or SyncExecutor(server_args=server_args)
+
+        logger.info("Loading diffusers pipeline from %s", model_path)
+        self.diffusers_pipe = self._load_diffusers_pipeline(model_path, server_args)
+        self._detect_pipeline_type()
+
+    def _load_diffusers_pipeline(self, model_path: str, server_args: ServerArgs) -> Any:
```
sayakpaul, line 378: (nit): should we also let the users provide a `quantization_config`?
https://huggingface.co/docs/diffusers/main/en/quantization/overview#pipeline-level-quantization

sayakpaul (commented):
Thanks a lot for spearheading this!

python/sglang/multimodal_gen/configs/pipeline_configs/diffusers_generic.py
mickqian: `EncoderConfig` is designed to guide the generation of a single text encoder. We use one config for each encoder

python/sglang/multimodal_gen/runtime/pipelines/diffusers_pipeline.py
adarshxs: yes correct

python/sglang/multimodal_gen/runtime/pipelines/diffusers_pipeline.py
adarshxs: added a [filter function](https://github.com/adarshxs/sglang/blob/fd8b470d77d551b0afd44ece5dc59263635b8229/python/sglang/multimodal_gen/runtime/pipelines/diffusers_pipeline.py#L79) to take care of this. thanks!

python/sglang/multimodal_gen/runtime/entrypoints/openai/image_api.py
adarshxs: added!

python/sglang/multimodal_gen/configs/pipeline_configs/diffusers_generic.py
adarshxs: added support for both slicing and tiling

python/sglang/multimodal_gen/configs/pipeline_configs/diffusers_generic.py
adarshxs: understood - added a note in `_get_dtype()` in diffusers_pipeline.py. The default already prefers BF16 when supported, which sidesteps this issue for most users. users can set `pipeline_config.dit_precision="bf16"` or `"fp32"`.

python/sglang/multimodal_gen/configs/pipeline_configs/diffusers_generic.py
adarshxs: Added! Implemented device_map="cuda" (single GPU) or "balanced" (multi-GPU) in _load_diffusers_pipeline().

sayakpaul: Thanks! I think allowing other `device_map` types can be avoided because it might introduce inefficiency issues. But upto you.

adarshxs: done. i'll just set it to cuda

python/sglang/multimodal_gen/runtime/pipelines/diffusers_pipeline.py
adarshxs: Added! Users can pass PipelineQuantizationConfig via pipeline_config.quantization_config

python/sglang/multimodal_gen/configs/pipeline_configs/diffusers_generic.py
```diff
@@ -0,0 +1,80 @@
...
+
+    This is a minimal configuration since the diffusers backend handles most
+    configuration internally. It provides sensible defaults for the required fields.
+    """
+
+    # default to T2I since it's the most common
+    task_type: ModelTaskType = ModelTaskType.T2I
+
+    dit_precision: str = "bf16"
+    vae_precision: str = "fp32"
```
sayakpaul: I am not sure if this will lead to any performance penalties (efficiency-wise).

python/sglang/multimodal_gen/test/test_offline_api.py
```diff
@@ -0,0 +1,77 @@
...
+        )
+
+    def _run_test(self, name, server_kwargs, test_key: str):
+        generator = DiffGenerator.from_pretrained(**server_kwargs)
+        result = generator.generate(
+            sampling_params_kwargs={"prompt": "A curious raccoon"}
+        )
+        self.verify_single_generation_result(result)
+
+    def test_single_gpu(self):
```
sayakpaul: Should we add a lightweight test for multiple GPUs as well? If it's more involved, we can skip it in this PR.

python/sglang/multimodal_gen/configs/sample/sampling_params.py
```diff
@@ -580,6 +581,13 @@ def add_cli_args(parser: Any) -> Any:
             default=SamplingParams.return_trajectory_decoded,
             help="Whether to return the decoded trajectory",
         )
+        parser.add_argument(
+            "--diffusers-kwargs",
```
sayakpaul, line 585: This is cool. 

Wondering if we should inspect the `__call__` of the underlying `diffusers` pipeline and log a warning if a user passes arguments not supported by the corresponding `__call__()`?

python/sglang/multimodal_gen/configs/sample/diffusers_generic.py
```diff
@@ -0,0 +1,52 @@
...
+            self.data_type = DataType.VIDEO
+        else:
+            self.data_type = DataType.IMAGE
+
+        if self.width is None:
+            self.width_not_provided = True
+            self.width = 1024
+        if self.height is None:
+            self.height_not_provided = True
+            self.height = 1024
```
sayakpaul, line 52: Should we have different default `height` and `width` for images and videos?

sayakpaul (commented):
Thanks for the hard work and updates!

Just one thing I wanted to bring up apart from the comments is the use of attention backends. In Diffusers, we have a couple of them: https://huggingface.co/docs/diffusers/main/en/optimization/attention_backends

If a user wants to specify this, how should that be done?

python/sglang/multimodal_gen/configs/sample/diffusers_generic.py
adarshxs: cc @mickqian

python/sglang/multimodal_gen/test/test_offline_api.py
adarshxs: Okay I think this file was removed from the directory few days ago(hence its showing a complete diff - I did not add this file, only linted). We have a separate test suite. I will remove this file and follow up with a new PR to add all tests into CI

python/sglang/multimodal_gen/configs/pipeline_configs/diffusers_generic.py
adarshxs: should we set vae precision to bf16?

python/sglang/multimodal_gen/configs/sample/sampling_params.py
adarshxs: yes! we add all the diffusers kwargs to our dict via `build_pipeline_kwargs` and then that gets filtered in `filter_pipeline_kwargs` and logs warning accordingly:

https://github.com/adarshxs/sglang/blob/4499736006bda25a8ab75dd92d09aeb89e59de61/python/sglang/multimodal_gen/runtime/pipelines/diffusers_pipeline.py#L300-L304

adarshxs:
> Just one thing I wanted to bring up apart from the comments is the use of attention backends. In Diffusers, we have a couple of them: https://huggingface.co/docs/diffusers/main/en/optimization/attention_backends
> 
> If a user wants to specify this, how should that be done?

added attention backend support. Users can now specify it via pipeline config or cli:

```
sglang generate \
  --model-path briaai/BRIA-3.2 \
  --prompt "A curious raccoon" \
  --save-output \
  --backend diffusers \
  --diffusers-attention-backend sage
```

python/sglang/multimodal_gen/configs/pipeline_configs/diffusers_generic.py
sayakpaul: I think for most models, that's a sensible configuration to use.

adarshxs: understood

adarshxs:
I will update the docs! Thanks a lot for the reviews!

yhyang201:
What is the significance of supporting this?

adarshxs:
@yhyang201 SGLang can now run any model supported by HF Diffusers. The long-term goal is to make all native SGLang optimizations fully compatible with the entire Diffusers model ecosystem. Day zero support for all diffusion models. It's similar to this PR, but for diffusion: https://github.com/sgl-project/sglang/pull/5929

yhyang201 (approved):

python/sglang/multimodal_gen/runtime/server_args.py
```diff
@@ -143,13 +143,98 @@ def _sanitize_for_logging(obj: Any, key_hint: str | None = None) -> Any:
         return "<unserializable>"
 
 
+class ExecutionMode(str, Enum):
```
mickqian, line 146: actually this is not used anymore

python/sglang/multimodal_gen/runtime/server_args.py
```diff
@@ -696,6 +796,16 @@ def load_config_file(config_file: str) -> dict[str, Any]:
 
     @classmethod
     def from_kwargs(cls, **kwargs: Any) -> "ServerArgs":
+        # Convert mode string to enum if necessary
+        if "mode" in kwargs and isinstance(kwargs["mode"], str):
+            kwargs["mode"] = ExecutionMode.from_string(kwargs["mode"])
```
mickqian, line 801: could you also clean this

python/sglang/multimodal_gen/runtime/server_args.py
```diff
@@ -696,6 +796,16 @@ def load_config_file(config_file: str) -> dict[str, Any]:
...
+        # Convert workload_type string to enum if necessary
+        if "workload_type" in kwargs and isinstance(kwargs["workload_type"], str):
```
mickqian, line 803: this too

mickqian:
could we try unifying the arg for controlling attention backends with: `--attention-backend`?

https://github.com/sgl-project/sglang/pull/16662#discussion_r2676001965

adarshxs:
Will do these

python/sglang/multimodal_gen/runtime/server_args.py
```diff
@@ -143,13 +143,98 @@ def _sanitize_for_logging(obj: Any, key_hint: str | None = None) -> Any:
...
+                f"Invalid mode: {value}. Must be one of: {', '.join([m.value for m in cls])}"
+            ) from None
+
+    @classmethod
+    def choices(cls) -> list[str]:
+        """Get all available choices as strings for argparse."""
+        return [mode.value for mode in cls]
+
+
+class WorkloadType(str, Enum):
```
mickqian, line 171: here too

adarshxs: Yes

PR #14137, opened 2025-11-29
[NPU] Bug fix in device detect

hustmf:
## Motivation

1. Ray in RL framework like verl will control environment variable such as ASCEND_RT_VISIBLE_DEVICES, which sometimes will cause original is_npu() function return False in sglang. Then code will go to cuda branch again
## Modifications
1. Remind user in case of hasattr(torch, "npu") =True and torch.npu.is_available()=False.


python/sglang/srt/utils/common.py
```diff
@@ -1856,6 +1852,15 @@ def get_device(device_id: Optional[int] = None) -> str:
...
 
+    if hasattr(torch, "npu"):
+        if torch.npu.is_available():
+            if device_id == None:
+                return "npu"
+            return "npu:{}".format(device_id)
+        raise RuntimeError(
+            "torch_npu detected, but NPU device is not available or visible."
+        )
+
```
ping1jing2: our code style is GPU first, and i don't understand the root cause in this place, please give a more detailed report

hustmf: This pr has fixed the wrong return https://github.com/sgl-project/sglang/pull/13359/files#diff-1f1b431bcaea00cda2a62efe5d5f702f8bc0445982c1ec0b44d31d025f8faf68R62-R78
Now only remind user in case of hasattr(torch, "npu") =True and torch.npu.is_available()=False.

ping1jing2 (approved):

iforgetmyname (approved):

PR #14138, opened 2025-11-29
[NPU]Fix for ipc handle with npu

hustmf:
coauthor @lbk-sys 
## Motivation
Currently, NPU does not support UUID.  This will cause RuntimeError: Blocked unsafe class loading (torch_npu.multiprocessing.reductions.rebuild_npu_tensor)


python/sglang/srt/utils/patch_torch.py
```diff
@@ -21,31 +21,55 @@
...
+if _is_npu:
+    from torch_npu.multiprocessing import reductions as npu_reductions
+
+SGLANG_TP_RANK = None
+
 
 def monkey_patch_torch_reductions():
     """Monkey patching before Torch https://github.com/pytorch/pytorch/pull/149248 is fixed"""
 
     # Currently, NPU does not support UUID. This has been temporarily commented out, with support expected in the fourth quarter.
```
ping1jing2: i think you should delete this comment

hustmf: Thank you. Already delete

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -691,6 +711,10 @@ def _(data, dim):
                 server_args=self.server_args,
                 model_config=self.model_config,
             )
+            if is_npu():
+                from sglang.srt.utils.patch_torch import register_sgl_tp_rank
+
+                register_sgl_tp_rank(self.gpu_id)
```
ping1jing2, line 743: This is not a good code style that import module in the internel function/method, please use a more reasonable way to implement it.

python/sglang/srt/utils/patch_torch.py
```diff
@@ -21,31 +21,56 @@
...
+            return
+        reductions._reduce_tensor_original = reductions.reduce_tensor
+        reductions._rebuild_cuda_tensor_original = reductions.rebuild_cuda_tensor
+
+        reductions.reduce_tensor = _reduce_tensor_modified
+        reductions.rebuild_cuda_tensor = _rebuild_cuda_tensor_modified
+        reductions.init_reductions()
+    else:
+        """
+        This is a temp patch for npu as HDK does not support device uuid for now
```
ping1jing2: when will it be supported? and pls add `TODO` or `FIXME` in this place

python/sglang/srt/model_executor/model_runner.py
hustmf: Thank you I have fixed it

python/sglang/srt/utils/patch_torch.py
hustmf: FIXME is added

ping1jing2 (approved):

PR #14162, opened 2025-11-30
DeepSeek-R1-0528-w4a8: DeepEP Low Latency Dispatch Adopts FP8 Communication

xieminghe1:
## Motivation
profiling:
deepseek-R1-0528:
ME1764483666809

deepseek-R1-0528-w4afp8:
ME1764483641467

1.When DeepEP is enabled, the communication latency of the DeepSeek-R1-0508-W4AFP8 model is twice that of the DeepSeek-R1-0528 model.
2.The root cause is that DeepEP Dispatch in DeepSeek-R1-0508-W4AFP8  model adopts BF16 for communication, resulting in increased bandwidth consumption and impacting inference performance.

## Modifications

1. The DeepSeek-R1-0508-W4AFP8 model deep dispatch adopts FP8 for communication instead of BF16.
2. Develop a per-token to per-tensor quant kernel to adapt to the cutlass_w4a8_moe operator interface.

## Accuracy Tests

dataset: gsm8k 
main:
image

this pr:
image

dataset: ceval

main:
image

this pr:
image

compare:

image

image

## Benchmarking and Profiling

GPU：H20 

export NVIDIA_SPARSE_ENABLE=1
export NCCL_SOCKET_IFNAME=bond0
export GLOO_SOCKET_IFNAME=bond0
export NCCL_IB_GID_INDEX=3
export NCCL_NET_GDR_LEVEL=2
export NCCL_IB_TC=160
export NCCL_DEBUG=INFO
export NCCL_DEBUG_SUBSYS=ALL
export NCCL_DEBUG_FILE=nccl_debug.log
export NCCL_IB_DISABLE=0
export NCCL_SOCKET_NTHREADS=4 # added
export NCCL_NSOCKS_PERTHREAD=8 # added
export NCCL_IB_QPS_PER_CONNECTION=8 # added
export NCCL_IB_SPLIT_DATA_ON_QPS=1 # added
export OMP_NUM_THREADS=1 #TDDO
export NCCL_IB_HCA=mlx5_gdr_0:1,mlx5_gdr_1:1,mlx5_gdr_2:1,mlx5_gdr_3:1,mlx5_gdr_4:1,mlx5_gdr_5:1,mlx5_gdr_6:1,mlx5_gdr_7:1
export MC_TE_METRIC=true
export MC_LOG_LEVEL=TRACE
export SGLANG_SET_CPU_AFFINITY=1
export SGLANG_ALLOW_OVERWRITE_LONGER_CONTEXT_LEN=1
export SGLANG_MOE_PADDING=1
export SGLANG_ENABLE_JIT_DEEPGEMM=1
export SGLANG_JIT_DEEPGEMM_COMPILE_WORKERS=32
export SGLANG_CHUNKED_PREFIX_CACHE_THRESHOLD=16384
export SGLANG_DEEPEP_NUM_MAX_DISPATCH_TOKENS_PER_RANK=64
export PYTHONUNBUFFERED=1
export UCX_TLS=tcp,sm
export UCX_NET_DEVICES=bond0
export METRICS_PREFIX=prefill
export SGLANG_HEALTH_CHECK_MEM_FREE=50
export ENABLE_PD_GUARDIAN=0
export SGLANG_DEEPEP_NORMAL_BF16_DISPATCH=1 \
    --model-path  /export/models/DeepSeek-R1-0528-W4AFP8 \
    --host 0.0.0.0 \
    --port 15000 \
    --trust-remote-code \
    --tp-size 8 \
    --dp-size 8 \
    --enable-dp-attention \
    --moe-dense-tp-size 1 \
    --enable-dp-lm-head \
    --ep-dispatch-algorithm dynamic \
    --eplb-algorithm deepseek \
    --disable-shared-experts-fusion \
    --moe-a2a-backend deepep \
    --moe-runner-backend cutlass \
    --deepep-mode low_latency \
    --mem-fraction-static 0.81 \
    --cuda-graph-max-bs 32 \
    --max-running-requests 256 \
    --context-length 16384 \
    --tokenizer-worker-num 8 \
    --enable-dynamic-batch-tokenizer \
    --dynamic-batch-tokenizer-batch-size 8 \
    --attention-backend flashinfer \
    --tool-call-parser deepseekv3 \
    --speculative-num-steps 1 \
    --speculative-eagle-topk 1 \
    --speculative-num-draft-tokens 2 \
    --disable-radix-cache \
    --page-size 128

sglang.bench_serving     --backend sglang     --host localhost     --port 15000    --dataset-path ShareGPT_V3_unfiltered_cleaned_split.json     --dataset-name sharegpt     --num-prompts 100     --sharegpt-output-len 256   --warmup-requests 10  --sharegpt-context-len 4096   --max-concurrency  64  --tokenizer /export/models/DeepSeek-R1-0528-W4AFP8  --request-rate 10  --model  /export/models/DeepSeek-R1-0528-W4AFP8

main:
image
this pr:
image

Throughput improved by approximately 10% after this pr.

## Checklist


python/sglang/srt/layers/moe/token_dispatcher/deepep.py
```diff
@@ -619,7 +617,7 @@ def _dispatch_core(
                 topk_ids,
                 self.num_max_dispatch_tokens_per_rank,
                 self.num_experts,
-                use_fp8=use_fp8,
+                use_fp8=True,
```
BBuf: I don't think `use_fp8` should be unconditional here. When `input_global_scale` is set we also pass `use_nvfp4=True`, so this ends up mixing two dispatch modes that used to be mutually exclusive. That looks risky for the NVFP4 path.

python/sglang/srt/environ.py
```diff
@@ -376,7 +376,7 @@ class Envs:
     SGLANG_CHUNKED_PREFIX_CACHE_THRESHOLD = EnvInt(8192)
 
     # DeepEP
-    SGLANG_DEEPEP_BF16_DISPATCH = EnvBool(False)
+    SGLANG_DEEPEP_NORMAL_BF16_DISPATCH = EnvBool(False)
```
BBuf: Renaming the env var here will break existing configs unless we keep a compatibility alias. There are still tests/docs using `SGLANG_DEEPEP_BF16_DISPATCH`, so normal-dispatch W4AFP8 will regress if this is the only change.

python/sglang/srt/layers/moe/ep_moe/kernels.py
```diff
@@ -1381,3 +1381,76 @@ def silu_and_mul_masked_post_per_tensor_quant_fwd(
...
+    k_offsets = pid_k * K_BLOCK_SIZE + tl.arange(0, K_BLOCK_SIZE)
+    scale_offsets = (k_offsets // K_SCALE_BLOCK_SIZE) * x_scale_stride2
+
+    x_ptrs = x_ptr + pid_e * m * k + k_offsets
+    output_ptrs = output_ptr + pid_e * m * k + k_offsets
+    x_scale_ptrs = x_scale_ptr + pid_e * x_scale_stride0 + scale_offsets
+
+    for tok_idx in tl.range(token_id, last_effective_id, pid_m_dim):
+        hidden = tl.load(x_ptrs + tok_idx * k).to(tl.float32)
+        scale_fp32 = tl.load(x_scale_ptrs + tok_idx * x_scale_stride1)
```
BBuf: This assumes `x_scale` is a regular fp scale. In the Blackwell path low-latency dispatch may hand back ue8m0/e8m0-style scales, so we probably need an explicit conversion here or a guard to keep that format out of this kernel？

BBuf (changes requested):
The direction makes sense, but I think this still needs a pass on compatibility and shared-path safety. In particular: the env var rename breaks existing normal-dispatch setups, `use_fp8=True` now overlaps with the NVFP4 path, and the new requant kernel does not seem to handle Blackwell scale formats yet.

python/sglang/srt/layers/moe/ep_moe/kernels.py
```diff
@@ -1381,3 +1381,76 @@ def silu_and_mul_masked_post_per_tensor_quant_fwd(
         NUM_STAGE=NUM_STAGES,
     )
     return output
+
+
+@triton.jit
```
BBuf, line 1386: It might be better to implement this kernel under the `jit_kernel` directory. We could also consider writing a simple CUDA kernel for it.

BBuf (approved):
Can you add more acc tests such as mmlu?

BBuf (approved):

PR #13824, opened 2025-11-24
[feat] update bucketed weights from distributed

ShawnY112358:
## Motivation

In the training-inference disaggregation RL scenario, we use update_weights_from_distributed to update SGLang's weights. However, update_weights_from_distributed currently does not support flattened tensors, resulting in significant time overhead during parameter synchronization.

## Modifications

support using accelerating update_weights_from_distributed with flattened tensor

## Benchmarking and Profiling

Qwen3 turbopp (30B) 5.5s -> 2.5s
Qwen3 Next (80B) 130s -> 7s

## Checklist

- [x] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [x] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.ai/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.ai/developer_guide/contribution_guide.html#benchmark-the-speed).
- [ ] Follow the SGLang code style [guidance](https://docs.sglang.ai/developer_guide/contribution_guide.html#code-style-guidance).
- [ ] Work with maintainers to merge your PR. See the [PR Merge Process](https://github.com/sgl-project/sglang/blob/main/.github/MAINTAINER.md#pull-request-merge-process)

PR #13125, opened 2025-11-12
support non-disturbing remote-instance-weight-loader

amysaq2023:
## Motivation

In https://github.com/sgl-project/sglang/pull/8215, SGLang has already supported a new load format: remote_instance, that allows new instance to load weights from another running instance. This approach can greatly improve weight loading time during instance initialization. However, since it use torch.distributed with NCCL as backend, it will disturb on-going inference requests: torch.distributed will always launch CUDA kernels for transferring weight tensors.

We come up with another backend option: TransferEngine, which will not disturbing any GPU workload and still, using RDMA to transfer weights.

## Modifications

We initialize one TransferEngine for each ModelRunner and will register its weights to RDMA channel during initialization.

When initializing a new instance who wants to use `remote_instance` load_format with TransferEngine backend:
1. It  will send an HTTP request to retrieve the source instance's TransferEngine metadata, including RDMA keys mapped to the corresponding GPU memory addresses.
2. Using these RDMA keys, the new instance directly loads weights from the source's GPU memory.

How to use:
`python -m sglang.launch_server [args] \`
`  --load-format remote_instance \`
`  --remote-instance-weight-loader-seed-instance-ip [seed_instance_ip] \`
`  --remote-instance-weight-loader-seed-instance-service-port [seed_instance_service_port] \`
`  --remote-instance-weight-loader-backend "transfer_engine"`

## Checklist

- [x] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [x] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [x] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.ai/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.ai/developer_guide/contribution_guide.html#benchmark-the-speed).

PR #14170, opened 2025-11-30
[Feature] Support p2p rdma weight transfer in RL training

Risc-lt:
[WIP] This is the **_test version_** of p2p rdma tranfer feature for review. More profiling and optimization will follow.

## Modifications

This PR introduces P2PTransferManager, a high-performance peer-to-peer weight synchronization system for RL training scenarios in SGLang. It enables efficient GPU-to-GPU direct weight transfers between training and rollout processes using RDMA, replacing the traditional PyTorch distributed broadcast approach with nccl backend.
Key Features

1. P2PTransferManager Architecture
   - Multi-engine design with a pool of MooncakeTransferEngine instances (configurable via SGLANG_TRANSFER_ENGINE_POOL_SIZE, defaults to 25% of CPU cores, max 8)
   - Task queue-based scheduling: automatically assigns transfer tasks to idle engines

image

2. Workflow
```
Training Process                 Rollout Process (P2PTransferManager)
     (GPU 0)                              (GPU 1)
        │                                     │
        │  1. Update weights                  │
        │     ┌─────────┐                     │
        ├────>│ Weights │                     │
        │     └─────────┘                     │
        │                                     │
        │  2. Register memory                 │   3. Submit transfer task
        │     via RDMA                        │     to task queue
        │                                     │         │
        │                                     │         ▼
        │                                     │   ┌──────────┐
        │                                     │   │ Scheduler│
        │                                     │   └──────────┘
        │                                     │         │
        │                                     │         ▼
        │                                     │   4. Assign to idle engine
        │                                     │      ┌─────────────────┐
        │                                     ├─────>│ Engine Pool     │
        │                                     │      │ [Eng0][Eng1]... │
        │                                     │      └─────────────────┘
        │                                     │         │
        │  5. ZMQ handshake                   │         ▼
        │     <───────────────────────────────┤   6. Register memory region
        │     (sync_status: ready)            │
        │                                     │   7. Release engine (early release)
        │                                     │      └──> Engine returns to pool
        │                                     │
        │  8. RDMA transfer                   │
        │     ──────────────────────────────> │   9. Wait for completion
        │     (direct GPU memory copy)        │
        │                                     │
        │  10. Send confirmation              │  11. Receive confirmation
        │     ──────────────────────────────> │
        │     (transfer_complete)             │
        │                                     │  12. Deregister memory
        │                                     │      Mark handle done
        │                                     ▼
```

## Integration Points
New server args: `--enable-p2p-transfer` and `--p2p-transfer-ib-device`
With `--enable-p2p-transfer=True`, sglang engine will automatically switche to P2P transfer. Otherwise, it will use the original `torch.distributed`. The flag `--p2p-transfer-ib-device` is used to specify the ib device for RDMA transfer. When it is nor specified, Mooncake transfer engine will automatcally search and find the suitable NICs.

## Test Coverage
Comprehensive test suite (test_p2p_transfer.py) with:
- Simple correctness tests (small tensor)
- Performance tests on Qwen3-32B model (64 layers, ~32B parameters)
- Dummy model tests for concurrency validation (1to1, 1to2, 2to1)

## Files Changed
New: python/sglang/srt/weight_sync/p2p_transfer.py (Core P2PTransferManager implementation)
New: test/srt/test_p2p_transfer.py (Comprehensive test suite)
Modified: python/sglang/srt/model_executor/model_runner.py (Integration with model weight update logic)
Modified: python/sglang/srt/server_args.py (Added P2P transfer configuration options)


Risc-lt:
Currently the pr only introduces the p2p recv side on sglang engine. E2E performance and P2P mapping will be added in a few days. Please have a look on it and welcome review. cc @zhaochenyang20 @stmatengss @jd-eth

stmatengss:
@Risc-lt Could you resolve the conflicts? So I can tag this PR to trigger CI.

python/sglang/srt/server_args.py
```diff
@@ -497,6 +497,10 @@ class ServerArgs:
     # FIXME: hack to reduce ITL when decode bs is small
     disaggregation_decode_polling_interval: int = 1
 
+    # P2P transfer settings: null by default
```
stmatengss: Move these two arguments to https://github.com/sgl-project/sglang/blob/879c72d9d8982b05d9829d4ae45c36b23d2df5d5/python/sglang/srt/server_args.py#L504. It will be clearer.

python/sglang/srt/weight_sync/p2p_transfer.py
```diff
@@ -0,0 +1,487 @@
...
+            ib_device: InfiniBand device name (optional)
+        """
+        self.hostname = hostname
+        self.gpu_id = gpu_id
+        self.ib_device = ib_device
+
+        # Initialize the engine pool
+        cpu_count = os.cpu_count() or 8
+        self.engine_pool_size = get_int_env_var(
+            "SGLANG_TRANSFER_ENGINE_POOL_SIZE",
```
stmatengss, line 95: Should the pool size be based on the TP size, or should we use alternative configurations?

python/sglang/srt/weight_sync/p2p_transfer.py
```diff
@@ -0,0 +1,487 @@
...
+import logging
+import os
+import threading
+import zmq
+
+from collections import defaultdict
+from queue import Queue
+from typing import Dict, Optional
+
+from sglang.srt.disaggregation.mooncake.transfer_engine import MooncakeTransferEngine
```
stmatengss, line 11: We might move transfer_engine.py to a common directory in the future.

test/srt/test_p2p_transfer.py
```diff
@@ -0,0 +1,1010 @@
+"""
```
stmatengss, line 1: Add this test to run_suite.py if it doesn't significantly increase runtime.

python/sglang/srt/weight_sync/p2p_transfer.py
```diff
@@ -0,0 +1,487 @@
...
+            context.term()
+            raise e
+
+    def get_session_id(self):
+        """
+        Get the session ID from the first transfer engine.
+        Note: In manager mode, each engine has its own session_id.
+        This returns the first one for compatibility.
+        """
+        return self.engine_pool[0].get_session_id()
```
stmatengss: It's unclear why engine pools are needed if each pool only has one engine.

python/sglang/srt/weight_sync/p2p_transfer.py
Risc-lt: We can expilictly set the env variable `SGLANG_TRANSFER_ENGINE_POOL_SIZE` for configuration. Maybe more detailed exploration on default engine pool size can be done after p2p mapping. cc @JD-ETH

python/sglang/srt/weight_sync/p2p_transfer.py
Risc-lt: Agree. Maybe not just for disaggregation.

test/srt/test_p2p_transfer.py
Risc-lt: Okay, I'll add it. FYI, this test may take about 1 min to complete.

python/sglang/srt/weight_sync/p2p_transfer.py
Risc-lt: Oh, this is one deprecated function that I forgot to remove. Sry for mistake.

python/sglang/srt/weight_sync/p2p_transfer.py
```diff
@@ -0,0 +1,479 @@
...
+            if ptr not in self.registered_ptrs[engine_idx][session_id]:
+                logger.debug(
+                    f"Registering memory region on engine {engine_idx} for session {session_id}: "
+                    f"ptr={ptr:#x}, length={length}"
+                )
+                self.engine_pool[engine_idx].register(ptr, length)
+                self.registered_ptrs[engine_idx][session_id].add(ptr)
+            else:
+                logger.debug(
+                    f"Memory region already registered on engine {engine_idx} for session {session_id}: ptr={ptr:#x}"
```
JD-ETH, line 314: when would this happen?

python/sglang/srt/weight_sync/p2p_transfer.py
```diff
@@ -0,0 +1,487 @@
...
+                # Block until a task is available
+                task_info = self.task_queue.get()
+
+                # Wait for an idle engine
+                engine_idx = None
+                while engine_idx is None:
+                    engine_idx = self._get_engine_index()
+                    if engine_idx is None:
+                        # All engines busy, wait a bit before retrying
+                        threading.Event().wait(0.01)
```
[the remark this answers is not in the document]
JD-ETH: this sounds fair, no?

python/sglang/srt/weight_sync/p2p_transfer.py
```diff
@@ -0,0 +1,479 @@
...
+        """
+        # Create a temporary socket
+        context = zmq.Context()
+        socket = context.socket(zmq.DEALER)
+        remote_ip, remote_port = training_p2p_session_id.split(":")[0], int(training_p2p_session_id.split(":")[1])
+        socket.connect(format_tcp_address(remote_ip, remote_port))
+
+        # Cache the socket for cleanup
+        with socket_lock:
+            socket_cache[task_id] = {
```
JD-ETH, line 370: indeed I'm not seeing the value of the cache here. Aren't they independent? Or could the task_id overlap?

test/srt/test_p2p_transfer.py
```diff
@@ -0,0 +1,1010 @@
...
+            ptr = tensor.data_ptr()
+            length = tensor.numel() * tensor.element_size()
+
+            if name in self.weight_buffers:
+                existing_ptr = self.weight_buffers[name]['ptr']
+                if existing_ptr == ptr:
+                    self.weight_buffers[name]['tensor'] = tensor
+                    continue
+
+            self.engine.register(ptr, length)
```
JD-ETH, line 313: I see you call this method twice. Would engine reuse its cache and skip the registration?

test/srt/test_p2p_transfer.py
```diff
@@ -0,0 +1,1010 @@
...
+                    logger.error(f"[TrainingSender] ZMQ error: {e}")
+                break
+            except Exception as e:
+                logger.error(f"[TrainingSender] Error: {e}", exc_info=True)
+
+    def _handle_sync_status(self, identity: bytes, message: dict):
+        """Handle sync_status and perform RDMA transfer."""
+        rollout_transfer_session_id = message.get("transfer_session_id", "")
+        rollout_ptr = message.get("ptr", 0)
+        rollout_length = message.get("length", 0)
```
JD-ETH, line 377: `rollout_msg_length` maybe?

test/srt/test_p2p_transfer.py
```diff
@@ -0,0 +1,1010 @@
...
+        task_id = message.get("task_id", "")
+
+        try:
+            # Find the weight with matching length
+            weight_info = None
+            weight_name = None
+            for name, info in self.weight_buffers.items():
+                if info['length'] == rollout_length:
```
JD-ETH, line 385: can you not match by name? You will easily mismatch with the qwen3 full model here, no?

test/srt/test_p2p_transfer.py
```diff
@@ -0,0 +1,1010 @@
...
+                weight += torch.randn_like(weight) * 0.01
+            torch.cuda.synchronize()
+
+            update_end = time.time()
+            logger.info(
+                f"[Training] Update {update_idx + 1}/{num_updates} completed in "
+                f"{(update_end - update_start)*1000:.2f}ms"
+            )
+
+            sender.register_weights(weights)
```
JD-ETH, line 474: why do you repeatedly register weights actually?

test/srt/test_p2p_transfer.py
```diff
@@ -0,0 +1,1010 @@
...
+            f"[Rollout-{rank}] All updates completed. Avg time: {avg_total_time*1000:.2f}ms, "
+            f"Avg bandwidth: {avg_bandwidth:.2f}Gbps, Sources: {num_training}"
+        )
+
+    except Exception as e:
+        logger.error(f"[Rollout-{rank}] Error: {e}", exc_info=True)
+        result_queue.put((f"rollout_{rank}_error", str(e)))
+
+
+class TestP2PTransferPerformance(unittest.TestCase):
```
JD-ETH, line 798: can we also test for correctness? Weights should be the same across train and rollout? Setting a fixed seed to know the expected weight delta?

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -1173,19 +1204,32 @@ def destroy_weights_update_group(self, group_name):
...
 
-    def update_weights_from_distributed(
-        self,
-        names,
-        dtypes,
-        shapes,
-        group_name,
-        load_format: Optional[str] = None,
-    ):
+    def init_p2p_transfer_engine(self):
```
JD-ETH, line 1142: hmm where is this called?

JD-ETH (changes requested):
Solid work, and lots of code! 

However I'd like to bring some attention to this: 
https://github.com/sgl-project/sglang/pull/13125
Where it will introduce a **singular engine** for each model rank. 

Things we could learn from that: 

1. moving the TransferEngine instantiation and weight registration at SGlang initialization time. We are not going to change a model memory layout once registered. 
2. memory registration itself is the bottleneck for real world transfer load. So why do we deregister memory after every transfer? In my mind, we should change two aspects of your current design 

1. Memory registration at initialization, once and for all 
2. Prioritize cached memory when selecting engine, not current availability 


I'm yet a bit skeptical about whether having more engine instances helps real world load, but let's profile and find out. Additionally, as we know flattening can reduce overhead: https://github.com/sgl-project/sglang/pull/13824, submitting each tensor as individual task may not make sense anyways. 

Could we make sure the following 
- the weight update is also correct, in addition to the performance tests right now 
- the single instance version of P2PManager is well supported and tested also 

for us to profile later?

PR #14201, opened 2025-12-01
[scheduler] enhance scheduler in dp_attention mixed case with spec

liupeng374:
## Motivation

In the scenario of dp_attention mixed configuration with MTP, due to the prefill priority, when all DP ranks reach full batch capacity under full load, and after one DP rank requests to complete and exit, a prefill request is then initiated. This causes the entire mixed configuration to enter a prefill/idle mode, where all dp rank except the one with the request remain in idle mode. I found that this situation has a significant impact on throughput and TPOT.

## Modifications

I tried to allow new prefill requests only when all DP ranks have an available slot, which would completely avoid the prefill/idle scenario. To prevent a situation where one DP rank remains fully loaded but does not exit, I added a condition to check for a stable state. This ensures that even if multiple scheduling attempts do not result in all DP ranks being fully loaded, prefill requests are still prioritized, thus avoiding excessively long TTFT (Time to First Transaction) in extreme scenarios. enable with **export SGLANG_SCHEDULER_DECREASE_PREFILL_IDLE=1**

image

## Benchmarking and Profiling
before
image
enable enhance
image


python/sglang/srt/managers/scheduler.py
```diff
@@ -209,6 +209,64 @@ class EmbeddingBatchResult:
...
+            group=self.cpu_group,
+        )
+        tp0_info = self.global_batch_size[:, 0, :]
+        return tp0_info
+
+    def get_schedule_decision(self, running_batch):
+        tp0_info = self.get_schedule_info(running_batch)
+        if (
+            int(tp0_info[:, 0].min().item()) < self.max_running_requests
+            and int(tp0_info[:, 0].max().item()) == self.max_running_requests
```
merrymercy: This heuristic is not good. The heuristic should look at the waiting queue instead of the current running batch.

python/sglang/srt/managers/scheduler.py
```diff
@@ -209,6 +209,64 @@ class EmbeddingBatchResult:
     embeddings: torch.Tensor
 
 
+class SchedulerEnhancer:
```
merrymercy: move this into a separate file.

merrymercy (changes requested):
There is another decision: put the queue on scheduler or put the queue on LB (sgl-router or data_parallel_controller.py).

The current code puts the queue on scheduler. For example, if there are 4 DP groups, the scheduler waits and prefers to run 4 new requests in a forward batch.

We can also put the queue on LB. For example, if there are 4 DP groups, then the LB only sends 4 requests as a group, each DP group will get one and run them according to the prefill-first policy. 

But I think the current code seems better.

python/sglang/srt/managers/scheduler.py
```diff
@@ -1745,6 +1756,11 @@ def get_num_allocatable_reqs(self, running_bs):
         return res
 
     def get_new_batch_prefill(self) -> Optional[ScheduleBatch]:
+        if self.schedule_enhancer and not self.schedule_enhancer.get_schedule_decision(
```
merrymercy, line 1759: 1. Do not call it scheduler enhander. Currently, it only delays prefill requests. `schedule_enhancer` is a name that is too general for its usage.
2. Maybe move later to merge the condition with "# Handle the cases where prefill is not allowed"

python/sglang/srt/managers/scheduler_enhancer.py
```diff
@@ -0,0 +1,59 @@
...
+            local_batch_size,
+            group=self.cpu_group,
+        )
+        tp0_info = self.global_batch_size[:, 0, :]
+        return tp0_info
+
+    def get_schedule_decision(self, running_batch):
+        tp0_info = self.get_schedule_info(running_batch)
+        if (
+            int(tp0_info[:, 0].min().item()) < self.max_running_requests
```
merrymercy, line 52: This heuristic is not good. The heuristic should look at the waiting queue instead of the current running batch.

PR #14208, opened 2025-12-01
feat(EvictionStrategy): Add a new adaptive mixed cache eviction strategy to improving cache hit rate

CLFutureX:
Add a new adaptive mixed cache eviction strategy called `AdapterMixStrategy`, which combines **hot/cold data separation** and **dynamic threshold adjustment** to fix the limitations of single LRU/LFU strategies, while improving cache hit rate and resource utilization.

**Limitations of Traditional Strategies:**
-**LRU Issue**: May evict frequently accessed cache data (due to overemphasizing recency over frequency).
-**LFU Issue**: May retain stale "zombie" high-frequency data (long-term unused but historically frequent) while evicting recently accessed items, leading to inefficient resource occupation.

### Initial Design
1. **Hot/Cold Data Separation**: Use a time threshold to split cache entries into two categories:
   - Hot data: Accessed recently;
   - Cold data: Not accessed for a long time.
2. **Differentiated Priority Sorting**:
   - Hot data: Sorted by "LFU (access count) + LRU (last access time)";
   - Cold data: Sorted by LRU;
   - Rule: Cold data is evicted first.

### Dynamic Threshold Adaptation (Simplifies Threshold Configuration)
To solve the problem that "time threshold requires manual configuration and is hard to evaluate", we designed a dynamic threshold adjustment logic:
- Adjust the time threshold based on the access ratio of hot/cold data to adapt to changing access patterns:
  - Shrink the threshold if there’s too much cold data (to classify more data as hot);
  - Expand the threshold if there’s too much hot data (to classify more data as cold);
- Default update interval: 60 seconds (to avoid frequent adjustments).

### Minimal Computational Overhead
1. Updates only once every 60 seconds—adds almost no extra computation;
2. Embeds the update logic into the existing `get_priority` method:
   - Users barely notice any changes;
   - No additional complexity for upper-layer code or callers.

### Benefits
- **Compared to LRU**: Prioritizes keeping frequently accessed data, reducing cases where "high-frequency data is mistakenly evicted";
- **Compared to LFU **: Eliminates "zombie high-frequency data" (data with high past access counts but no recent access) to avoid invalid cache occupying resources long-term;
 
With simple calculations, it fixes the drawbacks of single LRU/LFU strategies, while boosting cache hit rate and making better use of resources.

---
### CI States

Latest PR Test (Base): :x: [Run #20118663619](https://github.com/sgl-project/sglang/actions/runs/20118663619)
Latest PR Test (Extra): :warning: **Not enabled** -- add `run-ci-extra` label to opt in.


stmatengss:
@CLFutureX Sorry for the late reply. Will review it tomorrow!

PR #14209, opened 2025-12-01
[BUGFIX] fix edge case for qwen3-next

yizhang2077:
## Motivation

1. RL will offload memory and then call `clear`, it will cause IMA if call `zeros_` in `clear`, so we move `zeros_` in `alloc`.
2. fix repeated `release_kv_cache` when abort request in decode node.
## Checklist


ispobock (approved):

python/sglang/srt/mem_cache/memory_pool.py
```diff
@@ -266,23 +266,19 @@ def alloc(self, need_size: int) -> Optional[torch.Tensor]:
...
-        for i in range(len(self.mamba_cache.conv)):
-            self.mamba_cache.conv[i][:, free_index] = 0
-        self.mamba_cache.temporal[:, free_index] = 0
 
     def clear(self):
-        # Zero the entire mamba cache before resetting free_slots
-        # This ensures that when slots are reallocated, they start with clean state
-        for i in range(len(self.mamba_cache.conv)):
-            self.mamba_cache.conv[i].zero_()
-        self.mamba_cache.temporal.zero_()
```
hanming-lu, line 284: I'd recommend keeping this logic even though it may be repetitive. It is on par with the initial state that the pool was initialized with

hanming-lu (approved):

python/sglang/srt/mem_cache/memory_pool.py
yizhang2077: In rl colocate mode, it will offload memory firstly and then call `clear`. It will cause IMA if we call `zero_` in this case. that's why I want to move `zero_` logic into alloc

PR #14214, opened 2025-12-01
Add cuda event based on waiting value

hnyls2002:
For the scenarios where the record happens later than the wait.

```python
stream_a = torch.cuda.Stream()
stream_b = torch.cuda.Stream()

with torch.cuda.stream(stream_a):
    event.wait()

with torch.cuda.stream(stream_b):
    event.record()
```

### DIFF
--- a/python/sglang/jit_kernel/cuda_wait_value.py
+++ b/python/sglang/jit_kernel/cuda_wait_value.py
@@ -1,0 +1,64 @@
+from __future__ import annotations
+
+from functools import lru_cache
+from typing import TYPE_CHECKING
+
+import torch
+
+from sglang.jit_kernel.utils import load_jit
+
+if TYPE_CHECKING:
+    import torch
+    from tvm_ffi.module import Module
+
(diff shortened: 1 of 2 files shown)


python/sglang/jit_kernel/cuda_wait_value.py
```diff
@@ -0,0 +1,79 @@
...
+    module = _jit_stream_wait_value_module()
+    module.stream_wait_value(flag, value)
+
+
+class Event:
+    def __init__(self) -> None:
+        self.flag = torch.zeros(1, dtype=torch.int32, device="cuda")
+
+    def record(self, value: int = 1) -> None:
+        self.flag[0] = value
```
merrymercy, line 34: This seems blocking on host. We can use `torch.fill` to make it non blocking.

hnyls2002: Yes, I will fix it in a following PR.

PR #13965, opened 2025-11-26
[Piecewise] support disable decode cuda graph when enable piecewise cuda graph

hebiao064:
## Checklist

- [ ] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.ai/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.ai/developer_guide/contribution_guide.html#benchmark-the-speed).
- [ ] Follow the SGLang code style [guidance](https://docs.sglang.ai/developer_guide/contribution_guide.html#code-style-guidance).
- [ ] Work with maintainers to merge your PR. See the [PR Merge Process](https://github.com/sgl-project/sglang/blob/main/.github/MAINTAINER.md#pull-request-merge-process)

PR #14248, opened 2025-12-01
Fix piecewise CUDA graph for hybrid SWA models (Qwen3-Next)

alisonshao:
## Summary
- Fix `out_cache_loc_swa` condition in `replay_prepare` to check `self.out_cache_loc_swa` instead of `forward_batch.out_cache_loc_swa`
- This fixes TestQwen3NextPiecewiseCudaGraph test failure introduced in PR #13965

## Problem
PR #13965 introduced a bug where the condition for setting `out_cache_loc_swa` in `replay_prepare` was checking `forward_batch.out_cache_loc_swa` instead of `self.out_cache_loc_swa`. 

For hybrid SWA models like Qwen3-Next:
- `self.out_cache_loc_swa` is initialized (because `is_hybrid_swa=True`)
- But `forward_batch.out_cache_loc_swa` may be None
- This caused `out_cache_loc_swa` to be None in the static forward batch even when valid data existed

**Before fix:** TestQwen3NextPiecewiseCudaGraph had accuracy=0%, invalid=100%
**After fix:** TestQwen3NextPiecewiseCudaGraph has accuracy=95.5%, invalid=0%

## Test plan
- [x] Tested manually on B200 GPUs with the TestQwen3NextPiecewiseCudaGraph test
- [ ] CI tests should pass

### DIFF
--- a/python/sglang/srt/model_executor/piecewise_cuda_graph_runner.py
+++ b/python/sglang/srt/model_executor/piecewise_cuda_graph_runner.py
@@ -521,7 +521,7 @@ class PiecewiseCudaGraphRunner:
 
         out_cache_loc_swa = (
             self.out_cache_loc_swa[:static_num_tokens]
-            if forward_batch.out_cache_loc_swa is not None
+            if self.out_cache_loc_swa is not None
             else None
         )
 


alisonshao:
> QWen3-Next is not a SWA model, and I cannot reproduce the error you mentioned in your PR
> 
> Did you happen to launched a SWA Model?

You're right. We are trying to fix this error that keeps occurring: https://github.com/sgl-project/sglang/actions/runs/19782623297/job/56697338849

PR #14249, opened 2025-12-01
feat: DeepSeek new v3.2 encoding

Eva20150932-atlascloud:
## Motivation

https://github.com/sgl-project/sglang/issues/14227
DeepSeek official release a new encoding func to replace chat_template, and I made one workable version(though it is hard-coded and breaks other models)

if you still use old chat-template for the formal v3.2, then it works bad for tool_calling. so we need new encoding to run new v3.2 model.
## Accuracy Tests

start a server like `python3 -m sglang.launch_server --model-path deepseek-ai/DeepSeek-V3.2 --trust-remote-code --tp-size 8 --host 0.0.0.0 --tool-call-parser deepseekv32 --enable-metrics --max-queued-requests 3 --max-running-requests 64 --cuda-graph-max-bs 64 --reasoning-parser deepseek-v3` and my test for tool_calling passed.

## Checklist

- [*] Format your code according to the [Format code with pre-commit](https://docs.sglang.io/developer_guide/contribution_guide.html#format-code-with-pre-commit).


Johnsonms:
Verified it works

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -305,57 +317,82 @@ def _apply_jinja_template(
...
+        if self.use_dpsk_v32_encoding:
+            if request.chat_template_kwargs and request.chat_template_kwargs.get(
+                "thinking"
+            ):
+                thinking_mode = "thinking"
+            else:
+                thinking_mode = "chat"
+            messages = request.messages
+            messages = [msg.model_dump() for msg in messages]
+            if messages[0]["role"] != "system":
```
yudian0504, line 294: why a default system role here ?

JustinTong0323: Match here I think https://huggingface.co/deepseek-ai/DeepSeek-V3.2/blob/main/encoding/test_output.txt

yudian0504: The system role in this test_output is included in the message from test_input: https://huggingface.co/deepseek-ai/DeepSeek-V3.2/blob/main/encoding/test_input.json#L43

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -305,57 +317,82 @@ def _apply_jinja_template(
...
+                messages.insert(
+                    0, {"role": "system", "content": "You are a helpful Assistant."}
+                )
+            if request.tools:
+                messages[0]["tools"] = [tool.model_dump() for tool in request.tools]
+            real_input = encode_messages(
+                messages, thinking_mode=thinking_mode, drop_thinking=False
```
yudian0504, line 301: Based on the default values in the official code, the default behavior of drop_thinking should be **True**, which means that in thinking_mode, the previous think content will be dropped (this is also consistent with the historical behavior of the ds series).

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -12,6 +12,7 @@
 from fastapi.responses import ORJSONResponse, StreamingResponse
 from jsonschema import Draft202012Validator, SchemaError
 
+from sglang.srt.entrypoints.openai.encoding_dsv32 import encode_messages
```
yudian0504, line 15: Instead of directly copying the code into sglang, can we reference the code from the model directory (the local repo modification I made 😂)?

```python
import sys
import os
sys.path.append(os.path.join(model_path, 'encoding'))
from encoding_dsv32 import encode_messages
```

JustinTong0323: actually I prefer copying (maybe could avoid some env problem, but just my personal preference

python/sglang/srt/entrypoints/openai/serving_chat.py
JustinTong0323: Yes it was that before, but I noticed here: https://api-docs.deepseek.com/news/news251201 The thinking is involved in history

JustinTong0323: Oh I find this explanation: https://api-docs.deepseek.com/guides/thinking_mode#tool-calls, the reasoning content would be removed after each user input..

python/sglang/srt/entrypoints/openai/serving_chat.py
JustinTong0323: You are right... Let me check

python/sglang/srt/entrypoints/openai/serving_chat.py
yudian0504: right..Only ongoing tool-calls need to retain the thinking section, my understanding..

python/sglang/srt/entrypoints/openai/serving_chat.py
yudian0504: make sense, at least copying here can prevent some non-standard model directories, such as those for fine-tuned ds3.2.....

python/sglang/srt/entrypoints/openai/serving_chat.py
```diff
@@ -305,57 +317,82 @@ def _apply_jinja_template(
...
-                openai_compatible_messages,
-                tokenize=True,
-                add_generation_prompt=True,
-                tools=tools,
-                reasoning_effort=request.reasoning_effort,
-                **(
-                    request.chat_template_kwargs if request.chat_template_kwargs else {}
-                ),
-                return_dict=False,
+        if self.use_dpsk_v32_encoding:
```
yudian0504, line 285: The code above 'orjson.loads' for function.arguments, unnessary for ds v3.2
```python
            if (
                processed_msg["role"] == "assistant"
                and "tool_calls" in processed_msg
                and isinstance(processed_msg["tool_calls"], list)
            ):
                for item in processed_msg["tool_calls"]:
                    if "arguments" in item["function"] and isinstance(
                        item["function"]["arguments"], str
                    ):
                        item["function"]["arguments"] = orjson.loads(
                            item["function"]["arguments"]
                        )
```

Ref: 

https://huggingface.co/deepseek-ai/DeepSeek-V3.2/blob/main/encoding/test_input.json#L86

python/sglang/srt/entrypoints/openai/serving_chat.py
JustinTong0323: I just go through the encoding script and I think the logic has already processed by the script provided, so we shouldn't set drop_thinking to true.

python/sglang/srt/entrypoints/openai/serving_chat.py
Eva20150932: I tested. When first role is not system, the tools are ignored. 
 when there is an (added) system msg, `real_input` when tool-calling is like 
 ```
 real_input: <｜begin▁of▁sentence｜>You are a helpful Assistant.

## Tools
You have access to a set of tools you can use to answer the user's question.
You can invoke functions by writing a "<｜DSML｜function_calls>" block like the following as part of your reply to the user:
<｜DSML｜function_calls>
<｜DSML｜invoke name="$FUNCTION_NAME">
<｜DSML｜parameter name="$PARAMETER_NAME" string="true|false">$PARAMETER_VALUE</｜DSML｜parameter>
...
</｜DSML｜invoke>
<｜DSML｜invoke name="$FUNCTION_NAME2">
...
</｜DSML｜invoke>
</｜DSML｜function_calls>
String and scalar parameters should be specified as is without any escaping or quotes, while lists and objects should use JSON format. The "string" attribute should be set to "true" for string type parameters and "false" for other types (numbers, booleans, arrays, objects).
If the thinking_mode is enabled, then after function results you should strongly consider outputting a thinking block. Here is an example:
<｜DSML｜function_calls>
...
</｜DSML｜function_calls>
<function_results>
...
</function_results>
<think>...thinking about results</think>
Here are the functions available in JSONSchema format:
<functions>
{"description": "Get the current date and time", "name": "get_datetime", "parameters": {"type": "object", "properties": {"timezone": {"type": "string", "description": "The timezone, e.g. Asia/Shanghai, UTC"}}, "required": ["timezone"]}, "strict": false}
{"description": "Get the weather for a specific date and location", "name": "get_weather", "parameters": {"type": "object", "properties": {"location": {"type": "string", "description": "The city name, e.g. Beijing, Hangzhou"}, "date": {"type": "string", "description": "The date in YYYY-MM-DD format"}}, "required": ["location", "date"]}, "strict": false}
</functions>
<｜User｜>hello<｜Assistant｜><think>
```
 when there is no (added) system msg, `real_input` when tool-calling is like 
 ```
 real_input: <｜begin▁of▁sentence｜><｜User｜>hello<｜Assistant｜><think>
 ```

Comment (yudian0504) [reply python/sglang/srt/entrypoints/openai/serving_chat.py:294]: yes..I found that 'render_tools' function only be called for system or developer role..

Maybe default system role should be added when request.tools is not None?

https://huggingface.co/deepseek-ai/DeepSeek-V3.2/blob/main/encoding/encoding_dsv32.py#L160

Comment (JustinTong0323) [reply python/sglang/srt/entrypoints/openai/serving_chat.py:294]: It's not harmful to add a default system msg here I think, so let's keep it

File: python/sglang/srt/entrypoints/openai/serving_chat.py (hunk 4 above)
Comment (JustinTong0323) [reply python/sglang/srt/entrypoints/openai/serving_chat.py:285]: You are right, just refactored

File: python/sglang/srt/entrypoints/openai/serving_chat.py (hunk 2 above)
Comment (yudian0504) [reply python/sglang/srt/entrypoints/openai/serving_chat.py:301]: After checking again, it seems that setting drop_thinking=False is more intuitive, because the handling of reasoning_content in render_message is already quite thorough. For normal messages input, whether drop_thinking is True or False does not affect the final prompt.

File: python/sglang/srt/entrypoints/openai/serving_chat.py [hunk 5]
```diff
@@ -82,6 +83,17 @@ def __init__(
...
+
+    def _use_dpsk_v32_encoding(self) -> bool:
+        has_chat_template = (
+            self.tokenizer_manager.tokenizer is not None
+            and self.tokenizer_manager.tokenizer.chat_template is not None
+        )
+        architectures = self.tokenizer_manager.server_args.get_hf_config().architectures
+        is_dpsk_v32 = "DeepseekV3" in architectures[0] if architectures else False
+        return not has_chat_template and is_dpsk_v32
+
```
jimmy-evo, line 96:
```
self.use_dpsk_v32_encoding = self.tokenizer_manager.server_args.tool_call_parser == "deepseekv32"
```
just dont determine with "architectures", but "tool_call_parser"

jimmy-evo (changes requested):

python/sglang/srt/entrypoints/openai/serving_chat.py
JustinTong0323: should not as the tool_call_parser is not necessary in some cases, but this code path matters

python/sglang/srt/function_call/deepseekv32_detector.py
```diff
@@ -0,0 +1,321 @@
...
+
+                # Check if tool call is complete (has closing tag)
+                if is_tool_end:
+                    # Only emit the tool call when it's complete (saw </｜DSML｜invoke>)
+                    # This ensures each function returns at most once
+                    calls_for_this_invoke: list[ToolCallItem] = []
+
+                    # Check if invoke_content is empty or whitespace only
+                    # If so, skip this tool call entirely (it's likely incomplete or malformed)
+                    if not invoke_content.strip():
```
liaol, line 246: Here will ignore the non-parameter function. 
Like :
```
<｜DSML｜invoke name="get_current_time">

</｜DSML｜invoke>
```

python/sglang/srt/entrypoints/openai/serving_chat.py
jimmy-evo: we can just add an environ `SGLANG_USE_DPSKV32_ENCODING=True`, then no need to concern how to determine this

jimmy-evo: @JustinTong0323 i think this custom encode is a temp way. actually it is a kind of `def apply_chat_template`

Fridge003 (approved):

python/sglang/srt/entrypoints/openai/serving_chat.py
JustinTong0323: Not quite sure, do you mean we should not default it? But this code is adapted from deepseek's hf repo so I think it should be enabled by default

python/sglang/srt/function_call/deepseekv32_detector.py
JustinTong0323: Would fix that later.

python/sglang/srt/entrypoints/openai/serving_chat.py
jimmy-evo: > Not quite sure, do you mean we should not default it? But this code is adapted from deepseek's hf repo so I think it should be enabled by default

do you remember  when huggingface transforers~=4.2x (2023/2024), open source models usually provide a `tokenizer.py` with `def apply_chat_template`.

this `encoding_dsv32.py` is that `def apply_chat_template`.

Muqi1029:
@Eva20150932-atlascloud Thanks for you answering! I think maybe you are right, `while loop` should be kept!

But here I have met another question, why you use these markers here?

https://github.com/sgl-project/sglang/blob/ef3f8c97e180155e29c8a420ec8156974abf7bac/python/sglang/srt/function_call/deepseekv32_detector.py#L185-L201

I think model output in the token level, you can use the following scripts to see the tokens:
```py

tokenizer = AutoTokenizer.from_pretrained("deepseek-ai/DeepSeek-V3.2")

special_tokens = [
    "<｜DSML｜function_calls>",
    "</｜DSML｜function_calls>",
    "<｜DSML｜invoke",
    "</｜DSML｜invoke",
]

for tokens in special_tokens:
    print("\n\n")
    print(f" Processing {tokens} ".center(80, "-"))
    ids = tokenizer.encode(tokens, add_special_tokens=False)
    for id in ids:
        tokens = tokenizer.decode(id)
        print(f"'{tokens}' : {id}")

```
The output is as follows:
```

---------------------- Processing <｜DSML｜function_calls> -----------------------
'<' : 30
'｜DSML｜' : 128793
'function' : 8701
'_c' : 4941
'alls' : 12548
'>' : 32



---------------------- Processing </｜DSML｜function_calls> ----------------------
'</' : 1718
'｜DSML｜' : 128793
'function' : 8701
'_c' : 4941
'alls' : 12548
'>' : 32



--------------------------- Processing <｜DSML｜invoke ---------------------------
'<' : 30
'｜DSML｜' : 128793
'inv' : 40148
'oke' : 5406



-------------------------- Processing </｜DSML｜invoke ---------------------------
'</' : 1718
'｜DSML｜' : 128793
'inv' : 40148
'oke' : 5406
```

So `<｜` will not be generated forever. Right?

jxz542189:
@Eva20150932-atlascloud Can you ensure that the function calls are output in the expected streaming manner? https://github.com/sgl-project/sglang/issues/14711

whybeyoung:
When using smg and grpc mode, i think it  should do similar thing with this pr @slin1237

PR #14257, opened 2025-12-02
sync attention doc and ep doc to doctree

b8zhong:
Previously it was not included...
Screenshot 2025-12-01 at 5 18 15 PM

include it + sync attn document at the same time.

I had the change tht title of server args md bc, it was linking to each other (the titles were the same, so it was doing this weird appearing/disappearing effect, anyway, the server args are not only linked to EP itself anyway.


Fridge003 (approved):
LGTM

PR #14278, opened 2025-12-02
Revert PR #14044: Restore separate memory pool for piecewise CUDA graph

alisonshao:
## Summary
This PR reverts PR #14044 to test if it resolves the `TestQwen3NextPiecewiseCudaGraph` failure on H100 (accuracy=0.0, invalid=1.0).

**Hypothesis**: PR #14044 may have caused the issue by:
1. Sharing memory pool between piecewise and main cuda graph runners (instead of separate pools)
2. A potential bug in `set_graph_pool_id` where `if _graph_pool_id is not None` may prevent the variable from being set

This revert restores:
- Unconditional assignment in `set_graph_pool_id`
- Separate memory pool for piecewise cuda graph runner

## Test plan
- [ ] Verify if `TestQwen3NextPiecewiseCudaGraph` passes on H100 CI with this revert
- [ ] If it passes, PR #14044 is confirmed as the cause
- [ ] If it fails, need to continue bisecting other commits


python/sglang/srt/models/qwen3_next.py
```diff
@@ -377,7 +377,7 @@ def forward(
         hidden_states: torch.Tensor,
         forward_batch: ForwardBatch,
     ):
-        output = torch.empty_like(hidden_states)
+        output = torch.zeros_like(hidden_states)
         if forward_batch.forward_mode.is_extend() and get_forward_context() is not None:
```
Chen-0210: We can just move the `zeros_like` to this and copy the exact values into the output. This don't need to  modify the `o = q.new_empty` allocations in FLA and can avoids causing the overhead in the non–piecewise CUDA graph.

Fridge003 (approved):

PR #14293, opened 2025-12-02
[Feature]Report health status while worker exit

ZzuMrW:
When a worker service process exits, the health status interface is called to change the worker's health to False, maybe can also set the health to True.

## Checklist

### DIFF
--- a/python/sglang/srt/entrypoints/engine.py
+++ b/python/sglang/srt/entrypoints/engine.py
@@ -82,6 +82,7 @@ from sglang.srt.utils import (
     set_ulimit,
 )
 from sglang.srt.utils.torch_memory_saver_adapter import TorchMemorySaverAdapter
+from sglang.srt.utils.worker_exit_reporter import worker_exit_reporter
 from sglang.version import __version__
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


slin1237:
> > I'm not 100% following why we need this As of today, if worker is removed, dead, or over loaded, the health check will automatically become false gateway runs a async health check on the background Is there a specific issue you saw in current health check, or additional things you need
> 
> This is a requirement from Ascend. They might organize to merge it later. Sometimes, sub-processes (e.g., sglang::scheduler, sglang::detokenizer) are killed, but the HTTP server is still providing services externally. The router will continue to route requests to these processes, and the current request may be suspended. Under default conditions, it takes 180 seconds for the router to mark the worker as unhealthy. Ascend wants to quickly detect this failure and prevent routing new requests or retries to this worker.
> 
> 

This doesn't answer my question?

slin1237 (changes requested):
as commented

hnyls2002:
Hi @ZzuMrW, following up on the review discussion: the maintainer feedback was that worker exit should be handled on the SGLang worker side or via the gateway's existing health checks and circuit breaker, rather than a worker-to-router callback that creates a circular dependency, and the PR has been inactive since then. The touched sgl-router code has also been replaced by sgl-model-gateway. Closing - please reopen if I've missed something.

PR #14294, opened 2025-12-02
[scheduler] remove scheduler allgather for best throughout

liupeng374:
## Motivation

In the scheduler, when dp_attention is enabled, it introduces a global CPU allgather communication. This communication takes a long time when the cluster is large, and in scenarios with spec-overlap, the computation on the card side using draft_extend cannot fully mask this delay. To resolve this issue, additional hardware costs would be incurred, which is not cost-effective. However, in scenarios where padding is not required (i.e., when require_mlp_tp_gather is False), there is no need to use global_num_token for padding. In such cases, the allgather operation only serves to prevent all dp_ranks from being idle_batch. In some scenarios, the allgather operation on the CPU side can be very time-consuming, skip the all gather ops can actually yield better performance and throughput.

## Modifications

In the `scheduler_dp_attn_mixin.py` file, introduce the environment variable `SGLANG_SCHEDULER_SKIP_ALL_GATHER` by setting it to `export SGLANG_SCHEDULER_SKIP_ALL_GATHER=1`, which will allow you to skip the all_gather scheduling.

## Benchmarking and Profiling

image

## Checklist


iforgetmyname (approved):

ByronHsu:
this is amazing! I met similar issue and got around by using NCCL

PR #14297, opened 2025-12-02
[Diffusion]: support longcat video model

attack204:
## Motivation

support LongCatVideoModel

This implementation is heavily inspired by the FastVideo project, with the majority of the code in this PR adapted from it.

- example

```bash
sglang generate --model-path meituan-longcat/LongCat-Video \
  --prompt "A Logo With Bold Large Text: SGL Diffusion" 
  --save-output   > server.log 2>&1
```

https://github.com/user-attachments/assets/cf27b7cb-a644-4b2b-b881-92a107569ca6

- Limitation

1. The existing diffusion model in Longcat's Huggingface repository is incompatible with SGLang Diffusion Module, so the conversion script needs to be run first

```bash
python3 longcat_to_diffusion.py \
    --source /root/.cache/huggingface/hub/models--meituan-longcat--LongCat-Video/snapshots/03b55529b1d1d4045f5fbe14d65c8c6e8116b278 \
    --output /root/.cache/huggingface/hub/models--meituan-longcat--LongCat-Video/snapshots/03b55529b1d1d4045f5fbe14d65c8c6e8116b279 \
    --validate
```


python/sglang/multimodal_gen/runtime/pipelines/longcat_pipeline.py
```diff
@@ -0,0 +1,148 @@
+# SPDX-License-Identifier: Apache-2.0
+"""
```
mickqian, line 2: A lot of code can be replaced with those in SGL-diffusion, with better performance. CouldA lot of code can be replaced with those in SGL-diffusion, with better performance. Could you solve that first?

attack204: I'd be glad to, but I'm not familiar with the sgl-diffusion module. Could you give me some hints?

attack204: > A lot of code can be replaced with those in SGL-diffusion, with better performance. CouldA lot of code can be replaced with those in SGL-diffusion, with better performance. Could you solve that first?

I have already implemented part of the optimizations. Could you please help review them?

Optimization:

Integrate LongCatDenoisingStage into DenoisingStage.

Modify RotaryPositionalEmbedding3D to use the built-in NDRotaryEmbedding implementation.

Modify PatchEmbed3D to use the built-in PatchEmbed implementation.

Attempted but unsuccessful:

Attempted to modify TimestepEmbedder to use the built-in implementation: Found incompatibility. The TimestepEmbedder in LongCat is more similar to ZImage, but abstracting it into a common public method may not be elegant.

Attempted to introduce the cache-DiT mechanism. Unfortunately, the cache_dit/caching/block_adapters/__init__.py file does not support LongCat-Video.

python/sglang/multimodal_gen/runtime/layers/attention/backends/block_sparse_attention/__init__.py
yhyang201, line 1: Would it be better to move it under `runtime/layers/attention/backends` ?

attack204: fixed

PR #14301, opened 2025-12-02
Rename and refactor `get_available_gpu_memory` function

rauletorresc:
## Motivation

`get_available_gpu_memory` not only gets the available memory for GPU devices. It does it for XPUs, HPUs, CPUs and NPUs as well. 

## Modifications

* Rename to `get_available_device_memory` to better reflect its capabilities.
* Modify all related function calls accordingly.
* Keep the old function name as a wrapper of the new one for backwards compatibility, but mark it as deprecated.
* Refactor into a reusable function all the duplicated logic that checks if a device ID is valid.

## Accuracy Tests

Not applicable.

## Benchmarking and Profiling

Not applicable.

## Checklist


python/sglang/srt/utils/common.py
```diff
@@ -482,23 +483,43 @@ def inner_func(*args, **kwargs):
...
-    if device == "cuda":
-        num_gpus = torch.cuda.device_count()
-        assert gpu_id < num_gpus
+    get_available_device_memory(
+        device,
+        gpu_id,
+        distributed=distributed,
+        empty_cache=empty_cache,
+        cpu_group=cpu_group,
+    )
```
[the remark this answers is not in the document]
rauletorresc: Fixed

python/sglang/srt/utils/common.py
```diff
@@ -83,6 +83,7 @@
 import torch.distributed as dist
 import triton
 import zmq
+from deprecated import deprecated
```
ping1jing2, line 83: I think we have some problems in this code

1. if you want to add the 3rd-party lib named `deprecated` in SGLang, you should change `pyproject.toml`
2. if you want to use standard lib, maybe you should use `from warnings import deprecated` according to this [link](https://docs.python.org/3/library/warnings.html#warnings.deprecated), but we should ensure python >=3.13.( `Added in version 3.13: See [PEP 702](https://peps.python.org/pep-0702/).`)

@hnyls2002 maybe you can give some suggestions

python/sglang/srt/utils/common.py
```diff
@@ -482,23 +483,43 @@ def inner_func(*args, **kwargs):
     return wrapper
 
 
+@deprecated
```
ping1jing2: maybe it would be better you add some message in this place, just like `@deprecated("some user-friendly msg")`

python/sglang/srt/utils/common.py
```diff
@@ -507,64 +528,46 @@ def get_available_gpu_memory(
...
-        # TODO: rename the variables in the current function to be not GPU specific
         total_free_memory = psutil.virtual_memory().available
         n_numa_node: int = len(get_cpu_ids_by_node())
-        free_gpu_memory = round(total_free_memory / n_numa_node, 3)
+        free_device_memory = round(total_free_memory / n_numa_node, 3)
     elif device == "npu":
-        num_gpus = torch.npu.device_count()
-        assert gpu_id < num_gpus
+        assert device_id < torch.npu.device_count()
+        warn_not_device_id(torch.npu.current_device())
```
ping1jing2: maybe you can use `torch.get_device_module` to eliminate redundant code.

python/sglang/srt/utils/common.py
rauletorresc: You are right.
1. I have added it to pyproject_other.toml
2. I did not use `warnings` for maintaining backwards compatibility

python/sglang/srt/utils/common.py
rauletorresc: Done

python/sglang/srt/utils/common.py
rauletorresc: That's a great idea! I have used it now, and helps a lot to eliminate redundancy. Thanks!

python/pyproject_other.toml
```diff
@@ -25,6 +25,7 @@ runtime_common = [
   "compressed-tensors",
   "decord2",
   "datasets",
+  "deprecated",
```
ping1jing2, line 28: maybe you need to change all the `.toml` file, right?

rauletorresc: Correct. Apologies. Fixed now.

ping1jing2 (approved):

Makcum888e (approved):
LGTM

ssshinigami (approved):
LGTM

rauletorresc:
> Damn. I fucking needs this function. thanks so much. But does this PR support AMD/hip?

The PR just renames the function to something more general. It also restructures and improves its logic. As far as I know `torch.cuda` also cover AMD/hip devices. I don't know if that is what you mean :)

PR #6530, opened 2025-05-22
[sgl-kernel] add rotary embed kernel for trivial head_sizes

mickqian:
## Motivation

Previously for Attention with `head_size not in [64, 128, 256, 512]` (which is common for Multimodal Attention), sgl will adopt [`rotary_embedding`](https://github.com/vllm-project/vllm/blob/fa72f9a8126051abb9d00144f11aeb7615f36d21/csrc/pos_encoding_kernels.cu#L133) from vllm.

This pr copied and adapted the mentioned kernel with minor improvements.

## Modifications

1. Add kernel for rotary-embedding with common head_sizes
2. Add according tests

### test_rotary_embedding_benchmark[80-80-1000000.0-1000000.0-False-dtype1-cuda-1-4000-16-16]

| 实现             | Min           | Max           | Mean          | StdDev        | Median        | IQR           | Outliers   | OPS             | Rounds    | Iterations |
|------------------|---------------|---------------|---------------|---------------|---------------|---------------|------------|-----------------|-----------|------------|
| native           | 274.0989 (1.00) | 561.6210 (1.00) | 278.5268 (1.00) | 2.5004 (1.00) | 278.5940 (1.00) | 1.7020 (1.00) | 892;200    | 3,590.3191 (1.00) | 20000 (1.00) | 1 (1.00)   |
| original (vllm)  | 117.0728 (0.43) | 195.9121 (0.35) | 119.9629 (0.43) | 1.1579 (0.46) | 119.9138 (0.43) | 0.8049 (0.47) | 1595;562   | 8,335.9090 (2.32) | 20000 (1.00) | 1 (1.00)   |
| sgl              | 29.7832 (0.11)  | 731.3141 (1.30) | 32.2445 (0.12)  | 5.2406 (2.10) | 33.1532 (0.12)  | 3.0030 (1.76) | 25;21      | 31.0130 (0.01)    | 20000 (1.00) | 1 (1.00)   |

---

### test_rotary_embedding_benchmark[80-80-1000000.0-1000000.0-False-dtype0-cuda-1-8840-16-16]

| 实现             | Min           | Max             | Mean           | StdDev         | Median         | IQR           | Outliers   | OPS               | Rounds    | Iterations |
|------------------|---------------|-----------------|----------------|----------------|----------------|---------------|------------|-------------------|-----------|------------|
[...]

PR #14302, opened 2025-12-02
diffusion: rotary embedding kernel

RubiaCx:
## Motivation

- **close #12985**
- Diffusion models rely heavily on RoPE in both text and vision branches. This PR adds a rotary embedding CUDA kernel to `sgl-kernel`.
- Previous kernels had limited support for non‑standard head sizes (e.g., 80). This work follows the direction of the fast rotary embedding work in [#10527](https://github.com/sgl-project/sglang/pull/10527), and generalizes earlier rotary kernel efforts in [#6530](https://github.com/sgl-project/sglang/pull/6530) to cover more layouts and Diffusion‑style use cases.

## Modifications

- Add a generic rotary embedding CUDA kernel in `sgl-kernel/csrc/multimodal/rotary_embedding.cu`, supporting:
  - Separate `cos` / `sin` caches, NeoX and non‑NeoX layouts, 2D/3D Q/K, and GQA/MQA.
- Wire up the kernel into `sgl-kernel`:
  - Register C++ op in `include/sgl_kernel_ops.h` and `csrc/common_extension.cc`, add to `CMakeLists.txt`.
  - Expose a Python API via `python/sgl_kernel/rotary_embedding.py` and export from `python/sgl_kernel/__init__.py`.
- Add/update tests and benchmarks:
  - Add `tests/test_mm_rotary_embedding.py` and `benchmark/bench_mm_rotary_embedding.py`.

## Accuracy Tests

- `tests/test_mm_rotary_embedding.py`:
  - Parametrized coverage over multiple `(batch_size, seq_len, num_heads, num_kv_heads, head_size, dtype)` combinations.
  - Compare Q/K outputs against a float32 reference implementation; all cases pass within the expected tolerance.

## Benchmarking and Profiling

- `benchmark/bench_mm_rotary_embedding.py`:
  - Benchmarks our kernel vs. vLLM implementation and FlashAttention’s rotary path.
  - On representative configs (e.g., `batch_size ∈ {1, 32}`, `heads ∈ {32/8, 64/8, 8/8, 32/1}`, `head_size ∈ {64, 80, 128, 256}`), the new kernel achieves roughly **1.05–1.25×** speedup over the baseline and up to **2×+** over FlashAttention’s rotary.
- There are potential follow‑up optimizations can be introduced in future PRs without changing the Python API.

## Checklist

---
### CI States

Latest PR Test (Base): :x: [Run #33620844720](https://github.com/sgl-project/sglang/actions/runs/33620844720)
Latest PR Test (Extra): :x: [Run #33620844561](https://github.com/sgl-project/sglang/actions/runs/33620844561)
Latest PR Test (AMD ROCm 7.2): :x: [Run #33620844738](https://github.com/sgl-project/sglang/actions/runs/33620844738)


sgl-kernel/csrc/multimodal/rotary_embedding.cu
```diff
@@ -0,0 +1,276 @@
...
+  int x_index, y_index;
+
+  if (IS_NEOX) {
+    // NeoX-style: interleaved layout [x0, y0, x1, y1, ...].
+    // For NeoX, cos/sin have shape [..., rotary_dim/2]; 
+    // each index corresponds to one (x,y) pair.
+    x_index = 2 * rot_offset;
+    y_index = x_index + 1;
+
+    const float cos_val = static_cast<float>(SGLANG_LDG(cos_ptr + rot_offset));
```
BBuf: Can we use float4 to vectorize global memory read?

RubiaCx:
@BBuf I addressed the low-num_tokens under-utilization by updating the diffusion RoPE kernel. 
- Launch policy: when num_tokens <= 4 and blocks_per_token > 1, switch to a 2D launch (token × pair-tile) to increase parallelism for tiny-token regimes. Otherwise keep a 1D launch (one block per token) for general workloads.
- Vectorization: added a float2 fast path on the hot loop to improve memory throughput.

Results (Qwen-image): average kernel time improved from 29.551 µs to 24.557 µs (1.203× speedup).
image

Additionally, compared to the positions API (FlashInfer-based path), the diffusion RoPE kernel path is ~1.24× faster on average (weighted across configs).

python/sglang/multimodal_gen/runtime/models/dits/qwen_image.py
```diff
@@ -397,18 +402,34 @@ def forward(
...
-                txt_key, txt_cos, txt_sin, interleaved=True
-            )
+
+            if sgl_rotary_embedding is not None:
+                def _apply_sgl_rope(q: torch.Tensor, k: Optional[torch.Tensor], cos, sin):
+                    # sgl_kernel expects contiguous [num_tokens, num_heads, head_size]
+                    q_shape = q.shape
+                    q_view = q.contiguous().view(-1, q_shape[-2], q_shape[-1])
+                    k_view = k.contiguous().view(-1, k.shape[-2], k.shape[-1]) if k is not None else None
+                    # interleaved=True; kernel内部会根据 cos/sin 形状处理 Neox / GPT-J 布局
```
yuan-luo: Comment in English is preferred.

```
    """内部统一的 rotary dispatch：
    - mode="cos_sin": 使用 (cos, sin) 直接旋转
    - mode="positions": 使用 (positions, cos_sin_cache) 旋转
```

RubiaCx: sorry for the typo

sgl-kernel/python/sgl_kernel/rotary_embedding.py
```diff
@@ -0,0 +1,161 @@
...
+    interleaved: Optional[bool] = None,
+    is_neox: Optional[bool] = None,
+    # cos/sin mode
+    cos: Optional[torch.Tensor] = None,
+    sin: Optional[torch.Tensor] = None,
+    # positions mode
+    positions: Optional[torch.Tensor] = None,
+    cos_sin_cache: Optional[torch.Tensor] = None,
+) -> None:
+    """内部统一的 rotary dispatch：
```
BBuf: No chinese

sgl-kernel/python/sgl_kernel/rotary_embedding.py
```diff
@@ -0,0 +1,161 @@
...
+            torch.ops.sgl_kernel.rotary_embedding(
+                positions=positions,
+                query=query,
+                key=key if key is not None else None,
+                head_size=head_size,
+                cos_sin_cache=cos_sin_cache,
+                is_neox=effective_interleaved,
+            )
+            return
+        except Exception:
```
BBuf: Try catch is strange?

sgl-kernel/include/utils.h
```diff
@@ -459,11 +459,61 @@ inline uint32_t next_pow2(uint32_t x) noexcept {
...
 #endif
+
+// Define dispatch macro for Rotation Dimension (Pair Count)
+// Case 32 -> head_size 64
+// Case 64 -> head_size 128
+// Case 128 -> head_size 256
+#define DISPATCH_ROT_DIM(embed_dim, ...)  \
+  switch (embed_dim) {                    \
+    case 16: {                            \
+      constexpr int ROT_EMBED_DIM = 16;   \
```
BBuf: Maybe we can rename it to `_DIFFUSION_ROT_EMBED_DIM`.

RubiaCx:
@BBuf @DarkSharpness Hi! I migrated our SGL JIT RoPE implementation into `python/sglang/jit_kernel` and added benchmarks that report speedup using SGL JIT (cos/sin) as the baseline, with comparisons against FlashInfer (SGL AOT positions) and vLLM. The JIT kernel is faster for most tested shapes.
image
image

I also updated Qwen-Image to use our JIT RoPE implementation. Output quality matches the previous behavior.
image
image

Note: the JIT kernel incurs a warmup/compile overhead on first invocation (also visible in profiling), so the first run may be slower; steady-state runs show the intended speedup.
image

python/sglang/multimodal_gen/runtime/layers/layernorm.py
```diff
@@ -434,17 +434,19 @@ def apply_qk_norm(
...
     q_eps = q_norm.variance_epsilon
     k_eps = k_norm.variance_epsilon
-    # Only try fused path on CUDA and when it won't introduce implicit copies.
     if (
         q.is_cuda
+        and (torch.version.cuda is not None)
         and allow_inplace
+        and q.is_contiguous()
+        and k.is_contiguous()
         and (q_eps == k_eps)
```
DarkSharpness, line 896: inplace qk norm only requires the last 2 dimension (i.e, num_head, head_dim) to be contiguous, which should be applicable in most cases

python/sglang/multimodal_gen/runtime/layers/layernorm.py
```diff
@@ -458,7 +460,15 @@ def apply_qk_norm(
...
 
-    raise RuntimeError(
-        "apply_qk_norm: fused inplace QK-norm is not applicable "
-        "(expected CUDA, contiguous q/k, matching eps, and supported head_dim)"
-    )
+    # Fallback: apply the PyTorch norms.
+    q_out = q_norm(q)
+    k_out = k_norm(k)
+
+    if allow_inplace and q.is_contiguous() and q_out.shape == q.shape:
```
DarkSharpness: `allow_inplace` is just a performance hint, not a must (otherwise, it should be called `use_inplace`. if not applicable, simply use the fastest implementation is ok (i.e. out-of-place and without extra `copy_`)

python/sglang/multimodal_gen/runtime/models/dits/qwen_image.py
```diff
@@ -552,23 +549,14 @@ def forward(
...
-            )
-            txt_query, txt_key = apply_qk_norm(
-                q=txt_query,
-                k=txt_key,
-                q_norm=self.norm_added_q,
-                k_norm=self.norm_added_k,
-                head_dim=txt_query.shape[-1],
-                allow_inplace=True,
-            )
+        if self.norm_q is not None:
```
BBuf: If `_is_cuda()` is False and we need fallback?

python/sglang/jit_kernel/csrc/rotary_embedding_cos_sin.cuh
```diff
@@ -0,0 +1,1106 @@
...
+  scalar_t* k_ptr = (k != nullptr) ? static_cast<scalar_t*>(k->data_ptr()) : nullptr;
+
+  const dim3 grid2d((int)t, std::max(1, blocks_per_token));
+  const dim3 grid1d((int)t);
+  const dim3 block(threads_per_block);
+
+  // Compile-time specializations for common embed dims; fallback to runtime (ROT=0).
+  switch (embed_dim_for_rotation) {
+    case 32:
+      dispatch_rotary_launch<32, scalar_t>(
```
DarkSharpness: You can pass the dimension as a template args to eliminate dispatch code.

RubiaCx: Got it.

RubiaCx: `embed_dim_for_rotation` is a runtime value and depends on model config, so we still need a dispatch point to select a compile-time specialization. 
The current switch instantiates a few common `ROT` values and falls back to `ROT=0` for the rest. If desired, we can simplify redundant code (e.g., via a macro or something), but we can’t eliminate dispatch entirely without either moving it elsewhere or always using the generic kernel.

DarkSharpness: We have some utility functions in `jit_kernel` to help instantiate C++ template.  You may refer to the following code as an example:

https://github.com/sgl-project/sglang/blob/5b4f790200929beb52ba51a6c7b82fbf45e8c9b1/python/sglang/jit_kernel/norm.py#L19-L27

DarkSharpness: The point here is that for the same model, the embed_dim is identical across all layers, so we only need to compile once.

For different models, we can instantiate different templates to achieve the best performance.

RubiaCx:
```
sglang generate \
    --model-path Qwen/Qwen-image \
    --prompt "A Logo With Bold Large Text: SGL Diffusion" \
    --save-output \
    --num-inference-steps 40 \
    --seed 0
```
### apply_sglang_jit_rope_qk_inplace
```
[01-02 23:00:26] [DenoisingStage] started...
100%|████████████████████████████████████████████████████████████████████████████████| 40/40 [00:11<00:00,  3.43it/s]
[01-02 23:00:37] [DenoisingStage] average time per step: 0.2919 seconds
[01-02 23:00:37] [DenoisingStage] finished in 11.6819 seconds
```
### apply_flashinfer_rope_qk_inplace
```
[01-02 23:08:28] [DenoisingStage] started...
100%|████████████████████████████████████████████████████████████████████████████████| 40/40 [00:15<00:00,  2.63it/s]
[01-02 23:08:43] [DenoisingStage] average time per step: 0.3802 seconds
[01-02 23:08:43] [DenoisingStage] finished in 15.2134 seconds
```

python/sglang/jit_kernel/csrc/rotary_embedding_cos_sin.cuh
RubiaCx: Got it, I will try to move the ROT specialization to the Python JIT layer.

yuan-luo:
Please fix the lint. 
btw, I believe this jit-kernel is also helpful to non-diffusion inference.

python/sglang/jit_kernel/csrc/rotary_embedding_cos_sin.cuh
```diff
@@ -0,0 +1,1026 @@
...
+//   - layout: interleaved vs non-interleaved
+//   - compute: vectorized 16B tiles vs scalar
+//   - memory: q/k aligned 16B load/store vs scalar gather/scatter
+template <int ROT, typename scalar_t>
+__forceinline__ void dispatch_rotary_launch(
+    bool use_grid_2d,
+    dim3 grid2d,
+    dim3 grid1d,
+    dim3 block,
+    const decltype(host::LaunchKernel::resolve_device(std::declval<DLDevice>())) stream,
```
DarkSharpness: This type is just `cudaStream_t`. No need to use decltype.

python/sglang/jit_kernel/csrc/rotary_embedding_cos_sin.cuh
```diff
@@ -0,0 +1,1026 @@
...
+    int num_heads,
+    int num_kv_heads,
+    int head_size,
+    int blocks_per_token) {
+  // 2D grid path
+  if (use_grid_2d) {
+    if (interleaved) {
+      if (use_vec) {
+        if (qk_aligned16) {
+          host::LaunchKernel(grid2d, block, stream)(
```
DarkSharpness: too many if else here. actually, you can simplify this into 2 stage:

1. get the kernel pointer
2. launch the kernel with given config

since the arguments for different kernels are identical, we should only keep one `host::LaunchKernel`

DarkSharpness: BTW, i think our jit_kernel need some utilities to dispatch based on bool value. A temporary workaround could be passing all the boolean value as template from Python side, but I'm not sure whether it's good practice.

python/sglang/jit_kernel/csrc/rotary_embedding_cos_sin.cuh
```diff
@@ -0,0 +1,1026 @@
...
+      host::Panic("Unsupported dtype for rotary_embedding_cos_sin");
+    }
+  }
+
+  static void run_qk(
+      const tvm::ffi::TensorView cos,
+      const tvm::ffi::TensorView sin,
+      const tvm::ffi::TensorView query,
+      const tvm::ffi::TensorView key,
+      const tvm::ffi::AnyView positions_any,
```
DarkSharpness: you can use `tvm::ffi::Optional<tvm::ffi::TensorView>`

python/sglang/jit_kernel/benchmark/bench_rotary_embedding.py
```diff
@@ -0,0 +1,447 @@
...
+        q4_base=q4,
+        k4_base=k4,
+        cos_sin_cache_flashinfer=cos_sin_cache_flashinfer,
+    )
+
+
+def make_fn(
+    provider: str, case: Case, head_size: int, interleaved: bool
+) -> Callable[[], None]:
+    # 统一 clone：每次 bench 都从同一份 base 开始，避免“越跑越旋”
```
DarkSharpness: remove Chinese comments

python/sglang/multimodal_gen/runtime/layers/rotary_embedding.py
```diff
@@ -123,6 +127,149 @@ def apply_flashinfer_rope_qk_inplace(
...
+            and positions.dim() == 1
+        ):
+            raise ValueError("positions must be a 1D torch.long Tensor")
+        if positions.numel() != num_tokens:
+            raise ValueError(
+                f"positions length must be bsz*seqlen={num_tokens}, got {positions.numel()}"
+            )
+        positions = positions.to(q.device, non_blocking=True)
+
+    q3 = q.reshape(num_tokens, nheads, d)
```
DarkSharpness: avoid using `reshape`. try to use `view` in most cases. reshape may have some implicit memory copying, which can make the benchmark completely flawed

python/sglang/multimodal_gen/runtime/layers/rotary_embedding.py
```diff
@@ -123,6 +127,149 @@ def apply_flashinfer_rope_qk_inplace(
...
+    if not q3.is_contiguous():
```
DarkSharpness: same. a good implementation should avoid contiguous op, which may involve some implicit memory copying,

python/sglang/multimodal_gen/runtime/layers/rotary_embedding.py
```diff
@@ -123,6 +127,149 @@ def apply_flashinfer_rope_qk_inplace(
...
+    if head_size is None:
+        head_size = d
+    if head_size != d:
+        raise ValueError(f"head_size mismatch: inferred {d}, but head_size={head_size}")
+
+    try:
+        from sglang.jit_kernel.rotary_embedding import (
+            rotary_embedding_cos_sin as sglang_jit_rotary_embedding_cos_sin,
+        )
+    except Exception as e:
```
DarkSharpness: this try catch is not necessary. just let it crash at where the error is thrown

python/sglang/multimodal_gen/runtime/layers/rotary_embedding.py
```diff
@@ -38,6 +38,10 @@
 
 logger = init_logger(__name__)
 
+_JIT_ROPE_SPLIT_CAST_CACHE: "OrderedDict[tuple, tuple[torch.Tensor, torch.Tensor]]" = (
```
DarkSharpness: why we need to cache this? a common implementation like flashinfer/torch don't need something like this

python/sglang/jit_kernel/benchmark/bench_rotary_embedding.py
```diff
@@ -0,0 +1,447 @@
...
+def benchmark(
+    batch_size: int, seq_len: int, head_size: int, interleaved: bool, provider: str
+):
+    if DEVICE == "cuda" and not torch.cuda.is_available():
+        raise RuntimeError("CUDA not available")
+
+    case = prepare_case(batch_size, seq_len, head_size, interleaved)
+
+    global _SANITY_DONE
+    if not _SANITY_DONE and provider == "jit_fused":
```
DarkSharpness: I think sanity check should be put only in test file. For benchmark file, just keep it simple.

python/sglang/multimodal_gen/runtime/layers/rotary_embedding.py
DarkSharpness: such cache design is not elegant and very error-prune, so try to eliminate this if it won't hurt performance

python/sglang/jit_kernel/benchmark/bench_rotary_embedding.py
```diff
@@ -0,0 +1,447 @@
...
+        styles=None,
+        ylabel="us",
+        plot_name="rotary-embedding-performance",
+        args={},
+    )
+)
+def benchmark(
+    batch_size: int, seq_len: int, head_size: int, interleaved: bool, provider: str
+):
+    if DEVICE == "cuda" and not torch.cuda.is_available():
```
DarkSharpness: remove this check

python/sglang/jit_kernel/benchmark/bench_rotary_embedding.py
```diff
@@ -0,0 +1,447 @@
...
+from typing import Callable, Dict, Tuple
+
+import torch
+import triton.testing
+
+DEVICE = "cuda"
+DTYPE = torch.bfloat16
+MAX_SEQ_LEN = 8192
+NUM_HEADS = 32
+BS_RANGE = [1, 64]
```
DarkSharpness: add more benchmark when not in ci. you can take a look at  [python/sglang/jit_kernel/benchmark/bench_qknorm.py](https://github.com/sgl-project/sglang/blob/main/python/sglang/jit_kernel/benchmark/bench_qknorm.py)

python/sglang/multimodal_gen/runtime/layers/rotary_embedding.py
RubiaCx: Done.

BBuf (approved):

python/sglang/kernels/jit/csrc/rotary_embedding_cos_sin.cuh
```diff
@@ -0,0 +1,818 @@
...
+template <typename T>
+__device__ __forceinline__ T from_float(float v) {
+  return static_cast<T>(v);
+}
+template <>
+__device__ __forceinline__ half from_float<half>(float v) {
+  return __float2half_rn(v);
+}
+template <>
+__device__ __forceinline__ nv_bfloat16 from_float<nv_bfloat16>(float v) {
```
DarkSharpness, line 43: In #16884 we introduce `type.cuh`, which include some type conversion primitives in `namespace device`. You may try to replace these type casting with `cast<To>(value)`.

python/sglang/kernels/jit/csrc/rotary_embedding_cos_sin.cuh
```diff
@@ -0,0 +1,818 @@
...
+  if constexpr (interleaved) {
+    const int base = rot_offset * 2;
+
+    if constexpr (aligned_qk) {
+      data.vec = *reinterpret_cast<const float4*>(arr + base);
+    } else {
+      union VecU {
+        float4 v;
+        scalar_t e[kElePerVec];
+      } tmp;
```
DarkSharpness, line 91: Try to replace this with `device::AlignedVector`. This can also generate some effective aligned load/store instructions (just like `float4`).

python/sglang/kernels/jit/csrc/rotary_embedding_cos_sin.cuh
```diff
@@ -0,0 +1,818 @@
...
+        apply_token_rotary_embedding<scalar_t, interleaved>(
+            key_for_token_head, current_token_cos_ptr, current_token_sin_ptr, rot_offset, embed_dim_for_rotation);
+      }
+    }
+  }
+}
+
+// RoPE kernel with 1D grid: one block per token.
+template <typename scalar_t, bool interleaved, bool vectorized, bool aligned_qk, int ROT_EMBED_DIM>
+__launch_bounds__(512) __global__ void rotary_embedding_kernel_1d(
```
DarkSharpness, line 387: I'm not sure about the performance here. In my `qknorm` implementation, we use a persistent 2d kernel, where each warp handle one qk head dimension at a time. The code logic is as follows:

1. Each warp determines which q head or k head it will process at this iteration.
2. Set up the pointer information https://github.com/sgl-project/sglang/blob/e7df8bdc5c9b02b174e8b9deb4fd21a1ba301276/python/sglang/jit_kernel/csrc/elementwise/qknorm.cuh#L54-L59
3. Load and perform algorithm on the input https://github.com/sgl-project/sglang/blob/e7df8bdc5c9b02b174e8b9deb4fd21a1ba301276/python/sglang/jit_kernel/csrc/elementwise/qknorm.cuh#L60-L63

This can hopefully reduce code size and improve readability. You may refer to this as an example. I would suggest you take a similar approach (persistent 2d kernel + unified code logic) if it won't bring too much regression to performance.

python/sglang/kernels/jit/csrc/rotary_embedding_cos_sin.cuh
```diff
@@ -0,0 +1,818 @@
...
+  const int kElePerVec = kVecBytes / elem_bytes;
+  const int pairs_per_step = interleaved ? (kElePerVec / 2) : kElePerVec;
+
+  bool can_vec_compute = true;
+  if ((embed_dim_for_rotation % pairs_per_step) != 0) can_vec_compute = false;
+  if ((reinterpret_cast<uintptr_t>(cos.data_ptr()) % kVecBytes) != 0) can_vec_compute = false;
+  if ((reinterpret_cast<uintptr_t>(sin.data_ptr()) % kVecBytes) != 0) can_vec_compute = false;
+  if (((r * elem_bytes) % kVecBytes) != 0) can_vec_compute = false;
+
+  bool qk_aligned16 = true;
```
DarkSharpness, line 707: Do we really have to consider the unaligned cases? In most cases, I would expect the head bytes to be at least 64 bytes aligned. Do not optimize too early for rare cases.

python/sglang/kernels/jit/csrc/rotary_embedding_cos_sin.cuh
```diff
@@ -0,0 +1,818 @@
...
+  const int max_pairs_to_rotate_per_token =
+      (k == nullptr) ? ((int)hq * embed_dim_for_rotation)
+                     : std::max((int)hq * embed_dim_for_rotation, (int)hk * embed_dim_for_rotation);
+
+  constexpr int kVecBytes = 16;
+  const int elem_bytes = (int)sizeof(scalar_t);
+  const int kElePerVec = kVecBytes / elem_bytes;
+  const int pairs_per_step = interleaved ? (kElePerVec / 2) : kElePerVec;
+
+  bool can_vec_compute = true;
```
DarkSharpness, line 701: Do we really have to consider the unaligned cases? In most cases, I would expect the head bytes to be at least 64 bytes aligned. Do not optimize too early for rare cases.

PR #14307, opened 2025-12-02
[SMG][DS32][fix] support dsv32, add role developer

jimmy-evo:
Hi from [novita.ai](https://novita.ai/) team 👋

same as #14304 

add role developer for ChatMessage

### DIFF
--- a/sgl-model-gateway/src/protocols/chat.rs
+++ b/sgl-model-gateway/src/protocols/chat.rs
@@ -56,6 +56,14 @@ pub enum ChatMessage {
     },
     #[serde(rename = "function")]
     Function { content: String, name: String },
+    #[serde(rename = "developer")]
+    Developer {
+        content: MessageContent,
+        #[serde(skip_serializing_if = "Option::is_none")]
+        tools: Option<Vec<Tool>>,
+        #[serde(skip_serializing_if = "Option::is_none")]
+        name: Option<String>,
+    }
 }
 
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq)]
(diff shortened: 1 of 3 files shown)


sgl-model-gateway/src/protocols/chat.rs
```diff
@@ -56,6 +56,14 @@ pub enum ChatMessage {
     },
     #[serde(rename = "function")]
     Function { content: String, name: String },
+    #[serde(rename = "developer")]
+    Developer {
+        content: MessageContent,
+        #[serde(skip_serializing_if = "Option::is_none")]
+        tools: Option<Vec<Tool>>,
```
slin1237, line 63: there is no tool in developer?
https://platform.openai.com/docs/api-reference/chat/create#chat_create-messages-developer_message-role

jimmy-evo: > there is no tool in developer? https://platform.openai.com/docs/api-reference/chat/create#chat_create-messages-developer_message-role

i didn't found tool in developer on openai reference neither.

but deepseek use tool in developer...

slin1237: Can u send me reference where DeepSeek is using a different OpenAI spec? this pretty much means OpenAI SDK won't work for DeepSeek right?

jimmy-evo: > Can u send me reference where DeepSeek is using a different OpenAI spec? this pretty much means OpenAI SDK won't work for DeepSeek right?

https://huggingface.co/deepseek-ai/DeepSeek-V3.2/blob/main/encoding/test_input_search_w_date.json

PR #14325, opened 2025-12-03
[DeepseekV3.2][NSA][Indexer] Fix PAGED top-k transform for NSA indexer chunked execution on H200

YAMY1234:
## Motivation

In `extend` mode with NSA indexer enabled, H200 setups that select **PAGED** top-k transform (`flashmla_kv` + FP8 KV cache) may trigger chunked execution in `_get_topk_ragged` when the logits matrix becomes large.

Previously, chunked execution only sliced **logits / lengths / ks**, but continued to use full-batch `page_table_1` and full-batch query-sequence metadata.
This causes `fast_topk_transform_fused` to receive mismatched shapes:

* `prefill_bs = B_chunk` (from local cu_seqlens_q)
* `src_page_table.size(0) = batch_size` (full batch)

leading to kernel-level assertion failures such as:

```
Expected src_page_table.size(0) == prefill_bs to be true
```

B200 typically avoids this path because it uses **RAGGED** top-k transform, which does not depend on paged metadata and therefore does not require alignment.

This PR completes the missing logic required for **PAGED + chunked** execution.

---

## Modifications

This PR updates `_get_topk_ragged` to fully support chunked execution under **PAGED** top-k transform:

### Key changes

* Track the mapping from each **query token → its originating request index** (`token_to_batch_idx`).
* During chunked execution, construct:

  * a **local `cu_seqlens_q_chunk`** describing the B_chunk query tokens as *B_chunk sequences*, each of length 1.
  * a **local `batch_idx_list`**, selecting the correct rows from `page_table_1` for this chunk.
* Pass both of them into `NSAIndexerMetadata.topk_transform(...)` so that
  `src_page_table.size(0) == prefill_bs == logits_chunk.shape[0]`.

This makes the PAGED fused top-k kernel fully consistent under chunking, matching the semantics already used by the RAGGED variant.

No changes were made to the RAGGED path or normal (non-chunked) execution.

---

## Accuracy Tests

Validated correctness on both **H200** and **B200** using:

```
    --num-shots 40 \
    --num-questions 1319 \
    --parallel 1319
```

Results (B200 identical across devices and before/after patch, H200 will throw error before patch):

```
Accuracy: 0.958
Invalid: 0.000
Latency: 45.841 s
Output throughput: 2745.233 token/s
```

---

## Benchmarking and Profiling

---

## Checklist

* [ ] Format code with pre-commit
* [ ] Add unit tests if needed
* [ ] Update documentation if needed
* [ ] Provide accuracy & speed results
* [ ] Follow code style guidance
* [ ] Coordinate with maintainers for merge


python/sglang/srt/layers/attention/nsa/nsa_indexer.py
```diff
@@ -431,6 +434,10 @@ def _get_topk_ragged(
         token_nums, _, _ = q_fp8.shape
         device = q_fp8.device
 
+        token_to_batch_idx_tensor = torch.tensor(
```
Fridge003: Can we delay this operation to somewhere between line466, since if chunk isn't needed this operation will waste some time.

Also I'm not sure whether creating this tensor will cause sync between CPU and GPU.

Fridge003 (changes requested):

python/sglang/srt/layers/attention/nsa/nsa_indexer.py
YAMY1234: Revised! 
I’ve updated the patch so that this tensor is only created when chunking is actually needed, and in the PAGED path where batch_idx_list is consumed. 
For the sync, I think this performs a host→device copy on the current CUDA stream which does not introduce a global device synchronization

PR #14337, opened 2025-12-03
remove unecessary dual stream token threshold from the rest of models (qwen moe, kimi linear, etc.)

b8zhong:
in an earlier pr of llama 4, i found the same conclusion as https://github.com/sgl-project/sglang/pull/9405 (it is also done in glm now).

TODO there are still some in GDN and NSA backend, I'm not sure if those are needed either. it can be done seperately


Fridge003:
@b8zhong The usage of dual stream threshold for nsa/gdn is different from these models, so we need to check them further

Fridge003 (approved):

PR #14339, opened 2025-12-03
[Performance] Optimize radix cache eviction performance

YiXR:
## Motivation

Currently, the RadixCache.evict method calls _collect_leaves() to traverse the entire Radix Tree to find evictable nodes. This operation has a time complexity of O(N), where N is the total number of nodes in the tree.

In high-concurrency scenarios where the GPU memory is fully utilized, evict is triggered frequently. The O(N) traversal causes significant CPU overhead and leads to latency jitter (spikes) during the decoding phase, especially when the Radix Tree is large.

## Modifications

Introduced self.evictable_leaves:

Changed from a dynamic search to an incrementally maintained Set

Incremental Updates:

Updated insert, _delete_leaf, inc_lock_ref, and dec_lock_ref methods to add/remove nodes from self.evictable_leaves dynamically when their state changes (e.g., reference count drops to 0 or a node becomes a leaf).

### Radix Cache Test
Test on **H20 TP8 Qwen/Qwen3-0.6B**

Before optimze: 7 ms for each eviction
After optimize: 0.5ms for each eviction
image

### HiCache Test
image

## Checklist

### DIFF
--- a/python/sglang/srt/mem_cache/radix_cache.py
+++ b/python/sglang/srt/mem_cache/radix_cache.py
@@ -301,6 +301,8 @@ class RadixCache(BasePrefixCache):
             raise ValueError(
                 f"Unknown eviction policy: {self.eviction_policy}. Supported policies: 'lru', 'lfu', 'fifo', 'mru', 'filo', 'priority'."
             )
+
+        self.evictable_leaves = set()
         self.reset()
 
     @classmethod
@@ -333,6 +335,7 @@ class RadixCache(BasePrefixCache):
         self.root_node.hash_value = []
         self.evictable_size_ = 0
         self.protected_size_ = 0
+        self.evictable_leaves.clear()
         self._record_all_cleared_event()
 
     def maybe_bigram_convert(
@@ -566,7 +569,7 @@ class RadixCache(BasePrefixCache):
 
         start_time = time.perf_counter()
         num_tokens = params.num_tokens
-        leaves = self._collect_leaves()
+        leaves = list(self.evictable_leaves)
         eviction_heap = [
             (self.eviction_strategy.get_priority(node), node) for node in leaves
         ]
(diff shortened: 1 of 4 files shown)


python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -734,6 +741,24 @@ def _delete_leaf(self, node):
...
+
+    def _update_leaf_status(self, node: TreeNode):
+        if node.evicted or node.lock_ref > 0:
+            if node in self.evictable_leaves:
+                self.evictable_leaves.remove(node)
+            return
+
+        for child in node.children.values():
+            if not child.evicted:
+                if node in self.evictable_leaves:
```
xiezhq-hermann: should we remove this? as long as there are non-evicted child, the node should be removed from the list

YiXR: When a new leaf node (x) is added to the device pool, both _update_leaf_status(x) and _update_leaf_status(x.parent) should be called, as the parent node needs to update its status based on the child node's state. （In HiCache Insert() func)

xiezhq-hermann: I mean, should we remove the line 756

xiezhq-hermann: or alternatively should we do update leaf node update in a different order? update for itself and then check for the parent node.

YiXR: If the parent node was not previously in the list (e.g., when it had existing children and a new child is added), the remove operation will throw an error without this validation. 

If we only increment or decrement the lock reference count on a node without changing the tree structure, there is no need to update the parent's status.

I will always update node itself and check it's parent like _update_leaf_status(x) + _update_leaf_status(x.parent) only if the tree structure changed(insert, delete, evict, promote).

xiezhq-hermann (approved):

PR #12612, opened 2025-11-04
feat: Add FP4 (E2M1) KV Cache Support for MHA

JackChuang:
# Summary
This PR introduces support for **FP4 (float4_e2m1fn_x2) KV caching** in Multi-Headed Attention (MHA)  e.g., Qwen and GPT-OSS . See #10083, points 1-2, for more context.

Co-authored-by: @yicwang Yichen Wang [yichen.wang@bytedance.com](mailto:yichen.wang@bytedance.com)

# Usage
```
$ python3 -m sglang.launch_server --kv-cache-dtype fp4_e2m1 ... 
```

### **Motivation** and Benefits
Large models often face GPU memory constraints when storing KV cache.
By introducing **FP4 quantization with scale buffers**, this PR significantly reduces KV memory usage and improves efficiency:
- Supports significantly more tokens than KV8 (≈1.78×) and KV16 (≈3.56×) due to FP4 quantization with block_size = 16.
- Improves scalability for longer context windows and throughput for large batch requests
- Enables inference of larger models or longer context windows on memory-limited GPUs.
- Seamless integration with existing inference pipelines without breaking KV16/KV8 workflows.

# Key Changes
- **MHATokenToKVPool**
    - Added FP4 KV cache support with uint8 storage format.
    - Introduced k_scale_buffer and v_scale_buffer for per-block scaling factors.
    - Integrated batched quantization (on update) and dequantization (on access) using KVFP4QuantizeUtil.
- **ModelRunner**
    - Updated GPU memory estimation logic to account for FP4 cache and scale buffers.
- **Compatibility**
    - Preserves existing FP16/FP8 KV cache behavior without changes.

# Accuracy tests for KV4 MHA
- FP4 KV cache is well-suited for large-scale models, providing memory savings with minimal accuracy impact.
- For smaller models, careful evaluation is needed to balance memory efficiency and accuracy.
## Qwen3-235B-A22B

| Model          | Dataset      | Metric   | Subset  | Num  | Score  | Cat.0   |
| -------------- | ------------ | -------- | ------- | ---- | ------ | ------- |
| KV4 (fp4_e2m1) |              |          |         |      |        |         |
| KV4            | gsm8k        | mean_acc | main    | 6595 | 0.9186 | default |
| KV4            | aime25       | mean_acc | OVERALL | 150  | 0.6    | -       |
| KV4            | gpqa_diamond | mean_acc | default | 990  | 0.6778 | default |
[...]

PR #10078, opened 2025-09-05
feat: Add FP4 (E2M1) KV Cache Support with Quantization Utilities for MLA

JackChuang:
## Summary
This PR introduces FP4 (E2M1) support for Multi-Head Latent Attention (MLA) KV cache in SGLang, enabling low-precision caching to reduce memory usage and improve inference efficiency. It integrates FP4 quantization utilities, Triton kernels, and unit tests while remaining backward compatible with FP16/FP8. See #10083, points 1-1, for more context.

Co-authored-by: @yicwang  Yichen Wang <yichen.wang@bytedance.com> 

## Usage
Added `--kv-cache-dtype=fp4_e2m1` option.
``` 
$ python3 -m sglang.launch_server --kv-cache-dtype fp4_e2m1 ... 
```

## Key Changes
1. **Server Argument Extension**
   - Added `--kv-cache-dtype=fp4_e2m1` option.

2. **FP4 Quantization Utility**
   - Added `KVFP4QuantizeUtil` with `batched_quantize` and `batched_dequantize` methods for block-wise (16) processing of [M, N, K] tensors.

3. **Core KV Cache Integration**
   - Updated `ModelRunner` and `MLATokenToKVPool` to support FP4 KV cache.
   - Added `kv_scale_buffer` for FP4 scaling.
   - Implemented Triton kernel `set_mla_kv_scale_buffer_kernel` for efficient nope+rope tensor handling.
   - Maintains backward compatibility with FP16/FP8 KV cache.

4. **Unit Test & Benchmark**
   - Tests validate FP4 correctness via MSE, MAE, PSNR, and Relative Error.
   - Benchmarks FP4 vs FP8 performance on GPU.

## Impact
- Enables low-precision FP4 KV cache for MLA.
- Reduces GPU memory usage while maintaining accuracy.
- Provides measurable speedup over FP8 quantization.
- Backward compatible with existing FP16/FP8 workflows.

## Accuracy Tests
The results show that on simpler datasets, the accuracy is nearly lossless compared to the baseline. On more challenging datasets, there is some accuracy degradation, but it remains within an acceptable range.
``` 
| Model | Dataset | Metric | Subset. | Num | Score | Cat.0 |


| DeepSeek-R1-0528-FP4-KV16 | gsm8k | mean_acc | main | 6595 | 0.9157 | default |
| DeepSeek-R1-0528-FP4-KV8_e4m3 | gsm8k | mean_acc | main | 6595 | 0.9154 | default |
[...]

PR #14348, opened 2025-12-03
add doc for quantized kv cache

b8zhong:
frequently i see the usage of quantized kv cache w/ an incompatible backend, then it was reported to be very slow, make it clear that you should not do it usually.

### DIFF
--- a/docs/advanced_features/quantization.md
+++ b/docs/advanced_features/quantization.md
@@ -12,7 +12,7 @@ on-the-fly to convert high-precision weights into a lower-precision format.
 **Note: For better performance, usability and convenience, offline quantization is recommended over online quantization.**
 
 If you use a pre-quantized model, do not add `--quantization` to enable online quantization at the same time.
-For popular pre-quantized models, please visit [Unsloth](https://huggingface.co/unsloth), [ModelCloud](https://huggingface.co/collections/ModelCloud/vortex-673743382af0a52b2a8b9fe2)
+For popular pre-quantized models, please visit [Unsloth](https://huggingface.co/unsloth), [NVIDIA ModelOpt](https://huggingface.co/collections/nvidia/inference-optimized-checkpoints-with-model-optimizer)
 or [NeuralMagic](https://huggingface.co/collections/neuralmagic) collections on HF for some
 popular quality validated quantized models. Quantized models must be validated via benchmarks post-quantization
 to guard against abnormal quantization loss regressions.
--- a/docs/advanced_features/quantized_kv_cache.md
+++ b/docs/advanced_features/quantized_kv_cache.md
@@ -1,0 +1,124 @@
+# Quantized KV Cache
+
+Quantized KV cache reduces the memory footprint of key-value cache storage by using lower-precision data types (FP8 or FP4) instead of the default model precision in BF16. During autoregressive generation, LLMs cache previously computed key-value pairs to avoid redundant calculations. The KV cache typically consumes a significant portion of GPU memory, especially for long sequences.
+
+Quantized KV cache is a memory optimization technique that primarily benefits throughput by allowing more tokens to be cached, but may introduce minimal accuracy degradation depending on the quantization format used.
+
+```{warning}
+**Performance Warning**: When quantized KV cache must be dequantized before use in attention operations, performance can be extremely slow if dequantization is not fused with the attention kernel. Always verify that your chosen attention backend supports quantized KV cache. Backends without fused support may experience significant throughput degradation, potentially negating the memory benefits.
+
+**Backend Support**: Not all attention backends support quantized KV cache. Refer to [Attention Backend](attention_backend.md) for which backends support it.
+```
+
+## Supported Formats
+
+SGLang supports the following quantized KV cache formats:
+
+### FP8 Format
+
+[OCP (Open Compute Project)](https://www.opencompute.org) specifies two common 8-bit floating point formats:
+
+- **E5M2** (5 exponent bits, 2 mantissa bits): Larger dynamic range (±57344.0), lower precision
+- **E4M3** (4 exponent bits, 3 mantissa bits): Higher precision, smaller dynamic range (±240.0)
+
+### FP4 Format
+
+```{warning}
+FP4 quantization is currently experimental.
+```
+
+[OCP (Open Compute Project)](https://www.opencompute.org) specifies MXFP4 (Microscaling FP4), a 4-bit floating-point format:
+
+- **E2M1** (1 sign bit, 2 exponent bits, 1 mantissa bit): Uses block-based microscaling where tensors are divided into blocks of 32 consecutive elements, with each block sharing a single 8-bit exponential scaling factor
+
+## Usage
+
+### Enabling Quantized KV Cache
+
+To enable quantized KV cache, use the `--kv-cache-dtype` argument when launching the server:
+
+```bash
+# Enable FP8 E5M2 KV cache
+python3 -m sglang.launch_server \
+    --model-path deepseek-ai/DeepSeek-R1-0528 \
+    --kv-cache-dtype fp8_e5m2 \
+
+# Enable FP8 E4M3 KV cache
+python3 -m sglang.launch_server \
+    --model-path deepseek-ai/DeepSeek-R1-0528 \
+    --kv-cache-dtype fp8_e4m3 \
+
+# Enable FP4 E2M1 KV cache
+python3 -m sglang.launch_server \
+    --model-path nvidia/DeepSeek-R1-0528-NVFP4 \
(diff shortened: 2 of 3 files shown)


docs/advanced_features/quantization.md
```diff
@@ -12,7 +12,7 @@ on-the-fly to convert high-precision weights into a lower-precision format.
 **Note: For better performance, usability and convenience, offline quantization is recommended over online quantization.**
 
 If you use a pre-quantized model, do not add `--quantization` to enable online quantization at the same time.
-For popular pre-quantized models, please visit [Unsloth](https://huggingface.co/unsloth), [ModelCloud](https://huggingface.co/collections/ModelCloud/vortex-673743382af0a52b2a8b9fe2)
```
Fridge003, line 15: Why changing this line?

docs/advanced_features/quantized_kv_cache.md
```diff
@@ -0,0 +1,124 @@
...
+Scaling factors can be:
+
+- **Loaded from checkpoints**: Pre-quantized models (e.g., ModelOpt) may include `k_scale` and `v_scale` parameters that are automatically loaded
+- **Provided via JSON**: Supply scaling factors via `--quantization-param-path`:
+
+```bash
+python3 -m sglang.launch_server \
+    --model-path deepseek-ai/DeepSeek-R1-0528 \
+    --kv-cache-dtype fp8_e4m3 \
+    --quantization-param-path /path/to/kv_cache_scales.json \
```
Fridge003: Have you verified this usage?

Fridge003 (changes requested):

docs/advanced_features/quantization.md
b8zhong: Err, it seems that they don't have many new models avail (and the # of downloads is low), so I choose to remove it for now
Screenshot 2025-12-03 at 12 04 22 PM

(if you feel it should be added back I can do that too, I don't have a preference)

docs/advanced_features/quantized_kv_cache.md
b8zhong: Err, good point, I have not verified for dpsk, so I'll remove it (generally, I find it's accurate enough w/o it anyway).

docs/advanced_features/quantized_kv_cache.md
```diff
@@ -0,0 +1,116 @@
...
+
+Quantized KV cache reduces the memory footprint of key-value cache storage by using lower-precision data types (FP8 or FP4) instead of the default model precision in BF16. During autoregressive generation, LLMs cache previously computed key-value pairs to avoid redundant calculations. The KV cache typically consumes a significant portion of GPU memory, especially for long sequences.
+
+Quantized KV cache is a memory optimization technique that primarily benefits throughput by allowing more tokens to be cached, but may introduce minimal accuracy degradation depending on the quantization format used.
+
+```{warning}
+**Performance Warning**: When quantized KV cache must be dequantized before use in attention operations, performance can be extremely slow if dequantization is not fused with the attention kernel. Always verify that your chosen attention backend supports quantized KV cache. Backends without fused support may experience significant throughput degradation, potentially negating the memory benefits.
+
+**Backend Support**: Not all attention backends support quantized KV cache. Refer to [Attention Backend](attention_backend.md) for which backends support it.
```
JackChuang, line 10: Yes, I'm currently summarizing which backends support KV4. I will publish the PR soon. Thank you for mentioning it here.

docs/advanced_features/quantized_kv_cache.md
```diff
@@ -0,0 +1,116 @@
...
+
+### FP4 Format
+
+```{warning}
+FP4 quantization is currently experimental.
+```
+
+[OCP (Open Compute Project)](https://www.opencompute.org) specifies MXFP4 (Microscaling FP4), a 4-bit floating-point format:
+
+- **E2M1** (1 sign bit, 2 exponent bits, 1 mantissa bit): Uses block-based microscaling where tensors are divided into blocks of 32 consecutive elements, with each block sharing a single 8-bit exponential scaling factor
```
JackChuang: In our current implementation, scale_block_size is set to 16 (#12612). We may wanna clarify this somewhere?

docs/advanced_features/quantized_kv_cache.md
```diff
@@ -0,0 +1,116 @@
...
+If scaling factors are not provided and not found in the checkpoint, it will default to 1.0, which may cause accuracy issues.
+```
+
+## Performance Considerations
+
+### Memory Savings
+
+Quantized KV cache provides significant memory savings:
+- **BF16 → FP8**: 2x reduction (16 bits → 8 bits)
+- **BF16 → FP4**: 4x reduction (16 bits → 4 bits)
```
JackChuang: This represents the ideal reduction. In practice, our KV4 implementation requires some memory for the KV scale buffer, so the BF16 → FP4 reduction should be stated as “4× minus KV scale”, and similarly for KV8.

For example, as mentioned in #12612:
```
- Supports significantly more tokens than KV8 (≈1.78×) and KV16 (≈3.56×) due to FP4 quantization with block_size = 16.
```

JackChuang: Please correct me if I'm wrong. Even if KV8’s scales are saved in the checkpoint, they still need to be loaded into memory for calculating, and thus occupy the same amount of space as I mentioned.

docs/advanced_features/quantized_kv_cache.md
```diff
@@ -0,0 +1,116 @@
...
+    --model-path deepseek-ai/DeepSeek-R1-0528 \
+    --kv-cache-dtype fp8_e4m3 \
+
+# Enable FP4 E2M1 KV cache
+python3 -m sglang.launch_server \
+    --model-path nvidia/DeepSeek-R1-0528-NVFP4 \
+    --kv-cache-dtype fp4_e2m1 \
+```
+
+### Scaling Factors
```
JackChuang, line 57: This part only mentions KV8. Maybe also mention KV4?

For your reference:
Our KV4 implementation handles the scaling factors on the fly, so there is no need to pre-quantize the models.

docs/advanced_features/quantized_kv_cache.md
```diff
@@ -0,0 +1,116 @@
...
+
+### Memory Savings
+
+Quantized KV cache provides significant memory savings:
+- **BF16 → FP8**: 2x reduction (16 bits → 8 bits)
+- **BF16 → FP4**: 4x reduction (16 bits → 4 bits)
+
+This enables longer context lengths or more concurrent requests within the same memory budget.
+
+### Accuracy Impact
```
JackChuang, line 109: If we want to provide detailed KV4 accuracy numbers, the preliminary FP4 accuracy test results can be found in the “Accuracy Tests” sections of these two PRs (#10078 and #12612).

JackChuang:
@b8zhong Thanks for putting together this information. I’ve reviewed all the FP4 KV-cache parts and left comments. Please take a look and let me know if you have any questions.

b8zhong:
@JackChuang Thanks! Btw, I'm not sure if you only left one comment (or if you maybe you did not click submit review on github)

JackChuang (commented):
@b8zhong Thanks for putting together this information. I’ve reviewed all the FP4 KV-cache parts and left comments. Please take a look and let me know if you have any questions.

docs/advanced_features/quantized_kv_cache.md
b8zhong: Good point, I have mentioned it

docs/advanced_features/quantized_kv_cache.md
b8zhong: Btw, are most of the SF for KV 8 not per tensor (i.e, it may be minimal compared to fp4 with block size scaling factor)? Also based on your number, (3.56 / 1.78) -> 2 (i may be wrong too)

docs/advanced_features/quantized_kv_cache.md
b8zhong: Alright, I added a note 👍

docs/advanced_features/quantized_kv_cache.md
b8zhong: 👍 It's very comprehensive. I have added it. Thanks.
Screenshot 2025-12-03 at 9 23 36 PM

Fridge003 (approved):
Thanks @b8zhong @JackChuang , nice document~

docs/advanced_features/quantized_kv_cache.md
JackChuang: You are right. Per-tensor will save more memory than I had calculated, closer to 2× for FP8. My assumption was that both KV4 and KV8 were using per-block (block_size = 16) in that case.

PR #14364, opened 2025-12-03
diffusion: Fix CLIP text encoder attention mask and causal masking bugFix clip attention

niehen6174:
## Motivation

The CLIP text encoder has a critical bug where it does not properly apply attention masks and causal masking, leading to incorrect output results. The issues are:
1. Missing causal masking: CLIP's original design uses causal attention, but the current implementation has causal=False, causing incorrect attention patterns
2. Missing attention_mask support: The attention_mask parameter is accepted but never passed to the attention layers, so padding tokens are not properly masked, leading to incorrect embeddings
3. Commented-out mask handling: The code for handling attention masks is commented out, indicating this was a known issue

This bug causes:
- Incorrect attention computation when padding tokens are present
- Deviation from CLIP's original causal attention behavior
- Mismatch with diffusers implementation, leading to different outputs

The implementation used in diffusers is in the transformer library, and the corresponding code is in [there](https://github.com/huggingface/transformers/blob/v4.39.0/src/transformers/models/clip/modeling_clip.py#L288).
### 1. Force TORCH_SDPA backend for CLIP 
- Modified CLIPTextArchConfig to only support TORCH_SDPA backend
- Rationale: The current LocalAttention implementation does not support attention_mask input. To minimize code changes and avoid modifying the entire LocalAttention infrastructure (which is used by many other components that don't require attention_mask), we restrict CLIP to use TORCH_SDPA backend, which natively supports both attention_mask and is_causal parameters.

### 2. Enable causal masking
- Changed causal=False to causal=True in CLIPAttention.__init__
- Restores CLIP's original causal attention behavior

## Accuracy Tests
Use the flux model, promot is "Female furry Pixie with text 'hello world'" and the seed is 1024. The lora use the` XLabs-AI/flux-furry-lora`. 
The results show a significant difference between the original image before restoration and the diffuser, and the effect of using Lora before restoration was worse.

image

## Checklist


mickqian (approved):
LGTM, but some comments could be cleaned

niehen6174:
> LGTM, but some comments could be cleaned

Done.

mickqian:
how will it affect previous models, and have you verified those?

PR #14370, opened 2025-12-03
[BUG]fix: environ missing get() error

jimmy-evo:
missing `get()`


hnyls2002:
Thanks @jimmy-evo. This is already fixed on current main: the check moved to `sglang/srt/managers/scheduler_components/invariant_checker.py`, which calls `envs.SGLANG_ENABLE_STRICT_MEM_CHECK_DURING_IDLE.get()` correctly, and the original mixin file no longer exists. Closing as fixed elsewhere; please reopen if I've missed something.

PR #14379, opened 2025-12-04
Add FP8 Blockwise GEMM Backend Flag `--fp8-gemm-backend`

### LINKED ISSUE #13773: [Feature] Add a FP8 Gemm backend for choosing FP8 gemm kernel
### Checklist

- [ ] If this is not a feature request but a general question, please start a discussion at https://github.com/sgl-project/sglang/discussions. Otherwise, it will be closed.
- [ ] Please use English. Otherwise, it will be closed.

### Motivation

Currently in SGLang, the FP8 Gemm kernels we use is controlled by a series of environment variables or implicit dispatching logics, as in https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/layers/quantization/fp8_utils.py#L151

To make a better control, we need a server argument like `--fp8-gemm-runner-backend`, similar to `--moe-runner-backend`

### Related resources

_No response_

b8zhong:
It could either be an env var, or a flag. Flag for now.

The selection logic is still the same.

Currently, we only do this for the Blockwise GEMM, I feel there can be an improved design for sharing this with the scaled MM, or maybe FP4, etc. but it can be done in the future.

Fix https://github.com/sgl-project/sglang/issues/13773


python/sglang/srt/layers/quantization/fp8_utils.py
```diff
@@ -1,13 +1,18 @@
...
+from enum import Enum
+from typing import TYPE_CHECKING, Callable, List, Optional, Tuple
 
 import torch
 
-from sglang.srt.environ import envs
 from sglang.srt.layers import deep_gemm_wrapper
 from sglang.srt.layers.quantization.fp8_kernel import sglang_per_token_group_quant_fp8
 from sglang.srt.layers.quantization.mxfp4_tensor import MXFP4QuantizeUtil
 from sglang.srt.utils import ceil_div, is_blackwell_supported, offloader
```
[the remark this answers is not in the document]
b8zhong: Nice

python/sglang/srt/managers/scheduler.py
```diff
@@ -65,6 +65,8 @@
 from sglang.srt.eplb.expert_distribution import get_global_expert_distribution_recorder
 from sglang.srt.layers.dp_attention import compute_dp_attention_world_info
 from sglang.srt.layers.moe import initialize_moe_config
+from sglang.srt.layers.quantization.fp8_gemm_backend import initialize_fp8_gemm_config
```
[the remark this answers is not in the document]
b8zhong: Nice

python/sglang/srt/server_args.py
```diff
@@ -2907,6 +2921,22 @@ def add_cli_args(parser: argparse.ArgumentParser):
             type=str,
             choices=NSA_CHOICES,
         )
+        parser.add_argument(
+            "--fp8-gemm-backend",
+            type=str,
+            choices=FP8_GEMM_RUNNER_BACKEND_CHOICES,
+            default=ServerArgs.fp8_gemm_runner_backend,
+            dest="fp8_gemm_runner_backend",
```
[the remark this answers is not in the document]
b8zhong: For internal consistency with the runner interface, I'll leave it. I left the cli flag name for ease of use (e.g, I feel this doesn't add anything there.)

python/sglang/srt/utils/common.py
```diff
@@ -225,11 +225,7 @@ def is_blackwell():
 
 @lru_cache(maxsize=1)
 def is_blackwell_supported(device=None) -> bool:
-    if not is_cuda_alike():
-        return False
-    return (torch.cuda.get_device_capability(device)[0] in [10, 12]) and (
-        torch.version.cuda >= "12.8"
-    )
+    return is_sm100_supported(device) or is_sm120_supported(device)
```
Fridge003, line 230: Is it OK to remove `torch.version.cuda >= "12.8"`?

docs/references/environment_variables.md
```diff
@@ -74,9 +74,9 @@ SGLang supports various environment variables that can be used to configure its
 | `SGLANG_INT4_WEIGHT` | Enable INT4 weight quantization | `false` |
 | `SGLANG_MOE_PADDING` | Enable MoE padding (sets padding size to 128 if value is `1`, often set to `1` in Docker builds) | `0` |
 | `SGLANG_FORCE_FP8_MARLIN` | Force using FP8 MARLIN kernels even if other FP8 kernels are available | `false` |
-| `SGLANG_ENABLE_FLASHINFER_FP8_GEMM` | Use flashinfer kernels when running blockwise fp8 GEMM on Blackwell GPUs | `false` |
+| `SGLANG_ENABLE_FLASHINFER_FP8_GEMM` (deprecated) | Use flashinfer kernels when running blockwise fp8 GEMM on Blackwell GPUs. **DEPRECATED**: Please use `--fp8-gemm-backend=flashinfer_trtllm` instead. | `false` |
```
Fridge003, line 81: Also mark them as deprecated in `environ.py`, with `_print_deprecated_env` function

python/sglang/srt/managers/scheduler.py
```diff
@@ -305,6 +306,9 @@ def __init__(
         # Init moe config
         self.init_moe_config()
 
+        # Init FP8 GEMM config
+        self.init_fp8_gemm_config()
```
Fridge003: We can rename it to `init_gemm_config`
So initialization of other gemm related arguments/environs (e.g. fp4 gemm) can also be included in this part

Fridge003 (changes requested):

python/sglang/srt/utils/common.py
b8zhong: is_sm100 checks torch cuda version, same w/ is_sm120

Fridge003:
Also please add a per-commit CI test (on 4-GPU Blackwell) for fp8 gemm.
It should cover Triton/DeepGemm/Flashinfer/TRTLLM kernels (cutlass is unimportant), and uses qwen3-fp8 as tested model maybe)

docs/references/environment_variables.md
b8zhong: Done 👍

python/sglang/srt/managers/scheduler.py
b8zhong: I did something like:

```python
diff --git a/python/sglang/srt/managers/scheduler.py b/python/sglang/srt/managers/scheduler.py
index 20eb6f962..bc7456b7d 100644
--- a/python/sglang/srt/managers/scheduler.py
+++ b/python/sglang/srt/managers/scheduler.py
@@ -306,8 +306,8 @@ class Scheduler(
         # Init moe config
         self.init_moe_config()
 
-        # Init FP8 GEMM config
-        self.init_fp8_gemm_config()
+        # Init GEMM config (FP8 GEMM, etc.)
+        self.init_gemm_config()
 
         # Check whether overlap can be enabled
         if not self.is_generation:
@@ -967,9 +967,10 @@ class Scheduler(
         if hasattr(self.model_config.hf_config, "num_experts_per_tok"):
             initialize_moe_config(self.server_args)
 
-    def init_fp8_gemm_config(self):
-        # Initialize FP8 GEMM backend configuration
-        # This is needed for FP8 quantization which can be used independently of MoE
+    def init_gemm_config(self):
+        # Initialize GEMM-related configuration (currently FP8 Blockwise GEMM backend).
+        # Other GEMM backends (e.g. FP4, BF16, etc.) can be added here in the future.
+        # This is needed for FP8 quantization.
         initialize_fp8_gemm_config(self.server_args)
 
     @DynamicGradMode()
     ```

**Comment by Fridge003** [comment]
cc @kaixih @leejnau We might need to fix some documents/recipes after this change, since the flashinfer fp8 gemm flag is deprecated.

**Comment by Fridge003** [comment]
@b8zhong Please fix conflicts

File: python/sglang/srt/layers/quantization/fp8_utils.py [hunk 7]
```diff
@@ -125,42 +133,145 @@ def normalize_e4m3fn_to_e4m3fnuz(
...
+
+def _dispatch_auto_backend() -> Callable:
+    """Auto-select the best backend based on hardware capabilities."""
+    # Priority order for auto selection:
+    # 1. DeepGEMM (if enabled and available)
+    # 2. FlashInfer TRTLLM (if Blackwell GPU and FlashInfer available)
+    # 3. CUTLASS (if Hopper+ GPU and CUDA 12.0+)
+    # 4. AITER (if AMD GPU with AITER enabled)
+    # 5. Triton (fallback)
+
```
Fridge003, line 252: Can we handle the two deprecated environs here?
For example, if the user passes in `SGLANG_ENABLE_FLASHINFER_FP8_GEMM`, we need to set fp8 gemm backend to flashinfer. So the users will not be impacted by this usage change

Fridge003 (changes requested):

python/sglang/srt/layers/quantization/fp8_utils.py
b8zhong: Yeah, 👍 

I added these back. I added a warning too, that they will completely be removed in 0.5.7

```python
def initialize_fp8_gemm_config(server_args: ServerArgs) -> None:
    """Initialize FP8 GEMM configuration."""
    global FP8_GEMM_RUNNER_BACKEND

    backend = server_args.fp8_gemm_runner_backend

    # TODO(brayden): Remove env-based overrides in v0.5.7, they will be fully removed in v0.5.7.
    if envs.SGLANG_ENABLE_FLASHINFER_FP8_GEMM.get():
        backend = "flashinfer_trtllm"
    elif envs.SGLANG_SUPPORT_CUTLASS_BLOCK_FP8.get():
        backend = "cutlass"

    FP8_GEMM_RUNNER_BACKEND = Fp8GemmRunnerBackend(backend)
 ```

File: python/sglang/srt/layers/quantization/fp8_utils.py [hunk 8]
```diff
@@ -125,42 +134,154 @@ def normalize_e4m3fn_to_e4m3fnuz(
...
+    """Initialize FP8 GEMM configuration."""
+    global FP8_GEMM_RUNNER_BACKEND
+
+    backend = server_args.fp8_gemm_runner_backend
+
+    # TODO(brayden): Remove env-based overrides in v0.5.7, they will be fully removed in v0.5.7.
+    if envs.SGLANG_ENABLE_FLASHINFER_FP8_GEMM.get():
+        backend = "flashinfer_trtllm"
+    elif envs.SGLANG_SUPPORT_CUTLASS_BLOCK_FP8.get():
+        backend = "cutlass"
```
kaixih: I suggest we add a warning message to clarify that this environment variable currently overrides any simultaneously configured server arguments.

However, since the server arguments represent the newer, long-term approach, and users setting them are likely aware of the change of the PR, we might want to prioritize the arguments instead, making them override the environment variable.

kaixih (commented):
Sorry for the late reply. Just leave a minor comment.

python/sglang/srt/layers/quantization/fp8_utils.py
b8zhong: Good point. If the server args is set I made the server args override any config provided by the env var. I think this makes the most sense.

kaixih (approved):

Fridge003 (approved):

python/sglang/srt/layers/quantization/fp8_utils.py
```diff
@@ -125,42 +137,166 @@ def normalize_e4m3fn_to_e4m3fnuz(
...
+def _dispatch_auto_backend() -> Callable:
+    """Auto-select the best backend based on hardware capabilities."""
+    # Priority order for auto selection:
+    # 1. DeepGEMM (if enabled and available)
+    # 2. FlashInfer TRTLLM (if Blackwell GPU and FlashInfer available)
+    # 3. CUTLASS (if Hopper+ GPU and CUDA 12.0+)
+    # 4. AITER (if AMD GPU with AITER enabled)
+    # 5. Triton (fallback)
+
+    if deep_gemm_wrapper.ENABLE_JIT_DEEPGEMM:
```
Fridge003, line 253: Please put deepgemm backend after flashinfer/cutlass backend. Since on Blackwell deepgemm can only be applied  to ue8m0 scale, and might cause accuracy drops on most models.

Fridge003 (dismissed):

Fridge003:
Most tests have passed
https://github.com/sgl-project/sglang/actions/runs/20042689711/job/57497202945?pr=14379

python/sglang/srt/layers/quantization/fp8_utils.py
b8zhong: (discuss on slack)

motivation for leaving it as dg:

1 currently since the default is deepgemm we leave it, 
2 also, the drop in gsm8k is minimal (it may be more for other evals)

Fridge003:
Failing b200 tests have been fixed on main

PR #14395, opened 2025-12-04
Support FP8 MLA prefill and 128k context.

weireweire:
## Motivation
MLA prefill always use 16-bit despite set `kv-cache-dtype` to fp8. We have fp8 kernel now, so this fixed it.

## Modifications
- convert qkv to fp8 to make attention faster.
- increase workspace size to make 128k runable.
- ~~remove `tile_tokens_dim` to be compatible with newer flashinfer.~~

## Accuracy Tests
```
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value|   |Stderr|
|-----|------:|----------------|-----:|-----------|---|----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.956|±  |0.0056|
|     |       |strict-match    |     5|exact_match|↑  |0.956|±  |0.0056|
```

## Benchmarking and Profiling
my command:
```
DYN_SKIP_SGLANG_LOG_FORMATTING=1  MC_FORCE_MNNVL=1  NCCL_CUMEM_ENABLE=1  NCCL_MNNVL_ENABLE=1  PYTHONUNBUFFERED=1  SGLANG_DECODE_BOOTSTRAP_TIMEOUT=1000  SGLANG_DISABLE_TP_MEMORY_INBALANCE_CHECK=1  SGLANG_ENABLE_FLASHINFER_GEMM=true  SGLANG_ENABLE_JIT_DEEPGEMM=false  SGLANG_MOONCAKE_CUSTOM_MEM_POOL=True  SGLANG_USE_MESSAGE_QUEUE_BROADCASTER=0  TORCH_DISTRIBUTED_DEFAULT_TIMEOUT=1800  SGLANG_TORCH_PROFILER_DIR=/logs/profiles/prefill python3 -m sglang.launch_server --attention-backend trtllm_mla --chunked-prefill-size -1 --context-length 136000 --cuda-graph-max-bs 4 --data-parallel-size 1 --disable-radix-cache --disaggregation-bootstrap-port 30001 --disaggregation-transfer-backend nixl --enable-symm-mem --expert-parallel-size 1 --kv-cache-dtype fp8_e4m3 --load-balance-method round_robin --max-running-requests 4 --max-total-tokens 544000 --mem-fraction-static 0.95 --model-path /model/ --moe-dense-tp-size 1 --moe-runner-backend flashinfer_trtllm --pipeline-parallel-size 4 --quantization modelopt_fp4 --scheduler-recv-interval 1 --served-model-name deepseek-ai/DeepSeek-R1 --stream-interval 10 --tensor-parallel-size 1 --trust-remote-code --watchdog-timeout 1000000 --dist-init-addr 10.66.5.212:29500 --nnodes 1 --node-rank 0 --host 0.0.0.0
```

## Checklist

### DIFF
--- a/python/sglang/srt/layers/attention/trtllm_mla_backend.py
+++ b/python/sglang/srt/layers/attention/trtllm_mla_backend.py
@@ -198,7 +198,9 @@ def unpad_draft_extend_output_kernel(
 def _quantize_fp8_qkv(q, k, v, layer):
     q = q.to(torch.float8_e4m3fn)
 
-    k_scale = getattr(layer, "k_scale_float", 1.0)
+    k_scale = getattr(layer, "k_scale_float", None)
+    if k_scale is None:
+        k_scale = 1.0
     if k_scale != 1.0:
         assert hasattr(layer, "k_scale"), "k_scale is not set"
         k_2d, _ = scaled_fp8_quant(
@@ -208,7 +210,9 @@ def _quantize_fp8_qkv(q, k, v, layer):
     else:
         k = k.to(torch.float8_e4m3fn)
 
-    v_scale = getattr(layer, "v_scale_float", 1.0)
+    v_scale = getattr(layer, "v_scale_float", None)
+    if v_scale is None:
+        v_scale = 1.0
     if v_scale != 1.0:
         assert hasattr(layer, "v_scale"), "v_scale is not set"
         v_2d, _ = scaled_fp8_quant(
--- a/python/sglang/srt/layers/attention/trtllm_mla_backend.py
+++ b/python/sglang/srt/layers/attention/trtllm_mla_backend.py
@@ -21,6 +21,7 @@ from sglang.srt.layers.attention.utils import (
     get_num_page_per_block_flashmla,
 )
 from sglang.srt.layers.dp_attention import get_attention_tp_size
+from sglang.srt.layers.quantization.fp8_kernel import scaled_fp8_quant
 from sglang.srt.model_executor.forward_batch_info import ForwardBatch, ForwardMode
 from sglang.srt.server_args import get_global_server_args
 from sglang.srt.utils import is_cuda, is_flashinfer_available, is_float4_e2m1fn_x2
@@ -39,7 +40,7 @@ if _is_cuda:
     from sgl_kernel import concat_mla_absorb_q
 
 # Constants
-DEFAULT_WORKSPACE_SIZE_MB = 128  # Memory workspace size in MB
+DEFAULT_WORKSPACE_SIZE_MB = 150  # Memory workspace size in MB
 
 # Block constraint from flashinfer requirements
 # From flashinfer.decode._check_trtllm_gen_mla_shape:
@@ -192,6 +193,32 @@ def unpad_draft_extend_output_kernel(
         data,
         mask=head_mask[:, None] & dim_mask[None, :],
     )
+
+
+def _quantize_fp8_qkv(q, k, v, layer):
+    q = q.to(torch.float8_e4m3fn)
+
+    k_scale = getattr(layer, "k_scale_float", 1.0)
+    if k_scale != 1.0:
+        assert hasattr(layer, "k_scale"), "k_scale is not set"
+        k_2d, _ = scaled_fp8_quant(
+            k.reshape(-1, k.shape[-1]).contiguous(), layer.k_scale
+        )
+        k = k_2d.reshape(k.shape)
+    else:
+        k = k.to(torch.float8_e4m3fn)
+
+    v_scale = getattr(layer, "v_scale_float", 1.0)
+    if v_scale != 1.0:
+        assert hasattr(layer, "v_scale"), "v_scale is not set"
+        v_2d, _ = scaled_fp8_quant(
+            v.reshape(-1, v.shape[-1]).contiguous(), layer.v_scale
+        )
+        v = v_2d.reshape(v.shape)
+    else:
+        v = v.to(torch.float8_e4m3fn)
+
+    return q, k, v, k_scale, v_scale
 
 
 global_zero_init_workspace_buffer = None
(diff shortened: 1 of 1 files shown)


python/sglang/srt/layers/attention/trtllm_mla_backend.py
```diff
@@ -259,7 +259,7 @@ def __init__(
         self.req_to_token = model_runner.req_to_token_pool.req_to_token
 
         # Workspace allocation
-        self.workspace_size = DEFAULT_WORKSPACE_SIZE_MB * 1024 * 1024
```
ishandhanani, line 262: Can you maybe leave a small comment why we do `*1.2`?

python/sglang/srt/layers/attention/trtllm_mla_backend.py
```diff
@@ -1069,6 +1069,19 @@ def forward_extend(
...
+        out = None
+        if self.data_type == torch.float8_e4m3fn:
+            out = torch.empty(
+                q.shape[0],
+                q.shape[1],
+                v.shape[2],
+                device=q.device,
+                dtype=torch.bfloat16,
+            )
+            q =q.to(torch.float8_e4m3fn)
```
nvpohanh: nit: add space after `=`

python/sglang/srt/layers/attention/trtllm_mla_backend.py
weireweire: 128k prefill length will run out of the workspace.

python/sglang/srt/layers/attention/trtllm_mla_backend.py
weireweire: fixed.

python/sglang/srt/layers/attention/trtllm_mla_backend.py
hlu1: How did you come up with 1.2?

python/sglang/srt/layers/attention/trtllm_mla_backend.py
```diff
@@ -1069,6 +1069,19 @@ def forward_extend(
...
+            out = torch.empty(
+                q.shape[0],
+                q.shape[1],
+                v.shape[2],
+                device=q.device,
+                dtype=torch.bfloat16,
+            )
+            q = q.to(torch.float8_e4m3fn)
+            k = k.to(torch.float8_e4m3fn)
+            v = v.to(torch.float8_e4m3fn)
```
hlu1: Is layer.scaling always 1.0? If not, do q and k need to be scaled accordingly?

python/sglang/srt/layers/attention/trtllm_mla_backend.py
nvpohanh: Maybe we should just increase `DEFAULT_WORKSPACE_SIZE_MB` instead of the magic number 1.2?

Or should we ask FlashInfer if they can provide some util function to calculate this instead of hard-coding it in SGL code?

python/sglang/srt/layers/attention/trtllm_mla_backend.py
```diff
@@ -1069,6 +1069,19 @@ def forward_extend(
...
             )
 
+        out = None
+        if self.data_type == torch.float8_e4m3fn:
+            out = torch.empty(
+                q.shape[0],
+                q.shape[1],
+                v.shape[2],
+                device=q.device,
+                dtype=torch.bfloat16,
```
nvpohanh: should this be hard-coded to `bfloat16` or should we just the model's dtype?

python/sglang/srt/layers/attention/trtllm_mla_backend.py
nvpohanh: currently, all the known models we have uses q_scale/k_scale/v_scale = 1.0f. But if the models contain these scales, we should use them.

Could you follow the example in https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/layers/attention/trtllm_mha_backend.py to use `layer.q_scale/k_scale/v_scale` if you need device scales or `layer.q_scale_float/k_scale_float/v_scale_float` if you need host scales?

Or just use `scaled_fp8_quant()`?
https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/layers/quantization/fp8_kernel.py#L1429-L1463

python/sglang/srt/layers/attention/trtllm_mla_backend.py
weireweire: Ideally we should follow https://github.com/NVIDIA/TensorRT-LLM/blob/main/cpp/tensorrt_llm/common/attentionOp.cpp#L893-L899 to calculate the size. But for now I'll just increase the DEFAULT_WORKSPACE_SIZE_MB

python/sglang/srt/layers/attention/trtllm_mla_backend.py
weireweire: I saw other place just use scale 1.0, like `quantize_and_rope_for_fp8`, it's called in decode and draft extend.

But I do can add a check and call `scaled_fp8_quant` with real scale or 1.0

weireweire: updated, please review

python/sglang/srt/layers/attention/trtllm_mla_backend.py
weireweire: fixed

python/sglang/srt/layers/moe/fused_moe_triton/layer.py
```diff
@@ -1225,7 +1225,6 @@ def forward(self, hidden_states: torch.Tensor, topk_output: TopKOutput):
             local_expert_offset=self.moe_ep_rank * self.num_local_experts,
             local_num_experts=self.num_local_experts,
             routed_scaling_factor=self.moe_runner_config.routed_scaling_factor,
-            tile_tokens_dim=None,
```
Fridge003, line 1229: This argument cannot be removed until the upgrade of flashinfer

python/sglang/srt/layers/attention/trtllm_mla_backend.py
```diff
@@ -1066,6 +1067,35 @@ def forward_extend(
                 out=torch.zeros(*output_shape, dtype=q.dtype, device=q.device),
             )
 
+        out = None
```
Fridge003: Do we need to reset the data type in the `forward_batch.attn_attend_prefix_cache=True` branch?
Since it also uses `flashinfer.prefill.trtllm_ragged_attention_deepseek` kernel

nvpohanh (approved):

python/sglang/srt/layers/moe/fused_moe_triton/layer.py
weireweire: reverted, let do that in the upgrade pr.

python/sglang/srt/layers/attention/trtllm_mla_backend.py
weireweire: make sense, added fp8 support for it.

Fridge003:
@weireweire please fix lint

weireweire:
lint fixed

Fridge003 (approved):

PR #14396, opened 2025-12-04
[spec] fix dp-attn + spec + trtllm-mla

hnyls2002:
To completely fix the issue, it requires

- [ ] Move most of the metadata sync (in `prepare_mlp_sync_batch`) into `Scheduler.prepare_mlp_sync_batch`.
- [ ] Maybe deprecate the API `skip_attn_backend_init`.
- [ ] Refactor `ForwardMode` into more reliable status representation.

### DIFF
--- a/python/sglang/srt/layers/attention/trtllm_mla_backend.py
+++ b/python/sglang/srt/layers/attention/trtllm_mla_backend.py
@@ -594,8 +594,6 @@ class TRTLLMMLABackend(FlashInferMLAAttnBackend):
                 seq_lens = seq_lens + self.num_draft_tokens
                 self.forward_decode_metadata.seq_lens_k = seq_lens.to(torch.int32)
             elif forward_batch.forward_mode.is_draft_extend(include_v2=True):
-                max_seq = forward_batch.seq_lens_cpu.max().item()
-
                 sum_seq_lens_q = sum(forward_batch.extend_seq_lens_cpu)
                 max_seq_len_q = max(forward_batch.extend_seq_lens_cpu)
                 cu_seqlens_q = torch.nn.functional.pad(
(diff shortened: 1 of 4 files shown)


Fridge003:
cc @rainj-me

python/sglang/srt/layers/attention/trtllm_mla_backend.py
```diff
@@ -985,25 +983,6 @@ def forward_extend(
...
-                    # Dynamic allocation for non-CUDA graph mode
-                    padded_q = torch.zeros(
-                        bs,
-                        metadata.max_seq_len_q,
-                        layer.tp_q_head_num,
-                        layer.head_dim,
-                        dtype=q.dtype,
-                        device=q.device,
-                    )
-                q = self.pad_draft_extend_query(
```
rainj-me, line 1004: After removing the pad_draft_extend_query call, maybe we should already delete the relate code.

python/sglang/srt/layers/attention/trtllm_mla_backend.py
```diff
@@ -1022,15 +1001,6 @@ def forward_extend(
                 bmm1_scale=bmm1_scale,
             )
 
-            # Reshape output directly without slicing
-
-            if forward_batch.forward_mode.is_draft_extend(include_v2=True):
-                raw_out = self.unpad_draft_extend_output(
```
rainj-me, line 1028: Same here.

python/sglang/srt/speculative/eagle_worker_v2.py
```diff
@@ -512,7 +512,7 @@ def _draft_extend_for_decode(
             )
         else:
             draft_logits_output, _ = self.draft_runner.forward(
-                forward_batch, skip_attn_backend_init=True
+                forward_batch, skip_attn_backend_init=False
```
rainj-me, line 515: skip attn backend lead to the flashinfer backend cudagraph capture failure.

PR #14413, opened 2025-12-04
[NPU]minicpm3-4B for AscendAttnBackend

McZyWu:
## Motivation

1. previously npu_fused_infer_attention_score doesnot support this model.
2. for dp attn, if first dimension is 0, it will fail to reshape.
3. model config.json does not have index_head_dim.
4. it does not  pass qk_nope_head_di to model
## Modifications
as follow.

#supporting for A3,A2, tp 8

### DIFF
--- a/python/sglang/srt/layers/quantization/marlin_utils_fp8.py
+++ b/python/sglang/srt/layers/quantization/marlin_utils_fp8.py
@@ -62,7 +62,6 @@ def apply_fp8_marlin_linear(
         a=reshaped_x,
         c=None,
         b_q_weight=weight,
-        b_bias=bias,
         b_scales=weight_scale,
         global_scale=None,
         b_zeros=None,
@@ -76,6 +75,9 @@ def apply_fp8_marlin_linear(
         use_atomic_add=use_atomic_add,
         use_fp32_reduce=use_fp32_reduce,
     )
+
+    if bias is not None:
+        output.add_(bias)
 
     return output.reshape(out_shape)
 


python/sglang/srt/hardware_backend/npu/attention/ascend_backend.py
```diff
@@ -591,37 +592,64 @@ def forward_extend(
...
-
-            # Wait for the KV transfer to complete before performing attention computation.
             forward_batch.token_to_kv_pool.get_key_buffer(layer.layer_id)
             forward_batch.token_to_kv_pool.get_value_buffer(layer.layer_id)
 
             num_token_padding = q.shape[0]
             q, k, v = [
                 data[: forward_batch.num_token_non_padded_cpu] for data in [q, k, v]
             ]
+            if layer.tp_q_head_num%16==0:
```
khalil2ji3mp6: The value 16 here should not be hard-coded.

python/sglang/srt/hardware_backend/npu/attention/ascend_backend.py
```diff
@@ -591,37 +592,64 @@ def forward_extend(
...
+                assert (
+                    layer.qk_head_dim != layer.v_head_dim
+                ), "FIA only supports qk_head_dim != v_head_dim"
 
-            q_nope, q_rope = q.split([layer.v_head_dim, self.qk_rope_head_dim], dim=-1)
-            k_nope, k_rope = k.split([layer.v_head_dim, self.qk_rope_head_dim], dim=-1)
+                # Wait for the KV transfer to complete before performing attention computation.
```
khalil2ji3mp6: The comment here is unrelated to the code logic and can be removed.

python/sglang/srt/hardware_backend/npu/attention/ascend_backend.py
```diff
@@ -591,37 +592,64 @@ def forward_extend(
...
+                    )
+                query = q.reshape(-1, layer.tp_q_head_num, layer.qk_head_dim)
+                attn_output = torch.empty(
+                    (query.shape[0],layer.tp_q_head_num, layer.v_head_dim),
+                    dtype = query.dtype,
+                    device = query.device
+                )
+
+                max_len = max(self.forward_metadata.seq_lens_cpu_int)
+                self.ascend_attn_mask_builder.generate_attn_mask(max_len, "norm", self.mask_dtype)
```
khalil2ji3mp6: move generate_attn_mask to AscendAttnMaskBuilder init

python/sglang/srt/hardware_backend/npu/attention/ascend_backend.py
```diff
@@ -225,6 +225,7 @@ def __init__(self, model_runner: ModelRunner):
         self.req_to_token = model_runner.req_to_token_pool.req_to_token
         self.graph_mode = False
         self.use_fia = get_bool_env_var("ASCEND_USE_FIA", "False")
+        self.mask_dtype = model_runner.dtype
```
khalil2ji3mp6: You can remove this line of code; model_runner.dtype can be used directly in the __init__ function of AscendAttnMaskBuilder.

PR #14415, opened 2025-12-04
[bugfix][quark] Fixed an issue where `per_token` could not be properly recognized when the token count was 1.

haoyangli0109:
When the number of token is 1, per-token scale is [[1]], per-tensor scale is [1] or ().
This issue indirectly causes
when we use aiter to perform preshuffle on weight, if the number of tokens in inference equals 1, aiter's bpreshuffle will not be invoked correctly, resulting in garbled output.

The script that reproduces the error is as follows:

```
"""
Usage:
python3 offline_batch_inference.py  --model meta-llama/Llama-3.1-8B-Instruct
"""

import argparse
import dataclasses

import sglang as sgl
from sglang.srt.server_args import ServerArgs


def main(
    server_args: ServerArgs,
):
    # Sample prompts.
    prompts = [
        "Hello, my name is",
    ]
    # Create a sampling params object.
    sampling_params = {"temperature": 0.8, "top_p": 0.95}

    # Create an LLM.
    llm = sgl.Engine(**dataclasses.asdict(server_args))

    outputs = llm.generate(prompts, sampling_params)
    # Print the outputs.
    for prompt, output in zip(prompts, outputs):
        print("===============================")
        print(f"Prompt: {prompt}\nGenerated text: {output['text']}")


# The __main__ condition is necessary here because we use "spawn" to create subprocesses
# Spawn starts a fresh program every time, if there is no __main__, it will run into infinite loop to keep spawning processes from sgl.Engine
if __name__ == "__main__":
    parser = argparse.ArgumentParser()
    ServerArgs.add_cli_args(parser)
    args = parser.parse_args()
    server_args = ServerArgs.from_cli_args(args)
    main(server_args)

```


BowenBao (approved):
LGTM, thanks for the fix!

BowenBao:
@HaiShaw please review and help land this PR, thanks!

python/sglang/srt/layers/quantization/fp8_utils.py
```diff
@@ -845,7 +845,9 @@ def apply_fp8_linear(
     # torch.scaled_mm supports per tensor weights + activations only
     # so fallback to naive if per channel or per token
     per_tensor_weights = weight_scale.numel() == 1
-    per_tensor_activations = x_scale.numel() == 1
+    # When the number of token is 1,
+    # per-token scale is [[1]], per-tensor scale is [1] or ().
```
BowenBao: Per discussion with @HaiShaw, @haoyangli-amd let's update this comment to be more specific.


```suggestion
    # per-token scale has shape (1, 1), per-tensor scale has shape (1) or ().
```

HaiShaw (approved):

PR #14419, opened 2025-12-04
dst_attn_tp_size == 0 will trigger exception

strenuous-life:
Eg: The model of qwen3-32B num_key_value_heads is 8. If prefill tp is 8 and decode tp is 16, the value of dst_heads_per_rank will be zero

## Checklist

### DIFF
--- a/python/sglang/srt/disaggregation/mooncake/conn.py
+++ b/python/sglang/srt/disaggregation/mooncake/conn.py
@@ -411,6 +411,14 @@ class MooncakeKVManager(CommonKVManager):
         # Calculate head distribution
         src_heads_per_rank = num_kv_heads
         dst_heads_per_rank = num_kv_heads * self.attn_tp_size // dst_attn_tp_size
+        if dst_attn_tp_size == 0:
+            logger.error(
+                f"[{num_kv_heads}] num_kv_heads is ({num_kv_heads})"
+                f"[{self.attn_tp_size}] self.attn_tp_size is ({self.attn_tp_size}) "
+                f"dst_attn_tp_size is ({dst_attn_tp_size})."
+                f"dst_heads_per_rank will be set to zero which will trigger an exception."
+            )
+            return -1
         bytes_per_head_slice_to_send = (
             dst_kv_item_len // page_size // dst_heads_per_rank
         )


python/sglang/srt/disaggregation/mooncake/conn.py
```diff
@@ -411,6 +411,14 @@ def send_kvcache_slice(
         # Calculate head distribution
         src_heads_per_rank = num_kv_heads
         dst_heads_per_rank = num_kv_heads * self.attn_tp_size // dst_attn_tp_size
+        if dst_attn_tp_size == 0:
```
ShangmingCai, line 414: Can you elaborate more about why `dst_attn_tp_size` could equal 0? I think unless you have prefill attn_tp_size == 1 and num_kv_heads == 8 while dst_attn_tp_size == 16, but how come could num_kv_heads == 8 and dst_attn_tp_size == 16 happens?

strenuous-life: 1) prefill start with --tp 8 and decode start with --tp 16.

2) In Qwen3-32B config.json, num_key_value_heads is 8. 

3) According to function get_num_kv_heads and get_total_num_kv_heads, the result  of head_num is 1
def get_num_kv_heads(self, tensor_parallel_size):
    total_num_kv_heads = self.get_total_num_kv_heads()
    return max(1, total_num_kv_heads // tensor_parallel_size)
def get_total_num_kv_heads(self)
        attributes = [
            # For Falcon:
            "n_head_kv",
            "num_kv_heads",
            # For LLaMA-2:
            "num_key_value_heads",
            # For ChatGLM:
            "multi_query_group_num",
            # For Step3
            "num_attention_groups",
        ]
        for attr in attributes:
            num_kv_heads = getattr(self.hf_text_config, attr, None)
            if num_kv_heads is not None:
                return num_kv_heads

4) Then kv_args.kv_head_num = self.token_to_kv_pool.head_num in function _init_kv_manager

5) So in function of send_kvcache_slice, num_kv_heads = self.kv_args.kv_head_num, num_kv_heads is 1
      self.attn_tp_size is 8, and dst_attn_tp_size is 16

6) In the end, the result of num_kv_heads * self.attn_tp_size // dst_attn_tp_size  is 0

ShangmingCai: > 1. prefill start with --tp 8 and decode start with --tp 16.

But decode with --tp 16 and num_key_value_heads is 8 would make num_kv_heads at the decode side 0.5, is this a possible setup? I think this would be invalid at the first place.

strenuous-life: Sorry for replying to you so late. I created an issue to describe this scenario. https://github.com/sgl-project/sglang/issues/14709

ShangmingCai: Can you verify that Qwen3-32B with tp==16 can be launched successfully without PD disaggregation?

hnyls2002:
Thanks @strenuous-life. The dst_heads_per_rank == 0 scenario described here (e.g. 8 KV heads with decode TP 16) was fixed on main by #19086, which rewrote the slice transfer to use `max(1, total_kv_heads // dst_attn_tp_size)` in `mooncake/conn.py`. Closing as superseded; please reopen if I've missed something.

PR #14459, opened 2025-12-05
Add Mistral Large 3 to nightly CI tests

alisonshao:
## Summary
- Add nightly performance test for `mistralai/Mistral-Large-3-675B-Instruct-2512`
- Test configuration:
  - TP=8
  - `--attention-backend trtllm_mla`
  - `--model-loader-extra-config '{"enable_multithread_load": true}'`
  - `--chat-template mistral`
  - `SGLANG_ENABLE_JIT_DEEPGEMM=0`

## Test plan
- [ ] Verify nightly CI workflow runs successfully on 8-gpu-h200 runner
- [ ] Check performance traces are published correctly

### DIFF
--- a/.github/workflows/nightly-test-nvidia.yml
+++ b/.github/workflows/nightly-test-nvidia.yml
@@ -420,6 +420,13 @@ jobs:
         run: |
           IS_BLACKWELL=1 bash scripts/ci/ci_install_dependency.sh
 
+      - name: Pre-download Mistral-Large-3 model
+        timeout-minutes: 120
+        run: |
+          echo "Pre-downloading Mistral-Large-3 model to ensure cache is complete..."
+          python3 -c "from huggingface_hub import snapshot_download; snapshot_download('mistralai/Mistral-Large-3-675B-Instruct-2512')"
+          echo "Download complete."
+
       - name: Run Mistral-Large-3 nightly performance test
         timeout-minutes: 180
         env:
--- a/.github/workflows/nightly-test-nvidia.yml
+++ b/.github/workflows/nightly-test-nvidia.yml
@@ -420,7 +420,28 @@ jobs:
         run: |
           IS_BLACKWELL=1 bash scripts/ci/ci_install_dependency.sh
 
+      - name: Run Mistral-Large-3 nightly performance test
+        timeout-minutes: 180
+        env:
+          TRACE_BASE_URL: https://raw.githubusercontent.com/sglang-bot/sglang-ci-data/main/traces/${{ github.run_id }}
+          PERFETTO_RELAY_URL: ${{ vars.PERFETTO_RELAY_URL }}
+          GPU_CONFIG: "8-gpu-b200"
+          SGLANG_ENABLE_JIT_DEEPGEMM: "0"
+        run: |
+          rm -rf test/performance_profiles_mistral_large3/
+          cd test
+          IS_BLACKWELL=1 python3 nightly/test_mistral_large3_perf.py
+
+      - name: Publish Mistral-Large-3 traces to storage repo
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_PAT_FOR_NIGHTLY_CI_DATA }}
+          GITHUB_RUN_ID: ${{ github.run_id }}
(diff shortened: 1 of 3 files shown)


alisonshao:
testing at: https://github.com/sgl-project/sglang/actions/runs/19948050892/job/57201949750
waiting for result

test/nightly/test_mistral_large3_perf.py
```diff
@@ -0,0 +1,67 @@
...
+        cls.runner = NightlyBenchmarkRunner(PROFILE_DIR, cls.__name__, cls.base_url)
+        cls.runner.setup_profile_directory()
+
+    @classmethod
+    def tearDownClass(cls):
+        # Clean up environment variable
+        if "SGLANG_ENABLE_JIT_DEEPGEMM" in os.environ:
+            del os.environ["SGLANG_ENABLE_JIT_DEEPGEMM"]
+
+    def test_bench_one_batch(self):
```
ispobock, line 56: can we also add accuracy test for it?

python/sglang/srt/model_loader/weight_utils.py
```diff
@@ -309,8 +311,22 @@ def find_local_hf_snapshot_dir(
...
+    # if local snapshot exists, validate it contains at least one weight file
+    # matching allow_patterns before skipping download.
+    if found_local_snapshot_dir is None:
+        return None
+
+    # Use file lock to prevent multiple processes (TP ranks) from
+    # validating and cleaning up the same model cache simultaneously.
+    # This prevents race conditions where multiple ranks detect corruption
+    # and try to delete the same files at the same time.
+    with get_lock(model_name_or_path, cache_dir, suffix="-validation"):
```
ispobock, line 323: Why do we need this change for Mistral large 3 model?

.github/workflows/nightly-test-nvidia.yml
```diff
@@ -420,7 +420,35 @@ jobs:
         run: |
           IS_BLACKWELL=1 bash scripts/ci/ci_install_dependency.sh
 
+      - name: Pre-download Mistral-Large-3 model
```
ispobock: Why other models don't have the pre-download step?

python/sglang/srt/model_loader/weight_utils.py
alisonshao: it's a general bug fix discovered while debugging the Mistral test. When model weights are corrupted/missing, all 8 TP ranks detect the issue simultaneously and race to delete the cache directory. One rank succeeds, others fail with "No such file or directory" errors. This was causing CI failures.
the fix adds a file lock so only one rank performs validation and cleanup at a time. other models just happened to have complete caches so the bug wasn't triggered.

.github/workflows/nightly-test-nvidia.yml
alisonshao: We can remove it now, if preferred.
Other models (DeepSeek, Qwen3-235B, etc.) were already fully cached on the B200 runner from previous runs. Mistral is a newly added model that wasn't cached yet, so I added this step to support the first time running this model.

test/nightly/test_mistral_large3_perf.py
alisonshao: Yes, i can do that. What score threshold should we expect? How do we determine this value?

alisonshao: just added test_accuracy_mgsm method using the same eval framework as other text models. Set initial threshold to 0.90 (placeholder). Will calibrate the threshold after observing actual model performance.

alisonshao: testing at: https://github.com/sgl-project/sglang/actions/runs/19961215892/job/57241902940

alisonshao: MGSM accuracy for mistralai/Mistral-Large-3-675B-Instruct-2512: 0.972 
https://github.com/sgl-project/sglang/actions/runs/19961215892/job/57241902940#step:5:2471

ispobock (approved):

.github/workflows/nightly-test-nvidia.yml
ispobock: I see

Issue #14443, opened 2025-12-04
[Feature] Default to lower page size when not set


b8zhong:
### Checklist

- [x] If this is not a feature request but a general question, please start a discussion at https://github.com/sgl-project/sglang/discussions. Otherwise, it will be closed.
- [x] Please use English. Otherwise, it will be closed.

### Motivation

TRTLLM MLA can do 32 or 64, but it defaults to 64

TRTLLM MHA supports 16, 32, or 64, but it also defaults to 64

We can do 32 for MLA, 16 for MHA by default?

b8zhong:
Closing for now because it might affect PD scenario performance while only rarely benefiting prefix cache

PR #14474, opened 2025-12-05
feat: raise error in PD when page sizes are mismatched

### LINKED ISSUE #14439: [Feature] Raise an error when the page size of the prefill and decode attention backend are different
### Checklist

- [x] If this is not a feature request but a general question, please start a discussion at https://github.com/sgl-project/sglang/discussions. Otherwise, it will be closed.

### Motivation

The value of the prefill and decode `--page-size` must match, or else raise an error. Sometimes this is enforced by a backend which automatically selects this, so we can see the difference in the logs

Otherwise, the error can be cryptic (like in the case of TRTLLM MHA), which looks for page size cubin that does not exist.

### Related resources

_No response_

raayandhar:
## Motivation
See fix https://github.com/sgl-project/sglang/issues/14439

## Modifications

To check that the page sizes are the same, we have to communicate this between the prefill and decode servers, so I did that and on the prefill side it will throw an error. But I'm not really entirely happy with this design since the decode server is left hanging and the request will also hang. We should be throwing the error on both sides I think. Also, it only does this once we send a request, not during warmup / server start. I would really appreciate some feedback here as I'm sure there's a more intelligent way of doing this. I will continue working on that part.

Example (on prefill):
```
Exception in thread Thread-3 (bootstrap_thread):
  File "/usr/lib/python3.12/threading.py", line 1073, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.12/threading.py", line 1010, in run
    self._target(*self._args, **self._kwargs)
  File "/home/raayan/projects/sglang/python/sglang/srt/disaggregation/mooncake/conn.py", line 865, in bootstrap_thread
    raise ValueError(
ValueError: Page size mismatch: decode server has page_size=64, but prefill server has page_size=32. Both servers must use the same --page-size value.
```

## Checklist


python/sglang/srt/disaggregation/mooncake/conn.py
```diff
@@ -113,6 +113,7 @@ class KVArgsRegisterInfo:
     dst_tp_rank: int
     dst_attn_tp_size: int
     dst_kv_item_len: int
+    page_size: int
```
ShangmingCai: Maybe this would be better?
```suggestion
    dst_page_size: int
```

python/sglang/srt/disaggregation/mooncake/conn.py
```diff
@@ -127,6 +128,7 @@ def from_zmq(cls, msg: List[bytes]):
             dst_tp_rank=int(msg[7].decode("ascii")),
             dst_attn_tp_size=int(msg[8].decode("ascii")),
             dst_kv_item_len=int(msg[9].decode("ascii")),
+            page_size=int(msg[10].decode("ascii")),
```
ShangmingCai: ditto

python/sglang/srt/disaggregation/nixl/conn.py
```diff
@@ -74,6 +74,7 @@ class KVArgsRegisterInfo:
     decode_tp_size: int
     decode_tp_rank: int
     dst_kv_item_len: int
+    page_size: int
```
ShangmingCai: ditto

python/sglang/srt/disaggregation/nixl/conn.py
```diff
@@ -89,6 +90,7 @@ def from_zmq(cls, msg: List[bytes]):
             decode_tp_size=int(msg[8].decode("ascii")),
             decode_tp_rank=int(msg[9].decode("ascii")),
             dst_kv_item_len=int(msg[10].decode("ascii")),
+            page_size=int(msg[11].decode("ascii")),
```
ShangmingCai: ditto

python/sglang/srt/server_args.py
```diff
@@ -1388,9 +1388,9 @@ def _handle_attention_backend_compatibility(self):
 
             if self.page_size not in [32, 64]:
                 logger.warning(
-                    f"TensorRT-LLM MLA only supports page_size of 32 or 64, changing page_size from {self.page_size} to 64."
+                    f"TensorRT-LLM MLA only supports page_size of 32 or 64, changing page_size from {self.page_size} to 32."
                 )
-                self.page_size = 64
+                self.page_size = 32
```
ShangmingCai: Can you eloborate on this change?

python/sglang/srt/disaggregation/mooncake/conn.py
raayandhar: sure, will make the change tomorrow

ShangmingCai (commented):
The motivation has some points. But I don't think we should crash the prefill server. This could cause a chain reaction, leading to the entire cluster crashing due to the addition of only a misconfigured node. This is something we don't want to see. Maybe a warning should be enough.

python/sglang/srt/server_args.py
raayandhar: Sure, this is from this issue: https://github.com/sgl-project/sglang/issues/14443 that was opened.

ShangmingCai:
Sorry, I realize that the prefill page size wasn't registered to the bootstrap server, so the decode node doesn't know.

raayandhar:
> Maybe this check covers this scenario:
> 
> ```python
> # Sanity check: The data sub-slice to be sent should fit into the dst buffer.
>         # This means heads_bytes_per_token_to_send <= (dst_kv_item_len // page_size)
>         if heads_bytes_per_token_to_send > (dst_kv_item_len // page_size):
>             logger.error(
>                 f"[{mooncake_session_id}] slice size ({heads_bytes_per_token_to_send}) exceeds "
>                 f"target token slot size ({dst_kv_item_len // page_size})"
>             )
>             return -1
> ```

hmm, I will go and try tomorrow to see what happens and understand better. But I think at least some better messaging pointing to the issue (!= page sizes) could be nice.

b8zhong (commented):
When we are in the single node scenario we can use different attention backend for prefill and decode (which can have mismatched page sizes when incorrectly set), can you make sure it also works in the single node case?

raayandhar:
Sorry for lack of update, will work on this tomorrow

raayandhar:
I made some changes to this PR so that we can do this check in the common backend. Now the decode server gets runtime error with:
```
[2025-12-17 22:30:16] Page size mismatch: prefill server has page_size=32, but decode server has page_size=64. Both servers must use the same --page-size value.
```
I also tested this with different attention backends (triton / flashinfer different across prefill/decode). I have some comment about the recommended way to throw error which I made in the review.
@ShangmingCai when you get the chance, could you please take another look? Thanks!

python/sglang/srt/disaggregation/common/conn.py
```diff
@@ -256,19 +259,33 @@ def __init__(
...
+
+            if self.prefill_page_size is not None:
+                decode_page_size = self.kv_mgr.kv_args.page_size
+                if self.prefill_page_size != decode_page_size:
+                    error_msg = (
+                        f"Page size mismatch: prefill server has page_size={self.prefill_page_size}, "
+                        f"but decode server has page_size={decode_page_size}. "
+                        f"Both servers must use the same --page-size value."
+                    )
+                    logger.error(error_msg)
```
raayandhar, line 271: Previously here I did:
```
                    self.kv_mgr.record_failure(self.bootstrap_room, error_msg)
                    self.kv_mgr.update_status(self.bootstrap_room, KVPoll.Failed)
                    self.bootstrap_infos = None
                    return
```
but this just seemed to let all the servers hang until timeout. Right now the decode server crashes. I would really appreciate any feedback on what the right way to throw / do the error messaging should be?

ShangmingCai: We should never crash the prefill server, but when the decode server finds that its page size does not match to prefill's page size, it should crash and notify the users about this msg.

python/sglang/srt/disaggregation/common/conn.py
```diff
@@ -433,15 +453,16 @@ def _get_prefill_parallel_info_from_server(
                     int(prefill_parallel_info["prefill_attn_tp_size"]),
                     int(prefill_parallel_info["prefill_dp_size"]),
                     int(prefill_parallel_info["prefill_pp_size"]),
+                    prefill_parallel_info.get("prefill_page_size"),
```
ShangmingCai: Maybe align the code style here?

python/sglang/srt/disaggregation/common/conn.py
```diff
@@ -525,6 +547,7 @@ async def _handle_route_put(self, request: web.Request):
         system_dp_rank = data["system_dp_rank"]
         rank_ip = data["rank_ip"]
         rank_port = int(data["rank_port"])
+        page_size = data.get("page_size")  # optional for backward compatibility
```
ShangmingCai: ditto

python/sglang/srt/disaggregation/mooncake/conn.py
```diff
@@ -855,9 +855,8 @@ def bootstrap_thread():
                 room = waiting_req_bytes[0].decode("ascii")
                 mooncake_session_id = waiting_req_bytes[3].decode("ascii")
                 if room == "None":
-                    self.decode_kv_args_table[mooncake_session_id] = (
-                        KVArgsRegisterInfo.from_zmq(waiting_req_bytes)
-                    )
+                    register_info = KVArgsRegisterInfo.from_zmq(waiting_req_bytes)
+                    self.decode_kv_args_table[mooncake_session_id] = register_info
```
ShangmingCai: This change doesn't seem necessary now.

ShangmingCai (commented):
Others LGTM. Please fix the comments, then I can trigger the CI.

python/sglang/srt/disaggregation/common/conn.py
raayandhar: Thanks, good catch

ShangmingCai (approved):
Thx. LGTM now.

raayandhar:
@ShangmingCai when you get a chance, could you take a look at the CI failures? I took a look and they do not look related to my changes. Let me know what you think. Happy holidays!

raayandhar:
> /rerun-failed-ci

@ShangmingCai we didn't run into that error again (`test_disaggregation_basic` passing), and the failures are new I think so I'm inclined to think it's due to flaky CI -- let me know what you think

PR #14502, opened 2025-12-05
[Qwen3-Next]Optimize piecewise CUDA graph for Qwen3-Next

Chen-0210:
## Motivation
Based on https://github.com/sgl-project/sglang/pull/13081, this PR reduce TTFT  when piecewise CUDA graph is enabled.
The previous version kept the entire linear-attention layer outside CUDA graphs.
In this update, only the conv and GDN parts stay outside, while the in/out projection and fused_gdn_gating are moved into CUDA graphs.

## Modifications

1. Replace the previous split ops with new split `causal_conv1d_gdn` ops.
2. Remove `vars()` from memory pool state to be compatible with` torch.compile`.

May need to add a hybrid attention metadata, as it currently makes the function argument list too long.
## Accuracy Tests
```
python3 -m sglang.launch_server --model Qwen/Qwen3-Next-80B-A3B-Instruct/ --tp 2 --enable-piecewise-cuda-graph --piecewise-cuda-graph-compiler eager 
```

```
python3 benchmark/gsm8k/bench_sglang.py --parallel 1319 --num-questions 1319 
100%|██████████████████████| 1319/1319 [02:20<00:00,  9.38it/s]
Accuracy: 0.948
Invalid: 0.000
```

## Benchmarking and Profiling(Mean TTFT on bs=1 and H200x2)

`python3 -m sglang.bench_serving --backend sglang --dataset-name random --num-prompts 50 --random-input-len 4096 --random-output-len 1 --random-range-ratio 1 --max-concurrency 1`

| **Input length** | **1024** | **2048** | **4096** |
|:----------------:|:--------:|:--------:|:--------:|
| **without piecewise-cudagraph** | 99.17 ms | 101.52 ms | 127.20 ms |
| **with piecewise-cudagraph** | 67.83 ms | 68.72 ms | 105.29 ms |
| **with piecewise-cudagraph optimized** | 48.21 ms | 64.33    ms | 102.53 ms |

## Checklist


python/sglang/srt/model_executor/piecewise_cuda_graph_runner.py
```diff
@@ -313,10 +313,10 @@ def warmup_torch_compile(self):
                 encoder_lens=None,
                 return_logprob=False,
                 extend_num_tokens=num_tokens,
-                extend_seq_lens=torch.tensor([num_tokens], device=self.device),
+                extend_seq_lens=torch.tensor([1], device=self.device),
```
yizhang2077: why do we need change here? the same as following

Chen-0210: has some issues, fix it. intend to sets `extend_prefix_lens` from `num_token` to 0. Actually there is no prefix.

Chen-0210: Since it doesn’t affect this PR, I’ll move it to another one.

python/sglang/srt/model_executor/piecewise_cuda_graph_runner.py
```diff
@@ -569,7 +569,7 @@ def replay_prepare(
             encoder_lens=forward_batch.encoder_lens,
             return_logprob=False,
             extend_seq_lens=forward_batch.extend_seq_lens,
-            extend_prefix_lens=forward_batch.extend_prefix_lens,
+            extend_prefix_lens=forward_batch.extend_prefix_lens.long(),
```
yizhang2077: why do we need long here?

Chen-0210:
@gogongxt yes, not supported yet. I'm working on  https://github.com/sgl-project/sglang/issues/17039 for Qwen3-next + PCG. I will continue after it is fixed.

PR #14535, opened 2025-12-06
[Bug fix] Add /model_info endpoint to mini_lb

alisonshao:
## Summary
- Add `/model_info` endpoint to mini_lb to fix regression introduced by #13607
- Keep `/get_model_info` for backwards compatibility  
- Update backend call to use `/model_info` (the non-deprecated endpoint)

## Root Cause
PR #13607 changed `runtime_endpoint.py` to use `/model_info` instead of `/get_model_info`, but `mini_lb.py` was not updated to support the new endpoint. This caused `test_disaggregation_basic.py` to fail with 404 errors when accessing `/model_info` through the mini load balancer.

Error example: https://github.com/sgl-project/sglang/actions/runs/19977384736/job/57312528074

## Test plan
- [x] Verify `test_disaggregation_basic.py` passes in CI after this fix

### DIFF
--- a/sgl-model-gateway/bindings/python/sglang_router/mini_lb.py
+++ b/sgl-model-gateway/bindings/python/sglang_router/mini_lb.py
@@ -341,8 +341,7 @@ async def get_server_info():
         }
 
 
-@app.get("/get_model_info")
-async def get_model_info():
+async def _get_model_info_impl():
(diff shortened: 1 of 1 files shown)


b8zhong (approved):
Thanks, I think it's the right fix.

slin1237 (approved):

PR #14543, opened 2025-12-06
diffusion: fix LoRA dtype handling and weight attribute access for z-image model

niehen6174:
## Motivation

This PR fixes two critical issues when using LoRA with the zimage model:

1. **Weight attribute access in LoRA-wrapped layers**: The zimage model accesses `self.mlp[0].weight.dtype` in `TimestepEmbedder.forward()`, but when layers are wrapped with LoRA, the `weight` attribute is not directly accessible. This PR adds `weight` and `bias` properties to `BaseLayerWithLoRA` that proxy to the underlying `base_layer`, allowing code like `layer.weight.dtype` to work correctly.

2. **Incorrect output dtype in LoRA forward pass**: When using LoRA with autocast enabled, the output dtype should follow the weight's dtype (bf16) rather than the input dtype (fp32), matching the behavior without LoRA. The previous implementation used `out.to(x.dtype)`, which incorrectly converted the output to the input dtype (fp32) instead of preserving the autocast behavior (bf16). This PR removes the unnecessary dtype conversion, allowing `F.linear` to handle dtype promotion correctly under autocast, ensuring consistent behavior with and without LoRA.

These fixes ensure that:
- LoRA-wrapped layers maintain the same dtype behavior as unwrapped layers
- Mixed precision (autocast) works correctly with LoRA
- The zimage model can access weight attributes on LoRA-wrapped layers without errors

## Accuracy Tests
Use the lora is `elusarca-anime-style-lora-z-image-turbo` , prompt is "elusarca anime style.  A female demon sitting on a thrown chair.  Fire burns around her"
After resolving the issues mentioned above, Lora was able to function normally.

image

## Checklist

### DIFF
--- a/python/sglang/multimodal_gen/runtime/layers/lora/linear.py
+++ b/python/sglang/multimodal_gen/runtime/layers/lora/linear.py
@@ -59,6 +59,14 @@ class BaseLayerWithLoRA(nn.Module):
         self.lora_A = None
         self.lora_B = None
 
+    @property
+    def weight(self):
+        return self.base_layer.weight
+
+    @property
+    def bias(self):
+        return getattr(self.base_layer, 'bias', None)
+
     @torch.compile()
     def forward(self, x: torch.Tensor) -> torch.Tensor:
         lora_A = self.lora_A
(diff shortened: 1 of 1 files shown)


mickqian (approved):
LGTM. We should keep an eye on the unwanted autocast behavior for non-LoRA cases

niehen6174:
> LGTM. We should keep an eye on the unwanted autocast behavior for non-LoRA cases

Yes，If `autocast-disable` is used with `zimage`, there will be a `dtype` issue in `self.cap_embedder`. I will test this later, and if there are any problems, I will submit a PR to fix it.

mickqian:
@niehen6174. thanks! But the reason I mentioned it is, I saw this type conversion error with `autocast` before, for other models

PR #14550, opened 2025-12-06
Revert "Add fused FP8 KV cache write kernel for TRTLLM MHA backend"

Qiaolin-Yu:
Reverts sgl-project/sglang#14093 because it breaks b200 ci


harvenstar:
@Qiaolin-Yu Thanks for reverting. I reproduced the B200 Triton error locally with a minimal script, root cause is k_scale / v_scale being 0D tensors seen as pointer<fp32> in the FP8 KV kernel. 

A fix is already in progress. I am preparing a follow up PR with a corrected and 
type-safe version of the fused kernel that passes the LLaMA-3.1 FP4 B200 test.

PR #14553, opened 2025-12-06
Fix FP8 KV Triton type issue and add regression test

harvenstar:
Fix B200 CI failure that caused #14093 to be reverted in #14550.

Original issue discovered in CI run:
https://github.com/sgl-project/sglang/actions/runs/19985523381/job/57335123826?pr=14493

Root cause:
- layer.k_scale/v_scale are torch.nn.Parameter (0-D tensors)
- Triton interprets tensor arguments as pointers
- Expression '1.0 / k_scale' in kernel caused IncompatibleTypeError

Solution:
- Convert tensor scales to Python float in wrapper before kernel launch
- Handles None, Tensor, and scalar cases explicitly
- Preserves per-tensor scaling semantics

Testing:
- Added regression test test_fp8_kv_kernel_accepts_tensor_scales
- Verified on local environment with ATTN_BACKEND=trtllm_mha
- All 11 kernel tests pass, including new regression test
- test_llama31_fp4.py::TestLlama31FP4B200::test_gsm8k passes

## Motivation

Fix the Triton `pointer<fp32>` vs `float32` type error on B200 when using the fused FP8 KV cache path for TRTLLM MHA, which breaks Llama-3.1-FP4 CI.

## Modifications

- Normalize `layer.k_scale` / `layer.v_scale` into Python float scalars in `fused_fp8_set_kv_buffer` so Triton does not treat them as `pointer<fp32>`.
- Add a regression test `test_fp8_kv_kernel_accepts_tensor_scales` to cover the 0-D tensor scale case.

## Accuracy Tests

- `pytest test/manual/test_trtllm_fp8_kv_kernel.py -vv`
- `ATTN_BACKEND=trtllm_mha pytest test/srt/test_llama31_fp4.py::TestLlama31FP4B200::test_gsm8k -vv`

Both pass on a B200 box.

@Qiaolin-Yu @Fridge003


python/sglang/srt/layers/attention/trtllm_fp8_kv_kernel.py
```diff
@@ -343,15 +343,33 @@ def fused_fp8_set_kv_buffer(
...
+            k_scale_val = float(k_scale)
+        else:
+            k_scale_val = k_scale
+
+        if v_scale is None:
+            v_scale_val = 1.0
+        elif isinstance(v_scale, torch.Tensor):
+            v_scale_val = float(v_scale)
+        else:
+            v_scale_val = v_scale
```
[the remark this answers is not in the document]
harvenstar: The duplicated scale normalization code is now replaced by a GPU-only inverse-scale path, so that suggestion should be obsolete after the latest commits.

b8zhong:
Before on main

```
python3 -m sglang.launch_server --model-path nvidia/Llama-3.1-8B-Instruct-FP4 --trust-remote-code --quantization modelopt_fp4 --kv-cache-dtype fp8_e4m3
```

```
+-------------+--------+------------+-----------------+
| Latency (s) | Tokens | Acc Length | Speed (token/s) |
+-------------+--------+------------+-----------------+
|    2.097    |  512   |   1.000    |     244.20      |
+-------------+--------+------------+-----------------+
```

After the original PR + this fix:

```
+-------------+--------+------------+-----------------+
| Latency (s) | Tokens | Acc Length | Speed (token/s) |
+-------------+--------+------------+-----------------+
|    1.783    |  512   |   1.000    |     287.12      |
+-------------+--------+------------+-----------------+
```

```
python test/srt/test_llama31_fp4.py
...
Accuracy: 0.650
Invalid: 0.000
Latency: 2.116 s
Output throughput: 3933.404 token/s
{'accuracy': np.float64(0.65), 'invalid': np.float64(0.0), 'latency': 2.1162333069951274, 'output_throughput': 3933.4037378985295}
.
Ran 1 test in 33.798s

OK
```

So it seems fine. Thanks!

b8zhong (approved):

python/sglang/srt/layers/attention/trtllm_fp8_kv_kernel.py
b8zhong, line 1: Tiny: later, can we move this codes to `python/sglang/srt/layers/attention/triton_ops`, instead of in the root dir of attention folder

harvenstar: Thanks for the suggestion! I will move the kernel into `attention/triton_ops` in a follow-up PR to keep this one focused and avoid unnecessary back-and-forth.

PR #14592, opened 2025-12-08
Fix prefill latency performance drop of bench serving

gaopengff:
## Motivation
When use prompt as input for bench serving,  the server will get incorrect input length as its encode operation added special tokens. This may cause bad performance of prefill latency. We should truncate it.
Old:
```bash
input_ids(1024) -> client decode -> prompt -> server  encode -> input_ids(1025)
```
New:
```bash
input_ids(1024) -> truncate -> input_ids(1023) -> client decode -> prompt -> server encode -> input_ids(1024)
```


mingfeima:
@zhyncs could you please find someone to review this one? a fix for benchmark scripts~

PR #14601, opened 2025-12-08
fix the deepep 8 gpu unit test

rainj-me:
## Motivation

Fix the unit test failure for test/srt/ep/test_deepep_large.py

## Checklist


python/sglang/srt/layers/moe/utils.py
```diff
@@ -284,7 +284,7 @@ def speculative_moe_a2a_backend_context():
     global MOE_A2A_BACKEND
     original_backend = MOE_A2A_BACKEND
     try:
-        MOE_A2A_BACKEND = MoeA2ABackend.NONE
+        MOE_A2A_BACKEND = get_speculative_moe_a2a_backend()
```
Fridge003: We don't need to explicitly set `--moe-runner-backend` in `test_deepep_large.py`.
This line is enough for fixing this bug, since get_speculative_moe_a2a_backend will return MoEA2Abackend.DeepEP.

We were failing the test since speculative a2a backend was None

Fridge003 (approved):

PR #14611, opened 2025-12-08
[Loader] Add weight FQN remapper

yinghai:
## Modifications

In RL, trainer might create FQN a bit differently from HF file. We can allow this flexibility by adding a weight remapper. Make this register based and allow private mapper since it's usually RL loop specific. 

## Checklist

### DIFF
--- a/python/sglang/srt/model_loader/loader.py
+++ b/python/sglang/srt/model_loader/loader.py
@@ -79,6 +79,7 @@ DEFAULT_GPU_MEMORY_FRACTION_FOR_CALIBRATION = (
     0.8  # Reserve 20% GPU memory headroom for ModelOpt calibration
 )
 from sglang.srt.environ import envs
+from sglang.srt.model_loader.weight_remapper import WeightsMapper, get_weights_mapper
 from sglang.srt.model_loader.weight_utils import (
     download_safetensors_index_file_from_hf,
     download_weights_from_hf,
@@ -328,6 +329,13 @@ class DefaultModelLoader(BaseModelLoader):
                 f"{load_config.load_format}: "
                 f"{unexpected_keys}"
             )
+        
+        try:
+            from sglang.private.weights_mapping import register_private_weights_mapper
+            
+            register_private_weights_mapper()
+        except ImportError:
+            pass
 
     def _maybe_download_from_modelscope(
         self, model: str, revision: Optional[str]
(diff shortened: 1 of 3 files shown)


python/sglang/srt/models/qwen2_5_vl.py
```diff
@@ -688,7 +688,7 @@ def load_weights(self, weights: Iterable[Tuple[str, torch.Tensor]]):
                     if name in params_dict.keys():
                         param = params_dict[name]
                     else:
-                        continue
```
yinghai, line 691: I found it weird to silently pass a mismatched FQN. On the other hand, when I load HF weights, this won't trigger.

python/sglang/srt/model_loader/weight_remapper.py
```diff
@@ -0,0 +1,67 @@
...
+            if key.startswith(prefix):
+                if new_key is None:
+                    return None
+                key = key.replace(prefix, new_key, 1)
+
+        for suffix, new_key in self.orig_to_new_suffix.items():
+            if key.endswith(suffix):
+                if new_key is None:
+                    return None
+                key = new_key.join(key.rsplit(suffix, 1))
```
fzyzcjy, line 46: optional nit: I would personally suggest refactor a bit for this repeated pattern

python/sglang/srt/model_loader/loader.py
```diff
@@ -328,6 +329,13 @@ def __init__(self, load_config: LoadConfig):
                 f"{load_config.load_format}: "
                 f"{unexpected_keys}"
             )
+        
+        try:
+            from sglang.private.weights_mapping import register_private_weights_mapper
+            
+            register_private_weights_mapper()
+        except ImportError:
+            pass
```
fzyzcjy, line 338: I see patterns like https://github.com/sgl-project/sglang/blob/2970f22917cbfb559e73e254ea884394e556d5e0/python/sglang/srt/server_args.py#L177, https://github.com/sgl-project/sglang/blob/2970f22917cbfb559e73e254ea884394e556d5e0/python/sglang/srt/models/deepseek_v2.py#L251 for adding extra things in SGLang codebase, and is not very sure whether this directly importing is ok or not (maybe cc @merrymercy who introduced the private-folder approach)

python/sglang/srt/model_loader/loader.py
```diff
@@ -605,6 +613,11 @@ def load_model(
 
     @staticmethod
     def load_weights_and_postprocess(model, weights, target_device):
+        model_class_name = model.__class__.__name__
```
fzyzcjy, line 616: optional nit: maybe add a todo comment mentioning some places call model.load_weights directly instead of calling load_weights_and_postprocess, and we should let every place call load_weights_and_postprocess later to make it unified

fzyzcjy (approved):
LGTM w/ optional nits

python/sglang/srt/models/qwen2_5_vl.py
fzyzcjy: totally agree

python/sglang/srt/model_loader/loader.py
yinghai: I'll take a look at add_load_format_choices

PR #14626, opened 2025-12-08
[PD] Support KV transfer with MORI-IO

maning00:
## Motivation

[MORI-IO](https://github.com/ROCm/mori) is AMD's high-performance, point-to-point communication library that leverages GDR (GPU Direct RDMA) to achieve ultra-low latency and high bandwidth for KVCache transfer in LLM inference. To enable efficient PD (Prefill-Decode) disaggregation on AMD hardware, we adopt MORI-IO transfer engine as the transport layer for SGLang.

### Architecture Overview

The implementation follows a similar pattern to the mooncake transfer engine integration, with MORI-IO-specific optimizations:

#### 1. **MoriKVManager** - Core Transfer Management

- **Initialization**:
  - Creates `IOEngine` with RDMA backend configuration
  - Registers all GPU memory buffers with the engine
  - Spawns dedicated threads for bootstrap or status polling
  
- **Configuration Options** (via environment variables):
  - `SGLANG_MORI_QP_PER_TRANSFER`: Number of queue pairs per transfer (default: 1)
  - `SGLANG_MORI_POST_BATCH_SIZE`: RDMA post batch size (default: -1)
  - `SGLANG_MORI_NUM_WORKERS`: Number of worker threads (default: 1)

#### 2. **MoriKVSender** (Prefill Side)

- **Workflow**:
  1. Waits for decode instance registration via bootstrap thread
  2. Receives transfer metadata (destination memory descriptors, indices) from decode
  3. Issues RDMA writes using `batch_write` API for KV cache transfer
  4. Sends auxiliary data via TCP (ZMQ)
  5. Monitors transfer status and notifies decode instance upon completion

#### 3. **MoriKVReceiver** (Decode Side)

- **Workflow**:
  1. Registers local engine descriptor and memory descriptors with prefill instance
  2. Polls for transfer completion status
  3. Receives auxiliary data via dedicated TCP handler
  4. Updates request status based on prefill notifications

#### 4. **Dockerfile**
Add the `NIC_BACKEND` option to enable mori support for different network interface cards (NICs).

### Usage

**Installation**: Install MORI-IO library following the [MORI installation guide](https://github.com/ROCm/mori):

```bash
cd mori 
pip install -r requirements-build.txt 
git submodule update --init --recursive 
pip3 install .
```

**SGLang PD Disaggregation with MORI-IO**: Use `--disaggregation-transfer-backend mori` to enable MORI-IO transfer engine:

### Known Limitations
State data transfer not implemented: Currently, MORI-IO implementation does not support state data transfer for hybrid models (Mamba, SWA, NSA).

### End-to-End PD Disaggregation
**Hardware Configuration:**
- **GPUs**: 8x AMD Instinct MI355X per node
- **CPUs**: 2x AMD EPYC per node  
- **Network**: 8x AMD Pensando Pollara 400 AI-NIC per node
- **Model**: DeepSeek-V3 with TP=8
- **Setup**: 3-node configuration (1 prefill + 1 decode + 1 router)

**Benchmark Command:**

Prefill instance (node 1):
```bash
  --model-path DeepSeek-V3 \
  --host 0.0.0.0 --port 30002 \
  --tp-size 8 \
  --disaggregation-transfer-backend mori \
  --disable-radix-cache \
  --trust-remote-code
```

Decode instance (node 2):
```bash
  --model-path DeepSeek-V3 \
  --host 0.0.0.0 --port 30003 \
  --tp-size 8 \
  --disaggregation-transfer-backend mori \
  --disable-radix-cache \
  --trust-remote-code
```

Router (node 3):
```bash
python -m sglang_router.launch_router \
  --pd-disaggregation --mini-lb \
  --prefill http://node1:30002 \
  --decode http://node2:30003 \
  --host 0.0.0.0 --port 8000
```

Benchmark client:
```bash
  --backend sglang \
  --base-url http://127.0.0.1:8000 \
  --dataset-name random \
  --num-prompts 1 \
  --random-input <1024/2048/4096/8192> \
  --random-output 16
```

Accuracy test:
```bash
python3 -m sglang.test.few_shot_gsm8k \
    --host http://127.0.0.1 \
    --port 8000 \
    --num-questions 200 \
    --parallel 128 \
    --num-shots 5
```

**Performance Results:**

Comparison of STANDALONE (no PD disaggregation) vs MORI vs MOONCAKE backends. Each test was run 3 times and averaged.

| Input Tokens | Metric              | STANDALONE | MORI   | MOONCAKE |
|--------------|---------------------|------------|--------|----------|
| **1024**     | Throughput (tok/s)  | **21.91**  | 21.22  | 21.24    |
|              | TTFT Mean (ms)      | **155.35** | 168.13 | 165.71   |
| **2048**     | Throughput (tok/s)  | **21.49**  | 20.63  | 20.78    |
|              | TTFT Mean (ms)      | **162.24** | 176.57 | 173.54   |
| **4096**     | Throughput (tok/s)  | **21.51**  | 20.68  | 20.74    |
|              | TTFT Mean (ms)      | **159.93** | 176.05 | 174.75   |
| **8192**     | Throughput (tok/s)  | **16.93**  | 16.02  | 16.10    |
|              | TTFT Mean (ms)      | **294.41** | 326.19 | 322.70   |

FP8 KV Cache Results(`--kv-cache-dtype fp8_e4m3`):

| Input Tokens | Metric              | STANDALONE | MORI   | MOONCAKE |
|--------------|---------------------|------------|--------|----------|
| **1024**     | Throughput (tok/s)  | **56.47**  | 51.79  | 51.37    |
|              | TTFT Mean (ms)      | **101.15** | 103.98 | 104.54   |
| **2048**     | Throughput (tok/s)  | **54.43**  | 49.47  | 49.69    |
|              | TTFT Mean (ms)      | **104.72** | 109.89 | 108.40   |
| **4096**     | Throughput (tok/s)  | **54.47**  | 49.72  | 49.16    |
|              | TTFT Mean (ms)      | **103.23** | 108.74 | 109.21   |
| **8192**     | Throughput (tok/s)  | **36.16**  | 31.63  | 32.11    |
|              | TTFT Mean (ms)      | **208.80** | 229.15 | 223.93   |

**gsm8k Accuracy Test Results:**
| Metric              | STANDALONE | MORI   | MOONCAKE |
|---------------------|------------|--------|----------|
| Accuracy |  0.965 | 0.965 | 0.970 |
| Invalid | 0.000 | 0.000 | 0.000 |

Both MORI and MOONCAKE leverage RDMA effectively, with near-identical performance profiles, validating MORI-IO as a production-ready alternative for AMD hardware.

cc @inkcherry @TianDi101 @Duyi-Wang


python/sglang/srt/disaggregation/mori/conn.py
```diff
@@ -0,0 +1,1120 @@
...
+            self.prefill_response_tracker: Dict[int, Set[int]] = defaultdict(set)
+            self.addr_to_rooms_tracker = defaultdict(set)
+            self.room_to_bootstrap_addr: Dict[int, str] = {}
+            self._start_decode_thread()
+
+    def _init_engine(self) -> IOEngine:
+        if self.kv_args.ib_device:
+            os.environ["MORI_RDMA_DEVICES"] = self.kv_args.ib_device
+
+        port = get_free_port()
```
ShangmingCai: This util is not robust, which might cause port conflict in some situations. Is it possible to borrow some idea from `get_zmq_socket_on_host`?

ShangmingCai (commented):
This PR seems very complete. Would it be convenient to add a test to AMD's CI to verify the correctness?

python/sglang/srt/disaggregation/mori/conn.py
maning00: Thank you! Fixed by:
1. Using `port=0` to let the OS atomically bind an available port
2. Retrieve the actual bound port from Mori's TCP stack

This eliminates the race condition entirely. Thanks for catching this!

maning00:
> This PR seems very complete. Would it be convenient to add a test to AMD's CI to verify the correctness?

@ShangmingCai I added a `per-commit-amd` correctness test for the MORI integration (protocol/serialization/message-handling) using a lightweight fake mori module, since the current AMD CI image (rocm/sgl-dev) doesn’t include mori.
Would you prefer we (1) keep this as-is, or (2) additionally add a real MORI E2E test after mori is available in the AMD CI image?

ShangmingCai:
> > This PR seems very complete. Would it be convenient to add a test to AMD's CI to verify the correctness?
> 
> @ShangmingCai I added a `per-commit-amd` correctness test for the MORI integration (protocol/serialization/message-handling) using a lightweight fake mori module, since the current AMD CI image (rocm/sgl-dev) doesn’t include mori. Would you prefer we (1) keep this as-is, or (2) additionally add a real MORI E2E test after mori is available in the AMD CI image?

@maning00 I prefer 2, we can add the E2E test later. Do you mind moving this test to the manual dir? Most of the disaggregation tests are in the `test/srt` dir, adding a new disaggregation dir would bring some confusion. And also, this simulation test can not guarantee the accuracy of the mori backend (no need to waste the CI for it), so we should put it in the manual dir, and add a real E2E test when it is ready.

HaiShaw (changes requested):
Please incorporate mori.Dockerfile into existing rocm.Dockerfile.

maning00:
> @maning00 have a perf table as above with fp8 kv cache?

@HaiShaw  Not yet — I haven’t run the FP8 KV cache perf sweep. I can benchmark it and add a perf table in a follow-up update to this PR (same format as the existing table), once the runs finish.

HaiShaw:
> > Please incorporate mori.Dockerfile into existing rocm.Dockerfile.
> 
> Sure. `mori.Dockerfile` mainly adds AMD AINIC-related dependencies plus MORI build/config flags. I can incorporate it into the existing `rocm.Dockerfile` and make it an optional feature (e.g., gated by a build arg), so the default image stays unchanged.

Can think to add NIC options to support AMD AINIC, Broadcom Thor2, etc.

maning00:
> > @maning00 have a perf table as above with fp8 kv cache?
> 
> @HaiShaw Not yet — I haven’t run the FP8 KV cache perf sweep. I can benchmark it and add a perf table in a follow-up update to this PR (same format as the existing table), once the runs finish.

@HaiShaw Added FP8 KV cache perf results

docker/rocm.Dockerfile
```diff
@@ -75,6 +75,16 @@ ARG TILELANG_COMMIT="ae938cf885743f165a19656d1122ad42bb0e30b8"
 ARG FHT_REPO="https://github.com/jeffdaily/fast-hadamard-transform.git"
 ARG FHT_BRANCH="rocm"
 ARG FHT_COMMIT="46efb7d776d38638fc39f3c803eaee3dd7016bd1"
+
+ARG ENABLE_MORI=0
+ARG NIC_BACKEND=none
+
+ARG MORI_REPO="https://github.com/ROCm/mori.git"
+ARG MORI_COMMIT="main"
```
HaiShaw: Can we move this onto a stable commit/tag for versioning?

docker/rocm.Dockerfile
```diff
@@ -293,6 +303,70 @@ RUN python3 -m pip install --no-cache-dir \
...
+          libionic-dev \
+          ionic-common \
+      ; \
+      rm -rf /var/lib/apt/lists/*; \
+      ;; \
+    bnxt) \
+      export USE_IONIC="OFF"; \
+      export USE_BNXT="ON"; \
+      echo "[MORI] NIC_BACKEND=bnxt: USE_BNXT=ON. Add Broadcom bnxt packages/repos here when available."; \
+      ;; \
```
HaiShaw: When we can have bnxt support added here?
cc @Lzy17

python/sglang/srt/disaggregation/utils.py
```diff
@@ -233,6 +233,7 @@ class TransferBackend(Enum):
     MOONCAKE = "mooncake"
     NIXL = "nixl"
     ASCEND = "ascend"
+    MORI = "mori"
```
HaiShaw: move to under MOONCAKE

docker/rocm.Dockerfile
```diff
@@ -293,6 +303,70 @@ RUN python3 -m pip install --no-cache-dir \
...
+  esac; \
+  \
+  # Build/install MORI
+  export MORI_GPU_ARCHS="${GPU_ARCH_LIST}"; \
+  echo "[MORI] MORI_GPU_ARCHS=${MORI_GPU_ARCHS} USE_IONIC=${USE_IONIC} USE_BNXT=${USE_BNXT}"; \
+  rm -rf /sgl-workspace/mori; \
+  git clone "${MORI_REPO}" /sgl-workspace/mori; \
+  cd /sgl-workspace/mori; \
+  git checkout "${MORI_COMMIT}"; \
+  git submodule update --init --recursive; \
```
HaiShaw, line 356: Need `requirements.txt` check?

HaiShaw (changes requested):
Is `--page_size > 1` supported?
Please also add description for tuning following variables:
```
SGLANG_MORI_QP_PER_TRANSFER: Number of queue pairs per transfer (default: 1)
SGLANG_MORI_POST_BATCH_SIZE: RDMA post batch size (default: -1)
SGLANG_MORI_NUM_WORKERS: Number of worker threads (default: 1)
```

docker/rocm.Dockerfile
maning00: Update: To ensure full functionality for mori (io and ep), BRCM support will be integrated later once IBGDA support is fully available in the official library.

docker/rocm.Dockerfile
maning00: No need to check here; the only dependency is torch.

maning00:
> Is `--page_size > 1` supported? Please also add description for tuning following variables:
> 
> ```
> SGLANG_MORI_QP_PER_TRANSFER: Number of queue pairs per transfer (default: 1)
> SGLANG_MORI_POST_BATCH_SIZE: RDMA post batch size (default: -1)
> SGLANG_MORI_NUM_WORKERS: Number of worker threads (default: 1)
> ```

Sure, it is supported. 
Added description for these variables

maning00:
> @maning00 Please add a basic accuracy test (gsm8k, etc.) on DPSK from 1P1D. cc @kkHuang-amd @Lzy17

Added gsm8k test results

docker/rocm.Dockerfile
HaiShaw: Do you mean mori yet to support BRCM rdma-core?

docker/rocm.Dockerfile
HaiShaw: @kkHuang-amd let's keep an eye on this onwards

docker/rocm.Dockerfile
maning00: No, that is not the case. This is mainly because `mori-ep` and `mori-io` are built simultaneously. `mori-ep` currently depends on a pre-release version of the BRCM library (for IBGDA).

docker/rocm.Dockerfile
```diff
@@ -293,6 +303,65 @@ RUN python3 -m pip install --no-cache-dir \
...
+      echo "deb [arch=amd64 signed-by=/etc/apt/keyrings/amdainic.gpg] https://repo.radeon.com/amdainic/pensando/ubuntu/${AINIC_VERSION} ${UBUNTU_CODENAME} main" \
+        > /etc/apt/sources.list.d/amdainic.list; \
+      apt-get update && apt-get install -y --no-install-recommends \
+          libionic-dev \
+          ionic-common \
+      ; \
+      rm -rf /var/lib/apt/lists/*; \
+      ;; \
+    *) \
+      echo "ERROR: unknown NIC_BACKEND=${NIC_BACKEND}. Use one of: none, ainic, bnxt"; \
```
HaiShaw: where `bnxt` is handled, or should we change this echo message?

HaiShaw (changes requested):
Should add BXNT, etc. support later - w.r.t. `NIC_BACKEND`

HaiShaw (approved):

PR #14627, opened 2025-12-08
[Bugfix] Fix KeyError for Mistral-Large-3 rope_scaling config

alisonshao:
## Summary
- Fix `KeyError: 'factor'` when loading Mistral-Large-3 model
- Use `.get()` to safely access `rope_scaling["factor"]` which may not exist in Mistral-Large-3's config structure

## Root Cause
Mistral-Large-3's HuggingFace config uses a different `rope_scaling` structure (based on `rope_parameters`) that doesn't have a top-level `"factor"` key. The code at `model_config.py:361` assumed all MLA models with `rope_scaling` have this key.

## Error Reference
https://github.com/sgl-project/sglang/actions/runs/20017692847/job/57401685911?pr=14571

## Test plan
- [x] Verify Mistral-Large-3 CI test passes

### DIFF
--- a/python/sglang/srt/configs/model_config.py
+++ b/python/sglang/srt/configs/model_config.py
@@ -358,9 +358,10 @@ class ModelConfig:
                 mscale_all_dim = self.hf_config.rope_scaling.get(
                     "mscale_all_dim", False
                 )
-                scaling_factor = self.hf_config.rope_scaling["factor"]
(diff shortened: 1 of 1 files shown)


Fridge003 (approved):

ispobock:
We need to check why the factor is not loaded https://huggingface.co/mistralai/Mistral-Large-3-675B-Instruct-2512/blob/main/params.json#L114

PR #14636, opened 2025-12-08
[VLM] feat: true on policy for vlm + fsdp

nanjiangwill:
## Motivation

Support true on policy for VLM + FSDP

Exp: Qwen3-VL

Text-only: [#12058](https://github.com/sgl-project/sglang/pull/12058) 

## Modifications

when `get_global_server_args().rl_on_policy_target is not None`
1. change `MRotaryEmbedding.forward()` to be `forward_native()` 
2. PIL -> `pixel_values` conversion under `cpu` not `cuda`
3. `position_embeddings` to be same type as `inputs` in ViT
6. change `fast_pos_embed_interpolate` from `numpy` to `torch` implementation
7. set env `SGLANG_VLM_CACHE_SIZE_MB=0` 
8. add `input_deepstack_embeds` to `hidden_states` after `hidden_states + residual` since `hidden_states + residual + input_deepstack_embeds != hidden_states + input_deepstack_embeds + residual`

## Checklist


python/sglang/srt/layers/rotary_embedding.py
```diff
@@ -1510,14 +1511,12 @@ def forward(
...
-        elif _is_npu:
-            return self._forward_npu(positions, query, key)
-        else:
-            return self._forward_native(positions, query, key)
+        # Use Triton kernel for multimodal (2D positions) with mrope
+        if positions.ndim == 2 and self.mrope_section:
+            return self.forward_triton(positions, query, key)
+        return self.forward_native(positions, query, key, fused_set_kv_buffer_arg)
 
-    def _forward_triton(
```
zhaochenyang20, line 1553: no need to change it from `_forward` to `forward`?

python/sglang/srt/layers/rotary_embedding.py
```diff
@@ -1566,15 +1565,19 @@ def _forward_triton(
         key = torch.cat((key_rot, key_pass), dim=-1).reshape(key_shape)
         return query, key
 
-    def _forward_npu(
```
zhaochenyang20, line 1574: no need to change it from _forward to forward?

python/sglang/srt/layers/rotary_embedding.py
nanjiangwill: this inherits RotaryEmbedding and has CustomOp, we should just keep on standard. so we just follow the format in class RotaryEmbedding

python/sglang/srt/layers/rotary_embedding.py
nanjiangwill: same rationale


this inherits RotaryEmbedding and has CustomOp, we should just keep on standard. so we just follow the format in class RotaryEmbedding

python/sglang/srt/layers/layernorm.py
```diff
@@ -121,6 +121,10 @@ def forward_cuda(
                 self.variance_epsilon,
             )
         if residual is not None:
+            # TODO: Ideally we want to have (a+b)+c. but right now we can only have a+(b+c).
+            # (a+b)+c != a+(b+c), we probably need to add another parameter to fused_add_rmsnorm
```
zhaochenyang20: what is a b c here stands for?

zhaochenyang20 (changes requested):

python/sglang/srt/layers/layernorm.py
nanjiangwill: updated!

zhaochenyang20 (approved):
LGTM.

mickqian:
could you attach the mmmu accuracy change in here? since this PR modifies core implementation of qwen3-vl

mickqian (approved):

python/sglang/srt/layers/layernorm.py
```diff
@@ -179,13 +190,23 @@ def forward_native(
...
-            x = x + residual.to(torch.float32)
+            x = (
+                x
+                + residual.to(torch.float32)
+                + (
+                    post_residual_addition.to(torch.float32)
+                    if post_residual_addition is not None
+                    else 0.0
+                )
+            )
```
JustinTong0323, line 210: Maybe it is cleaner to write in this way:
```suggestion
        post_residual_addition = kwargs.get("post_residual_addition", 0.0).to(torch.float32)
        x = x.to(torch.float32)
        if residual is not None:
            x = x + residual.to(torch.float32)
            x = (
                x
                + residual.to(torch.float32)
                + post_residual_addition
            )
```

nanjiangwill: we need to make to(float) inside the if. but i will make the default value 0.0

python/sglang/srt/multimodal/processors/base_processor.py
```diff
@@ -318,7 +319,9 @@ def process_mm_data(
             and isinstance(processor.image_processor, BaseImageProcessorFast)
             and not self.server_args.disable_fast_image_processor
         ):
-            if not _is_npu:
+            if get_global_server_args().rl_on_policy_target is not None:
+                kwargs["device"] = "cpu"
```
JustinTong0323, line 321: Could you please explain more about this? Why using fast processor would break onpolicy🤔

JustinTong0323:
Also please resolve the conflicts~

python/sglang/srt/multimodal/processors/base_processor.py
nanjiangwill: because of the image processing, it will generate `pixel_values` in float, and doing that on GPU will give mismatch between Sglang and HF

yxchng:
is this going to be merged in new version soon?

python/sglang/srt/server_args.py
```diff
@@ -2329,6 +2329,9 @@ def _handle_deterministic_inference(self):
                 "Enable deterministic inference because of rl_on_policy_target."
             )
             self.enable_deterministic_inference = True
+
+            # For VLM
+            os.environ["SGLANG_VLM_CACHE_SIZE_MB"] = "0"
```
JustinTong0323, line 2329: why default it to 0?

nanjiangwill: this will only be triggered when using true on policy mode. we need to evict image cache for now as it will introduce mismatch. will optimize this in the future.

JustinTong0323: Thanks for the explanation, that makes sense.

JustinTong0323 (approved):

python/sglang/srt/layers/layernorm.py
```diff
@@ -105,21 +105,30 @@ def forward_cuda(
...
                 self.variance_epsilon,
             )
         if residual is not None:
+            # TODO: Ideally we want to have (hidden_states+residual)+post_residual_addition.
+            # but right now we can only have hidden_states+(residual+post_residual_addition).
+            # (hidden_states+residual)+post_residual_addition != hidden_states+(residual+post_residual_addition),
+            # we probably need to add another parameter to fused_add_rmsnorm
+            post_residual_addition = kwargs.get("post_residual_addition")
+            residual = residual + (
+                post_residual_addition if post_residual_addition is not None else 0.0
```
merrymercy, line 130: This adds unnecessary overhead. Please be super careful next time. https://github.com/sgl-project/sglang/pull/16474/changes

Also, please fix all other places where you did the same mistake

python/sglang/srt/layers/layernorm.py
```diff
@@ -105,21 +105,30 @@ def forward_cuda(
         self,
         x: torch.Tensor,
         residual: Optional[torch.Tensor] = None,
+        **kwargs,
```
merrymercy, line 108: `kwargs` is a very bad pattern. Use it unless you do not have any other methods to deal with it.
Why is this needed?

python/sglang/srt/layers/layernorm.py
```diff
@@ -179,13 +191,23 @@ def forward_native(
...
         x = x.to(torch.float32)
         if residual is not None:
-            x = x + residual.to(torch.float32)
+            x = (
+                x
+                + residual.to(torch.float32)
+                + (
+                    post_residual_addition.to(torch.float32)
+                    if post_residual_addition is not None
+                    else 0.0
```
merrymercy, line 208: Do not add unnecessary 0.0

PR #14668, opened 2025-12-08
[NVIDIA] Add flashinfer all-to-all MOE dispatcher

trevor-m:
~Draft PR since https://github.com/flashinfer-ai/flashinfer/pull/2102 is not yet merged into flashinfer.~

https://github.com/flashinfer-ai/flashinfer/pull/2102 is now merged.

## Motivation

This PR integrates the latest TRT-LLM moe all-to-all kernels into sglang (AKA nvlink one sided allotall or mnnvlthroughput alltoall):

> NVLINK one-sided comm AllToAll strategy for throughput scenarios.
> 
> This implementation utilizes symmetric memory to enable peer-to-peer access between GPUs over NVLink.
> The kernels only take the role as one side of the communication: the dispatch kernel puts the data
> into peer ranks' symmetric memory from local buffer, while the combine kernel gets the data from peer
> ranks' symmetric memory and reduces the data into local buffer. It is the most efficient implementation
> by now, but requires symmetric memory size proportional to `max_num_tokens * n_ranks`, which may not
> scale well for very large-scale parallelization.

Currently I have tested it with the flashinfer_cutlass moe runner backend and use fp4 quantize before communication. It also allows flashinfer_cutlass moe to write directly to the workspace buffer.

Remaining issues: 
1. Need to improve max_num_tokens - should multiply the per_rank value by ep_size. Or calculate this automatically based on server args.
2. Currently flashinfer a2a doesn't support when ranks have 0 tokens, so we need to create a dummy token in those cases. Doing the tensor allocation prevents us from using streams to overlap with shared experts.

## Modifications

* Add `--moe-a2a-backend=flashinfer`

## Accuracy Tests

```
SGLANG_MOE_NVFP4_DISPATCH=1 python3 -m sglang.launch_server --model-path nvidia/DeepSeek-R1-0528-FP4-v2 --trust-remote-code --quantization modelopt_fp4 --tp 4 --moe-runner-backend flashinfer_cutlass --ep-size 4 --dp 4 --enable-dp-attention --mem-fraction-static 0.85 --max-running-requests 2048 --stream-interval 5 --enable-dp-lm-head --attention-backend triton --cuda-graph-bs 1 2 4 8 16 32 64 128 256 512 --disable-radix-cache --moe-a2a-backend flashinfer
python3 -m sglang.bench_serving --backend sglang --dataset-name random --num-prompt 2048 --random-input 1024 --random-output 1024 --random-range-ratio 1 --max-concurrency 2048
Accuracy: 0.955
Invalid: 0.000
Latency: 166.179 s
Output throughput: 874.407 token/s
```

## Benchmarking and Profiling

Single node results are below. Working on multinode benchmarking

### Single node 4xGB200 results and profiles
```
SGLANG_MOE_NVFP4_DISPATCH=1 python3 -m sglang.launch_server --model-path nvidia/DeepSeek-R1-0528-FP4-v2 --trust-remote-code --quantization modelopt_fp4 --tp 4 --moe-runner-backend flashinfer_cutlass --ep-size 4 --dp 4 --enable-dp-attention --mem-fraction-static 0.85 --max-running-requests 2048 --stream-interval 5 --enable-dp-lm-head --cuda-graph-bs 1 2 4 8 16 32 64 128 256 512 --disable-radix-cache --moe-a2a-backend flashinfer --disable-shared-experts-fusion
python3 -m sglang.bench_serving --backend sglang --dataset-name random --num-prompt 2048 --random-input 1024 --random-output 1024 --random-range-ratio 1 --max-concurrency 2048
Max request concurrency:                 2048      
Successful requests:                     2048      
Benchmark duration (s):                  105.78    
Total input tokens:                      2097152   
Total input text tokens:                 2097152   
Total generated tokens (retokenized):    2093741   
Request throughput (req/s):              19.36     
Input token throughput (tok/s):          19826.37  
Output token throughput (tok/s):         19826.37  
Peak output token throughput (tok/s):    41273.00  
Peak concurrent requests:                2048      
Total token throughput (tok/s):          39652.75  
Concurrency:                             2033.67   
Mean E2E Latency (ms):                   105036.00 
Median E2E Latency (ms):                 105036.65 
Mean TTFT (ms):                          19192.10  
Median TTFT (ms):                        18945.35  
P99 TTFT (ms):                           35753.33  
Mean TPOT (ms):                          83.91     
Median TPOT (ms):                        84.31     
P99 TPOT (ms):                           96.69     
Mean ITL (ms):                           83.91     
Median ITL (ms):                         64.86     
P95 ITL (ms):                            134.52    
P99 ITL (ms):                            187.70    
Max ITL (ms):                            6041.95   

With SGLANG_MOE_NVFP4_DISPATCH=0 (BF16 dispatch)
Max request concurrency:                 2048      
Successful requests:                     2048      
Benchmark duration (s):                  113.61    
Total input tokens:                      2097152   
Total input text tokens:                 2097152   
Total generated tokens (retokenized):    2094097   
Request throughput (req/s):              18.03     
Input token throughput (tok/s):          18459.37  
Output token throughput (tok/s):         18459.37  
Peak output token throughput (tok/s):    39440.00  
Peak concurrent requests:                2048      
Total token throughput (tok/s):          36918.74  
Concurrency:                             2032.17   
Mean E2E Latency (ms):                   112730.76 
Median E2E Latency (ms):                 112745.83 
Mean TTFT (ms):                          22471.12  
Median TTFT (ms):                        22192.53  
P99 TTFT (ms):                           42519.44  
Mean TPOT (ms):                          88.23     
Median TPOT (ms):                        88.62     
P99 TPOT (ms):                           104.26    
Mean ITL (ms):                           88.23     
Median ITL (ms):                         66.35     
P95 ITL (ms):                            139.80    
P99 ITL (ms):                            192.80    
Max ITL (ms):                            7388.84   

With --moe-a2a-backend=none
Max request concurrency:                 2048      
Successful requests:                     2048      
Benchmark duration (s):                  107.98    
Total input tokens:                      2097152   
Total input text tokens:                 2097152   
Total generated tokens (retokenized):    2093493   
Request throughput (req/s):              18.97     
Input token throughput (tok/s):          19420.79  
Output token throughput (tok/s):         19420.79  
Peak output token throughput (tok/s):    43370.00  
Peak concurrent requests:                2048      
Total token throughput (tok/s):          38841.58  
Concurrency:                             2036.25   
Mean E2E Latency (ms):                   107365.61 
Median E2E Latency (ms):                 107376.61 
Mean TTFT (ms):                          19164.99  
Median TTFT (ms):                        18753.23  
P99 TTFT (ms):                           35803.22  
Mean TPOT (ms):                          86.22     
Median TPOT (ms):                        86.69     
P99 TPOT (ms):                           98.54     
Mean ITL (ms):                           86.22     
Median ITL (ms):                         66.88     
P95 ITL (ms):                            149.64    
P99 ITL (ms):                            224.95    
Max ITL (ms):                            5954.79   
```

#### Dispatch (bs=512) with --moe-a2a-backend=flashinfer
Screenshot 2025-12-10 at 2 28 05 PM

#### Dispatch (bs=512) with --moe-a2a-backend=none (FP4 allgather)
Screenshot 2025-12-08 at 11 48 14 AM

#### Combine (bs=512) with --moe-a2a-backend=flashinfer
Screenshot 2025-12-08 at 11 50 12 AM

#### Combine (bs-512) --moe-a2a-backend=none (reduce-scatter)
Screenshot 2025-12-08 at 11 50 15 AM
 (bs=512) with --moe-a2a-backend=none (reduce-scatter)

## Checklist


python/sglang/srt/layers/moe/token_dispatcher/flashinfer.py
```diff
@@ -0,0 +1,215 @@
...
+        self.group = group
+        self.router_topk = router_topk
+        self.hidden_size = hidden_size
+        self.num_experts = num_experts
+
+        # TODO: Can this be a server arg and shared with deepep/mooncakeep?
+        self.max_num_tokens = get_int_env_var(
+            "SGLANG_FLASHINFER_NUM_MAX_DISPATCH_TOKENS_PER_RANK", 4096
+        )
+        self.workspace_mb = get_int_env_var("TRTLLM_MOE_A2A_WORKSPACE_MB", 2048)
```
djns99: We should calculate the workspace size using the helper function instead of doing this now

python/sglang/srt/layers/moe/token_dispatcher/flashinfer.py
```diff
@@ -0,0 +1,215 @@
...
+        output_dtype = hidden_states.dtype
+        topk_ids = topk_output.topk_ids
+        topk_weights = topk_output.topk_weights
+
+        # Handle case where there are no tokens on this DP worker
+        # moe_a2a.dispatch requires at least one token
+        self.has_dummy_token = False
+        if hidden_states.shape[0] == 0:
+            self.has_dummy_token = True
+            hidden_states = torch.zeros(
```
djns99: Should we use empty here instead of zeros to save the memset?

python/sglang/srt/layers/moe/token_dispatcher/flashinfer.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+class FlashinferDispatchOutput(NamedTuple):
+    """Flashinfer EP dispatch output."""
+
+    hidden_states: torch.Tensor
+    hidden_states_scale: Optional[torch.Tensor]
+    topk_output: TopKOutput
+    # Provide an output tensor to fused_moe so it writes directly to our buffer
+    moe_output: Optional[torch.Tensor] = None
```
djns99, line 48: Will it always be an option to write the result here? Or should we support opting out of this path

djns99 (approved):

python/sglang/srt/layers/moe/token_dispatcher/flashinfer.py
trevor-m: The dummy token still has to go through fp4_quantize and moe so I want the values initialized to something valid that won't affect the accuracy.

python/sglang/srt/layers/moe/token_dispatcher/flashinfer.py
trevor-m: Thanks, for now I will only enable this for flashinfer_cutlass moe path.

python/sglang/srt/layers/moe/token_dispatcher/flashinfer.py
trevor-m: Thanks, done. FYI I had to pad to 512mb alignment otherwise i hit this error. Maybe we should add an alignment check somewhere.
```
 File "/trevor/flashinfer/flashinfer/comm/trtllm_moe_alltoall.py", line 290, in moe_a2a_dispatch                                                                                                           
    moe_a2a_wrap_payload_tensor_in_workspace(                                                                                                                                                               
  File "/trevor/flashinfer/flashinfer/comm/trtllm_moe_alltoall.py", line 231, in moe_a2a_wrap_payload_tensor_in_workspace                                                                                   
    workspace_base = workspace.view(-1).view(dtype=torch.uint8)                                                                                                                                             
                     ^^^^^^^^^^^^^^^^^^                                                                                                                                                                     
RuntimeError: view size is not compatible with input tensor's size and stride (at least one dimension spans across two contiguous subspaces). Use .reshape(...) instead.
```

python/sglang/srt/layers/moe/token_dispatcher/flashinfer.py
djns99: Ah does sglang do dynamic quantization?

python/sglang/srt/layers/moe/token_dispatcher/flashinfer.py
djns99: Hmm good catch, that seems like a bug in the implementation. We shouldn't be flattening it if we have non-contiguous MNNVL blocks. I believe each slice in that function should come from a contiguous subrange, not sure why the FI tests didnt catch this

djns99:
```
    workspace_base = workspace.view(dtype=torch.uint8)
    assert slice_end - slice_start <= workspace_base.shape[1], "slice_end - slice_start must belong to a single rank"
    assert slice_start // workspace_base.shape[1] == slice_end // workspace_base.shape[1], "slice_start and slice_end must belong to the same rank"
    slice_rank = slice_start // workspace_base.shape[1]
    local_slice_start = slice_start % workspace_base.shape[1]
    slice_length = slice_end - slice_start
    local_slice_end = local_slice_start + slice_length
    result = (
        workspace_base[slice_rank, local_slice_start:local_slice_end].view(dtype=dtype).view(*leading_shape, -1)
    )
    return result
```
Can you test making this change to `moe_a2a_wrap_payload_tensor_in_workspace` in flashinfer and seeing if it works for you. I haven't tested this at all, but I think that should address the issue

trevor-m: I tried your change and it appears to have the same issue:
```
  File "/trevor/flashinfer/flashinfer/comm/trtllm_moe_alltoall.py", line 232, in moe_a2a_wrap_payload_tensor_in_workspace
    assert slice_end - slice_start <= workspace_base.shape[1], "slice_end - slice_start must belong to a single rank"
```

Based on the log, it looks like the internal Mnnvl buffer has padded the size, but perhaps the tensor view has the unpadded size.
```
[MnnvlMemory] creating address with stride=536870912
```

djns99: Hopefully the fix pushed to https://github.com/flashinfer-ai/flashinfer/pull/2102 should resolve the issues

trevor-m: Issue is now fixed

python/sglang/srt/layers/moe/token_dispatcher/flashinfer.py
```diff
@@ -0,0 +1,265 @@
...
+
+    class TorchDistributedCommBackend(CommBackend):
+        """
+        Use torch distributed instead of MPI to set up flashinfer MNNVL workspaces during initialization
+        """
+
+        def __init__(self, group: dist.ProcessGroup):
+            self._group = group
+
+        def Get_rank(self) -> int:
```
ch-wan:
```suggestion
        def get_rank(self) -> int:
```

python/sglang/srt/layers/moe/token_dispatcher/flashinfer.py
```diff
@@ -0,0 +1,265 @@
...
+    def format(self) -> DispatchOutputFormat:
+        return DispatchOutputFormat.STANDARD
+
+
+assert isinstance(FlashinferDispatchOutput, DispatchOutput)
+
+if is_flashinfer_available():
+    from flashinfer.comm.mnnvl import CommBackend
+
+    class TorchDistributedCommBackend(CommBackend):
```
ch-wan: Can we move it to a new file `flashinfer_utils.py`?

python/sglang/srt/layers/moe/token_dispatcher/flashinfer.py
```diff
@@ -0,0 +1,265 @@
...
+        group: torch.distributed.ProcessGroup,
+        router_topk: int,
+        num_experts: int = None,
+        num_local_experts: int = None,  # Unused
+        hidden_size: int = None,
+        params_dtype: torch.dtype = None,  # Unused
+    ):
+        super().__init__()
+        try:
+            from flashinfer.comm import MoeAlltoAll, moe_a2a_get_workspace_size_per_rank
```
ch-wan: We can move all import to the beginning of this file

python/sglang/srt/layers/moe/token_dispatcher/flashinfer.py
```diff
@@ -0,0 +1,265 @@
...
+            gpus_per_node=torch.cuda.device_count(),
+            pp_size=1,
+            cp_size=1,
+        )
+        self.moe_a2a = MoeAlltoAll(
+            mapping=self.mapping,
+            max_num_tokens=self.max_num_tokens,
+            top_k=self.router_topk,
+            num_experts=self.num_experts,
+            workspace_size_per_rank=self.workspace_size,
```
ch-wan, line 145: What's the size of one workspace? Each layer may have an independent workspace, which wastes a lot of memory. Can we have a global static workspace?

python/sglang/srt/layers/moe/token_dispatcher/flashinfer.py
```diff
@@ -0,0 +1,265 @@
...
+        self.workspace_size = moe_a2a_get_workspace_size_per_rank(
+            ep_size=self.ep_size,
+            max_num_tokens=self.max_num_tokens,
+            total_dispatch_payload_size_per_token=total_dispatch_payload_size_per_token,
+            combine_payload_size_per_token=combine_payload_size_per_token,
+        )
+
+        self.mapping = Mapping(
+            rank=self.ep_rank,
+            tp_size=self.ep_size,
```
ch-wan, line 133: Is this true? What's the difference with `world_size`?

python/sglang/srt/layers/moe/token_dispatcher/flashinfer.py
```diff
@@ -0,0 +1,265 @@
...
+
+MOE_NVFP4_DISPATCH = envs.SGLANG_MOE_NVFP4_DISPATCH.get()
+
+
+class FlashinferDispatchOutput(NamedTuple):
+    """Flashinfer EP dispatch output."""
+
+    hidden_states: torch.Tensor
+    hidden_states_scale: Optional[torch.Tensor]
+    topk_output: TopKOutput
```
ch-wan:
```suggestion
    topk_output: StandardTopKOutput
```

python/sglang/srt/layers/moe/token_dispatcher/flashinfer.py
```diff
@@ -0,0 +1,265 @@
...
+    # Provide an output tensor to fused_moe so it writes directly to our buffer
+    moe_output: Optional[torch.Tensor] = None
+
+    @property
+    def format(self) -> DispatchOutputFormat:
+        return DispatchOutputFormat.STANDARD
```
ch-wan:
```suggestion
        return DispatchOutputFormat.FLASHINFER
```

python/sglang/srt/layers/moe/token_dispatcher/flashinfer.py
```diff
@@ -0,0 +1,265 @@
...
+        topk_weights = topk_output.topk_weights
+
+        # Handle case where there are no tokens on this DP worker
+        # moe_a2a.dispatch requires at least one token
+        self.has_dummy_token = False
+        topk_ids_current_rank = None
+        if x.shape[0] == 0:
+            self.has_dummy_token = True
+            x = torch.empty(
+                (1, x.shape[1]),
```
ch-wan: Why is the shape `(1, *)`?

python/sglang/srt/layers/moe/token_dispatcher/flashinfer.py
```diff
@@ -0,0 +1,265 @@
...
+            max_num_tokens=self.max_num_tokens,
+            top_k=self.router_topk,
+            num_experts=self.num_experts,
+            workspace_size_per_rank=self.workspace_size,
+            mnnvl_config=MnnvlConfig(comm_backend=TorchDistributedCommBackend(group)),
+        )
+
+    def dispatch(
+        self, hidden_states: torch.Tensor, topk_output: TopKOutput
+    ) -> DispatchOutput:
```
ch-wan:
```suggestion
    ) -> FlashinferDispatchOutput:
```

python/sglang/srt/layers/moe/token_dispatcher/flashinfer.py
```diff
@@ -0,0 +1,265 @@
...
+                self.runtime_max_tokens_per_rank, self.hidden_size, output_dtype
+            ).view(-1, self.hidden_size)
+        return FlashinferDispatchOutput(
+            x,
+            x_sf,
+            StandardTopKOutput(topk_weights, topk_ids, topk_output.router_logits),
+            moe_output,
+        )
+
+    def combine(self, combine_input: CombineInput) -> torch.Tensor:
```
ch-wan:
```suggestion
    def combine(self, combine_input: FlashinferCombineInput) -> torch.Tensor:
```

python/sglang/srt/layers/quantization/modelopt_quant.py
```diff
@@ -1793,20 +1795,26 @@ def apply(
...
 
-            # If x_sf is not None, x is FP4 packed (half size), so we need * 2
-            # If x_sf is None, x is not packed, so output_col = x.shape[1]
-            output_col = x.shape[1]
-            if x_sf is not None and layer.moe_runner_config.is_gated:
-                output_col *= 2
-            with use_symmetric_memory(
-                get_tp_group(), disabled=not is_allocation_symmetric()
+            if (
+                hasattr(dispatch_output, "moe_output")
```
ch-wan: `hasattr` is hacky. We can use `DispatchOutputChecker`.

ch-wan (commented):
We also need to update this document: https://github.com/sgl-project/sglang/blob/94e1251131ca27260cb0e8938aeb7b4a4e630b19/docs/advanced_features/expert_parallelism.md#backends-for-all-to-all-communication

python/sglang/srt/layers/moe/token_dispatcher/flashinfer.py
trevor-m: This "Mapping" class is a bit confusing because it comes from the internals of TRT-LLM and has a bunch of functionality that is unused for this case. It probably shouldn't be the user facing API in flashinfer, but this combination of parameters is what works for our use cases.

python/sglang/srt/layers/moe/token_dispatcher/flashinfer.py
trevor-m: The workspace is global and shared between the layers

python/sglang/srt/layers/moe/token_dispatcher/flashinfer.py
trevor-m: Currently the flashinfer a2a requires at least 1 token per rank, but in sglang some dp workers can have 0 tokens. To work around this, we add a dummy token before dispatch and remove it after combine.

python/sglang/srt/layers/moe/token_dispatcher/flashinfer.py
trevor-m: Thanks @ch-wan for reviewing!

Done.

python/sglang/srt/layers/moe/token_dispatcher/flashinfer.py
trevor-m: Done

python/sglang/srt/layers/quantization/modelopt_quant.py
trevor-m: Done, thanks

fzyzcjy:
Hi, I am wondering whether this will be helpful for flashinfer_trtllm moe, and whether there will be support for it? Thanks!

trevor-m:
> Hi, I am wondering whether this will be helpful for flashinfer_trtllm moe, and whether there will be support for it? Thanks!

@fzyzcjy It looks like it should work based on https://github.com/NVIDIA/TensorRT-LLM/commit/e4bf29bc66013da39c641b85d0265590af082ea2
I can give it a try

ch-wan (commented):
LGTM in general. Can we add a unittest for this new feature? Also, we need to update this document: https://github.com/sgl-project/sglang/blob/94e1251131ca27260cb0e8938aeb7b4a4e630b19/docs/advanced_features/expert_parallelism.md#backends-for-all-to-all-communication

ch-wan (approved):

python/sglang/test/test_flashinfer_dispatcher.py
```diff
@@ -0,0 +1,322 @@
+import unittest
```
Fridge003, line 1: Can we move this test to `test/srt/ep` and register it at nightly test.
Can open a following PR for this

PR #14291, opened 2025-12-02
Tiny use trtllm_mha as default when possible

fzyzcjy:
## Motivation

wait for ci to see perf regression or errors

## Checklist

- [ ] Format your code according to the [Format code with pre-commit](https://docs.sglang.io/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.io/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.io/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.io/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.io/developer_guide/contribution_guide.html#benchmark-the-speed).
- [ ] Follow the SGLang code style [guidance](https://docs.sglang.io/developer_guide/contribution_guide.html#code-style-guidance).
- [ ] Work with maintainers to merge your PR. See the [PR Merge Process](https://github.com/sgl-project/sglang/blob/main/.github/MAINTAINER.md#pull-request-merge-process)

PR #14699, opened 2025-12-09
[CI] Fix Llama 3.1 8B FP4 CI

b8zhong:
This threshold has too little questions, after https://github.com/sgl-project/sglang/pull/14291

trtllm and fmha have no difference. It can now pass.

```
Accuracy: 0.559
Invalid: 0.003
Latency: 22.624 s
Output throughput: 7241.530 token/s
{'accuracy': np.float64(0.558756633813495), 'invalid': np.float64(0.003032600454890068), 'latency': 22.624223445018288, 'output_throughput': 7241.530318074861}
.
Ran 1 test in 60.780s

OK
```

Seperate question:
Why is

```
python3 benchmark/gsm8k/bench_sglang.py --num-questions 1319 --parallel 1319
```

Different from the other gsm8k. ? I get two different numbers.

### DIFF
--- a/test/srt/test_llama31_fp4.py
+++ b/test/srt/test_llama31_fp4.py
@@ -21,8 +21,6 @@ class TestLlama31FP4(unittest.TestCase):
         cls.base_url = DEFAULT_URL_FOR_TEST
         other_args = [
             "--trust-remote-code",
-            "--attention-backend",
-            "flashinfer",
             "--quantization",
             "modelopt_fp4",
         ]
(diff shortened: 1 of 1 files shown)


test/srt/test_llama31_fp4.py
```diff
@@ -40,18 +38,18 @@ def tearDownClass(cls):
...
-            parallel=128,
+            parallel=200,
             host=f"{parsed_url.scheme}://{parsed_url.hostname}",
             port=parsed_url.port,
         )
         metrics = run_eval_few_shot_gsm8k(args)
         print(metrics)
 
-        self.assertGreater(metrics["accuracy"], 0.61)
+        self.assertGreater(metrics["accuracy"], 0.54)
```
Fridge003, line 52: Is 0.54 the expected precision? Lowering the threshold will be risky

b8zhong: i think it is fine here yes, bc the num questions and num shots changed (mainly num q from 200 -> 1319 so it wil make a large difference)

Fridge003 (approved):

PR #14619, opened 2025-12-08
[Sparse & HICache]: Enables hierarchical sparse KV cache management and scheduling for DeepSeek V32.

hzh0425:
## Motivation

Previously, Sparse Attention was introduced in DeepSeekV32 Mode, which selects only the top-2048 tokens to participate in attention computation, significantly improving inference performance in long-context scenarios.

However, this model still requires **caching the full KV cache in GPU memory,** meaning the storage footprint of the KV cache is not reduced. This limits the number of concurrent long-context requests the system can handle.

Therefore, we aim to integrate HICache to implement a hierarchical DSA—where the full KV cache is stored in CPU or remote store, **and only 2,048 tokens per request are kept in GPU memory**. This significantly increases the batch size during the decode phase and improves overall decode throughput.

CoAuthors:  @huangtingwei9988  @xiezhq-hermann @LingYeAI and hicache team
Upstream Issue：https://github.com/sgl-project/sglang/issues/12826
**This is the upstream branch, and we will split it into multiple smaller PRs to move forward. （**ready now**）**
Todo PR:
- [x] Define Sparse Interface: https://github.com/sgl-project/sglang/pull/14741
- [x] Seperate Index_k pool and kvcache pool https://github.com/sgl-project/sglang/pull/15807
- [x] Adding SparseCoodinator and attention backend adaptor https://github.com/sgl-project/sglang/pull/16086
- [x] Support sparse diff kenrel  & Support IO Transfer Logic https://github.com/sgl-project/sglang/pull/16984
- [x] Support DeepSeek DSA & GLM5  https://github.com/sgl-project/sglang/pull/20343
- [ ] Support two batch overlap

## Modifications

1. We have **integrated HICache to build a unified hierarchical sparse framework** that supports various sparse attention algorithms—such as [Quest](https://arxiv.org/pdf/2406.10774) and [ClusterKV](https://arxiv.org/pdf/2412.03213)—for hierarchical, sparsity-aware KV cache management.
2. Implementation of Hierarchical DSA:

      - Decoupled the DSA indexer from KV cache allocation and the req_to_token_pool mechanism: the full indexer remains on the GPU.
      - During the decode phase, each request is allocated a fixed token space of 2,048 tokens in GPU memory.
      - Before each attention layer begins, a diff kernel identifies the missing KV cache entries in GPU memory, and HICache incrementally fetches only those missing portions from CPU memory. Our observations show that, on average, only about 20% of the KV cache needs to be fetched incrementally per layer.

image

[...]

PR #14741, opened 2025-12-09
[1/N][Sparse With Hicache]: Add Sparse Interface

hzh0425:
## Motivation

This PR primarily introduces a generic Retrievable Sparse Algorithm interface.
Upstream PR：https://github.com/sgl-project/sglang/pull/14619

## Checklist


python/sglang/srt/mem_cache/sparsity/algorithms/base_algorithm.py
```diff
@@ -0,0 +1,175 @@
...
+if TYPE_CHECKING:
+    from sglang.srt.model_executor.forward_batch_info import ForwardBatch
+
+
+class SparseMode(Enum):
+    """Sparse attention granularity mode."""
+
+    PAGE_WISE = "page_wise"
+    TOKEN_WISE = "token_wise"
+    DEEPSEEK_TOKEN_WISE = "deepseek_token_wise"
```
xiezhq-hermann: I am still having a hard time understand why Enum is more preferrable than polymorphism, can't we implement sparse type specific functions as member functions?

python/sglang/srt/mem_cache/sparsity/algorithms/deepseek_nsa.py
```diff
@@ -0,0 +1,74 @@
...
+)
+
+if TYPE_CHECKING:
+    from sglang.srt.layers.attention.nsa.nsa_indexer import Indexer
+    from sglang.srt.model_executor.forward_batch_info import ForwardBatch
+
+logger = logging.getLogger(__name__)
+
+
+class DeepSeekNSAAlgorithm(BaseSparseAlgorithm):
```
xiezhq-hermann: can't we inherit it from a common TokenSparseAlgorithm class? Or if there is no much difference between token and page based algorithms, we can just have more public functions implemented in base algorithm

python/sglang/srt/mem_cache/sparsity/algorithms/deepseek_nsa.py
```diff
@@ -0,0 +1,74 @@
...
+    def retrieve_topk(
+        self,
+        queries: torch.Tensor,
+        layer_id: int,
+        req_pool_indices: torch.Tensor,
+        sparse_mask: torch.Tensor,
+        attn_metadata: Optional[Any],
+        **kwargs,
+    ) -> tuple:
+        indexer: Optional["Indexer"] = kwargs.get("indexer")
```
xiezhq-hermann: should we have indexer as an member of the class?

python/sglang/srt/mem_cache/sparsity/algorithms/deepseek_nsa.py
hzh0425: Sure, just like what's done in 'page_wise_algorithm.py'.

It's just that currently, the token-wise implementation doesn't have a common parent class yet. I'll refactor it again.

python/sglang/srt/mem_cache/sparsity/algorithms/deepseek_nsa.py
hzh0425: I'll take a try.

hzh0425: After some thought, making the indexer a member variable would introduce strong invasiveness into the DeepSeek model layer in terms of implementation.

Since the indexer is initialized at the MLA layer, registering it into the algorithm would require fairly complex modifications.

Passing it directly here might be more convenient.

xiezhq-hermann (approved):

PR #14744, opened 2025-12-09
Re-add the API serving timing metrics.

hnyls2002:
To fix the CI broken in #13432 and unify the API serving metrics into a mixin class.


python/sglang/srt/managers/io_struct.py
zhanghaotong, line 769: Maybe we should delete this line?

PR #14745, opened 2025-12-09
fix rope parameter initialization error caused by transformers v5.0 update

yhyang201:
## Motivation

Many model configurations (e.g., Qwen3) follow a structure where `rope_theta` is located at the top level of the configuration dictionary.

**Example Config:**
```json
{
  "architectures": ["Qwen3ForCausalLM"],
  "model_type": "qwen3",
  "rope_scaling": null,
  "rope_theta": 1000000,
  "transformers_version": "4.51.0",
  ...
}
```

Legacy modeling code typically retrieves this parameter directly via `getattr(config, "rope_theta", None)`.

### The Issue
With the upgrade to **Transformers 5.0.0rc0+**, RoPE-related parameters (including `rope_theta`) are no longer guaranteed to be at the root level. They must often be retrieved from `config.rope_scaling` or `config.rope_parameters`.

This causes backward compatibility issues where models fail to load the correct RoPE configuration, leading to incorrect calculations or runtime errors.

### Temporary Solution
To quickly resolve this and maintain compatibility without rewriting every model architecture immediately, we introduce a temporary helper function: `handle_rope_parameters`.

```python
def handle_rope_parameters(config: PretrainedConfig) -> PretrainedConfig:
    """
    Temporary workaround to flatten rope_scaling parameters back into the main config
    for backward compatibility with Transformers v5.0+.
    """
    if hasattr(config, "rope_scaling"):
        rope_scaling = config.rope_scaling
        if isinstance(rope_scaling, dict):
            for k, v in rope_scaling.items():
                # If the parameter is missing in the root config, inject it from rope_scaling
                if not hasattr(config, k):
                    setattr(config, k, v)
    return
```

**Logic:**
This function checks if `rope_scaling` exists and is a dictionary. It then iterates through the keys (e.g., `rope_theta`, `type`, etc.) and sets them as direct attributes of the `config` object if they are missing. This allows existing modeling code to continue using `config.rope_theta` without modification.

### Long-term Goal
While the above solution works as a patch, the correct long-term fix is to refactor the modeling code for each architecture. We should update the models to natively fetch parameters from `config.rope_scaling` or `config.rope_parameters` instead of relying on this flattening workaround.

## Checklist

### DIFF
--- a/python/sglang/srt/configs/model_config.py
+++ b/python/sglang/srt/configs/model_config.py
@@ -80,7 +80,7 @@ def get_nsa_index_n_heads(config: PretrainedConfig) -> int:
     return config.index_n_heads
 
 
-def handle_rope_parameters(config: PretrainedConfig) -> PretrainedConfig:
+def handle_rope_parameters(config: PretrainedConfig):
     if hasattr(config, "rope_scaling"):
         rope_scaling = config.rope_scaling
         if isinstance(rope_scaling, dict):
--- a/python/sglang/srt/configs/model_config.py
+++ b/python/sglang/srt/configs/model_config.py
@@ -80,6 +80,16 @@ def get_nsa_index_n_heads(config: PretrainedConfig) -> int:
     return config.index_n_heads
 
 
+def handle_rope_parameters(config: PretrainedConfig) -> PretrainedConfig:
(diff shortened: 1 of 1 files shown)


byjiang1996 (approved):
👍

PR #14752, opened 2025-12-09
[SMG] perf: optimize tokenizer for reduced CPU and memory overhead

slin1237:
- Use Arc<Encoding> in L0 cache for zero-copy cache hits
- Use Arc<[TokenIdType]> in L1 cache for efficient token sharing
- Replace clone() with std::mem::take() in stop.rs to avoid allocations
- Use String::split_off() instead of drain().collect() for zero-copy buffer splitting
- Add #[inline] hints to hot path methods in sequence.rs, stream.rs, and traits.rs
- Pre-allocate vectors with known capacity in process_tokens() and step_batch()
- Combine multiple windows() calls into single-pass scan in is_likely_sentencepiece()

## Checklist

### DIFF
--- a/sgl-model-gateway/src/tokenizer/sequence.rs
+++ b/sgl-model-gateway/src/tokenizer/sequence.rs
@@ -86,11 +86,13 @@ impl Sequence {
     }
 
     /// Check if the sequence is empty
+    #[inline]
     pub fn is_empty(&self) -> bool {
         self.token_ids.is_empty()
     }
 
     /// Get the length of the sequence
+    #[inline]
     pub fn len(&self) -> usize {
         self.token_ids.len()
     }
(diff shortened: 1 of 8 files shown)


sgl-model-gateway/src/tokenizer/sequence.rs
```diff
@@ -86,11 +86,13 @@ impl Sequence {
     }
 
     /// Check if the sequence is empty
+    #[inline]
```
CatherineSue, line 89: Do we need the inline comment? I thought most of the time it is a default in Rust implicitly.

slin1237: yeah
i dont think its always, added just in case

PR #14771, opened 2025-12-10
Add retry logic for scheduled CI tests

alisonshao:
## Summary

This PR implements smart retry logic for CI tests that fail due to accuracy/performance assertion errors (not code errors). The implementation uses Python instead of shell scripts for better maintainability and enables per-file fine-grained retry.

### Key Features

- **Pattern-based error classification**: Distinguishes retriable failures (accuracy/performance assertions) from non-retriable failures (SyntaxError, ImportError, TypeError, etc.)
- **Per-file retry**: Retries only the specific test file that failed, saving time compared to re-running entire test suites
- **Output capture**: Captures test output to make intelligent retry decisions
- **GitHub Step Summary**: Writes retry information to GitHub Actions summary for visibility

### Changes

**`python/sglang/test/ci/ci_utils.py`**
- Added `RETRIABLE_PATTERNS` and `NON_RETRIABLE_PATTERNS` for error classification
- Added `is_retriable_failure()` function to determine if a failure should be retried
- Added `write_github_step_summary()` for GitHub Actions integration
- Extended `run_unittest_files()` with retry support (`enable_retry`, `max_retry_attempts`, `retry_wait_seconds`)

**`test/srt/run_suite.py`**
- Added `--enable-retry` flag to enable smart retry
- Added `--max-retry-attempts` flag (default: 3)
- Added `--retry-wait-seconds` flag (default: 60)

**`.github/workflows/pr-test.yml`**
- Added `SGLANG_CI_ENABLE_RETRY: true` environment variable
- Added `--enable-retry` to all `run_suite.py` invocations
- Increased step timeouts by ~10 minutes to account for potential retries

### Retriable vs Non-Retriable Errors

**Will retry:**
- `AssertionError: 0.77 not greater than 0.8`
- Failures containing "accuracy", "score", "latency", "throughput"
- Generic `AssertionError` (assumes retriable)

**Will NOT retry:**
- `SyntaxError`, `ImportError`, `ModuleNotFoundError`
- `TypeError`, `AttributeError`, `NameError`
- `RuntimeError`, `CUDA out of memory`, `OOM`
- `Segmentation fault`, `ConnectionRefusedError`, `FileNotFoundError`

### Example Output

```
[CI Retry] test_accuracy.py failed with retriable pattern: accuracy
[CI Retry] Waiting 60s before retry...

[CI Retry] Attempt 2/3 for test_accuracy.py

✓ PASSED on retry (attempt 2): test_accuracy.py
```

### DIFF
--- a/python/sglang/test/ci/ci_utils.py
+++ b/python/sglang/test/ci/ci_utils.py
@@ -107,7 +107,7 @@ def run_unittest_files(
     timeout_per_file: float,
     continue_on_error: bool = False,
     enable_retry: bool = False,
-    max_retry_attempts: int = 3,
+    max_attempts: int = 2,
     retry_wait_seconds: int = 60,
 ):
     """
@@ -120,7 +120,7 @@ def run_unittest_files(
                           If False, stop at first failure (default behavior for PR tests).
         enable_retry: If True, retry failed tests that appear to be accuracy/performance
                      assertion failures (not code errors).
-        max_retry_attempts: Maximum number of retry attempts per file (default: 3).
+        max_attempts: Maximum number of attempts per file including initial run (default: 2).
         retry_wait_seconds: Seconds to wait between retries (default: 60).
     """
     tic = time.perf_counter()
@@ -177,10 +177,10 @@ def run_unittest_files(
         file_passed = False
         was_retried = False
 
-        while attempt <= (max_retry_attempts if enable_retry else 1):
+        while attempt <= (max_attempts if enable_retry else 1):
             if attempt > 1:
                 print(
-                    f"\n[CI Retry] Attempt {attempt}/{max_retry_attempts} for {filename}\n",
+                    f"\n[CI Retry] Attempt {attempt}/{max_attempts} for {filename}\n",
                     flush=True,
                 )
                 was_retried = True
@@ -205,7 +205,7 @@ def run_unittest_files(
                     break
                 else:
                     # Check if we should retry
-                    if enable_retry and attempt < max_retry_attempts:
+                    if enable_retry and attempt < max_attempts:
                         output = "".join(output_lines)
                         is_retriable, reason = is_retriable_failure(output)
 
--- a/python/sglang/test/ci/ci_utils.py
+++ b/python/sglang/test/ci/ci_utils.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import re
 import subprocess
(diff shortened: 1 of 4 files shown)


scripts/ci/ci_run_with_retry.sh
Kangyan-Zhou, line 1: I wonder whether we should incorporate the logic in run_suite.py instead of adding a shell script wrapper. The shell script is not maintainable and hard to extend with additional features.

alisonshao:
https://github.com/sgl-project/sglang/actions/runs/20093883866

scripts/ci/ci_run_with_retry.sh
merrymercy: I agree.
- Please use python over shell
- In `run_suite.py`, we can do per-file fine-grained retry, which can save some time
- Last thing, when you retry, we also need to think about that it increased the total time so it will likely exceed the default step timeout

alisonshao: changed from shell to python. also please review and check if the change in timeout make sense. @Kangyan-Zhou @merrymercy

test/srt/run_suite.py
```diff
@@ -510,6 +510,24 @@ def main():
...
+    arg_parser.add_argument(
+        "--enable-retry",
+        action="store_true",
+        default=False,
+        help="Enable smart retry for accuracy/performance assertion failures (not code errors)",
+    )
+    arg_parser.add_argument(
+        "--max-retry-attempts",
+        type=int,
+        default=3,
```
Kangyan-Zhou: nit: let's change the default to 2, and maybe change the variable name to be max-attempt so that the naming is more clear (people don't have to think about whether the first run count as retry or not)

python/sglang/test/ci/ci_utils.py
```diff
@@ -49,77 +118,154 @@ def run_unittest_files(
...
-            # Otherwise continue to next test for nightly tests
+
+    elapsed_total = time.perf_counter() - tic
 
     if success:
-        print(f"Success. Time elapsed: {time.perf_counter() - tic:.2f}s", flush=True)
+        print(f"Success. Time elapsed: {elapsed_total:.2f}s", flush=True)
     else:
-        print(f"Fail. Time elapsed: {time.perf_counter() - tic:.2f}s", flush=True)
+        print(f"Fail. Time elapsed: {elapsed_total:.2f}s", flush=True)
```
Kangyan-Zhou: Can we change all the print to logger.info?

fzyzcjy:
sure

PR #14772, opened 2025-12-10
handling incomplete rope_scaling confign ci after transformers upgrade

yhyang201:
## Motivation

In previous versions, if `rope_scaling` was not empty, sglang would assume that certain keys existed in `rope_scaling` by default.

During the upgrade, many RoPE-related parameters were moved into `rope_scaling`, which caused many `rope_scaling` fields to become non-empty. As a result, the above assumption no longer holds.

Specifically for this issue:

The `rope_scaling` in the config of `lmsys/sglang-ci-dsv3-test` is `null` (which is incomplete, since it is inconsistent with dpskv3).
Under older versions of Transformers, sglang detected that `rope_scaling` was `null`, so it did not throw an error.
In newer versions of Transformers, `rope_scaling` is no longer `null`; instead, it contains values such as `rope_type` and `rope_theta`. sglang then assumes that keys like `factor` and `original_max_position_embeddings` also exist in `rope_scaling`, which leads to errors.

I believe that a proper model config would not trigger this problem. The root cause is the incomplete config of `lmsys/sglang-ci-dsv3-test`, so I applied a targeted fix for it.

## Checklist

### DIFF
--- a/python/sglang/srt/models/deepseek_v2.py
+++ b/python/sglang/srt/models/deepseek_v2.py
@@ -1287,7 +1287,7 @@ class DeepseekV2AttentionMLA(nn.Module):
         if rope_scaling:
             rope_scaling["rope_type"] = "deepseek_yarn"
             if "factor" not in rope_scaling:
-                rope_scaling["factor"] = 40
+                rope_scaling["factor"] = 4096
             if "original_max_position_embeddings" not in rope_scaling:
                 rope_scaling["original_max_position_embeddings"] = 4096
 
--- a/python/sglang/srt/models/deepseek_v2.py
+++ b/python/sglang/srt/models/deepseek_v2.py
@@ -1286,6 +1286,10 @@ class DeepseekV2AttentionMLA(nn.Module):
         # NOTE modification to rope_scaling must be done early enough, b/c e.g. Indexer needs it
         if rope_scaling:
             rope_scaling["rope_type"] = "deepseek_yarn"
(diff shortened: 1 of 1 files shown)


python/sglang/srt/models/deepseek_v2.py
```diff
@@ -1286,6 +1286,10 @@ def __init__(
         # NOTE modification to rope_scaling must be done early enough, b/c e.g. Indexer needs it
         if rope_scaling:
             rope_scaling["rope_type"] = "deepseek_yarn"
+            if "factor" not in rope_scaling:
+                rope_scaling["factor"] = 40
```
Fridge003: Will hardcoding them to 40/4096 be risky

zhyncs: @yhyang201 can we just handle `lmsys/sglang-ci-dsv3-test` here

yhyang201: It should apply only to the `lmsys/sglang-ci-dsv3-test` model, because all the DPSK models have rope_scaling configured, and only `lmsys/sglang-ci-dsv3-test` does not.
Do we need to add an additional check specifically for `lmsys/sglang-ci-dsv3-test`?

yhyang201: > Will hardcoding them to 40/4096 be risky

I think not; this will only be hard-coded if these keys do not exist.

yhyang201:
move to #14784

PR #14781, opened 2025-12-10
[Performance] optimize NSA backend metadata computation for multi-step speculative decoding

Johnsonms:
## Summary

This PR optimizes metadata initialization in the NSA backend for multi-step speculative decoding.

The key idea is to replace repeated per-backend metadata computation with a **precompute-once-copy-many** strategy. Shared metadata is computed once and then copied/reused across backend instances, reducing decode-path overhead without changing model computation or speculative decoding behavior.

## Approach

The old path computed metadata separately for each backend instance:

```text
Old:
compute metadata N times, one per backend ≈ 210µs × N
```

The optimized path computes shared metadata once and then copies/reuses it:

```text
New:
compute shared metadata once copy / reuse it for each backend ≈ 210µs + 70µs × N
```

This reduces repeated metadata preparation, especially when using **4–8 speculative steps**, where the old path repeatedly initialized equivalent metadata across backend instances.

### Metadata Preparation

- **Metadata computation:** **31.9% faster**, 642µs → 437µs
- **Kernel launches:** **33.3% fewer**, 60 → 40

### End-to-End Testing

- **Single request, 16K sequence length:** **+12.4% TPS**
- **4 concurrent requests:** **+5.5% to +7.5% TPS** across 16K–64K sequence lengths
- **TTFT impact:** minimal, within approximately **±3%**
- **E2E benchmark tests:** approximately **+5% to +15% TPS**

Overall, the metadata initialization path is **1.7×–2.2× faster** with **4–8 speculative steps**, where the old path repeatedly initialized equivalent metadata across backend instances.

image
image

Opt: 437us per forwards in draft meta forward part, will lanuch 40 kernels

 
image

image

## Implementation

- Added `PrecomputedMetadata` to store shared replay metadata.
- Added `_precompute_replay_metadata()` to compute metadata once for all replay backends.
- Added `init_forward_metadata_replay_cuda_graph_from_precomputed()` as the fast replay initialization path.
- Added mode-specific precomputation for:
  - `decode`
  - `target_verify`
  - `draft_extend`

Conceptually, the change is:

```text
Before:
metadata recomputed per backend instance

After:
metadata computed once and reused across backend instances
```

## Correctness

This optimization only changes metadata preparation. It does not change the speculative decoding algorithm, model computation, attention semantics, or acceptance behavior.


Fridge003:
Also can you post the accuracy result of GPQA/AIME benchmark?

Johnsonms:
> Also can you post the accuracy result of GPQA/AIME benchmark?

Added the test result, Thanks @Fridge003

python/sglang/srt/layers/attention/nsa_backend.py
```diff
@@ -231,6 +231,39 @@ def compute_cu_seqlens(seqlens: torch.Tensor) -> torch.Tensor:
     )
 
 
+@dataclass
```
Fridge003: Can we create a new file `nsa_backend_mtp_precompute.py`, and put all the precompute related codes there? Seems all the precomputation logics are standalone

A Mixin class `NativeSparseAttnBackendMTPPrecomputeMixin` can be created to inherit methods.

python/sglang/srt/layers/attention/nsa_backend.py
```diff
@@ -231,6 +231,39 @@ def compute_cu_seqlens(seqlens: torch.Tensor) -> torch.Tensor:
...
+class PrecomputedMetadata:
+    """Precomputed metadata shared across multiple backend instances.
+
+    Used for multi-step speculative decoding where multiple backends
+    need identical metadata. Precomputing once and copying N times
+    is much faster than computing N times.
+
+    Performance: ~3-5x speedup for 4-8 speculative steps.
```
Fridge003: What's the baseline of ~3-5x speedup? Add some description here

python/sglang/srt/layers/attention/nsa_backend.py
```diff
@@ -858,6 +891,350 @@ def init_forward_metadata_replay_cuda_graph(
...
+        forward_mode: ForwardMode,
+        spec_info: Optional[SpecInput],
+    ) -> PrecomputedMetadata:
+        """Precompute all shared metadata for multi-step backends.
+
+        This function extracts and computes all operations that are
+        identical across different backend instances in multi-step
+        speculative decoding.
+
+        Performance: Saves ~(N-1) × 175μs for N backend instances.
```
Fridge003: Maybe remove 175us here since the number changes

python/sglang/srt/layers/attention/nsa_backend.py
```diff
@@ -1559,14 +1936,22 @@ def init_forward_metadata_capture_cuda_graph(self, forward_batch: ForwardBatch):
...
+
+        # Fast copy to each backend (1-2x faster than computing N times)
         for i in range(self.speculative_num_steps):
-            self.attn_backends[i].init_forward_metadata_replay_cuda_graph(
-                bs,
-                forward_batch.req_pool_indices,
-                forward_batch.seq_lens,
-                seq_lens_sum=-1,
-                encoder_lens=None,
+            self.attn_backends[
```
Fridge003: Please add an environ that controls the precomputing of metadata, and put it under `python/sglang/srt/layers/attention/nsa/utils.py`

python/sglang/srt/layers/attention/nsa_backend.py
```diff
@@ -1559,14 +1936,22 @@ def init_forward_metadata_capture_cuda_graph(self, forward_batch: ForwardBatch):
     def init_forward_metadata_replay_cuda_graph(
         self, forward_batch: ForwardBatch, bs: int
     ):
+        # Precompute metadata once (shared across all backends)
+        precomputed = self.attn_backends[0]._precompute_replay_metadata(
+            bs=bs,
+            req_pool_indices=forward_batch.req_pool_indices,
+            seq_lens=forward_batch.seq_lens,
+            seq_lens_cpu=forward_batch.seq_lens_cpu,
+            forward_mode=ForwardMode.DECODE,
```
Fridge003: It's correct to fix the forward_mode to decode here.
But since it's the only place for precomputing metadata, will be the verify/draft_extend metadata computation be useless?

python/sglang/srt/layers/attention/nsa_backend.py
```diff
@@ -858,6 +891,350 @@ def init_forward_metadata_replay_cuda_graph(
...
+    def init_forward_metadata_replay_cuda_graph_from_precomputed(
+        self,
+        bs: int,
+        precomputed: PrecomputedMetadata,
+        forward_mode: ForwardMode,
+    ):
+        """Fast path: copy precomputed metadata to this backend's metadata.
+
+        This function only performs copy operations, no computation.
+        Performance: ~20μs (vs ~175μs for full computation).
```
Fridge003: Also remove the hardcoded data here

Fridge003: Also remove the hardcoded numbers here

Fridge003 (changes requested):

python/sglang/srt/layers/attention/nsa_backend.py
Johnsonms: This is currently only for draft decoding. I’ll work on the verify / draft-extend metadata next, but since it’s a single pass, there may not be much room for further optimization.

python/sglang/srt/layers/attention/nsa_backend.py
Johnsonms: Yeap, get it done

python/sglang/srt/layers/attention/nsa_backend.py
Johnsonms: Removed

python/sglang/srt/layers/attention/nsa_backend.py
Johnsonms: Removed, Thanks @Fridge003

python/sglang/srt/layers/attention/nsa_backend.py
Johnsonms: Get it done, Thanks @Fridge003

Johnsonms:
> Do you have data of acceptance length? Will it drop after this PR?
I verified this and confirmed there is no change in the acceptance rate. The optimization focuses on NativeSparseAttnMultiStepBackend::init_forward_metadata_replay_cuda_graph. In this path, the same backend metadata is initialized multiple times (at least three) inside a loop, **without any changes between iterations**. By precomputing the metadata once and reusing it, we replace three compute-and-copy passes with three simple copies, which saves time.


Before:
image

After:
image

Fridge003 (approved):

Issue #14510, opened 2025-12-05
[CI] Use smaller models in CI


b8zhong:
### Checklist

- [x] If this is not a feature request but a general question, please start a discussion at https://github.com/sgl-project/sglang/discussions. Otherwise, it will be closed.
- [x] Please use English. Otherwise, it will be closed.

### Motivation

If the model is the exact same architecture, we can use a smaller model to speed up CI? Currently some tests are taking a long time.

E.g, Gemma 27B -> 1B

We may also have to adjust accuracy related tests if this happens.

PR #14786, opened 2025-12-10
fix b200 ci

b8zhong:
After rope fix the accuracy went up...

1. Llama FP4 tests takes only around 49 seconds. Keep 90 buffer in case anything happens.

2. GPT-OSS is hogging CI, especially the weight shuffle part for MXFP4 variant. 

E.g `[2025-12-10 04:13:52] Shuffling MoE weights for FlashInfer MXFP4 moe kernel (layer: model.layers.1.mlp.experts), it might take a while...`, for the 120B it takes around 5 minutes. https://github.com/sgl-project/sglang/issues/14510

20b is the exact same thing but less layers (confirmed in HF `config.json`). Just use this it's around 2x faster.

`python test/srt/test_llama31_fp4.py`
```
Accuracy: 0.650
Invalid: 0.002
Latency: 17.051 s
Output throughput: 6909.456 token/s
{'accuracy': np.float64(0.6497346474601972), 'invalid': np.float64(0.001516300227445034), 'latency': 17.05141551700217, 'output_throughput': 6909.455691964357}
.
Ran 1 test in 49.492s

OK
```

`CUDA_VISIBLE_DEVICES=7 python test/srt/test_gpt_oss_1gpu.py`
```
.
Ran 2 tests in 252.578s

OK
```

### DIFF
--- a/test/srt/run_suite.py
+++ b/test/srt/run_suite.py
@@ -171,7 +171,7 @@ suites = {
         TestFile("test_deepseek_v3_fp4_4gpu.py", 1800),
         TestFile("test_flash_attention_4.py", 300),
         TestFile("test_fp8_blockwise_gemm.py", 280),
-        TestFile("test_gpt_oss_1gpu.py", 300),
+        TestFile("test_gpt_oss_4gpu.py", 700),
         TestFile("test_llama31_fp4.py", 90),
         TestFile("test_eagle_infer_beta_dp_attention.py", 300),
     ],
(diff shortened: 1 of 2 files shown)


Fridge003 (approved):

test/srt/run_suite.py
```diff
@@ -171,8 +171,8 @@
         TestFile("test_deepseek_v3_fp4_4gpu.py", 1800),
         TestFile("test_flash_attention_4.py", 300),
         TestFile("test_fp8_blockwise_gemm.py", 280),
-        TestFile("test_gpt_oss_4gpu.py", 600),
-        TestFile("test_llama31_fp4.py", 300),
+        TestFile("test_gpt_oss_1gpu.py", 300),
```
Fridge003: It should be `test_gpt_oss_4gpu`, 1gpu is not for b200 test

Fridge003 (changes requested):

PR #14792, opened 2025-12-10
[Qwen3-next] support mamba radix cache for overlap scheduler

hanming-lu:
## Motivation
- Currently, qwen3-next doesn't support 1) overlap scheduler; 2) page_size > 1; 3) branching point caching

## Modifications

1. Support overlap scheduler for qwen3-next
2. Support page_size > 1 for qwen3-next
3. Support branching point caching for qwen3-next
4. tested for (ps = 1, ps > 1) x (non-spec dec, sd topk1, sd topk>1). All work except for ps > 1 + sd topk > 1, which is not supported on main yet
5. enable with `--mamba-scheduler-strategy extra_buffer`
6. Better memory allocation for ssm spec dec - instead of coupling spec dec intermediate state size with total ssm states, couple it with max running requests.

## Accuracy Tests

1. Added mamba radix cache KL tests for prefill and decode
2. cover both `--mamba-scheduler-strategy extra_buffer` and `no_buffer`

## Checklist


python/sglang/srt/managers/scheduler_output_processor_mixin.py
```diff
@@ -359,6 +361,34 @@ def process_batch_result_decode(
                 req.output_ids.extend(next_token_id)
                 new_accepted_len = len(next_token_id)
 
+            # Update Mamba
+            seq_len = len(req.origin_input_ids) + len(req.output_ids) - 1
```
yizhang2077: wrap it into a function named like `mamba_prefix_cache_xxx`

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -1404,6 +1404,54 @@ def prepare_for_extend(self):
                 req.already_computed = seq_len
             req.is_retracted = False
 
+            if self.is_hybrid_gdn_cache:
+                mask = (req.extend_input_len // FLA_CHUNK_SIZE) * FLA_CHUNK_SIZE > 0
+                mamba_track_mask_cpu.append(mask)
+                mamba_track_indices_cpu.append(
+                    req.mamba_ping_pong_track_buffer[req.mamba_next_track_idx].item()
```
yizhang2077: it will cause device/host sync here?

python/sglang/srt/managers/scheduler_output_processor_mixin.py
hanming-lu: done.

python/sglang/srt/managers/schedule_batch.py
hanming-lu: yes

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -1404,6 +1427,54 @@ def prepare_for_extend(self):
                 req.already_computed = seq_len
             req.is_retracted = False
 
+            if get_global_server_args().enable_mamba_radix_cache_v2:
```
yizhang2077: wrap it into a function named `mamba_radix_cache_xxx`

python/sglang/srt/managers/schedule_batch.py
yizhang2077: will it break overlap schedule?

hanming-lu: Checked the profile, before it, some functions also triggers sync such as `alloc_for_extend()`. This new one (the one with arrow) is pretty small compared to existing ones, so no impact.
Screenshot 2025-12-10 at 9 24 08 PM

python/sglang/srt/managers/schedule_batch.py
hanming-lu: done

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -1520,6 +1568,60 @@ def prepare_for_extend(self):
...
+            if req.mamba_branching_seqlen is not None:
+                # track branching point in this forward if the branching point
+                # is within the current extend batch.
+                branching_seqlen_aligned_mask = (
+                    req.mamba_branching_seqlen - len(req.prefix_indices)
+                ) % FLA_CHUNK_SIZE == 0
+                if (
+                    req.mamba_branching_seqlen > len(req.prefix_indices)
+                    and req.mamba_branching_seqlen < mamba_track_seqlen
+                    and branching_seqlen_aligned_mask
```
yizhang2077: it looks like `req.mamba_branching_seqlen` always be multiple of `FLA_CHUNK_SIZE`, but `prefix_indices` is not like that. I wonder if the `branching_seqlen_aligned_mask` is false in the most case

python/sglang/srt/server_args.py
```diff
@@ -299,6 +300,7 @@ class ServerArgs:
     swa_full_tokens_ratio: float = 0.8
     disable_hybrid_swa_memory: bool = False
     radix_eviction_policy: str = "lru"
+    mamba_track_interval: int = 256
```
yizhang2077: duplicate define

test/srt/models/test_qwen3_next_models.py
```diff
@@ -71,9 +91,35 @@ def test_input_output_logprobs_match_decode_cache_hit(self):
...
+        # Second request won't have cache hit, but will cache the branching point
+        # Third request will have cache hit on the branching point
+        for i, (suffix, cache_hit) in enumerate(
+            zip(suffix_list, cache_hit_list, strict=True)
+        ):
+            result = send_request_helper(self.base_url, text_prefix + suffix)
+            cached_tokens = result["meta_info"]["cached_tokens"]
+            if cache_hit:
+                assert (
+                    cached_tokens > 0
```
yizhang2077: I think the shape of tree in test can be more complex, and cached_tokens we can directly predict (256)

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -1511,6 +1542,23 @@ def prepare_for_extend(self):
         self.extend_logprob_start_lens = [r.extend_logprob_start_len for r in reqs]
         self.extend_input_logprob_token_ids = extend_input_logprob_token_ids
 
+        if get_global_server_args().enable_mamba_radix_cache_v2:
```
yizhang2077: I think this can also be wrapped in `_mamba_radix_cache_v2_prepare_for_extend`

python/sglang/srt/mem_cache/mamba_radix_cache.py
```diff
@@ -384,10 +435,12 @@ def match_prefix(self, key: RadixKey, **kwargs) -> MatchResult:
                 last_host_node=self.root_node,
             )
 
-        value, last_node = self._match_prefix_helper(key)
+        value, last_node, mamba_branching_seqlen = self._match_prefix_helper(key)
 
         # copy mamba state to req local space if cow is true
         if cow_mamba and last_node.mamba_value is not None:
+            assert req.req_pool_idx is None  # req_pool_idx is uninitialed
```
yizhang2077: remove this assertion

python/sglang/srt/mem_cache/mamba_radix_cache.py
```diff
@@ -793,7 +954,19 @@ def _match_prefix_helper(
             )
             node_update = node_update.parent
 
-        return value[:best_value_len], best_last_node
+        # Calculate the branching point. It is defined as the last aligned position that
+        # does not have a mamba value.
+        if len(value) > best_value_len:
+            fla_chunk_aligned_seqlen = (
+                sum(len(v) for v in value) // FLA_CHUNK_SIZE
```
yizhang2077, line 959: why we do not use  `(len(value) // FLA_CHUN_SIZE) * FLA_CHUN_SIZE` directly here?

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -1520,6 +1568,60 @@ def prepare_for_extend(self):
...
+                    req.mamba_branching_seqlen > len(req.prefix_indices)
+                    and req.mamba_branching_seqlen < mamba_track_seqlen
+                    and branching_seqlen_aligned_mask
+                ):
+                    # NOTE: See the comment above for mamba_track_seqlen, the +1 is necessary
+                    # because the branching point is not the last aligned position, so we need
+                    # to retrieve its state from h. Adding 1 will give us the correct index in h,
+                    # otherwise the calculation will retrieve the state from the last_recurrent_state,
+                    # which is not correct.
+                    mamba_track_seqlen = req.mamba_branching_seqlen + 1
```
yizhang2077, line 1621: why do we need +1 in branching point while in non-branching point we do not need?

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -823,11 +832,13 @@ def init_next_round_input(self, tree_cache: Optional[BasePrefixCache] = None):
                 self.last_node,
                 self.last_host_node,
                 self.host_hit_length,
+                self.mamba_branching_seqlen,
```
yizhang2077, line 836: I think when disable radix cache, it will cause error for extra input/output

python/sglang/srt/server_args.py
hanming-lu: good catch

python/sglang/srt/managers/schedule_batch.py
hanming-lu: `_mamba_radix_cache_v2_prepare_for_extend` is called per request, and here depends on all requests to be processed before sending to gpu. So probably couldn't wrap it in

python/sglang/srt/mem_cache/mamba_radix_cache.py
hanming-lu: I can remove, but what's the reason for that?

python/sglang/srt/mem_cache/mamba_radix_cache.py
hanming-lu: value is a list of Tensor, each tree node match append a Tensor to it. So if a cache hit is 4k long, and spread into 2 tree nodes, then `len(value)==2` and `sum(len(v) for v in value)==4k`

python/sglang/srt/managers/schedule_batch.py
hanming-lu: class MatchResult has `mamba_branching_seqlen: Optional[int] = None` so it defaults to None. `kimi linear` disables radix cache and the test `test/srt/models/test_kimi_linear_models.py` passes

python/sglang/srt/managers/schedule_batch.py
hanming-lu: Good question. From our design, `mamba_branching_seqlen % CHUNK_SIZE == 0` and `mamba_branching_seqlen` < `chunked prefill size for the request`. So based on the logic below, we should always retrieve the state from h. But the math afterwards, i.e. that calculates if we should retrieve from `h` or `final`, doesn't know if it is `chunked prefill size for the request` or `mamba_branching_seqlen`, so we add 1 to ask it to retrieve from `h` instead of `final` (i.e. trigger 3. instead of 2. below)

```
1. If prefill length is < 64, then only final state has value. In this case we cache final state
2. if prefill length == 64, then only final state has value. In this case we cache pos 64, from `final` state
3. if prefill length >64 and < 128, then both h and final state have value. We cache pos 64 from `h` state
4. if prefill length ==128, then both h and final state have value. We cache pos 128 from final state. Note h doesn't include the pos 128.
```

test/srt/models/test_qwen3_next_models.py
hanming-lu: 1. both gsm8k and prefill kl test covers more complex tree shape for branching, specifically the fewshot point is the branching point for gsm8k. This test is only for testing if branching is on.
2. done.

python/sglang/srt/managers/schedule_batch.py
yizhang2077: according to our discussion, the prefix length is less than 64 or the multiple of 64, could we add some assertion here?

hanming-lu: > prefix length is less than 64 or the multiple of 64

I think the assertion will fail. For each forward iteration, this is true. But it is possible that a cache hit with multiple of 64 and then next chunked prefill is less than 64, let's say M. Then the next request that hits this cache point will be 64 * N + M

test/srt/models/test_qwen3_next_models.py
```diff
@@ -98,6 +145,8 @@ def setUpClass(cls):
                 "0.8",
                 "--tp",
                 "4",
+                "--chunked-prefill-size",
```
yizhang2077, line 149: I think for test we need to test for both `enable_mamba_radix_cache_v2` or not, currently `enable_mamba_radix_cache_v2` is not tested

python/sglang/srt/server_args.py
```diff
@@ -1289,12 +1292,6 @@ def _handle_model_specific_adjustments(self):
                         f"{model_arch}"
                     )
         elif model_arch in ["Qwen3NextForCausalLM"]:
-            if not self.disable_radix_cache:
-                logger.warning(
```
yizhang2077, line 1337: I think if we use mamba radix cache while not set `enable_mamba_radix_cache_v2`, we still need disable overlap schedule?

python/sglang/srt/speculative/eagle_worker.py
```diff
@@ -751,6 +753,25 @@ def verify(self, batch: ScheduleBatch, spec_info: EagleVerifyInput):
                 )
                 + 1
             )
+            cumulative_accepted_lengths = torch.cumsum(accepted_length, dim=0)
```
yizhang2077: I think currently linear models verify logic is too heavy/long in main path, let's wrap it into a separate function

python/sglang/srt/managers/schedule_batch.py
yizhang2077: I see

test/srt/models/test_qwen3_next_models.py
hanming-lu: we test `--enable-mamba-radix-cache-v2` for `TestQwen3Next` and `TestQwen3NextMTPTopk`, and v1 for `TestQwen3NextMTP`. 

I think this split makes sense because non-sd and sd are two paths for v2, so covered. and all v1 is the same path, so also covered.

python/sglang/srt/mem_cache/mamba_radix_cache.py
yizhang2077: I think it may have trouble in retract since `req_pool_idx` is not set to none?

python/sglang/srt/server_args.py
hanming-lu: yup, handled here: https://github.com/sgl-project/sglang/pull/14792/files#diff-700b5118b493d60d7b5994857f5f1e6a7e842ad702392b8ab199945764dfc8edR1350-R1355

python/sglang/srt/speculative/eagle_worker.py
hanming-lu: done.

python/sglang/srt/mem_cache/mamba_radix_cache.py
hanming-lu: I see. We have retract test internally and we haven't hit any issue, it's possible that retract is done differently in oss, will delete it for now then.

test/srt/models/test_qwen3_next_models.py
yizhang2077: I see

python/sglang/srt/server_args.py
yizhang2077: I see

python/sglang/srt/disaggregation/decode.py
```diff
@@ -155,8 +155,14 @@ def __init__(
...
         self.enable_memory_saver = enable_memory_saver
+        self.enable_mamba_radix_cache_v2 = False
         self._init_mamba_pool(
-            size + pre_alloc_size, cache_params, device, speculative_num_draft_tokens
+            size + pre_alloc_size,
+            size + pre_alloc_size,
+            cache_params,
+            device,
+            self.enable_mamba_radix_cache_v2,
+            speculative_num_draft_tokens,
```
ShangmingCai: Does it mean v2 is not supported for PD disaggregation? What about the prefill side when PD is enabled? Should we set this flag to False as well?

hanming-lu: I am not familiar with PD in oss so not able to add it yet. Added an assert. It should be straightforward to add for anyone familiar with PD

yizhang2077 (approved):

hebiao064 (commented):
1st Pass: Kernel change LGTM, @byjiang1996 pls also take a look to ensure spec topk>1 won't break

hebiao064 (approved):
Overall LGTM

python/sglang/srt/server_args.py
```diff
@@ -457,6 +458,8 @@ class ServerArgs:
     max_mamba_cache_size: Optional[int] = None
     mamba_ssm_dtype: str = "float32"
     mamba_full_memory_ratio: float = 0.9
+    mamba_track_interval: int = 256
+    enable_mamba_radix_cache_v2: bool = False
```
merrymercy: move this to choices.

merrymercy: ## Code change
1. Change all code references to `enable_branch`, `enable_overlap`
     - if enable_overlap
          -`enable_branch` is True for qwen-next and is False for all other models. (This depends on the kernel support).
    - if not enable_overlap:
         - `enable_branch` is False for all models. (Although it is possible, we do not have a plan to implement it).
2. Change the server arg to an env var `SGLANG_ENABLE_MAMBA_CACHE_OVERLAP`
    - If it is True, follow the argument `--disable-overlap-schedule`.
    - If it is False, forcely set `--disable-overlap-schedule` to True.
    - Currently, it is False by default (if not set), we change it to `True` and remove it in one month.

## Document the tradeoff
There are four cases: [branch, non-branch] x [overlap, non-overlap]
- Branch is strictly better than non-branch, but it needs kernel support. Some models cannot run branch.
- Overlap doubles the memory usage of mamba state. It is strictly better for non-kvcache bound cases. It might be better or worse for kvcache bound cases.

python/sglang/srt/managers/scheduler.py
```diff
@@ -402,7 +402,7 @@ def __init__(
 
         # Hybrid memory pool
         self.is_hybrid_swa = self.tp_worker.is_hybrid_swa
-        self.is_ssm_model = (
+        self.is_hybrid_gdn = (
```
merrymercy: todo: is_hybrid_ssm

python/sglang/srt/disaggregation/decode.py
merrymercy: @ShangmingCai Once this is merged. Can you help support PD + this PR?

python/sglang/srt/managers/scheduler.py
hanming-lu: done.

docs/advanced_features/server_arguments.md
```diff
@@ -303,6 +303,8 @@ Please consult the documentation below and [server_args.py](https://github.com/s
...
 | `--mamba-ssm-dtype` | The data type of the SSM states in mamba cache. | `float32` | `float32`, `bfloat16` |
 | `--mamba-full-memory-ratio` | The ratio of mamba state memory to full kv cache memory. | `0.2` | Type: float |
+| `--mamba-scheduler-strategy` | The strategy to use for mamba scheduler. `auto` currently defaults to `no_buffer`. 1. `no_buffer` does not support overlap scheduler due to not allocating extra mamba state buffers. Branching point caching support is feasible but not implemented. 2. `extra_buffer` supports overlap schedule by allocating extra mamba state buffers to track mamba state for caching (increases mamba state usage per running req by 2 for non-speculative decoding, 1 for speculative decoding). 2a. `extra_buffer` is strictly better for non-KV-cache-bound cases; for KV-cache-bound cases, the tradeoff depends on whether enabling overlap outweighs reduced max running requests. 2b. mamba caching at radix cache branching point is strictly better than non-branch but requires kernel support (currently only FLA backend), currently only extra_buffer supports branching. | `auto` | `auto`, `no_buffer`, `extra_buffer` |
```
merrymercy: What is the meaning of `(increases mamba state usage per running req by 2 for non-speculative decoding)`? Can we use descriptions such as increase by 1.2x or  1.5x?

merrymercy (approved):

mingyuedang:
求指教，非投机解码开启extra_buffer给**每个请求**分配**2个slot**作用是啥

PR #14842, opened 2025-12-10
fix: trtllm mha attention auto-selection on sm120

b8zhong:
TRTLLM MHA does not support SM120, SM120 has Flashinfer has the best performing attention backend. 

Fix https://github.com/sgl-project/sglang/issues/14814


b8zhong:
/tag-and-rerun-ci

Also, this case can't be covered by CI (there is no SM120 device)

Fridge003 (approved):

Qiaolin-Yu (approved):

PR #14465, opened 2025-12-05
Fix safetensors validation to catch corruption after download

alisonshao:
## Summary
- Always validate local cache first (removed `is_in_ci()` gate around validation)
- Add post-download validation via new `_validate_weights_after_download()` function
- Add `SGLANG_IS_IN_CI=true` to GPU CI workflows (pr-test.yml, nightly-test-nvidia.yml)

## Problem
The safetensors validation only ran when `SGLANG_IS_IN_CI=true` was set, but this env var was missing from GPU workflows. Additionally, validation only checked cached files - freshly downloaded files were never validated.

This caused CI failures like:
```
safetensors_rust.SafetensorError: Error while deserializing header:
invalid JSON in header: EOF while parsing a value at line 1 column 0
```

Related CI failure: https://github.com/sgl-project/sglang/actions/runs/19948236909/job/57203231359

## Test plan
- [x] Verify existing CI tests pass
- [x] Verify corrupted safetensors files are detected both in cache and after download

PR #14849, opened 2025-12-10
fix: restrict cache validation behaviors to CI only

alisonshao:
## Summary
Address review feedback from #14465 - cache validation and cleanup behaviors should only run in CI environments to avoid unnecessary overhead for regular users.

## Changes
- Gate incomplete file detection/cleanup with `is_in_ci()`
- Gate sharded model validation with `is_in_ci()`
- Gate post-download validation retry loop with `is_in_ci()`

## Behavior
- **In CI** (`SGLANG_IS_IN_CI=true`): Full validation, corruption detection, cleanup, and retry logic is enabled
- **Outside CI**: Simple file checks and downloads without validation overhead

## Test plan
- [ ] Verify CI tests still pass with validation enabled
- [ ] Verify local usage doesn't trigger validation overhead

### DIFF
--- a/python/sglang/srt/model_loader/weight_utils.py
+++ b/python/sglang/srt/model_loader/weight_utils.py
@@ -47,6 +47,7 @@ from sglang.srt.model_loader.weight_validation import (
     _validate_sharded_model,
 )
 from sglang.srt.utils import find_local_repo_dir, log_info_on_rank0, print_warning_once
+from sglang.utils import is_in_ci
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


Kangyan-Zhou:
@alisonshao issue with rebase?

python/sglang/srt/model_loader/weight_utils.py
merrymercy, line 430: do not print this line

merrymercy, line 267: you should skip the whole function if not in CI

PR #14883, opened 2025-12-11
[DLLM] Implement initial dynamic batching for diffusion LLM

ClawSeven:
## Motivation

The current dLLM implementations lack batching capabilities, with batch size set to 1 by default. Additionally, the existing dLLM implementation is tightly coupled with chunked prefill execution, limiting flexibility and extensibility.

## Modifications

This PR introduces two key improvements:
- Basic Batching Support: Enables the dLLM to batch incoming requests when the number of running requests has not reached the configured limit. New requests will be grouped with existing batches to improve resource utilization.
- Decoupling of dLLM and Chunked Prefill: Separates the dLLM logic from the chunked prefill execution, laying the foundation for future enhancements such as radix cache support and overlap scheduling.

No new arguments are required—this relies solely on the existing max-running-requests configuration. Note that this is an initial implementation of DLLM batching and is not yet optimized. Currently, prefill and decoding requests are batched together, which may lead to redundant computations. We will address these optimizations in subsequent PRs

Currently, we still recommend using batch size 1 for running dLLM until the batching optimization PRs are merged.

## Accuracy Tests

Tested on H20
```
Accuracy: 0.900
Invalid: 0.000
Latency: 263.453 s
Output throughput: 97.638 token/s
metrics={'accuracy': 0.9, 'invalid': 0.0, 'latency': 263.45285430923104, 'output_throughput': 97.63796284327712}
```

Here I added a dLLM batching‑accuracy test instead of changing the max‑running‑requests argument in the previous LLaDA unit test, since the current batching performance is suboptimal. Once the optimization PRs are merged, I’ll remove this test.

## Checklist


python/sglang/srt/managers/schedule_batch.py
```diff
@@ -1134,6 +1134,59 @@ def __repr__(self):
         )
 
 
+class DllmReqs:
```
Monstertail: I noticed that the DllmReqs will not be passed as Forwardbatch to the algorithm interface yet. I wonder whether we will pass it to the algorithm interface in the future to enable the token shift for algorithms like fast-dllm-v2.  @ClawSeven It looks like passing the DllmReq to the algorithm interface can enable more flexibility in the algorithm side.

python/sglang/srt/dllm/config.py
```diff
@@ -38,6 +40,12 @@ def from_server_args(
                 f"Unknown diffusion LLM: {model_config.hf_config.architectures[0]}"
             )
 
+        max_running_requests = (
+            1
```
Monstertail, line 44: Shall we set it to a more reasonable default value rather than 1 in the future? As in speculative decoding, it is set to 48. https://github.com/sgl-project/sglang/blob/d0fb24ee7bf5d024d23dbb829d9abd7766489489/python/sglang/srt/server_args.py#L2146 


Even if we want to show a latency-sensitive case, maybe 2  would be better?

ClawSeven: The current dLLM implementation does not separate prefill and decode batching, meaning the PR is still in the initial version and only supports dynamic batching. Performance optimization will be added later. The default batch size will be set to 8 in the future.

python/sglang/srt/managers/schedule_batch.py
ClawSeven: I believe we will provide more reqs information for the DLLM algorithm, but I may not parse all of the reqs.

python/sglang/srt/dllm/algorithm/low_confidence.py
```diff
@@ -37,31 +55,50 @@ def run(
...
+                    curr_block_start:curr_block_end,
+                ]
+                block_mask_index = block_input_ids == self.mask_id
+                if torch.sum(block_mask_index).item() == 0:
+                    continue
+                curr_logits = logits_output.full_logits[
+                    curr_block_start:curr_block_end,
+                ]
+
+                x = torch.argmax(curr_logits, dim=-1)
```
Monstertail, line 72: In the future, we may need to replace the argmax to support sampling. https://github.com/sgl-project/sglang/issues/16615

ClawSeven: Yes, we have a PR for implementing more sampling algorithms.

python/sglang/srt/dllm/config.py
Monstertail: I see... That makes sense

python/sglang/srt/managers/schedule_batch.py
Monstertail: We may discuss this issue in the other algorithm support PR like fast dllm v2. Keep this for now.

python/sglang/srt/managers/utils.py
```diff
@@ -24,7 +24,7 @@
 class GenerationBatchResult:
     logits_output: Optional[LogitsProcessorOutput] = None
     pp_hidden_states_proxy_tensors: Optional[PPProxyTensors] = None
-    next_token_ids: Optional[torch.Tensor] = None
+    next_token_ids: Union[Optional[torch.Tensor], List[torch.Tensor]] = None
```
hnyls2002: I think the type hints are `Optional[Union[torch.Tensor, List[torch.Tensor]]]`?

hnyls2002 (changes requested):
Add a unit-test for that.

python/sglang/srt/managers/scheduler.py
```diff
@@ -1732,33 +1736,44 @@ def handle_batch_embedding_request(
         for tokenized_req in recv_req:
             self.handle_embedding_request(tokenized_req)
 
+    def _post_process_chunked_request(self, req: Req):
```
hnyls2002: Do not use ambigous name like "post process" or "pre process". I suggest modify the name to 

```suggestion
    def stash_chunked_request(self, req: Req):
```

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -1160,6 +1160,59 @@ def __repr__(self):
...
+            req.init_next_round_input()
+
+    def has_running_reqs(self) -> bool:
+        return self.dllm_config is not None and len(self.reqs) > 0
+
+    def update_chunked_status(self):
+        for req in self.reqs:
+            req.is_chunked += 1
+
+    def flush_finished_reqs(self):
```
hnyls2002: Why is `flush` here, I think `filter` would be better.

python/sglang/srt/managers/scheduler.py
```diff
@@ -1732,33 +1736,44 @@ def handle_batch_embedding_request(
...
 
         # Merge the prefill batch into the running batch
         chunked_req_to_exclude = set()
-        if self.chunked_req:
+
+        if self.dllm_config is not None and self.dllm_reqs.has_running_reqs():
+            chunked_req_to_exclude.update(self.dllm_reqs)
+            for req in self.dllm_reqs:
+                self._post_process_chunked_request(req)
+        elif self.chunked_req is not None:
```
hnyls2002: Do not assert here; assert before the above dllm requests process. Like

```python
if self.dllm_config is not None:
    assert self.chunked_req is None
```

That is to state the exclusivity first and keep processing logics clean.

python/sglang/srt/managers/scheduler.py
```diff
@@ -1846,20 +1861,21 @@ def _get_new_batch_prefill_raw(
...
             return None
 
         running_bs = len(self.running_batch.reqs)
         # Ignore the check if self.chunked_req is not None.
         # In the non-PP case, when self.chunked_req is not None, num_allocatable_reqs should always be greater than 0,
-        # as the space for the chunked request has just been released.
-        # In PP case, a chunked req can start in one microbatch and end in another microbatch, so the max_running_requests per microbatch should not be strict.
-        # Instead, we should always allow chunked request to be added, otherwise, there will be a memory leak.
+        # as the space for the chunked requests has just been released.
+        # In PP case, a chunked reqs can start in one microbatch and end in another microbatch, so the max_running_requests per microbatch should not be strict.
```
hnyls2002:
```suggestion
        # In PP case, chunked requests (or dllm requests) can start in one microbatch and end in another microbatch, so the max_running_requests per microbatch should not be strict.
```

python/sglang/srt/managers/scheduler.py
```diff
@@ -1897,10 +1913,16 @@ def _get_new_batch_prefill_raw(
             running_bs if self.is_mixed_chunk else 0,
             self.priority_scheduling_preemption_threshold,
             prefill_max_requests=self.server_args.prefill_max_requests,
+            dllm_config=self.dllm_config,
             prefill_delayer_single_pass=prefill_delayer_single_pass,
         )
 
-        if self.chunked_req is not None:
+        if self.dllm_config is not None and self.dllm_reqs.has_running_reqs():
```
hnyls2002: Also, add an assertion before the dllm processing and chunked requst processing. Then do not use `elif` to handle these branches.

python/sglang/srt/managers/scheduler.py
```diff
@@ -1979,14 +2003,19 @@ def _get_new_batch_prefill_raw(
             for req in adder.preempt_list:
                 self._add_request_to_queue(req)
 
-        # Update chunked prefill
-        if adder.new_chunked_req is not None:
+        if self.dllm_config is not None and adder.dllm_reqs.has_running_reqs():
+            self.dllm_reqs.add_reqs(adder.dllm_reqs)
+        elif adder.new_chunked_req is not None:
```
hnyls2002: also here. Just want to let every developers to know the two cases are mutually exclusive.

hnyls2002 (changes requested):

python/sglang/srt/managers/schedule_batch.py
ClawSeven: I believe that “flush” more accurately expresses the intended semantics than “filter.”
- “flush” conveys the idea of retaining the useful items while dropping the useless or dirty ones. (like [flush syscall](https://man7.org/linux/man-pages/man3/fflush.3.html)
- “filter,” by contrast, is broader and could be interpreted as either keeping or dropping the finished requests.

Therefore, I prefer to use the “flush” semantic in this context.

python/sglang/srt/managers/utils.py
ClawSeven: Done

python/sglang/srt/managers/scheduler.py
ClawSeven: Done

python/sglang/srt/managers/scheduler.py
ClawSeven: Done

python/sglang/srt/managers/scheduler.py
ClawSeven: Done

python/sglang/srt/managers/scheduler.py
ClawSeven: Done

python/sglang/srt/managers/scheduler.py
ClawSeven: Done

hnyls2002 (approved):

PR #14915, opened 2025-12-11
fix: config change after transformers upgrade results in dummy mode failure

mqhc2020:
## Motivation

To run dummy test for grok, hit an issue related to transformers revision. The root cause is 5.0.0rc0 changes the parameters in PretrainedConfig, which is used by dummy test to set up its rope scaling related parameters.

## Modifications

If "rope_parameters" is just a small dict, no real need to prepare rope scaling parameters for later rotary emb.

## Checklist


mqhc2020:
There is a PR #17784 that fixes this problem. Same purpose: to upgrade transformers.

PR #14920, opened 2025-12-11
Eagle: GPT-OSS Eagle v2 support

IzzyPutterman:
## Motivation

EAGLE v2/v3 models from GPT-OSS introduce an optional auxiliary hidden-state mechanism that improves speculative decoding quality by exposing additional per-layer representations. SGLang’s current EAGLE implementation assumes a single hidden-state stream and does not read any model-specific Eagle configuration, preventing correct execution for Eagle v2 models and causing mismatch with expected tensor shapes.

This PR adds full support for the Eagle v2/v3 aux-hidden pipeline while keeping existing EAGLE behavior unchanged for models that do not use it.

Related to [[Tracking] GPT-OSS B200/GB200 performance optimization tracker](https://github.com/sgl-project/sglang/issues/15243)

## Modifications

- Adds complete support for Eagle v2/v3’s optional auxiliary hidden-state design.
- Keeps default EAGLE execution unchanged for models that omit eagle_config.
- Moves all Eagle configuration into model-driven logic for correctness and maintainability.

## Checklist


python/sglang/srt/model_executor/model_runner.py
```diff
@@ -518,29 +545,10 @@ def initialize(self, min_per_gpu_memory: float):
...
-                is_draft_model=True,
-            )
-
-            try:
-                # get the aux layer from draft model config
-                eagle_config = getattr(
-                    draft_model_config.hf_config, "eagle_config", None
+            if self.eagle_use_aux_hidden_state:
+                self.model.set_eagle3_layers_to_capture(
+                    self.eagle_aux_hidden_state_layer_ids
```
IzzyPutterman: I think I have a typo here and I need to unindent these lines

IzzyPutterman: fixed, but might need to kill/rerun CI

python/sglang/test/speculative/test_eagle_config.py
```diff
@@ -0,0 +1,49 @@
+import unittest
```
Qiaolin-Yu: What are these tests for? Could you describe more? It seems it's not related to sglang.

python/sglang/srt/speculative/eagle_worker.py
```diff
@@ -186,6 +186,15 @@ def __init__(
         self.draft_tp_context = (
             draft_tp_context if server_args.enable_dp_attention else empty_context
         )
+        self.eagle_use_aux_hidden_state = False
```
Qiaolin-Yu, line 189: Do we also need to modify `eagle_worker_v2.py`?

Qiaolin-Yu (changes requested):
Could you provide the gpqa accuracy before and after this change? Also, for your provided test result, are you running on b200 or h200? The throughput number seems weird.

IzzyPutterman:
> Could you provide the gpqa accuracy before and after this change? Also, for your provided test result, are you running on b200 or h200? The throughput number seems weird.

Ahh sorry, this was H200, but I increased the concurrency for the second run.  The throughput is irrelevant mostly, just the acceptance rate/length being higher is what matters.

IzzyPutterman:
> Could you provide the gpqa accuracy before and after this change? Also, for your provided test result, are you running on b200 or h200? The throughput number seems weird.

Also this is just an eagle input change, so just the quality of the draft tokens change, but the output target should remain the same.  I can rerun in any case.

Qiaolin-Yu:
> > Could you provide the gpqa accuracy before and after this change? Also, for your provided test result, are you running on b200 or h200? The throughput number seems weird.
> 
> Also this is just an eagle input change, so just the quality of the draft tokens change, but the output target should remain the same. I can rerun in any case.

Yes. But I still think it's necessary.

python/sglang/srt/speculative/eagle_worker.py
IzzyPutterman: No need to modify eagle_worker_v2 it seems, I tried with setting the env var and False for disable overlap, and output was coherent and acceptance rates consistent.

python/sglang/test/speculative/test_eagle_config.py
IzzyPutterman: removed

Qiaolin-Yu (approved):

ishandhanani (approved):

PR #14925, opened 2025-12-11
extend timeout for b200 test

b8zhong:
Keeps on failing due to timeout (from trt allreduce fusion, flashinfer autotune), like

https://github.com/sgl-project/sglang/actions/runs/20119197436/job/57802226526?pr=14357
https://github.com/sgl-project/sglang/actions/runs/20118348681/job/57804217560?pr=13798

This env var increases to 50 min (already used elsewhere)

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -1214,7 +1214,7 @@ jobs:
         timeout-minutes: 30
         run: |
           cd test/srt
-          python3 run_suite.py --suite per-commit-4-gpu-b200 --auto-partition-id ${{ matrix.part }}  --auto-partition-size 3 --timeout-per-file 1800
+          IS_BLACKWELL=1 python3 run_suite.py --suite per-commit-4-gpu-b200 --auto-partition-id ${{ matrix.part }}  --auto-partition-size 3 --timeout-per-file 1800
 
   # TODO: Add gb200 tests back after the ci runner is fixed
   # unit-test-backend-4-gpu-gb200:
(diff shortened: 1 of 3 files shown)


python/sglang/srt/environ.py
```diff
@@ -151,6 +151,7 @@ class Envs:
     # Test & Debug
     SGLANG_IS_IN_CI = EnvBool(False)
     SGLANG_IS_IN_CI_AMD = EnvBool(False)
+    IS_BLACKWELL = EnvBool(False)
```
Kangyan-Zhou, line 154: Should we use SGLANG_IS_BLACKWELL to follow the naming convention?

.github/workflows/pr-test.yml
```diff
@@ -1214,7 +1214,7 @@ jobs:
         timeout-minutes: 30
         run: |
           cd test/srt
-          python3 run_suite.py --suite per-commit-4-gpu-b200 --auto-partition-id ${{ matrix.part }}  --auto-partition-size 3 --timeout-per-file 1800
+          IS_BLACKWELL=1 python3 run_suite.py --suite per-commit-4-gpu-b200 --auto-partition-id ${{ matrix.part }}  --auto-partition-size 3 --timeout-per-file 1800
```
zhyncs, line 1217: I think the real issue is that tvm-ffi is time-consuming during JIT generation.

b8zhong: @zhyncs I don't disagree. Btw, we also already installing `flashinfer-jit-cache`, so I'm unsure if there's anything we can do on SGL side.

PR #14933, opened 2025-12-12
[registry] Add a strict mode to model registration

yinghai:
## Motivation

The default behavior of model registration is to warn if import model failed. This is a bit inconvenient sometimes as downstream error can be obscure and the original import warning is swallowed by wall of logs. Adding a strict mode to fail fast. This PR does not change default behavior. 

## Checklist


python/sglang/srt/models/registry.py
```diff
@@ -19,8 +19,10 @@ class _ModelRegistry:
     # Keyed by model_arch
     models: Dict[str, Union[Type[nn.Module], str]] = field(default_factory=dict)
 
-    def register(self, package_name: str, overwrite: bool = False):
-        new_models = import_model_classes(package_name)
+    def register(
```
fzyzcjy, line 22: optional nit: shall we enable strict mode by `environ.py` adding a env variable

python/sglang/srt/models/registry.py
```diff
@@ -100,6 +102,8 @@ def import_model_classes(package_name: str):
             try:
                 module = importlib.import_module(name)
             except Exception as e:
+                if strict:
+                    raise e
```
fzyzcjy: tiny optional nit

```suggestion
                    raise
```

(iirc `raise e` will lose trace)

python/sglang/srt/models/registry.py
```diff
@@ -19,8 +19,10 @@ class _ModelRegistry:
     # Keyed by model_arch
     models: Dict[str, Union[Type[nn.Module], str]] = field(default_factory=dict)
 
-    def register(self, package_name: str, overwrite: bool = False):
-        new_models = import_model_classes(package_name)
+    def register(
+        self, package_name: str, overwrite: bool = False, strict: bool = False
+    ):
+        new_models = import_model_classes(package_name, strict)
```
fzyzcjy: optional tiny nit

```suggestion
        new_models = import_model_classes(package_name, strict=strict)
```

fzyzcjy (approved):

python/sglang/srt/models/registry.py
yinghai: Oh i actually don't know what sglang folks prefer :)

fzyzcjy: to be honest I also do not know :) I personally think making env var may be a little bit better b/c then people can use it w/o modifying register code

fzyzcjy:
btw ping me on slack if I do not reply on gihub - I get too many messages everywhere and can miss things :(

PR #14953, opened 2025-12-12
[model-gateway] Parallelize metrics requests

ppraneth:
<!DOCTYPE html>

## Motivation

The `fan_out_simple_request` function in the worker manager currently executes HTTP requests to workers sequentially inside a loop. This creates a "waterfall" effect where the total latency for endpoints like `/engine_metrics` is the sum of all individual worker latencies **(O(N)**).

If a single worker is slow or experiencing network latency, it blocks the entire metrics collection process, potentially causing timeouts in downstream monitoring systems (e.g., Prometheus) and creating gaps in observability data. This change aims to parallelize these requests to improve scalability and resilience.

## Modifications

  * Modified `fan_out_simple_request` in `sgl-model-gateway/src/core/worker_manager.rs`.
  * Refactored the request loop to collect `async` futures instead of awaiting them immediately.
  * Used `futures::future::join_all` to execute all worker requests concurrently.
* The total latency is now determined by the slowest single worker (max(L)) rather than the sum of all workers (sum(L)).

## Accuracy Tests

  * **Aggregation Correctness:** Validated that metrics from multiple workers are still correctly merged into the final response body.
  * **Fault Tolerance:** Verified that the parallel implementation correctly handles partial failures. If one worker returns an error (e.g., HTTP 500), the gateway successfully returns metrics from the remaining healthy workers without crashing.

## Benchmarking and Profiling

I created a reproduction test using 3 mock workers, each configured with a **2-second** artificial delay.

| Metric | Before Fix (Serial) | After Fix (Parallel) | Improvement |
| :--- | :--- | :--- | :--- |
| **Total Latency** | **6.01s** | **2.00s** | **3x Speedup** |
| **Behavior** | Waterfall (2s + 2s + 2s) | Concurrent (max(2s, 2s, 2s)) | Non-blocking |

**Test Output:**

```text
-> Starting 3 mock workers with 2000ms delay each...
-> Requesting engine metrics (calling Gateway)...
-> Metrics request completed in: 2.00 seconds
---------------------------------------------------
Expected Time (Serial)   : ~6.00 seconds (Sum of delays)
Expected Time (Parallel) : ~2.00 seconds (Max of delays)
Actual Time              : 2.00 seconds
---------------------------------------------------
RESULT: Confirmed PARALLEL execution (Issue Fixed)
```

## Checklist


sgl-model-gateway/src/core/worker_manager.rs
```diff
@@ -276,48 +276,62 @@ impl WorkerManager {
             return Err((StatusCode::SERVICE_UNAVAILABLE, "No available workers").into_response());
         }
 
-        let mut responses = vec![];
-        // May do parallel requests later
+        let mut tasks = Vec::new();
```
slin1237: since we know worker size
perhaps
`let mut tasks = Vec::with_capacity(num_backends);`

sgl-model-gateway/src/core/worker_manager.rs
slin1237, line 275:
```rust
use futures::stream::{FuturesUnordered, StreamExt};
use std::time::Duration;

const REQUEST_TIMEOUT: Duration = Duration::from_secs(5);
const MAX_CONCURRENT: usize = 32;

async fn fan_out_simple_request(
    worker_registry: &WorkerRegistry,
    client: &reqwest::Client,
    endpoint: &str,
    method: Method,
) -> Result<Vec<(String, String)>, Response> {
    let workers = worker_registry.get_all();
    if workers.is_empty() {
        return Err((StatusCode::SERVICE_UNAVAILABLE, "No available workers").into_response());
    }

    let futures: FuturesUnordered<_> = workers
        .into_iter()
        .map(|worker| {
            let client = client.clone();
            let worker_url = worker.url().to_string();
            let url = format!("{worker_url}/{endpoint}");
            let api_key = worker.api_key().clone();
            let method = method.clone();

            async move {
                let mut req = client.request(method, &url).timeout(REQUEST_TIMEOUT);

                if let Some(key) = api_key {
                    req = req.bearer_auth(key);
                }

                match req.send().await {
                    Ok(res) if res.status().is_success() => {
                        match res.text().await {
                            Ok(body) => Some((worker_url, body)),
                            Err(e) => {
                                warn!("Failed reading response from {url}: {e}");
                                None
                            }
                        }
                    }
                    Ok(res) => {
                        warn!("Request to {url} failed: {}", res.status());
                        None
                    }
                    Err(e) => {
                        warn!("Request to {url} failed: {e}");
                        None
                    }
                }
            }
        })
        .collect();

    let responses: Vec<_> = futures
        .buffer_unordered(MAX_CONCURRENT)
        .filter_map(|r| async { r })
        .collect()
        .await;

    if responses.is_empty() && !workers.is_empty() {
        return Err((
            StatusCode::BAD_GATEWAY,
            "All backend requests failed"
        ).into_response());
    }

    Ok(responses)
}
```
this might be better
1. Futures aren't spawned - These are just futures in a Vec, not actual tasks. They run concurrently via join_all but not in parallel (one slow future can block polling others)
2. Silent failures - Returns Ok(empty_vec) if all workers fail
3. Unnecessary clones - client.clone() is cheap (Arc), but method.clone() and the string allocations add up
4. StatusCode conversion - reqwest and http crate use the same type

sgl-model-gateway/tests/metrics_performance_test.rs
```diff
@@ -0,0 +1,95 @@
+use std::{sync::Arc, time::Instant};
```
slin1237: delete this test

slin1237 (changes requested):

sgl-model-gateway/tests/metrics_performance_test.rs
ppraneth: Done

sgl-model-gateway/src/core/worker_manager.rs
ppraneth: Thanks for the feedback! I've made the changes

sgl-model-gateway/src/core/worker_manager.rs
ppraneth: Thanks for the feedback! I've made the changes

ppraneth: Fixed

slin1237 (approved):

PR #14983, opened 2025-12-12
[CI] Enable retry logic for flaky CI tests

alisonshao:
## Summary
- Add `errors="ignore"` to `subprocess.Popen` in CI retry logic to handle non-UTF-8 bytes in test output

## Motivation
PR #14771 introduced retry logic that reads subprocess stdout to classify errors. When tests output binary data or non-standard encoded characters, this causes:

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x9e in position 6489: invalid start byte
```

This fix ignores invalid bytes instead of crashing, since the error classification only needs to match ASCII patterns like `AssertionError`, `RuntimeError`, etc.

## Test plan
- CI should no longer crash with UnicodeDecodeError when tests output non-UTF-8 bytes


alisonshao:
Verified locally that this fix works.

Test setup - created a script that outputs the same invalid UTF-8 byte (0x9e) that caused the CI failure:

```python
import sys
print("Starting test...")
sys.stdout.buffer.write(b"Binary data: \x9e\x9f\xa0 here\n")  # Invalid UTF-8 bytes
print("Test complete")
sys.exit(1)
```

Without `errors='ignore'`:
```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x9e in position 13: invalid start byte
```
Crashes immediately when iterating over `process.stdout`, same as the CI failure.

With `errors='ignore'`:
```
Got line: Binary data:  here
Got line: Starting test...
Got line: Test complete
Result: Success! Exit code: 1
```
Reads the output fine. Invalid bytes get dropped but all the ASCII text we need for error pattern matching (`AssertionError`, `RuntimeError`, etc.) is preserved.

This is safe because the retry logic only matches ASCII error patterns anyway - the non-UTF-8 bytes are just binary noise from GPU libraries or corrupted output that we don't need.

.github/workflows/pr-test.yml
```diff
@@ -1215,12 +1215,23 @@ jobs:
...
 
       - name: Evaluate accuracy
-        timeout-minutes: 20
+        timeout-minutes: 25
         run: |
           cd test/srt
-          python3 test_eval_accuracy_large.py
-        env:
-          SGLANG_CI_ENABLE_RETRY: ${{ needs.check-changes.outputs.enable_retry }}
+          # Retry logic for flaky accuracy tests
```
Kangyan-Zhou: Can we not have bash logic in the pr-test.yml file?

PR #14994, opened 2025-12-12
Introduce session cache

zhangrui1203:
## Motivation

SGLang provides session-level KV cache management, enabling user-defined business logic to explicitly control advanced features such as the lifecycle of the KV cache, hierarchical caching, and replica count management. In large-scale LLM service scenarios within data centers, SGLang serves as a high-performance data processing node, better supporting business logic and achieving superior results in performance, cost efficiency, and user experience.

## Modifications

- implement session cache framework
- improve hicache local file to support session cache
- support GD2FS as session cache storage backend
- improve performance for get_dummy_flat_data_page
- bug fix for cache_finished_req

## Checklist

### DIFF
--- a/benchmark/hicache/bench_multiturn.py
+++ b/benchmark/hicache/bench_multiturn.py
@@ -1,12 +1,14 @@
 import argparse
 import asyncio
 import json
+import os
 import queue
 import random
 import threading
 import time
 from datetime import datetime
-from typing import Optional
+from pathlib import Path
+from typing import Any, Dict, List, Optional
 
 import aiohttp
 import numpy as np
@@ -140,6 +142,11 @@ def parse_args():
         default="",
         help="String of LoRA path. Currently we only support benchmarking on a single LoRA adaptor.",
     )
+    parser.add_argument(
+        "--enable-session-cache",
+        action="store_true",
+        help="If set, enable session cache.",
+    )
     return parser.parse_args()
 
 
@@ -164,6 +171,8 @@ async def async_request_sglang_generate(
                 if response.status == 200:
                     prompt_tokens = 0
                     cached_tokens = 0
+                    fresh_kv_cache = None
+
                     async for chunk_bytes in response.content:
                         chunk_bytes = chunk_bytes.strip()
                         if not chunk_bytes:
@@ -195,6 +204,9 @@ async def async_request_sglang_generate(
 
                                 most_recent_timestamp = timestamp
                                 generated_text = data["text"]
+                                fresh_kv_cache = data.get("session_params", {}).get(
+                                    "fresh_kv_cache", None
+                                )
 
                     output.generated_text = generated_text
                     output.success = True
@@ -202,6 +214,7 @@ async def async_request_sglang_generate(
                     output.prompt_len = prompt_tokens
                     output.cached_tokens = cached_tokens
                     output.generated_len = len(output.itl) + 1
+                    output.fresh_kv_cache = fresh_kv_cache
                 else:
                     output.error = response.reason or ""
                     output.success = False
@@ -215,7 +228,7 @@ async def async_request_sglang_generate(
     return output
 
 
-def gen_payload(prompt, output_len, lora_path=""):
+def gen_payload(prompt, output_len, lora_path="", session_params=None):
     payload = {
         "text": prompt,
         "sampling_params": {
@@ -228,6 +241,7 @@ def gen_payload(prompt, output_len, lora_path=""):
         "lora_path": lora_path,
         "return_logprob": False,
         "logprob_start_len": -1,
+        "session_params": session_params or {},
     }
     return payload
 
@@ -312,11 +326,17 @@ class WorkloadGenerator:
             random_sample=not args.disable_random_sample,
         )
 
+        self.enable_session_cache = args.enable_session_cache
+        self.prepare_session_cache(args.num_clients)
+
         init_requests = [
             (
                 i,
                 gen_payload(
-                    self.candidate_inputs[i], args.output_length, args.lora_path
+                    self.candidate_inputs[i],
+                    args.output_length,
+                    args.lora_path,
+                    session_params=self.gen_session_params(i),
                 ),
             )
             for i in range(args.num_clients)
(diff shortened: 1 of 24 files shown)


python/sglang/srt/managers/io_struct.py
```diff
@@ -143,6 +144,21 @@ class SessionParams:
     offset: Optional[int] = None
     replace: Optional[bool] = None
     drop_previous_output: Optional[bool] = None
+    # The session cache segments.
+    # Each segment is a dict with 'token_start', 'token_length',
+    # 'kv_uri', 'kv_start', and 'kv_length' keys.
+    old_kv_cache: Optional[List[Dict]] = None
+    new_kv_cache: Optional[List[Dict]] = None
```
xiezhq-hermann: I suggest to use a different pair of names here, making `old_kv_cache` to be `stored_kv_cache`. We can keep `new_kv_cache`, or use `fresh_kv_cache`. The current old vs new is a bit confusing.

python/sglang/srt/managers/io_struct.py
```diff
@@ -143,6 +144,21 @@ class SessionParams:
...
+
+    def __post_init__(self):
+        old = SessionCache(self.old_kv_cache)
+        new = SessionCache(self.new_kv_cache)
+        if len(old) > 0 and len(new) > 0 and old.token_end != new.token_start:
```
xiezhq-hermann, line 156: this is a very strong assumption of execution, do you have following plans on fault tolerance?

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -1003,7 +1009,7 @@ def _check_str_based_finish(self):
             if len(self.sampling_params.stop_regex_strs) > 0:
                 for stop_regex_str in self.sampling_params.stop_regex_strs:
                     if re.search(stop_regex_str, tail_str):
-                        self.finished_reason = FINISHED_MATCHED_REGEX(
+                        self.finished_reason = FINISH_MATCHED_REGEX(
```
xiezhq-hermann: let's have all sorts of style fix in a separate PR

python/sglang/srt/managers/scheduler.py
```diff
@@ -1968,6 +1999,13 @@ def get_new_batch_prefill(self) -> Optional[ScheduleBatch]:
                 ):
                     break
 
+            if self.enable_session_cache:
+                session_id = req.session_id
+                if session_id is not None:
+                    done = self.tree_cache.check_session_prefetch_progress(session_id)
+                    if not done:
+                        continue
```
xiezhq-hermann, line 2052: so here is a blocking call for prefetching for the session cache, will there be more policies for it?

python/sglang/srt/mem_cache/memory_pool.py
```diff
@@ -809,6 +816,31 @@ def get_cpu_copy(self, indices):
...
+            )
+
+        kv_host = torch.zeros(
+            (2, self.layer_num, len(indices), self.head_num, self.head_dim),
+            dtype=self.store_dtype,
+            device="cpu",
+            pin_memory=True,
+        )
+
+        torch.cuda.synchronize()
```
xiezhq-hermann, line 880: probably better use stream sync for lower overhead

python/sglang/srt/mem_cache/session_cache.py
```diff
@@ -0,0 +1,615 @@
...
+
+            flat_data = mem_pool_host.get_dummy_flat_data_page(page_num)
+
+            storage, filepath = SessionCacheStorageManager.get_storage(
+                seg.kv_uri, storage_config, mem_pool_host
+            )
+
+            storage.load(filepath, seg.kv_start, flat_data)
+
+            mem_pool_host.set_from_flat_data(mem_indices, flat_data)
```
xiezhq-hermann, line 509: will there be a zero-copy mechanism for loading data from the storage?

python/sglang/srt/mem_cache/session_cache.py
```diff
@@ -0,0 +1,615 @@
...
+
+        kv_cache = self._shift_real_token()
+        for seg in kv_cache:
+            seg_data = flat_data[:, :, seg.token_start : seg.token_end, :, :]
+
+            storage, filepath = SessionCacheStorageManager.get_storage(
+                seg.kv_uri, storage_config, mem_pool_host
+            )
+
+            storage.save(filepath, seg.kv_start, seg_data)
```
xiezhq-hermann, line 543: same comments for zero-copy

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -447,7 +453,7 @@ def cache_finished_req(self, req: Req, is_insert: bool = True):
         ]
 
         # Maybe convert to bigram keys for EAGLE
-        keys = convert_to_bigram_key(req.fill_ids) if self.is_eagle else req.fill_ids
+        keys = convert_to_bigram_key(token_ids) if self.is_eagle else token_ids
```
xiezhq-hermann, line 463: not sure about this one, if it's more of a style change please make it in a separate PR, otherwise will need @ispobock's opinion on it

xiezhq-hermann:
Thank you @zhangrui1203 @pizhenwei for the PR, overall a very neat PR and improved a lot since my initial prototype: https://github.com/sgl-project/sglang/tree/xiezhq-session. I left some comments and have a few addition questions:
1. I do think the session cache abstraction can be compatible with other storage backends with only minor interface changes, e.g., providing a file interface to load and save session specific KV caches. It's ok for now to only support your backend but please help think about it in future refactoring.
2. Zero-copy from storage to memory is still missing, leaving fairly substantial performance on the table.
3. Can you prepare a test case just like other storage backends do, that way we have a correctness guarantee. It would be better to have a simple performance benchmark too, we can adapt the existing multi-turn benchmark script for it a bit. A validation on the session abstraction achieving close or even better performance over existing pages would be great.

python/sglang/srt/managers/io_struct.py
zhangrui1203: OK

python/sglang/srt/managers/io_struct.py
zhangrui1203: If users intend to utilize the session cache (by specifying session_id along with stored_kv_cache and fresh_kv_cache), they must strictly adhere to the following requirements:

1. The internal segments of both stored_kv_cache and fresh_kv_cache must be contiguous in terms of sequence positions.
2. The fresh_kv_cache must be directly appendable to the tail of the stored_kv_cache, maintaining sequential continuity.

If these conditions are not satisfied, SGLang must return an error. This is because the session cache relies on metadata that is managed and persisted by the control plane (i.e., the user). As a data-plane component, SGLang cannot interpret or recover from malformed or inconsistent control-plane metadata.

python/sglang/srt/managers/schedule_batch.py
zhangrui1203: OK

python/sglang/srt/managers/scheduler.py
zhangrui1203: Currently, a generate request will block indefinitely until the session cache prefetch operation completes. In the future, timeout support will be added to prevent the generate request from hanging due to excessively long prefetch operations; however, this capability depends on the implementation of the backend storage system.

python/sglang/srt/mem_cache/memory_pool.py
zhangrui1203: OK

python/sglang/srt/mem_cache/radix_cache.py
zhangrui1203: We noticed that `req.fill_ids` here only includes `req.origin_input_ids` and does **not** include `req.output_ids`, which causes the cached KV cache to be missing the KV entries corresponding to `req.output_ids`.

zhangrui1203:
Hi, @xiezhq-hermann 
The main changes in the new version are as follows:

- rename `old_kv_cache`/`new_kv_cache` to `stored_kv_cache`/`fresh_kv_cache`
- remove style fix
- use `torch.cuda.stream()` instead of `torch.cuda.synchronize()` in `get_flat_data`
- rebase to the latest commit

The test case is currently being added.

zhangrui1203:
@xiezhq-hermann 
One more thing: While adding accuracy tests, we noticed they depend on @sgl.function, which would require API changes. Are there simpler or more standard ways to test this?

benchmark/hicache/bench_multiturn.py
```diff
@@ -356,6 +376,58 @@ def __init__(self, args):
...
+        if not self.enable_session_cache:
+            return
+
+        self.sessions: Dict[int, Any] = {
+            i: {"id": f"session_{i}"} for i in range(num_clients)
+        }
+
+        # Create session cache files
+        hicache_file_path = os.getenv(
+            "SGLANG_HICACHE_FILE_BACKEND_STORAGE_DIR", "/tmp/hicache"
```
alphabetc1: nit: please put all environment variables in `environ.py` and use `envs.VAR_NAME.get()`

zhangrui1203:
Hi, @xiezhq-hermann 
All CI runs have completed. The failures appear unrelated to this PR, and some may pass with a retry.

xiezhq-hermann:
@zhangrui1203 thank you for all your effort. How has the accuracy testing been going? My understanding is that as long as we’re using the same session, the session cache should be in effect.

Regarding the merge, I’d suggest splitting it into multiple independent PRs. That should make the code review easier and keep each commit atomic.

PR #14997, opened 2025-12-12
support non disturbing remote instance weight loader v2

amysaq2023:
## Motivation

In https://github.com/sgl-project/sglang/pull/8215, SGLang has already supported a new load format: remote_instance, that allows new instance to load weights from another running instance. This approach can greatly improve weight loading time during instance initialization. However, since it use torch.distributed with NCCL as backend, it will disturb on-going inference requests: torch.distributed will always launch CUDA kernels for transferring weight tensors.

We come up with another backend option: TransferEngine, which will not disturbing any GPU workload and still, using RDMA to transfer weights.

## Modifications

We initialize one TransferEngine for each ModelRunner and will register its weights to RDMA channel during initialization.

When initializing a new instance who wants to use `remote_instance` load_format with TransferEngine backend:
1. It  will send an HTTP request to retrieve the source instance's TransferEngine metadata, including RDMA keys mapped to the corresponding GPU memory addresses.
2. Using these RDMA keys, the new instance directly loads weights from the source's GPU memory.

How to use:
seed instance:
`python -m sglang.launch_server [args] \`
`  --remote-instance-weight-loader-support-transfer-engine`

client instance:
`python -m sglang.launch_server [args] \`
`  --load-format remote_instance \`
`  --remote-instance-weight-loader-seed-instance-ip [seed_instance_ip] \`
`  --remote-instance-weight-loader-seed-instance-service-port [seed_instance_service_port] \`
`  --remote-instance-weight-loader-backend "transfer_engine"`
`  --remote-instance-weight-loader-support-transfer-engine`

## Checklist

### DIFF
--- a/python/sglang/srt/entrypoints/http_server.py
+++ b/python/sglang/srt/entrypoints/http_server.py
@@ -1615,10 +1615,11 @@ def launch_server(
     1. The HTTP server, Engine, and TokenizerManager all run in the main process.
     2. Inter-process communication is done through IPC (each process uses a different port) via the ZMQ library.
     """
-    tokenizer_manager, template_manager, scheduler_info, port_args = (
+    tokenizer_manager, template_manager, scheduler_infos, port_args = (
         launch_subprocesses_func(server_args=server_args)
     )
 
+    scheduler_info = scheduler_infos[0]
     set_global_state(
         _GlobalState(
             tokenizer_manager=tokenizer_manager,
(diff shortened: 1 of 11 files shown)


python/sglang/srt/entrypoints/http_server.py
```diff
@@ -152,6 +155,15 @@ class _GlobalState:
...
     scheduler_info: Dict
+    # Dict{
+    #   rank: Tuple(
+    #           session_id,
+    #           Dict{
+    #               name: Tuple (d_ptr, numel, element_size)
+    #           }
+    #         )
+    # }
+    remote_instance_transfer_engine_info: Optional[Dict] = None
```
merrymercy, line 166: Can you drop this?

We can parse from `scheduler_info` directly.

merrymercy (approved):

python/sglang/srt/entrypoints/http_server.py
amysaq2023: For now, we only change the return value of launch_subprocesses from scheduler_info to the list of scheduler_info. But when it is stored to global_state, it only keep the first scheduler info item as before.

Would you suggest it will be better to store the whole list of scheduler_info in global_state as well? I'm a little concern about whether it will be too much redundant scheduler info in global state.

Issue #13823, opened 2025-11-24
diffusion: communication kernel replacement


mickqian:
communication ops in [group_coordinator.py](https://github.com/sgl-project/sglang/blob/main/python/sglang/multimodal_gen/runtime/distributed/group_coordinator.py) should be replaced by custom ones

PR #15073, opened 2025-12-13
[Feat/diffusion] comm kernel replacement

Ther-LF:
#### 📝 Description
This PR addresses issue #13823 by refactoring the communication operations in `GroupCoordinator`.

Previously, the coordination logic relied on `torch.distributed` primitives. This PR replaces `all_gather`, `all_reduce`, and `gather` operations with their counterparts in `self.device_communicator`. This change aligns the diffusion pipeline with the SGLang runtime architecture and improves performance in specific stages.

**Key Changes:**
1.  **Refactor `group_coordinator.py`**:
    - Switched from `torch.distributed` ops to `DeviceCommunicator` kernels.
    - Ensured compatibility with existing tensor parallel logic.
2.  **Add Tests**:
    - Added `python/sglang/multimodal_gen/test/distributed/test_comm_ops.py`.
    - This script verifies the correctness of the new communication kernels and provides a basic benchmark suite.

#### 📊 Benchmark Results

I evaluated the performance using **FLUX.1-dev** with **TP=4**. The results demonstrate a **7.1% reduction in end-to-end latency**, with the most significant improvement observed in the `TextEncodingStage`.

**Reproduction Command:**
```bash
sglang generate \
  --model-path black-forest-labs/FLUX.1-dev \
  --prompt "A beautiful landscape" \
  --tp-size 4 \
  --perf-dump-path flux_tp4_replace_comm.json
```

**Performance Comparison Report:**

| Metric | Baseline | New | Diff | Status |
| :--- | :--- | :--- | :--- | :--- |
| **E2E Latency** | 26308.28 ms | 24452.11 ms | **-1856.17 ms (-7.1%)** | ✅ |
| **Throughput** | 0.04 req/s | 0.04 req/s | - | - |

**Stage Breakdown:**

| Stage Name | Baseline (ms) | New (ms) | Diff (ms) | Diff (%) | Status |
| :--- | :--- | :--- | :--- | :--- | :--- |
| InputValidationStage | 0.03 | 0.06 | +0.03 | +95.5% | ⚪️ |
| **TextEncodingStage** | **2660.89** | **871.64** | **-1789.24** | **-67.2%** | 🟢 |
| ConditioningStage | 0.02 | 0.02 | -0.00 | -1.7% | ⚪️ |
| TimestepPreparationStage | 19.35 | 14.61 | -4.73 | -24.5% | ⚪️ |
| LatentPreparationStage | 0.20 | 0.22 | +0.02 | +12.0% | ⚪️ |
| DenoisingStage | 22873.69 | 22821.18 | -52.52 | -0.2% | ⚪️ |
| DecodingStage | 692.90 | 689.62 | -3.29 | -0.5% | ⚪️ |

> **Note:** The `TextEncodingStage` shows a significant speedup (-67.2%), likely due to reduced overhead in the custom communication kernels for the specific tensor shapes used during encoding. The main `DenoisingStage` performance remains stable.

#### ✅ Verification
- [x] Passed local correctness tests (`test_comm_ops.py`).
- [x] Verified generation quality visually (output images are consistent with baseline).


Ther-LF:
### Investigation on `separate_tensors` Correctness

I conducted a deeper investigation into the `all_gather` implementation. While my current PR successfully maintains **parity** with the original PyTorch fallback implementation (meaning `res_custom == res_torch`), I discovered that **the original logic itself is fundamentally incorrect** when `dim != 0` and `separate_tensors=True`.

#### The Issue
The existing logic uses a "flatten and slice" approach:
```python
# Current Logic
tensor_list = [
    output_tensor.reshape(-1)
    .narrow(0, input_.numel() * i, input_.numel())
    .view_as(input_)
    for i in range(world_size)
]
```
This assumes that the data for each rank is physically contiguous in memory. However, when gathering along `dim != 0` (e.g., `dim=1`), the data becomes **interleaved** in row-major memory layout, breaking this assumption.

#### Concrete Example
Consider `World Size = 2`, `Input Shape = [2, 2]`.
- **Rank 0 Data**: `[[A, A], [A, A]]`
- **Rank 1 Data**: `[[B, B], [B, B]]`

**1. All Gather along `dim=1`:**
Resulting Tensor (2x4):
```
[[A, A, B, B],
 [A, A, B, B]]
```

**2. Current Logic (`reshape(-1)`):**
Flattening proceeds row by row:
`[A, A, B, B, A, A, B, B]`

**3. Current Logic (`narrow` & `view` for Rank 0):**
Takes the first 4 elements: `[A, A, B, B]` -> Reshapes to `[2, 2]`:
```
[[A, A],
 [B, B]]  <-- ❌ WRONG! Mixed data from Rank 1.
```
Rank 0 should have received `[[A, A], [A, A]]`.

#### The Fix
To address this, I will update **both** the new `DeviceCommunicator` path and the legacy `torch.distributed` fallback path to use `torch.chunk`, which correctly handles splitting along dimensions:

```python
# Correct Logic
return list(torch.chunk(output_tensor, world_size, dim=dim))
```

I will push a commit shortly to fix this logic error in both implementations and update the tests to cover `dim=1` cases to ensure absolute correctness.

python/sglang/multimodal_gen/runtime/distributed/group_coordinator.py
```diff
@@ -333,10 +329,14 @@ def all_reduce(
         # Bypass the function if we are using only 1 GPU.
         if self.world_size == 1:
             return input_
-        else:
-            torch.distributed.all_reduce(
-                input_, op=op, group=self.device_group, async_op=async_op
-            )
+
+        if self.device_communicator is not None and self.use_device_communicator:
```
BBuf: Should we add a `async_op is not True` check or pass `async_op` args ?

python/sglang/multimodal_gen/test/distributed/test_comm_ops.py
```diff
@@ -0,0 +1,237 @@
...
+    time_custom = benchmark_op(coord_custom.gather, (t_in, dst_rank, dim_to_test))
+    time_torch = benchmark_op(coord_torch.gather, (t_in, dst_rank, dim_to_test))
+
+    if rank == 0:
+        print(
+            f"[Gather dim={dim_to_test}      | {label:<6}] Custom: {time_custom:.4f} ms | Torch: {time_torch:.4f} ms | Speedup: {time_torch/time_custom:.2f}x"
+        )
+
+
+def main():
```
BBuf: We can rewrite this in pytest style and skip when there's `no CUDA, world_size < 2, or no NCCL`. Also, should we add this test to CI?

python/sglang/multimodal_gen/runtime/distributed/group_coordinator.py
Ther-LF: Thanks for the suggestion. I have updated the logic to fallback to the PyTorch implementation when `async_op` is set to `True`.

python/sglang/multimodal_gen/test/distributed/test_comm_ops.py
Ther-LF: I have refactored the script into unittest style as suggested. I added unittest.SkipTest logic for the following conditions:

No CUDA available.

NCCL backend not available.

World size < 2.

Regarding CI: If added, since most standard CI environments run as a single process or lack GPUs, this test will automatically result in a SKIPPED status (instead of failing or crashing). This ensures it won't break the build pipeline, while still remaining runnable locally via torchrun for benchmarking.

Ther-LF: I noticed that the CI workflow (e.g., multimodal-gen-test-2-gpu) indeed supports multi-GPU testing. I agree that adding this test to the CI is valuable to ensure the correctness of the custom communication kernels remains consistent with the PyTorch fallback, especially to prevent regression in future optimizations.

PR #15075, opened 2025-12-13
[CPU] support LayerNorm with 3D shape

ZailiWang:
## Modifications

Qwen3-Omni modeling applies 3D [batch_size, seq_len, hidden_size] shape for LayerNorm inputs. So update the CPU LayerNorm kernel in #14074 to support both 2D and 3D inputs.

Also changed to returning the output in a new tensor instead of inplace updates.

bias is also supported per Qwen3-omni requirement.

## Accuracy Tests

UT in `test_norm.py` has been updated accordingly.

## Checklist


sgl-kernel/csrc/cpu/norm.cpp
```diff
@@ -304,24 +304,30 @@ void fused_rmsnorm_gated_kernel_impl(
...
+    scalar_t* __restrict__ output,
+    const scalar_t* __restrict__ input,
     scalar_t* __restrict__ residual,
     const scalar_t* __restrict__ weight,
     float* __restrict__ buffer,
     int64_t batch_size,
+    int64_t seq_len,
     int64_t hidden_size,
     int64_t input_strideN,
     float eps = 1e-5) {
```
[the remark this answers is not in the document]
ZailiWang: `fused_add_layernorm_kernel_impl` is not a user-facing function, so such error is unlikely to be triggered. Anyway it would be good to add this check for code clarity.

sgl-kernel/csrc/cpu/norm.cpp
```diff
@@ -555,36 +582,54 @@ void fused_add_rmsnorm_cpu(at::Tensor& input, at::Tensor& residual, at::Tensor&
...
+  } else {
+    CHECK_EQ(input.size(1), weight.size(0));
+  }
+
+  int64_t batch_size{input.size(0)}, seq_len{0}, hidden_size{input.size(1)}, input_strideN{input.stride(0)};
+  if (inp_dim == 3) {
+    seq_len = input.size(1);
+    hidden_size = input.size(2);
+    input_strideN = input.stride(1);
+  }
```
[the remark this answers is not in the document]
ZailiWang: The checking is in the respective functions for better readability.

sgl-kernel/csrc/cpu/torch_extension_cpu.cpp
```diff
@@ -330,15 +330,15 @@ TORCH_LIBRARY_FRAGMENT(sgl_kernel, m) {
...
+  m.def("layernorm_cpu(Tensor input, Tensor weight, float eps) -> Tensor");
   m.impl("layernorm_cpu", torch::kCPU, &layernorm_cpu);
   m.def("l2norm_cpu(Tensor input, float eps) -> Tensor");
   m.impl("l2norm_cpu", torch::kCPU, &l2norm_cpu);
   m.def("fused_rmsnorm_gated_cpu(Tensor input, Tensor weight, Tensor gate, float eps) -> Tensor");
   m.impl("fused_rmsnorm_gated_cpu", torch::kCPU, &fused_rmsnorm_gated_cpu);
   m.def("fused_add_rmsnorm_cpu(Tensor(a!) input, Tensor residual, Tensor weight, float eps) -> ()");
   m.impl("fused_add_rmsnorm_cpu", torch::kCPU, &fused_add_rmsnorm_cpu);
-  m.def("fused_add_layernorm_cpu(Tensor(a!) input, Tensor residual, Tensor weight, float eps) -> ()");
+  m.def("fused_add_layernorm_cpu(Tensor input, Tensor residual, Tensor weight, float eps) -> Tensor");
```
mingfeima: use outplace tensor is actually better, make graph runner work easier. @CaoE please comment!

sgl-kernel/csrc/cpu/norm.cpp
```diff
@@ -304,24 +304,33 @@ void fused_rmsnorm_gated_kernel_impl(
...
     int64_t batch_size,
+    int64_t seq_len,
     int64_t hidden_size,
     int64_t input_strideN,
     float eps = 1e-5) {
   using bVec = at::vec::Vectorized<scalar_t>;
   using fVec = at::vec::Vectorized<float>;
+  TORCH_CHECK(
+      !((residual != nullptr) && (buffer == nullptr)),
+      "fused_add_layernorm_kernel_impl with a residual input requires a buffer.");
```
mingfeima: usually we don't do checks like this:

* check at the entry function. residual can be set to be optional, and check whether it has_value() for your scenarios.
* skip buffer check. it is defined by yourself.

sgl-kernel/csrc/cpu/norm.cpp
```diff
@@ -304,24 +304,33 @@ void fused_rmsnorm_gated_kernel_impl(
...
   constexpr int kVecSize = bVec::size();
-  at::parallel_for(0, batch_size, 0, [&](int64_t begin, int64_t end) {
-    int tid = at::get_thread_num();
-    float* __restrict__ buffer_ptr = buffer + tid * hidden_size;
+  int64_t parallel_size{(seq_len == 0) ? batch_size : batch_size * seq_len};
+  at::parallel_for(0, parallel_size, 0, [&](int64_t begin, int64_t end) {
+    float* __restrict__ buffer_ptr{(float*)nullptr};
+    if (buffer != nullptr) {
+      buffer_ptr = buffer + at::get_thread_num() * hidden_size;
+    }
```
mingfeima: allocate buffer at the entry function, not here.

sgl-kernel/csrc/cpu/norm.cpp
```diff
@@ -304,24 +304,33 @@ void fused_rmsnorm_gated_kernel_impl(
...
 
     for (int64_t i = begin; i < end; ++i) {
-      scalar_t* __restrict__ input_ptr = input + i * input_strideN;
+      scalar_t* __restrict__ out_ptr = output + i * hidden_size;
+      const scalar_t* __restrict__ input_ptr = input + i * input_strideN;
       scalar_t* __restrict__ residual_ptr{(scalar_t*)nullptr};
       if (residual != nullptr) {
         residual_ptr = residual + i * hidden_size;
```
mingfeima: usually i would prefer to use a template parameter 

<template typename has_residual>

something like that,

and use `AT_DISPATCH_BOOL` to skip the bool checks

sgl-kernel/csrc/cpu/norm.cpp
```diff
@@ -456,33 +477,42 @@ at::Tensor rmsnorm_cpu(at::Tensor& input, at::Tensor& weight, double eps) {
...
+  int64_t inp_dim{input.dim()};
+  TORCH_CHECK(inp_dim == 2 || inp_dim == 3, "Expected input dim to be 2 or 3, but got ", inp_dim);
   CHECK_DIM(1, weight);
-  CHECK_EQ(input.size(1), weight.size(0));
-  int64_t batch_size = input.size(0);
-  int64_t hidden_size = input.size(1);
-  int64_t input_strideN = input.stride(0);
 
-  int64_t num_threads = at::get_num_threads();
-  at::Tensor buffer = at::empty({num_threads, hidden_size}, input.options().dtype(at::kFloat));
```
mingfeima, line 561: a simple way to treat this is **make 2d cases as a special 3d case** and use the same set of kernels.

mingfeima: seqlen = 1 for 2d cases.

mingfeima (changes requested):

sgl-kernel/csrc/cpu/norm.cpp
ZailiWang: updated. As offline discussed, `buffer` is utilized for both has_residual and non_residual cases, so this checking is removed.

sgl-kernel/csrc/cpu/norm.cpp
ZailiWang: updated per comment

sgl-kernel/csrc/cpu/norm.cpp
ZailiWang: I studied the [example usage of `AT_DISPATCH_BOOL`](https://github.com/sgl-project/sglang/blob/702426b06ac5675f1c2362b3a8153700c8e05c96/sgl-kernel/csrc/cpu/gemm_fp8.cpp#L387). However, the usage intends to templatize the `has_xxx` conditions and pass this condition into a deeper kernel (like [here](https://github.com/sgl-project/sglang/blob/702426b06ac5675f1c2362b3a8153700c8e05c96/sgl-kernel/csrc/cpu/gemm_fp8.cpp#L404). 
My case does not have a deeper kernel, so seems `AT_DISPATCH_BOOL` should not applied here. In fact, when I set the core process (containing `#pragma GCC unroll 4`) in the function to be the lamdba function as the 3rd arg of `AT_DISPATCH_BOOL`, seems it's violating the syntax requirement. The compilation error is raised:
```
      /sgl-workspace/sglang/sgl-kernel/csrc/cpu/norm.cpp:409:9: error: ‘#pragma’ is not allowed here
        409 | #pragma GCC unroll 4
            |         ^~~
```

So I simply define a `const bool has_residual` at the beginning and use `has_residual` for the following `if` condition judgements.

ZailiWang:
The current kernel impl has a known bug when the input tensor has a `.transpose()` op in prior, which is met in Qwen3-VL. But except that, the kernel is verified and can be merged so that it can be utilized by other models.

python/sglang/srt/layers/layernorm.py
```diff
@@ -413,9 +413,14 @@ def forward_cpu(
...
-                x, self.weight.data, self.variance_epsilon
-            )
+            if self.use_bias:
+                return torch.ops.sgl_kernel.layernorm_cpu(
+                    x, self.weight.data, self.bias.data, self.variance_epsilon
+                )
+            else:
+                return torch.ops.sgl_kernel.layernorm_cpu(
+                    x, self.weight.data, None, self.variance_epsilon
+                )
```
mingfeima:
```suggestion
            bias_data = self.bias.data if self.use_bias else None
            return torch.ops.sgl_kernel.layernorm_cpu(
                x, self.weight.data, bias_data, self.variance_epsilon
)
```

mingfeima (approved):
LGTM. Just some minor changes needed.

PR #15086, opened 2025-12-13
[NSA] Fix NSA backend assertion error when running DeepSeek-V3.2 PP with radix-cache

YAMY1234:
## Motivation

Fix NSA backend assertion error when running DeepSeek-V3.2 triggered when running DeepSeek-V3.2 PP with radix-cache.

The assertion `num_tokens <= total_num_tokens` in `dequant_k_cache.py` assumes that logical token count cannot exceed physical KV cache size. However, prefix sharing allows multiple sequences to share the same KV cache slots, making `seq_lens_sum` larger than `kv_cache_size`.

**Reproduction:**
```
    --model-path deepseek-ai/DeepSeek-V3.2-Exp \
    --tp-size 2 \
    --pp-size 4 \
    --max-running-requests 128 \
    --max-total-tokens 100000 \
    --chunked-prefill-size 8192 \
    --mem-fraction-static 0.80
```
```
python3 benchmark/gsm8k/bench_sglang.py --num-shots 20 --num-questions 1319 --parallel 1319
```

**Error:**:
```
  File "/sgl-workspace/sglang/python/sglang/srt/layers/attention/nsa_backend.py", line 983, in forward_extend
    kv_cache = dequantize_k_cache_paged(
               ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/sgl-workspace/sglang/python/sglang/srt/layers/attention/nsa/dequant_k_cache.py", line 194, in dequantize_k_cache_paged
    assert num_tokens <= total_num_tokens
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

## Modifications

1. **`nsa_backend.py`**: Add page table index bounds validation in `init_forward_metadata()` (once per batch) to ensure all indices are within valid KV cache range.

2. **`dequant_k_cache.py`**: Remove incorrect assertion and add clarifying comments.

## Accuracy Tests
After fix:
```
    --model-path deepseek-ai/DeepSeek-V3.2-Exp \
    --tp-size 2 \
    --pp-size 4 \
    --max-running-requests 128 \
    --max-total-tokens 100000 \
    --chunked-prefill-size 8192 \
    --mem-fraction-static 0.80
```
```
python3 benchmark/gsm8k/bench_sglang.py --num-shots 20 --num-questions 1319 --parallel 1319
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1319/1319 [03:10<00:00,  6.91it/s]
Accuracy: 0.958
Invalid: 0.000
Latency: 192.569 s
Output throughput: 678.099 token/s
```

## Checklist

### DIFF
--- a/python/sglang/srt/layers/attention/nsa_backend.py
+++ b/python/sglang/srt/layers/attention/nsa_backend.py
@@ -486,6 +486,14 @@ class NativeSparseAttnBackend(AttentionBackend):
                     page_table_1_flattened.shape[0] == forward_batch.seq_lens_sum
                 ), f"{page_table_1_flattened.shape[0] = } must be the same as {forward_batch.seq_lens_sum = }"
 
+                # Validate page table indices (once per batch)
+                kv_cache_size = forward_batch.token_to_kv_pool.size
+                if page_table_1_flattened.numel() > 0:
+                    max_idx = page_table_1_flattened.max().item()
+                    assert (
+                        max_idx < kv_cache_size
+                    ), f"Invalid page table index: max={max_idx}, kv_cache_size={kv_cache_size}"
+
             if topk_transform_method == TopkTransformMethod.RAGGED:
                 topk_indices_offset = torch.repeat_interleave(
                     cu_seqlens_k[:-1],
--- a/python/sglang/srt/layers/attention/nsa_backend.py
+++ b/python/sglang/srt/layers/attention/nsa_backend.py
@@ -489,6 +489,7 @@ class NativeSparseAttnBackend(AttentionBackend):
                 ), f"{page_table_1_flattened.shape[0] = } must be the same as {forward_batch.seq_lens_sum = }"
 
                 # Validate indices when logical tokens exceed physical capacity
+                # This is likely to be triggered by PP with high kv reuse & parallelism
                 kv_cache_capacity = (
                     forward_batch.token_to_kv_pool.size
                     + forward_batch.token_to_kv_pool.page_size
--- a/python/sglang/srt/layers/attention/nsa_backend.py
+++ b/python/sglang/srt/layers/attention/nsa_backend.py
@@ -463,14 +463,16 @@ class NativeSparseAttnBackend(AttentionBackend):
                 ]
             )
 
-            # Check if MHA with FP8 needs page_table_1_flattened for dequantization
+            # Check if MHA FP8 dequantization is needed
             mha_dequantize_needed = (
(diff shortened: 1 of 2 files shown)


python/sglang/srt/layers/attention/nsa_backend.py
```diff
@@ -486,6 +486,14 @@ def init_forward_metadata(self, forward_batch: ForwardBatch):
                     page_table_1_flattened.shape[0] == forward_batch.seq_lens_sum
                 ), f"{page_table_1_flattened.shape[0] = } must be the same as {forward_batch.seq_lens_sum = }"
 
+                # Validate page table indices (once per batch)
+                kv_cache_size = forward_batch.token_to_kv_pool.size
+                if page_table_1_flattened.numel() > 0:
+                    max_idx = page_table_1_flattened.max().item()
```
Fridge003, line 498: Is page_table_1_flattened located on CPU memory? Since this operation might introduce synchronization

YAMY1234: I had the same concern, which is why I moved the validation into init_forward_metadata so it runs once per batch (instead of once per layer inside dequantize_k_cache_paged), where the impact should be limited.
Another option is we can completely remove the check ---- but the issue when it goes wrong, Triton may just hit an illegal memory access with little context.
What's your opinion?

Fridge003: Yes that's also true...
Maybe we can keep the original assertion when pp is disabled, and use the new assertion when pp is enabled?

YAMY1234: Good point! PP may just make that easier to trigger and I made some optimization:
Current checking will only be applied when forward_batch.seq_lens_sum > kv_cache_capacity (similar to original check), so in most normal cases this will not be triggered~

Also I changed `page_table_1_flattened`'s constructing logic, it will only be constructed when `any(forward_batch.extend_prefix_lens_cpu)`, I just found all the use cases are with  prefix sharing enabled.

Fridge003: Maybe add a comment mentioning this case only happens under PP?

YAMY1234: Updated comment~

Fridge003 (approved):

PR #7725, opened 2025-07-02
Support FlashAttention3 page_size > 1 and topk > 1 case with paged attn and spec decode

yubofredwang:
## Motivation

Fully enable flash attention 3 backend with speculative decoding and paged attention use case where topk > 1 and page_size > 1. 

The current implementation is still a fake support for page size > 1. In the `assign_draft_cache_locs`. we directly move the indices instead of the real kv cache. This only works when the kernel backend runs with page size = 1. If the kernel backend runs with page size > 1, we need to duplicate the real KV cache. The overhead of duplicating KV cache seems okay because the draft KV cache only has one layer.

Also fixed bunch of bugs in the original topk > 1 and page_size > 1 flash attention backend.

## Modifications

1. General support of the case.
2. Cuda graph support of the case.
3. Fixed kv cache handling during draft forward.

Produce `source_cache_loc` and `target_cache_loc` in assign_draft_cache_locs, and call `self.draft_model_runner.token_to_kv_pool.move_kv_cache(target_cache_loc, source_cache_loc)`. To copy the actual kv cache.

Example case:
Given: speculative_num_steps=5, page_size=4, topk=8, num_new_pages_per_topk_=2
assuming prefix is: [4,5,6,7,8,9,10], the last_page_len to be duplicated = 3, the generated source_cache_loc and source_cache_loc(len=21): [8,9,10] * 7
[8,9,10,8,9,10...]
target_cache_loc(len=21):
[16,17,18,24,25,26...] 
out_cache_loc to be filled:
[11,12,13,14,15,19,20,21,22,23,24...]

Duplicate KV Cache

## Benchmark Result:
```
SGLANG_ALLOW_OVERWRITE_LONGER_CONTEXT_LEN=1 python3 -m sglang.launch_server --model /shared/public/elr-models/meta-llama/Meta-Llama-3.1-8B-Instruct/07eb05b21d191a58c577b4a45982fe0c049d0693 --speculative-algorithm EAGLE3 --speculative-draft-model-path /shared/public/elr-models/jamesliu1/sglang-EAGLE3-Llama-3.1-Instruct-8B/e5ed08d66f528a95ce89f5d4fd136a28f6def714 --speculative-num-steps 2 --speculative-eagle-topk 2 --speculative-num-draft-tokens 5 --mem-fraction 0.9 --page-size 4 --attention-backend fa3 --dtype bfloat16 --trust-remote-code
```

Send one with page_size=4
```
+-------------+--------+------------+-----------------+
[...]

PR #15107, opened 2025-12-14
Fix IMA when fa3 + spec + topk + page

hnyls2002:
@yubofredwang 

debugging branch: #15077 

This fixes #14624 

Follow-up of #7725


yubofredwang:
Overall looks good. I think the CUDA graph change is also needed. I can do it tomorrow and add some related unit test.

PR #15113, opened 2025-12-14
Add NanoV3 reasoning parser support

danielafrimi:
- Added `NanoV3Detector` class inheriting from `DeepSeekR1Detector` (handles <think>...</think> format with force_reasoning=True by default)
 
- Added "nano_v3" to the conditional check for enable_thinking parameter support.


ispobock (commented):
Could you fix the lint issue?

ispobock (approved):

PR #15135, opened 2025-12-15
fix(function_call): Add schema-based type conversion to Qwen3CoderDetector

qywu:
## Summary
- Add schema-based type conversion for parameter values (int, float, bool, object, array) similar to vLLM's implementation
- Update parameter regex to handle missing `</parameter>` tags more robustly
- Add explicit null value handling
- Add tool schema caching for performance

This change has been well tested with Qwen3 Coder models.


qywu:
cc @zhyncs

PR #15151, opened 2025-12-15
MoE Refactor: Refactor `fp8.py` -> `flashinfer_trllm.py`

b8zhong:
Part of https://github.com/sgl-project/sglang/issues/8715. There are many bugs recently related to Flashinfer MoE, partially bc this file and `modelopt_quant.py` is getting too complex.

### DIFF
--- a/python/sglang/srt/layers/quantization/fp8.py
+++ b/python/sglang/srt/layers/quantization/fp8.py
@@ -69,7 +69,6 @@ from sglang.srt.utils import (
     is_sm90_supported,
     is_sm100_supported,
     log_info_on_rank0,
-    next_power_of_2,
     print_warning_once,
     set_weight_attrs,
     use_intel_amx_backend,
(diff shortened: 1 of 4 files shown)


python/sglang/srt/layers/quantization/fp8.py
```diff
@@ -1113,7 +1040,16 @@ def create_moe_runner(
...
                 moe_runner_backend = MoeRunnerBackend.TRITON
-        if moe_runner_backend.is_deep_gemm() or moe_runner_backend.is_triton():
+        if (
+            moe_runner_backend.is_deep_gemm()
+            or moe_runner_backend.is_triton()
+            or moe_runner_backend.is_flashinfer_trtllm()
+        ):
+            if moe_runner_backend.is_flashinfer_trtllm():
+                # Import to register the fused function
+                from sglang.srt.layers.moe.moe_runner.flashinfer_trtllm import (  # noqa: F401
```
ch-wan: How about moving this import to the beginning of this file? Our current strategy for avoiding circular import is to delay import for quant kernels in moe runner. Example: https://github.com/sgl-project/sglang/blob/5e1a495c65e19403fdc7d2054e97120902333526/python/sglang/srt/layers/moe/moe_runner/deep_gemm.py#L215-L221

python/sglang/srt/layers/quantization/fp8.py
ch-wan: As this function is much simpler, we can merge it with the original `apply` function.

b8zhong: Merged this into apply logic 👍

ch-wan (approved):

PR #14038, opened 2025-11-27
[diffusion][MUSA] Support diffusion model

yeahdongcn:
## Motivation

This PR adds diffusion model support on [Moore Threads](https://en.mthreads.com/) GPUs.

## Modifications

1. Enable MUSA backend auto-selection based on the existing code structure.
2. Replace many hard-coded device names with get_device to ensure the correct device is selected.
3. Add workarounds for several torch_musa issues (with comments indicating items to be addressed in future torch_musa releases).
4. Currently, only the SDPA backend is supported. FA support will be available soon.

## Accuracy Tests

1. Tested text -> video model: Wan2.1-T2V-1.3B-Diffusers
2. Tested text -> image model: FLUX.1-dev

Test script:
```python
from sglang.multimodal_gen import DiffGenerator


def main():
    # Create a diff generator from a pre-trained model
    generator = DiffGenerator.from_pretrained(
        # model_path="/ws/Wan2.1-T2V-1.3B-Diffusers",
        model_path="/ws/FLUX.1-dev",
        # model_path="/ws/Qwen-Image",
        num_gpus=1,  # Adjust based on your hardware
    )

    # Provide a prompt for your video
    prompt = "A dynamic, cinematic scene of Po from Kung Fu Panda playing badminton in a bright, stylized bamboo forest arena. Po wears traditional kung fu training clothes with subtle fabric movement as he breathes and prepares. The camera starts with a smooth dolly-in shot from behind Po, revealing soft sunlight rays passing through tall bamboo. Po performs agile kung fu-inspired footwork, mixing martial-arts movements with badminton techniques—quick sidesteps, spinning reaches, and exaggerated squash-style smashes. His expressions are energetic and humorous, with Po’s signature clumsy-but-determined charm. The badminton shuttlecock leaves soft motion trails, and each hit produces a gust of stylized wind and floating bamboo leaves. The atmosphere is playful and cinematic with high-motion clarity, dynamic lighting, and smooth animation. The style is high-quality CG animation similar to the Kung Fu Panda films, with detailed textures, expressive squash-and-stretch animation, and vibrant colors. The final shot shows Po jumping for a dramatic slow-motion smash, camera rotating around him with heroic lighting and particles floating in the air."

    # Generate the video
    video = generator.generate(
        prompt,
        # return_frames=True,  # Also return frames from this call (defaults to False)
        output_path="my_videos/",  # Controls where videos are saved
        save_output=True,
    )


if __name__ == "__main__":
    main()
```

[...]

PR #15232, opened 2025-12-16
[Diffusion] Use current_platform.device_type to replace hard-coded cuda device

yeahdongcn:
## Motivation

While adapting `multimodal_gen` to MThreads GPUs (see previous PR: https://github.com/sgl-project/sglang/pull/14038), I noticed several hard-coded references to the CUDA device. For MTGPU, `musa` is used as the device type instead of `cuda`.  

To enable smoother migration and reduce follow-up changes in future PRs, this PR refactors the device selection logic by using `current_platform.device_type` to replace hard-coded `cuda` strings.

## Modifications

- Replaced hard-coded `cuda` device strings with `current_platform.device_type`

## Checklist

### DIFF
--- a/python/sglang/multimodal_gen/configs/sample/sampling_params.py
+++ b/python/sglang/multimodal_gen/configs/sample/sampling_params.py
@@ -430,8 +430,8 @@ class SamplingParams:
             "--generator-device",
             type=str,
             default=SamplingParams.generator_device,
-            choices=["cuda", "cpu"],
-            help="Device for random generator (cuda or cpu). Default: cuda",
+            choices=["cuda", "musa", "cpu"],
+            help="Device for random generator (cuda, musa or cpu). Default: cuda",
         )
         parser.add_argument(
             "--num-frames",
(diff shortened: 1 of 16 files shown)


python/sglang/multimodal_gen/envs.py
```diff
@@ -422,10 +422,14 @@ def get_torch_distributed_backend() -> str:
         )
 
 
-def get_device(local_rank: int) -> torch.device:
+def get_device(local_rank: Optional[int] = None) -> torch.device:
```
mickqian: actually, could we use `Platform::get_device_name`

mickqian (changes requested):

python/sglang/multimodal_gen/envs.py
yeahdongcn: Okay, I’ll update it that way. Another thing I’d like to discuss is whether the `Platform` interface will also be applied to the sglang srt side, since we are currently importing `get_device` from `sglang.srt.utils` there.

mickqian: No, we prefer duplication for now. Will consider code reuse in the future

yeahdongcn: Done. Please take a look, thanks.

mickqian (commented):
I didn't see any CI. after this pr, will sgl-diffusion be available in this platform? and how should we guard it?

PR #15264, opened 2025-12-16
MiniMax-M2: pipeline-parallel safe forward/weight-loading + comm

jtreal:
## Motivation

Fixes initialization/cuda-graph capture failures when using --pp-size > 1.
Prevents partial/incorrect weight loading under PP.
Aligns MiniMax-M2 model code with current communicator API.

## Modifications

Make MiniMax-M2 model forward pipeline-parallel aware: Non-first PP stages consume pp_proxy_tensors.
Non-last PP stages return PPProxyTensors and skip logits/head.
Make load_weights() robust under PP when checkpoint iterator yields global keys: Skip only parameters that are expected to be absent on this PP rank. Keep raising on unexpected missing keys (avoids silently skipping real mapping bugs).
Update LayerScatterModes.init_new(...) invocation to pass is_next_layer_sparse (required by current communicator context).

## Checklist

### DIFF
--- a/python/sglang/srt/models/minimax_m2.py
+++ b/python/sglang/srt/models/minimax_m2.py
@@ -16,6 +16,7 @@
 """Inference-only MiniMax M2 model compatible with HuggingFace weights."""
 
 import logging
+import re
 from typing import Iterable, Optional, Set, Tuple, Union
 
 import torch
@@ -71,6 +72,8 @@ from sglang.srt.utils import (
 )
 
 logger = logging.getLogger(__name__)
+
+_LAYER_RE = re.compile(r"^model\.layers\.(\d+)\.")
 
 
 class MiniMaxM2RMSNormTP(nn.Module):
@@ -347,16 +350,11 @@ class MiniMaxM2Attention(nn.Module):
         self.total_num_kv_heads = config.num_key_value_heads
 
         if self.total_num_kv_heads >= tp_size:
-            # Number of KV heads is greater than TP size, so we partition
-            # the KV heads across multiple tensor parallel GPUs.
             assert self.total_num_kv_heads % tp_size == 0
         else:
-            # Number of KV heads is less than TP size, so we replicate
-            # the KV heads across multiple tensor parallel GPUs.
             assert tp_size % self.total_num_kv_heads == 0
         self.num_kv_heads = max(1, self.total_num_kv_heads // tp_size)
 
-        # Use head_dim from config if available, otherwise calculate
         self.head_dim = getattr(
             config, "head_dim", self.hidden_size // self.total_num_heads
         )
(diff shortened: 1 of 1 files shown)


python/sglang/srt/models/minimax_m2.py
```diff
@@ -347,31 +350,22 @@ def __init__(
         self.total_num_kv_heads = config.num_key_value_heads
 
         if self.total_num_kv_heads >= tp_size:
-            # Number of KV heads is greater than TP size, so we partition
```
rogeryoungh, line 350: Let's keep the original comments as they are helpful for maintenance.

python/sglang/srt/models/minimax_m2.py
```diff
@@ -515,16 +505,29 @@ def __init__(
...
+            # Newer API
+            self.layer_scatter_modes = LayerScatterModes.init_new(
+                layer_id=layer_id,
+                num_layers=config.num_hidden_layers,
+                is_layer_sparse=self.is_layer_sparse,
+                is_previous_layer_sparse=is_previous_layer_sparse,
+                is_next_layer_sparse=is_next_layer_sparse,
+            )
+        except TypeError:
+            # Older API (no is_next_layer_sparse)
```
rogeryoungh, line 522: The older API implementation for LayerScatterModes doesn't seem to exist in the current codebase? We can probably remove this part.

python/sglang/srt/models/minimax_m2.py
```diff
@@ -790,32 +787,65 @@ def forward(
         positions: torch.Tensor,
         forward_batch: ForwardBatch,
         input_embeds: torch.Tensor = None,
-    ) -> torch.Tensor:
-        # _print_tensor_info(input_ids, "input_ids")
-        hidden_states = self.model(input_ids, positions, forward_batch, input_embeds)
+        pp_proxy_tensors: Optional[PPProxyTensors] = None,
+    ) -> Union[torch.Tensor, PPProxyTensors]:
+        pp_group = get_pp_group()
```
rogeryoungh, line 792: It's better to move `get_pp_group()` to `__init__` instead of calling it in forward.

python/sglang/srt/models/minimax_m2.py
```diff
@@ -790,32 +787,65 @@ def forward(
...
         )
 
+    def _pp_should_skip_param(self, param_name: str) -> bool:
+        pp_group = get_pp_group()
+        if pp_group is None or pp_group.world_size <= 1:
+            return False
+
+        m = _LAYER_RE.match(param_name)
+        if m is not None:
+            layer_id = int(m.group(1))
```
rogeryoungh, line 824: We can use the existing `get_layer_id` helper here to stay consistent with other model implementations.

rogeryoungh (commented):
Thanks for the PR! I've left a few suggestions above regarding the implementation details.

hnyls2002:
@jtreal pipeline-parallel support for MiniMax-M2 landed on main via #19577, so this is superseded. Please reopen if I've missed something.

PR #15287, opened 2025-12-16
fix mindspore import warning

b8zhong:
Screenshot 2025-12-16 at 2 13 55 PM

Screenshot 2025-12-16 at 2 27 30 PM

### DIFF
--- a/python/sglang/srt/models/mindspore.py
+++ b/python/sglang/srt/models/mindspore.py
@@ -291,7 +291,7 @@ class MindSporeForCausalLM(torch.nn.Module):
         input_ids: torch.Tensor,
         positions: torch.Tensor,
         forward_batch: ForwardBatch,
-    ) -> Tensor:
+    ) -> "ms.Tensor":
         # prepare base inputs
         model_inputs = self.prepare_inputs(input_ids, positions, forward_batch)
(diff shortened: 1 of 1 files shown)


python/sglang/srt/entrypoints/engine.py
```diff
@@ -97,6 +98,7 @@ def _launch_subprocesses(
     # Configure global environment
     configure_logger(server_args)
     _set_envs_and_config(server_args)
+    torch.set_float32_matmul_precision(envs.SGLANG_FLOAT32_MATMUL_PRECISION.get())
```
merrymercy: move this to `_set_envs_and_config`

merrymercy (approved):

python/sglang/srt/models/mindspore.py
iforgetmyname:
```suggestion
        position_ids: "ms.Tensor",
```
this type checking causes NameError on non-NPU platforms

iforgetmyname:
```suggestion
    ) -> "ms.Tensor":
```
also the same problem

iforgetmyname (approved):
lgtm

PR #15298, opened 2025-12-17
Clean up __init__ function of the scheduler and event loop for PD

merrymercy:
- Organize the  `__init__` function of scheduler into many subfunctions
- Make the event loop of PD 99% the same as unified.

### DIFF
--- a/python/sglang/srt/server_args.py
+++ b/python/sglang/srt/server_args.py
@@ -26,8 +26,6 @@ import random
 import tempfile
 from typing import Any, Callable, Dict, List, Literal, Optional, Union
 
-import orjson
-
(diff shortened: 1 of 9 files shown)


python/sglang/srt/server_args.py
```diff
@@ -752,10 +739,23 @@ def _handle_missing_default_values(self):
             self.random_seed = random.randint(0, 1 << 30)
         if self.mm_process_config is None:
             self.mm_process_config = {}
+
+        # Handle ModelScope model downloads
```
Hide-on-bushsh, line 743: Is there any dependency between this code and this PR?  @merrymercy

PR #15327, opened 2025-12-17
[Gemma] SWAchunk cache for gemma

angry-crab:
## Motivation

Hi, I'm working with @giantleap2012 on #11346. This PR adds SWA chunk cache support to Gemma3/Gemma2 models. Quote from #11346 for detailed description and use case.

>Add support for SWAChunkCache to support sliding window attention. Currently SWAChunkCache evicts the whole chunk when it reaches the attention_chunk_size. But for SWA, we will need to keep the most recent sliding_window_size tokens.
>
>A model architecture heavily rely on SWA layers, which means the SWA tokens pool will be much smaller than full tokens pool
>Long decoding length which far exceeds sliding_window_size

`python -m sglang.launch_server --model-path /gemma3_1b_it/ --tp 1 --trust-remote-code --port 30000 --load-format auto --max-prefill-tokens 16384 --chunked-prefill-size 1024 --context-length=32768 --max-running-requests 2048 --cuda-graph-bs 2 4 8 16 32 64 128 192 256 320 384 448 512 576 640 704 768 832 896 960 1024 1536 2048 --grammar-backend outlines --attention-backend triton --mem-fraction-static 0.7 --swa-full-tokens-ratio 0.2 --disable-radix-cache`

## Modifications

>Modify the SWAChunkCache to support SWA by keeping the most recent sliding_window_size tokens instead of evicting the whole chunk. If no sliding_window_size is passed in, the behavior is exactly the same as before.

Adds support to Gemma3/Gemma2 models.

## Checklist

### DIFF
--- a/python/sglang/srt/configs/model_config.py
+++ b/python/sglang/srt/configs/model_config.py
@@ -158,6 +158,9 @@ class ModelConfig:
         self.attention_chunk_size = getattr(
             self.hf_text_config, "attention_chunk_size", None
         )
+        self.sliding_window_size = getattr(
+            self.hf_text_config, "sliding_window", -1
+        )
         self.is_hybrid_swa = is_hybrid_model(
             self.hf_config.architectures,
             hybrid_kvcache_ratio=hybrid_kvcache_ratio,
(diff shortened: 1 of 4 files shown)


hnyls2002:
@angry-crab SWA chunk-cache eviction based on the sliding window size has since been implemented on main (#16507, #17220, #17397; see SWAChunkCache and ScheduleBatch.maybe_evict_swa), so this is superseded. Please reopen if there is still a gap for Gemma.

PR #15363, opened 2025-12-18
[NPU]mindspore model support moe

zzzzzzzxh:
## Motivation
- Enable per‑architecture expert‑location config for MindSpore models, instead of a single generic path.
- Keep dense Qwen3 free from MoE‑specific config while Qwen3 MoE provides it when needed.
- Decouple upstream from downstream details so future MindSpore models can plug in their own behavior cleanly.

## Modifications

- Change MindSporeForCausalLM.get_model_config_for_expert_location to delegate to the selected MindSpore subclass based on HF architectures .
- Implement get_model_config_for_expert_location only in the Qwen3 MoE subclass of sgl_mindspore ; dense Qwen3 does not override (returns None ).
- No changes to the forward/logits path; only the config lookup is updated.
## Checklist


python/sglang/srt/models/mindspore.py
```diff
@@ -301,5 +302,19 @@ def forward(
         logits_result = LogitsProcessorOutput(next_token_logits=tensor_ms2torch(logits))
         return logits_result
 
+    @classmethod
+    def get_model_config_for_expert_location(cls, config):
```
chz34, line 306: Check and call the corresponding func from self.model,  to support different configuration in sgl_mindspore.  Instead of a general configuration there.

python/sglang/srt/models/mindspore.py
```diff
@@ -28,6 +28,19 @@
 logger = logging.getLogger(__name__)
 
 
+def _get_arch_from_config(config):
+    mindspore_models = import_model_classes("sgl_mindspore.models")
+    architectures = getattr(config, "architectures", [])
+    if isinstance(architectures, str):
+        architectures = [architectures]
+    if not architectures:
+        logger.warning("No model architectures are specified")
```
ping1jing2: considering the `ValueError`, is it more appropriate for this branch to raise a error or print warning.

zzzzzzzxh: Done. Changed it to raise a ValueError.

ping1jing2 (approved):

PR #15393, opened 2025-12-18
fix: handle custom model outputs without to_tuple() method

majiayu000:
## Summary
- Use `return_dict=True` instead of `return_dict=False` in transformers wrapper
- Extract hidden states from output object using attribute access (`last_hidden_state` or `hidden_states`)
- Adds fallback for tuple-like access

## Test plan
- Added unit test `test/srt/test_transformers_output_handling.py`
- Test covers various output formats (last_hidden_state, hidden_states, tuple access)

Fixes #15339


majiayu000:
Hi maintainers, could you please add the `run-ci` label to trigger CI tests? Thank you!

PR #15432, opened 2025-12-19
[FusedMoE] Fix fused w13 tp sharded weight loading

yinghai:
## Motivation
For fused w13 weights [w1; w3], we need to shard w1 and w3 independently, then concatenate them back to maintain the
[w1_shard; w3_shard] structure per TP rank, because our gelu_and_mul kernel is assuming this kind of layout:

https://github.com/sgl-project/sglang/blob/4b4050e27e219a4d5bfadc4b4846414a093ed56a/python/sglang/srt/layers/elementwise.py#L530-L532

This is an RFC as I don't know how the broader community is using w13 weights with TP. 

## Checklist


fzyzcjy (approved):
LGTM

python/sglang/srt/layers/moe/fused_moe_triton/layer.py
fzyzcjy, line 390: tiny nit: I may personally rename it to param_start to avoid ambiguous whether it is param or weight start (which I found when reading this code), but it is a separate thing and do not need to be in this pr

Issue #15399, opened 2025-12-18
[Bug] speculative EAGLEWorker forward_draft_extend_after_decode cuda graph has bug


llc-kc:
### Checklist

- [x] I searched related issues but found no solution.
- [x] The bug persists in the latest version.
- [x] Issues without environment info and a minimal reproducible demo are hard to resolve and may receive no feedback.
- [x] If this is not a bug report but a general question, please start a discussion at https://github.com/sgl-project/sglang/discussions. Otherwise, it will be closed.
- [x] Please use English. Otherwise, it will be closed.

### Describe the bug

the root cause of https://github.com/sgl-project/sglang/issues/15143
DeekSeek both V3.2 and V3.1 DP+MTP decode can be failed by vectorized gather kernel index out of bounds, since EAGLEWorker forward_draft_extend_after_decode cuda graph replay make scheduler.running_batch.spec_info.topk_index bacome a weird tensor like [-4886055425978319163 4510424140297813696]

### Reproduction

please check https://github.com/sgl-project/sglang/issues/15143

### Environment

Prefill: 2x H800 GPU, decode: 2x H800 GPU
sglang:
lmsysorg/sglang:v0.5.6.post1-cu129-amd64

PR #15457, opened 2025-12-19
Fix DeepSeek PD decode DP + MTP failed by vectorized gather kernel in…

llc-kc:
## Motivation
Fix DeepSeek PD TP+DP+MTP, decode failed by vectorized gather kernel index out of bounds
as Descripted in 
https://github.com/sgl-project/sglang/issues/15143
https://github.com/sgl-project/sglang/issues/15399

## Modifications

Using clone to avoid tensor of output in cuda graph output_buffers used at other place simultaneously.

## Accuracy Tests

None

## Benchmarking and Profiling

None


yudian0504:
https://github.com/sgl-project/sglang/pull/16551

Possibly different fixes for the same issue.

llc-kc:
@yudian0504 thanks, If your pr fix this problem and merged, I will close this pr.

yudian0504:
Additionally, it seems this issue is **NOT** exclusive to the PD disagg,  we've encountered it in our non-PD setups as well, although the probability is quite low, making it difficult to reproduce.

wcsjtu:
it works for me

llc-kc:
@hnyls2002 Hi, Is there any progress on this issue? Have any other PRs already fixed it?

hnyls2002:
@llc-kc thanks for the investigation. This patches forward_draft_extend_after_decode in eagle_worker.py, which was removed when spec V1 was deprecated in #25464, so the patched code path no longer exists on main. Closing as obsolete - please reopen if the issue still reproduces with the V2 spec workers on current main.

PR #15506, opened 2025-12-20
fix(qwen_vl): add semaphore to serialize video decoding for thread safety

ShuhangGe:
This commit fixes a concurrency bug in the Qwen VL video processor where
multiple concurrent requests decoding different videos would cause crashes.

## Problem

When multiple async requests process different videos simultaneously,
the decord library's VideoReader.get_batch() call fails with:

    DECORDError: Check failed: avcodec_send_packet(dec_ctx_.get(), pkt.get()) >= 0 (-11 vs. 0)
    Thread worker: Error sending packet.

Error code -11 (EAGAIN) indicates resource contention in FFmpeg's
internal threaded decoder, which is not thread-safe for concurrent
multi-file access.

## Root Cause

The decord library uses FFmpeg internally for video decoding. When
multiple VideoReader instances decode different video files at the same
time, FFmpeg's internal thread workers collide, causing the EAGAIN error.
This is a known issue with decord due to C package conflicts that are
difficult to resolve at the library level.

## Solution

Add an asyncio.Semaphore(1) to serialize the get_batch() call, ensuring
only one video decoding operation runs at a time.

## Note: This is a Quick Fix

This is a minimal fix to quickly resolve the crash. The semaphore
serializes video decoding, which may slightly reduce throughput when
processing multiple different videos concurrently.

### Future Improvement: OpenCV-based Decoding

A more comprehensive solution would be to use OpenCV (cv2.VideoCapture)
as the primary video decoder with decord as a fallback. This approach
is being adopted by QwenLM/Qwen3-VL in PR #1078, which:

1. Uses OpenCV as the primary decoder (more thread-safe and compatible)
2. Adds a timeout mechanism for decord to prevent hanging
3. Falls back to decord only when OpenCV fails

To implement this in SGLang would require:

1. **Modify `load_video()` in `sglang/srt/utils/common.py`**:
   - Keep temporary video files alive (currently deleted immediately)
   - Attach the file path to the VideoReader object for later use

2. **Modify `preprocess_video()` in `qwen_vl.py`**:
   - Use OpenCV as the primary decoder:
     def _decode_with_opencv(video_path: str, frame_indices: np.ndarray) -> np.ndarray:
         cap = cv2.VideoCapture(video_path)
         frames = []
         for idx in frame_indices:
             cap.set(cv2.CAP_PROP_POS_FRAMES, int(idx))
             ret, frame = cap.read()
             if ret:
                 frames.append(cv2.cvtColor(frame, cv2.COLOR_BGR2RGB))
         cap.release()
         return np.stack(frames)
        - Fall back to decord with semaphore protection if OpenCV fails

This would allow full parallel video decoding while maintaining reliability.
Happy to submit a follow-up PR for the OpenCV implementation.

## References

- https://github.com/QwenLM/Qwen3-VL/pull/1078 (OpenCV fallback in Qwen3-VL)
- https://github.com/dmlc/decord/issues/269 (get_batch hang issue)
- https://github.com/dmlc/decord/issues/220 (deadlock in multi-threaded training)
- https://github.com/dmlc/decord/issues/1 (crashes with concurrent VideoReader)

### DIFF
--- a/python/sglang/srt/multimodal/processors/qwen_vl.py
+++ b/python/sglang/srt/multimodal/processors/qwen_vl.py
@@ -1,3 +1,4 @@
+import asyncio
 import math
 import os
 import re
@@ -44,6 +45,8 @@ FRAME_FACTOR = 2
 FPS = 2.0
 FPS_MIN_FRAMES = 4
 FPS_MAX_FRAMES = 768
+
+_video_decode_semaphore = asyncio.Semaphore(1)
 
 
 def smart_resize(
@@ -156,7 +159,8 @@ async def preprocess_video(
     )
     idx = np.linspace(0, total_frames - 1, num=nframes, dtype=np.int64)
     idx = np.unique(idx)
-    video_np = vr.get_batch(idx).asnumpy()
+    async with _video_decode_semaphore:
+        video_np = vr.get_batch(idx).asnumpy()
     video = torch.from_numpy(video_np).pin_memory()
     video = video.permute(0, 3, 1, 2)  # Convert to TCHW format
 


hnyls2002:
@ShuhangGe thanks for the fix. Video decoding was moved out of the Qwen-VL processor and decord was replaced by torchcodec as the primary decoder in #20055 (now in python/sglang/srt/utils/video_decoder.py), so the decord concurrency path this PR patches no longer exists. Closing as obsolete - please reopen if this still reproduces on current main.

ShuhangGe:
Hi Liangsheng Yin,

Thank you for the explanation and for taking the time to review my PR.I
really appreciate the work the SGLang community is doing. SGLang is a great
open-source project, and I’m very interested in contributing more to it. As
someone deeply interested in open-source collaboration, I’d like to learn
the right way to submit useful PRs to the project.

Could you share any suggestions on how new contributors should get started?
For example, should I focus on existing issues, performance-related
improvements, tests, documentation, or specific areas of the codebase that
currently need help?

I would be very happy to continue learning from the community and
contribute to SGLang development if there is an opportunity.

Thanks again!

Best,
Shuhang

PR #15511, opened 2025-12-20
[Diffusion] Wan video model support zero-cost weight offload and overlap with compute

BBuf:
## Motivation

I have identified the reason for the slow speed of wan2.2. A full profiler would reveal that the first step and the 19th step are significantly slower and have similar speeds, while all other intermediate steps are fast and fully meet the expected 2x speed improvement from cp4 to cp8. The first and 19th steps are about 7 times slower than a normal step, equivalent to 7 normal steps. 

The reason is that wan2.2 uses dual transformers, and we have `dit_cpu_offload` enabled. Therefore, the weights are on cpu after loaded, and during the first step, the weights of both `transformer` and `transformer_2` are copied from the CPU to the CUDA device, making the first step very slow. Then, at the 19th step, a dual-stream switch occurs, requiring both `transformer` and `transformer_2` to be offloaded back to the CPU, their weights swapped, and then copied back to the CUDA device. 

## main branch:

```bash
sglang generate   --model-path Wan-AI/Wan2.2-T2V-A14B-Diffusers   --text-encoder-cpu-offload   --pin-cpu-memory   --num-gpus 8   --ulysses-degree 8 --attention-backend sage_attn  --enable-torch-compile --prompt "A cat walks on the grass, realistic" --num-frames 81 --height 720 --width 1280 --num-inference-steps 27 --guidance-scale 3.5 --guidance-scale-2 4.0 --perf-dump-path /home/lmsys/bbuf/dump/wan_step_profile_cp8_main.json

100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 27/27 [02:28<00:00,  5.52s/it]
[12-19 08:37:22] [DenoisingStage] average time per step: 5.5156 seconds
[12-19 08:37:23] [DenoisingStage] finished in 149.6943 seconds



"denoise_steps_ms": [
    35999.06893167645,
    3261.483933776617,
    3270.5406425520778,
    3267.8588768467307,
    3260.3964526206255,
    3263.016454875469,
    3268.026988953352,
    3264.5184732973576,
    3264.636719599366,
    3267.1875776723027,
    3268.562350422144,
    3268.1023878976703,
    3266.7769035324454,
    3268.044295720756,
    3264.268895611167,
    3271.0087513551116,
    3267.674465663731,
    3266.1060262471437,
    31282.590138725936,
    3263.5639663785696,
    3262.301029637456,
    3262.3210102319717,
    3261.833382770419,
    3264.719443395734,
    3265.314467251301,
    3267.530156299472,
    3261.9533529505134
  ],
```

### pr

```bash
sglang generate   --model-path Wan-AI/Wan2.2-T2V-A14B-Diffusers   --text-encoder-cpu-offload   --pin-cpu-memory   --num-gpus 8   --ulysses-degree 8 --attention-backend sage_attn  --enable-torch-compile --prompt "A cat walks on the grass, realistic" --num-frames 81 --height 720 --width 1280 --num-inference-steps 27 --guidance-scale 3.5 --guidance-scale-2 4.0 --dit-layerwise-offload true --perf-dump-path /home/lmsys/bbuf/dump/wan_step_profile_cp8_async_offload.json


100%|██████████████████████████████████████████████████████████████████████████████████████████████████| 27/27 [01:33<00:00,  3.46s/it]
[12-20 02:59:10] [DenoisingStage] average time per step: 3.4553 seconds
[12-20 02:59:10] [DenoisingStage] finished in 94.2283 seconds

"denoise_steps_ms": [
    7717.6975486800075,
    3275.042257271707,
    3280.2467988803983,
    3282.276245765388,
    3292.9044039919972,
    3286.5121429786086,
    3273.5616639256477,
    3271.6003246605396,
    3275.5934856832027,
    3291.9061705470085,
    3293.934356421232,
    3289.3909830600023,
    3298.0582248419523,
    3300.408118404448,
    3305.0247132778168,
    3299.3013756349683,
    3302.0150866359472,
    3299.040620215237,
    3291.0401169210672,
    3296.6199973598123,
    3290.26335850358,
    3302.190547809005,
    3295.942653901875,
    3297.3329443484545,
    3297.713255509734,
    3295.0284238904715,
    3290.1963284239173
  ],
```

**149.69->94.22s, speed up 58%** 。

图片

memcpy and compute are completely overlapped, with no additional overhead. 

## Warmup

Whether torch compile is enabled or not, I found that the performance of the first step is about 7 times slower than the subsequent steps, which is unacceptable. I profiled the full stage for both scenarios and discovered that the reason for the long initialization time is the initialization overhead of the NCCL All2All. This overhead should not be in the denoise stage but should be handled in advance. Therefore, I implemented a pre-warmup logic specifically for the All2All operation to eliminate this overhead. Now, without torch compile, the time for the first step is almost the same as that of the subsequent steps. Even with compile enabled, the time for the first step is only about twice that of the subsequent steps. 

5f69d545-37be-4156-b92a-c8397fb7359a

## Checklist

### DIFF
--- a/python/sglang/multimodal_gen/runtime/distributed/parallel_state.py
+++ b/python/sglang/multimodal_gen/runtime/distributed/parallel_state.py
@@ -423,6 +423,9 @@ def initialize_model_parallel(
         ring_group=PROCESS_GROUP.RING_PG,
     )
 
+    if ulysses_degree > 1:
+        _warmup_ulysses_communication()
+
     global _TP
     assert _TP is None, "Tensor parallel group is already initialized"
     _TP = init_parallel_group_coordinator(
@@ -933,6 +936,51 @@ def get_ring_parallel_world_size():
 
 def get_ring_parallel_rank():
     return get_sp_group().ring_rank
+
+
+def _warmup_ulysses_communication():
+    """
+    Warmup NCCL communication for Ulysses all-to-all to avoid first-step latency.
+    
+    This function performs a dummy all-to-all operation to initialize NCCL communication
+    channels, which can take several seconds on the first call.
+    """
+    import logging
+    import time
+    
+    logger = logging.getLogger(__name__)
+    logger.info("Warming up Ulysses all-to-all communication...")
+    
+    try:
+        import torch.distributed._functional_collectives as ft_c
+        
+        ulysses_pg = get_sp_group().ulysses_group
+        if ulysses_pg is None:
+            logger.warning("Ulysses group not initialized, skipping warmup")
+            return
+        
+        warmup_start = time.time()
+        
+        device = torch.device(f"cuda:{get_world_group().local_rank}")
+        dummy_tensor = torch.randn(1024, device=device, dtype=torch.float32)
+        
+        output = ft_c.all_to_all_single(
+            dummy_tensor,
+            output_split_sizes=None,
+            input_split_sizes=None,
+            group=ulysses_pg
+        )
+        
+        if isinstance(output, ft_c.AsyncCollectiveTensor):
+            output = output.wait()
+        
+        torch.cuda.synchronize()
+        
+        warmup_time = (time.time() - warmup_start) * 1000
+        logger.info(f"Ulysses communication warmup completed in {warmup_time:.2f}ms")
+        
+    except Exception as e:
+        logger.warning(f"Ulysses communication warmup failed: {e}. Continuing without warmup.")
 
 
 # PP
--- a/python/sglang/multimodal_gen/runtime/distributed/parallel_state.py
+++ b/python/sglang/multimodal_gen/runtime/distributed/parallel_state.py
@@ -957,7 +957,7 @@ def _warmup_ulysses_communication():
         warmup_start = time.time()
 
         device = torch.device(f"cuda:{get_world_group().local_rank}")
-        dummy_tensor = torch.randn(1024, device=device, dtype=torch.float32)
+        dummy_tensor = torch.zeros(1024, device=device, dtype=torch.float32)
 
         output = ft_c.all_to_all_single(
             dummy_tensor,
--- a/python/sglang/multimodal_gen/runtime/distributed/parallel_state.py
+++ b/python/sglang/multimodal_gen/runtime/distributed/parallel_state.py
@@ -32,6 +32,7 @@ If you only need to use the distributed environment without model parallelism,
 """
 import contextlib
 import os
+import time
 import weakref
 from collections import namedtuple
 from collections.abc import Callable
@@ -66,8 +67,6 @@ _CFG: Optional[GroupCoordinator] = None
 _DP: Optional[GroupCoordinator] = None
 _DIT: Optional[GroupCoordinator] = None
 _VAE: Optional[GroupCoordinator] = None
-
-logger = init_logger(__name__)
 
 TensorMetadata = namedtuple("TensorMetadata", ["device", "dtype", "size"])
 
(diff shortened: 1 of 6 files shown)


python/sglang/multimodal_gen/runtime/distributed/parallel_state.py
```diff
@@ -945,6 +937,49 @@ def get_ring_parallel_rank():
...
+        import torch.distributed._functional_collectives as ft_c
+
+        ulysses_pg = get_sp_group().ulysses_group
+        if ulysses_pg is None:
+            logger.warning("Ulysses group not initialized, skipping warmup")
+            return
+
+        warmup_start = time.time()
+
+        device = torch.device(f"cuda:{get_world_group().local_rank}")
```
mickqian, line 959: could we use `current_platform.get_device_name`?

python/sglang/multimodal_gen/runtime/distributed/parallel_state.py
```diff
@@ -945,6 +937,49 @@ def get_ring_parallel_rank():
...
+        dummy_tensor = torch.randn(1024, device=device, dtype=torch.float32)
```
mickqian: will it be faster if we use `torch.zeros`?

python/sglang/multimodal_gen/runtime/loader/component_loader.py
```diff
@@ -709,6 +700,23 @@ def load_customized(
...
+
+            if isinstance(num_layers, int) and num_layers > 0:
+                mgr = LayerwiseOffloadManager(
+                    model,
+                    module_list_attr="blocks",
+                    num_layers=num_layers,
+                    enabled=True,
+                    pin_cpu_memory=server_args.pin_cpu_memory,
+                )
+                mgr.initialize()
```
mickqian: should we move this to `LayerwiseOffloadManager.__init__`?

python/sglang/multimodal_gen/runtime/models/dits/wanvideo.py
```diff
@@ -792,10 +792,20 @@ def forward(
...
                 original_hidden_states = hidden_states.clone()
 
-            for block in self.blocks:
-                hidden_states = block(
-                    hidden_states, encoder_hidden_states, timestep_proj, freqs_cis
-                )
+            offload_mgr = getattr(self, "_layerwise_offload_manager", None)
+            if offload_mgr is not None and getattr(offload_mgr, "enabled", False):
+                for i, block in enumerate(self.blocks):
+                    offload_mgr.prefetch_layer(i + 1, non_blocking=True)
```
mickqian: nit: a context-manager impl for prefetch-and-release may provide better readability

python/sglang/multimodal_gen/runtime/utils/layerwise_offload.py
```diff
@@ -0,0 +1,155 @@
...
+    def __init__(
+        self,
+        model: torch.nn.Module,
+        *,
+        module_list_attr: str,
+        num_layers: int,
+        enabled: bool,
+        pin_cpu_memory: bool = True,
+    ) -> None:
+        self.model = model
```
mickqian, line 33: we might need more documentation for this later

python/sglang/multimodal_gen/runtime/models/dits/wanvideo.py
Edenzzzz: I'm probably missing something, but does FSDP also prefetch and offload layerwise?

python/sglang/multimodal_gen/runtime/distributed/parallel_state.py
BBuf: It's ok. Done.

python/sglang/multimodal_gen/runtime/models/dits/wanvideo.py
BBuf: Good advice. Done.

python/sglang/multimodal_gen/runtime/utils/layerwise_offload.py
BBuf: Done.

python/sglang/multimodal_gen/runtime/loader/component_loader.py
BBuf: Done.

python/sglang/multimodal_gen/runtime/distributed/parallel_state.py
BBuf: Need to obtain each rank and then perform the all2all operation. Getting the device_name only gives us one rank?

python/sglang/multimodal_gen/runtime/models/dits/wanvideo.py
BBuf: Not in the same way. FSDP CPU offload handles parameter movement/sharding at the FSDP boundary (and may offload/reshard according to its policies), but it does not provide an explicit per-layer async H2D prefetch pipeline aligned with the transformer block loop. `dit_layerwise_offload` here is a separate mechanism that explicitly offloads per-block weights to CPU and overlaps H2D prefetch of the next block with compute of the current block for supported DiT models. Also note it is mutually exclusive with `dit_cpu_offload / use_fsdp_inference` in our server args checks.

python/sglang/multimodal_gen/runtime/distributed/parallel_state.py
BBuf: I tried it, but the warmup time didn't decrease.

python/sglang/multimodal_gen/runtime/models/dits/wanvideo.py
Edenzzzz: Got it, so in your experience, FSDP does not achieve good overlap?

BBuf: It might simply be because FSDP hasn't been heavily optimized for inference; it still focuses more on training optimizations.

mickqian (approved):

PR #15522, opened 2025-12-20
Optimize FP8 MLA KV cache writes with Triton kernel

### LINKED ISSUE #15104: [Perf][DeepSeekV3.2][Decode] Reduce small-kernel overhead in _get_topk_paged
### Background
This is a sub-task of the DeepSeek V3.2 optimization roadmap (#15025) and NSA optimization roadmap (#11989).
Goal: optimize decode-side `_get_topk_paged` where there are many small kernels / excessive kernel launches.

### Motivation
Profiling shows many small CUDA kernels around the TopK path, causing noticeable overhead in low-latency decode cases.

### Reproduction
Server:
python3 -m sglang.launch_server --model-path deepseek-ai/DeepSeek-V3.2 --trust-remote-code --tp-size 8 --dp-size 8 --enable-dp-attention --tool-call-parser deepseekv32 --reasoning-parser deepseek-v3

Bench (bs=1):
python3 -m sglang.bench_serving --model deepseek-ai/DeepSeek-V3.2 --dataset-name random --backend sglang-oai --random-range-ratio 1 --random-input-len 1200 --random-output-len 128 --max-concurrency 1 --num-prompts 5 --profile

### Environment
- SGLang commit:
- sgl-kernel / deepgemm / flashinfer versions:
- GPU: (e.g., B200 x8)
- CUDA / driver:
- TP/DP config:

### Profiling Evidence
- Profiler trace path:
- Top kernels / op stacks: (paste summary)
- Kernel launch count around `_get_topk_paged`:

### Proposed Approach
1) Reproduce and pinpoint hotspots via torch profiler / nsight
2) Reduce launches by caching / reusing metadata and consolidating tensor construction
3) Fuse where feasible (Triton/C++ op) with a safe fallback

### Acceptance Criteria
- Reduced kernel launches in the `_get_topk_paged` path
- Improved decode latency / throughput on bs=1 and bs=32 (attach numbers)
- No correctness regression (unit tests + serving sanity checks)

### Progress Updates
- Owner: @harvenstar 
- Timeline: target 1 to 2 weeks for a first PR + measurable improvement

harvenstar:
# Updated Summary

  Optimize FP8 MLA KV cache writes in NSA decode by:
  - skipping `torch.cat([k_nope, k_rope])` in the FP8 path, and
  - reusing the existing `set_mla_kv_buffer_triton` two-tensor write kernel.

  **No new Triton write kernel is added in the final version.**

  ## Motivation

  In DeepSeek-V3.2 decode, each step writes FP8 MLA KV cache per layer. The baseline path incurs noticeable CPU-side overhead when using PyTorch fancy indexing (`index_put_`) for small writes.

  ## Changes

  1. **Add `quantize_k_cache_separate(k_nope, k_rope)`** in `python/sglang/srt/layers/attention/nsa/quant_k_cache.py`.
     - Reuses existing fast quantization kernel (`_quantize_k_cache_fast`) without requiring an upstream concat.
     - Returns `(nope_part, rope_part)` as uint8 byte views, matching the MLA KV buffer layout.

  2. **Update FP8 NSA KV write path** in `python/sglang/srt/mem_cache/memory_pool.py`:
     - Remove upstream concat.
     - Reuse `set_mla_kv_buffer_triton(kv_buffer, loc, nope_part, rope_part)` for FP8 writes.

  3. **Cleanup:**
     - Remove the previously proposed FP8 scatter kernel and its env flag `SGLANG_USE_TRITON_FP8_SCATTER`.
     - Remove legacy tests and debug scripts for the old scatter path.

  ## Testing

  - **Byte-level equivalence**: `quantize_k_cache_separate` output matches the old concat+quantize path.
  - **End-to-end KV write** matches baseline behavior.
  - `pre-commit` passes.

 **Server:**
  ```bash
    --model-path deepseek-ai/DeepSeek-V3.2 \
    --trust-remote-code \
    --tp-size 8 \
    --dp-size 8 \
    --enable-dp-attention \
    --reasoning-parser deepseek-v3

  Benchmark:
    --dataset-name random \
    --random-input-len 1024 \
    --random-output-len 1024 \
    --random-range-ratio 0.98 \
    --num-prompts 80 \
    --max-concurrency 16
```
  ## Results

  | Metric             | Baseline (disabled) | Optimized (default) | Improvement |
  |--------------------|---------------------|---------------------|-------------|
  | Benchmark duration | 153.09s             | 150.03s             | -2.0%       |
  | Total throughput   | 1059.16 tok/s       | 1080.93 tok/s       | +2.0%       |

  ## GPU Kernel Change (from nsys profile):
  - Removed: at::native::index_elementwise_kernel<...index_put_kernel_impl...>
  - Added: set_mla_kv_buffer_kernel (Triton)

  ## Correctness Validation:
      --num-shots 8 \
      --num-questions 1319 \
      --parallel 1319
  - Accuracy: 0.951 (unchanged)
  - Invalid: 0.000

  ## Backward Compatibility

  - Default enabled: Users get performance improvement automatically
  - Fallback available: Set SGLANG_USE_TRITON_FP8_SCATTER=0 to disable
  - No API changes: Transparent to end users
  - Scope limited: Only affects MLA + FP8 KV cache path

  ### Related Issues

  Fixes #15104
  Related: #15025
   @Fridge003 
  ```

### CODE REVIEW

File: python/sglang/srt/layers/attention/nsa/quant_k_cache.py [hunk 1]
```diff
@@ -13,6 +13,113 @@ def quantize_k_cache(cache_k):
         return _quantize_k_cache_slow(cache_k)
 
 
+def quantize_k_cache_separate(
```
Fridge003, line 16: This implementation is not correct.
The quantization of k_nope and k_rope should be done in a single triton kernel.

You can refer to the implementation of `_quantize_k_cache_fast` kernel
The separate one just needs some simple refactors on the basis of this kernel

Fridge003 (changes requested):

test/manual/layers/attention/nsa/test_quantize_separate.py
```diff
@@ -0,0 +1,95 @@
+"""
```
Fridge003: The test can be put under the `if __name__ == "__main__":` part of `quant_k_cache.py`
We don't need a new manual test for this kernel

python/sglang/srt/layers/attention/nsa/quant_k_cache.py
harvenstar: @Fridge003 Thanks for the feedback. I've refactored the implementation:

  - Now uses a single Triton kernel by reusing the existing _quantize_k_cache_fast_kernel
  - Added _quantize_k_cache_fast_separate() that directly allocates two separate uint8 buffers
  - Avoids Python-side pack/view/split operations
  - Replaced all input validation asserts with explicit ValueError checks
  - Auto-fixes non-contiguous tensors

harvenstar: The new implementation passes typed views to the existing kernel to avoid code duplication. Sanity tests pass with byte-wise equality vs the concat path.

test/manual/layers/attention/nsa/test_quantize_separate.py
harvenstar: Done. Deleted the manual test file and moved the test to the __main__ section in quant_k_cache.py. See commits 5dec9a5c.

harvenstar:
## Updated Performance Results 
@Fridge003 
  I've run the latest benchmark and the results show **significant improvement**:

  ### Performance Metrics

  | Metric | Baseline (disabled) | Optimized (current) | Improvement |
  |--------|---------------------|---------------------|-------------|
  | Benchmark duration | 153.35s | **146.84s** | **-4.25%**  |
  | Total throughput | 1059.16 tok/s | **1104.20 tok/s** | **+4.25%**  |


  **This is a 2x improvement compared to the previous result **

  ### GPU Kernel Changes (from nsys profile)

  **Removed:**
  - `at::native::index_elementwise_kernel<...index_put_kernel_impl...>` (PyTorch native indexing)
  - `CatArrayBatchedCopy` kernel

  **Added:**
  - `set_mla_kv_buffer_kernel` (Triton) - our optimized single-kernel implementation

  The kernel consolidation successfully eliminated the overhead from separate indexing and concatenation operations, resulting in cleaner execution and better performance and the profiler results support this. 
image

python/sglang/srt/mem_cache/memory_pool.py
```diff
@@ -22,7 +22,10 @@
 from sglang.srt.configs.mamba_utils import BaseLinearStateParams
 from sglang.srt.environ import envs
 from sglang.srt.layers.attention.nsa import index_buf_accessor
-from sglang.srt.layers.attention.nsa.quant_k_cache import quantize_k_cache
+from sglang.srt.layers.attention.nsa.quant_k_cache import (
+    quantize_k_cache,
```
Fridge003: Not need to import line26

harvenstar:
## Bug Fix: Kernel Boundary Handling for FP8 Quantization

  ### Issue Found
  After applying the optimization, accuracy dropped from 0.8 to 0 in evaluation tests.

  ### Root Cause
  The `set_mla_kv_buffer_kernel` had a boundary bug when handling FP8 quantization:
  - The original two-branch logic assumed `nope_dim % BLOCK == 0`
  - This works for non-FP8 (nope_dim=512, BLOCK=128) 
  - FP8 has nope_dim=528, causing blocks to span nope/rope boundary 
  - When pid_blk=4 (base=512), the kernel would read wrong data with negative offsets

  ### Fix Applied
  Changed to three-way branching:
  1. Fast path: entire block in nope region (preserved)
  2. Fast path: entire block in rope region (preserved)  
  3. Boundary case: per-element masking for cross-boundary blocks

  ### Result
  -  Accuracy restored to baseline (~0.85)

Fridge003 (approved):

Issue #15527, opened 2025-12-20
[Bug] Incorrect output in PP mode when server_args.pp_async_batch_depth > 1


lynncui00:
### Checklist

- [x] I searched related issues but found no solution.
- [x] The bug persists in the latest version.
- [x] Issues without environment info and a minimal reproducible demo are hard to resolve and may receive no feedback.
- [x] If this is not a bug report but a general question, please start a discussion at https://github.com/sgl-project/sglang/discussions. Otherwise, it will be closed.
- [x] Please use English. Otherwise, it will be closed.

### Describe the bug

code link: https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/managers/scheduler_pp_mixin.py#L105

                if self.cur_batch:
                    server_is_idle = False
                    pp_proxy_tensors = self._pp_recv_proxy_tensors()
                next_pp_outputs = None
                next_batch_result = None
                d2h_event = None
                if self.server_args.pp_async_batch_depth > 0:
                    next_pp_outputs, next_batch_result, d2h_event = (
                        self._pp_commit_send_output_work_and_preprocess_output_tensors(
                            next_first_rank_mb_id,
                            next_mb_id,
                        )
                    )
                self._pp_commit_comm_work(self.send_proxy_work)
                if self.cur_batch:
                    result, self.launch_event = self._pp_launch_batch(
                        mb_id,
                        pp_proxy_tensors,
                        self.mb_metadata,
                        self.last_rank_comm_queue,
                    )

When the parameter server_args.pp_async_batch_depth is greater than 1, the output becomes completely incorrect. The reason is that the variable pp_proxy_tensors loses its reference during the run_batch process, causing PyTorch's gc mechanism to free the tensor, resulting in incorrect output.

After fixing this issue, we can move the _pp_commit_send_output_work_and_preprocess_output_tensors function to after run_batch to increase the overlap effect, thereby improving performance.

### Reproduction

python -m sglang.launch_server --model-path Qwen3-Coder-480B-A35B-Instruct-FP8 --disable-radix-cache --pp-size 2 --trust-remote --host 0.0.0.0 --port 8001 --mem-fraction-static 0.8  --tp-size 4 --pp-async-batch-depth 2 --max-running-requests 64

### Environment

Python: 3.12.8 | packaged by Anaconda, Inc. | (main, Dec 11 2024, 16:31:09) [GCC 11.2.0]
CUDA available: True
[...]

PR #15545, opened 2025-12-20
[fix] fix incorrect output in pp mode when server_args.pp_async_batch_depth > 1

qimcis:
Addresses #15527 

Summary:

To prevent premature freeing of PP proxy tensors, this change records them on the forward stream right after they're received; this keeps CUDA allocations alive while kernels are in flight when ```pp_async_batch_depth``` > 1.

The main change is that I moved ```_pp_commit_send_output_work_and_preprocess_output_tensors``` to after ```_pp_launch_batch``` in all PP event loops and call it unconditionally, so we keep a single ordering for all depths.

## Checklist

### DIFF
--- a/python/sglang/srt/managers/scheduler_pp_mixin.py
+++ b/python/sglang/srt/managers/scheduler_pp_mixin.py
@@ -92,6 +92,15 @@ class SchedulerPPMixin:
                 next_pp_outputs = None
                 next_batch_result = None
                 d2h_event = None
+                if self.server_args.pp_async_batch_depth > 0:
+                    (
+                        next_pp_outputs,
+                        next_batch_result,
+                        d2h_event,
+                    ) = self._pp_commit_send_output_work_and_preprocess_output_tensors(
+                        next_first_rank_mb_id,
+                        next_mb_id,
+                    )
                 self._pp_commit_comm_work(self.send_proxy_work)
                 if self.cur_batch:
                     result, self.launch_event = self._pp_launch_batch(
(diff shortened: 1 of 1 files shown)


qimcis:
@lynncui00 I just pushed a change reordering the PP event loops, where now for `pp_async_batch_depth > 0`, we now run `_pp_commit_send_output_work_and_preprocess_output_tensors` before `_pp_launch_batch`, so the NCCL recv/preprocess for the next microbatch should complete before we launch compute. For `pp_async_batch_depth == 0`, the ordering stays the same. This should restore the non-overlapped recv/compute behavior that avoids `nccl::recv` slowing `run_batch` (and performance degradations), lmk if it works out on your setup

lynncui00:
> @lynncui00 I just pushed a change reordering the PP event loops, where now for `pp_async_batch_depth > 0`, we now run `_pp_commit_send_output_work_and_preprocess_output_tensors` before `_pp_launch_batch`, so the NCCL recv/preprocess for the next microbatch should complete before we launch compute. For `pp_async_batch_depth == 0`, the ordering stays the same. This should restore the non-overlapped recv/compute behavior that avoids `nccl::recv` slowing `run_batch` (and performance degradations), lmk if it works out on your setup

Fixing this bug is all about moving _pp_commit_send_output_work_and_preprocess_output_tensors after _pp_launch_batch. Only this way can we truly eliminate the bubble and achieve an overlapping effect. The actual result is that the throughput increased from 2.64 to 3.15.

This PR itself is fine. I’m just discussing here whether there’s further room for improvement. If we can resolve the issue of nccl::recv slowing down _pp_launch_batchwhile maintaining the overlapping, the throughput could be further increased.

By the way, I’m not sure who wrote this pp async mode. I assume you were involved in the related work, so I’m discussing it here. If not, feel free to ignore this discussion.

Lastly, it’s essential to place _pp_commit_send_output_work_and_preprocess_output_tensors after _pp_launch_batch.

PR #15555, opened 2025-12-21
[Diffusion] Refactor attention backend checking to use backend enum

yeahdongcn:
## Motivation

While adapting `multimodal_gen` for MThreads GPUs, I noticed that it is not necessary to import the concrete attention backend implementations just to determine which backend is being used. Instead, the backend can be identified by its enum.

This approach is preferable because each platform is already responsible for determining which attention backends are available. Directly importing a specific backend in a given stage introduces unnecessary coupling and reduces portability across platforms.
## Modifications

Use `AttentionBackendEnum` to retrieve the backend enum and perform enum-based comparison instead of importing concrete backend implementations.

## Checklist

### DIFF
--- a/python/sglang/multimodal_gen/runtime/layers/attention/backends/aiter.py
+++ b/python/sglang/multimodal_gen/runtime/layers/attention/backends/aiter.py
@@ -11,6 +11,7 @@ from sglang.multimodal_gen.runtime.layers.attention.backends.attention_backend i
     AttentionMetadata,
     AttentionMetadataBuilder,
 )
+from sglang.multimodal_gen.runtime.platforms import AttentionBackendEnum
 
 
 class AITerBackend(AttentionBackend):
@@ -19,8 +20,8 @@ class AITerBackend(AttentionBackend):
     """
 
     @staticmethod
-    def get_name() -> str:
-        return "AITER"
+    def get_enum() -> AttentionBackendEnum:
+        return AttentionBackendEnum.AITER
 
     @staticmethod
     def get_impl_cls() -> type["AITerImpl"]:
--- a/python/sglang/multimodal_gen/runtime/pipelines_core/stages/denoising.py
+++ b/python/sglang/multimodal_gen/runtime/pipelines_core/stages/denoising.py
@@ -38,13 +38,6 @@ from sglang.multimodal_gen.runtime.distributed.parallel_state import (
     get_cfg_group,
     get_classifier_free_guidance_rank,
 )
-
-try:
-    from sglang.multimodal_gen.runtime.layers.attention.backends.flash_attn import (
-        FlashAttentionBackend,
-    )
-except ImportError:
-    FlashAttentionBackend = None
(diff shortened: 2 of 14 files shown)


python/sglang/multimodal_gen/configs/models/dits/base.py
```diff
@@ -27,7 +27,7 @@ class DiTArchConfig(ArchConfig):
         default_factory=lambda: {
             AttentionBackendEnum.SLIDING_TILE_ATTN,
             AttentionBackendEnum.SAGE_ATTN,
-            AttentionBackendEnum.FA,
+            AttentionBackendEnum.FLASH_ATTN,
```
mickqian: we'll leave it, to be consistent with `srt`

python/sglang/multimodal_gen/runtime/layers/attention/backends/aiter.py
```diff
@@ -20,7 +21,7 @@ class AITerBackend(AttentionBackend):
 
     @staticmethod
     def get_name() -> str:
-        return "AITER"
+        return str(AttentionBackendEnum.AITER)
```
mickqian: `AttentionBackendEnum.AITER.name` is better

python/sglang/multimodal_gen/runtime/pipelines_core/stages/denoising.py
```diff
@@ -563,7 +531,7 @@ def _prepare_denoising_loop(self, batch: Req, server_args: ServerArgs):
             ]
 
         # Prepare STA parameters
-        if st_attn_available and self.attn_backend == SlidingTileAttentionBackend:
+        if self.attn_backend.get_name() == str(AttentionBackendEnum.SLIDING_TILE_ATTN):
```
mickqian: should we add a `get_enum()`?

mickqian (changes requested):

python/sglang/multimodal_gen/configs/models/dits/base.py
yeahdongcn: No problem.

python/sglang/multimodal_gen/runtime/layers/attention/backends/aiter.py
yeahdongcn: Okay. Let me update that way.

python/sglang/multimodal_gen/runtime/pipelines_core/stages/denoising.py
yeahdongcn: Sure.

yeahdongcn: I just realized that we could replace `get_name()` with `get_enum()`, since `get_name()` is currently only used to retrieve the enum in the selector. Using `get_enum().name` would directly give us the attention backend name and simplify the logic.

mickqian (approved):

stmatengss:
Great feature! Looking forward to it.

PR #15559, opened 2025-12-21
Add MTP speculative decoding support for Triton attention backend for Mimo-v2-Flash

chriswritescode-dev:
The PR adds Triton backend support for MTP speculative decoding, which is important because:                                   
     - Triton works on all GPUs (including Ampere, Hopper, and Blackwell)                                                           
     - Previously MTP was hardcoded to use FlashAttentionBackend (FA3)                                                              
     - Now users on non-Hopper hardware or those preferring Triton can use MTP speculative decoding  
      - Adds `DRAFT_EXTEND_V2` mode support to the Triton attention backend for MTP speculative decoding
      - Uses `DraftBackendFactory` in `mtp_worker_v2.py` to allow backend flexibility 
      -  instead of hardcoding FlashAttention
      - Test system 4x 6000 Pro Blackwell workstation cards

### DIFF
--- a/python/sglang/srt/layers/attention/triton_backend.py
+++ b/python/sglang/srt/layers/attention/triton_backend.py
@@ -173,6 +173,7 @@ class TritonAttnBackend(AttentionBackend):
         self.forward_metadata: ForwardMetadata = None
 
         self.cuda_graph_custom_mask = None
+        self.cuda_graph_kv_indices = None
 
     def get_num_kv_splits(
         self,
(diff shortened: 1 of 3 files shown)


python/sglang/srt/layers/attention/triton_backend.py
```diff
@@ -624,6 +625,62 @@ def init_forward_metadata_capture_cuda_graph(
...
+                bs * num_tokens_per_bs + 1,
+                step=num_tokens_per_bs,
+                dtype=torch.int32,
+                device=self.device,
+            )
+            kv_indptr = self.kv_indptr[: bs + 1]
+            kv_indptr[1 : bs + 1] = torch.cumsum(seq_lens, dim=0)
+
+            # Ensure cuda_graph_kv_indices exists (set by init_cuda_graph_state)
+            if self.cuda_graph_kv_indices is None:
```
acelyc111, line 650: Is it needed to judge it also before line 618?

acelyc111 (approved):

acelyc111:
@hnyls2002 Please take a look

hnyls2002:
@chriswritescode-dev thanks for the PR. The Triton attention backend on main now supports the spec-v2 MTP paths this adds, including draft-extend-v2 and target-verify CUDA-graph metadata in triton_backend.py (see #26651 and #27545 which maintain these paths), so MTP with the Triton backend works without this patch. Closing as superseded - please reopen if something is still missing on current main.

PR #15601, opened 2025-12-22
Fix BatchMLAPagedAttentionWrapper query/qo_inptr mismatch for EAGLE

wenscarl:
## Motivation
When the batch dimension of the query does not match `qo_indptr[-1]` in `BatchMLAPagedAttentionWrapper` (as described in FlashInfer issue #[2236](https://github.com/flashinfer-ai/flashinfer/issues/2236)), an invalid memory access can occur as below. 
```
  File "/workspace/build/aot/generated/batch_mla_attention_dtype_q_bf16_dtype_kv_bf16_dtype_o_bf16_dtype_idx_i32_head_dim_ckv_512_head_dim_kpe_64_profiler_False/batch_mla_run.cu", line 123, in BatchMLAPagedAttentionRun(tvm::ffi::TensorView, tvm::ffi::TensorView, tvm::ffi::Array<long int>, tvm::ffi::TensorView, tvm::ffi::TensorView, tvm::ffi::TensorView, tvm::ffi::TensorView, tvm::ffi::TensorView, tvm::ffi::TensorView, tvm::ffi::Optional<tvm::ffi::TensorView>, int64_t, int64_t, int64_t, double, bool)::<lambda()>
RuntimeError: Check failed: (status == cudaSuccess) is false: Failed to run MLA, error: an illegal memory access was encountered

```

The root cause is that the forward_metadata—including `qo_indptr`—is not updated in the TRTLLM MLA backend, leading to inconsistent metadata being used during execution.

## Benchmarking and Profiling

```
CUDA_LAUNCH_BLOCKING=1 \
SGLANG_MOE_NVFP4_DISPATCH=1 \
HF_TOKEN=<your token> \
  --model-path nvidia/DeepSeek-R1-0528-FP4-v2 \
  --trust-remote-code \
  --kv-cache-dtype fp8_e4m3 \
  --attention-backend trtllm_mla \
  --quantization modelopt_fp4 \
  --moe-runner-backend flashinfer_cutlass \
  --disable-radix-cache \
  --disable-chunked-prefix-cache \
  --disable-shared-experts-fusion \
  --disable-cuda-graph \
  --stream-interval 50 \
  --decode-log-interval 1000 \
  --watchdog-timeout 1000000 \
  --context-length 2176 \
  --max-total-tokens 131072 \
  --max-prefill-tokens 32768 \
  --chunked-prefill-size 65536 \
  --max-running-requests 30000 \
  --mem-fraction-static 0.84 \
  --enable-dp-attention \
  --tp-size 4 \
  --dp-size 4 \
  --ep-size 4 \
  --eplb-algorithm deepseek \
  --speculative-num-steps 1 \
  --speculative-eagle-topk 1 \
  --speculative-num-draft-tokens 2

client:
```

The flashinfer attn backend:
```
Accuracy: 0.961
Invalid: 0.000
Latency: 132.733s
Output throughput: 1099.874 token/s
```
trtllm_mla attn backend:
```
Accuracy: 0.959
Invalid: 0.000
Latency: 130.433 s
Output throughput: 1133.730 token/s

```

## Checklist


python/sglang/srt/layers/attention/trtllm_mla_backend.py
```diff
@@ -658,6 +658,11 @@ def init_forward_metadata(self, forward_batch: ForwardBatch):
             self.forward_decode_metadata.batch_size = bs
 
             forward_batch.decode_trtllm_mla_metadata = self.forward_decode_metadata
+            if (
```
Fridge003: Can we combine this condition with the else branch below (line 666-668)?

python/sglang/srt/layers/attention/trtllm_mla_backend.py
```diff
@@ -955,6 +955,10 @@ def forward_extend(
         if (
             self.forward_prefill_metadata is not None
             and self.forward_prefill_metadata.fallback_to_flashinfer_impl
+            and not (
```
Fridge003: According the codes here
https://github.com/sgl-project/sglang/blob/f9808ae5422d85d9714d84052758efeb1d229984/python/sglang/srt/layers/attention/trtllm_mla_backend.py#L587

`forward_prefill_metadata` should not be set when the forward mode is target verify or draft extend...
Maybe the bug is elsewhere?

Fridge003 (changes requested):

python/sglang/srt/layers/attention/trtllm_mla_backend.py
wenscarl: From the symptom, when the forward mode is `target_verify` or `draft_extend`, it's possible to fall into the super class's [forward_extend](https://github.com/sgl-project/sglang/blob/f9808ae5422d85d9714d84052758efeb1d229984/python/sglang/srt/layers/attention/trtllm_mla_backend.py#L963). From [here](https://github.com/sgl-project/sglang/blob/f9808ae5422d85d9714d84052758efeb1d229984/python/sglang/srt/layers/attention/trtllm_mla_backend.py#L611-L613), the `forward_prefill_metadata` shouldn't be set for `target_verify` or `draft_extend`.

Fridge003: That's true. So why target_verify/draft_extend batches don't terminate early at 
`self.forward_prefill_metadata is not None`

wenscarl: In [this ](https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/layers/attention/trtllm_mla_backend.py#L615)branch, the following assertion actually failed,
```
assert self.forward_prefill_metadata is None
```
which indicates that `self.forward_prefill_metadata is not None`  is possible for `target_verify` or `draft_extend`.

nvpohanh: @Fridge003 do you think @wenscarl 's reason makes sense? Or do you think we should dive deeper into your question?

Fridge003: @nvpohanh
Will this PR be useful for solving the out-of-boundary bug? If so we can merge it temporarily, and come up with some better solution in the future

nvpohanh: @Fridge003 Sorry for confusion. This is NOT related to the index out-of-bound issue. This is for a completely unrelated issue. I was just asking if we have come to an agreement here or whether this PR still need more discussion. Thanks!

python/sglang/srt/layers/attention/flashinfer_mla_backend.py
```diff
@@ -571,6 +571,22 @@ def forward_extend(
                 logits_soft_cap=logits_soft_cap,
             )
         else:
+            if (
+                hasattr(prefill_wrapper_paged, "_qo_indptr_buf")
+                and prefill_wrapper_paged._qo_indptr_buf is not None
+            ):
+                planned_total_queries = prefill_wrapper_paged._qo_indptr_buf[-1].item()
```
Fridge003: Will this operation cause extra sync (since there is an item() here)
Looks like this branch is only for sanity check, so we might remove it?

wenscarl: Yes. Removed.

Fridge003 (approved):

Fridge003 (approved):

PR #15164, opened 2025-12-15
add decode round robin policy

Hexq0210:
## Motivation

In the PD separation scenario, based on the DeepSeek-R1 model, we find that the load requests of each DP channel of the Decode instance are unbalanced during the pressure test. (The following figure shows the pressure test process.)
2C3B87F5-CDB0-43B7-D25C-A3A78AAFD1B7

To solve this problem, the roundrobin scheduling policy is added for the Decode instance in the PD separation scenario. The test result is as follows:
3150DA89-8700-4E95-DDFA-96BFC8FF0CF1

## Modifications

Add the decode_round_robin policy to enable the roundrobin policy of the decode instance.

## Accuracy Tests

NA

## Checklist

- [ ] Format your code according to the [Format code with pre-commit](https://docs.sglang.io/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.io/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.io/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.io/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.io/developer_guide/contribution_guide.html#benchmark-the-speed).
- [ ] Follow the SGLang code style [guidance](https://docs.sglang.io/developer_guide/contribution_guide.html#code-style-guidance).
- [ ] Work with maintainers to merge your PR. See the [PR Merge Process](https://github.com/sgl-project/sglang/blob/main/.github/MAINTAINER.md#pull-request-merge-process)

PR #15617, opened 2025-12-22
[Fix] Set DataParallelController to round robin per dp rank in PD.

miter6:
## Motivation
image

Bootstrap_room info is generated using random function in sgl_router.  
DataparallelController use it as round-robin index.   
This will dispatch the request randomly not round-robin in both prefill and decode instance.  

It seems this patch doesn't work anymore.  
https://github.com/sgl-project/sglang/pull/15164 
## Checklist


thefacetakt:
I think this is the wrong way to handle the issue -- decode worker should know dp-rank of prefill worker in order to be able to fetch kv caches. Proposed solution may cause `round_robin_counter` on prefill and decode instances to diverge.

I believe that the right way is to use `--dp-aware` option of sgl_router. This way router will pass `data_parallel_rank` to both prefill and decode, and it will affect worker selection (`maybe_external_dp_rank_routing` function).

And the router will change `data_parallel_rank` in a round-robin-way.

hnyls2002:
@miter6 thanks for the PR. The DP controller load balancing was reworked on main: round_robin_scheduler no longer special-cases PD mode, and bootstrap-room-following is now its own FOLLOW_BOOTSTRAP_ROOM load-balance method (#16110, #16258), which covers this use case. Closing as superseded - please reopen if I've missed something.

PR #13425, opened 2025-11-17
feat: Naive support Spec V2 + Constrained Decoding

Ubospica:
This PR supports enabling constrained decoding and Speculative decoding v2 at the same time, resolving #13019.

When the batch contains grammars to handle, the overlap is temporarily disabled.

Further TODO:

- [ ] Split the forward launch into two phases: before verify sampling and after verify sampling
- [ ] Apply the delayed launch of phase-2, which requires
  - Launch phase-2 after the last batch's result is processed (grammar is synced)
  - Sync the draft token ids to the CPU before launching phase-2 or make mask preparation on the GPU.

Signed-off-by: Ubospica <ubospica@gmail.com>

cc @merrymercy @hnyls2002 @jiapingW 

Output:
```
result {'text': '{"name": "John", "age": 30}', 'output_ids': [6377, 978, 1115, 376, 11639, 613, 376, 482, 1115, 29871, 29941, 29900, 29913, 2], 'meta_info': {'id': '05c466dca22d4397b91b31c8f7422aaf', 'finish_reason': {'type': 'stop', 'matched': 2}, 'prompt_tokens': 58, 'weight_version': 'default', 'total_retractions': 0, 'completion_tokens': 14, 'cached_tokens': 0, 'spec_accept_rate': 0.45, 'spec_accept_length': 3.5, 'spec_verify_ct': 4, 'spec_accept_token_num': 9, 'spec_draft_token_num': 20, 'e2e_latency': 0.6285512447357178, 'response_sent_to_client_ts': 1764062075.5853202}}
[...]

Issue #13019, opened 2025-11-10
XGrammar support for Spec V2


JustinTong0323:
Required when `tool_choice="required"`

jiapingW:
> [@hnyls2002](https://github.com/hnyls2002) [@JustinTong0323](https://github.com/JustinTong0323) need to look deeper but i assume it's because there's no grammar backend reference in v2 file https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/speculative/eagle_info_v2.py unlike v1 file https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/speculative/eagle_info.py#L9

I debugged the program and found that since Spec v2 does not introduce any grammar-related operations, it does not perform the relevant processing. The `batch.spec_info.grammar` is None.

pdasgup:
thanks @jiapingW . will you be raising the fix for this?

hnyls2002:
Adapt these two logics from worker v1 to worker v2

- https://github.com/sgl-project/sglang/blob/85b8c5c4cde2b3da7d63fba10e657b79b11484ab/python/sglang/srt/speculative/eagle_worker.py#L670-L687
- https://github.com/sgl-project/sglang/blob/85b8c5c4cde2b3da7d63fba10e657b79b11484ab/python/sglang/srt/speculative/eagle_worker.py#L653-L658

jiapingW:
I'm modifying its logic. In v1, the batch's `spec_info` included grammar information, but in v2, it's not being retrieved here: https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/speculative/eagle_worker_v2.py#L587. It seems this parameter wasn't initialized in a previous initialization step.

jiapingW:
I implemented a similar solution using the patch below, which does use grammar and handle logits, but issues such as duplicate output still exist and I'm still working on it.
```python

diff --git a/python/sglang/srt/speculative/eagle_info_v2.py b/python/sglang/srt/speculative/eagle_info_v2.py
index 8d334a460..cd886e18a 100644
--- a/python/sglang/srt/speculative/eagle_info_v2.py
+++ b/python/sglang/srt/speculative/eagle_info_v2.py
@@ -1,13 +1,14 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Optional
 
 import torch
 import torch.nn.functional as F
 import triton
[...]

Issue #11762, opened 2025-10-17
[Feature] Overlap Spec Support


hnyls2002:
### Motivation

We have already implemented the initial support for eagle speculative decoding with the overlap scheduler, and here is the roadmap for more feature optimizations and support.

***

### page size & topk support

- [ ] Support page size > 1 @cicirori @hnyls2002 
- [ ] Support topk > 1 
- [ ] Support topk > 1 + page size > 1

### memory allocation

- [ ] over-allocation optimization @hnyls2002 
- [ ] over-allocation with page size > 1 + topk > 1

### Attention backend support

- [ ] Remove or make `verify_done.synchronize()` an option @hnyls2002 
- [ ] Different attention backend support @Fridge003 

### sampling

- [ ] xgrammar support
- [ ] penalty support 
- [ ] logprob support

### speculative methods

- [x] new speculative model worker interface (https://github.com/sgl-project/sglang/pull/11643)
- [ ] standalone speculative support @Qiaolin-Yu 
- [ ] ngram speculative support @a4zhangfei 
- [ ] Top `SpecTpWorker` for all speculative decoding backends @hnyls2002 
- [ ] Make `SpecTpWorker` compatible with all `TpModelWorker` features.

### dp attention support

- [ ] Support idle batch @ch-wan 
- [ ] cover testcases with dp-attention + overlap + spec @ch-wan 

### EP support

- [ ] Check compatibility with DeepEP / EP @fzyzcjy 
- [ ] Cover testcases with EP + overlap + spec @fzyzcjy 

### PD disaggregation

- [ ] Event loop adjust in Prefill / Decode worker @ShangmingCai @ByronHsu 
- [ ] Cover testcases with PD-Disagg + overlap + spec @ShangmingCai @ByronHsu 

### Aggressive Optimizations

- [ ] Enable a separate `plan_stream`

vincentzed:
I impl `Support topk > 1 + page size > 1`

hnyls2002:
@vincentzed Great, added you to the roadmap. Please try to implement the topk > 1 first, which is much easier than topk > 1 + page size > 1. Also, for the current main branch without overlap, our implementation of topk > 1 + page size > 1 is wrong; you can look at this [PR](https://github.com/sgl-project/sglang/pull/7725) (which may still be wrong, but can give you some hints) about the issue.

lixuwei2333:
[...]

PR #15623, opened 2025-12-22
feat: Full Support for Overlapped Constrained Decoding + Spec V2

Ubospica:
This PR supports fully overlapped constrained decoding with Spec V2.

See #13425 for the non-overlapped version. See #11762, #13019 for more background. This PR is on top of #15465.

## Overlapping Pattern

image

 

## Benchmark

E2E Latency:

| Test Case | Baseline | Optimized | Speedup |
|-----------|-------------------|---------------------|---------|
| test_json_generate | 0.644s | 0.609s | **1.06x** |
| test_mix_json_and_other (batch JSON) | 0.17-0.21s | 0.12-0.16s | **~1.3x** |
| test_regex_generate_complex_json | 0.175s | 0.164s | **1.07x** |
| test_regex_generate_custom_log_format | 0.19-0.26s | 0.18-0.23s | **~1.1x** |
| test_regex_generate_email | 0.07s | 0.04-0.06s | **~1.2x** |
| test_regex_generate_phone | 0.08-0.11s | 0.07-0.08s | **~1.2x** |

### Key Findings

1. **Overall Speedup: 5%-30%** reduction in e2e_latency across most test cases
2. **Best improvement** observed in batch JSON constrained generation (`test_mix_json_and_other`): **~30% faster**
3. **Correctness verified**: All 12 tests pass with identical output formats

Signed-off-by: Ubospica <ubospica@gmail.com>

## Checklist

### DIFF
--- a/benchmark/speculative/run_bench_traverse_tree.sh
+++ b/benchmark/speculative/run_bench_traverse_tree.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+SCRIPT_DIR=$(dirname "$0")
+
+python "$SCRIPT_DIR/bench_traverse_tree.py" --spec-steps 1 --spec-topk 4 --spec-draft-tokens 4
+python "$SCRIPT_DIR/bench_traverse_tree.py" --spec-steps 5 --spec-topk 4 --spec-draft-tokens 16
+python "$SCRIPT_DIR/bench_traverse_tree.py" --spec-steps 5 --spec-topk 4 --spec-draft-tokens 64
--- a/python/sglang/srt/managers/schedule_batch.py
+++ b/python/sglang/srt/managers/schedule_batch.py
@@ -92,6 +92,7 @@ from sglang.srt.utils.cuda_ipc_transport_utils import CudaIpcTensorTransportProx
 
 if TYPE_CHECKING:
     from sglang.srt.configs.model_config import ModelConfig
+    from sglang.srt.managers.utils import GenerationBatchResult
     from sglang.srt.speculative.eagle_info import EagleDraftInput
     from sglang.srt.speculative.spec_info import SpecInput, SpeculativeAlgorithm
 
@@ -1260,8 +1261,8 @@ class ScheduleBatch(ScheduleBatchDisaggregationDecodeMixin):
     spec_info: Optional[SpecInput] = None
 
     # Pending accept info from last batch for overlapped grammar processing
-    # list of (request, accepted_token_ids)
-    pending_accept_info: Optional[List[Tuple[Req, List[int]]]] = None
+    # Tuple of (last_batch, last_result)
+    pending_accept_info: Optional[Tuple[ScheduleBatch, GenerationBatchResult]] = None
 
     # Whether to return hidden states
     return_hidden_states: bool = False
@@ -2278,8 +2279,8 @@ class ModelWorkerBatch:
     reqs: Optional[List[Req]] = None
     has_grammar: bool = False
     # Pending accept info from last batch for overlapped grammar processing
-    # Format: List[Tuple[Req, List[int]]] - list of (request, accepted_token_ids)
-    pending_accept_info: Optional[List] = None
+    # Tuple of (last_batch, last_result)
+    pending_accept_info: Optional[Tuple[ScheduleBatch, GenerationBatchResult]] = None
 
     # For hidden states before normal
     return_hidden_states_before_norm: bool = False
--- a/python/sglang/srt/managers/schedule_batch.py
+++ b/python/sglang/srt/managers/schedule_batch.py
@@ -1259,6 +1259,10 @@ class ScheduleBatch(ScheduleBatchDisaggregationDecodeMixin):
     # spec_info: Optional[SpecInput] = None
     spec_info: Optional[SpecInput] = None
 
+    # Pending accept info from last batch for overlapped grammar processing
+    # list of (request, accepted_token_ids)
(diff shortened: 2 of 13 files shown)


python/sglang/srt/managers/scheduler.py
```diff
@@ -1130,6 +1130,22 @@ def pop_and_process():
             if disable_overlap_for_batch:
                 pop_and_process()
 
+            # Prepare pending accept info for overlapped grammar processing
+            # The accept tokens from the last batch will be processed during
+            # the current batch's verify forward to overlap CPU and GPU operations.
+            if (
+                batch
+                and batch.is_eagle_v2
```
acelyc111, line 1138: AttributeError: 'ScheduleBatch' object has no attribute 'is_eagle_v2'

acelyc111:
Also consider to adapt to python/sglang/srt/speculative/multi_layer_eagle_worker_v2.py ?

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -2165,6 +2172,7 @@ def copy(self):
             is_prefill_only=self.is_prefill_only,
             seq_lens_cpu=self.seq_lens_cpu,
             enable_overlap=self.enable_overlap,
+            has_grammar=self.has_grammar,
```
hanming-lu: should we copy `pending_accept_info` here as well?

benchmark/speculative/run_bench_traverse_tree.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
hanming-lu, line 1: maybe we don't need this file?

python/sglang/srt/speculative/eagle_worker_v2.py
```diff
@@ -609,6 +615,59 @@ def target_worker(self):
     def draft_worker(self):
         return self._draft_worker
 
+    def _init_grammar_pinned_buffers(self, bs: int, num_draft_tokens: int):
```
hanming-lu, line 618: what's the reason we want to lazy init here? seems like non-lazy init is fine?

python/sglang/srt/speculative/eagle_worker_v2.py
```diff
@@ -609,6 +615,59 @@ def target_worker(self):
...
+        )
+        retrieve_next_sibling_cpu.copy_(
+            verify_input.retrive_next_sibling, non_blocking=True
+        )
+        draft_tokens_cpu.copy_(
+            verify_input.draft_token.view(bs, num_draft_tokens), non_blocking=True
+        )
+
+        # Record event to synchronize before using CPU tensors
+        copy_event = torch.cuda.Event()
```
hanming-lu, line 661: `grammar_copy_done` for clarity and consistency since we have a `copy_done` and `verify_done`

python/sglang/srt/speculative/eagle_worker_v2.py
```diff
@@ -763,6 +833,63 @@ def verify(self, batch: ModelWorkerBatch):
...
+                        i * stride : i * stride + accept_lens[i]
+                    ]
+                    try:
+                        for token_id in accepted_tokens:
+                            req.grammar.accept_token(token_id)
+                    except ValueError as e:
+                        logger.error(
+                            f"Grammar accept_token failed for req {req.rid} "
+                            f"with tokens {accepted_tokens}: {e}"
+                        )
```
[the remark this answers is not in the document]
hanming-lu: the suggestion looks not bad

python/sglang/srt/managers/scheduler.py
```diff
@@ -1130,6 +1130,22 @@ def pop_and_process():
...
+                and batch.has_grammar
+                and batch.forward_mode.is_decode()
+                and len(self.result_queue) > 0
+            ):
+                last_batch, last_result = self.result_queue[-1]
+                if last_batch.has_grammar:
+                    batch.pending_accept_info = (last_batch, last_result)
+                    # Mark that grammar accept will be processed in the next batch's verify
+                    last_result.grammar_accept_processed = True
```
hanming-lu, line 1147: I am trying to understand why we need this `grammar_accept_processed` flag. for spec v2 + decode + not disable_overlap_for_batch, all batches will have `grammar_accept_processed = True`, and all other cases `grammar_accept_processed = False`.

In the output processing, can we just check the conditions above instead?

hanming-lu: Also, IIUC, `grammar_accept_processed` only works for decode batch. How do you check confirm that for the last_batch? 

I see you check `batch.forward_mode.is_decode()`, but it's checking if batch N being decode, last_batch is batch N-1, how can we infer batch N-1 from batch N?

hanming-lu: I think I got the reason now, the batch N's verify handles batch N-1's grammar_accept_processed

python/sglang/srt/managers/schedule_batch.py
hanming-lu, line 2175: do we need to copy `pending_accept_info` here as well?

python/sglang/srt/speculative/spec_utils.py
Paiiiiiiiiiiiiii, line 697: `allocate_token_bitmask` resides in host-side pageable memory; consequently, the line `vocab_mask = vocab_mask.to(verify_input.retrive_next_token.device, non_blocking=True)` fails to execute asynchronously, causing the CPU to block at this point.

PR #15648, opened 2025-12-22
Tiny fix inconsistent probs processing between FlashInfer sampling paths leads to distribution drift

narutolhy:
## Motivation

In the flashinfer sampling backend, when min_p_sampling is enabled together with top_k and/or top_p, the current code path applies top_k_renorm_prob and top_p_renorm_prob directly to the input probs tensor.

These renormalization functions are destructive, which causes the original probability distribution to be modified unintentionally. As a result:

The same probs tensor is reused for both min-p thresholding and sampling, but with altered semantics

Multiple renormalization steps are applied implicitly, leading to incorrect probability distributions

Sampling behavior becomes inconsistent with other backends and reference implementations

This PR fixes the issue by ensuring that top_k / top_p renormalization does not mutate the original probs, and that each probability tensor has a clear, well-defined role.

https://github.com/sgl-project/sglang/issues/15046

## Modifications

Preserve the original probs tensor and avoid in-place modification

Introduce a separate sampling_probs tensor used exclusively for sampling

Apply top_k_renorm_prob and top_p_renorm_prob only on sampling_probs

Keep the original probs unchanged for:

min_p threshold checks

Clarify the separation between probability filtering logic and sampling distribution

## Accuracy Tests

PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -q sgl-kernel/tests/test_sampling.py 
150 passed, 3 skipped in 25.84s

## Checklist

### DIFF
--- a/python/sglang/srt/layers/sampler.py
+++ b/python/sglang/srt/layers/sampler.py
@@ -210,10 +210,12 @@ class Sampler(nn.Module):
                     sampling_info.sampling_seed is None
                 ), "Sampling seed is not supported for flashinfer backend"
                 if sampling_info.need_min_p_sampling:
-                    probs = top_k_renorm_prob(probs, sampling_info.top_ks)
-                    probs = top_p_renorm_prob(probs, sampling_info.top_ps)
+                    sampling_probs = top_k_renorm_prob(probs, sampling_info.top_ks)
+                    sampling_probs = top_p_renorm_prob(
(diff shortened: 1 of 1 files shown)


Qiaolin-Yu (approved):
LGTM

Qiaolin-Yu (commented):
It seems #17885 has solved this issue, so we may not merge this. But thanks for the contribution!

Issue #12987, opened 2025-11-10
diffusion: accuracy and performance test


mickqian:
@adarshxs

PR #15649, opened 2025-12-22
diffusion(tests): Add per-test sampling overrides

qimcis:
## Summary
Enabled per-test sampling overrides for diffusion video runs so we can exercise resolution/steps/FPS controls via the `/videos` API without changing global defaults. Allows for some tests to be run with lower resolution/steps/FPS, to increase speed.

Aligns with the TODO points described in #12987 (Reducing ```num_inference_steps``` to shorten testing time, Lower the video/image resolution to reduce testing time.)

## Modifications
- Added optional override fields (`num_inference_steps`, `height`, `width`, `guidance_scale`) to `DiffusionSamplingParams`.
- Added a helper to build `extra_body` for `/videos` calls and wired it through all video generation paths so `steps/height/width/num_frames/fps/guidance_scale` are sent alongside `size` (including reference URL / input-reference flows).
- Added an env override that let you opt into lower steps/res for faster runs without changing the baseline configs; the alternative would be to hardcode reductions into the test cases or add separate fast cases, wasn't sure if this is what was what we were looking for, lmk if it is though.

## Checklist

### DIFF
--- a/python/sglang/multimodal_gen/test/server/test_server_utils.py
+++ b/python/sglang/multimodal_gen/test/server/test_server_utils.py
@@ -649,8 +649,13 @@ def get_generate_fn(
     sampling_params: DiffusionSamplingParams,
 ) -> Callable[[str, Client], str]:
     """Return appropriate generation function for the case."""
+
+    # Pick size from explicit width/height when provided; otherwise use configured output_size.
     # Allow override via environment variable (useful for AMD where large resolutions cause slow VAE)
-    output_size = os.environ.get("SGLANG_TEST_OUTPUT_SIZE", sampling_params.output_size)
+    if sampling_params.width is not None and sampling_params.height is not None:
+        output_size = f"{sampling_params.width}x{sampling_params.height}"
+    else:
+        output_size = os.environ.get("SGLANG_TEST_OUTPUT_SIZE", sampling_params.output_size)
     n = sampling_params.num_outputs_per_prompt
 
     def _create_and_download_video(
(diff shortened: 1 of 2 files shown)


python/sglang/multimodal_gen/test/server/test_server_utils.py
```diff
@@ -768,6 +768,50 @@ def _create_and_download_video(
...
+        """Compose extra_body for video generation with optional overrides."""
+        body: dict[str, Any] = {}
+        steps_override = _get_int_env("SGLANG_TEST_VIDEO_STEPS")
+        height_override = _get_int_env("SGLANG_TEST_VIDEO_HEIGHT")
+        width_override = _get_int_env("SGLANG_TEST_VIDEO_WIDTH")
+
+        if steps_override is not None:
+            body["steps"] = steps_override
+        elif sampling_params.num_inference_steps is not None:
+            body["steps"] = sampling_params.num_inference_steps
```
Prozac614: You read sampling_params.num_inference_steps but write it as body["steps"]. The server expects "num_inference_steps" though — was the rename intentional?btw, VideoGenerationsRequest doesn't have a steps field. Did you actually test this works end-to-end?

python/sglang/multimodal_gen/test/server/testcase_configs.py
```diff
@@ -547,7 +576,12 @@ def from_req_perf_record(
             modality="video",
             custom_validator="video",
         ),
-        TI2V_sampling_params,
+        replace(
+            TI2V_sampling_params,
+            width=832,
```
mickqian: could we avoid these three lines from duplicating?

qimcis: yes! done

PR #15693, opened 2025-12-23
SGLang Tracing: fix attribute errors (header extraction & bootstrap span closing) 

vladnosiv:
## Motivation

1. Header extraction

The `external_trace_header` parameter is marked as optional:
https://github.com/sgl-project/sglang/blob/6a5764a71937d135f22c1dd025f743693783d31c/python/sglang/srt/tracing/trace.py#L438

However, it is passed to extract as it is:
https://github.com/sgl-project/sglang/blob/6a5764a71937d135f22c1dd025f743693783d31c/python/sglang/srt/tracing/trace.py#L465-L467

Which causes the request to crash:
```
  File "/sgl-workspace/sglang/python/sglang/srt/managers/tokenizer_manager.py", line 2150, in _trace_request_start
    trace_req_start(
  File "/sgl-workspace/sglang/python/sglang/srt/tracing/trace.py", line 465, in trace_req_start
    external_trace_context = _trace_context_propagator.extract(
  File "/home/dynamo/.local/lib/python3.10/site-packages/opentelemetry/trace/propagation/tracecontext.py", line 49, in extract
    header = getter.get(carrier, self._TRACEPARENT_HEADER_NAME)
  File "/home/dynamo/.local/lib/python3.10/site-packages/opentelemetry/propagators/textmap.py", line 87, in get
    val = carrier.get(key, None)
AttributeError: 'NoneType' object has no attribute 'get'
```

2. Bootstrap span closing

Here the code always tries to close the bootstrap room span:
https://github.com/sgl-project/sglang/blob/6a5764a71937d135f22c1dd025f743693783d31c/python/sglang/srt/tracing/trace.py#L535-L538

Which in my tests crashes in the case of an aggregated setup:
```
File "/sgl-workspace/sglang/python/sglang/srt/tracing/trace.py", line 538, in trace_req_finish
  req_context.bootstrap_room_span.end(end_time=ts)
AttributeError: 'NoneType' object has no attribute 'end'
```


ishandhanani (approved):
LGTM but I'll let @sufeng-buaa take a quick look just in case

python/sglang/srt/tracing/trace.py
```diff
@@ -463,7 +463,7 @@ def trace_req_start(
     if str(bootstrap_room) not in remote_trace_contexts:
         attrs = {"bootstrap_room": str(hex(bootstrap_room))}
         external_trace_context = _trace_context_propagator.extract(
-            external_trace_header
+            external_trace_header or {}
```
sufeng-buaa, line 468: Indeed, `external_trace_header` being `None` can cause a crash. but `external_trace_header` should have already been processed by `extract_trace_headers()`, which would convert it to `{}` if it were None. How did your trigger this crash?

python/sglang/srt/tracing/trace.py
```diff
@@ -534,7 +534,7 @@ def trace_req_finish(
     req_context.root_span.end(end_time=ts)
     if str(req_context.bootstrap_room) in remote_trace_contexts:
         del remote_trace_contexts[str(req_context.bootstrap_room)]
-    else:
+    elif req_context.bootstrap_room_span:
         req_context.bootstrap_room_span.end(end_time=ts)
```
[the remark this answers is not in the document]
sufeng-buaa: If the "trace_context" is not passed from `mini_lb`, `bootstrap_room_span` will definitely exist. I'm also curious how you trigger this crash?

python/sglang/srt/tracing/trace.py
sufeng-buaa: Did you also add the "trace_context" field in the request headers from the upstream service?

vladnosiv: Both bugs were found in tests with Nvidia Dynamo. Then I was immediately confused by the type signature, so I didn't dig deeper here. Actually, I had this bug when I didn't enable tracing in the higher-level components of Dynamo, but enabled it in sglang.

vladnosiv:
(@sufeng-buaa idk why github won't let me reply to the second comment)

Here again, the bug always shoots for me when using Dynamo + aggregated setup. It is quite possible that there is a bug there, but in any case, checking here does not look like overkill, because now the inability to close the span will fail the entire request.

python/sglang/srt/tracing/trace.py
vladnosiv: I think root cause here

https://github.com/sgl-project/sglang/blob/ddb3970eb47483de4f38ab497f2cf3f7969a30c6/python/sglang/srt/managers/tokenizer_manager.py#L2173

and here

https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/tracing/trace.py#L415

Seems like dynamo sent smth like "trace_context": "" in headers

sufeng-buaa:
> (@sufeng-buaa idk why github won't let me reply to the second comment)
> 
> Here again, the bug always shoots for me when using Dynamo + aggregated setup. It is quite possible that there is a bug there, but in any case, checking here does not look like overkill, because now the inability to close the span will fail the entire request.

ok, LGTM. I'll try to reproduce it locally. Could you provide some steps for reproduction?

python/sglang/srt/tracing/trace.py
sufeng-buaa: Setting a custom "trace_context" indeed causes this issue. I'm planning to remove it in the next PR and keep only the standard W3C trace header. https://github.com/sgl-project/sglang/pull/13152

vladnosiv:
> ok, LGTM. I'll try to reproduce it locally. Could you provide some steps for reproduction?

My exact setup looks pretty overloaded for a complete repeat. I can try to provide a minimal setup for playback today. 

I guess everything will work out if you take any [simple example ](https://github.com/ai-dynamo/dynamo/blob/main/examples/backends/sglang/deploy/agg.yaml) and add enable-trace in sglang. 
It is important that I launched dynamo from the main branch, I am not sure that the feature for linking dynamo and sglang tracing managed to get into theirs latest release.

vladnosiv:
@sufeng-buaa 
Judging by the Dynamo code, since they themselves wrap the sglang engine in a worker, they themselves control the propagation of the context further. This is where they pass the context on (the `trace_set_remote_propagate_context` call is just lower down the chain, and no custom trace_context btw):


https://github.com/ai-dynamo/dynamo/blob/2ae466268ee20352f63e4f749d618b1fbaea240a/components/src/dynamo/sglang/request_handlers/llm/decode_handler.py#L125


At the same time, they rely on their own `enable_trace` flag, which I had turned off during that crash. And not on the SGLang flag, which was enabled for me. As a result, no one provided the context, while SGLang was waiting for it.
There's really a Dynamo bug here, but I still don't think it's overkill to make the change I've made here, because it's pretty easy to step on this when framing SGLang from the outside. And in general, tolerance to None on a critical query path looks reasonable in my opinion.

vladnosiv:
@sufeng-buaa 

About the `bootsrap_room_span`:

SGLang set this only if room id not in the `remote_trace_contexts`:

https://github.com/sgl-project/sglang/blob/d6108166bc7f1f157407fe5e52111299b17c9993/python/sglang/srt/tracing/trace.py#L463-L474

But Dynamo in aggregated case set the room id = 0 always:

https://github.com/ai-dynamo/dynamo/blob/2ae466268ee20352f63e4f749d618b1fbaea240a/components/src/dynamo/sglang/request_handlers/llm/decode_handler.py#L165

and

https://github.com/ai-dynamo/dynamo/blob/main/components/src/dynamo/sglang/request_handlers/handler_base.py#L150

So this is essentially a Dynamo bug entirely (however, it exists because of a not very trivial contract). I would still prefer to leave the tolerance to None, since this is the default of the `bootstrap_room_span` field, but I can also remove this change.

vladnosiv:
> Are you the developer responsible for Dynamo's tracing functionality, or do you know the person who is?

I'm more like an early adopter of SGLang and Dynamo integration here, and the one responsible for dynamo + sglang tracing (at least at first glance) is @ishandhanani

sufeng-buaa:
This PR LGTM @ishandhanani

sufeng-buaa (approved):

ishandhanani:
Got it - enabled auto-merge. 

@vladnosiv - @sufeng-buaa and I have also discussed offline and will be doing slight refactor of the tracing for dynamo/`sgl.Engine` api starting here https://github.com/sgl-project/sglang/pull/15814. TLDR is that we will pass in W3C headers directly into the engine which will make things a lot cleaner

PR #15699, opened 2025-12-23
feat(metrics): add prefix cache monitoring statistics

maggie26375:
Add new Prometheus metrics for cache monitoring:
- sglang:cache_entry_count - number of entries in the prefix cache
- sglang:cache_total_tokens - total tokens stored in the prefix cache

This helps operators monitor cache efficiency in production.

Changes:
- Extended RadixCacheMetricsCollector with 2 new Gauge metrics
- Added get_cache_stats() method to RadixCache
- Updated metrics reporting in scheduler_metrics_mixin.py
- Added tests for new metrics

## Checklist

### DIFF
--- a/python/sglang/srt/managers/scheduler_metrics_mixin.py
+++ b/python/sglang/srt/managers/scheduler_metrics_mixin.py
@@ -249,6 +249,22 @@ class SchedulerMetricsMixin:
                     self.disagg_decode_transfer_queue.queue
                 )
 
+            # Cache monitoring metrics
+            # Only call get_cache_stats if tree_cache supports it (e.g., RadixCache)
+            if (
+                hasattr(self, "tree_cache")
+                and self.tree_cache is not None
+                and self.tree_cache.metrics_collector is not None
+                and hasattr(self.tree_cache, "get_cache_stats")
+            ):
+                cache_stats = self.tree_cache.get_cache_stats()
+                self.tree_cache.metrics_collector.set_cache_entry_count(
+                    cache_stats["entry_count"]
+                )
+                self.tree_cache.metrics_collector.set_cache_total_tokens(
+                    cache_stats["total_tokens"]
+                )
+
             # Others
             self.calculate_utilization()
             self.update_lora_metrics()
--- a/python/sglang/srt/mem_cache/radix_cache.py
+++ b/python/sglang/srt/mem_cache/radix_cache.py
@@ -325,6 +325,7 @@ class RadixCache(BasePrefixCache):
         self.root_node.hash_value = []
         self.evictable_size_ = 0
         self.protected_size_ = 0
+        self.entry_count_ = 0
         self._record_all_cleared_event()
 
     def maybe_bigram_convert(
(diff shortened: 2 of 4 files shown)


python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -607,6 +607,25 @@ def protected_size(self):
...
+    def _count_entries(self) -> int:
+        """Count total number of cache entries (tree nodes with values)."""
+        count = 0
+        stack = [self.root_node]
+        while stack:
+            node = stack.pop()
+            if node.value is not None and len(node.value) > 0:
+                count += 1
+            stack.extend(node.children.values())
+        return count
```
kartikx: This function is really similar to [_total_size_helper](https://github.com/sgl-project/sglang/blob/96655749379cd6bf9ab0cae84084309cc61a81fc/python/sglang/srt/mem_cache/radix_cache.py#L738-L748). For performance, it might be better to compute both total_tokens and entry_count in the same pass.

python/sglang/srt/managers/scheduler_metrics_mixin.py
```diff
@@ -236,6 +236,20 @@ def log_prefill_stats(
                 prefill_cache_tokens=adder.log_hit_tokens,
             )
 
+            # Cache monitoring metrics
+            if (
```
kartikx, line 254: If `self.tree_cache` is not of type `RadixCache` then `.get_cache_stats` will fail, since it is not implemented for all sub-classes of `BasePrefixCache`.

We should either check `isinstance(self.tree_cache, RadixCache)`, or implement the function for the base class.

kartikx (commented):
Thank you for the changes! I have left a few comments.

python/sglang/srt/mem_cache/radix_cache.py
maggie26375: Combined `entry_count` and `total_tokens` computation into a single pass in `get_cache_stats()`, removing the separate `_count_entries()` method.

python/sglang/srt/managers/scheduler_metrics_mixin.py
maggie26375: Added `hasattr(self.tree_cache, "get_cache_stats")` check to ensure compatibility with all `BasePrefixCache` subclasses.

maggie26375:
Thank you for the review @kartikx! I've addressed both comments, please take another look when you have a chance

python/sglang/srt/mem_cache/radix_cache.py
kartikx: I am concerned for the overhead that this function adds (O (n) tree traversal), since it will be called on every metric invocation. Can we perhaps store `entry_count` and `total_tokens` as variables that are updated whenever we add or remove nodes? Then this function can just return that value, similar to `self.evictable_size()`

Let me know what you think

maggie26375: Good catch! Implemented as suggested - added entry_count_ maintained incrementally like evictable_size_, and total_size() now returns evictable_size_ + protected_size_. Both metrics are now O(1).

PR #15716, opened 2025-12-24
Tiny fix test eagle infer b.

hnyls2002:
Due to the wrong usage of the small KV cache size environment, the other tests in `test_eagle_infer_b.py` are also infected by this small KV cache size, which leads to an out-of-memory problem.


hnyls2002:
All eagle UTs passed.

https://github.com/sgl-project/sglang/actions/runs/20477218774/job/58843926374

PR #7415, opened 2025-06-21
[RL] support update_weights_from_tensor for mtp

zhuzilin:
## Motivation

This PR does 2 things:

~1. Put the `monkey_patch_torch_reductions` into `TpModelWorker.update_weights_from_tensor` to greatly accelerate the update.~
2. Add `update_weights_from_tensor` to `EAGLEWorker` so we can update the mtp head of a model.

Thank you your your time on reviewing this PR :)

## Checklist

- [x] Format your code according to the [Code Formatting with Pre-Commit](https://docs.sglang.ai/references/contribution_guide.html#code-formatting-with-pre-commit).
- [ ] Add unit tests as outlined in the [Running Unit Tests](https://docs.sglang.ai/references/contribution_guide.html#running-unit-tests-adding-to-ci).
- [ ] Update documentation / docstrings / example tutorials as needed, according to [Writing Documentation](https://docs.sglang.ai/references/contribution_guide.html#writing-documentation-running-docs-ci).
- [ ] Provide throughput / latency benchmark results and accuracy evaluation results as needed, according to [Benchmark and Profiling](https://docs.sglang.ai/references/benchmark_and_profiling.html) and [Accuracy Results](https://docs.sglang.ai/references/accuracy_evaluation.html).
- [ ] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co-author when merging the PR.
- [ ] Please feel free to join our Slack channel at https://slack.sglang.ai to discuss your PR.

PR #15726, opened 2025-12-24
[Fix] Add --disable-draft-model-update to control draft model updates(especially in RL)

bingyang-lei:
## Motivation
This PR does:
- Add `--disable-draft-model-update` parameter to determine whether to update the draft model.

The motivation is as follows:

In PR https://github.com/sgl-project/sglang/pull/7415, they support update_weights_from_tensor for mtp. However, in many cases, the RL framework only supports mtp layer updates, not external draft model updates. So If I only want to use an external draft model (e.g., trained by SpecForge) to accelerate inference during the RL rollout process without updating the draft model, the current SGlang engine will incorrectly load the main model weights into the draft model, causing issues like https://github.com/THUDM/slime/issues/1177

Therefore, we need to introduce an additional parameter to determine whether to update the draft model. If the external draft model is solely used to enhance inference speed, weight updates should not be performed during RL training.

## Checklist

### DIFF
--- a/python/sglang/srt/managers/io_struct.py
+++ b/python/sglang/srt/managers/io_struct.py
@@ -1337,6 +1337,8 @@ class UpdateWeightsFromTensorReqInput(BaseReq):
     abort_all_requests: bool = False
     # Optional: Update weight version along with weights
     weight_version: Optional[str] = None
+    # Optional: Determine whether to disable updating the draft model
+    disable_draft_model: Optional[bool] = None
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


fzyzcjy (commented):
wondering whether this means today we are not able to let RL framework have trainable EAGLE3 draft model (and update weights), if so maybe one way is to add a flag like `target_model: Literal["main", "draft"]` in the update weights call to specify which model to update

fzyzcjy (approved):

PR #15737, opened 2025-12-24
enable xpu graph

huaiyuzh:
enable xpu graph


airMeng:
is it workable with triton backend?

PR #15744, opened 2025-12-24
[diffusion] chore: add a switch to control whether to use runai

### LINKED ISSUE #15728: [Bug] multimodal_gen: Need a way to opt out use of runai_model_streamer
### Checklist

### Describe the bug

# Background

We store model files on S3 like object store and FUSE mount them as local dir with tools like s3fs.

We know that loading tensors with `safetensors.safe_open().keys()` can cause random io and it works so bad with s3fs that sometimes the io offsets are so random that the load process looks as if it hangs.

To work around this, we modify s3fs to intercept open call from app and prefetch the whole file into system pagecache so that later when app reads the model file the random io will be on mem and perf issue mitigated.  

The work around fits well with `safetensors.load_file()` and the old `safetensors.safe_open().keys()` iterator.

# Issue

For multimodal_gen, a recent commit (https://github.com/sgl-project/sglang/commit/fa0ca97694a90c5aa10efd4a6c2e5a8ea9e73) introduces `runai_model_streamer` by default. 

`runai_model_streamer` uses threaded workers (default 16) and each file can be open/close many times.  This breaks the work around for the following observations
 - Likely the pagecache is not big enough to hold all the model files.  This is further worsened by the fact runai_model_streamer defaults to a "unlimited" buffer policy ([RUNAI_STREAMER_MEMORY_LIMIT](https://github.com/run-ai/runai-model-streamer/blob/master/docs/src/env-vars.md#runai_streamer_memory_limit))
 - With tensor parallel, all the subprocesses replicates the same behavior, meaning the runai buffer cost will be multiplied and this can further add pressure on pagecache eviction.

No doubt that runai_model_streamer is very good at saturating storage throughput and good for load times on SSD etc.  But it's too much in this FUSE mount case and I wonder if we could introduce a command line argument or environment variable to opt out its use.  I am new with sglang and comments are welcome cc @zyksir @mickqian 

### Reproduction

Wan-AI/Wan2.2-I2V-A14B-Diffusers

### Environment

N/A

mickqian:
## Motivation
fix #15728

## Checklist

### DIFF
--- a/python/sglang/multimodal_gen/runtime/entrypoints/openai/image_api.py
+++ b/python/sglang/multimodal_gen/runtime/entrypoints/openai/image_api.py
@@ -231,7 +231,7 @@ async def edits(
         )
 
     # Save all input images; additional images beyond the first are saved for potential future use
-    uploads_dir = os.path.join("outputs", "uploads")
+    uploads_dir = os.path.join("inputs", "uploads")
     os.makedirs(uploads_dir, exist_ok=True)
     image_list = merge_image_input_list(images, urls)
 
--- a/python/sglang/multimodal_gen/runtime/loader/adapter_loader.py
+++ b/python/sglang/multimodal_gen/runtime/loader/adapter_loader.py
@@ -27,7 +27,7 @@ class AdapterLoader(ComponentLoader):
     expected_library = "diffusers"
 
     def load_customized(
-        self, component_model_path: str, server_args: ServerArgs, *args
+        self, component_model_path: str, server_args: ServerArgs, *args, **kwargs
     ):
         config = get_diffusers_component_config(model_path=component_model_path)
 
--- a/python/sglang/multimodal_gen/runtime/loader/bridge_loader.py
+++ b/python/sglang/multimodal_gen/runtime/loader/bridge_loader.py
@@ -26,7 +26,12 @@ class BridgeLoader(ComponentLoader):
     expected_library = "diffusers"
 
     def load_customized(
-        self, component_model_path: str, server_args: ServerArgs, component_name: str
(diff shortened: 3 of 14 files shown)


python/sglang/multimodal_gen/envs.py
```diff
@@ -396,6 +398,9 @@ def maybe_convert_int(value: str | None) -> int | None:
             os.getenv("SGLANG_CACHE_DIT_TS_ORDER", "1"),
         )
     ),
+    "SGLANG_USE_RUNAI_MODEL_STREAMER": lambda: get_bool_env_var(
+        os.getenv("SGLANG_USE_RUNAI_MODEL_STREAMER", "true")
```
yousong: The `os.getenv()` call is not needed.  Otherwise all good👍

```diff
diff --git a/python/sglang/multimodal_gen/envs.py b/python/sglang/multimodal_gen/envs.py
index e855087d7..73d092f8f 100644
--- a/python/sglang/multimodal_gen/envs.py
+++ b/python/sglang/multimodal_gen/envs.py
@@ -399,7 +399,7 @@ environment_variables: dict[str, Callable[[], Any]] = {
         )
     ),
     "SGLANG_USE_RUNAI_MODEL_STREAMER": lambda: get_bool_env_var(
-        os.getenv("SGLANG_USE_RUNAI_MODEL_STREAMER", "true")
+        "SGLANG_USE_RUNAI_MODEL_STREAMER", default="true",
     ),
 }
 
```

mickqian:
rebased. just curious, based on the conversation in #15728, I thought `runai_model_streamer` has solved your problem? @yousong

Issue #15643, opened 2025-12-22
[Bug] v0.5.6 import vllm error: pooler.py introduces strict hardware dependency via activation.py


sharonyu-115:
### Checklist

- [x] I searched related issues but found no solution.
- [x] The bug persists in the latest version.
- [ ] Issues without environment info and a minimal reproducible demo are hard to resolve and may receive no feedback.
- [ ] If this is not a bug report but a general question, please start a discussion at https://github.com/sgl-project/sglang/discussions. Otherwise, it will be closed.
- [ ] Please use English. Otherwise, it will be closed.

### Describe the bug

I encountered an issue in sglang v0.5.6 where simply importing sglang components on a CPU-only environment (without vllm installed) causes a crash. This breaks distributed workflows (e.g., Ray-based verl drivers/orchestrators) that run on CPU nodes but need to import sglang utilities for coordination.

The issue seems to stem from a new dependency chain that propagates strict hardware requirements to lightweight modules: 
1. `pooler.py`: Change: In v0.5.6, `sglang.srt.layers.pooler` added a top-level import:    `from sglang.srt.layers.activation import get_cross_encoder_activation_function`
2. `activation.py`:  The activation module executes strict logic at import time. If `is_cuda()` is False (CPU node), it mandates vllm presence:
    ```
       if not (_is_cuda or ...):        
           from vllm.model_executor.layers.activation import GeluAndMul... 
    ``` 
    Crashes if vllm is missing.
3. Central modules like `model_runner.py` import `pooler.py`. Consequently, any module importing `model_runner` (even for basic utility purposes like `weight_sync`) now transitively triggers the `activation.py` hardware check.
4. Result: A CPU-based orchestrator importing `sglang.srt.weight_sync.utils` immediately crashes with `ModuleNotFoundError: No module named 'vllm'`, whereas in v0.5.5 this import chain was clean and worked on CPU.

### Reproduction

The following script can reproduce the problem: 

```
import ray
import sys
import os

# We define a TaskRunner actor that mimics the one in verl/recipe/dapo/main_dapo.py
@ray.remote(num_cpus=1)
class TaskRunner:
    def __init__(self):
        print(f"TaskRunner started on PID: {os.getpid()}")

    def run(self):
        print("TaskRunner: Running initialization flow...")
        from sglang.srt.weight_sync.utils import update_weights

def main():
    if ray.is_initialized():
        ray.shutdown()
        
    print("Initializing Ray...")
    ray.init()
    
[...]

PR #14867, opened 2025-12-11
[cpu/arm64] support run sglang on arm64 cpu

cyb70289:
## Motivation
To support run sglang on Arm64 CPU with torch_native attention backend.

## Modifications
Enabled Arm CPU to run some code path which were only available for Intel AMX.

## Accuracy Tests
Model accuray = 0.680 for meta-llama/Llama-3.2-3B-Instruct model, few_shot_gsm8k dataset, 200 questions.

## Checklist

- [x] Format your code according to the [Format code with pre-commit](https://docs.sglang.io/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.io/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.io/developer_guide/contribution_guide.html#write-documentations).
- [x] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.io/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.io/developer_guide/contribution_guide.html#benchmark-the-speed).
- [x] Follow the SGLang code style [guidance](https://docs.sglang.io/developer_guide/contribution_guide.html#code-style-guidance).
- [ ] Work with maintainers to merge your PR. See the [PR Merge Process](https://github.com/sgl-project/sglang/blob/main/.github/MAINTAINER.md#pull-request-merge-process)

PR #15749, opened 2025-12-24
[Fix] CPU-only chained import crashes due to no vllm import in activation

RongMu00:
## Motivation

In sglang v0.5.6, importing sglang components on CPU-only environment leads to a crash due to no vllm installed. This may cause a range of crashes due to the dependency between components (according to #15643). Specifically, in python/sglang/srt/layers/activation.py, at the end of the code, there is a manual import of vllm for CPU-only environment. However, there is no vllm installed by running this single test:

```
sglang git:(main) CUDA_VISIBLE_DEVICES="" python -c "from sglang.srt.weight_sync import utils"
...
  File "<string>", line 1, in <module>
  File "/sgl-workspace/sglang/python/sglang/srt/weight_sync/utils.py", line 10, in <module>
    from sglang.srt.model_executor.model_runner import LocalSerializedTensor
  File "/sgl-workspace/sglang/python/sglang/srt/model_executor/model_runner.py", line 106, in <module>
    from sglang.srt.layers.pooler import EmbeddingPoolerOutput
  File "/sgl-workspace/sglang/python/sglang/srt/layers/pooler.py", line 12, in <module>
    from sglang.srt.layers.activation import get_cross_encoder_activation_function
  File "/sgl-workspace/sglang/python/sglang/srt/layers/activation.py", line 389, in <module>
    from vllm.model_executor.layers.activation import (  # noqa: F401
ModuleNotFoundError: No module named 'vllm'
```

This causes crashes in pooler.py when importing activation, in model_runner.py when importing pooler, and in any files import model_runner.

## Modifications

To avoid chained crashes in CPU-only environment, I simply moved the activation import line of code from the top to CrossEntropyPooler class.
`from sglang.srt.layers.activation import get_cross_encoder_activation_function`

It can also be fixed by adding error handling in the cpu-only condition. Please let me know if this is a more anticipated way to fix it.

## Checklist

- [ x] Format code

### DIFF
--- a/python/sglang/srt/layers/pooler.py
+++ b/python/sglang/srt/layers/pooler.py
@@ -9,7 +9,7 @@ import torch
 import torch.nn as nn
 from transformers import PretrainedConfig
 
-from sglang.srt.layers.activation import get_cross_encoder_activation_function
+
 from sglang.srt.model_executor.forward_batch_info import ForwardBatch
 
 
@@ -96,6 +96,7 @@ class CrossEncodingPooler(nn.Module):
         super().__init__()
         self.classifier = classifier
         self.pooler = pooler
+        from sglang.srt.layers.activation import get_cross_encoder_activation_function
         self.default_activation_function = get_cross_encoder_activation_function(config)
 
     def forward(


ishandhanani (approved):

hnyls2002:
Thanks @RongMu00! The root cause here was fixed by #14867, which removed the vllm fallback import from activation.py entirely, so the CPU-only import chain no longer crashes on current main. Closing this as fixed elsewhere; please reopen if I've missed something.

PR #15333, opened 2025-12-17
[GLM-4.7] GLM-4.7 Tool Parser and Doc Update

zRzRzRzRzRzRzR:
Support for GLM-4.7 model's Tool Parser and partial documentation

PR #15753, opened 2025-12-24
Fix GLM-4.7 MoE Detector complex JSON Schema type parsing

Leoyzen:
- Modify get_argument_type function to correctly parse complex JSON Schema structures including anyOf/oneOf/allOf and OpenAI-style array types like ["string", "null"]
- Improve _format_value_complete function to properly handle object/array types as JSON rather than strings
- Add additional parsing strategies in parse_arguments for object/array type handling
- Add comprehensive test cases to verify complex type definitions are correctly parsed as their appropriate data types (e.g., arrays) rather than strings

This fixes the issue where array values like [] were being parsed as string "[]" instead of actual array [].

## Motivation

This pull request addresses a bug in the GLM-4.7 MoE Detector where complex JSON Schema types were not being parsed correctly. Specifically, when a tool parameter was defined with complex structures like `anyOf` (e.g., `{"anyOf": [{"type": "array"}, {"type": "null"}]}`), the values were being treated as strings instead of their actual types (like arrays). For example, an array value `[]` would be parsed as the string `"[]"` rather than the actual array `[]`.

## Modifications

- Modified `get_argument_type` function in `glm47_moe_detector.py` to correctly parse complex JSON Schema structures including anyOf/oneOf/allOf and OpenAI-style array types like ["string", "null"]
- Improved `_format_value_complete` function to properly handle object/array types as JSON rather than strings
- Added additional parsing strategies in `parse_arguments` for object/array type handling
- Added comprehensive test cases in `test_function_call_parser.py` to verify complex type definitions are correctly parsed as their appropriate data types (e.g., arrays) rather than strings

## Accuracy Tests

The fix has been validated with test cases that confirm complex JSON Schema types are now correctly parsed. The tests verify that array values like `[]` are correctly interpreted as arrays instead of strings.

## Benchmarking and Profiling

No performance impact is expected from these changes as they only affect the parsing logic when complex types are encountered.

## Checklist


JustinTong0323 (approved):

wwzeng1:
Hi, I tried this and this PR introduces a bug where a partial tool call can be early matched and incorrectly returned. For example if the tool should be `<tool_call>read_file` you'll early match `<tool_call>read` which causes the tool to fail.

I'll leave a separate comment with the fix

python/sglang/srt/function_call/glm47_moe_detector.py
```diff
@@ -408,17 +487,33 @@ def parse_streaming_increment(
...
-                func_args_raw = partial_match.group(2).strip()
+            # Only proceed if we have a non-empty function name
+            if partial_match and (func_name := partial_match.group(1)) is not None:
+                # Check if we have a valid function name before proceeding
+                func_name = func_name.strip()
+                func_args_raw = partial_match.group(2)
+                func_args_raw = (
+                    func_args_raw.strip() if func_args_raw is not None else ""
+                )
                 is_tool_end = partial_match.group(3)
```
wwzeng1: add a check below:

```
                known_tool_names = {tool.function.name for tool in tools}
                func_name_is_complete = (
                    func_args_raw or  # saw <arg_key>
                    is_tool_end == self.eot_token or  # saw </tool_call>
                    func_name in known_tool_names  # matches a known tool exactly
                )

                if not func_name_is_complete:
                    # Function name might be incomplete, keep buffering
                    return StreamingParseResult(normal_text=normal_text_before, calls=[])
```

Leoyzen: @wwzeng1 The bug was introduced by original implementation #15333:
```python
        try:
            # Try to match a partial or complete tool call
            partial_match = re.search(
                pattern=r"<tool_call>(.*?)(<arg_key>.*?)?(</tool_call>|$)",
                string=current_text,
                flags=re.DOTALL,
            )
            if partial_match:
                func_name = partial_match.group(1).strip()
                func_args_raw = partial_match.group(2).strip()
                is_tool_end = partial_match.group(3)

                # Initialize state if this is the first tool call
                if self.current_tool_id == -1:
                    self.current_tool_id = 0
                    self.prev_tool_call_arr = []
                    self.streamed_args_for_tool = [""]
                    self._streamed_raw_length = 0
                    self.current_tool_name_sent = False
                    self._reset_streaming_state()

                # Ensure we have enough entries in our tracking arrays
                while len(self.prev_tool_call_arr) <= self.current_tool_id:
                    self.prev_tool_call_arr.append({})
                while len(self.streamed_args_for_tool) <= self.current_tool_id:
                    self.streamed_args_for_tool.append("")

                # Send tool name first if not sent yet
                if not self.current_tool_name_sent:
                    assert func_name, "func_name should not be empty"
                    calls.append(
                        ToolCallItem(
                            tool_index=self.current_tool_id,
                            name=func_name,
                            parameters="",
                        )
                    )
                    self.current_tool_name_sent = True
                    self._streamed_raw_length = 0
                    self._reset_streaming_state()
                    # Store the tool call info
                    self.prev_tool_call_arr[self.current_tool_id] = {
                        "name": func_name,
                        "arguments": {},
                    }
```

When the model output partial function name tokens, the regex still match(`(</tool_call>|$)`), and send the function name, then set `current_tool_name_sent=True`.

This will cause the remain function name will not being send after.

CatherineSue:
LGTM. Could you take a look at CI failures?

Leoyzen:
@JustinTong0323  I'm not sure how to fix these failures. I checked the details, but the errors seem unrelated to the changes in this PR. What should I do?

Leoyzen:
@JustinTong0323 @CatherineSue Could you take a look about this PR?

PR #15760, opened 2025-12-24
fix(test): use temp directory in VLM test to avoid cached results

### LINKED ISSUE #14760: [Bug][CI] `test/srt/models/test_vlm_models.py` is reading local results
### Checklist

### Describe the bug

The test uses a hardcoded output path "./logs". If `lmms_eval` finds existing JSON result files from a previous run in that directory, it skips re-running the evaluation and uses cached results. The test then reads those old cached files instead of running fresh evaluations, so it's reading the same result.

### Reproduction

`python test/srt/models/test_vlm_models.py`

### Environment

N/A

yurekami:
## Summary

Fixes #14760

The `test_vlm_models.py` test was using a hardcoded `./logs` directory which could contain cached result files from previous `lmms_eval` runs. When `lmms_eval` finds existing JSON result files from a previous run in that directory, it skips re-running the evaluation and uses cached results. The test then reads those old cached files instead of running fresh evaluations.

## Changes

- Use `tempfile.mkdtemp()` to create a unique temporary directory for each model test
- Clean up the directory after each test with `shutil.rmtree()`
- This ensures each test run uses fresh evaluation results

## Test plan

- [ ] Run `python test/srt/models/test_vlm_models.py` multiple times and verify it runs fresh evaluations each time
- [ ] Verify the temporary directories are cleaned up after tests complete

### DIFF
--- a/test/srt/models/test_vlm_models.py
+++ b/test/srt/models/test_vlm_models.py
@@ -1,12 +1,14 @@
 import argparse
 import random
+import shutil
 import sys
+import tempfile
 import unittest
 from types import SimpleNamespace
 
 from sglang.srt.utils import is_hip
-from sglang.test.mmmu_vlm_mixin import DEFAULT_MEM_FRACTION_STATIC, MMMUVLMMixin
-from sglang.test.test_utils import CustomTestCase, is_in_ci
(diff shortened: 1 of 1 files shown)


yurekami:
Closing as duplicate of #15777 which fixes the same issue.

PR #15764, opened 2025-12-24
feat(openai): use --preferred-sampling-params as defaults in OpenAI API

yurekami:
## Summary
- Makes `--preferred-sampling-params` CLI arg work as actual defaults for the OpenAI-compatible chat completion API
- Priority: user value > preferred_sampling_params > model generation_config > OpenAI defaults
- Fixes the responses endpoint to not override parent's default_sampling_params

This addresses the feature request to have `--preferred-sampling-params` actually set default sampling parameters for models that don't have a `generation_config.json`, such as the recent [MiMo-V2-Flash](https://huggingface.co/XiaomiMiMo/MiMo-V2-Flash) model.

## Changes
- `serving_chat.py`: Merge `preferred_sampling_params` with model generation config, with CLI params taking precedence
- `serving_responses.py`: Don't override parent class's `default_sampling_params` with empty dict

## Test plan
- [ ] Start server with `--preferred-sampling-params '{"temperature": 0.7, "top_p": 0.9}'`
- [ ] Verify chat completions use those defaults when not specified in request
- [ ] Verify request-level params still override the defaults

Fixes #15487

### DIFF
--- a/python/sglang/srt/entrypoints/openai/serving_chat.py
+++ b/python/sglang/srt/entrypoints/openai/serving_chat.py
@@ -69,12 +69,23 @@ class OpenAIServingChat(OpenAIServingBase):
         self.reasoning_parser = self.tokenizer_manager.server_args.reasoning_parser
 
         # Get default sampling parameters from model's generation config
-        self.default_sampling_params = (
-            self.tokenizer_manager.model_config.get_default_sampling_params()
+        # and merge with preferred_sampling_params from CLI args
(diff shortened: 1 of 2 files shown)


wengsnow:
Hi, I think this modification is useful because --preferred-sampling-params is currently not working with the chat/completions api. Is there an alternative solution planned, or did I miss some context on why this change wasn't accepted? Thanks! @merrymercy

PR #15773, opened 2025-12-24
[diffusion] benchmark: enable performance benchmark with warmup

fsygd:
## Motivation

Performance benchmark will be better with warmup to get more accurate performance, especially when something like DeepGemm is involved. (e.g. #15403)

without warmup on H200:

image

with warmup on H100:

image

## Checklist


python/sglang/multimodal_gen/docs/contributing.md
```diff
@@ -29,12 +29,12 @@ For PRs that impact **latency**, **throughput**, or **memory usage**, you **shou
 
 1.  **Baseline**: run the benchmark (for a single generation task)
     ```bash
-    $ sglang generate --model-path <model> --prompt "A benchmark prompt" --perf-dump-path baseline.json
+    $ python python/sglang/multimodal_gen/benchmarks/run_perf.py --model-path <model> --warmup 10 --perf-dump-path baseline.json
```
mickqian: I'm worrying that modifying this place will:
1. not be able to inform the users that `--perf-dump-path` is a valid sampling params
2. increase duplication

To achieve the same effect, I suggest we turn on `warmup` automatically when `--perf-dump-path` is turned on

fsygd: Reasonable. I'll do it.

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -1104,7 +1104,11 @@ def log_time_stats(self):
         if self.has_log_time_stats:
             return
 
-        bootstrap_info = f", bootstrap_room={self.bootstrap_room}" if self.bootstrap_room is not None else ""
+        bootstrap_info = (
```
mickqian, line 1107: why this😂

mickqian:
@BBuf would you take a look?

python/sglang/srt/managers/schedule_batch.py
fsygd: Pre-commit did this, not me. Should I rollback?

BBuf:
If warmup is enabled by default here, it might mask some performance issues that would otherwise be discoverable, such as the problem found previously here: https://github.com/sgl-project/sglang/pull/15511. For offline inference, you must accept the warmup cost. For serving, the first request also has to accept the warmup cost. For subsequent requests, if the resolution changes or similar things happen, deepgemm still needs to recompile, which also has a cost. These should not be removed, especially for profiling.

fsygd:
> If warmup is enabled by default here, it might mask some performance issues that would otherwise be discoverable, such as the problem found previously here: #15511. For offline inference, you must accept the warmup cost. For serving, the first request also has to accept the warmup cost. For subsequent requests, if the resolution changes or similar things happen, deepgemm still needs to recompile, which also has a cost. These should not be removed, especially for profiling.

I see, warmup by default may hide something we concern. May i add warmup for option maybe sometimes we need it, or just close the pr?

PR #15799, opened 2025-12-25
[model-gateway] Optimize L1 Cache Insertion with Incremental Hashing and Tokenization

ppraneth:
## Motivation

The L1 Tokenizer Cache is designed to optimize repeated prefixes by caching at special token boundaries. However, the previous implementation of `insert_at_boundaries` in `l1.rs` had  complexity. For every special token boundary discovered, the system performed a full re-hash and a full re-tokenization of the entire growing prefix.

In long conversation histories or prompts with many special tokens , this redundant work resulted in significant millisecond-scale latency spikes during cache misses. The goal of this PR is to transition to an incremental, single-pass approach to achieve  linear complexity.

## Modifications

* **Incremental Hashing**: Replaced `blake3::hash(prefix)` with a persistent `blake3::Hasher` that updates incrementally with only the new text segment between boundaries.
* **Segmented Tokenization**: Updated the loop to tokenize only the new segment between the current and previous boundary rather than the entire prefix. These segments are then accumulated to form the full prefix token sequence.
* **BOS Token Management**: Added logic to detect if the tokenizer automatically prepends a Beginning of Sentence (BOS) token. The system now strips these redundant tokens from segments after the first one to ensure the accumulated token sequence remains correct and consistent with full-string encoding.
* **Performance Stability**: Maintained the existing architectural guarantees that splits only occur at atomic special tokens, which are safe boundaries for BPE tokenizers.

## Accuracy Tests

* **Correctness Suite**: Verified via `cargo test --test tokenizer_cache_correctness_test`.

## Benchmarking and Profiling

The optimization was validated using the `benches/l1_cache_insertion_bench.rs` stress test.

| Metric | Before Optimization | Post Optimization | Change |
| --- | --- | --- | --- |
| **Mean Latency (50 Turns)** | **6.3181 ms** | **407.91 µs** | **~15.5x Speedup** |
| **Algorithmic Complexity** | $O(N^2)$ | $O(N)$ | **Linear Scaling** |
| **Confidence Interval** | 6.06ms - 6.56ms | 407.2µs - 408.6µs | **Highly Stable** |

_Happy holidays and Merry Christmas to the sglang team. Hope this optimization brings a little extra speed cheer 🎄_
## Checklist


ppraneth:
@slin1237 looks like ci failure happens on main branch

PR #15802, opened 2025-12-25
Cleanup `ModelRunner`

hnyls2002:
- `SGL_DISABLE_TP_MEMORY_INBALANCE_CHECK` -> `SGLANG_ENABLE_TP_MEMORY_INBALANCE_CHECK`
- Move some logic into a separate function in `ModelRunner`: `configure_kv_cache_dtype`, `get_cell_size_per_token`.
- Remove `SGLANG_CI_SMALL_KV_SIZE`, which is a duplicate of `--max-total-tokens`.

### DIFF
--- a/python/sglang/srt/model_executor/model_runner.py
+++ b/python/sglang/srt/model_executor/model_runner.py
@@ -1493,7 +1493,7 @@ class ModelRunner:
 
         return result
 
-    def get_kv_cache_cell_size(self, num_layers: int) -> int:
+    def get_cell_size_per_token(self, num_layers: int) -> int:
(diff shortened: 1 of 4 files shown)


python/sglang/srt/model_executor/model_runner.py
```diff
@@ -1835,14 +1820,11 @@ def init_memory_pool(
 
         log_info_on_rank0(logger, f"Using KV cache dtype: {self.kv_cache_dtype}")
 
+    def init_memory_pool(self, total_gpu_memory: int, server_args: ServerArgs):
```
ispobock, line 1823: This function is still too long. We may consider to split it later.

ispobock (approved):

PR #15806, opened 2025-12-25
 Add correctness validation for decode_attention test

harvenstar:
Noticed the TODO in test_decode_attention asking for a correctness check. Currently the test just runs decode_attention_fwd without validating output.

  Implemented a float32 reference using PyTorch einsum that:
  - Handles both MHA and GQA configurations (expands KV heads with repeat_interleave when needed)
  - Uses stable softmax (subtract max before exp)
  - Follows the same kv_indices semantics with index_select
  - Validates with allclose (atol=2e-2, rtol=1e-2)

### DIFF
--- a/test/registered/attention/test_triton_attention_kernels.py
+++ b/test/registered/attention/test_triton_attention_kernels.py
@@ -472,7 +472,7 @@ class TestTritonAttention(CustomTestCase):
 
     def test_decode_attention(self):
         # Here we just to ensure there is no error
-        # TODO: correctnesss test
+        # TODO: correctness test
 
         # Test configurations
         configs = [
--- a/test/registered/attention/test_triton_attention_kernels.py
+++ b/test/registered/attention/test_triton_attention_kernels.py
@@ -525,7 +525,7 @@ class TestTritonAttention(CustomTestCase):
 
         max_abs_err = (o.to(torch.float32) - o_ref).abs().max().item()
         self.assertTrue(
-            torch.allclose(o.to(torch.float32), o_ref, atol=2e-2, rtol=1e-2),
+            torch.allclose(o.to(torch.float32), o_ref, atol=1e-2, rtol=1e-2),
             msg=f"decode_attention mismatch, max_abs_err={max_abs_err}",
         )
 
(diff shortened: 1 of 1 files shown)


test/registered/attention/test_triton_attention_kernels.py
```diff
@@ -470,10 +470,43 @@ def _test_decode_attention_once(self, B, H_Q, H_KV, D):
...
-    def test_decode_attention(self):
-        # Here we just to ensure there is no error
-        # TODO: correctnesss test
+        # Correctness reference (float32, stable softmax)
+        assert H_Q % H_KV == 0, "H_Q must be divisible by H_KV for GQA"
+        group_size = H_Q // H_KV
+
+        o_ref = torch.empty((B, H_Q, D), dtype=torch.float32, device=device)
+
+        for b in range(B):
```
ispobock: Can we move it to a torch deocde reference function and call it here?

test/registered/attention/test_triton_attention_kernels.py
```diff
@@ -470,10 +518,18 @@ def _test_decode_attention_once(self, B, H_Q, H_KV, D):
...
-        # Here we just to ensure there is no error
-        # TODO: correctnesss test
+        # Correctness reference (float32, stable softmax)
+        o_ref = decode_attention_fwd_torch(
+            q, k_buffer, v_buffer, kv_indptr, kv_indices, sm_scale
+        )
 
+        max_abs_err = (o.to(torch.float32) - o_ref).abs().max().item()
+        self.assertTrue(
+            torch.allclose(o.to(torch.float32), o_ref, atol=2e-2, rtol=1e-2),
```
ispobock: For float32, can we set atol and rtol to 1e-3?

harvenstar: Tested with atol=1e-3, rtol=1e-3 but it fails. Observed max errors across configs: 3.1e-3, 3.5e-3, 5.8e-3, and 8.3e-3.

  The errors come from the kernel output being bfloat16 (even though reference is fp32). 
  Using atol=1e-2, rtol=1e-2 gives about 20% margin over max observed error while being 2x stricter than the original 2e-2. Does this make sense? Thank you~

ispobock (approved):

harvenstar:
image

image


@ispobock  The test in this pr, test/registered/attention/test_triton_attention_kernels.py, runs in the PR Test workflow under stage-b-test-small-1-gpu, partition (3). It passed here, link: https://github.com/sgl-project/sglang/actions/runs/20536522211/job/58997026746?pr=15806

PR #8710, opened 2025-08-02
[PD metrics] Add latency Histogram metrics of each stage for generate requests

acelyc111:
## Motivation

In PD disaggregation deployments, the request on either prefill or decode side will experience several stages. When trouble shooting online issue, we can only observe the queue lengths of the request in these stages (e.g. `num_prefill_bootstrap_queue_reqs`, `num_decode_prealloc_queue_reqs`), but we didn't observe the duration of requests on each stage, so it's not easy to distinguish the queue lengths are dynamically changing but keep in the number, or just staying motionless.

## Modifications

Introduce a new metric named `sglang:request_latency_seconds`, it is labeled as other scheduler metrics with an extra label `stage`, which can be the following values:
```
    # prefill
    PREFILL_WAITING = "prefill_waiting"

    # disaggregation prefill
    PREFILL_PREPARE = "prefill_prepare"
    PREFILL_BOOTSTRAP = "prefill_bootstrap"
    PREFILL_FORWARD = "prefill_forward"
    PREFILL_TRANSFER_KV_CACHE = "prefill_transfer_kv_cache"

    # disaggregation decode
    DECODE_PREPARE = "decode_prepare"
    DECODE_BOOTSTRAP = "decode_bootstrap"
    DECODE_WAITING = "decode_waiting"
    DECODE_TRANSFERRED = "decode_transferred"
```
The metric values on each stage are easy to be record by using the method:
```
// stage A operations
req.add_latency("stage_A")
// stage B operations
req.add_latency("stage_B")
// stage C operations
req.add_latency("stage_C")
```

## Checklist

- [x] Format your code according to the [Code Formatting with Pre-Commit](https://docs.sglang.ai/references/contribution_guide.html#code-formatting-with-pre-commit).
- [ ] Add unit tests as outlined in the [Running Unit Tests](https://docs.sglang.ai/references/contribution_guide.html#running-unit-tests-adding-to-ci).
- [ ] Update documentation / docstrings / example tutorials as needed, according to [Writing Documentation](https://docs.sglang.ai/references/contribution_guide.html#writing-documentation-running-docs-ci).
- [ ] Provide throughput / latency benchmark results and accuracy evaluation results as needed, according to [Benchmark and Profiling](https://docs.sglang.ai/references/benchmark_and_profiling.html) and [Accuracy Results](https://docs.sglang.ai/references/accuracy_evaluation.html).
[...]

PR #15809, opened 2025-12-25
[metrics] Add more useful metrics

Baidu-AIAK:
## Motivation

These are practical monitoring metrics used in our production environment:

- **`sglang:prefill_chunk_util`**
  - The utilization ratio of chunked prefill, calculated as the number of tokens involved in the chunked prefill computation divided by `server_args.chunked_prefill_size`.

- **`sglang:request_waiting_time`**
  - **For Centralized architecture:** The time a request spends waiting in the `waiting_queue`.
  - **For Prefill instances:** The time from when a request enters the `bootstrap queue` until the start of the forward pass.
  - **For Decode instances:** The time from when a request enters the `transferring queue` until the start of the forward pass.

- **`sglang:request_zmq_time_seconds`**
  - The latency between the tokenizer process sending a request and the scheduler process receiving it.

- **`sglang:request_first_token_forward_time`**
  - **For Centralized architecture:** The time elapsed from entering the `waiting_queue` to the completion of `run_batch`.
  - **For Prefill instances:** The time elapsed from entering the `bootstrap queue` to the completion of `run_batch`.

- **`sglang:request_time_to_tokenizer_ready_seconds`**
  - The processing time consumed by the tokenizer.

- **`sglang:all_gather_latency_us`**
  - Measures the time taken to aggregate local token counts from all ranks and complete a cross-DP `all_gather` operation before each `run_batch` in the Decode phase (during DP attention/MLP preparation). This indicates the overhead of Data Parallel (DP) communication and synchronization during decoding.

- **`sglang:decode_bs_util`**
  - Batch size utilization during the Decode phase (the proportion of active decode requests per DP worker).

- **`sglang:dp_balance`**
  - Data Parallel (DP) load balance metric, calculated using the Gini coefficient based on DP token counts (1.0 indicates perfect balance).

- **`sglang:generation_time`**
  - The total time required to generate a single token.

- **`sglang:idle_batch_ratio`**
  - The proportion of idle workers among all DP workers.

- **`sglang:run_batch_time`**
  - The execution time for a batch (duration of the forward pass).

- **`sglang:iter_token_process_time`**
  - The processing overhead between tokens, calculated as `generation_time - run_batch_time`.

## Checklist


Baidu-AIAK:
> This is super helpful. Thank you.
> 
> Do you mind sharing some before and after benchmarks with these metrics emitted?

It needs some work. Please wait a moment

ishandhanani:
Sounds good. Let me know when this is re-ready to review

sufeng-buaa:
The metrics and collection method for `request stage latency` are somewhat redundant with the existing implementation in https://github.com/sgl-project/sglang/pull/8710. If we merge this patch, we should revert https://github.com/sgl-project/sglang/pull/8710, consolidate all `request stage latency metrics` within `TimeStats` and `SchedulerMetricsMixin`, and abandon integrating the `request stage latency metric` into the Tracing package in my PR https://github.com/sgl-project/sglang/pull/13152. What do you think? @ShangmingCai @ishandhanani @acelyc111 @Baidu-AIAK @scottjlee

python/sglang/srt/entrypoints/http_server.py
```diff
@@ -1190,8 +1190,10 @@ async def continue_generation(obj: ContinueGenerationReqInput, request: Request)
 @app.post("/v1/completions", dependencies=[Depends(validate_json_request)])
 async def openai_v1_completions(request: CompletionRequest, raw_request: Request):
     """OpenAI-compatible text completion endpoint."""
+    # Timestamp when the HTTP request is received and handed off to the tokenizer
+    tokenizer_rev_request_time = time.time()
```
sufeng-buaa, line 1210: Somewhat redundant with https://github.com/sgl-project/sglang/pull/13432

python/sglang/srt/disaggregation/decode.py
```diff
@@ -861,7 +861,7 @@ def event_loop_overlap_disagg_decode(self: Scheduler):
         self.last_batch: Optional[ScheduleBatch] = None
 
         while True:
-            # Receive requests
+            self.iter_start_time = time.time()
```
merrymercy: 1. Do not delete comment
2. use perf counter?

python/sglang/srt/managers/scheduler_metrics_mixin.py
```diff
@@ -269,6 +288,62 @@ def log_prefill_stats_late(self: Scheduler, batch: Optional[ScheduleBatch]):
                 dp_cooperation_info=batch.dp_cooperation_info,
             )
 
+    def log_prefill_dp_balance_stats(self: Scheduler, batch: ScheduleBatch) -> None:
+        """Log DP-level load-balancing metrics for the prefill stage."""
```
merrymercy, line 301: @fzyzcjy take a review?

python/sglang/srt/managers/scheduler_metrics_mixin.py
```diff
@@ -282,6 +357,57 @@ def log_decode_stats(
         num_running_reqs = len(batch.reqs)
         num_running_reqs_offline_batch = 0
 
+        self.iter_finish_time = time.time()
+        generation_time = self.iter_finish_time - self.iter_start_time
+        run_batch_time = self.iter_forward_finish_time - self.iter_forward_start_time
+        iter_token_process_time = generation_time - run_batch_time
+
+        # Add metrics related to DP balancing
```
merrymercy: split this big function into multiple smaller ones

merrymercy (commented):
Now we have many observability features. There seems to be some redundancy among them. All the related code is scattered around the whole codebase without documentation. The existing related code:
- TimeStats (`python/sglang/srt/metrics/collector.py`)
- per request tracing (`python/sglang/srt/tracing/trace.py`)
- Scheduler metrics for prefill and decode batch  (`python/sglang/srt/managers/scheduler_metrics_mixin.py`)


Questions:
- Can we move all observability code into a single place?
- How can we avoid ad-hoc change to insert metrics collection code all around the code base? We should have a unified interface to collect the latency breakdown of a request.

python/sglang/srt/entrypoints/http_server.py
zhanghaotong: > Somewhat redundant with [#13432](https://github.com/sgl-project/sglang/pull/13432)

Yes, I think this variable is essentially no different from received_time, and it can also be used for embedding requests placed within the serving_base.

python/sglang/srt/metrics/collector.py
```diff
@@ -359,6 +405,56 @@ def __init__(
...
         )
 
+        # Run batch
+        self.generation_time = Gauge(
+            name="sglang:generation_time",
+            documentation="The generation time",
+            labelnames=labels.keys(),
+            multiprocess_mode="mostrecent",
+        )
+        self.run_batch_time = Gauge(
```
sufeng-buaa, line 433: Do `generation_time`,`run_batch_time` and `iter_token_process_time` have any practical observational value in your experience? In `event_loop_overlap`, the `batch` in `run_batch` and the `batch` in `process_batch_result` are not the same, so I don't think adding this observation is necessary. If you simply want to measure the time spent in `process_inputs`, `run_batch`, and `process_batch_result`, it would suffice to record timestamps at the beginning and end of each function. Even then, however, I still find the value limited, since their latency is directly correlated with batch size.

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -703,6 +708,17 @@ async def _tokenize_one_request(
...
 
+        # Record tokenize request ready time
+        tokenizer_ready_time = time.time()
+        if (
+            tokenizer_rev_request_time is not None
+            and tokenizer_rev_request_time > 0
+            and self.enable_metrics
+        ):
+            self.metrics_collector.observe_request_time_to_tokenizer_ready(
+                tokenizer_ready_time - tokenizer_rev_request_time
```
zhanghaotong, line 700: Instead of passing it through layer by layer, could we use obj.received_time?

python/sglang/srt/entrypoints/http_server.py
Baidu-AIAK: > > Somewhat redundant with [#13432](https://github.com/sgl-project/sglang/pull/13432)
> 
> Yes, I think this variable is essentially no different from received_time, and it can also be used for embedding requests placed within the serving_base.

yes, it's redundant. I will reuse `received_time`

python/sglang/srt/managers/scheduler_metrics_mixin.py
Baidu-AIAK: > split this big function into multiple smaller ones

I already splitted  `log_decode_stats` into sub `log_decode_*`

python/sglang/srt/metrics/collector.py
llllllooooox: @sufeng-buaa Hello. Yes, firstly, the granularity of the time-consuming data collection is indeed quite fine-grained, which has helped us pinpoint some difficult performance issues in our production environment. Secondly, in overlap mode, these values ​​are not entirely accurate, and we haven't found a good way to accurately collect these metrics.  Currently, in our actual usage scenarios, we have compared and analyzed these metrics with nsys-profile data, and they generally match the end-to-end time of the profile (under conditions where the batch size is not frequently and significantly changed). This still helps us quickly identify some performance problems and is of great reference value. For now, we can use this method to statistically measure the time spent on this part, and we welcome any suggestions for better implementation methods.

PR #15866, opened 2025-12-26
[feat] reduce overhead when TP and constrained-decoding are enabled

ghost:
From my understanding, the issue https://github.com/sgl-project/sglang/issues/13809 asks the following: 

When TP and constrained decoding are enabled, avoid redundant work across TP ranks by:
1. Applying the grammar token mask only on rank 0
2. Broadcasting the next token id from rank 0 when grammars are present in the batch
3. Keeping the old code path when there’s no grammar

In: https://github.com/sgl-project/sglang/pull/13947 a few cases were mentioned (I solve them in this pr, but they dont seem to be solved in 13947):
1. “What happens when tp>1 and grammar_backend is None?”
2. Can we “compile grammar only on rank 0; other ranks use a dummy/no grammar backend”? Yes. achieved by adding extra broadcast.
3. Can we change “`all_reduce(max)` to a broadcast for ready-count synchronization because rank 0 and rank>0 logic are different.” Yes. Switched all_reduce to broadcast in move_ready_grammar_requests.

In this PR:
- in model_runner.sample we compute normal_decode = forward_batch.spec_algorithm is None and apply_vocab_mask = normal_decode and (tp_size == 1 or tp_rank == 0).
- the existing `_preprocess_logits` uses this flag. 
- i didnt change the spec decode path because it didnt seem in scope of the issue.

- in `sampler.forward`, when grammars are present and broadcast_from_rank0 is set (we pass normal_decode), rank 0 broadcasts the batch next token ids to other ranks.
- if there's no grammar, or if not in normal decode, we use all_reduce, same as before.

### DIFF
--- a/python/sglang/srt/layers/sampler.py
+++ b/python/sglang/srt/layers/sampler.py
@@ -209,28 +209,43 @@ class Sampler(nn.Module):
                 batch_next_token_ids,
             ]
 
-        if broadcast_from_rank0 and sampling_info.grammars and self.tp_size > 1:
+        self._sync_batch_token_ids(
+            batch_next_token_ids, sampling_info, broadcast_from_rank0
+        )
+        return batch_next_token_ids
+
+    def _sync_batch_token_ids(
+        self,
+        batch_next_token_ids: torch.Tensor,
+        sampling_info: SamplingBatchInfo,
+        broadcast_from_rank0: bool,
+    ) -> None:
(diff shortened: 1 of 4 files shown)


raayandhar:
hi, if you want to collaborate on this effort, can you slack me? I'm "Raayan Dhar" in the SGLang slack.

PR #15870, opened 2025-12-26
[Bug] fix piggyback load report return `None` bug

hnyls2002:
This PR fixes #15823.

Also, the server would not crash after the dp_controller is done, which should be fixed.

It only happens in mult-tokenizer mode.


whybeyoung (commented):
approved

PR #15905, opened 2025-12-26
Add a test case for crash dump

merrymercy:


python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -2056,6 +1920,107 @@ def background_task():
...
+                self.dump_requests_before_crash()
+                break
+
+        kill_process_tree(os.getpid(), include_parent=True)
+        sys.exit(0)
+
+    def force_exit_handler(self):
+        """Put some custom force exit logic here."""
+        pass
+
```
jimmy-evo, line 2023: is `force_exit_handler` used for "resources collect and close", like close files? if so, why not just call before `kill_process_tree`

PR #15908, opened 2025-12-26
feat: add --admin-api-key for finer-grained endpoint auth

alphabetc1:
## Motivation
Previously, there was only a single **“global API key” mechanism**: if server_args.api_key was set, auth was enabled and applied almost uniformly to all routes (except `/health*`, `/metrics*`, and OPTIONS). Anyone with that one api_key could call **sensitive management endpoints** like `/update_weights`, `/set_internal_state`, `/abort_request`, etc.

BTW, the current gateway already supports a more comprehensive authorization model https://github.com/sgl-project/sglang/pull/15850, and future auth logic can eventually be centralized there.  
However, there are still some existing production services running **without** a gateway, so we still need a simple and effective auth mechanism to prevent users from touching **high-risk endpoints**.

We could also rely on network-level whitelist for control, but right now:
- The number of endpoints is large
- Endpoint may sets differ between versions
This makes both whitelist- and blacklist-based approaches hard to maintain.

## Modifications
This patch introduces `--admin-api-key` for **sensitive management endpoints** and defines a clear three-tier authorization model:
* Normal routes
* Optional-auth routes (admin auth applied if configured)
* Force-auth routes (require admin, or are denied)

Behavior by sglang's configuration:
- **Only `api_key` set**  
  - Behavior remains identical to the legacy “global API key” gate.
- **Only `admin_api_key` set**  
  - Normal routes can stay open (no `api_key` required).  
  - Optional/force admin routes require `admin_api_key`.  
  - Force-auth routes are rejected if `admin_api_key` is not configured.
- **Both `api_key` and `admin_api_key` set**  
  - Normal routes require `api_key`.  
  - Optional/force admin routes require `admin_api_key`.  
  - Admin operations are clearly separated from normal user access.

**Note:** When `admin-api-key` is **not** configured, all authorization behavior remains exactly the same as it is today.

The patch also adds a small, testable auth decision function and a middleware to enforce these rules, plus a focused unit test that exercises the authorization matrix and validates compatibility with the previous behavior.

## Checklist


xiezhq-hermann:
this can be a very useful feature guarding endpoint APIs, how do folks think? @CatherineSue @ispobock @JustinTong0323 @slin1237

python/sglang/srt/utils/auth.py
```diff
@@ -0,0 +1,112 @@
...
+      or with api_key/admin_api_key depending on server config.
+    - force-auth: requires admin_api_key when configured; if admin_api_key is NOT configured,
+      it must be rejected (403) even if api_key is provided.
+    """
+    if method == "OPTIONS":
+        return AuthDecision(allowed=True)
+    if path.startswith("/health") or path.startswith("/metrics"):
+        return AuthDecision(allowed=True)
+
+    def _is_bearer(token: str) -> bool:
```
slin1237:
```python
import secrets

def _check_bearer_token(authorization_header: Optional[str], expected_token: str) -> bool:
    """Check bearer token with constant-time comparison."""
    if not authorization_header:
        return False
    parts = authorization_header.split(" ", 1)
    if len(parts) != 2 or parts[0].lower() != "bearer":
        return False
    return secrets.compare_digest(parts[1], expected_token)
```

python/sglang/srt/utils/auth.py
```diff
@@ -0,0 +1,112 @@
...
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import AbstractSet, Optional
+
+from fastapi.responses import ORJSONResponse
+
+
+@dataclass(frozen=True)
```
slin1237, line 14:
```python
@dataclass(frozen=True)
class AuthDecision:
    allowed: bool
    error_status_code: int = 401  # Only meaningful when allowed=False
```

python/sglang/srt/utils/auth.py
```diff
@@ -0,0 +1,112 @@
...
+    Path categories:
+    - normal: any path not in admin_optional_auth_paths or admin_force_auth_paths
+    - optional-auth: can be accessed without any key (if no keys configured),
+      or with api_key/admin_api_key depending on server config.
+    - force-auth: requires admin_api_key when configured; if admin_api_key is NOT configured,
+      it must be rejected (403) even if api_key is provided.
+    """
+    if method == "OPTIONS":
+        return AuthDecision(allowed=True)
+    if path.startswith("/health") or path.startswith("/metrics"):
```
slin1237, line 100: `startswith` for health/metrics but exact matching for admin paths:
This could cause confusion. consider normalizing or documenting the expectation clearly.

python/sglang/srt/entrypoints/http_server.py
```diff
@@ -158,6 +157,38 @@
 HEALTH_CHECK_TIMEOUT = int(os.getenv("SGLANG_HEALTH_CHECK_TIMEOUT", 20))
 WAIT_WEIGHTS_READY_TIMEOUT = int(os.getenv("SGLANG_WAIT_WEIGHTS_READY_TIMEOUT", 120))
 
+# Endpoint auth categories (enforced by middleware).
```
slin1237:
```python
def auth_level(level: AuthLevel):
    """Mark endpoint with auth level (stored in route metadata)."""
    def decorator(func):
        func._auth_level = level
        return func
    return decorator
```

then use
```python
@auth_level(AuthLevel.ADMIN_OPTIONAL)
```

slin1237 (changes requested):
overall, this is useful
please address those comments
lgtm

python/sglang/srt/entrypoints/http_server.py
alphabetc1: done

python/sglang/srt/utils/auth.py
alphabetc1: done

python/sglang/srt/utils/auth.py
alphabetc1: done

python/sglang/srt/utils/auth.py
alphabetc1: startswith is kept for backward compatibility with existing logic, and the behavior has now been documented.

slin1237 (approved):

PR #12115, opened 2025-10-25
Fix Illegal Instruction/IMA errors when using DP attention -- num_tokens_for_logprob calculation

YAMY1234:
## Motivation

Finding the root fix for https://github.com/sgl-project/sglang/pull/12052

Issue: When running with --enable-dp-attention, the DP gather operation in logits stage uses incorrect metadata, causing Triton kernel to access out-of-bounds memory.

```
AssertionError: src OOB: sz=661, chunk_size=7168, required=4738048, src.numel=21504
```

Root cause: The scheduler calculates num_tokens_for_logprob incorrectly by always assuming all tokens need logits computation, but when return_logprob=False, LogitsProcessor only constructs pruned_states with the last token per request (batch_size tokens). This mismatch causes DP gather to expect 661 tokens but only receive 3 tokens, leading to out-of-bounds memory access.

## Modifications

Key debug log before crash:
```

================================================================================
ERROR at DP1 TP1
Exception: src OOB: sz=661, chunk_size=7168, required=4738048, src.numel=21504
================================================================================
hidden_states.shape: (1111, 7168)
local_hidden_states.shape: (3, 7168)
hidden_states.dtype: torch.bfloat16
local_hidden_states.dtype: torch.bfloat16
hidden_states.device: cuda:1
local_hidden_states.device: cuda:1

--- LogitsMetadata ---
global_dp_buffer_len: 1111
dp_local_start_pos: 67
dp_local_num_tokens: 661
global_num_tokens_gpu: [67, 899, 54, 63, 67, 80, 61, 58]
global_num_tokens_for_logprob_gpu: [67, 661, 54, 63, 67, 80, 61, 58]
global_num_tokens_for_logprob_cpu: [67, 661, 54, 63, 67, 80, 61, 58]

--- Batch Info ---
extend_seq_lens_cpu: [253, 291, 355]
extend_logprob_start_lens_cpu: [62, 58, 118]
extend_logprob_pruned_lens_cpu: False
extend_return_logprob: False

--- Calculated Values ---
local_hidden_states.shape[0]: 3
max(extend_len - start_len, 1) per req: [191, 233, 237]
sum of above: 661
================================================================================

```

`sglang/python/sglang/srt/managers/scheduler.py`

In `prepare_mlp_sync_batch_raw()`, distinguish two cases when calculating `num_tokens_for_logprob` for extend mode:
[...]

PR #15922, opened 2025-12-27
[scheduler] fix: correcting `extend_logprob_start_len` calculation

ch-wan:
## Motivation

The original calculation for `extend_logprob_start_len` was incorrect ([code](https://github.com/sgl-project/sglang/blob/656f4d69a1bc8bc9c2addd1a33b0409b454136e5/python/sglang/srt/managers/schedule_batch.py#L1509)). When kv retraction happens, it should be determined based on `len(fill_ids)` (the total number of original input ids and output ids) when `return_logprob` is `False` but `logprob_start_len` is set as `len(origin_input_ids) - 1`.

## Modifications

We set `logprob_start_len` as -1 by default when `return_logprob` is `False` so that `extend_logprob_start_len` can be set as `len(fill_ids) - 1` during `prepare_for_extend`. We also reverted changes in #12115 as it's no longer needed.

## Accuracy Tests, Benchmarking and Profiling

Tests in #9748 and #12115 are reproduced for verifying the robustness, accuracy and performance of this PR.

Test `test_logprobs.py`:

```
max Δ=0
mean Δ=0
logprobs returned for 500 samples (expected: 500)

✅ Test completed successfully!
```

## Checklist


python/sglang/srt/managers/scheduler_dp_attn_mixin.py
```diff
@@ -121,18 +121,24 @@ def prepare_mlp_sync_batch_raw(
...
+            max(extend_len - logprob_start_len, 1)
+            for logprob_start_len, extend_len in zip(
+                local_batch.extend_logprob_start_lens,
+                local_batch.extend_lens,
             )
-        else:
-            # When return_logprob = False, only need last token per request
-            num_tokens_for_logprob = local_batch.batch_size()
+        )
+        if not local_batch.return_logprob:
```
ch-wan: it should be removed before this PR getting merged

python/sglang/srt/managers/scheduler_dp_attn_mixin.py
```diff
@@ -121,18 +121,24 @@ def prepare_mlp_sync_batch_raw(
...
+            # sanity check
+            if num_tokens_for_logprob != local_batch.batch_size():
+                print(
+                    f"extend_logprob_start_lens={local_batch.extend_logprob_start_lens}"
+                )
+                print(f"extend_lens={local_batch.extend_lens}")
+            assert (
```
merrymercy: add a test case so the old code will fail / crash

python/sglang/srt/managers/scheduler.py
```diff
@@ -1524,24 +1524,23 @@ def handle_generate_request(
...
+                # For prefill-only requests with logprob_start_len == -1, set logprob_start_len
+                # beyond input sequence to skip input logprob computation entirely
                 req.logprob_start_len = len(req.origin_input_ids)
-            else:
-                # TODO: For text generation, evaluate setting logprob_start_len to len(req.origin_input_ids) as well
+            elif recv_req.return_logprob:
+                # If return_logprob is True, return the logprobs for output tokens by default
                 req.logprob_start_len = len(req.origin_input_ids) - 1
+            else:
+                # If return_logprob is False, only the last token requires logprob computation
```
sundar24295s, line 1536: In the previous code, we didn't propogate `logprob_start_len = -1` beyond the Scheduler. The value was reset to `len(req.origin_input_ids) - 1`. Hence, for prefill-only, we followed the same way.

Given, with this change we are propogating `logprob_start_len = -1` into scheduler_batch, we can do the same logic for `prefill_only` as well instead of resetting to `len(req.origin_input_ids)` in line: 1531.

But, looks like,  the current change doesn't break the prefill-only logprobs computation, so we are good.

PR #15925, opened 2025-12-27
[diffusion] chore: remove useless params

yhyang201:
## Motivation

1. There is no practical need to use `--override-transformer-cls-name`.
2. The `--mode` option,  sgl-diffusion supports inference mode only.
3. For `--distributed-executor-backend`, there is currently only a single available backend: `mp`.
4. The `--output-type` option supports only one output format, `pli`.

The only uncertainty in the code changes is whether
5. `--enable-stage-verification` should be removed.

My view is that verification can be enabled by default, and the current usage frequency of this option is likely low. It is also unclear under what circumstances disabling verification would be preferable (possibly for higher performance). However, the performance impact of verification should be limited, so I have removed this option in the current changes.

@mickqian 

## Checklist

### DIFF
--- a/python/sglang/multimodal_gen/runtime/server_args.py
+++ b/python/sglang/multimodal_gen/runtime/server_args.py
@@ -145,31 +145,6 @@ def _sanitize_for_logging(obj: Any, key_hint: str | None = None) -> Any:
         return "<unserializable>"
 
 
-class ExecutionMode(str, Enum):
-    """
-    Enumeration for different pipeline modes.
-
-    Inherits from str to allow string comparison for backward compatibility.
-    """
-
-    INFERENCE = "inference"
-
-    @classmethod
-    def from_string(cls, value: str) -> "ExecutionMode":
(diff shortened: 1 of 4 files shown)


python/sglang/multimodal_gen/runtime/server_args.py
```diff
@@ -266,9 +232,6 @@ class ServerArgs:
 
     scheduler_port: int = 5555
 
-    # Stage verification
-    enable_stage_verification: bool = True
```
mickqian, line 270: How much overhead will it introduce, can you do some tests? if they are not negligible, we should turn it off be default

yhyang201: 1. At present, because the default value is set to `True`, stage verification is enabled regardless of whether `--enable-stage-verification` is specified. As a result, this parameter is currently ineffective.
2. Enabling or disabling this feature has minimal impact on performance. The detailed metrics are provided below.

mickqian (approved):

PR #15943, opened 2025-12-27
Fix temp_prefill_info assertion error in PP disaggregation mode

harvenstar:
In pipeline parallelism disaggregation mode, log_prefill_stats_late was not being called after processing prefill batches, causing temp_prefill_info to remain set and trigger an assertion error on the next iteration.

This fix adds the missing log_prefill_stats_late call after batch processing to properly clear temp_prefill_info.

## Checklist

### DIFF
--- a/python/sglang/srt/disaggregation/prefill.py
+++ b/python/sglang/srt/disaggregation/prefill.py
@@ -523,6 +523,7 @@ class SchedulerDisaggregationPrefillMixin:
                 )
 
         self.maybe_send_health_check_signal()
+        self.log_prefill_stats_late(batch)
 
     def process_disagg_prefill_inflight_queue(
         self: Scheduler, rids_to_check: Optional[List[str]] = None


fzyzcjy:
oops I saw the wrong one, I thought you mean the one in screenshot (which is reasonable), but may I know why you add one extra in process batch results?

image

harvenstar:
> oops I saw the wrong one, I thought you mean the one in screenshot (which is reasonable), but may I know why you add one extra in process batch results?
> 
> image

The reason for the extra call is to make the temp_prefill_info lifecycle consistent in PP disaggregation.

In PP mode, process_prefill_chunk() can populate temp_prefill_info, and get_new_batch_prefill() calls log_prefill_stats(...) which asserts temp_prefill_info is None. If we do not consume and clear the temp state right after get_new_batch_prefill(), it can carry over to the next iteration and trigger the assertion. Putting log_prefill_stats_late(batch) in the PP event loop ensures it always runs, and matches the non-PP ordering.

There is still a CI run reproducing the same assertion on main (20535260036/job/58994087455), so this change is meant to clear temp_prefill_info reliably before the next get_new_batch_prefill().

harvenstar:
Lastest CI report 
image

@fzyzcjy I opened a new PR that moves this change to a more appropriate location (right after get_new_batch_prefill() in the PP event loop), and removes the old call site.

Could you please take a quick look and let me know if this placement makes sense to you?

PR #15946, opened 2025-12-27
Move log_prefill_stats_late to correct location in PP mode

harvenstar:
Address review feedback from #15943.

  Move `log_prefill_stats_late()` call from `prefill.py:526` to `scheduler_pp_mixin.py:224` (right after `get_new_batch_prefill()`).

  The original location was incorrect because `event_loop_overlap_disagg_prefill()` doesn't create new batches. The new location matches the calling pattern in non-PP mode (`prefill.py:334`).

## Checklist

### DIFF
--- a/python/sglang/srt/disaggregation/decode.py
+++ b/python/sglang/srt/disaggregation/decode.py
@@ -937,8 +937,7 @@ class SchedulerDisaggregationDecodeMixin:
         # 2. decode + prebuilt -> decode + idle (idle forward, prebuilt returns)
         # 3. prebuilt + None -> None (None forward, prebuilt returns) + None
         # 4. prebuilt + decode + None -> idle (idle forward, prebuilt returns) + decode + idle
-        if self.require_mlp_sync:
-            ret = self.prepare_mlp_sync_batch(ret)
+        ret = self.maybe_prepare_mlp_sync_batch_and_log_stats(ret)
 
         if ret:
             trace_event_batch("schedule", ret.reqs)
--- a/python/sglang/srt/disaggregation/prefill.py
+++ b/python/sglang/srt/disaggregation/prefill.py
@@ -325,13 +325,10 @@ class SchedulerDisaggregationPrefillMixin:
         self.process_prefill_chunk()
 
         batch = self.get_new_batch_prefill()
-        if self.require_mlp_sync:
-            batch = self.prepare_mlp_sync_batch(batch)
+        batch = self.maybe_prepare_mlp_sync_batch_and_log_stats(batch)
 
         if batch:
             trace_event_batch("schedule", batch.reqs)
-
-        self.log_prefill_stats_late(batch)
 
         return batch
 
@@ -523,7 +520,6 @@ class SchedulerDisaggregationPrefillMixin:
                 )
 
         self.maybe_send_health_check_signal()
-        self.log_prefill_stats_late(batch)
 
     def process_disagg_prefill_inflight_queue(
         self: Scheduler, rids_to_check: Optional[List[str]] = None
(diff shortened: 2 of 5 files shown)


ShangmingCai (commented):
Why there are many fixes of this `log_prefill_stats_late` recently? I am seeing like 3 PRs are addressing this problem. If this is need for `event_loop_pp_disagg_prefill`, should `event_loop_pp` also add this line?

harvenstar:
> > @ShangmingCai Why there are many fixes of this log_prefill_stats_late recently? I am seeing like 3 PRs are addressing this problem. If this is need for event_loop_pp_disagg_prefill, should event_loop_pp also add this line?
> 
> 1. I introduced that for a useful metric in DP attention mode, but as is suggested by @hnyls2002 I do not run CI but use CI on main to save resources
> 2. thus the CI expectedly fails in several places and thus there are fixes to this
> 3. `If this is need for event_loop_pp_disagg_prefill, should event_loop_pp also add this line?` - generally speaking every `.log_prefill_stats` (called inside `get_new_batch_prefill`) should be paired with one `.log_prefill_stats_late` after prepare_mlp_sync_batch is called, since the "late" function exists b/c I need information obtained in prepare_mlp_sync_batch for metrics, thus yes I guess so. cc @harvenstar .
>    
>    * Another way may be somehow add a maybe_prepare_mlp_sync_batch and include this, based on how many places we changed (not realize this when I firstly impl it since lack of CI) this may be reasonable.

@fzyzcjy Thanks for the clarification. I took a look at the calling patterns:

event_loop_pp (around line 85):
self.mbs[mb_id] = self.get_next_batch_to_run()
get_next_batch_to_run() already pairs log_prefill_stats (from get_new_batch_prefill()) with log_prefill_stats_late() (after prepare_mlp_sync_batch()), so I suppose no change is needed there.

event_loop_pp_disagg_prefill (around line 220-224):

batch = self.get_new_batch_prefill()  # log_prefill_stats called here
if self.require_mlp_sync:
    batch = self.prepare_mlp_sync_batch(batch)
self.log_prefill_stats_late(batch)    # this PR adds the pairing


This path directly calls the lower-level functions, so it needs the manual pairing. This PR fixes that.

So event_loop_pp is already correct. The missing pairing was only in event_loop_pp_disagg_prefill (and get_next_disagg_prefill_batch_to_run, which was addressed in #15943).

On the refactoring suggestion, I am happy to add a helper like maybe_prepare_mlp_sync_batch_and_log_stats() to reduce duplication. Would you prefer that change to be included in this PR, or should I send it as a small follow-up PR to keep this one minimal?

fzyzcjy:
> @fzyzcjy: Another way may be somehow add a maybe_prepare_mlp_sync_batch and include this, based on how many places we changed (not realize this when I firstly impl it since lack of CI) this may be reasonable.
> @harvenstar: Would you prefer that change to be included in this PR, or should I send it as a small follow-up PR to keep this one minimal?

I guess just refactor it here is enough

fzyzcjy:
shall we change all, instead of some, of the calls

image

harvenstar:
> shall we change all, instead of some, of the calls
> 
> 
> 
> image
> 
> 

sure!

ShangmingCai (approved):
Thx. The logic looks very clear now.

ShangmingCai:
cc: @ch-wan

ch-wan (approved):
LGTM

fzyzcjy (approved):
LGTM, and we should add back the temporarily removed assertion again once everything is good

fzyzcjy:
I guess ensure not break ci then everything is ok (but here is dilemma - we do not want to use precious ci resource on this pr but w/ that we are hard to verify it is correct...)

PR #15995, opened 2025-12-28
[Refactor] Clean up custom op

DarkSharpness:
## Motivation

We observe that there're many direct use `torch.ops.sglang` in code, which is not friendly to IDE (especially for type checking). This is a result of custom op registration. Currently, registering a custom op requires:

1. Define the custom op `example_op`.
2. Define a fake implementation.
3. Call `direct_register_custom_op`.
4. After that, we can only use `torch.ops.sglang.example_op`.

The fake impl is mainly for torch compile compatibility and should return a tensor of correct shape. However, in most cases, the shape of the result is either `None` (in-place ops) or the same as one input tensor.

As a result, we propose to use a `register_custom_op` decorator to simplify this. This could be very useful in making new JIT kernel compatible with `torch.compile`.

## Modifications

1. Add a decorator `register_custom_op` to simplify the registration.
2. Clean up most custom op registration.
3. Remove`supports_custom_op`. Currently, torch version in sglang dependency is greater than 2.4.0, which already supports custom op. This function will always return True, so we remove it.

## Checklist

### DIFF
--- a/python/sglang/jit_kernel/utils.py
+++ b/python/sglang/jit_kernel/utils.py
@@ -1,25 +1,9 @@
 from __future__ import annotations
 
 import functools
-import inspect
 import pathlib
 from functools import lru_cache
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    Callable,
-    List,
-    Optional,
-    Tuple,
-    TypeAlias,
-    TypeVar,
-    Union,
-    overload,
-)
-
-import torch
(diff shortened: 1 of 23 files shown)


BBuf (approved):
Great work.

merrymercy:
Guideline: For all newly added kernels, enforce inplace ops and no fake registration

PR #15997, opened 2025-12-28
fix(dp): Add SIGQUIT propagation and child process monitoring to DataParallelController

ishandhanani:
## Summary

When running sglang with data parallel workers (`dp_size > 1`), if a scheduler process dies unexpectedly (e.g., killed via `kill -9` or crashes), the parent `Engine` process remains alive indefinitely instead of exiting. This prevents proper cleanup and restart by orchestration systems.

## Root Cause

The `DataParallelController` (DPC) had two issues:

1. **No SIGQUIT propagation**: When a scheduler sends SIGQUIT to its parent (the DPC), the DPC used the default signal handler which just terminates the process without propagating the signal to the Engine.

2. **No child process monitoring**: The DPC's `event_loop()` is an infinite `while True` loop with no mechanism to detect when child scheduler processes die. When a scheduler is killed with SIGKILL, it cannot send SIGQUIT (SIGKILL is uncatchable), leaving the DPC unaware of the death.

### Process Hierarchy
```
Engine Process (python main)
  └── DataParallelController (DPC)
        ├── scheduler_DP0_TP0
        ├── scheduler_DP0_TP1
        └── ...
  └── detokenizer
```

### 1. Added SIGQUIT handler to DPC
When the DPC receives SIGQUIT from a child scheduler, it now propagates the signal to the parent Engine process before exiting.

### 2. Added child process monitoring in `event_loop()`
The DPC now periodically checks (every 5 seconds) if its scheduler child processes are still alive. If any scheduler dies, it raises a `RuntimeError` which triggers the exception handler to send SIGQUIT to the parent Engine.

### Reproduction Steps
1. Start sglang with disaggregated prefill/decode (`dp_size=2`)
2. Wait for model to load
3. Kill a scheduler process: `kill -9 <scheduler_pid>`

### Before Fix
```bash
# logs hang here on the ranks
[2025-12-28 05:47:54 DP1 TP1] Scheduler hit an exception: Traceback (most recent call last):
  File "/home/ubuntu/sglang/python/sglang/srt/managers/scheduler.py", line 2739, in run_scheduler_process
    scheduler.event_loop_overlap_disagg_decode()
  File "/home/ubuntu/dynamo/.venv/lib/python3.12/site-packages/torch/utils/_contextlib.py", line 120, in decorate_context
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/sglang/python/sglang/srt/disaggregation/decode.py", line 812, in event_loop_overlap_disagg_decode
    recv_reqs = self.recv_requests()
                ^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/sglang/python/sglang/srt/managers/scheduler.py", line 1122, in recv_requests
    control_reqs = broadcast_pyobj(
                   ^^^^^^^^^^^^^^^^
  File "/home/ubuntu/sglang/python/sglang/srt/utils/common.py", line 1229, in broadcast_pyobj
    dist.broadcast(tensor_size, src=src, group=dist_group)
  File "/home/ubuntu/dynamo/.venv/lib/python3.12/site-packages/torch/distributed/c10d_logger.py", line 81, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/dynamo/.venv/lib/python3.12/site-packages/torch/distributed/distributed_c10d.py", line 2830, in broadcast
    work.wait()
RuntimeError: [/pytorch/third_party/gloo/gloo/transport/tcp/pair.cc:535] Read error [172.31.44.124]:29350: Connection reset by peer
```

### After Fix
```bash
  File "/home/ubuntu/dynamo/.venv/lib/python3.12/site-packages/torch/distributed/distributed_c10d.py", line 4056, in all_gather_into_tensor
    work.wait()
RuntimeError: [/pytorch/third_party/gloo/gloo/transport/tcp/pair.cc:538] Read error [172.31.44.124]:13997: Connection reset by peer

2025-12-28T06:10:28.996700Z ERROR data_parallel_controller.sigquit_handler: DataParallelController received SIGQUIT from child. Propagating to parent Engine process.
2025-12-28T06:10:28.997210Z ERROR tokenizer_manager.running_phase_sigquit_handler: SIGQUIT received. signum=None, frame=None. It usually means one child failed.
```

## Checklist
- [x] Fix handles SIGQUIT propagation from DPC to Engine
- [x] Fix handles child process death detection via periodic monitoring
- [x] Tested with `kill -9` on scheduler process
- [x] Parent process exits within seconds of child death


python/sglang/srt/managers/data_parallel_controller.py
```diff
@@ -583,6 +571,19 @@ def run_data_parallel_controller_process(
...
+    # Set up SIGQUIT handler to propagate to parent
+    # This ensures that if a scheduler sends SIGQUIT to DPC, we propagate it to Engine
+    def sigquit_handler(signum, frame):
+        logger.error(
+            "DataParallelController received SIGQUIT from child. "
+            "Propagating to parent Engine process."
+        )
+        parent_process.send_signal(signal.SIGQUIT)
+        # Use os._exit to avoid running any cleanup that might hang
+        os._exit(1)
```
merrymercy, line 583: Should we remove this line?
This is because when the Engine gets SIGQUIT, it will call `kill_process_tree` to clean up all the child processes. If we remove this line, the `kill_process_tree` can correctly finish the cleaning.
If we keep this line, `kill_process_tree` won't be able to kill all scheduler processes because DPC is dead and engine process cannot find the child scheduler processes anymore.

python/sglang/srt/managers/data_parallel_controller.py
```diff
@@ -571,6 +549,16 @@ def event_loop(self):
                     break
                 self._request_dispatcher(recv_req)
 
+            current_time = time.time()
```
merrymercy, line 552: We should create a watchdog thread instead of adding code into the event loop.

You can reuse this class 
https://github.com/sgl-project/sglang/blob/be3828a13b32820c21965946b630d0fc632b9c85/python/sglang/srt/managers/scheduler_runtime_checker_mixin.py#L356-L363

you can pass in your custom alive function (which is `all(proc.is_alive() for proc in self.scheduler_procs)`)

merrymercy (changes requested):

PR #16043, opened 2025-12-29
optimize get_topk_ragged by fusing get k and k_scale triton kernel

BJWang-ant:
- **Summary**

Optimize the get_index_k_scale_buffer function to reduce the number of concatenation operations of k_fp8 and k_scale.
image


BJWang-ant:
@Fridge003 Please help me review code. Thanks!

test/manual/layers/attention/nsa/test_get_k_scale_triton_kernel.py
```diff
@@ -0,0 +1,253 @@
+import torch
+import triton
+import triton.language as tl
+
+
+@triton.jit
+def _get_k_and_s_triton_kernel(
```
Fridge003: Can we import this kernel from srt?
Since if anybody modify this kernel later, it's hard to keep the two copies synced

python/sglang/srt/mem_cache/memory_pool.py
```diff
@@ -1782,8 +1782,10 @@ def get_index_k_scale_continuous(
     def get_index_k_scale_buffer(
```
Fridge003, line 1837: Can we open a new method for fused kernel (`get_fused_index_k_scale_buffer`)
This one can be a correctness baseline

python/sglang/srt/layers/attention/nsa/index_buf_accessor.py
```diff
@@ -592,32 +605,44 @@ def _get_k_and_s_triton(
...
+    s_out = torch.empty((seq_len_sum, 4), dtype=torch.uint8, device=buf.device)
+
+    _, buf_numel_per_page = buf.shape
+    _, page_indice_batch_offset = page_indices.shape
+    s_offset_in_page = page_size * index_head_dim
 
     # Launch kernel with one thread per token
-    grid = (seq_len,)
+    seq_num = seq_lens.shape[0]
+    grid = (seq_num, max_seq_len)
```
Fridge003, line 646: Are we launching too many blocks for this grid?
Like when seq_len is really long, this might be inefficient.
Can we handle multiple tokens (such as 128 or 256) for each block?

(or maybe we can leave this for next PR)

test/manual/layers/attention/nsa/test_get_k_scale_triton_kernel.py
BJWang-ant: OK.No problem.I will rewrite later.

python/sglang/srt/mem_cache/memory_pool.py
BJWang-ant: ok. No problem

python/sglang/srt/layers/attention/nsa/index_buf_accessor.py
BJWang-ant: Yeah.I agree to postpone it for the next PR.

python/sglang/srt/mem_cache/memory_pool.py
Fridge003: Oh wait... The former `get_index_k_scale_buffer` can be deprecated since it's only called once in nsa_indexer.py.
Then I think no need to open a new method. The current PR is OK

Fridge003:
@BJWang-ant Please fix the conflict

test/manual/layers/attention/nsa/test_get_k_scale_triton_kernel.py
Fridge003: cc @BJWang-ant  Please resolve this

BJWang-ant: OK

Fridge003 (approved):

BJWang-ant:
> Hi, after this PR was merged, DeepSeek-v3.2 encounters the following error when running with a long context (> 65,536 tokens): `Triton Error [HIP]: Code: 1, Messsage: invalid argument` This occurs because the Triton kernel grid is configured as `grid = (seq_num, max_seq_len)`. When max_seq_len exceeds 65,536, the kernel launch fails since HIP does not support launching more than 65,536 blocks in the Y dimension. I suspect that CUDA GPUs may also have the same limitation and may not support launching more than 65,536 blocks in the Y dimension either.
> 
> Server:
> 
> ```
> python3 -m sglang.launch_server \
>     --model /data/huggingface/hub/deepseek-ai/DeepSeek-V3.2-Exp \
>     --mem-fraction-static 0.7 --tp 8 --port 8552 --trust-remote-code \
>     --disable-radix-cache --chunked-prefill-size 131072 \
>     --nsa-prefill-backend tilelang --nsa-decode-backend tilelang
> ```
> 
> Client:
> 
> ```
> python3 -m sglang.bench_serving \
>     --port  8552 \
>     --backend sglang \
>     --model /data/huggingface/hub/deepseek-ai/DeepSeek-V3.2-Exp  \
>     --dataset-name random \
>     --random-input 65537 \
>     --random-output 600 \
>     --random-range-ratio 1.0 \
>     --num-prompts 1 \
>     --max-concurrency 1
> ```

I will push code either today or tomorrow.

Fridge003:
Hi @BJWang-ant, I also found this PR will cause errors on some extremely long context. So we need to revert this PR temporarily. Please combine this PR with fix together in the next PR.

BJWang-ant:
> Hi @BJWang-ant, I also found this PR will cause errors on some extremely long context. So we need to revert this PR temporarily. Please combine this PR with fix together in the next PR.

OK. Couldo you please give me some bad case?

xu-yfei:
@BJWang-ant The original code has a bug in the CP (context parallel) scenario. Could you please include this fix as well?
https://github.com/sgl-project/sglang/pull/18280

BJWang-ant:
> @BJWang-ant The original code has a bug in the CP (context parallel) scenario. Could you please include this fix as well? #18280

OK

PR #16068, opened 2025-12-29
[bugfix]: add check, stream_options could only be set when stream=true

Bruce-x-1997:
## Motivation

I meet a bad req, that includes stream_options but it's stream=false. And we find model-gateway returns a 400 response, but sglang engine could just handle this req. I read [openai api doc](https://platform.openai.com/docs/api-reference/chat/create#chat_create-stream_options), and it seems that 400 is an expected response. So I fix this bug.

## Modifications

just add a check that stream_options could only be set when stream=true.
and fix one related bench script.

## Checklist


Bruce-x-1997:
@slin1237 could you help trigger it again, only 3 failing jobs now

PR #16077, opened 2025-12-29
[http-server] Add kv cache info to server_info endpoint

peleg-yair:
## Motivation

Add kv cache information to model_info endpoint.
Closes #16075

The information consists of:

- max kv cache tokens
- kv cache size in bytes
- kv cache size in GB

## Modifications

- Add `kv_cache_size_bytes` to the dict sent to `pipe_writer` in `python/sglang/srt/managers/scheduler.py`
- Read it in `python/sglang/srt/entrypoints/http_server.py`, and add as a dict in `model_info`.

## Checklist


python/sglang/srt/entrypoints/http_server.py
```diff
@@ -547,6 +547,24 @@ async def model_info():
         "model_type": getattr(model_config.hf_config, "model_type", None),
         "architectures": getattr(model_config.hf_config, "architectures", None),
     }
+
+    # Add KV cache information from scheduler_info if available
+    if _global_state.scheduler_info:
```
slin1237: this is hardware specific, not model related?
because this could change
is this already part of metrics? and server_info?

slin1237 (changes requested):

python/sglang/srt/entrypoints/http_server.py
slin1237: can we do that instead?
add to `server_info` and expose it to metrics

peleg-yair: It is run-specific. It does relate to the model and the hardware in the way that each combination will have another amount of kv cache tokens\size but it is not all as you can control it by passing `--mem-fraction-static 0.8` as you wish.

I thought it was a better fit in model_info as it strongly-related to the model but now I do see that `scheduler_info` is already referenced on `server_info` so I can have it there in addition to the metrics. 

Will update soon enough (:

peleg-yair: @slin1237 I added a metric and it is not contained in server_info, please review again
Thanks!

peleg-yair: @slin1237

PR #15216, opened 2025-12-16
Refactor: separate CI-specific weight validation into dedicated module

alisonshao:
## Summary

- Rename `weight_validation.py` to `ci_weight_validation.py` to clearly indicate its CI-specific purpose
- Add wrapper functions (`ci_validate_and_cleanup_local_snapshot`, `ci_validate_weights_after_download`) that encapsulate CI validation and cleanup logic
- Simplify `weight_utils.py` by removing inline CI logic and using lazy imports gated by `is_in_ci()` checks
- `is_in_ci()` now appears only twice in `weight_utils.py`

**For users:** Simple download without validation overhead  
**For CI:** Automatic validation, cleanup, and retry on corruption

## Test plan

- [ ] CI tests pass
- [ ] Local testing as user (non-CI environment) to verify download works without overhead

PR #16121, opened 2025-12-30
ci: enable offline mode when local cache is complete to avoid HF Hub …

harvenstar:
## Motivation

  CI tests frequently experience timeouts due to HuggingFace Hub network instability. When HF Hub has transient network issues (e.g., RemoteDisconnected on HEAD requests for tokenizer_config.json), server initialization fails with /health_generate returning 503, leading to popen_launch_server timing out after 600s.

  This PR addresses the root cause by enabling HF_HUB_OFFLINE mode when local cache is verified to be complete, avoiding unnecessary network requests that can fail.

  ## Changes

  ### 1. Cache Validation Logic (ci_weight_validation.py)
  - Added validation for config files: config.json, tokenizer_config.json, generation_config.json
  - Added validation for tokenizer files: tokenizer.json, tokenizer.model, tiktoken.model
  - Integrated with existing weight validation from PR #15216 for safetensors files
  - All validations check file existence, non-empty size, and JSON parsing where applicable

  ### 2. Offline Mode Enablement (test_utils.py)
  - Added _try_enable_offline_mode_if_cache_complete() helper function
  - Before launching server subprocess, validate local cache completeness
  - If validation passes, set HF_HUB_OFFLINE=1 in subprocess environment
  - Smart weight file collection: prioritizes .safetensors.index.json, falls back to recursive glob

  ### 3. Safety Guards
  - Require at least one weight file to enable offline mode
  - HF snapshot path sanity check (must contain "snapshots")
  - Weight file count limit (max 1000) to avoid excessive scanning
  - Double guards in both helper and validation functions

  ### 4. Logging
  - All logs use CI_OFFLINE: prefix for easy grepping in GitHub Actions
  - Logs include: cache validation status, HF_HUB_OFFLINE value, model name, snapshot info
  - Neutral wording: "skip offline, keep online allowed" instead of "failed"

  ## Test Plan

  - Run existing CI tests to verify offline mode is enabled when cache is complete
  - Search for "CI_OFFLINE:" in CI logs to verify behavior
  - Verify that tests with incomplete cache still work (online mode)
  - Check that weight_files guard prevents false positives

  ## Related

  Based on PR #15216 which added CI-specific weight validation infrastructure.


harvenstar:
## Solution Overview:

  The changes enable HF_HUB_OFFLINE mode when local cache is verified as complete, eliminating unnecessary network requests. Key components include:

  1. **Cache Validation** – Checks config files (config.json, tokenizer_config.json, generation_config.json), tokenizer files, and weight files for existence and validity

  2. **Snapshot-Level Marker Caching** – Addresses the issue of redundant validation across tests:
     - **Problem**: Initial approach validated cache on every test launch, causing repeated overhead
     - **Solution**: Cache validation results per snapshot using markers in `/tmp`
     - Marker keyed by `sha256(snapshot_dir)` to prevent collisions
     - Only successful validations are cached (failures retry to allow downloads)
     - Atomic writes via `os.replace` for cross-platform consistency
     - **Result**: Each snapshot validated at most once per runner, subsequent tests skip validation

  3. **Offline Mode Enablement** – Sets HF_HUB_OFFLINE=1 in subprocess environment only (not globally) when cache is complete

  4. **Logging** – Uses "CI_OFFLINE:" prefix with marker filenames for traceability

PR #16134, opened 2025-12-30
[diffusion] lora: Fix/qkv fuse lora breakage

MikukuOvO:
## Motivation

Some diffusion pipelines/models use **fused projections** (e.g., `attn.to_qkv` / `attn.to_added_qkv`, and fused SwiGLU `ffn.w13`). However, many public LoRAs (including Qwen-Image / Z-Image / Lightning-style LoRAs) are authored for **split projections** (e.g., `to_q/to_k/to_v`, `add_q_proj/add_k_proj/add_v_proj`, `w1/w3`).

When the target model expects fused modules but the LoRA provides split keys, plain name-format normalization is not enough: LoRA keys may not match the target layers (becoming ineffective), or weight injection can fail with **rank/shape mismatches** (especially when exact fusions expand effective rank).

This PR adds **target-aware LoRA preparation + exact fusing** and a **dynamic-rank injection fallback** so split-projection LoRAs can be reliably applied to fused targets without breaking inference.

### 1) `lora_format_adapter.py`
- **Target-aware LoRA preparation** via `prepare_lora_state_dict_for_target(...)`:
  - Normalize external LoRA formats into canonical `.lora_A/.lora_B`.
  - Detect fused targets (QKV/add_qkv, SwiGLU) from target layer names.
  - Perform **exact fusion** when the target is fused but the LoRA is split:
    - QKV/add_qkv: `q/k/v → to_qkv / to_added_qkv`
    - FFN: `w1/w3 → w13`
    - Shared `A`: concatenate `B`; different `A`: block-diagonal exact fusion.
- **Robust injection** via `set_lora_weights_dynamic_rank(...)`:
  - Try `set_lora_weights(...)` first; on rank/shape mismatch, dynamically replace `lora_A/B` (inference-safe).
- **Correct scaling**:
  - Bake `(alpha / rank)` into `lora_B` and drop `.alpha`.

### 2) `lora_pipeline.py`
- **Load-time preparation**:
  - Collect target LoRA layers across components and call `prepare_lora_state_dict_for_target(...)`.
- **Application path**:
  - Use `set_lora_weights_dynamic_rank(...)` instead of direct setting to handle fused targets and rank mismatches.

## Compatibility notes

- If the target does not use fused modules (or the LoRA already matches fused keys), no fusing is performed and behavior remains unchanged.
- If the canonical `layer.set_lora_weights(...)` accepts the provided shapes, the dynamic-rank fallback is not triggered.

## Accuracy Tests

This PR affects LoRA weight mapping/injection logic. I ran local functional tests focusing on:
- no crashes / no shape assertion failures
- LoRA effectively applied (qualitative output changes as expected)
- logs indicate target-aware fuse behavior where applicable

### ✅ Local tests executed

1. **Qwen-Image + `lightx2v/Qwen-Image-Lightning-4` & `8steps-V2.0-bf16`**
2. **Z-Image + `Z-Image-Turbo-DeJPEG-Lora/dejpeg_v3`**
3. **Qwen-Edit + `lightx2v/Qwen-Image-Edit-2511-Lightning/Qwen-Image-Edit-2511-Lightning-4steps-V1.0-fp32`**

## Benchmarking and Profiling

This PR does not affect runtime inference performance. Normalization occurs only once at LoRA load time and has negligible overhead.

## Checklist

- [ ] Provide accuracy and speed benchmark results according to the [Test the accuracy](https://docs.sglang.io/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.io/developer_guide/contribution_guide.html#benchmark-the-speed).

### DIFF
--- a/python/sglang/multimodal_gen/runtime/pipelines_core/lora_format_adapter.py
+++ b/python/sglang/multimodal_gen/runtime/pipelines_core/lora_format_adapter.py
@@ -364,7 +364,11 @@ def fuse_qkv_like_keys_inplace(
     skipped = 0
     a_mismatch = 0
     for q, k, v, fused_name in patterns:
-        pat = re.compile(rf"^(?P<prefix>.+\.attn)\.{re.escape(q)}\.lora_A\.weight$")
+        # Support both "...attn.*" and "...attention.*"
+        pat = re.compile(
+            rf"^(?P<prefix>.+\.(?:attn|attention))\.{re.escape(q)}\.lora_A\.weight$"
+        )
+
         prefixes = sorted(
             {
                 m.group("prefix")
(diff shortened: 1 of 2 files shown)


MikukuOvO:
@mickqian PR is ready for review.

python/sglang/multimodal_gen/runtime/pipelines_core/lora_format_adapter.py
```diff
@@ -21,9 +22,767 @@ class LoRAFormat(str, Enum):
...
+    if Ak.dtype != dtype:
+        Ak = Ak.to(dtype)
+    if Av.dtype != dtype:
+        Av = Av.to(dtype)
+    if Bq.dtype != dtype:
+        Bq = Bq.to(dtype)
+    if Bk.dtype != dtype:
+        Bk = Bk.to(dtype)
+    if Bv.dtype != dtype:
+        Bv = Bv.to(dtype)
```
niehen6174: The validation and conversion process here could probably be simplified using a single function to reduce the amount of code.

MikukuOvO:
> The amount of code here seems quite large; can it be simplified?

Okay! I'll take a look and try.

MikukuOvO:
> The amount of code here seems quite large; can it be simplified?

Already simplified the code, and retest the function locally to make sure it works.

PR #14091, opened 2025-11-28
vlm: Refactor engine vlm params and support precessor output as input

minleminzui:
# Summary

This PR refactors how the offline **Engine** handles multimodal / VLM inputs and exposes a consistent `image_data` API across models.

---

## Motivation

Previously, each VLM had its own ad-hoc way to pass images (raw pixel values, custom dicts, etc.), and multi-image requests could easily break the engine path.

We want a single, well-defined `image_data` contract that works for:

- raw images (e.g., `PIL`, `numpy`, `torch`)
- HuggingFace processor outputs
- precomputed visual embeddings

---

### Engine API

- Clarified and unified the supported formats for `image_data` in `Engine.generate` / `async_generate`, including:
  - **Plain images:**  
    `image_data=[image]` or `image_data=[[image1, image2, ...]]`
  - **HuggingFace processor outputs:**  
    `image_data=[dict(processor_output, format="processor_output")]`
  - **Precomputed embeddings:**  
    `image_data=[dict(processor_output, format="precomputed_embedding", feature=precomputed_embeddings)]`

- Centralized multimodal validation / normalization in the scheduler (`mm_utils`, `schedule_batch`)  
  so batched and multi-image requests follow the same path.

---

### Model / Processor Side

- Updated **Qwen2.5-VL** model wrapper and multimodal processors to use the new `MultimodalInputFormat` helper instead of model-specific dicts.  
- Fixed the engine multi-image bug reproduced in `refactor-engine-vlm-params` — the engine now correctly handles multiple images per request.

---

### Tests

Added `test_vlm_input_format.py`, which verifies that both **Qwen2.5-VL** and **Gemma-3-VLM** work with:

- direct `image_data=[PIL.Image]`
- processor output as `image_data`
- precomputed embeddings passed via `format="precomputed_embedding"`

It also checks that the model can correctly understand a **2-image input** (e.g., *taxi + SGL logo*) for all three code paths.

---

### Docs

Updated [`docs/advanced_features/vlm_query.ipynb`](docs/advanced_features/vlm_query.ipynb) to:

- describe the three supported calling patterns (Basic / Processor Output / Precomputed Embeddings)
- show concrete examples for **Qwen2.5-VL** using the offline `Engine` API

---

## Testing

```bash
pytest test/srt/test_vlm_input_format.py

✅ All 6 tests pass on Qwen2.5-VL-3B-Instruct and Gemma-3-4B-IT with multimodal enabled.

PR #16152, opened 2025-12-30
[VLM][Reland] Refactor load_mm_data to improve performance

yuan-luo:
## Motivation

Per discussed with @JustinTong0323 , we do need this enhancement as a part for token-in and token-out implementation, request to re-land https://github.com/sgl-project/sglang/pull/14644.

## Checklist

## Review Process

### DIFF
--- a/python/sglang/srt/multimodal/processors/base_processor.py
+++ b/python/sglang/srt/multimodal/processors/base_processor.py
@@ -418,6 +418,47 @@ class BaseMultimodalProcessor(ABC):
         except Exception as e:
             raise RuntimeError(f"Error while loading data {data}: {e}")
 
+    def _submit_mm_data_loading_tasks_simple(
+        self,
+        data_list: Optional[list],
+        modality: Modality,
+        audio_sample_rate: Optional[int],
+        discard_alpha_channel: bool,
+    ) -> List[Tuple[Modality, int, concurrent.futures.Future]]:
+        """
+        Simple version: For one modal data submit IO load task.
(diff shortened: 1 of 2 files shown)


yuan-luo:
In SGLang, the essence of --skip-tokenizer-init is that it splits the responsibility of “turning a prompt into model-usable inputs” and outsources part of it to the client.

skip-tokenizer-init = the server does not load a tokenizer (and therefore does not accept text prompts). It only accepts already-tokenized input_ids (or an even lower-level form like input_embeds), and then runs backend inference directly.

The current implementation, is “multimodal” which is more tricky than LLM case:

- When skip-tokenizer-init is enabled: the client sends input_ids
- At the same time, it still allows / requires sending image_data
- The server-side mm_data_processor still performs an additional multimodal step via process_mm_data_async

This creates the core contradiction we ran into:
The client has already expanded image-related tokens into input_ids (commonly a large number of image_pad tokens), but the server decodes input_ids back into text, then uses the processor to realign again in a “text + images” manner.

As a result, the “expanded image tokens” get misinterpreted as “multiple image placeholders,” causing alignment to fail (out-of-bounds).

Given the analysis, the fix is to fallback to legacy slow path for skip-tokenizer-init.

yuan-luo:
This PR has some impact on https://github.com/sgl-project/sglang/pull/14091. 
@minleminzui Could you help to review this PR as well? Thanks.

JustinTong0323 (approved):

PR #16168, opened 2025-12-30
Refactor speculative algorithm registry.

hnyls2002:
Prev code is overly complex and provides no practical benefit for the current design.

- Speculative algorithms are static in practice and do not require a dynamic registry.
- The set of speculative algorithms is small and well-defined, and can be safely hardcoded.
- The lock is unnecessary, as there is no runtime mutation or concurrent registration.
- The alias mechanism is unused in practice, since `NEXTN` is already normalized to `EAGLE` in server args.
- The tag-based design is redundant; existing semantics are fully covered by `is_eagle()`, which naturally includes `is_eagle3()`.
- `_registration_order` is unused, as we never iterate over speculative algorithms and there are only a few fixed entries.

**If private forks need additional customization**, they can add explicit enum values or auxiliary dicts/lists alongside the enum, without introducing a global dynamic registry.


python/sglang/srt/managers/scheduler.py
```diff
@@ -534,10 +529,16 @@ def init_draft_worker(self):
                     dp_rank=self.dp_rank,
                 )
         else:
-            self.draft_worker = self.spec_algorithm.create_draft_worker(
-                **draft_worker_kwargs
+            WorkerClass = self.spec_algorithm.create_worker(
```
ispobock, line 532: use SpecWorkerClass since there are two model workers (target and draft) here?

hnyls2002: Oh, this is just a local variable. I will optimize the draft model initialization in the following PRs.

hnyls2002:
All spec algorithms' CI are hit.

image

PR #16211, opened 2025-12-31
Download missing shards in model weights files when not in CI

Kangyan-Zhou:
This logic was removed in favor of CI missing weight file check, add it back to avoid confusing errors when the mode files are partially downloaded.


Fridge003:
https://github.com/sgl-project/sglang/actions/runs/20615702372/job/59207972592

Fridge003 (approved):

PR #16286, opened 2026-01-02
Fix: ensure device consistency in alloc_for_extend for paged allocation

Simon-Li:
## Motivation
Fix #14160. 

Fix `CUDA error: invalid argument` in `alloc_for_extend()` during paged allocation.

The error occurs at `torch.cat(last_loc)` when `prefix_indices` tensors are on 
different devices (CPU vs GPU). This can happen because:

1. `Req.prefix_indices` is initialized as a CPU tensor in `Req.__init__()`
2. Various code paths update `prefix_indices` to GPU tensors via `match_prefix()`
3. However, device consistency is not explicitly enforced when building `last_loc`

## Modifications

Add explicit device transfer to ensure all tensors are on `batch.device`:

```python
t[-1:].to(batch.device, non_blocking=True)
```
This is a defensive fix that guarantees device consistency regardless of prefix_indices's current device.

After run a hot-fix in our prods and the issue was no longer occurred.

## Accuracy Tests

N/A - This fix only addresses tensor device placement and does not affect model computation or outputs.

## Checklist

- [X] Update documentation according to [Write documentations](https://docs.sglang.io/developer_guide/contribution_guide.html#write-documentations).
- [X] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.io/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.io/developer_guide/contribution_guide.html#benchmark-the-speed).
- [X] Follow the SGLang code style [guidance](https://docs.sglang.io/developer_guide/contribution_guide.html#code-style-guidance).


Simon-Li:
@merrymercy Pls have a look at it.

cctry:
> > Is this the root cause? should pytorch raise error for concatenating two tensors with different device?
> 
> This is likely an initialization issue as said, it's not occurred with the fix.
> 
> * Req.prefix_indices is initialized as a CPU tensor in Req.**init**()
> * Various code paths update prefix_indices to GPU tensors via match_prefix()

Hmm, my opinions are:
1. `invalid argument` comes from calling CUDA runtime API with wrong argument, while `torch.cat` should not call them. (I am assuming `CUDA_LAUNCH_BLOCKING=1` is set)
2. I generally advise against silently moving tensors to GPU because it can mask real device-placement bugs elsewhere.

PR #15812, opened 2025-12-25
[Diffusion] Improve qwen image edit performace to align with LightX2V

BBuf:
## main

```shell
sglang generate --model-path Qwen/Qwen-Image-Edit-2511 --prompt "Change the person to a standing position, bending over to hold the dog's front paws."  --image-path "/home/lmsys/bbuf/LightX2V/examples/qwen_image/1.png"
```

```shell
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 40/40 [00:30<00:00,  1.29it/s]
[12-25 07:11:20] [DenoisingStage] average time per step: 0.7722 seconds
[12-25 07:11:20] [DenoisingStage] finished in 30.8956 seconds
[12-25 07:11:20] [DecodingStage] started...
[12-25 07:11:20] [DecodingStage] finished in 0.5390 seconds
[12-25 07:11:20] Output saved to outputs/Change_the_person_to_a_standing_position_bending_over_to_hold_the_dog_s_front_paws._20251225-071047_08c591d4.jpg
[12-25 07:11:20] Pixel data generated successfully in 33.13 seconds
```

图片

## pr

```shell
sglang generate --model-path Qwen/Qwen-Image-Edit-2511 --prompt "Change the person to a standing position, bending over to hold the dog's front paws."  --image-path "/home/lmsys/bbuf/LightX2V/examples/qwen_image/1.png"
```

```shell
[12-25 07:00:34] [DenoisingStage] started...
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 40/40 [00:25<00:00,  1.58it/s]
[12-25 07:00:59] [DenoisingStage] average time per step: 0.6327 seconds
[12-25 07:00:59] [DenoisingStage] finished in 25.3114 seconds
[12-25 07:00:59] [DecodingStage] started...
[12-25 07:01:00] [DecodingStage] finished in 0.5667 seconds
```

图片

### use upstream fa3 , not sgl-kernel fa3 : 1.7ms->1.2ms

main:

b5f74563-ea0e-40fd-8416-f447ec1805fe

pr:

d1ae0892-8e3a-4044-a468-6daa162450e8

###  flashinfer rope: 241us->82us

main

图片

pr:

图片

 
### revert pack qkv to avoid unaligned cat kernel:   154.5us->70us

main:

图片

pr:

图片

### fuse_scale_shift_gate_select01_kernel: 148us->66us。

main:

图片

pr:

图片

[...]

PR #16287, opened 2026-01-02
fixed amd multimodal CI failures caused by refactor in #15812 #15813

sunxxuns:
fixed the failure initially caused by #15812 , which was merged without passing mm CI on AMD (pls avoide this future on)
image
1 fixed flashinfer call, since it's only supporting nvidia gpu now, should not be on the main path for other hardwares
2 attn_mask && is_causal doesn't work for sdpa/rocm, flux actually didn't use attn_mask anyway;  fixed it for amd path and added a warning
also 2-gpu is broken by #15813 
for lora weight load path.


Qiaolin-Yu (commented):
Could you please fix the lint? Thanks!

PR #16311, opened 2026-01-02
[diffusion] Add CI validation for multimodal model downloads

alisonshao:
## Summary

Fixes recurring CI failures in diffusion tests caused by corrupted/incomplete HuggingFace cache.

**Example failure:** https://github.com/sgl-project/sglang/actions/runs/20620327597/job/59220900139?pr=16213#step:5:4686

The issue: CI tests were failing with `FileNotFoundError` when safetensors shards were missing from subdirectories like `transformer_2/`. The existing validation only checked for top-level directories (`transformer/`, `vae/`, `model_index.json`) but not for missing shards within those directories.

**Root cause:** The file `transformer_2/diffusion_pytorch_model-00009-of-00012.safetensors` was missing from the HF cache, but the index file still referenced it. This caused:
1. Customized loader to fail with `ValueError` (missing parameters)
2. Native fallback to fail with `FileNotFoundError`

### Changes

- Add `_check_index_files_for_missing_shards()` - validates all subdirectories by checking safetensors index files and verifying all referenced files exist
- Add `_cleanup_model_cache()` - removes corrupted HF cache directories to force clean re-download  
- Add `_ci_validate_diffusers_model()` - CI-specific validation entry point
- Integrate validation into `maybe_download_model()` at three points:
  - After checking local path exists
  - After finding model in cache
  - After downloading from HF Hub

The validation is gated behind `is_in_ci()` so regular users won't experience any overhead.

## Test plan

- [ ] CI tests pass (specifically diffusion tests that were previously failing)
- [ ] Validation correctly detects missing shards and cleans up cache
- [ ] Re-download succeeds after cache cleanup

### DIFF
--- a/python/sglang/multimodal_gen/runtime/utils/hf_diffusers_utils.py
+++ b/python/sglang/multimodal_gen/runtime/utils/hf_diffusers_utils.py
@@ -19,8 +19,10 @@
 """Utilities for Huggingface Transformers."""
 
 import contextlib
+import glob
 import json
 import os
+import shutil
 import time
 from functools import reduce
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


mickqian (approved):
thanks sooo much!

PR #16329, opened 2026-01-03
[Diffusion] safe fallback for fused QK-norm

RubiaCx:
## Motivation

The fused QK-norm kernel used here was introduced in #16062, and this PR ensures it does not break on non-applicable layouts or ROCm (AMD CI) while keeping QK-norm behavior consistent when enabled.
1272abe947e43703ca5e5f99c10c5e47

## Modifications

Remove `_is_cuda()` gating at callsites; always call `apply_qk_norm` when` self.qk_norm` is enabled.

In `apply_qk_norm`, only call the fused kernel on NVIDIA CUDA (q.is_cuda and torch.version.cuda is not None) and valid layouts; if it raises "QK-norm is not applicable", fall back to the RMSNorm path; re-raise other errors.

## Checklist

### DIFF
--- a/python/sglang/multimodal_gen/runtime/layers/layernorm.py
+++ b/python/sglang/multimodal_gen/runtime/layers/layernorm.py
@@ -434,17 +434,19 @@ def apply_qk_norm(
 ) -> Tuple[torch.Tensor, torch.Tensor]:
     """Apply QK normalization for query and key tensors.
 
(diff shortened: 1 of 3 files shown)


RubiaCx:
Already sloved by https://github.com/sgl-project/sglang/pull/16287, closed.

PR #16345, opened 2026-01-03
Fix parallel tool call parsing bug when tool parameters contain arrays

Leoyzen:
## Motivation

This PR fixes a critical bug in parallel tool call parsing where the parser incorrectly identifies `[` characters inside array parameters as the start of a new tool call when processing the second and subsequent tools in a parallel tool call sequence.

**Bug Scenario:**
When a model outputs multiple parallel tool calls in JSON array format (e.g., `tool_choice='required'`), and both tools have array parameters like `"title": ["some value"]`, the parser would fail on the second tool with the error:
```
AttributeError: 'str' object has no attribute 'get'
```

This happens because the parser finds the `[` character in the second tool's array parameter (e.g., `"title": ["4.8"]`) and incorrectly treats it as the `bot_token` marking the start of a new tool array, instead of correctly identifying the tool separator comma.

### 1. Fixed parsing logic in [`base_format_detector.py`](python/sglang/srt/function_call/base_format_detector.py:168-183)

**Root Cause:** The if-elif logic checked for `bot_token` ('[') before checking if we're processing a subsequent tool (`current_tool_id > 0`). This caused the parser to find '[' inside array parameters and mistakenly identify it as the start of a new tool.

**Fix:** Reordered the if-elif logic to prioritize checking for the tool separator when `current_tool_id > 0`. This ensures that when processing subsequent tools in parallel calls, we correctly identify the tool separator comma first, rather than being misled by '[' characters in parameter values.

**Code Change:**
```python
# Before (incorrect):
tool_call_pos = current_text.find(self.bot_token)
if tool_call_pos != -1:
    start_idx = tool_call_pos + len(self.bot_token)
elif self.current_tool_id > 0 and current_text.startswith(self.tool_call_separator):
    start_idx = len(self.tool_call_separator)

# After (correct):
if self.current_tool_id > 0 and current_text.startswith(self.tool_call_separator):
    # Priority check for subsequent tools
    start_idx = len(self.tool_call_separator)
else:
    # Only search for bot_token if not processing subsequent tool
    tool_call_pos = current_text.find(self.bot_token)
    if tool_call_pos != -1:
        start_idx = tool_call_pos + len(self.bot_token)
```

### 2. Added comprehensive test coverage in [`test_parallel_tool_bug.py`](test/registered/function_call/test_parallel_tool_bug.py)

- `test_parallel_tool_calls_with_array_parameters`: Tests realistic streaming chunks where both tools have array parameters
- `test_simple_parallel_tool_calls`: Minimal test case for the bug scenario

Both tests verify that:
- No parsing errors occur
- Both tools are correctly parsed
- Tool names and parameters match expected values

## Accuracy Tests

N/A - This is a bug fix for the tool call parsing logic, not a change to model outputs.

## Benchmarking and Profiling

N/A - This fix does not impact inference speed. It only corrects the parsing logic to handle parallel tool calls with array parameters correctly.

## Checklist

  - Added 2 new tests in `test_parallel_tool_bug.py`
  - All existing 96 tests in `test_function_call_parser.py` continue to pass
  - No documentation update needed (bug fix)
  - N/A - bug fix does not affect model accuracy or speed

## Review Process


test/registered/function_call/test_parallel_tool_calls.py
Kangyan-Zhou, line 1: Can you follow the new setup below to register this new test? Also please change the filename to be more intuitive (_bug is not meaningful in this context)

from sglang.test.ci.ci_register import register_cuda_ci

register_cuda_ci(est_time=xxx, suite="stage-a-test-1")

Leoyzen: @Kangyan-Zhou Done.

Leoyzen:
@JustinTong0323 @Kangyan-Zhou Could you take a look at this PR?

JustinTong0323 (approved):
LGTM, thanks for finding it and fixing it~

PR #16369, opened 2026-01-04
fix【hicache】fix the KV cache resource occupation  and invalid loading from prefetch when pending requests are aborted.

CLFutureX:
**Background**
Currently, when processing pending scheduled requests ， if the number of requests in the waiting queue exceeds the set threshold, the system will interrupt low-priority requests and return directly based on the priority policy. However, the preempted and aborted low-priority requests may have triggered corresponding prefetch operators. These uninterrupted prefetch operations will continue to occupy KV cache resources, resulting in invalid occupation of cache space and reduced overall resource utilization.
path:python/sglang/srt/managers/scheduler.py#_abort_on_queued_limit

**Optimization**
Interrupt Corresponding Prefetch Requests When Pending Scheduled Requests Are Aborted to Release KV Cache Resources


python/sglang/srt/managers/scheduler.py
```diff
@@ -1746,6 +1699,8 @@ def _abort_on_queued_limit(self, recv_req: Req) -> bool:
                 direction * recv_req.priority < direction * candidate_req.priority
             )
             if abort_existing_req:
+                if self.enable_hierarchical_cache:
+                    self.tree_cache.terminate_prefetch()
```
xiezhq-hermann: did you forget to pass request id?

CLFutureX: > did you forget to pass request id?

yeah, adjusted already

xiezhq-hermann (approved):

PR #16370, opened 2026-01-04
【hicache】Optimize HiCache prefetch logic: adapt to remaining available memory when memory is insufficient

CLFutureX:
#### Background
In the original prefetch logic of HiCache's `prefetch_from_storage` method, if the requested prefetch memory (`prefetch_length`) cannot be allocated, the prefetch operation is directly abandoned even if the remaining available memory still meets the prefetch threshold. This leads to waste of memory resources and reduces prefetch efficiency.

#### Changes Made
After the failure of `host_indices` allocation, the following logic is added:
1. Get the available size of the current memory pool (`avaliable_size`);
2. Calculate the new prefetch length (`prefetch_length`) aligned with the page size;
3. If the new length ≥ prefetch threshold, adjust the input tokens length and reallocate memory;
4. Only release the host and return (keep the original logic) if the threshold is not met.


python/sglang/srt/mem_cache/hiradix_cache.py
```diff
@@ -730,9 +730,17 @@ def prefetch_from_storage(
...
+                new_input_tokens = new_input_tokens[:prefetch_length]
+                host_indices = self.cache_controller.mem_pool_host.alloc(
+                    prefetch_length
+                )
+            else:
+                last_host_node.release_host()
+                # no sufficient host memory for prefetch
+                return
         operation = self.cache_controller.prefetch(
             req_id, host_indices, new_input_tokens, last_hash, prefix_keys
```
hzh0425: If the prefetch length is changed, will the last_hash and prefix_keys no longer match here?

CLFutureX: > If the prefetch length is changed, will the last_hash and prefix_keys no longer match here?

I understand they don't interfere with each other. This is because last_hash and prefix_keys depend on the nodes that have already been matched earlier: req.fill_ids[:matched_len].
In contrast, the prefetch_length here targets the subsequent unmatched nodes:new_input_tokens = req.fill_ids[matched_len:]

hzh0425: Looks good;

hzh0425 (approved):

python/sglang/srt/mem_cache/hiradix_cache.py
hzh0425: could you take a review  @xiezhq-hermann

xiezhq-hermann (approved):

PR #13592, opened 2025-11-19
[FEAT] optimize tensor zmq transfer for multimodal inputs

liusy58:
## Motivation

This PR optimizes the transfer of multimodal inputs containing large tensors from tokenizer to scheduler by replacing `send_pyobj`/`recv_pyobj` with a zero-copy transfer mechanism. This significantly reduces serialization/deserialization overhead for vision-language models.

## Benchmarking and Profiling

* To launch a server

```
CUDA_VISIBLE_DEVICES=2,3 python3 -m sglang.launch_server --model-path /mnt/models/models--Qwen--Qwen2.5-VL-3B-Instruct/ --enable-torch-compile --disable-radix-cache --host 0.0.0.0 --port 34000 --trust-remote-code --tp-size 2 --log-level info --max-running-requests 128 --mem-fraction-static 0.85 --chunked-prefill-size 32000 --attention-backend fa3 --disable-cuda-graph
```

* To bench

```
NUM_PROMPTS=10
DATASET=image # mmmu
REQUEST_RATE=0.05
IMAGE_COUNT=10
IMAGE_RES=720p
INPUT_LEN=100
HOST=`hostname -I | awk '{print $1}'`
PORT=34000
MODEL=/mnt/models/models--Qwen--Qwen2.5-VL-3B-Instruct/
python -m sglang.bench_serving \
    --model $MODEL \
    --num-prompts $NUM_PROMPTS \
    --dataset-name $DATASET \
    --random-input-len $INPUT_LEN \
    --random-output-len 1 \
    --warmup-requests 1 \
    --image-count $IMAGE_COUNT \
    --image-resolution $IMAGE_RES \
    --host 0.0.0.0 \
    --port 34000 \
    --backend vllm-chat \
    --request-rate $REQUEST_RATE

```

This can reduce the transport cost from 400ms to 70ms for requests with large feature tensors in mm_inputs.
## Checklist

- [x] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
[...]

PR #16386, opened 2026-01-04
Revert "[FEAT] optimize tensor zmq transfer for multimodal inputs"

yhyang201:
Reverts sgl-project/sglang#13592

The code quality needs improvement. We could adopt an approach that requires fewer changes, avoids modifying key interfaces, and still achieves the same performance.

@merrymercy

### DIFF
--- a/python/sglang/srt/managers/data_parallel_controller.py
+++ b/python/sglang/srt/managers/data_parallel_controller.py
@@ -16,7 +16,6 @@
 import faulthandler
 import logging
 import multiprocessing as mp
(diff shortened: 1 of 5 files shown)


ShangmingCai:
Thank you for the timely response. Changes to the _manager_ dir or scheduler-related code need to be handled carefully and approved by the reviewer before merging, even if all CIs have passed. My bad, sorry for not double-checking the review status.

We should ping hnyls2002 next time, he should be available for this kind of final check.

PR #14717, opened 2025-12-09
[diffusion] kernel fusion: gated residual layernorm scale shift and layernorm scale shift kernel fusion for Qwen-Image, WAN and HunyuanVideo

jianyingzhu:
## Motivation

Profiling shows a lot of GPU bubbles, which can be mitigated via kernel fusion. 

This PR:
1. fused residual, gating, LayerNorm, and scale/shift into one single CUDA kernel named `fused_scale_residual_layernorm_scale_shift`
2. fused LayerNorm, and scale/shift into one single CUDA kernel named `fused_layernorm_scale_shift`. 

The kernel fusion reduces kernel launch overhead for Qwen-Image, WAN and HunyuanVideo pipelines.

The fused kernels support both 2D \([batch_size * seq_len, inner_dim]\), \([1, inner_dim]\), 3D \([1, 1, inner_dim]\), and 4D \([batch_size, num_frames, 1, inner_dim]\) modulation for diffusion models. Numerical parity is maintained with the existing PyTorch/Triton implementations across fp32, fp16, and bf16.

## Modifications

- In `sgl_kernel`, add the `fused_scale_residual_layernorm_scale_shift` and `fused_layernorm_scale_shift` CUDA kernels based on CUTLASS.
- In `layers/layernorm.py`, update the LayerNorm path to call the fused kernels when available.
- In the DiT models, update the Qwen-Image, WAN, and HunyuanVideo implementations to use the fused kernels in `layernorm.py`.
- Support both Layernorm and RMSNorm. 

with @AichenF and @yingluosanqian.
Special thanks to @yingluosanqian, he helped a lot.

Pass unit tests.
image

### Benchmark: fused_norm_scale_shift

| B | S    | D    | norm_type | affine | SGLang Native | CuTeDSL   | CUDA C   |
|---|------|------|-----------|--------|---------------|-----------|---------|
| 1 | 128  | 1024 | layer     | True   | 31.680000     | 7.616000  | 7.584   |
| 1 | 128  | 1024 | layer     | False  | 22.208000     | 7.536000  | 7.584   |
| 1 | 128  | 1024 | rms       | True   | 12.128000     | 7.328000  | 7.296   |
| 1 | 128  | 1024 | rms       | False  | 12.192000     | 7.104000  | 7.520   |
| 1 | 128  | 3072 | layer     | True   | 36.736000     | 8.352000  | 8.640   |
| 1 | 128  | 3072 | layer     | False  | 25.216000     | 8.352000  | 8.864   |
[...]

PR #16404, opened 2026-01-04
[diffusion]: optimize zimage with rmsnorm + add + (rmsnorm)

attack204:
# E2E IMPROMENT

10s/13% faster

```bash
python python/sglang/multimodal_gen/benchmarks/compare_perf.py baseline.json new.json
### Performance Comparison Report
```

#### 1. High-level Summary
| Metric | Baseline | New | Diff | Status |
| :--- | :--- | :--- | :--- | :--- |
| **E2E Latency** | 71805.84 ms | 62156.90 ms | **-9648.94 ms (-13.4%)** | ✅ |
| **Throughput** | 0.01 req/s | 0.02 req/s | - | - |

#### 2. Stage Breakdown
| Stage Name | Baseline (ms) | New (ms) | Diff (ms) | Diff (%) | Status |
| :--- | :--- | :--- | :--- | :--- | :--- |
| InputValidationStage | 0.06 | 0.06 | -0.01 | -9.4% | ⚪️ |
| TextEncodingStage | 929.59 | 686.98 | -242.61 | -26.1% | 🟢 |
| ConditioningStage | 0.03 | 0.03 | -0.00 | -7.0% | ⚪️ |
| TimestepPreparationStage | 2.25 | 2.12 | -0.13 | -5.7% | ⚪️ |
| LatentPreparationStage | 9.30 | 4.89 | -4.41 | -47.4% | ⚪️ |
| DenoisingStage | 69765.61 | 60430.47 | -9335.14 | -13.4% | 🟢 |
| DecodingStage | 1061.44 | 1000.04 | -61.40 | -5.8% | ⚪️ |

## DETAIL PERF

TODO

### DIFF
--- a/python/sglang/jit_kernel/csrc/diffusion/fused_norm_scale_shift.cuh
+++ b/python/sglang/jit_kernel/csrc/diffusion/fused_norm_scale_shift.cuh
@@ -1,0 +1,1816 @@
+/* Copyright 2025 SGLang Team. */
+#include <sgl_kernel/tensor.h>   // For TensorMatcher, SymbolicSize, SymbolicDevice
+#include <sgl_kernel/utils.cuh>  // For LaunchKernel
+#include <sgl_kernel/utils.h>    // For div_ceil, RuntimeCheck
+
+#include <cuda_fp16.h>
+#include <dlpack/dlpack.h>
+#include <tvm/ffi/container/tensor.h>
+#include <tvm/ffi/optional.h>
+
+#include "cutlass/numeric_types.h"
+
+namespace {
+
+namespace ffi = tvm::ffi;
+
+enum NormType : int {
+  LayerNorm = 0,
+  RMSNorm = 1,
+};
+
+template <NormType NT>
+struct NormTag {
+  static constexpr NormType value = NT;
+};
+
+template <int V>
+struct ItemPerThreadTag {
+  static constexpr int value = V;
+};
+
+template <typename T4_, typename T_>
+struct DTypeTag {
+  using T4 = T4_;
+  using T = T_;
+};
+
+// Minimal warp/block reduction helpers (sum) for small arrays
+template <typename T, int NumVals>
+__device__ __forceinline__ void warpReduceSum(T (&vals)[NumVals]) {
+  unsigned mask = 0xffffffffu;
+#pragma unroll
+  for (int offset = 16; offset > 0; offset >>= 1) {
+#pragma unroll
+    for (int i = 0; i < NumVals; ++i) {
+      vals[i] += __shfl_down_sync(mask, vals[i], offset);
+    }
+  }
+}
+
+template <typename T, int NumVals>
+__device__ __forceinline__ void blockReduceSum(T (&vals)[NumVals]) {
+  __shared__ T shared[32][NumVals];  // up to 32 warps (1024 threads)
+  int lane = threadIdx.x & 31;
+  int wid = threadIdx.x >> 5;
+  warpReduceSum<T, NumVals>(vals);
+  if (lane == 0) {
+#pragma unroll
+    for (int i = 0; i < NumVals; ++i) {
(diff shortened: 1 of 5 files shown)


python/sglang/jit_kernel/csrc/diffusion/fused_norm_scale_shift.cuh
```diff
@@ -0,0 +1,1816 @@
...
+#include <sgl_kernel/tensor.h>   // For TensorMatcher, SymbolicSize, SymbolicDevice
+#include <sgl_kernel/utils.cuh>  // For LaunchKernel
+#include <sgl_kernel/utils.h>    // For div_ceil, RuntimeCheck
+
+#include <cuda_fp16.h>
+#include <dlpack/dlpack.h>
+#include <tvm/ffi/container/tensor.h>
+#include <tvm/ffi/optional.h>
+
+#include "cutlass/numeric_types.h"
```
DarkSharpness, line 11: Can we try to eliminate this dependency? It seems only numeric types are used.

python/sglang/multimodal_gen/runtime/layers/layernorm.py
```diff
@@ -469,3 +469,140 @@ def apply_qk_norm(
     q_out = q_norm(q.view(-1, head_dim)).view(q_shape)
     k_out = k_norm(k.view(-1, head_dim)).view(k_shape)
     return q_out, k_out
+
+
+# ========== New Fused RMSNorm Classes for ZImage ==========
+
+class FusedRMSNormAddRMSNorm(nn.Module):
```
yingluosanqian: Could we use `FusedRMSNormAddRMSNorm(CustomOp)` instead? This way, we wouldn’t need to manually check `is_cuda`, since `CustomOp` already implements the dispatch logic, as shown below:
``` python
    def dispatch_forward(self) -> Callable:
        if current_platform.is_cuda():
            return self.forward_cuda
        elif current_platform.is_hip():
            return self.forward_hip
        elif current_platform.is_npu():
            return self.forward_npu
        elif current_platform.is_xpu():
            return self.forward_xpu
        else:
            return self.forward_native
```

yingluosanqian:
After this [MR](https://github.com/sgl-project/sglang/pull/16452)
 is merged, we could pass `dtype` as a C++ template parameter, eliminating the need for runtime dtype dispatch.

``` cuda
if (dtype.code == kDLFloat && dtype.bits == 32) {
    dispatch_ipt(DTypeTag<float4, float>{});
  } else if (dtype.code == kDLFloat && dtype.bits == 16) {
    dispatch_ipt(DTypeTag<half4, half>{});
  } else if (dtype.code == kDLBfloat && dtype.bits == 16) {
    dispatch_ipt(DTypeTag<bf16_4, cutlass::bfloat16_t>{});
  }
```

DarkSharpness:
What's the relationship of this PR and #14717?

It seems there's many redundant code between these 2 PRs. We may need to maximize code reuse and avoid too many similar code.

yingluosanqian:
> What's the relationship of this PR and #14717?
> 
> It seems there's many redundant code between these 2 PRs. We may need to maximize code reuse and avoid too many similar code.

We have a [kernel fusion plan](https://acntwz63wyi6.feishu.cn/wiki/XRfdwY9jFii2MKkTMGwcRJBhnuf?from=from_copylink) for diffusion kernel. Many of which involve `norm` operation. 

I summarized how we currently end up generating a large number of kernel source variants:
1. Some kernels must handle both 3D and 4D input shapes.
2. When N is smaller than a certain threshold, placing the input in shared memory yields better performance.
3. Different kernels have different fusion patterns. For example, some kernels fuse `norm + scale + shift`, while others fuse `norm + tanh + scale + shift`, or `norm + add + norm`. **This is especially problematic, and we may need a better abstraction to maximize code reuse.**

PR #16425, opened 2026-01-04
Add v1/models endpoint to diffusion model APIs so that they can be discovered by model gateway

Kangyan-Zhou:
## Motivation

- Introduce OpenAI compatible API v1/models with additional info from diffusion server
- Mark the models/ endpoint as deprecated

## Accuracy Tests

curl 207.210.107.189:30001/v1/models | jq
{
  "object": "list",
  "data": [
    {
      "id": "Wan-AI/Wan2.2-T2V-A14B-Diffusers",
      "object": "model",
      "created": 1767658791,
      "owned_by": "sglang",
      "root": "Wan-AI/Wan2.2-T2V-A14B-Diffusers",
      "parent": null,
      "max_model_len": null,
      "num_gpus": 1,
      "task_type": "T2V",
      "dit_precision": "bf16",
      "vae_precision": "fp32",
      "pipeline_name": "WanPipeline",
      "pipeline_class": "WanPipeline"
    }
  ]
}

### DIFF
--- a/python/sglang/multimodal_gen/runtime/entrypoints/openai/common_api.py
+++ b/python/sglang/multimodal_gen/runtime/entrypoints/openai/common_api.py
@@ -1,6 +1,7 @@
 from typing import Any, Optional
 
 from fastapi import APIRouter, Body, HTTPException
+from fastapi.responses import ORJSONResponse
 
 from sglang.multimodal_gen.runtime.entrypoints.openai.utils import (
     MergeLoraWeightsReq,
@@ -11,6 +12,7 @@ from sglang.multimodal_gen.runtime.pipelines_core.schedule_batch import OutputBa
 from sglang.multimodal_gen.runtime.scheduler_client import async_scheduler_client
 from sglang.multimodal_gen.runtime.server_args import get_global_server_args
 from sglang.multimodal_gen.runtime.utils.logging_utils import init_logger
+from sglang.srt.entrypoints.openai.protocol import ModelCard, ModelList
 
 router = APIRouter(prefix="/v1")
 logger = init_logger(__name__)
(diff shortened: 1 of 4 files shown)


python/sglang/multimodal_gen/runtime/entrypoints/openai/common_api.py
```diff
@@ -117,3 +119,46 @@ async def model_info():
         "model_path": server_args.model_path,
     }
     return result
+
+
+@router.get("/models", response_class=ORJSONResponse)
```
mickqian, line 136: actually we have a `get_models` in http_server.py already

Kangyan-Zhou: Yes but router uses v1/models as the standard settings, and I feel like both server interfaces should have the same APIs so that people won't get confused as I do

mickqian: in that case, I think we should delete the `get_models` and use `models` as the only endpoint, and also update the usage in https://github.com/sgl-project/sglang/blob/main/python/sglang/multimodal_gen/docs/openai_api.md#get-model-information, since two similar endpoints would cause confusion

But maybe we could do it later

BBuf: We are currently using the get_models API. Is it possible to keep it and just mark it as deprecated first?

Kangyan-Zhou: done and amend the new endpoint with diffusion related info, updated the PR description to reflect the change

mickqian (approved):

PR #16452, opened 2026-01-05
[jit kernel] support dtype as a cpp template parameter

yingluosanqian:
## Motivation

In the current jit kernel implementation, `dtype` can not be passed as a cpp template parameters, which in some cases leads to larger binary size and longer compilation time. For example, when a kernel supports multiple dtype we have to generate separate kernels for each dtype instead of generating a single templated kernel.

This PR support dtype as a cpp template parameter.

## Modifications

This PR modify `make_cpp_args` to deal with dtype template parameters. Then optimize the `norm` kernel to validate this modification.

## Accuracy Tests

run test_qknorm and passed.

```
python python/sglang/jit_kernel/tests/test_qknorm.py
```

## Benchmarking and Profiling

None.

## Checklist

None.

## Review Process

### DIFF
--- a/python/sglang/jit_kernel/csrc/norm.cuh
+++ b/python/sglang/jit_kernel/csrc/norm.cuh
@@ -146,6 +146,14 @@ struct QKNormKernel {
     auto dtype = SymbolicDType{};
     auto device = SymbolicDevice{};
 
+    /*
+     * We need the .template disambiguator here because this call happens in a dependent context.
+     * After switching to with_dtype<DType>(...) (where DType is a template parameter), the chained expression becomes
+     * dependent. In C++, when calling a member function template via ./-> on a dependent expression, the compiler may
+     * otherwise parse <kDLCUDA> as the < operator instead of template arguments. Adding .template forces correct
+     * parsing and fixes compilation errors (often seen with NVCC/clang). Ref:
+     * https://en.cppreference.com/w/cpp/language/dependent_name
+     */
     TensorMatcher({N, Q, D})  // q input
         .with_strides({Sq, D, 1})
         .with_dtype<DType>(dtype)
--- a/python/sglang/jit_kernel/csrc/norm.cuh
+++ b/python/sglang/jit_kernel/csrc/norm.cuh
@@ -119,10 +119,12 @@ __global__ void fused_qknorm(const QKNormParams __grid_constant__ params) {
   PDLTriggerSecondary<kUsePDL>();  // launch secondary kernel
 }
 
-template <int64_t kHeadDim, bool kUsePDL>
+template <int64_t kHeadDim, bool kUsePDL, typename DType>
 struct QKNormKernel {
-  template <typename PackedFloat, typename Float>
-  static constexpr auto qknorm_kernel = fused_qknorm<kHeadDim, kUsePDL, PackedFloat, Float>;
+  using DType2 = host::PackedDType<DType, 2>::type;
+
+  // only initialize once (static variable) to avoid overhead
+  static constexpr auto kernel = fused_qknorm<kHeadDim, kUsePDL, DType2, DType>;
 
   static void
   run(const tvm::ffi::TensorView q,
@@ -177,19 +179,10 @@ struct QKNormKernel {
         .num_tokens = num_tokens,
     };
 
-    // only initialize once (static variable) to avoid overhead
-    static constexpr auto bf16_kernel = qknorm_kernel<nv_bfloat162, nv_bfloat16>;
-    static constexpr auto fp16_kernel = qknorm_kernel<half2, half>;
-    static const uint32_t kMaxOccupancyTable[2] = {
-        runtime::get_blocks_per_sm(fp16_kernel, kThreadsPerBlock),
-        runtime::get_blocks_per_sm(bf16_kernel, kThreadsPerBlock),
-    };
+    static const uint32_t max_occupancy = runtime::get_blocks_per_sm(kernel, kThreadsPerBlock);
     static const uint32_t kNumSM = runtime::get_sm_count(device.unwrap().device_id);
 
     // choose kernel based on dtype
-    const bool use_bf16 = dtype.is_type<nv_bfloat16>();
-    const auto kernel = use_bf16 ? bf16_kernel : fp16_kernel;
-    const auto max_occupancy = kMaxOccupancyTable[use_bf16 ? 1 : 0];
     const auto num_works = (num_qo_heads + num_kv_heads) * num_tokens;
     const auto needed_blocks = div_ceil(num_works, kWarpsPerBlock);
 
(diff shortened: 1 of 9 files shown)


python/sglang/jit_kernel/include/sgl_kernel/utils.h
```diff
@@ -33,6 +33,8 @@
 #include "source_location.h"
 #endif
 
+#include <cuda_bf16.h>
```
DarkSharpness: Move the change in this file to `utils.cuh`. `utils.h` only contains CPU related operations (e.g. DebugInfo). For GPU related wrappers (like packed dtype, `bfloat162`), they should live in `xxx.cuh`

python/sglang/jit_kernel/include/sgl_kernel/utils.h
```diff
@@ -44,6 +46,53 @@
 
 namespace host {
 
+template <typename>
+inline constexpr bool dependent_false_v = false;
```
DarkSharpness, line 48: only exception: just keep `dependent_false_v` here. It's CPU-only op, and really useful in metaprogramming.

python/sglang/jit_kernel/csrc/norm.cuh
```diff
@@ -119,10 +119,12 @@ __global__ void fused_qknorm(const QKNormParams __grid_constant__ params) {
   PDLTriggerSecondary<kUsePDL>();  // launch secondary kernel
 }
 
-template <int64_t kHeadDim, bool kUsePDL>
+template <int64_t kHeadDim, bool kUsePDL, typename DType>
```
DarkSharpness, line 122: You should also replace `.with_dtype<nv_bfloat16, half>(dtype)` with `.with_dtype<Dtype>(dtype)` in line 148 and 153

python/sglang/jit_kernel/utils.py
```diff
@@ -54,6 +56,14 @@ def _convert(arg: CPP_TEMPLATE_TYPE) -> str:
...
         if isinstance(arg, (int, float)):
             return str(arg)
+        if isinstance(arg, (torch.dtype)):
+            if arg == torch.float:
+                return "float"
+            if arg == torch.float16:
+                return "__half"
+            if arg == torch.bfloat16:
+                return "nv_bfloat16"
+            raise TypeError(f"Not implement this arg wrapper yet: {arg}")
```
DarkSharpness, line 66: We should support more Python side type mapping in the future. But for now it's ok.

python/sglang/jit_kernel/include/sgl_kernel/utils.h
yingluosanqian: done

python/sglang/jit_kernel/csrc/norm.cuh
```diff
@@ -143,17 +148,17 @@ struct QKNormKernel {
...
-        .with_dtype<nv_bfloat16, half>(dtype)
-        .with_device<kDLCUDA>(device)
+        .with_dtype<DType>(dtype)
+        .template with_device<kDLCUDA>(device)
         .verify(k);
     TensorMatcher({D})  // weight
-        .with_dtype<nv_bfloat16, half>(dtype)
-        .with_device<kDLCUDA>(device)
+        .with_dtype<DType>(dtype)
+        .template with_device<kDLCUDA>(device)
```
DarkSharpness, line 169: why need `template` here? is this a must?

yingluosanqian: Yes. It causes a compilation error without `template`.
``` txt
error: no suitable user-defined conversion from "const tvm::ffi::TensorView" to "DLDevice" exists
          .verify(q);
```
I’m not very familiar with C++. After looking up some references, I have the following understanding:

Without template, I guess the compiler parses it as:
``` cpp
(xxx.with_device < kDLCUDA)(device)
```
Because in a dependent context (inside a template), when calling a member function template through an object expression, the compiler cannot know whether the name refers to a template or a normal member.
Therefore, the < that follows is assumed to be the less-than operator, not the start of a template argument list. To disambiguate, C++ requires the template keyword.

Some refs:
* https://en.cppreference.com/w/cpp/language/dependent_name.html?utm_source=chatgpt.com#:~:text=The%20template%20disambiguator%20for%20dependent%20names
* https://www.aerialmantis.co.uk/blog/2017/03/17/template-keywords/?utm_source=chatgpt.com

DarkSharpness: ok, looks good to me. C++ is too hard

python/sglang/jit_kernel/norm.py
```diff
@@ -28,13 +28,13 @@ def _jit_norm_module(head_dims: int) -> Module:
 
 
 @cache_once
-def can_use_fused_inplace_qknorm(head_dim: int) -> bool:
+def can_use_fused_inplace_qknorm(head_dim: int, dtype: torch.dtype) -> bool:
```
BBuf, line 31: It will block ci? Such as https://github.com/sgl-project/sglang/blob/main/python/sglang/multimodal_gen/runtime/layers/layernorm.py#L448

python/sglang/jit_kernel/utils.py
```diff
@@ -54,6 +56,14 @@ def _convert(arg: CPP_TEMPLATE_TYPE) -> str:
             return "true" if arg else "false"
         if isinstance(arg, (int, float)):
             return str(arg)
+        if isinstance(arg, (torch.dtype)):
```
BBuf:
```suggestion
        if isinstance(arg, torch.dtype):
```

python/sglang/jit_kernel/csrc/norm.cuh
BBuf: Add a comment would be better? Generated by gemini bot:

We need the `.template` disambiguator here because this call happens in a dependent context.

After switching to `with_dtype<DType>(...)` (where `DType` is a template parameter), the chained expression becomes dependent. In C++, when calling a member function template via `.`/`->` on a dependent expression, the compiler may otherwise parse `<kDLCUDA>` as the `<` operator instead of template arguments. Adding `.template` forces correct parsing and fixes compilation errors (often seen with NVCC/clang).

Ref: https://en.cppreference.com/w/cpp/language/dependent_name

DarkSharpness: Another workaround can be: call `device.set_options<kDLCUDA>()` beforehand, so that we don't need to specify the device constraint (CUDA only) again in `with_device`.

(Actually, `with_device<xxx>(device)` is just a syntax suger, which will call `set_options<xxx>` in implementation)

python/sglang/jit_kernel/utils.py
yingluosanqian: done, thanks.

python/sglang/jit_kernel/norm.py
yingluosanqian: Thanks, I’ve replaced all occurrences of `can_use_fused_inplace_qknorm` and successfully ran both QwenImage and ZImage for validation.

DarkSharpness (approved):

python/sglang/jit_kernel/include/sgl_kernel/utils.h
DarkSharpness: remove the unnecessary include

BBuf (approved):

yhyang201 (approved):

Issue #13808, opened 2025-11-23
[Roadmap] CI suites organization


hnyls2002:
## Existing Issues

- Many CI tests under `test/srt` are never executed by current workflows.
- Some contributors mistakenly added CI tests under `sglang.test`; tests in this module are not connected to any CI workflow and therefore never run.
- CI tests are tightly coupled to hardware backends (CUDA, AMD, XPU, NPU, etc.), reducing clarity and making it difficult to reuse unit tests across different backends.
- Cannot easily switch a CI test between nightly and per-commit pipelines; doing so currently requires physically moving the test files across directories.
- CI workflow pipelines and dependencies are messy and lack a proper fast-fail mechanism, leading to unnecessary CI resource usage.
- Timeout settings are not fine-grained and fail to catch most real timeout scenarios, causing severe waste of CI resources.
- Some performance tests are not managed by any suites and must be manually triggered inside CI workflow files, which limits scalability and reduces consistency.
- CI monitoring is tightly coupled to standalone summary steps and is not compatible with the `run_suite`-based triggering approach.
- There is no tracking or management of flaky tests.
- CPU-only tests are mixed with other platform tests, making it difficult to isolate test behavior and backend-specific failures.

### Workflow Pipeline Prototype

<img width="1103" height="611" alt="Image" src="https://github.com/user-attachments/assets/30862b57-fd12-4dfc-893d-b74e7f863724" />

### CI suites structure

- `test/manual`: unofficially maintained CI tests, kept mainly as code references for agents (cursor, codex), not guaranteed to run in CI.
- `test/registered` (previously `per-commit` + `srt` + `nightly`): officially maintained by the SGLang community and guaranteed to be triggered in at least one CI workflow (or temporarily disabled with clear justification).
- All test files (registered or manual) should be organized by features rather than behaviors.

### CI registry

Example for a test registered with AMD and CUDA (nightly only) workflows.

```python
register_cuda_ci(est_time=80, suite="stage-a-test-1", nightly=True)
register_amd_ci(est_time=120, suite="stage-a-test-1")
```

Example for a test that is temporarily disabled. 

```python
register_tmp_disabled(reason="flaky...")
```
(or add "tmp disabled" flag to existing registry)

Example of triggering a CI through `run_suite` API:

```bash
run_suite.py --nightly --hw cuda --suite "stage-a-test-1"
```

## Refactoring Steps

- [ ] Move tests into `test/registered` and `test/manual`, organizing them by feature.
[...]

PR #16457, opened 2026-01-05
Migrate tokenizer tests to test/registered/tokenizer/

alisonshao:
## Summary
- Creates new `test/registered/tokenizer/` directory
- Migrates 3 tokenizer-related tests: multi-tokenizer, skip init, and jinja utils
- All tests registered for `stage-b-test-small-1-gpu` suite
- AMD support included for all three tests

Part of #13808

## Test plan
- [ ] CI passes for stage-b-test-small-1-gpu (CUDA)
- [ ] CI passes for stage-b-test-small-1-gpu (AMD)
- [ ] Verify estimated times are accurate after CI run


test/registered/function_call/test_jinja_template_utils.py
Kangyan-Zhou, line 1: Rename this to jinja_template_utils_test and change it to a CPU only test since it does not load any models

Kangyan-Zhou, line 1: This should live in function_call folder

PR #16466, opened 2026-01-05
Migrate metrics tests to test/registered/metrics/

alisonshao:
Migrate metrics tests to test/registered/metrics/. test_metrics_utils.py moved to stage-a-cpu-only (pure CPU test). Part of #13808


alisonshao:
https://github.com/sgl-project/sglang/actions/runs/20739210587

PR #15356, opened 2025-12-18
Custom All Reduce for Piecewise Cuda Graph

Oasis-Git:
## Motivation

Enable Custom All Reduce in Piecewise Cuda Graph, equal contribution to @ByronHsu 

## Modifications

1. Split the compile period and warmup-capture period to make sure to replay during pcg init.
2. Enable `graph_capture` during capture 

## Accuracy Tests

```
# Server
python3 -m sglang.launch_server --model-path Qwen/Qwen3-8B --enable-piecewise-cuda-graph \
     --tp 2 \
     --piecewise-cuda-graph-max-tokens 2048

# Client
python3 sglang/benchmark/gsm8k/bench_sglang.py --num-questions 1319 --parallel 1319 --port 30000
```

```
bash client.sh 
100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1319/1319 [00:10<00:00, 127.84it/s]
Accuracy: 0.908
Invalid: 0.000
Latency: 10.396 s
Output throughput: 15629.471 token/s
```

For VL Model:
```
# Server
python -m sglang.launch_server --model Qwen/Qwen2.5-VL-7B-Instruct --tp 4 \
    --enable-piecewise-cuda-graph \
    --disable-radix-cache
    
# Client
python3 -m sglang.bench_serving \
  --backend sglang-oai-chat \
  --dataset-name image \
  --num-prompts 256 \
  --apply-chat-template \
  --random-input-len 128 \
  --random-output-len 32 \
  --image-resolution 560x560 \
  --image-format jpeg \
  --image-count 1 \
  --image-content random \
  --random-range-ratio 0.1 \
  --port 30000 \
  --max-concurrency 32
  
Created 256 random jpeg images with average 316335 bytes per request
Starting warmup with 1 sequences...
Warmup completed with 1 sequences. Starting main benchmark run...
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 256/256 [00:14<00:00, 18.11it/s]

============ Serving Benchmark Result ============
Backend:                                 sglang-oai-chat
Traffic request rate:                    inf       
Max request concurrency:                 32        
Successful requests:                     256       
Benchmark duration (s):                  14.14     
Total input tokens:                      126306    
Total input text tokens:                 23394     
Total input vision tokens:               102912    
Total generated tokens:                  4541      
[...]

PR #16489, opened 2026-01-05
[Bugfix] Remove unnecessary assert for PCG capture stream

mmdbhs:
## Motivation

After enabling --enable-piecewise-cuda-graph, if "capture new graph" is triggered during model runtime and the stream obtained by get_pcg_capture_stream() is None, an assert error will occur. However, the model can actually run normally even when the stream is None, so this assert should be removed to avoid process errors.

## Modifications

Remove unnecessary assert for PCG capture stream

## Checklist

## Review Process

### DIFF
--- a/python/sglang/srt/compilation/cuda_piecewise_backend.py
+++ b/python/sglang/srt/compilation/cuda_piecewise_backend.py
@@ -172,7 +172,6 @@ class CUDAPiecewiseBackend:
                     stack.enter_context(patch("torch.cuda.empty_cache", lambda: None))
                 # mind-exploding: carefully manage the reference and memory.
                 stream = get_pcg_capture_stream()
-                assert stream is not None, "PCG capture stream is not set"
                 with torch.cuda.graph(cudagraph, pool=self.graph_pool, stream=stream):
                     # `output` is managed by pytorch's cudagraph pool
                     output = entry.runnable(*args)


Qiaolin-Yu (commented):
It's actually necessary. We should not capture the graph during the runtime. It's caused by recompile, which only occurs in incorrect implementation.

mmdbhs:
> It's actually necessary. We should not capture the graph during the runtime. It's caused by recompile, which only occurs in incorrect implementation.

Before #15356, the stream had always defaulted to None.

Qiaolin-Yu:
> > capture the graph during the runtime
> 
> Thank you very much for your reply. I have some questions and would appreciate your clarification:
> 
> 1. During my testing of qwen3-vl-2b, I confirmed that it indeed captures the graph during runtime.
> 2. If there is an input shape that hasn't been captured before during runtime, how will the code handle it?

I think it will not enter this code path if the input shape is not in the `concrete_size_entries`? If the code reaches here, it means the shape should be captured and hasn't been captured, which is not a normal behaviour during runtime. Correct me if I'm wrong @ispobock

mmdbhs:
> > > capture the graph during the runtime
> > 
> > 
> > Thank you very much for your reply. I have some questions and would appreciate your clarification:
> > 
> > 1. During my testing of qwen3-vl-2b, I confirmed that it indeed captures the graph during runtime.
> > 2. If there is an input shape that hasn't been captured before during runtime, how will the code handle it?
> 
> I think it will not enter this code path if the input shape is not in the `concrete_size_entries`? If the code reaches here, it means the shape should be captured and hasn't been captured, which is not a normal behaviour during runtime. Correct me if I'm wrong @ispobock

However, the CUDAPiecewise graph for the prefill phase cannot exhaustively cover all possible shapes. As evident in the code, it already accounts for cases where the corresponding shape's CUDA graph is empty and captures it during runtime. This aligns with my actual testing observations.

image

mmdbhs:
> > > capture the graph during the runtime
> > 
> > 
> > Thank you very much for your reply. I have some questions and would appreciate your clarification:
> > 
> > 1. During my testing of qwen3-vl-2b, I confirmed that it indeed captures the graph during runtime.
> > 2. If there is an input shape that hasn't been captured before during runtime, how will the code handle it?
> 
> I think it will not enter this code path if the input shape is not in the `concrete_size_entries`? If the code reaches here, it means the shape should be captured and hasn't been captured, which is not a normal behaviour during runtime. Correct me if I'm wrong @ispobock

Moreover, in similar CUDA graph captures during the ViT phase, there are no restrictions imposed on the stream either.
image

mmdbhs:
> > > capture the graph during the runtime
> > 
> > 
> > Thank you very much for your reply. I have some questions and would appreciate your clarification:
> > 
> > 1. During my testing of qwen3-vl-2b, I confirmed that it indeed captures the graph during runtime.
> > 2. If there is an input shape that hasn't been captured before during runtime, how will the code handle it?
> 
> I think it will not enter this code path if the input shape is not in the `concrete_size_entries`? If the code reaches here, it means the shape should be captured and hasn't been captured, which is not a normal behaviour during runtime. Correct me if I'm wrong @ispobock

In the latest main branch, there appears to be a bug where all cudagraph entries in concrete_size_entries are empty. I will investigate and fix this issue.
image

hnyls2002:
Thanks @mmdbhs. As discussed in the thread, the assert is intentional for catching unexpected runtime recompilation; main now handles the None-stream case explicitly (cuda_piecewise_backend.py adds an eager fallback for the HIP recompile scenario via #22299 while keeping the assertion on other platforms). Closing since the blanket removal was not the agreed direction; please reopen if I've missed something.

PR #16500, opened 2026-01-05
Add thresh None

lxaw:
## Motivation

Prior did not have a thresh None case where one token denoised per step.

## Modifications

Change the `sglang/srt/dllm/algorithm/fastdiffuser.py` slightly to include a case for `self.threshold=="None"` that denoises one token per step.

### DIFF
--- a/README-DLM.md
+++ b/README-DLM.md
@@ -1,0 +1,16 @@
+# build
+```
+git clone https://github.com/hutm/sglang sglang-diffusion && cd sglang-diffusion && git checkout dev-dllm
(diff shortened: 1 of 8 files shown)


hnyls2002:
Thanks @lxaw. This PR modifies sglang/srt/dllm/algorithm/fastdiffuser.py and models/fastdiffuser.py, which do not exist in this repository (the upstream dllm algorithms are base/joint_threshold/low_confidence), and it also includes unrelated artifacts like README-DLM.md and a nemotron-diffusion-src gitlink, so it cannot be applied to main. Closing; please reopen with a rebased, self-contained change if I've missed something.

PR #16522, opened 2026-01-05
Fix AMD GPU visibility by setting HIP_VISIBLE_DEVICES

alisonshao:
## Summary
Fix flaky AMD CI failures caused by HIP device ordinal errors.

The `maybe_reindex_device_id` function was only setting `CUDA_VISIBLE_DEVICES` but not `HIP_VISIBLE_DEVICES` for AMD GPUs. On ROCm, `HIP_VISIBLE_DEVICES` is the more reliable way to control GPU visibility.

## Changes
- Set `HIP_VISIBLE_DEVICES` in addition to `CUDA_VISIBLE_DEVICES` when running on AMD GPUs
- Properly restore `HIP_VISIBLE_DEVICES` after subprocess launch
- Remove unnecessary comment from test file

## Test plan
- CI will validate the fix on AMD runners

### DIFF
--- a/python/sglang/srt/utils/common.py
+++ b/python/sglang/srt/utils/common.py
@@ -3488,6 +3488,8 @@ def maybe_reindex_device_id(gpu_id: int):
         return
 
     original_cuda_visible_devices = os.environ.get("CUDA_VISIBLE_DEVICES")
+    original_hip_visible_devices = os.environ.get("HIP_VISIBLE_DEVICES")
+
     if original_cuda_visible_devices:
         cuda_visible_devices = original_cuda_visible_devices.split(",")
     else:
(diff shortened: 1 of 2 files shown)


HaiShaw (approved):

yctseng0211:
@alisonshao Thanks for your help! 
I believe this might have already been fixed in (https://github.com/sgl-project/sglang/pull/15663). Let me know if you see otherwise.

The root cause is the missing is_in_ci() related environment variable, which triggers 2-gpu tests:

ALL_OTHER_MODELS = [ModelCase(DEFAULT_MODEL_NAME_FOR_TEST, tp_size=2)]

causing errors due to insufficient GPU resources.

alisonshao:
> @alisonshao Thanks for your help! I believe the CI failures might have already been fixed in (#15663) merged 5 hrs ago. Let me know if you see otherwise.
> 
> The root cause is the missing is_in_ci() related environment variable, which triggers 2-gpu tests:
> 
> ALL_OTHER_MODELS = [ModelCase(DEFAULT_MODEL_NAME_FOR_TEST, tp_size=2)]
> 
> causing errors due to insufficient GPU resources.

thank you, i think that fixes this bug

PR #16534, opened 2026-01-06
[Refactor] Add `-fp4-gemm-backend` to replace `SGLANG_FLASHINFER_FP4_GEMM_BACKEND`

b8zhong:
## Motivation

After https://github.com/sgl-project/sglang/issues/13773, it seems useful to also add a `--fp4-gemm-backend`, to avoid env vars like `SGLANG_FLASHINFER_FP4_GEMM_BACKEND`. Currently, it's absolute usefulness to decrease the complexity because many backends are specified through Flashinfer `mm_fp4` backend argument, but it should become more useful in the future.

## Modifications

Refactor the codes (similar to FP8 one). Also, add the CuDNN backend to the list, which has been an available choice for a while.
## Checklist

### DIFF
--- a/python/sglang/srt/layers/quantization/fp4_utils.py
+++ b/python/sglang/srt/layers/quantization/fp4_utils.py
@@ -44,17 +44,15 @@ def initialize_fp4_gemm_config(server_args: ServerArgs) -> None:
 
     # Handle deprecated env var for backward compatibility
     # TODO: Remove this in a future version
-    if backend == "auto":
+    if envs.SGLANG_FLASHINFER_FP4_GEMM_BACKEND.is_set():
         env_backend = envs.SGLANG_FLASHINFER_FP4_GEMM_BACKEND.get()
-        if env_backend:
+        if backend == "auto":
             logger.warning(
                 "SGLANG_FLASHINFER_FP4_GEMM_BACKEND is deprecated. "
                 f"Please use '--fp4-gemm-backend={env_backend}' instead."
             )
             backend = env_backend
-    else:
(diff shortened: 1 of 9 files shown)


python/sglang/srt/layers/quantization/fp4_utils.py
```diff
@@ -0,0 +1,72 @@
...
+def initialize_fp4_gemm_config(server_args: ServerArgs) -> None:
+    """Initialize FP4 GEMM configuration from server args."""
+    global FP4_GEMM_RUNNER_BACKEND
+
+    backend = server_args.fp4_gemm_runner_backend
+
+    # Handle deprecated env var for backward compatibility
+    # TODO: Remove this in a future version
+    if backend == "auto":
+        env_backend = envs.SGLANG_FLASHINFER_FP4_GEMM_BACKEND.get()
```
Fridge003, line 48: Deduplicate line48-line53 here

Fridge003:
> @Fridge003 Do you want a UT or E2E test (since the logics of different FP4 backends are mostly in scale preperation or smth, I suppose it can just be a short UT)

We need an E2E test to make sure the kernels are integrated correctly.
Just similar to `test_llama_fp4.py` (or we can change this file to a fp4 gemm test)

python/sglang/srt/layers/quantization/fp4_utils.py
b8zhong: Consolidated 👍

Fridge003 (approved):

PR #15683, opened 2025-12-23
[sgl-model-gateway] Add EPD (Encode-Prefill-Decode) Disaggregated Routing Support

chenzongyao200127:
## Motivation

Following the recent kernel-level support for EPD disaggregation in SGLang (see [PR #12263](https://github.com/sgl-project/sglang/pull/12263)), this PR implements EPD (Encode-Prefill-Decode) disaggregated routing for sgl-router, enabling efficient multimodal LLM inference by separating the encode, prefill, and decode stages across specialized workers. This extends the existing PD (Prefill-Decode) mode to support vision-language models like Qwen2.5-VL.

Key Problem Addressed:
- Multimodal models need to process images/videos separately from text
- Traditional single-worker or PD-mode serving underutilizes resources for vision-language workloads
- PD mode lacks the encode stage required for efficient multimodal input processing

Architecture Pattern:

~~~
User Request (with images/video)
↓
Encode Worker (HTTP) → Processes images/video, sends embeddings
↓
Prefill Worker (gRPC) → Receives embeddings, computes KV cache
↓
Decode Worker (gRPC) → Generates output tokens using KV cache
↓
Response
~~~

## Modifications

**Core Infrastructure**

Triple Dispatch Pattern:
- Extended the pipeline framework to support three-stage dispatch (EPD) alongside existing single dispatch (Regular) and dual dispatch (PD)
- New WorkerSelectionMode::EncodePrefillDecode for selecting HTTP encode + gRPC prefill/decode workers
- New ExecutionMode::TripleDispatch for coordinating encode HTTP call with prefill/decode gRPC streams

HTTP Client Module:
- Created dedicated HTTP client for encode worker REST API (src/http_client/)
- EncodeHttpClient with configurable timeout and connection pooling
- EncodeRequest/EncodeResponse types for multimodal encoding requests

Worker Types:
- WorkerType::Encode: HTTP-based encode workers with optional bootstrap port
- WorkerType::Prefill: gRPC-based prefill workers with ZMQ embedding receiver
- WorkerType::Decode: gRPC-based decode workers

**Pipeline Stages**

WorkerSelectionStage:
- select_epd_triple() method selects encode (HTTP) + prefill (gRPC) + decode (gRPC) workers
- Separate routing policies per worker type (encode_policy, prefill_policy, decode_policy)
- Falls back gracefully if encode workers unavailable

RequestBuildingStage:
- Injects EPD metadata (request IDs, worker IDs, prefill host info)
[...]

PR #16552, opened 2026-01-06
[model gateway][0/N] router EPD support: add encoder grpc server backend support

JasonZhang517:
## Motivation

This PR is part of a staged series adding EPD (Encode‑Prefill‑Decode) support to SGLang Model Gateway and its test harness. The series is organized to keep review scope small and incremental:

PR1 (epd/pr1-python-grpc): Python-side gRPC encoder support.
PR2‑foundation (epd/pr2-foundation): Introduce encoder worker type foundation in Model Gateway.
PR2‑epd‑mode (epd/pr2-epd-mode): Add EPD mode configuration + service discovery classification.
PR3 (epd/pr3-routing): EPD routing pipeline (triple dispatch) and worker selection improvements.
PR4 (epd/pr4-python-bindings): Python bindings + E2E test harness for EPD

## Modifications

We are trying to split the giant PR into a few smaller PRs. 
https://github.com/sgl-project/sglang/pull/15683#issuecomment-3689062241
This is the python gRPC implementation part of the implementation.
The test results can be referred in the above branch.

## Checklist

## Review Process


python/sglang/srt/disaggregation/encode_grpc_server.py
```diff
@@ -0,0 +1,314 @@
...
+
+        except Exception as e:
+            logger.error(f"SchedulerReceiveUrl error: {e}")
+            traceback.print_exc()
+            context.set_code(grpc.StatusCode.INTERNAL)
+            context.set_details(str(e))
+            return sglang_encoder_pb2.SchedulerReceiveUrlResponse()
+
+
+async def run_encoder_worker(
```
ZhengWG: Can we reuse the method from `encoder_server.py` here? I see similar logic duplicated in other places too.

chenzongyao200127: Thank you for pointing that out. I’ve updated it to reuse the logic.

python/sglang/srt/disaggregation/encode_receiver.py
```diff
@@ -28,6 +30,81 @@
...
+                mm_items=encode_request["mm_items"],
+                req_id=encode_request["req_id"],
+                num_parts=encode_request["num_parts"],
+                part_idx=encode_request["part_idx"],
+                prefill_host=encode_request["prefill_host"],
+                embedding_port=_normalize_embedding_ports(
+                    encode_request["embedding_port"]
+                ),
+            ),
+            timeout=1800,
```
ZhengWG: The default timeout seems too long for encode requests. Consider 60s or an env variable (same as above)?

chenzongyao200127: Good Point! I’ve reduced the default gRPC encoder timeout to 60s and made it configurable via `SGLANG_ENCODER_GRPC_TIMEOUT_SECS` (applied to both Encode and Send). Thx for pointing that:)

ShangmingCai:
Please use `pre-commit run --all-files` to fix lint.

ShangmingCai (commented):
Can we change the `SglangEncoder` prefix to `SGLangEncoder` when the code is on the Python side?

python/sglang/srt/disaggregation/encode_grpc_server.py
```diff
@@ -0,0 +1,296 @@
...
+        return health_pb2.HealthCheckResponse(
+            status=health_pb2.HealthCheckResponse.NOT_SERVING
+        )
+
+    async def Watch(self, request, context):
+        """Streaming health check."""
+        yield await self.Check(request, context)
+
+
+class SglangEncoderServicer(sglang_encoder_pb2_grpc.SglangEncoderServicer):
```
ShangmingCai: Will this look better?
```suggestion
class SGLangEncoderServer(sglang_encoder_pb2_grpc.SglangEncoderServicer):
```

python/sglang/srt/disaggregation/encode_receiver.py
```diff
@@ -28,6 +31,84 @@
...
+                receive_count=receive_count,
+            ),
+            timeout=60,
+        )
+    finally:
+        channel.close()
+
+
+def _grpc_encode_request(target, encode_request):
+    timeout_secs = int(os.getenv("SGLANG_ENCODER_GRPC_TIMEOUT_SECS", "60"))
```
ShangmingCai: Maybe put this env var in here https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/environ.py#L395-L411

And use `from sglang.srt.environ import envs`

test/registered/distributed/test_epd_disaggregation.py
```diff
@@ -422,5 +462,334 @@ def test_mmmu(self):
...
+
+        cls.wait_grpc_ready(cls.base_host, cls.encode_port, cls.process_encode)
+        cls.wait_server_ready(cls.prefill_url + "/health")
+        cls.wait_server_ready(cls.decode_url + "/health")
+
+        cls.launch_lb()
+
+        cls.api_key = "sk-123456"
+        os.environ["OPENAI_API_KEY"] = cls.api_key
+        os.environ["OPENAI_API_BASE"] = f"{cls.lb_url}/v1"
```
ShangmingCai, line 466: need to pop these env vars in `tearDownClass`

test/registered/distributed/test_epd_disaggregation.py
```diff
@@ -422,5 +462,334 @@ def test_mmmu(self):
...
+            "--trust-remote-code",
+            "--encoder-only",
+            "--grpc-mode",
+            "--encoder-transfer-backend",
+            "zmq_to_scheduler",
+            "--tp",
+            "1",
+            "--enable-prefix-mm-cache",
+        ]
+        cls.process_encode = subprocess.Popen(encode_command)
```
ShangmingCai, line 684: IIUC, this will start the encoder in GPU 0, which might also be the device for the prefill instance. Will this cause OOM?

python/sglang/srt/entrypoints/grpc_server.py
```diff
@@ -533,6 +543,26 @@ async def GetLoads(
...
+            return
+
+        encode_req = GenerateReqInput(
+            image_data=[ImageData(url=url) for url in image_urls],
+            rid=grpc_req.request_id,
+        )
+        self.mm_receiver.send_encode_request(encode_req)
+        tokenized_req.need_wait_for_image = bool(encode_req.need_wait_for_image)
+        tokenized_req.num_items_assigned = encode_req.num_items_assigned
+
```
ShangmingCai, line 559: @gty111, please help review this endpoint

python/sglang/srt/disaggregation/encode_receiver.py
```diff
@@ -463,75 +552,125 @@ async def encode(
...
+                            f"{encoder_url}/{endpoint_send}",
+                            json=response_json,
+                        )
+                    )
+                offset += embedding_size_list_sort[response_json["part_idx"]]
+
+            if http_metadata_tasks:
+                await asyncio.gather(*http_metadata_tasks)
+        if grpc_metadata_tasks:
+            await asyncio.gather(*grpc_metadata_tasks)
```
ShangmingCai: Maybe a stupid question: why are these awaits implemented with different indent?

chenzongyao200127:
> Can we change the `SglangEncoder` prefix to `SGLangEncoder` when the code is on the Python side?

Sure thing! I'll get those fixed this week. Thanks for reviewing.

python/sglang/srt/entrypoints/grpc_server.py
```diff
@@ -533,6 +543,26 @@ async def GetLoads(
             aggregate=_compute_aggregate_protobuf(loads),
         )
 
+    def _handle_epd_disaggregation_encode_request(
```
gty111, line 541: Maybe we can wrap this function into mm_receiver

python/sglang/srt/disaggregation/encode_receiver.py
chenzongyao200127: Good Question. The http_metadata_tasks await is inside the async with aiohttp.ClientSession(...) block, while the grpc_metadata_tasks await is outside it. Because the HTTP tasks need the session to still be open, but the gRPC tasks are run via asyncio.to_thread and don’t depend on the session, so they can run after the session is closed.

test/registered/distributed/test_epd_disaggregation.py
chenzongyao200127: Good Point, In that class (TestEPDDisaggregationGrpcEncoder) it only starts the encoder, so there is no prefill process running at all. Using GPU 0 there won’t collide with anything. While in the gRPC MMMU test class prefill is pinned to GPU 1 and decode to GPU 2, so GPU 0 is free for the encoder.

I've already renamed this class to improve readability.

test/registered/distributed/test_epd_disaggregation.py
chenzongyao200127: Thx for pointing that, Fixed

python/sglang/srt/disaggregation/encode_grpc_server.py
chenzongyao200127: Fixed

python/sglang/srt/disaggregation/encode_receiver.py
chenzongyao200127: Thx for pointing that, Fixed

python/sglang/srt/entrypoints/grpc_server.py
chenzongyao200127: Yes we can, but it might be a trade-off, if we do that:
Pros:
 - keeps gRPC entrypoint slimmer
 - centralizes encode-request logic in one place 
Cons:
 - MMReceiver would need to know about gRPC types (GenerateRequest) and TokenizedGenerateReqInput, which are higher-level than a “receiver”
- it mixes transport/encoder concerns with request parsing and scheduler state

Maybe we can add a small helper on MMReceiver that takes primitive data and returns the encode_req, and keep the gRPC-specific parsing and tokenized_req updates in the entrypoint？

gty111: I think a helper make sense

python/sglang/srt/disaggregation/encode_receiver.py
```diff
@@ -639,3 +773,277 @@ async def _recv_mm_data(self, req_id, recv_socket, mm_processor, prompt):
...
+        )
+    return MMReceiverHTTP(
+        server_args,
+        dtype=dtype,
+        hf_config=hf_config,
+        pp_rank=pp_rank,
+        tp_rank=tp_rank,
+        tp_group=tp_group,
+        scheduler=scheduler,
+    )
```
liusy58, line 1054: Simply use a parameter or an environment variable to determine which receiver class to use.

python/sglang/srt/disaggregation/encode_receiver.py
```diff
@@ -258,6 +384,10 @@ def __init__(
         self.encode_urls = server_args.encoder_urls
         self.encode_idx = list(range(len(self.encode_urls)))
         self.host = server_args.host
+        if any(_is_grpc_url(url) for url in self.encode_urls):
+            raise ValueError(
+                "MMReceiverHTTP requires http(s):// encoder URLs; use MMReceiverGrpc for grpc://"
+            )
```
liusy58: I’m not sure if you’re intending to mix gRPC and HTTP. If so, wouldn't that introduce a lot of other issues?

python/sglang/srt/disaggregation/encode_receiver.py
chenzongyao200127: You're right, i should not determine the kind of MMReceiver by infering the url. Thx for pointing that.
Fixed — receiver selection is now explicit via SGLANG_ENCODER_MM_RECEIVER_MODE (http|grpc), and create_mm_receiver reads that by default. No URL inference or extra CLI.

python/sglang/srt/disaggregation/encode_receiver.py
chenzongyao200127: Fixed, I am not intend to mix gRPC and HTTP transpot mode, Thx for pointing that.

python/sglang/srt/disaggregation/encode_receiver.py
```diff
@@ -666,3 +795,280 @@ async def _recv_mm_data(self, req_id, recv_socket, mm_processor, prompt):
 
         mm_inputs = mm_processor.get_mm_data(prompt, recv_embedding, img_grid_thw)
         return mm_inputs
+
+
+class MMReceiverGrpc(MMReceiverHTTP):
```
liusy58: I don't think `MMReceiverGrpc` should extend `MMReceiverHTTP`. They should probably share a common base class instead.

chenzongyao200127: Thx for pointing that. I have dropped the inheritance and made MMReceiverGrpc a sibling of MMReceiverHTTP under MMReceiverBase.

python/sglang/srt/disaggregation/encode_receiver.py
```diff
@@ -579,109 +859,204 @@ async def encode(
...
             await asyncio.gather(*metadata_tasks)
 
-    # For mooncake
-    async def allocate_embedding_buffer(self, req_id, embedding_length, embedding_dim):
-        embeddings = torch.zeros(
-            (embedding_length, embedding_dim),
-            dtype=self.dtype,
+    # For zmq_to_tokenizer and mooncake
+    async def recv_mm_data(self, img_data, mm_processor, prompt):
+        return await super().recv_mm_data(img_data, mm_processor, prompt)
```
liusy58: Redundant override. This is a direct passthrough to the parent class, so it isn't necessary. Recommend deleting it to reduce boilerplate.

chenzongyao200127: Thx for pointing that, that' s a stupid mistake, i have fixed it:)

python/pyproject_cpu.toml
```diff
@@ -68,7 +68,7 @@ dependencies = [
   "uvicorn",
   "uvloop",
   "xgrammar==0.1.27",
-  "smg-grpc-proto>=0.3.3",
+  "smg-grpc-proto>=0.4.1",
```
ShangmingCai, line 71: Better discuss this with Simo first

chenzongyao200127: Got it, i have opened a PR in smg repo about this change - https://github.com/lightseekorg/smg/pull/536 . I will discuss this with Simo in Slack. Thx for pointing that.

python/sglang/srt/entrypoints/grpc_server.py
```diff
@@ -159,6 +159,15 @@ def __init__(
...
         self.start_time = time.time()
         self.health_servicer = health_servicer
+        self.mm_receiver = None
+        if (
+            self.server_args.language_only
+            and self.server_args.encoder_transfer_backend == "zmq_to_scheduler"
+        ):
+            from sglang.srt.disaggregation.encode_receiver import create_mm_receiver
+
+            os.environ.setdefault("SGLANG_ENCODER_MM_RECEIVER_MODE", "grpc")
```
ShangmingCai: Why are these lines needed, if SGLANG_ENCODER_MM_RECEIVER_MODE is required before the start command?

python/sglang/launch_server.py
```diff
@@ -14,6 +14,16 @@
 def run_server(server_args):
     """Run the server based on server_args.grpc_mode and server_args.encoder_only."""
     if server_args.grpc_mode:
+        os.environ.setdefault("SGLANG_ENCODER_MM_RECEIVER_MODE", "grpc")
+    else:
+        os.environ.setdefault("SGLANG_ENCODER_MM_RECEIVER_MODE", "http")
```
ShangmingCai: Looks a little bit hacky. Should not put them here, maybe put it in here: https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/server_args.py#L2608

python/sglang/srt/entrypoints/grpc_server.py
chenzongyao200127: That' s a good question. They’re needed because the prefill worker doesn’t infer the receiver type on its own. If the env isn’t set, it defaults to HTTP and creates the wrong MMReceiver for grpc:// URLs. Setting it here guarantees the correct receiver is used (and setdefault won’t override a value that was already set before launch).

ShangmingCai: The logic here is a bit convoluted. Do not set environment variables in the code; this is only allowed in CI. Please import (use import xx as xxx) and configure the correct MMReceiver based on the server args.

chenzongyao200127: Ack. I removed the os.environ.setdefault lines (both in launch_server.py and server_args.py). 
The gRPC server now imports encode_receiver as an alias and calls create_mm_receiver(...) directly, relying on SGLANG_ENCODER_MM_RECEIVER_MODE to be set externally (as in the docs).

python/sglang/launch_server.py
chenzongyao200127: Fixed

python/sglang/launch_server.py
ShangmingCai: Maybe refactor it to:

```suggestion
    if server_args.encoder_only:
        if server_args.grpc_mode:
            from sglang.srt.disaggregation.encode_grpc_server import serve_grpc_encoder

            asyncio.run(serve_grpc_encoder(server_args))
        else:
            from sglang.srt.disaggregation.encode_server import launch_server
            launch_server(server_args)
    elif server_args.grpc_mode:
        from sglang.srt.entrypoints.grpc_server import serve_grpc
        asyncio.run(serve_grpc(server_args))
```

ShangmingCai: Try make encoder_only a top-level branch, then check grpc inside that branch

ShangmingCai (commented):
No other comments. Please get approval from Simo, then we can merge.

python/sglang/launch_server.py
chenzongyao200127: Thx for pointing that, i have cleaned it.

chenzongyao200127:
> No other comments. Please get approval from Simo, then we can merge.

Sure, I'll ping you once the encoder proto-related changes are merged (https://github.com/lightseekorg/smg/pull/536). Thanks again for your review!

liusy58:
Need to pass CI first.

python/sglang/launch_server.py
ShangmingCai: Seems like the update hasn't been pushed yet from the diff

docs/advanced_features/epd_disaggregation.md
```diff
@@ -78,3 +78,42 @@ python -m sglang_router.launch_router \
...
+```bash
+# gRPC encoder
+python -m sglang.launch_server \
+  --model-path Qwen/Qwen3-VL-8B-Instruct \
+  --encoder-only \
+  --grpc-mode \
+  --encoder-transfer-backend zmq_to_scheduler \
+  --port 30000
+
+# prefill (HTTP) - tell it to use gRPC receiver
```
liusy58, line 97: Curious, why use HTTP here?

ShangmingCai:
Why not use elif and else? Use early `return` reduce the code readability.

chenzongyao200127:
Hello @ShangmingCai! Now that [the dependency PR: add sglang encoder proto to smg-grpc-proto](https://github.com/lightseekorg/smg/pull/536#event-23123053693) is merged, do you think we’re good to merge this one too? Looking forward to your thoughts. Thank you

docs/advanced_features/epd_disaggregation.md
chenzongyao200127: That' s a good question. 

A full gRPC-for-all setup is possible in principle, but it requires additional router-side gRPC routing (i am working on it: https://github.com/lightseekorg/smg/pull/585) That is
  not fully implemented in current codebase yet.

Maybe we can update to all grpc-mode after that PR merge?

ShangmingCai (approved):
Sure. We can merge this now since the dependency issue has been resolved and the CI has passed.

PR #16579, opened 2026-01-06
Add CI metrics collection infrastructure for stage-b jobs

harvenstar:
## Background

  This PR implements step 2 of the roadmap outlined in #15996: Set up CI infrastructure to collect and upload test metrics artifacts.

  Builds on top of #16064 which added the `dump_metric()` function and integrated it into `run_eval.py`.

  ## Changes

  1. **New script: `scripts/ci/merge_metrics.py`**
     - Merges per-process JSONL files (`test_metrics.*.jsonl`) into single output
     - Pure stdlib implementation (no sglang dependency)
     - JSON validation with skip-on-error for robustness
     - UTF-8 encoding to avoid runner differences
     - Creates parent directories automatically

  2. **CI workflows: `pr-test.yml` and `pr-test-amd.yml`**
     - Set `SGLANG_TEST_METRICS_OUTPUT` in all stage-b jobs:
       - `stage-b-test-small-1-gpu` (8 partitions)
       - `stage-b-test-large-1-gpu`
       - `stage-b-test-large-2-gpu`
       - `stage-b-test-4-gpu-b200`
       - AMD equivalents (stage-b-test-small-1-gpu-amd, stage-b-test-large-2-gpu-amd)
     - Merge metrics with `if: always()` (runs even on test failure)
     - Upload entire `$METRICS_DIR/` as artifact with `if: always()`

  3. **Artifact contents**
     - `test_metrics.jsonl` - merged metrics for trend analysis
     - `test_metrics.*.jsonl` - per-worker raw files for debugging

  ## Reason

  - **Flaky test diagnosis**: Collect metrics even when tests fail
  - **Root cause analysis**: Raw per-worker files help identify which process had issues
  - **Regression detection**: Merged file enables historical trend analysis
  - **Future work**: Foundation for automated regression detection system

  ## Verification

  - CI jobs will produce `regression-metrics-{run_id}-{job}-[partition-{N}]` artifacts
  - Each artifact contains both merged and raw JSONL files
  - Works on both NVIDIA and AMD runners (handles container mount correctly)
  - Artifact names are unique for parallel workers

  ## Next Steps

  After this PR:
  - Step 3: Migrate `test_standalone_speculative_decoding.py` to use `dump_metric()`
  - Step 4: Migrate `test_hicache_variants.py` to use `dump_metric()`

  Related: #15996
  Depends on: #16064


.github/workflows/pr-test-amd.yml
```diff
@@ -282,6 +302,8 @@ jobs:
       matrix:
         runner: [linux-mi325-gpu-2]
     runs-on: ${{matrix.runner}}
+    env:
```
merrymercy: Setting this for every job is too redundant. Just use a global default value for all cases.

.github/workflows/pr-test.yml
```diff
@@ -993,6 +1057,19 @@ jobs:
...
+        if: always()
+        run: |
+          python3 scripts/ci/merge_metrics.py "$SGLANG_TEST_METRICS_OUTPUT" "$METRICS_DIR/test_metrics.jsonl"
+
+      - name: Upload metrics artifact
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: regression-metrics-${{ github.run_id }}-${{ github.job }}
+          path: ${{ env.METRICS_DIR }}/
```
merrymercy: Adding three steps for each job is too redundant! Please do the following

1. Unify your code into a github action like `actions/checkout@v4`, so we can update the code in a single place without changing everywhere.
2. You can register post processing in the action, so the last two steps "Merge metrics" and "Upload metrics artifact" do not need to be there. Please refer to the "Post checkout code". It will run but users do not need to add one more step in the workflow file. It is coming form the `actions/checkout@v4`. Also, please merge the last two steps.
Image

merrymercy (commented):
Can you show us an example run with example output file?

PR #16607, opened 2026-01-07
[diffusion] feat: add multi-LoRA batching support

gmixiaojin:
## Motivation
Enable processing multiple requests with different LoRA adapters in a single batch for diffusion models, improving throughput for multi-tenant inference scenarios.

## Modifications
- Added multi-LoRA batching support to `BaseLayerWithLoRA` class with per-sample LoRA tracking
- Implemented `_apply_multi_lora()` in `RowParallelLinearWithLoRA` for efficient batched LoRA delta computation
- Added tensor parallelism support for multi-LoRA (partial sums aggregated via all-reduce)
- Maintains backward compatibility with existing single-LoRA inference path


niehen6174:
I feel like something is missing. The entry points and settings for using multiple LoRA models don't seem to have corresponding implementations.
I think the `set_lora` entry point for using multiple LoRA models should have some adaptation here. `set_multi_lora_state` and `clear_multi_lora_state` are also not being used anywhere.

PR #16645, opened 2026-01-07
[BugFix] fix TTFT measure error when special token like video_pad random…

ZLkanyo009:
# Modifications
When testing multimodal models (such as qwen3vl) using bench_serving.py, since bench_serving.py randomly generates various tokens—including special tokens like <image_pad> and <video_pad>—sglang fails to correctly handle requests containing such special tokens. Some cases result in direct errors, as seen in https://github.com/sgl-project/sglang/pull/13254, while others return empty requests under the retry mechanism of TokenizerManager. This PR addresses these issues by ensuring that no special tokens are randomly generated anymore.

# Error manifestation
server:
```
model=/models/nvme0/Qwen3-VL-235B-A22B-Instruct-FP8/
   --model-path ${model} \
   --host 0.0.0.0 \
   --port 9000 \
   --tp 4
```

client:
```
  --backend sglang \
  --port 9000 \
  --model "/models/nvme0/Qwen3-VL-235B-A22B-Instruct-FP8/" \
  --dataset-name image \
  --image-count 1 \
  --max-concurrency 64 \
  --image-resolution 800x800 \
  --random-input-len 1000 \
  --random-output-len 1000 \
  --num-prompts 640 
```
Regardless of whether on NVIDIA GPUs or AMD GPUs
Directly returning empty text causes TTFT timing to be inaccurate.
image
image

The screenshot above shows the output I printed after making some modifications in the code. The first TTFT (Time To First Token) is normal, but the subsequent two attempts are problematic. This is because bench_serving, when randomly generating text, created special tokens like <image_pad> and <video_pad>. This caused the model to encounter an error while processing the request, leading it to return an empty text reply prematurely, resulting in a TTFT of 0. Consequently, this skewed the average TTFT measurement.

### DIFF
--- a/python/sglang/benchmark/datasets/common.py
+++ b/python/sglang/benchmark/datasets/common.py
@@ -74,10 +74,14 @@ def gen_prompt(tokenizer, token_num):
     return tokenizer.decode(selected_tokens)
 
 
-def gen_mm_prompt(tokenizer, image_pad_id, token_num):
+def gen_mm_prompt(tokenizer, image_pad_id, token_num, special_tokens=None):
     """Generate a random prompt of specified token length using tokenizer vocabulary."""
     all_available_tokens = list(tokenizer.get_vocab().values())
     if image_pad_id:
         all_available_tokens.remove(image_pad_id)
+    if special_tokens:
+        all_available_tokens = [
+            t for t in all_available_tokens if t not in special_tokens
+        ]
     selected_tokens = random.choices(all_available_tokens, k=token_num)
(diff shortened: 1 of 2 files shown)


hnyls2002:
Thanks @ZLkanyo009. This was fixed on main by #26864, which makes the synthetic multimodal benchmark prompt generation exclude all tokenizer special tokens (benchmark/datasets/common.py, get_available_multimodal_text_tokens). Closing as superseded; please reopen if I've missed something.

PR #16656, opened 2026-01-07
Fix qwen3_80_next radix error pd

whybeyoung:
FIx https://github.com/sgl-project/sglang/pull/14792


ShangmingCai (commented):
Great, I think this can address the register problem, but will it bring overhead since it requires a round of H2D copy?

ShangmingCai (approved):
LGTM!

hanming-lu (approved):

PR #16122, opened 2025-12-30
[Bugfix][DeepSeek-V3.2] Fix uniform condition judegement

Muqi1029:
## Motivation

Follow up #15155.  Fix #14695 #15042.

The internal bug stems from ｜DSML｜ being treated as a special token.

```shell
 grep -C 10 "｜DSML｜" tokenizer.json
```
image

Let me illustrate its influence with an example:
```py
ids =  [69440, 4482, 15429, 11514, 548, 14072, 3303, 303, 25414, 2386, 15831, 4915, 4482, 5109, 531, 91125, 2920, 3303, 320, 4482, 41897, 10539, 5189, 19, 16, 223, 4036, 22250, 8620, 637, 8916, 18780, 3303, 6692, 20, 16, 223, 3070, 5109, 531, 2541, 4, 2985, 2901, 4, 73984, 548, 397, 19598, 14072, 637, 22250, 21826, 35478, 8620, 4793, 4497, 1255, 2971, 653, 8093, 8620, 10821, 303, 11513, 4, 13933, 7907, 4, 4793, 2386, 3437, 16590, 28242, 1237, 2985, 2901, 1316, 4239, 4481, 1227, 548, 14072, 1237, 18, 15, 1349, 4850, 14164, 4482, 3796, 19682, 47652, 8767, 1959, 3099, 303, 34071, 4036, 22250, 1959, 8620, 637, 8916, 18780, 3303, 478, 10539, 48289, 22250, 8620, 5189, 30, 128793, 8701, 4941, 12548, 1018, 30, 128793, 40148, 5406, 2329, 1281, 13933, 7907, 3816, 30, 128793, 41523, 2329, 1281, 2852, 4, 3418, 1281, 11476, 3320, 2985, 2901, 1718, 128793, 41523, 1018, 30, 128793, 41523, 2329, 1281, 615, 4, 3418, 1281, 19836, 3320, 397, 1718, 128793, 41523, 1018, 1718, 128793, 40148, 5406, 1018, 1718, 128793, 8701, 4941, 12548, 32, 1]
>>> tokenizer.decode(ids)
[...]

Issue #15042, opened 2025-12-13
[Bug] deepseek v3.2 Under high concurrency, the function calling feature is experiencing issues.


huhu-tiger:
### Checklist

- [ ] I searched related issues but found no solution.
- [ ] The bug persists in the latest version.
- [ ] Issues without environment info and a minimal reproducible demo are hard to resolve and may receive no feedback.
- [ ] If this is not a bug report but a general question, please start a discussion at https://github.com/sgl-project/sglang/discussions. Otherwise, it will be closed.
- [ ] Please use English. Otherwise, it will be closed.

### Describe the bug

In a scenario with multiple concurrent requests, the function calling feature is experiencing issues.

### Reproduction

# request funcation calling params
```
{
  "model": "DeepSeek-V3.2",
  "messages": [
    {
      "role": "user",
      "content": "Check the weather in Beijing and Shanghai."
    }
  ],
  "priority":100,
  "tools": [
    {
      "type": "function",
      "function": {
        "name": "query_weather",
        "description": "Query weather information for a specified city.",
        "parameters": {
          "type": "object",
          "properties": {
            "city": {
              "type": "string",
              "description": "City names, for example: Beijing, Shanghai, Qingdao"
            }
          },
          "required": ["city"]
        }
      }
    }
  ],
  "temperature": 0,
  "max_tokens": 2000
}
```
# Under low concurrent session requests, the function calling request returns...
```
{
    "id": "27b2900f32a24ac9a034e39719064f97",
    "object": "chat.completion",
    "created": 1765608121,
    "model": "DeepSeek-V3.2",
    "choices": [
        {
            "index": 0,
            "message": {
                "role": "assistant",
                "content": "I'll check the weather for both Beijing and Shanghai for you.",
                "reasoning_content": null,
                "tool_calls": [
                    {
                        "id": "call_7ba9516b17e1425992e78c21",
                        "index": -1,
                        "type": "function",
                        "function": {
                            "name": "query_weather",
                            "arguments": "{\"city\": \"Beijing\"}"
                        }
                    },
                    {
                        "id": "call_d363341ad6594bad9ee4683b",
                        "index": -1,
                        "type": "function",
                        "function": {
                            "name": "query_weather",
                            "arguments": "{\"city\": \"Shanghai\"}"
                        }
                    }
                ]
            },
            "logprobs": null,
            "finish_reason": "tool_calls",
            "matched_stop": null
        }
    ],
    "usage": {
[...]

PR #16698, opened 2026-01-08
fix(function_call): group batch decode by options instead of fallback

ruokee:
## Motivation

Follow up #15155 #16122. Fix #14695 #15042.

DeepSeek v3.2 function calls require `skip_special_tokens=False`. Function call control markers (e.g. `"| DSML |"`) are encoded as special tokens and must not be stripped during decoding.

> At the very beginning, the bug arose because the detokenizer manager defaulted to using the `skip_special_tokens` value from the first request when calling `batch_decode`.
>
> But why does SGLang sometimes work as expected? This behavior is actually limited to the first request in the batch, which is random. For tool calls, the request's `skip_special_tokens` is explicitly set to `False` (it is `True` by default).
>
> Thanks to community feedback and PR #15155, the above abnormal behavior seems to be mitigated.
>
> However, the incorrect judgment still introduces bugs. `batch_decode` can only be safely used when the `skip_special_tokens` values of all requests in a batch are uniform.

This PR implements grouped batch decoding: requests are grouped by flags, `batch_decode` is applied to each group separately, and results are reassembled in original order. This preserves batch decoding efficiency while correctly handling heterogeneous decode options.

## Modifications

- Added `_grouped_batch_decode` helper method to `DetokenizerManager` that groups requests by `(skip_special_tokens, spaces_between_special_tokens)` and applies `batch_decode` to each group
- Modified `_decode_batch_token_id_output` to use the new grouped batch decode instead of assuming uniform options across all requests

## Checklist

## Review Process


python/sglang/srt/managers/detokenizer_manager.py
```diff
@@ -174,6 +174,35 @@ def handle_batch_embedding_out(self, recv_obj: BatchEmbeddingOutput):
...
+    def _grouped_batch_decode(
+        self,
+        ids_list: list[list[int]],
+        skip_list: list[bool],
+        space_list: list[bool],
+    ) -> list[str]:
+        """Batch decode with grouping by (skip_special_tokens, spaces_between_special_tokens)."""
+
+        assert self.tokenizer is not None
+
```
JustinTong0323, line 186: We could add a fast path to avoid overhead in most cases like:
```suggestion
# fast path
first_skip, first_space = skip_list[0], space_list[0]
    if all(s == first_skip for s in skip_list) and all(sp == first_space for sp in space_list):
        return self.tokenizer.batch_decode(
            ids_list, 
            skip_special_tokens=first_skip, 
            spaces_between_special_tokens=first_space
        )
```

python/sglang/srt/managers/detokenizer_manager.py
```diff
@@ -174,6 +174,35 @@ def handle_batch_embedding_out(self, recv_obj: BatchEmbeddingOutput):
...
+        results: list[str] = [""] * len(ids_list)
+        for (skip, space), items in groups.items():
+            indices, grouped_ids = zip(*items)
+            decoded = self.tokenizer.batch_decode(
+                list(grouped_ids),
+                skip_special_tokens=skip,
+                spaces_between_special_tokens=space,
+            )
+            for idx, text in zip(indices, decoded):
+                results[idx] = text
```
JustinTong0323, line 213: Avoid `zip(*items)` here to save memory. For large `ids_list`, unzipping creates large intermediate objects. Fetching by index in the loop is more efficient.

```suggestion
        # Only store indices in 'groups' to keep the dict small
        # Then fetch IDs only when needed for the batch call
        groups: Dict[Tuple[bool, bool], List[int]] = defaultdict(list)
        for idx, (skip, space) in enumerate(zip(skip_list, space_list)):
            groups[(skip, space)].append(idx)
    
        results: List[str] = [None] * n # type: ignore
        
        for (skip, space), indices in groups.items():
            indices = groups[(skip, space)]
            grouped_ids = [ids_list[i] for i in indices]
            
            decoded = self.tokenizer.batch_decode(
                grouped_ids,
                skip_special_tokens=skip,
                spaces_between_special_tokens=space,
            )
            
            for idx, text in zip(indices, decoded):
                results[idx] = text
```

JustinTong0323 (commented):
Thanks for your contribution! Leave some suggestions:

JustinTong0323 (approved):

PR #16711, opened 2026-01-08
Add `--stream-response-default-include-usage` server flag

syd520zy:
## Motivation

When streaming is enabled, usage info is only returned if the client sets `stream_options.include_usage = true`. Server operators who need token-level monitoring metrics cannot rely on clients to set this. This PR adds a server-side flag to force usage inclusion in streaming responses.

## Modifications

- Add `--stream-response-default-include-usage` server arg that forces a final usage chunk in streaming responses even when `stream_options` is not specified by the client
- Extract repeated `stream_options` checks into a shared `should_include_usage()` utility in `utils.py`
- Remove dead `enable_force_include_usage` param from `OpenAIServingResponses` and unused `stream_output` field from `ServerArgs`
- Fix test mock to include the new server arg attribute

## Checklist


syd520zy:
@CatherineSue @JustinTong0323 @ispobock @merrymercy @slin1237 Please review my pr, thank you very much

hnyls2002 (changes requested):
Why is **force**?

syd520zy:
Please review my pr, thank you very much @CatherineSue @merrymercy @slin1237 @ispobock @JustinTong0323

syd520zy:
> Why is **force**?

The "include_usage" configuration can currently be specified through the request body, but when the user does not explicitly enable this configuration, its default value is False, which can cause the server to be unable to obtain the actual usage information for the current request.
In practical business scenarios, the server expects to capture usage related information for all requests.
so i called it force.

hnyls2002 (approved):

PR #16744, opened 2026-01-08
support new qwen3_coder_detector

attack204:
# UT

39a93fe0305fa84ec404a08211d84853

# E2E Test

76e2386d0e007fb9c295f7f8b61f2904

The test results, provided by Zeyu Cui from the Qwen Team @cyente, have been confirmed to meet expectations despite the presence of some jitter.


python/sglang/srt/function_call/qwen3_coder_detector.py
```diff
@@ -17,334 +15,458 @@
...
+                        f"Parsed value '{param_value}' of parameter '{param_name}' cannot be parsed with json.loads in tool "
+                        f"'{func_name}', will try other methods to parse it."
+                    )
+            try:
+                param_value = ast.literal_eval(param_value)  # safer
+            except:
+                logger.warning(
+                    f"Parsed value '{param_value}' of parameter '{param_name}' cannot be converted via Python `ast.literal_eval()` in tool '{func_name}', degenerating to string."
+                )
+            return param_value
```
zhaochenyang20, line 170: Gemini is right, those `try-except` catches all the exceptions. This makes debugging impossible and can cause the server to hang or behave unpredictably instead of failing fast or handling specific errors.

python/sglang/srt/function_call/qwen3_coder_detector.py
```diff
@@ -17,334 +15,458 @@
...
+    def _reset_streaming_state(self):
+        """Reset internal streaming cursors."""
+        self.parsed_pos = 0
+        self.current_tool_param_count = 0
+        self.json_started = False
+        self.is_inside_tool_call = False  # [FIX] Reset state
+
+        # Base class state reset is handled by base class logic mostly,
+        # but we ensure our cursor aligns with buffer resets.
+        if hasattr(self, "_buffer") and not self._buffer:
```
zhaochenyang20: If a stream is interrupted mid-generation (user disconnects), `self.current_func_name` might retain a stale value. When the detector is reused, it could corrupt the next request.

Please Ensure self.current_func_name = None is explicitly set in _reset_streaming_state

python/sglang/srt/function_call/qwen3_coder_detector.py
```diff
@@ -17,334 +15,458 @@
...
+                break
 
-                normal += self._buf[:s]
-                self._buf = self._buf[s:]
+            # -------------------------------------------------------
+            # 1. Priority detection: check if it's the start of Tool Call
+            # -------------------------------------------------------
+            if current_slice.startswith(self.tool_call_start_token):
+                self.parsed_pos += len(self.tool_call_start_token)
+                self.is_inside_tool_call = True  # [FIX] Enter tool call region
```
zhaochenyang20: clean the comments

python/sglang/srt/function_call/qwen3_coder_detector.py
```diff
@@ -17,334 +15,458 @@
...
+        # Override parent class _buffer management, or work with parent class.
+        # SGLang BaseFormatDetector usually has its own _buffer, but we explicitly manage it here to ensure logic clarity
+        if not hasattr(self, "_buffer"):
+            self._buffer = ""
+
+        # Index pointing to the next character to be processed in buffer
+        self.parsed_pos = 0
+        # Parameter count inside the current tool being processed, used to determine whether to add comma
+        self.current_tool_param_count = 0
+        # Flag indicating whether current tool has already sent '{'
```
zhaochenyang20: In our style guide, try to avoid `hasattr` as much as possible. Like these two examples:

- **Avoid Dynamic Attributes**: Minimize the use of `getattr` or `setattr`. Code should be explicit for better traceability. [Example](https://github.com/THUDM/slime/pull/1141#discussion_r2654651150) [Example 2](https://github.com/THUDM/slime/pull/1141#discussion_r2658247124)

zhaochenyang20:
```python
class Qwen3CoderDetector(BaseFormatDetector):
    def __init__(self):
        super().__init__()
        
        # Explicitly define all attributes with Type Hints
        # No 'hasattr' checks needed. We control the lifecycle here.
        self._buffer: str = "" 
        self.parsed_pos: int = 0
        self.current_tool_param_count: int = 0
        self.json_started: bool = False
        self.is_inside_tool_call: bool = False
        
        # Initialize attributes that were missing in the original PR
        self.current_func_name: Optional[str] = None
```

python/sglang/srt/function_call/qwen3_coder_detector.py
```diff
@@ -17,334 +15,458 @@
...
+                if isinstance(params, dict) and "properties" in params:
+                    return params["properties"]
+                elif isinstance(params, dict):
+                    return params
+                else:
+                    return {}
+        logger.warning(f"Tool '{func_name}' is not defined in the tools list.")
+        return {}
+
+    def _convert_param_value(
```
zhaochenyang20, line 89: This function looks chaotic. Try to extract helper functions out like `_safe_parse_int, _safe_parse_bool`:

```python
def _convert_param_value(
        self, param_value: str, param_name: str, param_config: dict, func_name: str
    ) -> Any:
        # 1. Early return for trivial cases
        if param_value.lower() == "null":
            return None
            
        # 2. Schema validation
        if param_name not in param_config:
            if param_config:
                logger.warning(
                    f"Parameter '{param_name}' not defined in tool '{func_name}', returning raw string."
                )
            return param_value

        # 3. Determine type
        config = param_config[param_name]
        param_type = str(config.get("type", "string")).strip().lower() if isinstance(config, dict) else "string"

        # 4. Dispatch to specific parsers
        # logic extracted to avoid massive if-else block
        if param_type in ("string", "str", "text", "enum"):
            return param_value
            
        if param_type in ("integer", "int", "long"):
            return self._safe_parse_int(param_value, param_name, func_name)
            
        if param_type in ("float", "number"):
            return self._safe_parse_float(param_value, param_name, func_name)
            
        if param_type in ("boolean", "bool"):
            return self._safe_parse_bool(param_value, param_name, func_name)
            
        if param_type in ("object", "array", "list", "dict"):
            return self._safe_parse_json(param_value, param_name, func_name)

        # Fallback
        return param_value

    # --- Private Helpers (Strictly Typed & Safe) ---

    def _safe_parse_int(self, value: str, name: str, func: str) -> Union[int, str]:
        try:
            return int(value)
        except ValueError:
            logger.warning(f"Param '{name}' in '{func}' expected int, got '{value}'.")
            return value

    def _safe_parse_bool(self, value: str, name: str, func: str) -> Union[bool, str]:
        normalized = value.lower()
        if normalized == "true":
            return True
        if normalized == "false":
            return False
        logger.warning(f"Param '{name}' in '{func}' expected bool, got '{value}'.")
        return value

    def _safe_parse_json(self, value: str, name: str, func: str) -> Any:
        try:
            return json.loads(value)
        except json.JSONDecodeError:
            try:
                return ast.literal_eval(value)
            except (ValueError, SyntaxError):
                logger.warning(f"Param '{name}' in '{func}' failed JSON/AST parse.")
                return value
    # ... handle float similarly ...
```

zhaochenyang20 (changes requested):

zhaochenyang20 (commented):
You are great

PR #16770, opened 2026-01-09
fix: handle empty NUMA node in lscpu output on WSL

kuafou:
This is because WSL doesn't have real NUMA architecture, so the Node information is not available.

## Modifications

- Handle empty Node column by defaulting to node 0
- Skip empty lines during parsing  
- Add more robust validation of parsed fields (check `len(parts) >= 3` before accessing)

## Accuracy Tests

N/A - This is a bug fix for lscpu output parsing, does not affect model outputs.

## Benchmarking and Profiling

N/A - This change only affects CPU topology parsing during initialization, no impact on inference speed.

## Checklist

## Testing

Tested on WSL2 with AMD Ryzen 7 8845H (16 logical CPUs, no NUMA support). The fix correctly parses the lscpu output and defaults Node to 0.

### DIFF
--- a/python/sglang/srt/utils/common.py
+++ b/python/sglang/srt/utils/common.py
@@ -3102,12 +3102,19 @@ def parse_lscpu_topology():
     except Exception as e:
         raise RuntimeError(f"Unexpected error running 'lscpu': {e}")
 
-    # Parse only data lines (skip comments)
+    # Parse only data lines (skip comments and empty lines)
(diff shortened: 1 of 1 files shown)


hnyls2002:
Hi @kuafou, thanks for the fix. The empty NUMA node / malformed lscpu line handling has since landed on main via #18520 - parse_lscpu_topology now skips malformed lines and defaults empty Core/Socket/Node fields to 0, which covers the WSL case this PR addressed. Closing as fixed elsewhere; please reopen if I've missed something.

PR #16785, opened 2026-01-09
[Bugfix] fix recompile in qwen3 vl

narutolhy:
## Motivation

Qwen3-VL automatically injects input_deepstack_embeds into the language model inputs when requests contain mm_inputs (multimodal inputs). The current implementation does not handle this injected tensor consistently, so the model execution path differs between multimodal and text-only requests. In mixed traffic, this leads to unnecessary TorchDynamo recompilations at runtime (e.g., repeatedly specializing on the presence/absence of deepstack inputs), increasing latency and reducing throughput stability.

This PR makes the deepstack injection path explicit and stable to avoid redundant recompilation for Qwen3-VL multimodal inference.

## Modifications

Standardized the Qwen3-VL deepstack injection behavior so that the language model forward path remains stable when mm_inputs is present.

Ensured input_deepstack_embeds is handled in a consistent way across requests to prevent runtime recompilation churn (especially under mixed multimodal/text workloads).

Refactored the relevant code paths to make the deepstack injection explicit and easier to reason about.

## Accuracy Tests

The update in test/manual/nightly/test_vlms_vit_cuda_graph.py case covers the accuracy.

## Checklist

- [X] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.io/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.io/developer_guide/contribution_guide.html#benchmark-the-speed).

## Review Process

### DIFF
--- a/python/sglang/srt/model_executor/piecewise_cuda_graph_runner.py
+++ b/python/sglang/srt/model_executor/piecewise_cuda_graph_runner.py
@@ -245,6 +245,18 @@ class PiecewiseCudaGraphRunner:
                 mrope_positions = torch.zeros(
                     (3, self.max_num_tokens), dtype=torch.int64
                 )
+                if hasattr(self.model_runner.model, "num_deepstack_embeddings"):
+                    deepstack_slots = getattr(
+                        self.model_runner.model, "num_deepstack_embeddings", 3
+                    )
+                    self.input_deepstack_embeds = torch.zeros(
+                        (
+                            self.max_num_tokens,
+                            self.model_runner.model_config.hidden_size
+                            * deepstack_slots,
+                        ),
+                        dtype=self.model_runner.dtype,
+                    )
             else:
                 input_embeds = None
                 mrope_positions = None
@@ -392,6 +404,9 @@ class PiecewiseCudaGraphRunner:
         forward_batch.dp_local_start_pos = forward_batch.dp_local_num_tokens = None
         set_dp_buffer_len(None, num_tokens, forward_batch.dp_padding_mode.is_max_len())
         set_is_extend_in_batch(False)
+        kwargs = {}
+        if self.is_multimodal and hasattr(self, "input_deepstack_embeds"):
+            kwargs["input_deepstack_embeds"] = self.input_deepstack_embeds[:num_tokens]
         with set_forward_context(
             forward_batch,
             self.attention_layers,
@@ -403,6 +418,7 @@ class PiecewiseCudaGraphRunner:
                 forward_batch.input_ids,
                 forward_batch.positions,
                 forward_batch,
+                **kwargs,
             )
 
     def _cache_loc_dtype(self):
@@ -547,6 +563,10 @@ class PiecewiseCudaGraphRunner:
 
         self.model_runner.attn_backend.init_forward_metadata(forward_batch)
 
+        kwargs = {}
+        if self.is_multimodal and hasattr(self, "input_deepstack_embeds"):
+            kwargs["input_deepstack_embeds"] = self.input_deepstack_embeds[:num_tokens]
+
         # Run and capture
         def run_once():
             # Clean intermediate result cache for DP attention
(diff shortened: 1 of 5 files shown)


python/sglang/srt/model_executor/piecewise_cuda_graph_runner.py
```diff
@@ -357,11 +369,18 @@ def warmup_torch_compile(self, num_tokens: int):
...
-                forward_batch.input_ids,
-                forward_batch.positions,
-                forward_batch,
-            )
+            kwargs = {}
+            if self.is_multimodal and hasattr(self, "input_deepstack_embeds"):
+                kwargs["input_deepstack_embeds"] = self.input_deepstack_embeds[
+                    :num_tokens
+                ]
+            with torch.no_grad():
```
Oasis-Git: Which part of vlm will introduce non no_grad() mode?

python/sglang/srt/model_executor/piecewise_cuda_graph_runner.py
```diff
@@ -683,12 +707,19 @@ def replay(
                 self.moe_layers,
             ):
                 with set_compiled(True):
-                    output = self.model_runner.model.forward(
-                        static_forward_batch.input_ids,
-                        static_forward_batch.positions,
-                        static_forward_batch,
-                        **kwargs,
-                    )
+                    call_kwargs = {}
```
Oasis-Git: Can we move all the kwargs preparation outside the piecewise cuda graph context?

python/sglang/srt/models/qwen3_vl.py
```diff
@@ -566,19 +566,41 @@ def __init__(
...
+                )
+                self.deepstack_embeds_buffer = torch.zeros(
+                    new_len, total, dtype=dtype, device=device
+                ).contiguous()
+            sep = self.hidden_size * layer_idx
+            return self.deepstack_embeds_buffer[:seq_len, sep : sep + self.hidden_size]
         sep = self.hidden_size * layer_idx
         return input_deepstack_embeds[:, sep : sep + self.hidden_size]
 
+    @torch.no_grad()
```
Oasis-Git: With this no_grad() maybe we can remove the one in pcg file??

Oasis-Git (commented):
Previously when I run this model I noticed the recompilation is caused by grad mode. So could you please share which part introduce forward with gradient?

yuan-luo:
> > Could you help to verify with SGLANG_VIT_ENABLE_CUDA_GRAPH=1? Qwen3-VL ViT CUDA Graph considered injecting deepstack output into LLM layers, just want make sure the full ViT+LLM CUDA Graph path working correctly.
> > The issue also exists when SGLANG_VIT_ENABLE_CUDA_GRAPH is set to 0.
> 
> Hi @yuan-luo Thank you for your review. I tested with VIT CUDA Graph before. It has some error. This is the issue #16883 I think this is due to the latest optimizations, but after I removed them, it resulted in a CUDA address access error during graph replay. I think it's because different graphs are being used for the same `seq_len` length, and your planned refactoring should solve this problem. #16902 I'll try test it with your new pr. Thank you

@narutolhy Did https://github.com/sgl-project/sglang/pull/16902 fix your problem?

python/sglang/srt/models/qwen3_vl.py
narutolhy: Yes, I removed it.

python/sglang/srt/model_executor/piecewise_cuda_graph_runner.py
narutolhy: OK, I removed them outside the piecewise cuda graph context

python/sglang/srt/model_executor/piecewise_cuda_graph_runner.py
narutolhy: Removed

Oasis-Git (approved):
LGTM from the pcg side. @yuan-luo Could you take a next step look please? Thanks

PR #16859, opened 2026-01-10
[RL] DeepEP support for `--enable-return-routed-experts`

PrinsYin:
Linked issue: https://github.com/THUDM/slime/issues/1316

## Summary
- Add DeepEP a2a support to `RoutedExpertsCapturer` (rebased onto `srt/state_capturer/` after #24450)
- Each attn-TP rank only sees its scattered slice of `topk_ids` under DeepEP, so all-gather across attn-TP at capture time before staging into `device_cache`
- Alternative late-gather impl in #17892 by @ocss884 (gather at D2H sync); this PR keeps the early-gather approach
- Replace the R3 capture test with a config that actually exercises the gather path (`--tp 4 --dp 2 --moe-a2a-backend deepep`, `attn_tp_size == 2`); previous `tp=dp=2` config left the gather as dead code


hnyls2002:
/rerun-test test/registered/rl/test_return_routed_experts.py
(4 tries)

PR #16860, opened 2026-01-10
Expose Model Parallelism Information 

JD-ETH:
## Motivation

We want to enable model shards to be initialized outside of sglang in a way that's identical to sglang's ModelLoader.  
The external initialization workflow will look something like this: 

```
sglang.srt.server_args._global_server_args = server_args
model_parallelism_info = engine.get_parallelism_config(rank) 
with ParallelismContext(RankParallelismConfig.from_dict(model_parallelism_info)):
   model = get_model(
                model_config=model_config,
                load_config=load_config,
                device_config=device_config,
            )
```

The primary use case of this is to enable train -> inference weight transfer. with [14997](https://github.com/sgl-project/sglang/pull/14997), we already have transfer engine and weight registration, but the transfer can only happen between weights of identical shape. One option is to instantiate the model shards outside, register the weights, and send them to sglang via the existing registrations. 

I will however leave draft_tp for future work. The test currently verifies: TP, EP and Moe with DP Attention, with the CI test only has the most simple tp case.  

## Modifications
Whenever transfer engine is enabled, we additionally expose the full parallelism configurations of tp/pp/ep/attn_tp/attn_dp information through http API. The info is propagated to the _global_states just like the rdma weight register does at initialization time, so no performance impact. 

A small fix was made to pass dp_parallel_controller's scheduler_infos back to the global  engine instance. This allows remote_weight_info APIs to work with dp>1. 

## Accuracy Tests
`python -m sglang.launch_server --model-path qwen/qwen2.5-0.5b-instruct --remote-instance-weight-loader-start-seed-via-transfer-engine` 
`curl http://localhost:30000/get_parallelism_config` 
returns: 
`{"rank":0,"tp_size":1,"tp_rank":0,"pp_size":1,"pp_rank":0,"ep_size":1,"ep_rank":0,"attn_tp_size":1,"attn_tp_rank":0,"attn_dp_size":1,"attn_dp_rank":0,"world_size":1,"global_rank":0}` etc

the `test_parallelism_context_integration.py` shows the intended use case and that the model parameters match. 

## Checklist

## Review Process


python/sglang/srt/entrypoints/http_server.py
```diff
@@ -898,6 +900,30 @@ async def get_remote_instance_transfer_engine_info(rank: int = None):
         return Response(status_code=HTTPStatus.BAD_REQUEST)
 
 
+@app.get("/get_parallelism_config")
```
slin1237: this is server info?
can we not add another endpoint like this

slin1237 (changes requested):

python/sglang/srt/entrypoints/http_server.py
slin1237: please stop adding verb to http endpoint unless it can be justified

JD-ETH: this is the rank specific parallelism info, like tp_rank, ep_rank, dp_attention_rank, etc,

slin1237: I know. 
shouldn't this be server info?
why another endpoint to expose same category of info

JD-ETH: similar to `get_remote_instance_transfer_engine_info`, as the info is on a rank level not on a server level. Or do you prefer having the server level dumping parallelism info for every rank at once?

JD-ETH: as we are doing a P2P transfer, we want to query the exact sglang engine rank the current trainer rank will be responsible for.

python/sglang/srt/entrypoints/engine.py
```diff
@@ -200,6 +201,9 @@ def __init__(self, **kwargs):
                 scheduler_init_result.scheduler_infos
             )
         )
+        self.parallelism_config = parse_parallelism_config_from_scheduler_infos(
+            scheduler_infos
```
amysaq2023: should it be `scheduler_init_result.scheduler_infos` here?

JD-ETH: thanks for catching, let me fix!

PR #11910, opened 2025-10-21
[Perf] Optimize multimodal mm_inputs process in scheduler

yuan-luo:
## Motivation

In VL model, mm_input processing (mainly from_dict) in Scheduler is an important part. For a 2MB video file, the whole time cost range of the inference is as following:
image

After the Scheduler receives the request, processing mm_inputs takes about 500 ms, with most time frame spent in `from_dict`.

Although `from_dict` appears to do nothing more than `setattr`, the deeper reason is that recv_req is delivered via ZMQ recv_pyobj (or subsequent broadcast_pyobj). mm_inputs typically contains custom class instances (e.g., Video items), and during unpickling/__setstate__ or on first attribute access they may:

- decode base64/bytes into PIL.Image/np.ndarray;
- perform size/channel checks or copies;
- run normalization, pad-parameter calculations, etc.

In from_dict, the first traversal of obj["mm_items"] forces materialization of the entire object list and its internal state; this time is then attributed to from_dict.

In TP8 scenario, all the Scheduler TP ranks will run this duplicated computing portion in CPU which occupies the main thread CPU cycle. This computing logic actually needs to be run merely on TP0 and be broadcast to other TP ranks. 
Since the scheduler is single-threaded, any large CPU cost here will impact handling of other messages. For example, CPU hits 99.9% can significantly increase the cuda kernel launch time.

image

Moreover, due to video preprocessing in TokenizerManager leverages `decord` to do frame interpolation, it randomly picks up the CPU cores and can be overlapped with the core running Scheduler, in this point of view, this PR can potentially mitigate the preprocess performance downgrade. We attempted to leverage GPU offload to do video preprocessing, while there's another issue addressing that GPU context has worse performance than CPU in decord. https://github.com/dmlc/decord/issues/359

## Modifications

The change is to perform deserialization of multimodal mm_inputs (from_dict + validation + set_pad_value) only once on the entry Scheduler (rank 0 within the TP group / Attn-TP0). 
The entry rank (rank 0) then broadcasts the materialized MultimodalInputs to the other TP ranks via the CPU communication group using broadcast_object_list. The other ranks no longer call from_dict; they only do local pad_input_ids and extend_image_inputs. 
This reduces Scheduler CPU utilization and avoids contending with request scheduling.
image

## Accuracy Tests

```
➜  sglang python -m sglang.launch_server --model-path Qwen/Qwen2.5-VL-7B-Instruct --host 0.0.0.0 --port 30000  --trust-remote-code --tp-size=4

[...]

PR #16876, opened 2026-01-10
Clarify the meaning of `cpu_group` / `entry_rank` when dp + tp is enabled.

hnyls2002:
Introduce `dp_tp_*`, which is an request/data-plane coordination group (not tensor collectives). When DP attention is enabled, scope to the attention-TP group; otherwise use the base TP group. Entry rank is the local rank 0 in that group. 

Use the CPU (gloo) group to broadcast VLM Python objects and avoid CUDA tream/device coupling (#11910).

### DIFF
--- a/python/sglang/srt/disaggregation/prefill.py
+++ b/python/sglang/srt/disaggregation/prefill.py
@@ -603,7 +603,7 @@ class SchedulerDisaggregationPrefillMixin:
         """
         polls = poll_and_all_reduce(
             [req.disagg_kv_sender for req in self.disagg_prefill_inflight_queue],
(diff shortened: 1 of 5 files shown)


ShangmingCai (approved):
LGTM. Let us check if all dp- related CIs work normally.

hnyls2002:
@ch-wan, I just found that the DP attention to group will initialize, no matter whether the DP attention is enabled or not.

Shall we just:
- Disable initializing PD attention when there are only pure TPs.
- Or use the always-initialized DP-attention-TP-group when we need to do TP sync.

PR #16890, opened 2026-01-11
Add Liquid Foundation Model (LFM2)

tugot17:
## Summary
- Add support for LiquidAI's LFM2 hybrid architecture (attention + ShortConv layers)
- LFM2 uses gated 1D causal convolution (kernel=3) instead of attention in some layers, requiring SGLang's hybrid caching system
- Tested with [LiquidAI/LFM2.5-1.2B-Instruct](https://huggingface.co/LiquidAI/LFM2.5-1.2B-Instruct) and [LiquidAI/LFM2-2.6B-Exp](https://huggingface.co/LiquidAI/LFM2-2.6B-Exp)

## Changes
| File | Purpose |
|------|---------|
| `srt/models/lfm2.py` | Full model implementation with ShortConv layers |
| `srt/configs/lfm2.py` | LFM2 config with `mamba2_cache_params` |
| `srt/configs/mamba_utils.py` | Dynamic conv dtype selection (fixes CUDA graph capture) |
| `srt/model_executor/model_runner.py` | Add LFM2 to hybrid model detection |
| `srt/function_call/lfm2_detector.py` | Streaming parser for LFM2 tool call format |
| `srt/function_call/function_call_parser.py` | Register `lfm2` parser |
| `test/srt/models/test_generation_models.py` | Add LFM2 to generation test suite |
| `test/registered/function_call/test_function_call_parser.py` | Unit tests for `Lfm2Detector` |
| `test/registered/openai_server/function_call/test_tool_choice.py` | Integration tests (`TestToolChoiceLfm2`) |

### ShortConv Architecture
- ShortConv layers use fixed-size cache (2 tokens) vs attention's growing KV cache
- Uses `HybridReqToTokenPool` + `MambaPool` for hybrid caching
- Real convolution done by Triton kernel `causal_conv1d_fn()`

### Conv State Dtype Fix
The `causal_conv1d_update` kernel requires conv state dtype to match input dtype exactly. We fixed a dtype mismatch that caused CUDA graph capture to fail:

- **Problem**: `mamba_utils.py` hardcoded `CONV_DTYPE = torch.bfloat16`, but tests run models in `torch.float16`
- **Root cause**: Server path sets `SGLANG_MAMBA_SSM_DTYPE` env var via `ServerArgs`, but test path uses `Engine` directly (bypassing this)
- **Solution**:
  - Added `get_conv_dtype()` function that dynamically gets dtype from `torch.get_default_dtype()`
  - Added default fallback for `SGLANG_MAMBA_SSM_DTYPE` env var
  - Wrapped `init_memory_pool()` in `set_default_torch_dtype(self.model_config.dtype)` context

This fix is safe for other Mamba-based models (NemotronH, FalconH1, Qwen3Next) - server behavior is unchanged.

### Function Calling Support
Added `Lfm2Detector` for parsing LFM2's tool call format with special tokens:

<|tool_call_start|>[get_weather(city="Paris")]<|tool_call_end|>

Usage: `--tool-call-parser lfm2`

## Tests

Logprob accuracy test (compares SGLang vs HuggingFace):
```bash
SGLANG_MAMBA_CONV_DTYPE=float16 ONLY_RUN=LiquidAI/LFM2.5-1.2B-Instruct pytest test/registered/models/test_generation_models.py -v -s
```

LFM2 function call parser (unit test):
```
pytest test/registered/function_call/test_function_call_parser.py::TestLfm2Detector -v -s
```

LFM2 function calling (integration test):

```
pytest test/registered/openai_server/function_call/test_tool_choice.py::TestToolChoiceLfm2 -v -s
```

## Benchmark Performance

We run `MMLU-Pro` and `tau2` on our internal OAI-server compatible benchmarking suite and the scores match the [declared](https://huggingface.co/LiquidAI/LFM2.5-1.2B-Instruct#benchmarks) performance. 

Running on `1xH100 SXM5`

```bash
#!/usr/bin/env bash
set -e

MODEL="${MODEL:-LiquidAI/LFM2.5-1.2B-Instruct}"
# MODEL="${MODEL:-LiquidAI/LFM2-2.6B-Exp}"
HOST="${HOST:-127.0.0.1}"
PORT="${PORT:-30001}"

echo "Starting SGLang server: $MODEL on $HOST:$PORT"

sglang serve \
    --model-path "$MODEL" \
    --host "$HOST" \
    --port "$PORT" \
    --tool-call-parser lfm2

```

```bash
Max request concurrency:                 not set
Successful requests:                     500
Benchmark duration (s):                  3.77
Total input tokens:                      127216
Total input text tokens:                 127216
Total generated tokens:                  63834
Total generated tokens (retokenized):    62640
Request throughput (req/s):              132.79
Input token throughput (tok/s):          33786.60
Output token throughput (tok/s):         16953.32
Peak output token throughput (tok/s):    33486.00
Peak concurrent requests:                500
Total token throughput (tok/s):          50739.92
Concurrency:                             404.12
Mean E2E Latency (ms):                   3043.26
Median E2E Latency (ms):                 3341.52
Mean TTFT (ms):                          863.90
Median TTFT (ms):                        786.99
P99 TTFT (ms):                           1621.48
Mean TPOT (ms):                          32.01
Median TPOT (ms):                        18.84
P99 TPOT (ms):                           310.85
Mean ITL (ms):                           17.24
Median ITL (ms):                         8.57
P95 ITL (ms):                            18.11
P99 ITL (ms):                            120.99
Max ITL (ms):                            996.73
```


python/sglang/srt/function_call/lfm2_detector.py
```diff
@@ -0,0 +1,431 @@
...
+                string_char = None
+            elif not in_string:
+                if char == "[":
+                    bracket_count += 1
+                elif char == "]":
+                    bracket_count -= 1
+                    if bracket_count == 0:
+                        return i
+
+        return -1  # No matching bracket found
```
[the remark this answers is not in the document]
tugot17: @gemini-code-assist
Yeah fair, it was something that remained from the other detector. I removed the `_find_matching_bracket` and rerun the unit test and integration test (as described in the PR summary) all works fine.

python/sglang/srt/configs/mamba_utils.py
```diff
@@ -41,16 +41,25 @@ class Mamba2StateDType:
     temporal: torch.dtype
 
 
-CONV_DTYPE = torch.bfloat16
+def get_conv_dtype() -> torch.dtype:
+    """Get conv state dtype - uses torch default dtype which is set by model runner."""
+    default_dtype = torch.get_default_dtype()
```
yizhang2077: could we use env set conv type like mamba dtype does?

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -544,7 +545,9 @@ def initialize(self, min_per_gpu_memory: float):
         self.configure_kv_cache_dtype()
 
         # Init memory pool and attention backends
-        self.init_memory_pool(min_per_gpu_memory)
+        # Set default dtype so mamba2_cache_params picks up the correct dtype for conv state
+        with set_default_torch_dtype(self.model_config.dtype):
```
yizhang2077: I think it is unnecessary

tugot17: yeah, there was an issue when we initialized the model with fp16, but now I just propagate the dtype. 

@yizhang2077 wdyt?

python/sglang/srt/configs/mamba_utils.py
tugot17: undone this

tugot17:
Fixed! The issue was that LFM2 wasn't in the SM100 special handling in server_args.py.

On B200, the default backend is `trtllm_mha` which forces `page_size=64`, but `MambaRadixCache` 
requires `page_size=1`. Added `Lfm2ForCausalLM` with the same handling as `NemotronH` — uses 
`flashinfer` on `SM100` instead.

(Also had to block triton backend since LFM2's first layer is a conv layer, not attention.)

The tests pass now on B200 as well

ispobock:
@ChangyiYang Could you share the test results?

python/sglang/srt/configs/model_config.py
```diff
@@ -192,6 +192,10 @@ def __init__(
         )
         self.dtype = _get_and_verify_dtype(self.hf_text_config, dtype)
 
+        # Propagate runtime dtype to hf_config so that hybrid models (mamba, LFM2, etc.)
+        # can use it for conv state cache dtype
+        self.hf_config.torch_dtype = self.dtype
```
yizhang2077: could we use other way to pass dtype to conv, I doubt it may affect other models here.
Besides this, it looks good for me in other part

ChangyiYang:
I run model with command
```
#!/usr/bin/env bash
set -e

export CUDA_VISIBLE_DEVICES=1

MODEL="${MODEL:-LiquidAI/LFM2.5-1.2B-Instruct}"
# MODEL="${MODEL:-LiquidAI/LFM2-2.6B-Exp}"
HOST="${HOST:-127.0.0.1}"
PORT="${PORT:-30001}"

echo "Starting SGLang server: $MODEL on $HOST:$PORT"

sglang serve \
    --model-path "$MODEL" \
    --host "$HOST" \
    --port "$PORT" \
    --tool-call-parser lfm2
```

 The test command and result is:
```
root@2763e46d6169:/sgl-workspace/sglang# python3 benchmark/gsm8k/bench_sglang.py --num-questions 1000 --host http://127.0.0.1 --
port 30001
100%|███████████████████████| 1000/1000 [00:16<00:00, 60.97it/s]
Accuracy: 0.554
Invalid: 0.001
Latency: 16.421 s
Output throughput: 8042.606 token/s
```

Is this expected performace?

JustinTong0323:
We also get following result for `LiquidAI/LFM2.5-1.2B-Instruct` using lm_eval:
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.6126|±  |0.0134|
|     |       |strict-match    |     5|exact_match|↑  |0.6118|±  |0.0134|
|mmlu_pro|      2|custom-extract|      |exact_match|↑  |0.3661|±  |0.0043|
```

@tugot17  Could you please confirm it's the evaluation mismatch or the implementation's issue? Thanks~

python/sglang/srt/configs/lfm2.py
```diff
@@ -0,0 +1,110 @@
...
+        # get_attention_tp_size() requires initialization, default to 1 if not available
+        try:
+            tp_size = get_attention_tp_size()
+        except (AssertionError, RuntimeError):
+            tp_size = 1
+
+        # For ShortConv layers, we use a simplified Mamba2StateShape
+        # LFM2 doesn't use SSM state (state_size=0), only conv state
+        shape = Mamba2StateShape.create(
+            tp_world_size=tp_size,
```
yizhang2077, line 81: I think we need to refactor it later but it is ok for current pr. I think ShortConv-only models being mixed with mamba models is tricky here. cc @ispobock @hebiao064

ispobock: Yes, we need to do some refactor later.

python/sglang/srt/configs/model_config.py
ispobock: agree with that

ispobock: @tugot17 could you address this comment?

tugot17: @yizhang2077 what do you think about changing it in `mamba_utils.py`

```py
def mamba2_state_dtype() -> Mamba2StateDType:
    dtype_map = {
        "float32": torch.float32,
        "bfloat16": torch.bfloat16,
        "float16": torch.float16,
    }
    conv_dtype = dtype_map.get(
        os.environ.get("SGLANG_MAMBA_CONV_DTYPE", "bfloat16"), torch.bfloat16
    )
    ssm_dtype = dtype_map.get(
        os.environ.get("SGLANG_MAMBA_SSM_DTYPE", "float32"), torch.float32
    )
    return Mamba2StateDType(conv=conv_dtype, temporal=ssm_dtype)
```

This way we could modify `SGLANG_MAMBA_CONV_DTYPE` as an ENV VARIABLE, similar to how it is already done for other models. Than the tests that use `fp16` should just pass. 

Would this make sense? I agree the current version is kinda too hacky.

yizhang2077: I think we could use this way.

yizhang2077 (approved):

python/sglang/srt/configs/model_config.py
tugot17: @yizhang2077

I also added this to the tests so they don't require setting the flag manually 

```python
def assert_close_logits_and_output_strs(
        self,
        prompts: List[str],
        model_case: ModelCase,
        torch_dtype: torch.dtype,
    ) -> None:
        model_path = model_case.model_path
        prefill_tolerance, decode_tolerance, rouge_l_tolerance = (
            model_case.prefill_tolerance,
            model_case.decode_tolerance,
            model_case.rouge_l_tolerance,
        )
        max_new_tokens = 32

        # Set conv dtype for hybrid models to match inference dtype
        dtype_str = {torch.float16: "float16", torch.bfloat16: "bfloat16"}.get(
            torch_dtype, "bfloat16"
        )
        os.environ["SGLANG_MAMBA_CONV_DTYPE"] = dtype_str
```

tugot17:
@yizhang2077 Could you approve again? I added one commit to make tests more smooth, see [comment](https://github.com/sgl-project/sglang/pull/16890#discussion_r2713157406)

test/registered/models/test_generation_models.py
```diff
@@ -134,6 +138,12 @@ def assert_close_logits_and_output_strs(
         )
         max_new_tokens = 32
 
+        # Set conv dtype for hybrid models to match inference dtype
+        dtype_str = {torch.float16: "float16", torch.bfloat16: "bfloat16"}.get(
+            torch_dtype, "bfloat16"
+        )
+        os.environ["SGLANG_MAMBA_CONV_DTYPE"] = dtype_str
```
ispobock, line 145: could you use `temp_set_env` context https://github.com/sgl-project/sglang/blob/a95c9f5b815cc4fed373e1f3483868652a67a778/python/sglang/srt/environ.py#L10

tugot17: @ispobock It will raise `ValueError` I think

```python
@contextmanager
def temp_set_env(**env_vars: dict[str, Any]):
    """Temporarily set non-sglang environment variables, e.g. OPENAI_API_KEY"""
    for key in env_vars:
        if key.startswith("SGLANG_") or key.startswith("SGL_"):
            raise ValueError("temp_set_env should not be used for sglang env vars")
```

ispobock: oh, it's `override` https://github.com/sgl-project/sglang/blob/a95c9f5b815cc4fed373e1f3483868652a67a778/python/sglang/srt/environ.py#L76
you can use it like this: https://github.com/sgl-project/sglang/blob/f2ae066a6b77afa82800174e92efa7067db57830/test/manual/ep/test_eplb.py#L27
this can let the env will not affect other test case.

ispobock: we can update it later.

ispobock: updated in https://github.com/sgl-project/sglang/pull/17566

PR #16901, opened 2026-01-11
Implement the Top-2048 algorithm using Bitonic Sort

fan-wenjie:
Implemented Top-2048 using a BitonicSort-based algorithm, which improves performance compared to the previous RadixSort-based implementation. On a mobile RTX 4060 GPU, the execution time of test_topk.py decreased from 40 seconds to 38 seconds. Both code size and shared memory usage have been reduced.

### DIFF
--- a/sgl-kernel/csrc/elementwise/topk.cu
+++ b/sgl-kernel/csrc/elementwise/topk.cu
@@ -1,11 +1,3 @@
-/**
- * @NOTE: This file is adapted from
- * https://github.com/tile-ai/tilelang/blob/main/examples/deepseek_v32/topk_selector.py
- * We:
- * 1. adapt from tilelang to pure cuda
- * 2. optimize the performance a little
- * 3. fix the potential illegal memory access
- */
 #include <ATen/core/TensorBase.h>
 #include <ATen/core/TensorBody.h>
 #include <c10/cuda/CUDAStream.h>
@@ -21,19 +13,7 @@
 namespace {
 
 constexpr int TopK = 2048;
-constexpr int kThreadsPerBlock = 1024;
-
-#ifdef USE_ROCM
-// On ROCm, the per-workgroup LDS budget depends on the target arch, so we inject a
-// per-arch value from `setup_rocm.py` via `-DSGL_TOPK_DYNAMIC_SMEM_BYTES=...`.
-#ifdef SGL_TOPK_DYNAMIC_SMEM_BYTES
-constexpr size_t kSmem = static_cast<size_t>(SGL_TOPK_DYNAMIC_SMEM_BYTES);
-#else
-constexpr size_t kSmem = 48 * 1024;  // bytes
-#endif
-#else
-constexpr size_t kSmem = 32 * 1024 * sizeof(uint32_t);  // 128KB (bytes)
-#endif
+constexpr int kThreadsPerBlock = 512;
 
 struct FastTopKParams {
   const float* __restrict__ input;         // [B, input_stride]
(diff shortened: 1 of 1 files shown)


sgl-kernel/csrc/elementwise/topk.cu
```diff
@@ -43,208 +23,138 @@ struct FastTopKParams {
...
-          ::atomicAdd(&s_histogram[sub_bin], 1);
-        }
-      }
+    sorter.sort2048(value, true);
+#pragma unroll
+    for (int i = 0; i < bitonic_sort_t::kValueSize; ++i) {
+      result[i] = min(result[i], value[i]);
     }
-    __syncthreads();
+    sorter.merge2048(result, false);
```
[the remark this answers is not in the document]
fan-wenjie: > critical
> 
> The logic for combining the top-K results from different chunks seems incorrect. The element-wise `min` operation followed by an ascending merge (`merge2048(result, false)`) is not a valid way to merge two sets of candidates to find the overall top-K largest elements.
> 
> `result` holds the current top-K candidates, and `value` holds the sorted candidates from a new chunk. To find the new top-K, you need to perform a merge-select operation, not an element-wise `min`.
> 
> After `sorter.sort2048(value, true)`, `value` is sorted in descending order of keys. `result` is expected to hold the current top-K, also sorted. The `min` operation on the keys does not correctly combine these two sets. For example, if `result[i]` corresponds to a large value and `value[i]` to a small one, `min` will pick the small one, potentially discarding a valid top-K candidate.
> 
> Furthermore, `merge2048` is called with `descend=false`, which sorts `result` in ascending order, while `sort2048` on `value` sorts in descending order. This inconsistency is problematic.
> 
> The correct approach would be to merge the two sorted lists (`result` and `value`) and keep the top `TopK` elements. The fact that the tests allow for a non-zero error margin (`max_permit_error=5`) suggests this implementation may be inexact, which is a concern for a top-K algorithm.

bits = (bits & 0x80000000u) ? (bits - 1) : (bits ^ 0x7fffffffu); It can map relatively large values to relatively small values, transforming the problem of taking the maximum value of topk into the problem of taking the minimum value of topk.
By comparing the elements of an ascending sequence and a descending sequence, the smallest top-halfsize sequence can be quickly identified.

sgl-kernel/csrc/elementwise/topk.cu
```diff
@@ -43,208 +23,138 @@ struct FastTopKParams {
...
-  uint16_t bits = __half_as_ushort(h);
-  uint16_t key = (bits & 0x8000) ? static_cast<uint16_t>(~bits) : static_cast<uint16_t>(bits | 0x8000);
-  return static_cast<uint8_t>(key >> 8);
-}
-
-__device__ __forceinline__ auto convert_to_uint32(float x) -> uint32_t {
+__device__ __forceinline__ auto pack_float_uint32(float x, int y) -> uint64_t {
   uint32_t bits = __float_as_uint(x);
-  return (bits & 0x80000000u) ? ~bits : (bits | 0x80000000u);
+  bits = (bits & 0x80000000u) ? (bits - 1) : (bits ^ 0x7fffffffu);
```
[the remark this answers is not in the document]
fan-wenjie: > critical
> 
> The logic for converting a float to a sortable integer representation appears to be incorrect. The current implementation does not correctly preserve the relative order of floating-point numbers, especially across the positive/negative boundary, which is necessary for a correct descending sort.
> 
> A standard way to create a sortable integer key that preserves ascending float order is:
> 
> ```c++
> bits = (bits & 0x80000000u) ? ~bits : (bits | 0x80000000u);
> ```
> 
> Since the bitonic sort is called with `descend=true`, it performs a descending sort on the keys. To get the top-K largest float values, the keys should be ordered ascendingly with the float values. The current logic fails to do this correctly.
> 
> For example, with the current logic:
> 
> * `1.0f` (0x3f800000) maps to `0x407fffff`
> * `-1.0f` (0xbf800000) maps to `0xbf7fffff`
>   Here `1.0f > -1.0f`, but `0x407fffff < 0xbf7fffff`, which will lead to incorrect sorting when comparing a positive and a negative value.
> 
> I suggest using the well-established method for this conversion, which was also used in the previous implementation.
> 
> ```
>   bits = (bits & 0x80000000u) ? ~bits : (bits | 0x80000000u);
> ```

bits = (bits & 0x80000000u) ? (bits - 1) : (bits ^ 0x7fffffffu); It can map relatively large values to relatively small values, transforming the problem of taking the maximum value of topk into the problem of taking the minimum value of topk. when the float value is positive, bits < 0x80000000u. bits ^ 0x7fffffffu invert the lower 31 bits. when the float value is negative, bits - 1 can make +0.0f and -0.0f map to same value 0x7fffffffu. Concatenate the value and sequence number into uint64_t

sgl-kernel/csrc/elementwise/topk.cu
```diff
@@ -21,19 +13,7 @@
 namespace {
 
 constexpr int TopK = 2048;
-constexpr int kThreadsPerBlock = 1024;
```
yiakwy-xpu-ml-framework-team, line 24: The original algorithm is verified with fast implemenation of radix sort:

a) bitnoic topk complexity : logn * logk
b) radix top-k complexity : logn * w (w=16/32/64) (as suggestd by flashInfer author)

where n is the hidden size, and k is topk used for DeepSeek V3.2 lighting indexer :

k=2048, logk =11
k=1024, logk=10
k=512, logk=9

The theoretical imporovement could be achieved as maximum of half of the radix sorting algorithm.

I remember K=2048 is required by DeepSeek V3.2.

Let's work on it.

sgl-kernel/csrc/elementwise/topk.cu
```diff
@@ -21,19 +13,7 @@
...
-#ifdef USE_ROCM
-// On ROCm, the per-workgroup LDS budget depends on the target arch, so we inject a
-// per-arch value from `setup_rocm.py` via `-DSGL_TOPK_DYNAMIC_SMEM_BYTES=...`.
-#ifdef SGL_TOPK_DYNAMIC_SMEM_BYTES
-constexpr size_t kSmem = static_cast<size_t>(SGL_TOPK_DYNAMIC_SMEM_BYTES);
-#else
-constexpr size_t kSmem = 48 * 1024;  // bytes
-#endif
-#else
-constexpr size_t kSmem = 32 * 1024 * sizeof(uint32_t);  // 128KB (bytes)
```
yiakwy-xpu-ml-framework-team, line 35: This is valued used by ROCM SDK

sgl-kernel/csrc/elementwise/topk.cu
```diff
@@ -43,208 +23,140 @@ struct FastTopKParams {
   int64_t input_stride;
 };
 
-// when length <= TopK, we can directly write the indices
-__device__ void naive_topk_cuda(const float* __restrict__ score, int32_t* __restrict__ indice, int32_t length) {
-  const auto tid = threadIdx.x;
-  for (int i = tid; i < TopK; i += kThreadsPerBlock) {
-    indice[i] = (i < length) ? i : -1;
-  }
+__device__ __forceinline__ auto pack_float_uint32(float x, int y) -> uint64_t {
```
yiakwy-xpu-ml-framework-team, line 26: is it necessary to pack 32 bits into 64 bits ? (bits_x << 32 | bits_y)

fan-wenjie: > is it necessary to pack 32 bits into 64 bits ? (bits_x << 32 | bits_y)

Yes, It should pack double 32-bits into 64-bits. The main reason for the acceleration is the reduction of high-time-consuming instructions such as atomicAdd and syncthreads. Each iteration of 2048 numbers only requires the execution of 10 syncthreads (bar.sync)

sgl-kernel/csrc/elementwise/topk.cu
```diff
@@ -43,208 +23,140 @@ struct FastTopKParams {
...
-// keep the first `length` entries, set others to -1
-__device__ void naive_topk_transform(
-    const float* __restrict__ score,
-    int32_t length,
-    int32_t* __restrict__ dst_page_table,
-    const int32_t* __restrict__ src_page_table) {
-  const auto tid = threadIdx.x;
-  for (auto i = tid; i < TopK; i += kThreadsPerBlock) {
-    dst_page_table[i] = (i < length) ? src_page_table[i] : -1;
+template <typename T, int kThreadsPerBlock, int kWarpSize>
```
yiakwy-xpu-ml-framework-team, line 32: Have you tried some simpler one ?

```
/**
 * Applies the bitonic sorting algorithm to each thread. It swaps two
 * elements in the two lists if they are out of place.
 */
__global__ __forceinline__ void bitonicSort(float* a, int blockSize) {
	// First we need to figure out what index each thread will access
    int index = threadIdx.x + blockIdx.x * blockSize;
	atomicMin(&a[i + index], 
		atomicMax(&a[i + blockSize + index], a[i + index]));
	__syncthreads();
}

/**
 * Combines two bitonic sequences together to create a new bitonic sequence.
 * @param a Pointer to the start of the bitonic sequence.
 * @param blockSize The size of each sub-array partition.
 * @param t Determines when to switch between ascending and descending.
 */
__global__ __forceinline__ void bitonicBuild(float* a, int blockSize, int t) {
	int index = threadIdx.x + blockIdx.x * blockSize, x = 0, asc = 1;
	float* b = a + index + (blockSize / 2);
	while(x > index) {
		x += t;
		asc++;
	}
	
	if(asc % 2 == 1) {
		atomicMin(&a[index], atomicMax(&b, a[index]));
	}
	else {
		atomicMax(&a[index], atomicMin(&b, a[index]));
	}
}
```

And bitonic sort from CUDA SDK. CUDA has native support of bitonic sort just like radix sort.


Our problem can summarized :

1) try baseline bitnoic sort vs radix sort
2) develop better bitonic sort for K=2048 and hiddensize

yiakwy-xpu-ml-framework-team (commented):
To summary:

1) K=2048 is required by DeepSeek V3.2, we need to verify it
2) K=512 should have better performance, since flashinfer uses w=16
3) we need baseline bitonic sort from CUDA SDK to compare against radix sort (w=16)
4) we need align the twistted bitonic sort for the indexer with the baseline
5) we need test/benchmark scripts, since the codes is not trival; here is an example : https://github.com/Relaxed-System-Lab/Flash-Sparse-Attention/blob/main/benchmarks/full_attn_bench.py


TO DO:

1) better add a new function with test and benchmark scripts
2) add bitonic baseline

I can support you with MI355X and we need to verify the performance on the latest workhorse.
截屏2026-01-12 21 16 14

sgl-kernel/csrc/elementwise/topk.cu
fan-wenjie: > Have you tried some simpler one ?
> 
> ```
> /**
>  * Applies the bitonic sorting algorithm to each thread. It swaps two
>  * elements in the two lists if they are out of place.
>  */
> __global__ __forceinline__ void bitonicSort(float* a, int blockSize) {
> 	// First we need to figure out what index each thread will access
>     int index = threadIdx.x + blockIdx.x * blockSize;
> 	atomicMin(&a[i + index], 
> 		atomicMax(&a[i + blockSize + index], a[i + index]));
> 	__syncthreads();
> }
> 
> /**
>  * Combines two bitonic sequences together to create a new bitonic sequence.
>  * @param a Pointer to the start of the bitonic sequence.
>  * @param blockSize The size of each sub-array partition.
>  * @param t Determines when to switch between ascending and descending.
>  */
> __global__ __forceinline__ void bitonicBuild(float* a, int blockSize, int t) {
> 	int index = threadIdx.x + blockIdx.x * blockSize, x = 0, asc = 1;
> 	float* b = a + index + (blockSize / 2);
> 	while(x > index) {
> 		x += t;
> 		asc++;
> 	}
> 	
> 	if(asc % 2 == 1) {
> 		atomicMin(&a[index], atomicMax(&b, a[index]));
> 	}
> 	else {
> 		atomicMax(&a[index], atomicMin(&b, a[index]));
> 	}
> }
> ```
> 
> And bitonic sort from CUDA SDK. CUDA has native support of bitonic sort just like radix sort.
> 
> Our problem can summarized :
> 
> 1. try baseline bitnoic sort vs radix sort
> 2. develop better bitonic sort for K=2048 and hiddensize

Utilize the faster implementation of bitonicSort that I have developed, fully leverage data exchange within the warp, reduce the usage of __shared__ memory, and avoid atomic instructions. The implementation of bitonicSort in the CUDA SDK is not particularly good

fan-wenjie: > Have you tried some simpler one ?
> 
> ```
> /**
>  * Applies the bitonic sorting algorithm to each thread. It swaps two
>  * elements in the two lists if they are out of place.
>  */
> __global__ __forceinline__ void bitonicSort(float* a, int blockSize) {
> 	// First we need to figure out what index each thread will access
>     int index = threadIdx.x + blockIdx.x * blockSize;
> 	atomicMin(&a[i + index], 
> 		atomicMax(&a[i + blockSize + index], a[i + index]));
> 	__syncthreads();
> }
> 
> /**
>  * Combines two bitonic sequences together to create a new bitonic sequence.
>  * @param a Pointer to the start of the bitonic sequence.
>  * @param blockSize The size of each sub-array partition.
>  * @param t Determines when to switch between ascending and descending.
>  */
> __global__ __forceinline__ void bitonicBuild(float* a, int blockSize, int t) {
> 	int index = threadIdx.x + blockIdx.x * blockSize, x = 0, asc = 1;
> 	float* b = a + index + (blockSize / 2);
> 	while(x > index) {
> 		x += t;
> 		asc++;
> 	}
> 	
> 	if(asc % 2 == 1) {
> 		atomicMin(&a[index], atomicMax(&b, a[index]));
> 	}
> 	else {
> 		atomicMax(&a[index], atomicMin(&b, a[index]));
> 	}
> }
> ```
> 
> And bitonic sort from CUDA SDK. CUDA has native support of bitonic sort just like radix sort.
> 
> Our problem can summarized :
> 
> 1. try baseline bitnoic sort vs radix sort
> 2. develop better bitonic sort for K=2048 and hiddensize

It Pass test_topk.py. and TopK=2048. 
ThreadsPerBlock is 512.

yiakwy-xpu-ml-framework-team:
Related PR : 

clean topk : https://github.com/sgl-project/sglang/pull/16945

sgl-kernel/csrc/elementwise/topk.cu
```diff
@@ -43,208 +23,140 @@ struct FastTopKParams {
...
+struct bitonic_sort {
+  T (&buffer)[TopK];
+  static_assert(
+      kThreadsPerBlock * kThreadsPerBlock >= kWarpSize * TopK,
+      "assert failed: kThreadsPerBlock * kThreadsPerBlock >= kWarpSize * TopK");
+  static constexpr int kValueSize = TopK / kThreadsPerBlock;
```
DarkSharpness, line 38: `kValueSize` is confusing here. Maybe `kLocalSize` or `kElementsPerThread` is better.

sgl-kernel/csrc/elementwise/topk.cu
```diff
@@ -256,10 +168,11 @@ __global__ __launch_bounds__(kThreadsPerBlock)  // topk
   const auto length = lengths[bid];
   const auto indice = indices + bid * TopK;
   const auto score = input + bid * input_stride;
-  if (length <= TopK) {
-    return naive_topk_cuda(score, indice, length);
-  } else {
-    return fast_topk_cuda_tl(score, indice, row_start, length);
+  int index[bitonic_sort_t::kValueSize];
+  fast_topk_cuda_tl(score, index, row_start, length);
```
DarkSharpness, line 172: Why not using `naive_topk_cuda` or something like that here? For short length, we can skip the topk operation by simply filling from 0 to length.

sgl-kernel/csrc/elementwise/topk.cu
```diff
@@ -256,10 +168,11 @@ __global__ __launch_bounds__(kThreadsPerBlock)  // topk
...
+#pragma unroll
+  for (int i = 0; i < bitonic_sort_t::kValueSize; ++i) {
+    indice[threadIdx.x * bitonic_sort_t::kValueSize + i] = index[i];
```
DarkSharpness, line 175: Is this global memory access coalesced? If we don't care about the output order, `indices[i * kThreadsPerBlock + threadIdx.x]` might be more efficient since the memory op is coalesced.

If your profiling shows that current implementation is faster, skip my comment

sgl-kernel/csrc/elementwise/topk.cu
```diff
@@ -360,20 +255,11 @@ __global__ __launch_bounds__(kThreadsPerBlock)  // prefill, ragged kv
...
-    const auto pos_0 = s_indices[idx_0];
-    dst_indices_entry[idx_0] = pos_0 + offset;
-    const auto idx_1 = tid + kThreadsPerBlock;
-    const auto pos_1 = s_indices[idx_1];
-    dst_indices_entry[idx_1] = pos_1 + offset;
+  int index[bitonic_sort_t::kValueSize];
+  fast_topk_cuda_tl(score, index, row_start, length);
+#pragma unroll
+  for (int i = 0; i < bitonic_sort_t::kValueSize; ++i) {
+    dst_indices_entry[threadIdx.x * bitonic_sort_t::kValueSize + i] = index[i] != -1 ? index[i] + offset : -1;
```
DarkSharpness, line 262: same questions as above:

1. why not `naive_topk_cuda`
2. Is this global memory access coalesced?

DarkSharpness:
Thanks for this PR. Do you have any performance results on hopper/blackwell GPUs (e.g. H100, H200, B200) with different problem inpot (prefill for large BS input, decode for small BS input, page-table transform)?

BTW, you may implement this in sglang JIT kernel (#17035) if you're interested. JIT kernel can be integrated and tested faster than  AOT kernels.

hnyls2002:
Hi @fan-wenjie, thanks for exploring the bitonic-sort approach. The shared-memory/occupancy concern in topk.cu was since addressed on main by #17747 (128KB -> 32KB smem in the radix kernel), and the requested hopper/blackwell benchmark results never materialized, so this branch now conflicts with the rewritten kernel. Closing for now; please reopen with updated numbers if I've missed something.

PR #16909, opened 2026-01-11
[HiCache] Support host memory size smaller than device memory size

stmatengss:
## Motivation

To support host memory size can be smaller than device memory size:
1. In the evict method, adding a check for the return value of write_backup. If write_backup fails (e.g., due to host OOM), the code now gracefully falls back to _evict_regular, which evicts the node from the device without attempting to back it up.
2. Modifying evict_host to allow evicting host memory even for nodes that are still resident on the device (x.evicted is False). This is crucial when the host cache is smaller than the device cache, as we need to reclaim host memory occupied by redundant copies to make space for new backups.

## Checklist

## Review Process

4. Get approvals from [CODEOWNERS](https://github.com/sgl-project/sglang/blob/main/.github/CODEOWNERS) and other reviewers.
5. Trigger CI tests with [comments](https://docs.sglang.io/developer_guide/contribution_guide.html#how-to-trigger-ci-tests) or contact authorized users to do so.
6. After green CI and required approvals, ask Merge Oncalls to merge.


python/sglang/srt/mem_cache/hiradix_cache.py
```diff
@@ -397,23 +401,25 @@ def evict_host(self, num_tokens: int):
             _priority, x = heapq.heappop(eviction_heap)
             if x == self.root_node:
                 break
-            # only evict the host value of evicted nodes
-            if not x.evicted:
-                continue
-
+
```
stmatengss:
```suggestion

```

stmatengss (commented):
## Review: [HiCache] Support host memory size smaller than device memory size

### Key Issue: `evict_host` new logic for non-evicted nodes is unreachable

`evict_host` builds its heap from `self.evictable_host_leaves` (line 844). But `_update_host_leaf_status` (line 759-772) only adds nodes where `node.evicted is True` — non-evicted nodes are excluded by the early return at line 760:

```python
def _update_host_leaf_status(self, node: TreeNode):
    if not node.evicted or node.lock_ref > 0:
        if node in self.evictable_host_leaves:
            self.evictable_host_leaves.remove(node)
        return
```

This means backed-up nodes still on device will **never** enter the eviction heap. The new branches in `evict_host` that handle `x.evicted is False` (keeping the node in the tree, only clearing `host_value`) are dead code.

For the PR's goal to work — freeing redundant host copies of on-device nodes when host < device — `_update_host_leaf_status` needs to be updated to also consider non-evicted nodes that have `host_value is not None`.

### Minor items

1. **Commented-out assertion in `memory_pool_host.py`**: If the constraint is intentionally lifted, delete the assertion entirely rather than commenting it out.

2. **Magic number `0.5` in `prefetch_capacity_limit` fallback** (`cache_controller.py:456`): Worth a brief comment explaining the rationale for reserving half of host memory for prefetch when host <= device.

3. **Tree invariant is subtle but correct**: After `evict_host`, a node reachable in the tree with `evicted=True` will always have `host_value != None`, because `evict_host` removes evicted nodes from the tree when clearing their `host_value`. This invariant protects `match_prefix_return_kv` (line 1156-1157) from hitting `len(None)`. A comment documenting this invariant would help future readers.

python/sglang/srt/mem_cache/memory_pool_host.py
```diff
@@ -158,9 +158,9 @@ def __init__(
         self.start_layer = device_pool.start_layer
         self.end_layer = device_pool.end_layer
 
-        assert (
-            self.size > device_pool.size
-        ), "The host memory should be larger than the device memory with the current protocol"
+        # assert (
+        #     self.size > device_pool.size
+        # ), "The host memory should be larger than the device memory with the current protocol"
```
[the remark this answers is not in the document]
xiezhq-hermann: let's only skip this check when write back is selected, or when user specified smaller CPU memory size, fall back to write back policy

python/sglang/srt/mem_cache/hiradix_cache.py
```diff
@@ -977,22 +981,28 @@ def evict_host(self, num_tokens: int):
...
+            if x.host_value is None:
+                continue
+
             # Block deleted entirely (GPU already evicted, now CPU freed) --
             # emit BlockRemoved so the router removes this block from its index.
             self._record_remove_event(x)
             num_evicted += self.cache_controller.evict_host(x.host_value)
+            x.host_value = None
+
+            if x.evicted:
```
xiezhq-hermann, line 993: does this mean we can't evict the host if the data is present on GPU?

python/sglang/srt/mem_cache/memory_pool_host.py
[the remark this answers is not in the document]
stmatengss: Sure, we need a check here

python/sglang/srt/mem_cache/hiradix_cache.py
stmatengss: Data in L2 should be consistent with the GPU.

PR #16919, opened 2026-01-12
[diffusion] Skip negative prompt encoding when guidance_scale <= 1.0 or negative_prompt is None

zhuyuhua-v:
## Motivation
classifier-free guidance is enabled when both `guidance_scale > 1.0` and `negative_prompt is not None`.  
Consequently, the negative prompt needs to be encoded **only under this condition**.

When either `guidance_scale <= 1.0` or `negative_prompt is None`, the unconditional noise prediction (`noise_pred_uncond`) is not used, making the negative prompt encoding redundant. Skipping it in these cases avoids unnecessary computation in the text encoder and improves inference efficiency.

## Modifications
- Skip the negative prompt text encoding if `guidance_scale <= 1.0` or `negative_prompt is None`.

## Accuracy Tests
test cmd:
```bash

def main():
        model_path="/mnt/models/Qwen/Qwen-Image-Edit",
        num_gpus=2,
        ulysses_degree=2,
        tp_size=1,
        enable_torch_compile=True,
        dit_cpu_offload=False,
        text_encoder_cpu_offload=False,
        vae_cpu_offload=False,
        image_encoder_cpu_offload=False,
        image_encoder_precision="bf16",
        vae_precision="bf16",
    )

    generator.generate(
            sampling_params_kwargs=dict(
                prompt="make the clothes to red",
                image_path="input/768x1024.png",
                output_path="qwen_image_edit/",
                height=1024,
                width=768,
                num_inference_steps=8,
                guidance_scale=1, 
            )
        )

if __name__ == '__main__':
    main()
```
input image:
image

prompt: `make the clothes to red`

output with this pr:
image

output without this pr:
image

## Benchmarking and Profiling
performance without this pr on MI308:
```bash
[01-12 02:38:53] [ImageEncodingStage] finished in 0.7408 seconds
[01-12 02:39:05] Pixel data generated successfully in 12.66 seconds
```

performance with this pr on MI308:
```bash
[01-12 02:33:32] [ImageEncodingStage] finished in 0.3976 seconds
[01-12 02:33:44] Pixel data generated successfully in 12.33 seconds
```


python/sglang/multimodal_gen/runtime/pipelines_core/stages/image_encoding.py
```diff
@@ -150,20 +150,22 @@ def forward(
...
                     output_hidden_states=True,
                 )
-                neg_outputs = self.text_encoder(
-                    input_ids=neg_image_inputs.input_ids,
-                    attention_mask=neg_image_inputs.attention_mask,
-                    pixel_values=neg_image_inputs.pixel_values,
-                    image_grid_thw=neg_image_inputs.image_grid_thw,
-                    output_hidden_states=True,
-                )
+                if batch.guidance_scale > 1.0 and batch.negative_prompt is not None:
```
mickqian: could we generalize this logic to something like `batch::should_do_cfg`, and apply that logic to denoising stage as well?

mickqian: actually, we have a `do_classifier_free_guidance` already

zhuyuhua-v: Thanks a lot for the notification! I've updated these changes to use `do_classifier_free_guidance` for judgment.

python/sglang/multimodal_gen/runtime/pipelines_core/stages/image_encoding.py
mickqian: in this case, should be skip these too?

mickqian: also here

zhuyuhua-v: updated

zhuyuhua-v: updated

mickqian (approved):

PR #16944, opened 2026-01-12
[feat] Add ATOM model impl backend

zejunchen-zejun:
- [x] RFC for ATOM model impl backend

### DIFF
--- a/test/registered/models/test_atom_models.py
+++ b/test/registered/models/test_atom_models.py
@@ -1,0 +1,57 @@
+# ATOM Plugin tests
+
+import unittest
+from types import SimpleNamespace 
+
+
+from sglang.srt.utils import kill_process_tree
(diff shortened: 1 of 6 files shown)


test/registered/models/test_atom_models.py
```diff
@@ -0,0 +1,57 @@
...
+from sglang.test.test_utils import (
+    DEFAULT_TIMEOUT_FOR_SERVER_LAUNCH,
+    DEFAULT_URL_FOR_TEST,
+    CustomTestCase,
+    popen_launch_server,
+)
+
+register_amd_ci(est_time=600, suite="stage-b-test-small-1-gpu-amd")
+
+class TestQwen3MoeFp8(CustomTestCase):
```
zejunchen-zejun, line 19: only test interface in SGLang
model QA will be done on ATOM side

PR #16946, opened 2026-01-12
[HiCache] Prevent KV cache data loss when radix tree node is split b…

chenkaiyue:
…efore storage write

Fix KV cache data loss when radix tree node is split before storage write

## Motivation

When using HiCache with Mooncake Store as the storage backend, I discovered a critical bug that causes significant KV cache data loss during the write-through process.

### Problem Discovery

While testing the HiCache + Mooncake Store integration, I sent **two identical requests** to an SGLang instance and observed unexpected behavior:

**First Request:**
- Created a new radix tree node (`node_id=3`) with **375 tokens**
- The node was added to `ongoing_write_through` queue via `write_backup()` → `_inc_hit_count()`
- D2H (Device-to-Host) copy started asynchronously

**Second Request (same content):**
- `match_prefix()` found that 374 tokens matched the existing node
- Since `prefix_len (374) < len(node.key) (375)`, `_split_node()` was triggered
- The original node was split into:
  - `node_id=4` (new parent): **374 tokens** (key, hash_value, host_value all transferred).  ***node_id=4 will never be written to the cache because it inherits the backuped=True attribute from node_id=3 and will not be added to the ongoing_write_through subsequently.***
  - `node_id=3` (original): only **1 token** remaining

**When `writing_check()` executed:**
- Retrieved `node_id=3` from `ongoing_write_through` queue
- Called `write_backup_storage(node_id=3)`
- But at this point, `node_id=3` only had **1 token** instead of the original 375!
- Result: Only **1 key** was written to Mooncake instead of **375 keys**

### Root Cause Analysis

The issue is that `ongoing_write_through[node.id] = node` stores a **reference** to the node object, not a snapshot of its data. When the node is later split by `_split_node()`:

```
Before split:
  ongoing_write_through[3] ────→ node_id=3 (375 tokens)

After split:
  ongoing_write_through[3] ────→ node_id=3 (1 token)  ← Reference points to modified node!
                                       ↑
                                 Data was moved to node_id=4
```

This causes **massive data loss** as only the split remainder gets written to storage.

### Solution: Save Data Snapshot Instead of Node Reference

The fix saves a **snapshot** of the node's critical data (`key`, `hash_value`, `host_value`) when adding to the queue, rather than just storing the node reference.

#### 1. Modified `write_backup()` to save snapshot:

```python
# Before (stores reference - problematic)
self.ongoing_write_through[node.id] = node

# After (stores snapshot - safe)
self.ongoing_write_through[node.id] = {
    'node': node,
    'key': RadixKey(token_ids=node.key.token_ids.clone()),
    'hash_value': list(node.hash_value) if node.hash_value else None,
    'host_value': host_indices,
}
```

#### 2. Modified `writing_check()` to use snapshot data:

```python
snapshot = self.ongoing_write_through.pop(ack_id)
node = snapshot['node']
self.dec_lock_ref(node)
if self.enable_storage:
    # Temporarily set snapshot data for write_backup_storage
    current_key, current_hash_value, current_host_value = (
        node.key, node.hash_value, node.host_value
    )
    
    node.key = snapshot['key']
    node.hash_value = snapshot['hash_value']
    node.host_value = snapshot['host_value']
    
    self.write_backup_storage(node)
    
    # Restore current values to keep tree structure intact
    node.key = current_key
    node.hash_value = current_hash_value
    node.host_value = current_host_value
```

### Why This Approach?

1. **No new methods required**: Reuses existing `write_backup_storage()` by temporarily setting node attributes

2. **Tree structure preserved**: After `write_backup_storage()` returns, the node's attributes are restored to their current (post-split) values, maintaining tree integrity

3. **Python reference semantics**: Modifying `node.key = snapshot['key']` only changes which object `node.key` points to; `BackupOperation` created inside `write_backup_storage()` still holds the correct reference to the snapshot data

### Concurrency and Frequent Splitting Considerations

This solution has been designed with high-concurrency and frequent radix tree splitting scenarios in mind:

| Scenario | Status | Explanation |
|----------|--------|-------------|
| **Thread Safety** | ✅ Safe | All radix cache operations run in the scheduler's main thread; no concurrent access to `ongoing_write_through` |
| **Multiple Splits** | ✅ Safe | Snapshot captures data at queue-add time; any number of subsequent splits don't affect the saved snapshot |
| **TP Synchronization** | ✅ Safe | All TP ranks have identical tree structure and snapshot data |
| **Memory Overhead** | ✅ Minimal | ~25KB per node (token_ids clone + hash strings), released after write completion |
| **write_back Mode** | ✅ Compatible | Updated to pop snapshot dict instead of delete node reference |

### Alternative Approaches Considered

| Approach | Rejected Reason |
|----------|-----------------|
| Full node clone | Would break tree structure (parent/children references) and lock_ref management |
| Handle in `_split_node()` | Complex nested handling required for chains of splits; race conditions with ongoing D2H |
| Write lock | Performance impact unacceptable for high-throughput scenarios |

---
### CI States

Latest PR Test (Base): :white_check_mark: [Run #26444661404](https://github.com/sgl-project/sglang/actions/runs/26444661404)
Latest PR Test (Extra): :x: [Run #26444661225](https://github.com/sgl-project/sglang/actions/runs/26444661225)


hzh0425:
This bug should be reasonable; but it might require some discussion on how to resolve it more concisely  cc @xiezhq-hermann

xiezhq-hermann:
thank you for reporting, we had similar problem with L1+L2 too, but fixed it with an addtion reference list. I will get back to you on this and I think there might be easier solutions.

python/sglang/srt/mem_cache/hiradix_cache.py
```diff
@@ -302,10 +309,27 @@ def writing_check(self, write_back=False):
...
+                    current_key = node.key
+                    current_hash_value = node.hash_value
+                    current_host_value = node.host_value
+
+                    try:
+                        node.key = snapshot["key"]
+                        node.hash_value = snapshot["hash_value"]
+                        node.host_value = snapshot["host_value"]
+
+                        self.write_backup_storage(node)
```
xiezhq-hermann: I would suggest to directly change the `write_backup_storage` function to accommodate the snapshot rather than fitting a node for it.

chenkaiyue: OK~ I will fix it

xiezhq-hermann:
Hi @chenkaiyue can you help test whether this problem still persists on the latest main? If so can we resolve conflict and get this merged sooner, thanks a lot!

chenkaiyue:
> Hi @chenkaiyue can you help test whether this problem still persists on the latest main? If so can we resolve conflict and get this merged sooner, thanks a lot!

The bug still exists on the latest main (checked at 2a327f087). I have updated the code.

xiezhq-hermann:
I would suggest a lighter fix: record `(node_id, backup_len)` at enqueue, and at ack time walk node.parent accumulating `len(parent.key)` until it covers `backup_len`, then concat `key/hash_value/host_value` along the chain. `_split_node` slices into clones pointing at the same host slots, so concat recovers the original exactly;

Wins: no potentially expensive node snapshot, no mutate-then-restore dance, fast path (len(node.key) == backup_len) stays allocation-free. Only cost is ack-time concat on the split path, which should be rare and low in general. What do you think @chenkaiyue  @hzh0425 

Happy to take a shot at this if it sounds reasonable.

chenkaiyue:
> I would suggest a lighter fix: record `(node_id, backup_len)` at enqueue, and at ack time walk node.parent accumulating `len(parent.key)` until it covers `backup_len`, then concat `key/hash_value/host_value` along the chain. `_split_node` slices into clones pointing at the same host slots, so concat recovers the original exactly;
> 
> Wins: no potentially expensive node snapshot, no mutate-then-restore dance, fast path (len(node.key) == backup_len) stays allocation-free. Only cost is ack-time concat on the split path, which should be rare and low in general. What do you think @chenkaiyue @hzh0425
> 
> Happy to take a shot at this if it sounds reasonable.
 
I think your approach works well. I've made changes following this idea and submitted a new PR. @xiezhq-hermann

chenkaiyue:
## End-to-end verification

Verified this commit on a sglang **0.5.11** deployment with Mooncake (RDMA) as the storage backend, using the same repro method as the PR description — two identical long-prompt requests sent back-to-back.

### sglang server-side log (`log_requests=true, log_requests_level=2`)

A1 — fresh request, brings ~222 new KV tokens into the cache and enqueues the node for write-through:

```
[2026-05-13 12:27:25 TP0] Req Time Stats(rid=2f35db53f29f4eb98c49077ea88e4dd9, input len=253, output len=4, type=unified): queue_duration=0.71ms, forward_duration=109.62ms, start_time=14026922.615
Finish: obj=GenerateReqInput(rid='2f35db53f29f4eb98c49077ea88e4dd9', ..., input_ids=[...253 ids...], ...), out={'text': 'The user asks:', 'output_ids': [758, 3100, 20886, 58], 'meta_info': {'id': '2f35db53f29f4eb98c49077ea88e4dd9', 'finish_reason': {'type': 'length', 'length': 4}, 'prompt_tokens': 253, 'completion_tokens': 4, 'cached_tokens': 31, 'cached_tokens_details': {'device': 31, 'host': 0, 'storage': 0, 'storage_backend': 'MooncakeStore'}, ...}}
```

A2 — same prompt, sent immediately after A1 returned:

```
[2026-05-13 12:27:26 TP0] Req Time Stats(rid=8b6ae0b5364240de9fa46762695e8f48, input len=253, output len=4, type=unified): queue_duration=0.55ms, forward_duration=110.18ms, start_time=14026922.734
Finish: obj=GenerateReqInput(rid='8b6ae0b5364240de9fa46762695e8f48', ..., input_ids=[...253 ids...], ...), out={'text': 'The user asks:', 'output_ids': [758, 3100, 20886, 58], 'meta_info': {'id': '8b6ae0b5364240de9fa46762695e8f48', 'finish_reason': {'type': 'length', 'length': 4}, 'prompt_tokens': 253, 'completion_tokens': 4, 'cached_tokens': 252, 'cached_tokens_details': {'device': 252, 'host': 0, 'storage': 0, 'storage_backend': 'MooncakeStore'}, ...}}
```

Note `cached_tokens_details: device=252, host=0, storage=0` on A2 — A1's D2H had not finished yet, the entire 252-token hit came from L1 (GPU). This is exactly the in-flight window the bug describes: A2 arrived in time to slice A1's still-enqueued node.

### Mooncake master metrics (`curl /metrics`) — same instance, before and after the pair

Before (just before A1):
```
master_key_count 6244
master_value_size_bytes_count 6253
```

After (8 s after A2 returned):
```
master_key_count 8052
master_value_size_bytes_count 8061
master_allocated_bytes 255492096
```

Delta: **+1808 keys / +1808 values written to storage** across this single split-triggering pair.

With the unfixed code, A1's split node would have written only its post-split tail (1 token) to storage and A2 would not enqueue anything else through the storage path, so the delta would be in the single-digit-to-low-tens range. The observed +1808 is two orders of magnitude larger and matches the entire ~222-token pre-split range of A1 being recovered through the parent chain and written to storage — i.e. the data-loss path is closed.

chenkaiyue:
## 13.5h soak test under the same Mooncake deployment

Beyond the targeted repro above, also stress-tested the patched build with `bench_multiturn.py` in a loop overnight. The intent was to make sure the new walk-and-concat path doesn't break under sustained mixed-prefix traffic — split events happen continuously when many concurrent clients share/extend prefixes.

### Loop driver

A wrapper script ran the following inside the sglang pod, looping for 13.5h. Each round, after the bench finishes, it hits `/flush_cache` (with retry) and `/clear_hicache_storage_backend` so the next round starts with all three tiers empty:

```bash
while true; do
  python3 /sgl-workspace/sglang/benchmark/hicache/bench_multiturn.py \
    --host localhost --port 8001 \
    --model-path /dev/shm/MiniMax-M2.7 \
    --dataset-path .../ShareGPT_V3_unfiltered_cleaned_split.json \
    --disable-random-sample --output-length 1 --request-length 2048 \
    --num-clients 80 --num-rounds 10 --max-parallel 16 --request-rate 32 \
    --ready-queue-policy random --disable-auto-run --enable-round-barrier \
    --log-file round${N}.jsonl --tag "loop-r${N}"

  for i in 1..10; do curl -X POST .../flush_cache && break; sleep 3; done
  curl -X POST .../clear_hicache_storage_backend
  curl -X POST .../flush_cache
done
```

### Aggregate result over 13.5h

```
rounds attempted          : 103
bench exit=0              : 102 / 102  (last round was mid-flight when stopped)
/flush_cache final 200    : 100 / 102  (47 transient 400s recovered on retry,
                                        caused by pending hicache backups)
/clear_hicache_storage_backend 200 : 102 / 102
sglang /health after run  : HTTP 200
```

No sglang crash, no scheduler stack trace, no `master_put_start_failures` on the Mooncake side.

### Cache-tier traffic observed (sglang `/metrics`, cumulative over the loop)

```
sglang:cached_tokens_total{cache_source="device"}              208,774,847   # L1 GPU
sglang:cached_tokens_total{cache_source="host"}                206,258,994   # L2 host
sglang:cached_tokens_total{cache_source="storage_MooncakeStore"}  3,865,110  # L3
sglang:hicache_host_used_tokens                                    825,646
sglang:hicache_host_total_tokens                                   826,662   # L2 99.88% used
sglang:backuped_tokens_total{tp_rank="0"}                      ~3.00e+08     # writes to mooncake
sglang:backup_pgs_count{tp_rank="0"}                            2,316,018    # batch_put calls
sglang:prefetched_tokens_total{tp_rank="0"}                     9,491,333    # reads from mooncake
sglang:prefetch_pgs_count{tp_rank="0"}                             75,138    # prefetch ops
```

All three tiers were live: 208M+206M+3.9M ≈ 419M tokens served from cache across the loop, with the storage tier actively both written (2.3M batch_put / ~1.2B tokens total across 4 TP ranks) and read (75k prefetch ops, 9.5M tokens pulled back via RDMA).

### Mooncake master side at the end

```
master_allocated_bytes          96,990,077,952   # 96.4% of 100 GiB
master_total_capacity_bytes    107,374,182,400
master_key_count                 3,055,383
master_value_size_bytes_count    2,293,184,652   # ~2.29 B batch_set ops over the run
master_put_start_failures_total            0
master_active_clients                      4
```

Mooncake master reports zero failures over the whole run; the `BatchPut failed ... insufficient space` warnings that occasionally show up in the sglang log are client-side ENOSPC when the 100 GiB segment is nearly full — they're cleared by the loop's per-round `/clear_hicache_storage_backend`.

The data-loss path this PR closes is the bug's worst-case footprint per request. The fact that 4M+ token-level L3 hits across 103 rounds all came back as the expected payload (no corrupted text, no scheduler aborts) is the cumulative version of that guarantee under real concurrent traffic.

chenkaiyue:
@xiezhq-hermann   please review~~

python/sglang/srt/mem_cache/hiradix_cache.py
```diff
@@ -715,23 +719,70 @@ def write_backup(self, node: TreeNode, write_back=False) -> int:
...
+            # segment's last; drop the overlap on every segment except the first.
+            token_ids = list(chain[0].key.token_ids)
+            for n in chain[1:]:
+                token_ids.extend(n.key.token_ids[1:])
+        else:
+            token_ids = [t for n in chain for t in n.key.token_ids]
+        key = RadixKey(token_ids, top.key.extra_key, top.key.is_bigram)
+
+        hash_value = (
+            [h for n in chain for h in n.hash_value]
```
xiezhq-hermann: consider use vectorized operation for efficiency

xiezhq-hermann (approved):

python/sglang/srt/mem_cache/hiradix_cache.py
```diff
@@ -725,23 +729,70 @@ def write_backup(self, node: TreeNode, write_back=False) -> int:
 
         return len(host_indices)
 
-    def write_backup_storage(self, node: TreeNode):
+    def write_backup_storage(self, node: TreeNode, backup_len: Optional[int] = None):
+        # When `backup_len` is given and the node has been split since
```
xiezhq-hermann: let's make the comments more concise as well, claude code tends to generate verbose comments which can be annoying sometimes

chenkaiyue:
@xiezhq-hermann  I have finished the revisions as you suggested.

chenkaiyue:
Can you help merge the PR? @xiezhq-hermann

PR #16961, opened 2026-01-12
[DeepSeek v3.2] Opt MTP decode cuda batch sizes and nsa implementation

xu-yfei:
## Motivation
1、`draft_extend` and `target_verify` use `nsa_decode_backend` as backend implementation instead of `nsa_prefill_backend`
In the MTP scenario, the NSA attention implementation selected for `prefill`, `draft_extend`, and `target_verify` is all `nsa_prefill_backend`, which is typically set to `flashmla_sparse`. For `draft_extend` and `target_verify` scenarios with a small number of tokens, we observed that `fa3` delivers better performance. However, when `flashmla_sparse` is chosen for `prefill,` there is no interface to additionally specify the NSA attention implementation for `draft_extend` and `target_verify`. This PR modifies the backend implementation of `draft_extend` and `target_verify` to `nsa_decode_backend` —in practice, their CUDA Graph initialization and replay already use `nsa_decode_impl`.

num_draft_tokens=4, TPOT:
| batch size | flashmla_sparse | fa3   |
|------------|-----------------|-------|
| 1          | 11.87           | 9.44  |
| 2          | 13.34           | 13.08 |
| 8          | 15.25           | 15.87 |
| 16         | 21.25           | 21.50|
| 32         | 32.94           | 33.79 |

2、Opt cuda graph batch sizes when MTP:
In the Prefill CP scenario, `require_gathered_buffer` is set to True. Assuming the number of MTP num_draft_tokens is `4`, the CUDA Graph tokens in the MTP scenario are calculated as follows:
`get_batch_sizes_to_capture` returns batch sizes that are multiples of `attn_tp_size`, i.e., `[8, 16, 24, 32, ....]`.
In `capture_one_batch_size`, `num_tokens = bs * self.num_tokens_per_bs`, meaning the actual number of tokens is `[32, 64, 96, 128, ...]`.
When `bs=1`, the number of tokens is `4` but needs to be padded to `32`, resulting in significant performance overhead. In reality, `num_tokens` only needs to be a multiple of `attn_tp_size`.
Modification: `get_batch_sizes_to_capture` is updated to return batch sizes where `batch size * num_draft_tokens` is divisible by `attn_tp_size`, i.e., `[2, 4, 6, 8, ....]`.

## Accuracy Tests

```
MODEL_PATH=/home/models/DeepSeek-V3.2/

python3 -m sglang.launch_server --model-path $MODEL_PATH --trust-remote-code \
--port 8000 --host 0.0.0.0 --attention-backend  nsa \
--enable-metrics --mem-fraction-static 0.8 --max-running-requests 128 --enable-cache-report --page-size 64 \
--tp-size 8 --moe-dense-tp-size 1 \
--reasoning-parser deepseek-v3 \
--moe-dense-tp-size 1 \
--nsa-decode-backend fa3 \
--enable-nsa-prefill-context-parallel \
--nsa-prefill-cp-mode round-robin-split \
--speculative-num-steps 3 --speculative-eagle-topk 1 --speculative-num-draft-tokens 4
```

```
# gsm8k
Accuracy: 0.948
Invalid: 0.000
Latency: 192.326 s
Output throughput: 630.665 token/s
```

## Checklist

## Review Process

### DIFF
--- a/python/sglang/srt/model_executor/cuda_graph_runner.py
+++ b/python/sglang/srt/model_executor/cuda_graph_runner.py
@@ -186,7 +186,7 @@ def set_torch_compile_config():
     monkey_patch_torch_compile()
 
 
-def get_batch_sizes_to_capture(model_runner: ModelRunner):
+def get_batch_sizes_to_capture(model_runner: ModelRunner, num_tokens_per_bs=1):
     server_args = model_runner.server_args
     capture_bs = server_args.cuda_graph_bs
 
@@ -203,7 +203,7 @@ def get_batch_sizes_to_capture(model_runner: ModelRunner):
     if require_gathered_buffer(server_args):
         mul_base *= get_attention_tp_size()
 
-    capture_bs = [bs for bs in capture_bs if bs % mul_base == 0]
+    capture_bs = [bs for bs in capture_bs if bs * num_tokens_per_bs % mul_base == 0]
 
     capture_bs = [bs for bs in capture_bs if bs <= model_runner.req_to_token_pool.size]
     capture_bs = list(sorted(set(capture_bs)))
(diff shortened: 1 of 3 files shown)


python/sglang/srt/model_executor/cuda_graph_runner.py
```diff
@@ -203,7 +203,7 @@ def get_batch_sizes_to_capture(model_runner: ModelRunner):
     if require_gathered_buffer(server_args):
         mul_base *= get_attention_tp_size()
 
-    capture_bs = [bs for bs in capture_bs if bs % mul_base == 0]
+    capture_bs = [bs for bs in capture_bs if bs * num_tokens_per_bs % mul_base == 0]
```
Fridge003, line 208: Maybe add a comment on why we take `bs * num_tokens_per_bs % mul_base == 0` here as condition?

xu-yfei: done~

PR #16996, opened 2026-01-13
feat: Support 'use_audio_in_video' option for qwen3omnimoe model

srLi24:
## Motivation

This PR adds support for the `use_audio_in_video` configuration parameter in the `Qwen3-Omni-Moe` model.

## Modifications

- Passed this parameter to the input of 'Engine' class and openai request structure.
- Support load audio from the same video file source when using Qwen3-Omni-Moe model and open this option.
- Fix 'fps' calculation in qwenvl processor.

## Usage Example

**1. OpenAI Compatible Client**
Pass `use_audio_in_video` via `extra_body`:
```python
    model="Qwen/Qwen3-Omni-30B-A3B-Instruct",
    messages=messages,
    temperature=0,
    max_tokens=1024,
    extra_body={
        "use_audio_in_video": True,
    }
)
```

**2. Offline Model Engine**
Pass `use_audio_in_video` via `generate`:
```python
outputs = llm.generate(
    prompt=prompts, 
    video_data=["codes/video_audio.mp4"],
    use_audio_in_video=True,
)
```

## Accuracy Tests

| Modality | vLLM | SGLang (This PR) |
| :--- | :--- | :--- |
| **Audio + Video** | 74.35 | **74.60** |
| **Video Only** | 58.23 | **59.31** |

## Checklist

- [] Update documentation according to [Write documentations](https://docs.sglang.io/developer_guide/contribution_guide.html#write-documentations).
- [] Follow the SGLang code style [guidance](https://docs.sglang.io/developer_guide/contribution_guide.html#code-style-guidance).

## Review Process


JustinTong0323:
Could we add `use_audio_in_video` to the sampling parameters? Then we could "pop" it in the processor, which would be less invasive since we wouldn't need to add it into the existing protocol.

srLi24:
> Could we add `use_audio_in_video` to the sampling parameters? Then we could "pop" it in the processor, which would be less invasive since we wouldn't need to add it into the existing protocol.

Hi, we re-write the logic to support 'use_audio_in_video' through sampling_params and maintain the benchmark performance with the previous commit.

**1. OpenAI Compatible Client**
Pass `use_audio_in_video` via `extra_body`:
```python
    model="Qwen/Qwen3-Omni-30B-A3B-Instruct",
    messages=messages,
    temperature=0,
    max_tokens=1024,
    extra_body={
        "use_audio_in_video": True,
    }
)
```

**2. Offline Engine**
```python
sampling_params = {
        "temperature": 0.0, 
        "top_p": 1.0,
        "max_new_tokens": 2048,
        "use_audio_in_video": True,
    }
```

srLi24:
I've updated the dependency requirements to include `av`. This should resolve the `ModuleNotFoundError` in the CI pipeline.

li126com:
Hello, I have a question regarding the training scheme for Qwen3-Omni with audio-video data.

I noticed that in your PR it mainly addressed the use_audio_in_video issue, which is indeed a major pain point. However, my question is: under the audio-in-video training setting, how should we correctly construct and pass input_ids to SGLang?

To put it simply, in the current multimodal data pipeline, when we pass data to SGLang, the input_ids are constructed as:
the result of tokenizer.encode(text_prompt)
plus the serialized raw multimodal data sequence

Then inside SGLang, input_ids are decoded to obtain input_text. 
''''''
# sglang/srt/multimodal/processors/base_processor.py
def load_mm_data:
    ...
    prompt = self._processor.tokenizer.decode(prompt)
    ...
''''''
Finally, this input_text, together with the deserialized multimodal data, is fed into the Qwen3-Omni processor to produce the processor outputs.
''''''
# sglang/srt/multimodal/processors/base_processor.py
def process_and_combine_mm_data:
    ...
    if raw_images or raw_audios or raw_videos:
        collected_items, input_ids, ret = self._process_and_collect_mm_items(
            input_text=base_output.input_text,
            images=raw_images,
            audios=raw_audios,
            videos=raw_videos,
            **kwargs,
        )
    ...
''''''

My question is: when use_audio_in_video=True, how should we properly construct the input_ids for SGLang?

The issue is that tokenizer.encode() itself does not accept the use_audio_in_video argument. Therefore, the returned input_ids will only contain video tokens but not contain any audio-related tokens since actually we only have video data (which contains audio). This can lead to a mismatch during SGLang’s internal consistency checks between tokens and multimodal data. see `def submit_data_loading_tasks`

However, if we forcibly append audio tokens to the input_ids before sending them to SGLang, would that cause problems?

My understanding is that the final input_text fed into the processor inside SGLang should contain only the video placeholder. Then, with use_audio_in_video=True, the processor should internally handle audio extraction from the video and produce the expected outputs.

But since input_text is obtained by decoding the input input_ids, if we manually append audio tokens to the input input_ids, wouldn’t the resulting input_text that decoded from input_ids deviate from what the processor expects?

I would really appreciate your clarification on this.

PR #17009, opened 2026-01-13
feat: add support for bitwise weight verification via checksums

Ratish1:
## Motivation

   This PR adds a mechanism to verify that the model weights loaded in SGLang are bit-for-bit identical. For more info, see here  radixark/miles/pull/415.

## Modifications

- Protocol: Added compare_checksum to WeightChecker. It gathers local TP shards, reconstructs the full parameter (supporting both Column and Row Parallelism), and verifies the SHA256 hash against provided ground truth.
- Precision: Forces a cast to bfloat16 during hashing to ensure bit-perfect alignment.
- API: Updated CheckWeightsReqInput to accept an optional checksums payload and propagated it from the API to the ModelRunner.

## Checklist

## Review Process


zhaochenyang20 (commented):
Great PR for checking the bitwise verification. I have several comments:

1. In the actual weights update logic in RL frameworks, try to overlap the checksum and weight refit. The idea is that `check_sum` computation on the CPU is rather slower than moving data from the GPU to the CPU by PCIe. If we are currently using this:

```python
# blocking check sum
for name, tensor in tensors:
    t_cpu = tensor.cpu() 
    hash = hashlib.sha256(t_cpu).hexdigest()
```

The ideal Version is that we should turn it into non-blocking:

```python

import concurrent.futures

executor = concurrent.futures.ThreadPoolExecutor(max_workers=1)
future = None

for tensor in tensors:
    t_cpu = tensor.detach().cpu().contiguous()
    if future:
        check_hash(future.result())
    future = executor.submit(compute_sha256, t_cpu)
if future:
    check_hash(future.result())
```

2. But, for the first step, favor simplicity. No need to make the non-blocking version of `check_sum` right now; it should be okay. To this point, we shall not turn the `check_sum` default on. Let the users decide whether to `check_sum` in each weight update round.

3. I add some comments in `python/sglang/srt/utils/weight_checker.py`. I think we should add dedicated comments to users on `@app.post("/weights_checker")` for RL users.

python/sglang/srt/utils/weight_checker.py
```diff
@@ -16,18 +17,109 @@ def __init__(self, model_runner):
...
+
+        errors = []
+        matched_count = 0
+
+        for name in sorted(actual_state.keys()):
+            if name in expected_checksums:
+                param = actual_state[name]
+                expected_hash = expected_checksums[name]
+
+                data = param.data.to(torch.bfloat16)
```
zhaochenyang20: 1. Why should we copy the data? It doubled the size of GPU usage.
2. Why should we move it to bf16? If we are doing FP8 training, we want to compare everything in FP8.

python/sglang/srt/utils/weight_checker.py
```diff
@@ -16,18 +17,109 @@ def __init__(self, model_runner):
...
+                                full_p1.detach()
+                                .cpu()
+                                .contiguous()
+                                .view(torch.uint8)
+                                .numpy()
+                            ).hexdigest()
+                            == expected_hash
+                        ):
+                            matched_count += 1
+                            continue
```
zhaochenyang20: The code tries Dim 0, calculates hash, and if it fails, tries Dim 1. This is inefficient. For Row Parallel layers, you are performing the reconstruction and hashing twice. This wastes CPU/GPU bandwidth.

Please determine the correct sharding dimension based on the parameter name or metadata (Row vs. Column Parallel) and reconstruct only once.

python/sglang/srt/utils/weight_checker.py
```diff
@@ -16,18 +17,109 @@ def __init__(self, model_runner):
...
+                    matched_count += 1
+                    continue
+
+                # STAGE B: Shard Reconstruction (Handles TP layers)
+                if tp_group.world_size > 1:
+                    # Gather shards from all TP ranks
+                    all_shards = [
+                        torch.empty_like(data) for _ in range(tp_group.world_size)
+                    ]
+                    torch.distributed.all_gather(
```
zhaochenyang20: As I mentioned, this is an all_gather in a for loop. It launched some GPU kernels N times. We should use overlap methods to verify each parameter. Not verify all the parameters in a large `expected_checksums`.

In other words, if we can make checksum non-blocking, the passed-in `expected_checksums` should be a one-element dictionary of `{parameter_name, hex_digest}`.

python/sglang/srt/managers/io_struct.py
```diff
@@ -1402,6 +1402,7 @@ class ResumeMemoryOccupationReqOutput(BaseReq):
 @dataclass
 class CheckWeightsReqInput(BaseReq):
     action: str
+    checksums: Optional[list | dict] = None
```
zhaochenyang20: From your codes, I think this is indeed a `{name, hex}`. This type hint is wrong?

Also, please follow this:

https://github.com/sgl-project/sglang/pull/17009/changes#r2692231715

python/sglang/srt/utils/weight_checker.py
```diff
@@ -16,18 +17,109 @@ def __init__(self, model_runner):
...
             raise Exception(f"Unsupported {action=}")
 
+    def _compare_checksum(self, expected_checksums):
+        if expected_checksums is None:
+            return
+
+        # 1. Get raw model state (BF16 for Qwen2.5-3B)
+        actual_state = dict(self._model_state())
+
+        import sglang.srt.distributed.parallel_state as ps
```
zhaochenyang20: Do not import anything here.

zhaochenyang20 (changes requested):

python/sglang/srt/utils/weight_checker.py
```diff
@@ -16,18 +17,109 @@ def __init__(self, model_runner):
...
+        for name in sorted(actual_state.keys()):
+            if name in expected_checksums:
+                param = actual_state[name]
+                expected_hash = expected_checksums[name]
+
+                data = param.data.to(torch.bfloat16)
+
+                # STAGE A: Direct Match (Handles Replicated layers like Norms/Embeds)
+                t_cpu = data.detach().cpu().contiguous()
+                actual_hash = hashlib.sha256(
```
lilei199908: maybe use more fast hash?

python/sglang/srt/utils/weight_checker.py
Ratish1: Thanks for the feedback!

1)Regarding GPU Memory: You are absolutely right. I will refactor the code to process parameters one-by-one and delete the reconstructed tensors. 

2)Regarding BF16 vs FP8: Since the goal of issue [here](https://github.com/radixark/miles/issues/405) is to support 'bf16 training fp8 rollout' verification, we must align the precision to the trainer's source to verify the equality right?. Or do you suggest I should instead quantize the weights on the trainer side before hashing, or is dequantizing one-by-one in SGLang for the check acceptable?. Lmk which way you would prefer for this.

python/sglang/srt/utils/weight_checker.py
Ratish1: Thanks, that is something maybe we could consider. My initial choice was based on avoiding new dependencies. Do you have any suggestions for a faster algorithm or a more standard library that we can use for this PR, if so, then I have to update my miles pr aswell accordingly.

PR #17013, opened 2026-01-13
Feat/support nemotron h mtp

shaharmor98:
## Motivation
This PR adds Multi-Token Prediction (MTP) speculative decoding support for NemotronH hybrid Mamba2/Attention models.

#### New NemotronH MTP Model (nemotron_h_mtp.py)

- Introduced `NemotronHForCausalLMMTP` model class for MTP-based speculative decoding
- Added `NemotronHMTPAttentionDecoderLayer` and `NemotronHMTPMoEDecoderLayer` that extend base `NemotronH` layers
- Implemented `NemotronHMultiTokenPredictor` that processes the hybrid layer pattern

Note: Two decoder layers were added (for both attention and MoE), since future versions might reverse the MTP layers. That is, we assume the first layer has the layer projections, while the last has a layer norm. However, it’s not guaranteed that attention will appear immediately after the layer projections, nor that the MoE layer will appear before the layer norm. Therefore, the current implementation allows future architecture modifications without code changes.

## Accuracy Tests
MTP support has been internally tested on the upcoming Nemotron-Super checkpoint.
Since the model isn’t public yet, we can’t expose internal results.

## Review Process

### DIFF
--- a/python/sglang/srt/model_executor/model_runner.py
+++ b/python/sglang/srt/model_executor/model_runner.py
@@ -1485,13 +1485,13 @@ class ModelRunner(ModelRunnerKVCacheMixin):
     @property
     def mamba2_config(self):
         config = self.model_config.hf_config
-        if isinstance(config, FalconH1Config | NemotronHConfig):
+        if isinstance(config, NemotronHConfig) and self.is_draft_worker:
             # NemotronH MTP draft models have no Mamba layers (pattern like "*E")
             # so they shouldn't use HybridLinearAttnBackend
-            if self.is_draft_worker:
-                pattern = getattr(config, "mtp_hybrid_override_pattern", None)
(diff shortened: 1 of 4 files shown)


python/sglang/srt/model_executor/model_runner.py
```diff
@@ -1486,6 +1486,12 @@ def hybrid_gdn_config(self):
     def mamba2_config(self):
         config = self.model_config.hf_config
         if isinstance(config, FalconH1Config | NemotronHConfig):
```
Fridge003: Will this change also be applied to `FalconH1Config`? If not we need to split these two models

shaharmor98: You're absolutely right, good catch!  
Fixed

Fridge003 (approved):

PR #17026, opened 2026-01-13
feat: Priority-based scheduling optimization (including default priority, preemption toggle, priority-based metrics, etc.)

zhuxinjie-nz:
## Motivation

During the use of priority-based scheduling, we identified several issues:

1. The original priority scheduling logic does not support setting a default priority.

2. The preemption logic introduces certain risks in our specific business scenario, so we need a configurable toggle to enable or disable preemption.

3. We require metrics to support performance analysis based on priority-based scheduling.

## Modifications

Built upon the original priority-based scheduling, several optimizations have been implemented, including support for setting a default priority, adding a toggle for preemption logic, and introducing priority-based metrics.
image

## Benchmarking and Profiling
image

### Queueing behavior under high and low priority

Starting from concurrency level 40, requests begin to queue, with the number of queued low-priority requests significantly higher than that of high-priority ones, which aligns with expectations.
image

## Checklist


huangtingwei9988:
@hnyls2002 @xiezhq-hermann PTAL

xiezhq-hermann:
@harrisonlimh would appreciate if you can take a look at this PR.

python/sglang/srt/server_args.py
```diff
@@ -318,6 +318,8 @@ class ServerArgs:
     prefill_max_requests: Optional[int] = None
     schedule_policy: str = "fcfs"
     enable_priority_scheduling: bool = False
+    enable_try_preemption_by_priority: bool = False
```
harrisonlimh: Preemption is enabled by default in priority scheduling, so this flag is redundant. Would you be able to update it to `disable_try_preemption_by_priority` and use it as a toggle to disable the behavior instead?

```
        self.try_preemption = (
            self.enable_priority_scheduling
            and not self.server_args.disable_try_preemption_by_priority
        )
```

FYI - alternative way of disabling preemption is to set `priority_scheduling_preemption_threshold` to a value that is >  absl(highest_priority_int - lowest_priority_int). Sharing it in case that the mentioned use case of disabling preemption is urgent for the business need.

python/sglang/srt/managers/scheduler.py
harrisonlimh, line 1: Would it be possible to make one or two helper functions that calculate per-priority request counts in scheduler.py and scheduler_metrics_mixin.py and consolidate the usage? 

Perhaps that a) takes in the list of requests and return xxx_reqs_by_priority dictionary or b) additionally taking the dictionary and updating in place.

zhuxinjie-nz: Thank you very much — I’ll work on fixing these issues.

python/sglang/srt/server_args.py
zhuxinjie-nz: The parameters have already been updated as required. we still hope to be able to explicitly disable preemption via a parameter, which would make usage clearer and more explicit.

harrisonlimh: Thank you for updating it!

harrisonlimh:
LGTM! Thank you!

harrisonlimh (approved):

zhuxinjie-nz:
@harrisonlimh All required pipelines have passed — could you please merge it when you get a chance?
Thank you so much!🙏

hnyls2002:
Please avoid using confusing terminology, `try-preemption`, which only means allowing/enabling preemption or not.

This PR LGTM, but please clean the flag of preemption, then rebase this PR (or directly modify on this branch) @harrisonlimh.

- When priority scheduling is enabled, preemption is enabled by default.
- When priority scheduling is enabled but `--disable-preemption` is set, then disable the preemptive behavior.

huangtingwei9988:
@hnyls2002 Can you merge this PR?

python/sglang/srt/managers/scheduler.py
```diff
@@ -2004,6 +2007,7 @@ def _get_new_batch_prefill_raw(
             return None
 
         running_bs = len(self.running_batch.reqs)
+
```
stmatengss, line 2012: Duplicate blank

sufeng-buaa (approved):

PR #17051, opened 2026-01-14
[ConfigArgumentMerger] Improve ConfigArgumentMerger compatibility with external callers

YAMY1234:
## Motivation

The previous interface required external callers (e.g., Dynamo) to explicitly provide boolean argument metadata, which could become incompatible as SGLang’s argument parser evolved.

This PR improves robustness at the SGLang layer while preserving backward compatibility for existing callers.

## Checklist

## Review Process

### DIFF
--- a/python/sglang/srt/server_args_config_parser.py
+++ b/python/sglang/srt/server_args_config_parser.py
@@ -16,25 +16,37 @@ logger = logging.getLogger(__name__)
 class ConfigArgumentMerger:
     """Handles merging of configuration file arguments with command-line arguments."""
 
-    def __init__(self, parser: argparse.ArgumentParser):
+    def __init__(
+        self,
+        parser: argparse.ArgumentParser = None,
+        boolean_actions: List[str] = None,
+    ):
         """Initialize with list of store_true action names."""
         # NOTE: The current code does not support actions other than "store_true" and "store".
-        self.parser = parser
-        self.store_true_actions = [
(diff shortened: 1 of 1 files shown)


python/sglang/srt/server_args_config_parser.py
```diff
@@ -16,25 +16,37 @@
 class ConfigArgumentMerger:
     """Handles merging of configuration file arguments with command-line arguments."""
 
-    def __init__(self, parser: argparse.ArgumentParser):
+    def __init__(
+        self,
+        parser: argparse.ArgumentParser = None,
+        boolean_actions: List[str] = None,
```
hnyls2002, line 22: The `boolean_actions` is always `None`, is that necessary?

hnyls2002 (changes requested):

python/sglang/srt/server_args_config_parser.py
YAMY1234: Yeah boolean_actions is typically None here, but I added boolean_actions as a backward-compatible path for external/legacy callers (e.g., current Dynamo integrations) that initialize the merger without passing an argparse parser. Keeping this parameter avoids breaking those callers while leaving the in-repo behavior unchanged.

But you are right since this could not be used in the future, do you think we can add a deprecation warning? Once downstream users like Dynamo finish the upgrade completely we can remove this.

Issue #16327, opened 2026-01-03
[Bug] TMA issue when running long context PD+MTP with chunked-prefill-size: -1 in Prefill worker


YAMY1234:
### Checklist

- [ ] I searched related issues but found no solution.
- [ ] The bug persists in the latest version.
- [ ] Issues without environment info and a minimal reproducible demo are hard to resolve and may receive no feedback.
- [ ] If this is not a bug report but a general question, please start a discussion at https://github.com/sgl-project/sglang/discussions. Otherwise, it will be closed.
- [ ] Please use English. Otherwise, it will be closed.

### Describe the bug

When running **SGLang disaggregation (prefill+decode)** on **GB300 (SM103)** with **TRTLLM MLA attention backend** and with chunked-prefill-size=-1, the server intermittently crashes with repeated:

* `Error: Failed to initialize the TMA descriptor 1`

followed by a CUDA failure:

* `CUDA error: an illegal instruction was encountered` (reported by `ProcessGroupNCCL` watchdog)

This appears to be triggered during execution of a kernel that initializes a TMA descriptor (same `globalDim/globalStrides/boxDim` shown in logs), and the error repeats multiple times before the process terminates.

```
TMA Desc Addr:   0xffffffff5680
format         9
dim            3
gmem_address   0xfffd770e9c0c
globalDim      (7168,119384,1,1,1)
globalStrides  (2,14336,0,0,0)
boxDim         (32,128,1,1,1)
elementStrides (1,1,1,1,1)
interleave     0
swizzle        2
l2Promotion    2
oobFill        0
Error: Failed to initialize the TMA descriptor 1
TMA Desc Addr:   0xffffffff5680
format         9
dim            3
gmem_address   0xfffd770e9c0c
globalDim      (7168,119384,1,1,1)
globalStrides  (2,14336,0,0,0)
boxDim         (32,128,1,1,1)
elementStrides (1,1,1,1,1)
interleave     0
swizzle        2
l2Promotion    2
oobFill        0
Error: Failed to initialize the TMA descriptor 1
TMA Desc Addr:   0xffffffff5680
format         9
dim            3
gmem_address   0xfffd770e9c0c
globalDim      (7168,119384,1,1,1)
globalStrides  (2,14336,0,0,0)
boxDim         (32,128,1,1,1)
elementStrides (1,1,1,1,1)
interleave     0
swizzle        2
l2Promotion    2
oobFill        0
Error: Failed to initialize the TMA descriptor 1
[...]

PR #17089, opened 2026-01-14
Support symmetric memory pre-allocation to avoid fragmentation

nvcastet:
## Motivation

Related to #17038
PyTorch mempools never de-fragment memory in OOM scenarios, so we need to pre-allocate a large chunk of memory to limit fragmentation in symmetric memory pool.

Add `SGLANG_SYMM_MEM_PREALLOC_GB_SIZE` to specify the prealloc size in GiB.
Solved issues found in https://github.com/sgl-project/sglang/issues/16327 and https://github.com/sgl-project/sglang/issues/13863

CC @merrymercy @Fridge003

## Checklist

## Review Process


python/sglang/srt/environ.py
```diff
@@ -445,6 +445,9 @@ class Envs:
     # TokenizerManager
     SGLANG_REQUEST_STATE_WAIT_TIMEOUT = EnvInt(4)
 
+    # Symmetric Memory
+    SGLANG_SYMM_MEM_PREALLOC_GB_SIZE = EnvInt(-1)
```
Fridge003, line 452: Also update this flag in this file
https://github.com/sgl-project/sglang/blob/main/docs/references/environment_variables.md

Fridge003: A positive default value (maybe 4 or 8?)should be set when symm mem is enabled. We can do this in `server_args.py`

python/sglang/srt/managers/scheduler.py
```diff
@@ -960,7 +961,6 @@ def init_overlap(self):
         if self.device == "cpu":
             self.default_stream.synchronize = lambda: None  # No-op for CPU
 
-        self.forward_stream: CudaStream = self.device_module.Stream()
```
Fridge003, line 974: We only initialize `self.forward_stream` when overlap is enabled. It will change this behavior.

Maybe a better way is to pass the forward_stream from scheduler to model runner somehow? Changing the initialization of forward stream is quite risky

Fridge003 (dismissed):

python/sglang/srt/managers/scheduler.py
nvcastet: > We only initialize self.forward_stream when overlap is enabled. It will change this behavior.

I agree but having an unused cuda logical stream in the app should not have any side effect.

> Maybe a better way is to pass the forward_stream from scheduler to model runner somehow? Changing the initialization of forward stream is quite risky

Yes that was my first impl but the problem is that the stream needs to be created after `self.init_torch_distributed()` called in model runner init so that the correct current device is set.

hnyls2002: Ok, I agree there is no side effect. @Fridge003 Make sure the CIs are all passed before merge.

hnyls2002 (approved):

PR #17106, opened 2026-01-15
[diffusion] feat: support generating videos with specified resolution and aspect ratio in I2V

nono-Sang:
## Motivation

Currently in wan2.2 i2v, the input validation stage determines the video size based on the image size, using a core logic that relies on max_area and maintains the original aspect ratio through direct resizing. This approach has two issues:

1. **User-specified width/height are ignored**—the video size is determined by the image, not the user.  
2. **When the input image size changes, the tensor shape in the denoising stage becomes dynamic**, causing frequent recompilation when torch compile is enabled.

## Modifications

I have added a new configuration method to generate videos with specified resolution and aspect ratio. The key features are:

1. By **cropping and resizing** the original image, video distortion is avoided.  
2. When resolution and aspect ratio are configured as a fixed set of options (frame changes are not considered for now), we obtain a set of **static tensor shapes**, fully leveraging the benefits of torch compile.  
3. This method **coexists in isolation** with the original logic without causing conflicts.

Note: Setting aspect_ratio to "auto" will automatically choose the nearest preset (16:9/9:16/1:1) according to the image's original aspect ratio.

### Example
**[Old Method]** Using [832x1104](https://github.com/Wan-Video/Wan2.2/blob/990af50de458c19590c245151197326e208d7191/examples/i2v_input.JPG?raw=true) as the input image, regardless of the size setting, the video generated by wan2.2 is shown below, with dimensions **544x720**:

https://github.com/user-attachments/assets/5e391412-5c5b-4ec7-8156-464d75d2ad3f

**[New Method]** When resolution is set to 720p and aspect ratio to 9:16, wan2.2 generates the following video with dimensions **720x1280**:

https://github.com/user-attachments/assets/395f3975-3589-44f1-938e-9bc4b9c76958

## Checklist

## Review Process

### DIFF
--- a/python/sglang/multimodal_gen/configs/sample/sampling_params.py
+++ b/python/sglang/multimodal_gen/configs/sample/sampling_params.py
@@ -150,6 +150,10 @@ class SamplingParams:
     # whether to adjust num_frames for multi-GPU friendly splitting (default: True)
     adjust_frames: bool = True
 
+    # Resolution and aspect ratio for I2V/T2V
+    resolution: str | None = None
+    aspect_ratio: str | None = None
+
     def _set_output_file_ext(self):
         # add extension if needed
         if not any(
@@ -333,7 +337,11 @@ class SamplingParams:
 
         # Validate resolution against pipeline-specific supported resolutions
         if self.height is None and self.width is None:
-            if self.supported_resolutions is not None:
+            if self.resolution is not None and self.aspect_ratio is not None:
+                logger.info(
+                    f"Resolution and aspect ratio specified, height and width will be calculated based on them"
+                )
+            elif self.supported_resolutions is not None:
                 self.width, self.height = self.supported_resolutions[0]
                 logger.info(
                     f"Resolution unspecified, using default: {self.supported_resolutions[0]}"
@@ -717,6 +725,20 @@ class SamplingParams:
                 "and satisfy model temporal constraints. If disabled, tokens might be padded for SP."
                 "Default: true. Examples: --adjust-frames, --adjust-frames true, --adjust-frames false."
             ),
+        )
+        parser.add_argument(
+            "--resolution",
+            type=str,
+            default=SamplingParams.resolution,
+            choices=["480p", "580p", "720p"],
+            help="Target resolution for video generation. Options: 480p, 580p, 720p",
+        )
+        parser.add_argument(
+            "--aspect-ratio",
+            type=str,
+            default=SamplingParams.aspect_ratio,
+            choices=["auto", "16:9", "9:16", "1:1"],
+            help="Target aspect ratio for video generation. Options: auto, 16:9, 9:16, 1:1",
         )
         return parser
 
--- a/python/sglang/multimodal_gen/configs/sample/wan.py
+++ b/python/sglang/multimodal_gen/configs/sample/wan.py
@@ -247,6 +247,8 @@ class Wan2_2_T2V_A14B_SamplingParam(Wan2_2_Base_SamplingParams):
         default_factory=lambda: [
             (1280, 720),  # 16:9
             (720, 1280),  # 9:16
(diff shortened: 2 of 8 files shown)


python/sglang/multimodal_gen/runtime/utils/resolution_utils.py
```diff
@@ -0,0 +1,86 @@
...
+
+def isotropic_crop_resize_pil(
+    image: Image.Image, target_size: tuple[int, int]
+) -> Image.Image:
+    target_width, target_height = target_size
+    orig_width, orig_height = image.size
+    target_ratio = target_height / target_width
+    orig_ratio = orig_height / orig_width
+
+    if abs(orig_ratio - target_ratio) < 0.01:
```
BBuf: Avoid hardcoding 0.01. You can write it as an uppercase variable on line 26?

BBuf (approved):
Good job! cc @mickqian

python/sglang/multimodal_gen/runtime/utils/resolution_utils.py
nono-Sang: @BBuf done

python/sglang/multimodal_gen/configs/sample/sampling_params.py
```diff
@@ -718,6 +726,20 @@ def add_cli_args(parser: Any) -> Any:
                 "Default: true. Examples: --adjust-frames, --adjust-frames true, --adjust-frames false."
             ),
         )
+        parser.add_argument(
```
mickqian, line 729: we have `resolution_4k` and others as sampling params, please unify them.
In fact the previous way is better

python/sglang/multimodal_gen/configs/sample/wan.py
```diff
@@ -247,6 +247,8 @@ class Wan2_2_T2V_A14B_SamplingParam(Wan2_2_Base_SamplingParams):
         default_factory=lambda: [
             (1280, 720),  # 16:9
             (720, 1280),  # 9:16
+            (960, 512),  # 16:9
```
mickqian, line 250: @Brain97 could you take a look here, about the previous `supported resolution`

mickqian (changes requested):
If I'm understanding this correctly, when resolution and aspect_ratio are provided, the output resolution will be calculated in a different way with when `height and width` is provided?

It's a little confusing here, we need to provide more documentation and usage guide (in command line).

python/sglang/multimodal_gen/configs/sample/sampling_params.py
nono-Sang: @mickqian 
For the previous I2V implementation, **the user specified height and weight were actually ignored** because the generated video size was calculated based on the size of the input image.
**My goal is to provide a way to generate user specified sizes.** I provide a set of fixed optional sizes through resolution and aspect ratio. That is to say, (resolution, aspect_ratio) and video size are mapped one-to-one.
**Note that these two methods are completely isolated.** In fact, if you take a look at this [fal.ai-i2v](https://fal.ai/models/fal-ai/wan/v2.2-a14b/image-to-video) you can get a better understanding of what I want to do.

mickqian: make sense. we should explain the difference in help str and documentation, as it's easily confused

python/sglang/multimodal_gen/runtime/pipelines_core/schedule_batch.py
```diff
@@ -286,6 +286,8 @@ def log(self, server_args: ServerArgs):
                   image_path: {self.image_path}
                  save_output: {self.save_output}
             output_file_path: {self.output_file_path()}
+                  resolution: {self.resolution}
```
mickqian, line 289: move it under `height` to group similar params together

PR #17191, opened 2026-01-16
[BUGFIX] fix radix cache memory consumption to avoid OOM

yizhang2077:
## Motivation

When radix tree does splitting, new_node and child both hold the view for inserted value, so when child is evicted, `child.value` will not return to torch memory pool until all nodes along path hold the view of value are evicted. It may cause dynamic CUDA OOM when tree shares long prefix
This bug may also happen in other radix cache types.

## Modifications

use clone to release original value when do slice, it may cause overhead but I think it is small and can be hidden by overlap schedule

## Accuracy Tests
see ut, before this pr,  `torch_allocated_memory / tree total_size` is more than 400, which is abnormal. After this pr, it is less than 10 (it should be value dtype theoratically)

## Checklist

## Review Process

### DIFF
--- a/python/sglang/srt/mem_cache/radix_cache.py
+++ b/python/sglang/srt/mem_cache/radix_cache.py
@@ -654,10 +654,10 @@ class RadixCache(BasePrefixCache):
         new_node.parent = child.parent
         new_node.lock_ref = child.lock_ref
         new_node.key = child.key[:split_len]
-        new_node.value = child.value[:split_len]
+        new_node.value = child.value[:split_len].clone()
         child.parent = new_node
         child.key = child.key[split_len:]
-        child.value = child.value[split_len:]
+        child.value = child.value[split_len:].clone()
         new_node.parent.children[self.get_child_key_fn(key)] = new_node
 
         # Split hash_value if it was already computed, otherwise leave as None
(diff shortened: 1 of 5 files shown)


python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -654,10 +654,10 @@ def _split_node(self, key: RadixKey, child: TreeNode, split_len: int):
         new_node.parent = child.parent
         new_node.lock_ref = child.lock_ref
         new_node.key = child.key[:split_len]
-        new_node.value = child.value[:split_len]
+        new_node.value = child.value[:split_len].clone()
```
ispobock, line 657: We need to fix this for all radix cache variants.

Swipe4057:
Can I test it?

hanming-lu (approved):

python/sglang/srt/mem_cache/radix_cache.py
hnyls2002: Fixed.

Swipe4057:
> > Can I test it?
> 
> you can run unit test and print allocated memory. Bad case can be reproduced easily

the test lasted 15 minutes (2хH100)
start:
image

end:
image

Swipe4057:
> > > Can I test it?
> > 
> > 
> > you can run unit test and print allocated memory. Bad case can be reproduced easily
> 
> the test lasted 15 minutes (2хH100) start: image
> 
> end: image

OOM still happened, but later.

PR #17197, opened 2026-01-16
fix(hicache): enable prefetch from storage for root node in cross-pod…

chenkaiyue:
… deployment

## Motivation

In distributed deployments using HiCache with a shared storage backend (e.g., Mooncake Store), KV cache prefetch from remote storage fails when the requesting instance has an empty local RadixCache.

### Problem Scenario

1. **Pod A** processes a request and writes KV cache to Mooncake Store
2. **Pod B** (a new SGLang instance with empty cache) receives the same request
3. Pod B attempts to match prefix in its local RadixCache, but finds nothing
4. `req.last_node` points to the **root node** (id=0), which has `host_value = None`
5. `req.last_node.backuped` returns `False` (since root node has no backup)
6. **Prefetch is skipped**, even though Mooncake Store has the KV cache from Pod A

### Error Log Evidence

In our environment (k8s cluster with DeepSeek-V3 model + EAGLE speculative decoding), we observed:

```
# Pod B receives request but prefetch condition fails
# No [PREFETCH_QUERY] logs appear because prefetch is not triggered

# Expected behavior: Pod B should query Mooncake Store
# Actual behavior: Pod B skips prefetch entirely
```

The original code in `scheduler.py`:

```python
def _prefetch_kvcache(self, req: Req):
    if self.enable_hicache_storage:
        req.init_next_round_input(self.tree_cache)
        if req.last_node.backuped:  # <-- Root node fails this check!
            last_hash = req.last_host_node.get_last_hash_value()
            # ... prefetch logic
```

When `last_node` is root:
- `last_node.id == 0`
- `last_node.host_value == None`
- `last_node.backuped == False` (computed property based on `host_value`)
- **Prefetch never triggers** for new SGLang instances

### Impact

- **Cross-pod KV cache sharing completely broken** for new instances
- First request to a new pod cannot benefit from KV cache stored by other pods
- Defeats the purpose of using shared storage backend for KV cache in distributed deployments

### File Changed

- `python/sglang/srt/managers/scheduler.py`

### Code Changes

Extended the prefetch condition to also trigger when `last_node` is root node:

```python
def _prefetch_kvcache(self, req: Req):
    if self.enable_hicache_storage:
        req.init_next_round_input(self.tree_cache)
        # Modified condition: even if last_node is not backuped (e.g., root node),
        # we should still try to prefetch from storage backend since there might be
        # matching KV cache in remote storage (e.g., Mooncake Store from other pods).
        # This enables cross-pod KV cache sharing in distributed deployments.
        should_prefetch = req.last_node.backuped or (
            req.last_node.id == 0 and len(req.fill_ids) > 0
        )
        if should_prefetch:
            # For root node (id=0), use None as last_hash since it has no hash value
            last_hash = (
                req.last_host_node.get_last_hash_value()
                if req.last_node.backuped
                else None
            )
            # ... rest of prefetch logic unchanged
```

## Testing

Tested in production cluster with:
- **Model**: DeepSeek-V3.1-Terminus-W4AFP8 (MLA architecture)
- **Speculative Decoding**: EAGLE algorithm
- **Storage Backend**: Mooncake Store (RDMA protocol, P2P Handshake mode)
- **Scenario**: Multi-pod deployment with shared Mooncake Store

### Verification Results

After the fix:

```
# Pod A writes KV cache
[BATCH_SET] new_keys_to_write=12, already_exist=0
[BATCH_SET] write result: success=12/12

# Pod B (new instance) successfully prefetches
[PREFETCH_QUERY] EAGLE mode: converted to bigram, original_len=13, bigram_len=12
[BATCH_EXISTS] query_keys_len=12, hit_count=12, first_miss_idx=-1
Prefetch completed with 12 tokens

# Second request on Pod B shows cache reuse
Prefill batch, #new-seq: 1, #new-token: 2, #cached-token: 11
```

| Metric | Before Fix | After Fix |
|--------|------------|-----------|
| Cross-pod prefetch triggered | ❌ No | ✅ Yes |
| Mooncake Store hit rate | 0% | 100% |
| `#cached-token` on second request | 0 | 11 (out of 13) |

### DIFF
--- a/python/sglang/srt/managers/scheduler.py
+++ b/python/sglang/srt/managers/scheduler.py
@@ -1559,10 +1559,25 @@ class Scheduler(
     def _prefetch_kvcache(self, req: Req):
         if self.enable_hicache_storage:
             req.init_next_round_input(self.tree_cache)
-            if req.last_node.backuped:
-                # only to initiate the prefetch if the last node is backuped
-                # otherwise, the allocated GPU memory must be locked for integrity
-                last_hash = req.last_host_node.get_last_hash_value()
+            # Modified condition: even if last_node is not backuped (e.g., root node),
+            # we should still try to prefetch from storage backend since there might be
+            # matching KV cache in remote storage (e.g., Mooncake Store from other pods).
+            # This enables cross-pod KV cache sharing in distributed deployments.
+            should_prefetch = req.last_node.backuped or (
+                req.last_node.id == 0 and len(req.fill_ids) > 0
+            )
+            logger.debug(
+                f"[PREFETCH_DEBUG] rid={req.rid}, last_node.id={req.last_node.id}, "
+                f"last_node.backuped={req.last_node.backuped}, fill_ids_len={len(req.fill_ids)}, "
+                f"should_prefetch={should_prefetch}"
+            )
+            if should_prefetch:
+                # For root node (id=0), use None as last_hash since it has no hash value
+                last_hash = (
+                    req.last_host_node.get_last_hash_value()
+                    if req.last_node.backuped
+                    else None
+                )
                 matched_len = len(req.prefix_indices) + req.host_hit_length
                 new_input_tokens = req.fill_ids[matched_len:]
 


hnyls2002:
@chenkaiyue Main now initiates storage prefetch for the root node as well: scheduler.py checks last_host_node.backuped or last_host_node is self.tree_cache.root_node (landed via #19663), which covers the cross-pod first-request case this PR addressed. Closing as superseded; please reopen if you still see prefetch skipped on current main.

PR #17202, opened 2026-01-16
[Feat] Accelerate qwen3vl by remove cpu op

ZLkanyo009:
# Motivation
Accelerate qwen3vl by delete contiguous in vision_attention && delete D2D in torch.where and index_put

During the forward pass of the ViT in the Qwen3VL multimodal model, some D2D and H2D operations were observed, which impacted the forward propagation speed of the ViT. This PR eliminates two such instances, as shown below.

# command
```
export SGLANG_USE_AITER=1
model=/models/nvme3/data/models/Qwen3-VL-235B-A22B-Instruct-FP8-dynamic/
TP=8
EP=1
 
echo "launching ${model}"
echo "TP=${TP}"
echo "EP=${EP}"

    --model-path ${model} \
    --host localhost \
    --port 9000 \
    --tp-size ${TP} \
    --ep-size ${EP} \
    --trust-remote-code \
    --mem-fraction-static 0.8 \
    --disable-radix-cache \
    --max-prefill-tokens 16384 \
    --cuda-graph-max-bs 128 \
    --max-running-requests 128 \
    --mm-attention-backend aiter_attn
```

### DIFF
--- a/python/sglang/srt/layers/attention/vision.py
+++ b/python/sglang/srt/layers/attention/vision.py
@@ -1005,9 +1005,9 @@ class VisionAttention(nn.Module):
             q, k, v = qkv.split([self.q_size, self.kv_size, self.kv_size], dim=-1)
 
             # [b, s, embed_dim] --> [b * s, head, head_size]
-            q = q.reshape(bsz * s, head, -1).contiguous()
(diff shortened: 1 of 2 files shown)


ZLkanyo009:
# acc
Overall accuracy: 0.603

PR #17245, opened 2026-01-17
[CI] Add partition to stage-b-test-large-1-gpu (11->12)

alisonshao:
## Summary

Add one more partition to `stage-b-test-large-1-gpu` to reduce per-partition test time.

**Problem**: Some test jobs are exceeding 20 minutes after recent test additions.

**Fix**: Increase partitions from 11 to 12.

| Setting | Before | After |
|---------|--------|-------|
| Partitions | 11 | **12** |
| Est. time/partition | ~20 min | ~18 min |

Ref: https://github.com/sgl-project/sglang/commit/a87df2f

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -655,7 +655,7 @@ jobs:
       fail-fast: false
       max-parallel: ${{ fromJson(needs.check-changes.outputs.max_parallel) }}
       matrix:
-        partition: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-test.yml
```diff
@@ -655,7 +655,7 @@ jobs:
       fail-fast: false
       max-parallel: ${{ fromJson(needs.check-changes.outputs.max_parallel) }}
       matrix:
-        partition: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+        partition: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
```
Kangyan-Zhou: Should we do 12 so that we have more buffer?

PR #17260, opened 2026-01-17
[Feature] [Ngram spec] Support ngram spec v2

SYChen123:
## Motivation

Support spec v2 for ngram speculative decoding. Enable overlap schedule on ngram speculative decoding, as mentioned in #11762 and #21052.

## Modifications
The main modification is to add ngram_worker_v2 and modify NgramVerifyInput to be compatible with overlap scheduling. The modification includes

- Before drafting, prepare the complete output token ids (include the previous round). All the accepted_tokens of round N are stored in the batch.spec_info and they will be processed in round N+1 before draft.
- Use move_accepted_tokens_to_target_kvcache kernel to move the kvcache of accepted tokens to the correct slots.
- Modify _prepare_for_speculative_decoding, _prepare_draft_tokens, etc. to adapt the spec-overlap workflow.

## Accuracy Tests

**gsm8k test on Qwen25-VL-72B-Instruct.**
without spec v2
image

with spec v2
image

## Benchmarking and Profiling

Benchmarking on Qwen25-VL-72B-Instruct
`python -m sglang.bench_serving --backend sglang-oai-chat --dataset-name image --host 127.0.0.1 --port 9090 --model Qwen/Qwen2.5-VL-72B-Instruct/ --num-prompts 256 --image-count 1 --image-resolution 1080p --image-content random --random-input-len 2500 --random-range-ratio 1 --random-output-len 600 --max-concurrency 32 --request-rate 0.7`

The E2E latency decreases from 5681ms to 5021ms (**11%**), and TPOT decreases from 8.50ms to 7.36ms (**13%**)

without spec v2
image

with spec v2
image

During benchmark, we found the performance fluctuates and depends on token accept rate which may vary across different runs of benchmark. To further verify the effectiveness of spec v2 and rule out the influence of accept rate on the performance, we **mock the draft tokens** generated by ngram and replace all of them with 0 to make sure only 1 token is accepted. In such case, we do the benchmark and the result is as below. The TPOT decreases from 24.43 to 23.44ms. The E2E latency decreases from 15232ms to 14626ms.
Without spec v2
image

spec v2
image

---
### CI States

Latest PR Test (Base): :x: [Run #27261275928](https://github.com/sgl-project/sglang/actions/runs/27261275928)
Latest PR Test (Extra): :x: [Run #27261275550](https://github.com/sgl-project/sglang/actions/runs/27261275550)


python/sglang/srt/speculative/ngram_worker_v2.py
```diff
@@ -0,0 +1,288 @@
+import logging
```
hnyls2002: Explain the modification compared to the spec v1 version of ngram.

python/sglang/srt/speculative/ngram_worker_v2.py
```diff
@@ -0,0 +1,288 @@
...
+from typing import Optional
```
hnyls2002: Show me the profiling (nsys / torch) to prove that there is no CPU time during two continuous decoding batches.

hnyls2002 (changes requested):
In Eagle spec v2 (current), we use this pipeline:

```md
(CUDA) draft         -> verify              -> draft extend    ->  draft

(CPU)  launch verify -> launch draft extend -> sync & schedule & launch draft
```

How do you solve the CPU overhead when there is no draft / draft extend (they are purely on CPU)

python/sglang/srt/speculative/ngram_worker_v2.py
SYChen123: **Main observation**: The gap duration between each 2 decode batches decreases from 4.95ms (spec v1) to 3.96ms (spec v2). The profile is shown as below. 

**1. GPU side**

- spec v1. The gap between 2 decode batches are 4.95ms. 

image

- spec v2. The gap between 2 decode batches are 3.96ms. The gap decreases **1ms** between every 2 decode batches. 

image



**2. CPU side**

- spec v1. After CudaGraphLaunch, it basically synchronize immediately. All the operations afterwards contribute to the gap. 
image

- spec v2. Synchronization is delayed to get_next_batch_to_run and the CPU overhead of operations in between can be eliminated.
image

python/sglang/srt/speculative/ngram_worker_v2.py
SYChen123: Main modifications are listed below.

- The accepted_tokens and accept_lens of round N are stored in the batch.spec_info. They will be used to obtain the correct accepted tokens before draft of round N+1. Thus draft can have correct input tokens with overlap schedule enabled.

- Use move_accepted_tokens_to_target_kvcache kernel to move the kvcache of accepted tokens to the correct slots, because the accept indices in ngram spec can be non-consecutive and need to reorder the kvcache.

- Add _prepare_for_speculative_decoding_v2, _prepare_draft_tokens_v2, etc. to adapt the spec-overlap workflow.

python/sglang/srt/speculative/ngram_worker.py
```diff
@@ -174,32 +211,49 @@ def _prepare_for_speculative_decoding(self, batch: ScheduleBatch):
...
+                seq_len = batch.seq_lens_cpu[i]
+                req_mask = torch.ones((self.draft_token_num, seq_len)).to(
+                    device=self.device, non_blocking=True
+                )
                 req_mask = torch.cat(
-                    (req_mask, torch.from_numpy(mask[i]).cuda()), dim=1
+                    (
+                        req_mask,
+                        torch.from_numpy(mask[i]).to(
+                            device=self.device, non_blocking=True
```
SYChen123, line 298: Do not use .to()

Ratish1:
Hey @SYChen123 , I traced the accept-length failure to the NGRAM spec-v2 overlap path. (FAILURE TEST HERE: https://github.com/sgl-project/sglang/actions/runs/24176255887/job/70558819211)

Root cause: `_prepare_draft_tokens` correctly builds the draft context from:

`origin_input_ids + output_ids + prev_tokens`

where `prev_tokens` comes from `batch.spec_info`, because `req.output_ids` is incomplete under overlap scheduling.
However, the length passed to `NgramCorpus.batch_get` was still only:

`len(origin_input_ids) + len(output_ids)`

So the trie received a context that included the pending spec-v2 tokens, while `total_len` described a shorter logical
sequence. That can make the per-request trie match state advance/rebuild from the wrong logical length. The resulting draft tree can miss the target continuation, and the verifier then rejects correctly, lowering average accept length below the `1.8` threshold.

The fix I tested is to include the pending spec-v2 tokens in the length passed to `batch_get` ( I have included the fix below in the review):

`len(origin_input_ids) + len(output_ids) + len(prev_tokens)`

Comparison against main:

| Variant | Main score | PR + fix score | Main avg accept | PR + fix avg accept |
|---|---:|---:|---:|---:|
| Base / fa3 | 0.805 | 0.805 | 1.8846 | 1.8018 |
| Flashinfer | 0.810 | 0.805 | 1.8826 | 1.8040 |
| Triton | 0.810 | 0.805 | 1.8879 | 1.8119 |

Main’s observed NGRAM score range in this run was `0.805-0.810`; PR + fix is `0.805`, so accuracy matches the observed main range. Thanks.

cc: @hnyls2002

python/sglang/srt/speculative/ngram_worker.py
```diff
@@ -152,20 +166,42 @@ def _init_preallocated_tensors(self):
...
             check_token = self._efficient_concat_last_n(
-                req.origin_input_ids, req.output_ids, self.max_trie_depth
+                req.origin_input_ids,
+                req.output_ids + prev_tokens,
+                self.max_trie_depth,
             )
             req_ids.append(req.rid)
             batch_tokens.append(check_token)
+            i += 1
             total_lens.append(len(req.origin_input_ids) + len(req.output_ids))
```
Ratish1:
```suggestion
            total_lens.append(len(req.origin_input_ids) + len(req.output_ids) + len(prev_tokens))
```

SYChen123:
Hi @Ratish1 . Thanks for your comments and work. Your analysis and suggestions are solid and greatly appreciated. However, while trying to resolve this accept length issue, I found a potentially more severe problem: 

If I **modify the _update_ngram_corpus to put the prev_tokens into the ngram trie**, the gsm8k accuracy will significantly **drop**. 

This is wired because no matter what the draft tokens are, the accuracy should be normal if the target_verify works. I've done several debugging explorations which are as follows.
1. Reset to commit 499106b5. The accuracy drop still exists. This commit has all the core codes of ngram spec v2 but without recent refactors on ngram corpus. Thus, we can rule out the suspicion of ngram corpus refactor.
2. Run the test with CUDA_LAUNCH_BLOCKING=1 or use torch.cuda.synchronize. The accuracy drop still exists. Thus the root cause is not related to the CPU-GPU synchronization mechanism in spec v2.
3. If I force the accept length to be 1, the gsm8k accuracy becomes normal. 

Could you please take a look at this problem? You may use the above debug information as a starting point. 
cc @hnyls2002 


### How to reproduce this problem

#### Reproduce accuracy drops
1. Modify _update_ngram_corpus as following.
````
def _update_ngram_corpus(self, batch: ModelWorkerBatch):
  stride = self.draft_token_num
  batch_tokens = []
  i = 0
  for req in batch.reqs:
      prev_tokens = (
          self.prev_token_ids[i * stride : i * stride + self.prev_accept_lens[i]]
          if not batch.has_grammar
          else []
      )
      put_ids = self._efficient_concat_last_n(
          req.origin_input_ids, req.output_ids + prev_tokens, self.max_trie_depth
      )
      batch_tokens.append(put_ids)
      i += 1
  self.ngram_corpus.batch_put(batch_tokens)
````
2. Run the test. You are supposed to observe the accuracy is around **0.5~0.6**.
`python ./test/registered/spec/test_ngram_speculative_decoding.py`

#### Force the accept length to be 1.

1. Modify _update_ngram_corpus the same as above.
2. Insert the following piece of code into ngram_worker's forward_batch_generation function, right after `verify_input.sample(model_worker_batch, logits_output, vocab_mask)`
````
import os
_debug_cap = os.environ.get("SGLANG_NGRAM_DEBUG_CAP_ACCEPT")
if _debug_cap:
    cap = int(_debug_cap)
    accept_length = accept_length.clamp(max=cap)
    for bx in range(len(accept_length)):
        al = accept_length[bx]
        accept_index[bx, al:] = -1
````
3. Run the test with the env set. You should observe the gsm8k accuracy recovers to around 0.8.
`SGLANG_NGRAM_DEBUG_CAP_ACCEPT=1 python ./test/registered/spec/test_ngram_speculative_decoding.py`

SYChen123:
### **04/17 Update**
I dump the input of batch_put and batch_get and compare the dumped result of ngram spec v1 and ngram spec v2 on gsm8k. I found that even for ngram spec v1, after the correct answer is predicted, it will predict the subsequent question-answer pair as well.
image

The only difference is that ngram spec v1 stops after "Question" generated (gsm8k set Question as stop string), while ngram spec v2 does **NOT** stop correctly and finally generates a new question-answer pair with wrong answer.
image

When walking through the code, I noticed that in req.check_finished(), it truncates the output_ids for stop string matching. If the number of accepted tokens becomes large, the stop token may fall out of the truncate window. So I modify the process_batch_result_decode to do req.check_finished() in a token-by-token manner.

After the modification, appending prev_token to ngram tree will not cause accuracy degradation, for both concurrency=1 and concurrency>1. That's what we expect. No matter what draft tokens are, accuracy should be ok because verify will reject the wrong token.
image

The test_ngram_speculative_decoding.py is stably passed.
image


@hnyls2002 @Ratish1 

----------------------------------------------

### **04/16 Update**
The real root cause of the accuracy issue may **not** be the previously mentioned reason. Otherwise, the gsm8k accuracy should be normal if concurrency is 1 (without prefill-decode interleaved). However, gsm8k accuracy is still around 0.5 if concurrency set to 1, and the accept length is still abnormally high. 
When setting concurrency to 1 and comparing the output of ngram spec v1 and spec v2, I found the generated output of spec v2 is often with plausible but incorrect continuation, which leads to the finally wrong answer. For example, in spec v2 (left of the figure below), the answer 104 is correctly predicted. But it doesn't stop and instead generates some plausible tokens.
image

Still working on it. Trying to visualize the ngram trie for comparison and see what really happens inside the trie.

------------------------------------------------------


### **04/15 Update**
For the accuracy problem, I think I found the root cause. It's not related to the data race on cpu or gpu, but the **prev_token corrupts the ngram trie** if inserting prev_token to ngram during forward. 

Let's say we have such an execution order: `Decode A => Prefill B => Decode A+B`. The detailed operation is like `Decode A => Prefill B => process A's result => update A's output_ids => merge batch A & B => Decode A+B`. When Decode A+B, the output_ids of requests in batch A is already complete. The `prev_token` is actually the latest verified tokens. Therefore, `A.output_ids + prev_token` contains repeated tokens and inserting it into ngram corpus will corrupt the trie. 

As for the drafting phase, in this case, for requests in batch B, they need to use prev_token to make the latest verified token included in the draft. For requests in batch A, it's **not** necessary to use prev_token because req.output_ids already contains the latest verified tokens. Use `A.output_ids + prev_token` to match the trie will not affect the accuracy. But not sure if it'll affect the quality of generated draft tokens.

@hnyls2002 @Ratish1

hnyls2002:
/rerun-test test_spec_ngram.py test_spec_ngram_extra.py test_spec_eagle_parity.py test_spec_standalone.py

hnyls2002 (approved):

SYChen123:
**Profile result update**

From v1 to v2, the bubble between 2 decode batches decreases from **7.25ms to 2.52ms** (bs=4, model is Qwen3-8B).

ngram v1 (baseline)
image
ngram v2 (with this pr)
image

commit id
ngram v1: 255843d4546289c1b4fdafb9d06e85e4bd6a1c9e
ngram v2: 111009ea542e3430d8eed6e8e99dd8558982699f

cc @hnyls2002

PR #17270, opened 2026-01-17
Harden OpenAI upload filename handling

MaxMiksa:
## Motivation
Fixes #17267. Prevent path traversal in OpenAI-compatible multimodal upload filenames.

## Modifications
- Sanitize uploaded filenames (basename + safe allowlist) and preserve original filename as metadata.
- Enforce uploads-root boundary checks in OpenAI image/video endpoints.
- Add unit tests for traversal filenames and boundary enforcement.
- Update OpenAI API docs with filename sanitization notes.

## Accuracy Tests
Not applicable (no model forward changes). Tests not run on Windows because `triton` is unavailable.

## Benchmarking and Profiling
Not applicable.

## Checklist
- [ ] Format your code according to the [Format code with pre-commit](https://docs.sglang.io/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [x] Add unit tests according to the [Run and add unit tests](https://docs.sglang.io/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [x] Update documentation according to [Write documentations](https://docs.sglang.io/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.io/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.io/developer_guide/contribution_guide.html#benchmark-the-speed).
- [ ] Follow the SGLang code style [guidance](https://docs.sglang.io/developer_guide/contribution_guide.html#code-style-guidance).

PR #17271, opened 2026-01-17
Add OpenAI media URL guardrails

MaxMiksa:
## Motivation
Fixes #17268. Adds guardrails for OpenAI-compatible image/video URL fetches to mitigate SSRF.

Depends on #17270 (stacked on the filename path traversal fix).

## Modifications
- Validate URL scheme and resolve host/IP to block private/reserved ranges.
- Add allowlist configuration and a global enable/disable gate for remote URL fetches.
- Enforce redirect limits, timeouts, and max response size in `httpx` fetches.
- Update OpenAI image/video endpoints to use the new validator.
- Add unit tests for private IPs, redirects, and size limits.
- Update OpenAI API docs with new security settings.

## Accuracy Tests
Not applicable (no model forward changes). Tests not run on Windows because `triton` is unavailable.

## Benchmarking and Profiling
Not applicable.

## Checklist


python/sglang/multimodal_gen/runtime/entrypoints/openai/utils.py
```diff
@@ -83,85 +88,310 @@ def _parse_size(size: str) -> tuple[int, int] | tuple[None, None]:
...
+        )
+    except (TypeError, ValueError):
+        timeout = default_timeout
+
+    schemes = {
+        scheme.strip().lower()
+        for scheme in allowed_schemes.split(",")
+        if scheme.strip()
+    }
+    if not schemes:
```
Kangyan-Zhou, line 207: Could the check be affected by TOCTOU DNS Rebinding Vulnerability?

MaxMiksa: @Kangyan-Zhou Thanks for flagging this.

As implemented, the current hostname-to-IP validation can be affected by a TOCTOU / DNS rebinding scenario, because the DNS result used during validation may differ from what the HTTP client resolves at connection time.

I’m going to address this in this PR by binding the fetch to already-validated targets (DNS pinning), meaning I will resolve and validate first, then connect to the validated IP while preserving the original hostname for Host/SNI, and re-validate on every redirect hop. I’ll also add/adjust tests to cover this rebinding/TOCTOU case.

If there is a preference for the exact mitigation (e.g. pin the first validated IP vs iterate validated IPs, and whether you want HTTPS-only as the default policy), please let me know — otherwise I’ll proceed with the above.

MaxMiksa: Update: 

✅ Just pushed a follow-up commit (435c378) that fixes the DNS rebinding/TOCTOU issue by pinning the fetch to the validated IP. It still keeps the original hostname for the Host header and SNI. I also added logic to re-validate on every redirect and included unit tests for IPv4/IPv6.

❓Inspired by the Check Status below, would you mind adding run-ci to kick off the CI?

Kangyan-Zhou: https://github.com/sgl-project/sglang/actions/runs/21166894484/job/60941803164?pr=17271

Can you fix the CI issue?

PR #17276, opened 2026-01-18
Add Qwen3VL Eagle3 Inference Support

ardenma:
## Motivation

Add support for EAGLE3 for Qwen3VL. I have not extensively tried to optimize benchmark performance, but the accuracy/performance numbers at the bottom are there to show that inference works.

## Modifications

Adds a few things to qwen3vl model:
- `get_embed_and_head`
- `set_eagle3_layers_to_capture`
- and wires the aux hidden states into the `logits_processor`

Starts hidden state capture after the deepstack injection to avoid mismatch with specforge training code. In sglang we capture the hidden states after the `hidden_states + residual` but specforge/hf captures after `hidden_states + residual + deepstack`. We could probably fix the timing sync... but this is less error prone.

## Accuracy Test

See https://github.com/sgl-project/SpecForge/pull/431 for the specforge training implementation and results.

## Benchmark & Profiling

Sglang command: `python3 -m sglang.launch_server --model Qwen/Qwen3-VL-8B-Instruct --speculative-algorithm EAGLE3 --speculative-draft-model-path ../SpecForge/outputs/Qwen3-VL-8B-eagle3/epoch_11_step_140000 --speculative-num-steps 6 --speculative-eagle-topk 4 --spec
ulative-num-draft-tokens 16 --mem-fraction-static 0.75 --cuda-graph-max-bs 1 --tp 1 --trust-remote-code --host 0.0.0.0 --port 30000 --dtype bfloat16`
Benchmark command: `python3 benchmarks/bench_eagle3.py --model-path Qwen/Qwen3-VL-8B-Instruct --port 30000 --config-list 1,0,0,0 1,3,1,4 --benchmark-list mmstar:100 --skip-launch-server`

```
{
    "mmstar": [
        {
            "batch_size": 1,
            "steps": null,
            "topk": null,
            "num_draft_tokens": null,
            "metrics": [
                {
                    "latency": 108.34728950168937,
                    "output_throughput": 172.75928254493292,
                    "accept_length": 2.1631803998613197,
                    "accuracy": 0.24,
                    "num_questions": 100,
                    "num_valid_predictions": 56,
                    "categorical_performance": null
                }
            ],
            "num_samples": 100
        }
    ]
}
```

## Checklist

## Review Process

### DIFF
--- a/python/sglang/srt/models/qwen3_vl.py
+++ b/python/sglang/srt/models/qwen3_vl.py
@@ -733,6 +733,9 @@ class Qwen3VLForConditionalGeneration(nn.Module):
         self.num_deepstack_embeddings = len(self.deepstack_visual_indexes)
         self.use_deepstack = {Modality.IMAGE: True, Modality.VIDEO: True}
 
+        # For EAGLE3 support
+        self.capture_aux_hidden_states = False
+
     def separate_deepstack_embeds(self, embedding):
         assert (
             embedding.shape[-1] % (1 + self.num_deepstack_embeddings) == 0
@@ -920,6 +923,10 @@ class Qwen3VLForConditionalGeneration(nn.Module):
             pp_proxy_tensors=pp_proxy_tensors,
         )
 
+        aux_hidden_states = None
+        if self.capture_aux_hidden_states:
+            hidden_states, aux_hidden_states = hidden_states
+
         if self.pp_group.is_last_rank:
             if not get_embedding:
                 return self.logits_processor(
@@ -927,6 +934,7 @@ class Qwen3VLForConditionalGeneration(nn.Module):
                     hidden_states,
                     self.lm_head,
                     forward_batch,
+                    aux_hidden_states,
                 )
             else:
                 return self.pooler(hidden_states, forward_batch)
@@ -1011,6 +1019,30 @@ class Qwen3VLForConditionalGeneration(nn.Module):
 
                 weight_loader = getattr(param, "weight_loader", default_weight_loader)
                 weight_loader(param, loaded_weight)
+
+    def get_embed_and_head(self):
+        return self.model.embed_tokens.weight, self.lm_head.weight
+
+    def set_eagle3_layers_to_capture(self, layer_ids: Optional[List[int]] = None):
+        """Configure layers for EAGLE3 auxiliary hidden state capture."""
+        self.capture_aux_hidden_states = True
+        self.model.capture_aux_hidden_states = True
+        # NOTE: there's a +1 offset because sglang captures input to layer (output of previous)
+        if layer_ids is None:
+            # Qwen3VL uses config directly, text_config is only for Qwen3Omni
+            if hasattr(self.config, "text_config") and self.config.text_config is not None:
+                num_layers = self.config.text_config.num_hidden_layers
+            else:
+                num_layers = self.config.num_hidden_layers
+            # Qwen3VL uses deepstack at decoder layers 0, 1, 2
+            # Start at layer 4 to capture output of layer 3 (avoids deepstack timing mismatch)
+            self.model.layers_to_capture = [
+                4,
+                num_layers // 2,
+                num_layers - 3,
+            ]
+        else:
+            self.model.layers_to_capture = [val + 1 for val in layer_ids]
 
 
 EntryClass = Qwen3VLForConditionalGeneration
--- a/python/sglang/srt/models/qwen3_vl.py
+++ b/python/sglang/srt/models/qwen3_vl.py
@@ -1030,7 +1030,10 @@ class Qwen3VLForConditionalGeneration(nn.Module):
         # NOTE: there's a +1 offset because sglang captures input to layer (output of previous)
         if layer_ids is None:
             # Qwen3VL uses config directly, text_config is only for Qwen3Omni
-            if hasattr(self.config, "text_config") and self.config.text_config is not None:
+            if (
+                hasattr(self.config, "text_config")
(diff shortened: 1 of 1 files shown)


EanWang211123:
Hi, thanks for this PR.

I’ve recently been trying speculative decoding on **Qwen3-VL** as well, and I observed the same issue in both **my own fork** and **your PR**: **after enabling speculative decoding, the output becomes different from the baseline output**.

Interestingly, I did **not** observe this issue on **Qwen2.5-VL** under similar testing, so I wanted to ask:

**Have you also seen this behavior on Qwen3-VL (output mismatch when speculative decoding is enabled)?**

### Environment

* **GPU:** 4090D
* **Models:** `Qwen3-VL-8B-Instruct` and `Qwen3-VL-8B-Instruct-Eagle3`

### Launch Commands

**Baseline**

```bash
CUDA_VISIBLE_DEVICES=1 \
  --model-path /models/Qwen3-VL-8B-Instruct \
  --tp-size 2 \
  --dtype bfloat16 \
  --mem-fraction-static 0.75 \
  --cuda-graph-max-bs 32 \
  --context-length 40960 \
  --port 30000
```

**EAGLE3 (speculative decoding enabled)**

```bash
CUDA_VISIBLE_DEVICES=1 \
  --model-path /models/Qwen3-VL-8B-Instruct \
  --speculative-algorithm EAGLE3 \
  --speculative-draft-model-path /models/Qwen3-VL-8B-Instruct-Eagle3-bak/ \
  --tp-size 1 \
  --dtype bfloat16 \
  --mem-fraction-static 0.75 \
  --cuda-graph-max-bs 32 \
  --context-length 40960
```

### Test Method

I used the test file from this PR diff:
`https://github.com/sgl-project/sglang/pull/18387/changes#diff-37ee3735f2eae0c7f3bedf7931ccca202759474126de338a0dcfa6f6ef1e4c04`

### Result

Using a **single MMStar sample**, the outputs are **completely inconsistent** between baseline and EAGLE3 (including the final answer).

#### Baseline output (full)

```text
Looking at the image, we can see a brown suitcase sitting inside what appears to be the trunk of a car. On top of the suitcase, there are various stickers. In front of the suitcase, there is an open magazine or book lying on the car's trunk floor.

Let's evaluate the options:

- **A: The suitcase is on the book.** — This is incorrect. The suitcase is behind the book/magazine, and the book is in front of the suitcase. The suitcase is not resting on top of the book.

- **B: The suitcase is beneath the cat.** — There is no cat visible in the image.

- **C: The suitcase is beneath the bed.** — There is no bed in the image.

- **D: The suitcase is beneath the book.** — This is incorrect. The book is in front of the suitcase, not above it. The suitcase is behind the book.

None of the options seem to describe the relationship correctly based on the image. However, if we must choose the *least incorrect* or the one that might be intended, we need to reconsider.

Actually, looking again, the book is lying *in front of* the suitcase, not *on top of* it. The suitcase is *behind* the book. So none of the options accurately describe the spatial relationship.

But let's re-examine option D: "The suitcase is beneath the book." This would mean the book is on top of the suitcase. In the image, the book is *in front* of the suitcase, not on top of it. So D is also incorrect.

Wait — perhaps there's a misinterpretation. The book is lying on the trunk floor, and the suitcase is sitting on the trunk floor behind it. So the suitcase is *behind* the book, not beneath it.

Given that, none of the options are correct. However, if we consider the book as being "in front of" the suitcase, and the suitcase is "behind" the book, then the suitcase is not beneath the book — it's behind it.

Perhaps the question is flawed, or perhaps we are meant to interpret "beneath" as "behind" in a loose sense, but that's not accurate.

Let me check the image again. The book is lying flat on the trunk floor, and the suitcase is sitting upright behind it. So the suitcase is not beneath the book — it's behind it.

Therefore, **none of the options are correct**
```

#### EAGLE3 output (full)

```text
Looking at the image, we can see:

- A brown suitcase is sitting inside what appears to be the trunk of a car.
- A magazine (book) is lying on the floor of the trunk, partially underneath the suitcase.
- The suitcase is positioned on top of the magazine.

Therefore, the suitcase is beneath the book? No, that's incorrect. The suitcase is *on top of* the book.

Let's evaluate the options:

A: The suitcase is on the book. — This is incorrect. The suitcase is *on top of* the book, but the book is *underneath* the suitcase. The phrasing "on the book" is ambiguous, but in spatial relationships, "on" typically means directly above and resting on top of. However, looking at the options, this is the only one that describes the suitcase being in a position relative to the book.

B: The suitcase is beneath the cat. — There is no cat visible in the image.

C: The suitcase is beneath the bed. — There is no bed visible in the image.

D: The suitcase is beneath the book. — This is incorrect. The suitcase is *on top of* the book, not beneath it.

Actually, re-examining the image, the suitcase is resting on the magazine, which is lying on the floor. So the suitcase is *on* the magazine (the book). The book is *underneath* the suitcase.

Option A says "The suitcase is on the book." This is the most accurate description among the given choices, even if the wording is slightly imprecise. The suitcase is positioned on top of the book.

Therefore, the correct answer is **A: The suitcase is on the book.**
```

In my case, this is not just a stylistic difference — it changes the **final answer**.

If you have seen this as well, do you have any ideas about the root cause (e.g. VL-specific preprocessing / vision token handling / speculative verification path differences for Qwen3-VL)?

hnyls2002:
@ardenma Thanks for the work on this. EAGLE3 support for Qwen3-VL has since been merged via #22230, and main's qwen3_vl.py already includes the EAGLE3 capture hooks. Closing as superseded; please reopen if this PR covers cases that #22230 does not.

PR #17292, opened 2026-01-18
support model teleflm for npu

McZyWu:
## Motivation

support model teleflm for npu.
The configuration (config.json) of model teleflm doesn't have the key 'architectures', resulting in failure of entering the model.

## Modifications

As follows. the same as pr:https://github.com/sgl-project/sglang/pull/17285

## Accuracy Tests

image


McZyWu:
revise in config.json

PR #17300, opened 2026-01-18
[FIX] Always support TP > 4 for FP4 Gemm

danielafrimi:
**Summary**
This PR enables FP4 (NVFP4) quantization to work with TP >= 4

**Background**
Previously, FP4 quantized models would fail to initialize with TP=4/8 due to kernel alignment requirements. The FlashInfer-CUTLASS FP4 GEMM kernels require the N/K-dimension to be divisible by 32.

**Changes**
This PR makes the following changes to support TP >= 4, by padding the weights accordingly. In addition, in cases where we pad the K dim of the weights, we need to pad the activation in the forward pass.

### DIFF
--- a/python/sglang/srt/layers/quantization/modelopt_quant.py
+++ b/python/sglang/srt/layers/quantization/modelopt_quant.py
@@ -168,23 +168,22 @@ def round_up_to_multiple(x: int, m: int) -> int:
     return (x + m - 1) // m * m
 
 
-def pad_nvfp4_weight_for_cutlass(
+def pad_nvfp4_weight(
     weight: torch.Tensor,
     n_alignment: int = FP4_GEMM_ALIGNMENT,
     k_alignment: int = FP4_GEMM_ALIGNMENT,
 ) -> tuple[torch.Tensor, int]:
     """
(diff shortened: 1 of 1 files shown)


python/sglang/srt/layers/quantization/modelopt_quant.py
```diff
@@ -1174,7 +1269,14 @@ def process_weights_after_loading(self, layer: torch.nn.Module) -> None:
 
             layer.weight_scale_interleaved = Parameter(scale, requires_grad=False)
             layer.weight = Parameter(weight, requires_grad=False)
+            layer.weights_padding_cols = 0
             return
+
+        # Pad weights for CUTLASS/FlashInfer kernel alignment (K and N divisible by 32)
```
b8zhong, line 1317: QQ: if this comment is accurate, we also need to pad it for the `trtllm` backend? should we also do it under the block above

b8zhong (commented):
Thanks. Btw, what models with weird hidden dim are there issues with? (guess: nemotron? Just curious). Because recently we fixd something related to GLM scale divisibility as well

b8zhong (approved):

b8zhong:
Hi @danielafrimi can you fix the merge conflicts

PR #17322, opened 2026-01-19
[Bugfix] Improve error message for dynamic LoRA loading incompatibility

karanb192:
## Summary
- Fixes issue where dynamic LoRA loading fails with a generic error message that doesn't help users understand what's wrong
- Adds a new `get_incompatibility_reason()` method to `LoRAMemoryPool` that provides detailed diagnostics
- Updates `validate_new_adapter()` to use the new method and show specific failure reasons

## Problem
When dynamically loading a LoRA adapter fails due to incompatibility with the memory pool configuration, users see a generic error:
```
LoRA adapter xxxx with rank 8 is incompatible with the current LoRA memory pool configuration. 
Please ensure that the LoRA adapter's rank is within the configured `--max-lora-rank` and that 
the target modules are included in `--lora-target-modules`.
```

This doesn't tell users which specific check failed, making it hard to diagnose and fix the issue.

## Solution
The error message now provides specific details about which validation check failed:
1. **LoRA rank**: "LoRA rank 8 exceeds the configured maximum rank 4. Please increase `--max-lora-rank` to at least 8."
2. **Added tokens**: "LoRA adapter requires 100 added tokens, but the memory pool was initialized with support for only 0 added tokens..."
3. **Target modules**: "LoRA adapter contains target modules ['embed_tokens'] that are not supported by the current memory pool configuration..."

## Test plan
- [x] Verified pre-commit checks pass
- [x] Syntax validation passed
- [ ] Manual testing with dynamic LoRA loading scenarios
- [ ] CI tests

Fixes #17096

### DIFF
--- a/python/sglang/srt/lora/mem_pool.py
+++ b/python/sglang/srt/lora/mem_pool.py
@@ -122,6 +122,38 @@ class LoRAMemoryPool:
         else:
             return all(_can_support(x) for x in config)
 
+    def get_incompatibility_reason(self, config: LoRAConfig) -> Optional[str]:
+        """
+        Get a detailed reason why a LoRA adapter is incompatible with the memory pool.
+        Returns None if the adapter is compatible.
+        """
+        if config.r > self.max_lora_rank:
+            return (
+                f"LoRA rank {config.r} exceeds the configured maximum rank {self.max_lora_rank}. "
+                f"Please increase `--max-lora-rank` to at least {config.r}."
+            )
+
+        if config.lora_added_tokens_size > self.lora_added_tokens_size:
+            return (
+                f"LoRA adapter requires {config.lora_added_tokens_size} added tokens, but the memory pool "
+                f"was initialized with support for only {self.lora_added_tokens_size} added tokens. "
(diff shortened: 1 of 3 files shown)


karanb192:
Hi maintainers! Could you please add the `run-ci` label to trigger CI? The pre-commit checks passed locally. Thank you!

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -1106,7 +1106,8 @@ async def _wait_one_response(
                 )
             except asyncio.TimeoutError:
                 if (
-                    request is not None
+                    not is_stream
```
Fridge003, line 1109: Why adding this?

python/sglang/srt/lora/mem_pool.py
```diff
@@ -122,6 +122,38 @@ def _can_support(config: LoRAConfig) -> bool:
         else:
             return all(_can_support(x) for x in config)
 
+    def get_incompatibility_reason(self, config: LoRAConfig) -> Optional[str]:
```
Fridge003, line 125: We can move this into `can_support`

Issue #17311, opened 2026-01-18
[Feature] [Tracking] Enable RDMA weight transfer for RL use cases


JD-ETH:
As part of [roadmap](https://github.com/sgl-project/sglang/issues/12780), several features are in the works for enabling zero-copy weight update cross hosts between training and inference instances during distributed RL training. 

<img width="1696" height="1141" alt="Image" src="https://github.com/user-attachments/assets/7591047f-ff65-41eb-9a3e-fdc805efaf77" />
Illustration of current design. A CPU-offloadable engine replica is created on each trainer rank, which receives local bucketed weight update and sends to sgl instance via TransferEngine. 

- [ ] _Review Pending_ [Expose parallelism info](https://github.com/sgl-project/sglang/pull/16860) to enable engine replica creation. 
- [ ] _WIP_ Unified Weight Mapping Interface : To allow weight mapping between incoming parameter -> model weight shard.
- [ ] _WIP_ Extending remote_weight_info and parallelism_info http query to cross node via nccl communication. 
- [ ] _Planned_ Demo implementation of the weight transfer via RDMA on external RL framework. 

@zhaochenyang20

lfopensource:
Hi team, I'm interested in working on this.

I plan to implement the 'Source Replica' (Proposal 2) architecture as outlined in the roadmap.

zhaochenyang20:
`register_memory_region_v2`: Iterates through `torch.cuda.memory.memory_snapshot()` to identify contiguous physical blocks where weights are stored, then registers these merged blocks.

Any PR that modifies memory layout (e.g., weight offloading, quantization) may break the merging logic in v2. This merging mechanism relies on weights being located within contiguous blocks of the CUDA allocator.

JD-ETH:
compared to how nccl in `update_weight_from_distributed` API, we are only sending a shard of the original size --- the part that individual SGL enging rank actually wants.

JD-ETH:
merged https://www.lmsys.org/blog/2026-04-29-p2p-update/

PR #17326, opened 2026-01-19
Parameter Mapper to convert hugging face parameter to sglang param and shard info

JD-ETH:
## Motivation
Create a unified weight name mapping interface for a couple of key models. In order to pipeline the RDMA weight update process, we need to understand which model parameter has been updated during model.load_weight and whether the entire shard has been completed. 

Basically we want to know 

`sglang_param, num_shards, shard_idx = model.load_weight(hf_tensor)`

https://github.com/sgl-project/sglang/issues/17311 
This is the second PR of the planned solution. 

## Modifications
We just want to expose the parameter mapping layer of several key models, and move that mapping logic at initialization time. We let those models extend a `Mixin` to provide an unified interface for model-agnostic mapping resolution. It does not support NextN or NPU. 

Most of the models already have `stacked_params_mapping` and `expert_params_mapping` layers exposed to them, we just move those definitions up to the __init__ . the standard kv scale remap is supported directly in the mixin interface. 
This already make Qwen models work. Additionally, two hooks were added to handle scenarios
- we want to apply a pre-transforms of the weights (like the shared experts) 
- we have special parameter direct mapping directly (like the mistral related changes on llama model) 

We currently only consider the default model loader. Currently, we are targeting the below text model classes with their model specific adaptations (if any):
 
1. Llama3: Handling additional scale remap for mistral (activation_scale, weight_scale_inv) 
2. Qwen2: no additional change
3. Qwen3: no additional change
4. Qwen3-MoE: no additional change
5. GLM4: no additional change
6. GLM4-MOE: shared routing expert in the mutation hook 
7. DeepseekV2: shared routing expert, additional scale remap (kv_scale -> attn_mqa), and a special handling of the qkv mapping. 

Created unit test for the interface module including standard mapping as well as model specific mapping logics. 

## Accuracy Tests
Only expose additional interface, no change to the logic 

## Checklist

## Review Process

6. Get approvals from [CODEOWNERS](https://github.com/sgl-project/sglang/blob/main/.github/CODEOWNERS) and other reviewers.
8. Trigger CI tests with [comments](https://docs.sglang.io/developer_guide/contribution_guide.html#how-to-trigger-ci-tests) or contact authorized users to do so.
9. After green CI and required approvals, ask Merge Oncalls to merge.


python/sglang/srt/models/llama.py
JD-ETH, line 565: note this was duplicated so deleting has no impact

python/sglang/srt/models/deepseek_common/deepseek_weight_loader.py
```diff
@@ -93,6 +93,14 @@ class DeepseekV2WeightLoaderMixin:
     pp_group: GroupCoordinator
     num_fused_shared_experts: int
 
+    def mutate_weight_preload(self, name: str) -> str:
```
JD-ETH, line 99: since currently there is only one implementation of the `DeepseekV2WeightLoaderMixin`, we are defining the `mutate` and `custom_scale_remap` hooks on the actual definition of the architecture. If this is intended to be shared, we can move it backup?

PR #17241, opened 2026-01-17
[Perf] fuse q, k norm for Flux2Attention

zminglei:
## Motivation
fuse q, k norm for Flux2Attention

## Benchmarking and Profiling
`sglang generate --model-path /shared/public/elr-models/black-forest-labs/FLUX.2-dev   --prompt "A Logo With Bold Large Text: SGL Diffusion"   --save-output`

Before:
```
[01-17 00:21:11] [DenoisingStage] average time per step: 0.4453 seconds
[01-17 00:21:11] [DenoisingStage] finished in 22.6614 seconds
[01-17 00:21:11] [DecodingStage] started...
[01-17 00:21:12] [DecodingStage] finished in 0.2300 seconds
[01-17 00:21:12] Peak GPU memory: 64.49 GB, Remaining GPU memory at peak: 75.91 GB. Components that can stay resident: ['text_encoder']
[01-17 00:21:12] Output saved to outputs/A_Logo_With_Bold_Large_Text_SGL_Diffusion_20260117-002048_b81248a0.png
[01-17 00:21:12] Pixel data generated successfully in 24.32 seconds
[01-17 00:21:12] Completed batch processing. Generated 1 outputs in 24.32 seconds
[01-17 00:21:12] Memory usage - Max peak: 66041.99 MB, Avg peak: 66041.99 MB
```
After:
```
[01-17 00:23:49] [DenoisingStage] average time per step: 0.4346 seconds
[01-17 00:23:49] [DenoisingStage] finished in 22.1209 seconds
[01-17 00:23:49] [DecodingStage] started...
[01-17 00:23:49] [DecodingStage] finished in 0.2593 seconds
[01-17 00:23:49] Peak GPU memory: 64.49 GB, Remaining GPU memory at peak: 75.91 GB. Components that can stay resident: ['text_encoder']
[01-17 00:23:50] Output saved to outputs/A_Logo_With_Bold_Large_Text_SGL_Diffusion_20260117-002326_178f3467.png
[01-17 00:23:50] Pixel data generated successfully in 23.75 seconds
[01-17 00:23:50] Completed batch processing. Generated 1 outputs in 23.76 seconds
[01-17 00:23:50] Memory usage - Max peak: 66041.99 MB, Avg peak: 66041.99 MB
```
24.32s -> 23.76s, around 2.5% e2e latency improvement

## Checklist

[...]

PR #17332, opened 2026-01-19
Revert "[Perf] fuse q, k norm for Flux2Attention (#17241)"

BBuf:
Reverts sgl-project/sglang#17241

First, not implementing a fallback when applying qk_norm will break the [AMD diffusion CI](https://github.com/sgl-project/sglang/actions/runs/21085222138/job/60680696502). Second, the decision to use qknorm should be based on whether `can_use_fused_inplace_qknorm` returns true, rather than being applied directly. Third, converting non-contiguous q and k tensors directly to contiguous here is a performance-sensitive operation and should not be done. Even though this is dead code, it should not be written this way. Finally, a warmup should be add in performace test command.

Please refer to: https://github.com/sgl-project/sglang/pull/17305

And I will refactor diffusion qk_norm apply logic to make it easier to use in the future.

cc @mickqian @yhyang201 @zminglei

### DIFF
--- a/python/sglang/multimodal_gen/runtime/layers/layernorm.py
+++ b/python/sglang/multimodal_gen/runtime/layers/layernorm.py
@@ -443,15 +443,6 @@ def apply_qk_norm(
     """
 
     batch_size = q.size(0)
-    q_shape = q.shape
-    k_shape = k.shape
-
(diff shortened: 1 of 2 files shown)


yhyang201:
OK, I’ll pay attention.

PR #17334, opened 2026-01-19
[FIX] fix mambaish model pp kv cache compute

GuoweiWangU:
## Motivation

In the Mambaish model, the calculation of `max_new_tokens` considers all full attention layers, without taking into account the scenario in which, under parallel processing (PP), only the effective layers need to be considered.

## Modifications
` 
def profile_max_num_token():
    ...       
    elif mambaish := self.mambaish_config:
            effective_layer_ids = [
                i
                for i in mambaish.full_attention_layer_ids
                if self.start_layer <= i < self.end_layer
            ]
            num_layers = len(effective_layer_ids)`

## Checklist

## Review Process

### DIFF
--- a/python/sglang/srt/model_executor/model_runner_kv_cache_mixin.py
+++ b/python/sglang/srt/model_executor/model_runner_kv_cache_mixin.py
@@ -124,7 +124,12 @@ class ModelRunnerKVCacheMixin:
                 self.num_effective_layers,
             )
         elif mambaish := self.mambaish_config:
-            num_layers = len(mambaish.full_attention_layer_ids)
+            effective_layer_ids = [
+                i
+                for i in mambaish.full_attention_layer_ids
+                if self.start_layer <= i < self.end_layer
+            ]
+            num_layers = len(effective_layer_ids)
         else:
             num_layers = self.num_effective_layers
 


python/sglang/srt/model_executor/model_runner_kv_cache_mixin.py
```diff
@@ -124,7 +124,12 @@ def profile_max_num_token(self: ModelRunner, total_gpu_memory: int):
                 self.num_effective_layers,
             )
         elif mambaish := self.mambaish_config:
-            num_layers = len(mambaish.full_attention_layer_ids)
+            effective_layer_ids = [
```
ispobock, line 127: Maybe we can have a more general way to handle pp? 
ref: https://github.com/sgl-project/sglang/blob/8916b9d080acec83c23f1d3c8f5d74d1738ebc8f/python/sglang/srt/model_executor/model_runner.py#L504
cc: @yizhang2077

GuoweiWangU: I‘ll try that

GuoweiWangU: `full_attention_layer_ids` in `model_runner.full_attention_layer_ids` only works for `is_hybrid_swa_model`, in other cases, `full_attention_layer_ids` will be empty. Therefore, a more general approach requires other mambaish models to be adapted to the computation of `full_attention_layers`.

ispobock (approved):

PR #17353, opened 2026-01-19
Move fa4 from sgl-kernel to jit kernel

BBuf:
## Motivation

Flash Attention V4 is implemented using the cute-dsl and is inherently JIT-compiled, so moving it to jit_kernel is a fairly natural transition. 

`python/sglang/jit_kernel/flash_attention/cute` directory is copied from https://github.com/Dao-AILab/flash-attention/tree/main/flash_attn/cute , so this directory do not need review.

图片

unit-test passed.

## Checklist

## Review Process


zhyncs (changes requested):
This change looks good. In the future, we can evolve FA4 independently in the JIT kernel.

Given that the new FA4 has a different import path, the existing FA4 in sgl-kernel does not need to be removed.

Since sgl-kernel is depended on by other projects—and some users may not rely on PyPI installs—keeping the original FA4 is important for backward compatibility. Please avoid removing it, and apply the same approach to future kernel migrations.

BBuf:
> This change looks good. In the future, we can evolve FA4 independently in the JIT kernel.
> 
> Given that the new FA4 has a different import path, the existing FA4 in sgl-kernel does not need to be removed.
> 
> Since sgl-kernel is depended on by other projects—and some users may not rely on PyPI installs—keeping the original FA4 is important for backward compatibility. Please avoid removing it, and apply the same approach to future kernel migrations.

Flash Attention V4 is implemented purely in Python using the cute-dsl and can run independently of flash-attention and sgl-kernel. The removal of sgl-kernel's fa4 here is also to avoid being constrained by sgl-kernel during future upgrades of fa4. Additionally, I inquired about the release plan for this Python package in [here](https://github.com/Dao-AILab/flash-attention/issues/2193). If a Python package is released later, it could avoid the need to copy the `python/sglang/jit_kernel/flash_attention/cute` directory in jit_kernel. The motivation behind this work is that we will not update fa4 or release new versions within sgl-kernel. Even if it were retained as you suggested, it would only leave behind an outdated interface.

merrymercy (approved):

johnnynunez:
> > I think that at some point, we can move everything to cute dsl, so It would be nice to remove C++ FA from sglang
> > It saves time compilation == Less pressure CI == reduce costs.
> > FA cute dsl it has FA2 and FA3(but i don't know if it has all the features)
> > Now i close to cutlass team, so i'll divulgate the future changes
> 
> Indeed, this can significantly reduce the size of the sgl-kernel package.

yes, and we can incorporate more archs.

zhyncs (approved):

PR #17369, opened 2026-01-19
[Fix] Revert back to using CUTLASS `mm_fp4` backend

b8zhong:
## Motivation

After manually installing `nvidia-cudnn-cu13==9.18.0.77`, it still seems to be worse (`mm_fp4` picks this when cudnn >= 9.15.xx and CUDA 13, so hold off on this change for now (otherwise API will pick incorrect backend)

image

Issue reported in https://github.com/flashinfer-ai/flashinfer/issues/2375

## Modifications

Change it back to earlier behaviour.


Fridge003 (approved):

Fridge003 (approved):

Fridge003:
B200 fp4 gemm tests passed

PR #17373, opened 2026-01-20
refactor Qwen3-Next with a new RadixLinearAttention

zminglei:
## Motivation
Refactor: Abstract linear attention backend calls through `RadixLinearAttention`

## Modifications

- `qwen3_next.py`: Replaced direct forward_batch.attn_backend.forward() calls with RadixLinearAttention interface

- `hybrid_linear_attn_backend.py`: Backend now extracts layer parameters (conv_weights, A_log, dt_bias, bias, activation, dimensions) from the layer object instead of kwargs

Init-time params stored on layer: `conv_weights, bias, A_log, dt_bias, activation, head dimensions`
Runtime tensors passed via kwargs: `mixed_qkv, a, b, z`

## Accuracy Tests

```
python3 -m sglang.launch_server --model-path /shared/public/elr-models/Qwen/Qwen3-Next-80B-A3B-Instruct/9c7f2fbe84465e40164a94cc16cd30b6999b0cc7/ --tp 4

python benchmark/gsm8k/bench_sglang.py --data-path /shared/public/data/gsm8k/test.jsonl
100%|████████████████████████████████████████████████████████████████████████████████████████████████████| 200/200 [00:20<00:00,  9.57it/s]
Accuracy: 0.960
Invalid: 0.000
Latency: 21.029 s
Output throughput: 1522.310 token/s
```

With piecewise cuda graph:
```
python3 -m sglang.launch_server --model-path /shared/public/elr-models/Qwen/Qwen3-Next-80B-A3B-Instruct/9c7f2fbe84465e40164a94cc16cd30b6999b0cc7/ --tp 4 --enable-piecewise-cuda-graph

python benchmark/gsm8k/bench_sglang.py --data-path /shared/public/data/gsm8k/test.jsonl
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 200/200 [00:14<00:00, 13.86it/s]
Accuracy: 0.955
Invalid: 0.000
Latency: 14.492 s
Output throughput: 2261.951 token/s
```

## Checklist

## Review Process


python/sglang/srt/layers/attention/hybrid_linear_attn_backend.py
```diff
@@ -864,20 +864,23 @@ def forward_decode(
...
-        attn_tp_size = kwargs["attention_tp_size"]
-        head_k_dim = kwargs["head_k_dim"]
-        head_v_dim = kwargs["head_v_dim"]
         a = kwargs["a"]
         b = kwargs["b"]
-        A_log = kwargs["A_log"]
-        dt_bias = kwargs["dt_bias"]
-        layer_id = kwargs["layer_id"]
+
+        if layer is not None:
```
hebiao064: is it possible that layer is none?

zminglei: No, layer should always be passed as a non-None value from the RadixLinearAttention, I've removed the check here to avoid confusion.

python/sglang/srt/layers/attention/hybrid_linear_attn_backend.py
```diff
@@ -831,25 +832,27 @@ def forward_decode(
         q: torch.Tensor,
         k: torch.Tensor,
         v: torch.Tensor,
-        layer: RadixAttention,
+        layer: RadixLinearAttention,
```
hebiao064, line 832: change this interface to be

```
    def forward_decode(
        self,
        layer: RadixLinearAttention,
        forward_batch: ForwardBatch,
        mixed_qkv: torch.Tensor,
        a: torch.Tensor,
        b: torch.Tensor,
    )
```

hebiao064: the goal is to remove all kwargs

hebiao064: for the hybrid linear attention backend, we can do 
```
 def forward(
        self,
        layer: RadixLinearAttention,
        forward_batch: ForwardBatch,
        save_kv_cache: bool = True,
        q: Optional[torch.Tensor] = None,  # For full attention
        k: Optional[torch.Tensor] = None,  # For full attention
        v: Optional[torch.Tensor] = None,  # For full attention
        mixed_qkv: Optional[torch.Tensor] = None,  # For Linear Attention
        a: Optional[torch.Tensor] = None,  # For Linear Attention
        b: Optional[torch.Tensor] = None,  # For Linear Attention
        **kwargs,
    )
```

python/sglang/srt/layers/attention/hybrid_linear_attn_backend.py
```diff
@@ -909,26 +912,30 @@ def forward_extend(
...
-        head_v_dim = kwargs["head_v_dim"]
         a = kwargs["a"]
         b = kwargs["b"]
-        A_log = kwargs["A_log"]
-        dt_bias = kwargs["dt_bias"]
-        layer_id = kwargs["layer_id"]
-        seq_len = kwargs["seq_len"]
+
+        # seq_len can be derived from mixed_qkv or passed explicitly
+        seq_len = kwargs.get("seq_len", mixed_qkv.shape[0])
```
hebiao064: we don't need to pass it from outside, we can just do `seq_len = mixed_qkv.shape[0]`

python/sglang/srt/layers/attention/hybrid_linear_attn_backend.py
```diff
@@ -909,26 +912,30 @@ def forward_extend(
...
+        bias = layer.bias
+        activation = layer.activation
+        key_dim = layer.key_dim
+        value_dim = layer.value_dim
+        attn_tp_size = layer.attention_tp_size
+        head_k_dim = layer.head_k_dim
+        head_v_dim = layer.head_v_dim
+        A_log = layer.A_log
+        dt_bias = layer.dt_bias
+        layer_id = layer.layer_id
```
hebiao064: some of the variables are only being used once, we don't need to reassign to a new value, for example: `A_log`

python/sglang/srt/models/qwen3_next.py
```diff
@@ -305,6 +304,25 @@ def __init__(
             prefix=add_prefix("out_proj", prefix),
         )
 
+        # Create the 2D view of conv weights once at init
+        conv_weights = self.conv1d.weight.view(
```
hebiao064: do we need this?

python/sglang/srt/layers/attention/hybrid_linear_attn_backend.py
zminglei: updated it, ptal

python/sglang/srt/layers/attention/hybrid_linear_attn_backend.py
zminglei: updated

python/sglang/srt/layers/attention/hybrid_linear_attn_backend.py
zminglei: updated

python/sglang/srt/models/qwen3_next.py
zminglei: Yes, it's needed inside the GDN attention, I replaced it with a squeeze op.

hebiao064 (approved):

yizhang2077 (approved):

yizhang2077:
Could we also move other hybrid models like kimi linear into `RadixLinearAttention` later?

hebiao064 (approved):

python/sglang/srt/layers/radix_linear_attention.py
```diff
@@ -0,0 +1,83 @@
...
+from typing import TYPE_CHECKING, Optional
+
+import torch
+from torch import nn
+
+if TYPE_CHECKING:
+    from sglang.srt.model_executor.forward_batch_info import ForwardBatch
+
+
+class RadixLinearAttention(nn.Module):
```
ispobock: Can it also be applied to other linear models like kimi-linear, nemotron?

ispobock (approved):

python/sglang/srt/layers/attention/hybrid_linear_attn_backend.py
```diff
@@ -828,30 +829,23 @@ def __init__(self, model_runner: ModelRunner):
...
-
-        layer_cache = self.req_to_token_pool.mamba2_layer_cache(layer_id)
+        conv_weights = layer.conv_weights
+        bias = layer.bias
+        activation = layer.activation
+        key_dim = layer.key_dim
+        value_dim = layer.value_dim
+        attn_tp_size = layer.attention_tp_size
+        head_k_dim = layer.head_k_dim
+        head_v_dim = layer.head_v_dim
```
yuan-luo, line 846: Suggest also add:
```
A_log = layer.A_log
dt_bias = layer.dt_bias
```
to make the parameter list consistent.

python/sglang/srt/layers/attention/hybrid_linear_attn_backend.py
```diff
@@ -1024,7 +1012,7 @@ def forward_extend(
         key = key.view(1, actual_seq_len, num_heads, head_k_dim)
         value = value.view(1, actual_seq_len, num_value_heads, head_v_dim)
 
-        g, beta = fused_gdn_gating(A_log, a, b, dt_bias)
+        g, beta = fused_gdn_gating(layer.A_log, a, b, layer.dt_bias)
```
yuan-luo, line 1015: ditto, in that way, we can avoid to use layer.A_log style here and there.

python/sglang/srt/models/qwen3_next.py
```diff
@@ -419,41 +432,12 @@ def _forward(
...
-            "head_v_dim": self.head_v_dim,
-            "a": a,
-            "b": b,
-            "A_log": self.A_log,
-            "dt_bias": self.dt_bias,
-            "layer_id": self.layer_id,
-            "seq_len": seq_len,
-            "num_k_heads": self.num_k_heads,
-            "num_v_heads": self.num_v_heads,
-            "z": z,
```
yuan-luo, line 447: @zminglei I realized here `z` was not passed to AttentionBackend. In Qwen3-Next model, `z`'s calculation rely on hidden_states, it is neither passed by kwargs in forward(kwargs is gone) nor can be deduced in RadixLinearAttention initialization as there's no hidden_states. So this value is lost. 
When refactoring Kimi-Linear to reuse this RadixLinearAttention arch I also encountered this circumstance. There are more parameters depend on hidden_states, but there's no way to pass it as kwargs is wiped off. I think this is an issue in this architecture. The kwargs seems to be inevitable. The parameters which can be calculated can be removed from kwargs and put them into LinearAttention initialization phase.
cc: @ispobock @yizhang2077 @hebiao064 

After double confirming the code, GDN and KDA can use the same parameters:
z: norm_gate
a: forget_gate
b: beta
The z don't need to pass into AttentionBackend. z's shape can be deduced from other parameters.

This code refactor is solid.

yuan-luo (approved):

python/sglang/srt/models/qwen3_next.py
zminglei: Yes, the z is not used in the attention_backend, only its shape was used, and the shape could be deduced from other parameters like `mixed_qkv`, so we're not passing z in forward here.

python/sglang/srt/layers/attention/hybrid_linear_attn_backend.py
zminglei: For the variables which are only used once, we didn't reassign them to a new value/variable. See [comment](https://github.com/sgl-project/sglang/pull/17373#discussion_r2709841444).

python/sglang/srt/layers/attention/hybrid_linear_attn_backend.py
zminglei: same as above.

PR #17374, opened 2026-01-20
 [4/N] (Elastic EP) Back up Expert Weights in DRAM

ympcMark:
## Motivation
Currently, in Elastic EP, when rank failures or expert weight loss occur, the system must reload missing weights from disk. This process significantly increases service interruption time.

We observed that servers often have unused spare DRAM capacity. Leveraging this available DRAM, we propose maintaining a hot backup of expert weights in memory. In the event of a failure, weights can be restored directly from DRAM over the high-speed RDMA network, updating the model weights in GPU memory much faster than disk-based recovery.

Experimental results show that this approach outperforms the original disk-based method—even when the original already employs optimizations such as multi-threading and OS page cache.

Furthermore, in multi-node deployment scenarios, each node only needs to store a subset of expert weights, significantly reducing per-node DRAM overhead.

Our changes are incremental and controlled by a server argument that is disabled by default, ensuring no impact on existing deployments.

## Modifications
- **expert_backup_manager.py**: Introduced a new `ExpertBackupManager` class and process responsible for loading a subset of expert weights from disk into a continuous CPU buffer and managing their transfer via Mooncake.

- **expert_backup_client.py**: Implemented ZMQ-based communication between the `ExpertBackupClient` and the `ExpertBackupManager` to coordinate weight backup requests and transfer expert weight pointer information and session IDs across distributed nodes. Then initialize a client-side TransferEngine and dynamically load expert weights from DRAM when needed.

- **io_struct.py**: `UpdateExpertBackupReq` and `BackupDramReq`.
- **engine.py**: Launch `ExpertBackupManager`.
- **test_expert_backup.py**: Add unit tests.

## Checklist

## Review Process


python/sglang/srt/elastic_ep/expert_backup_manager.py
```diff
@@ -0,0 +1,182 @@
...
+        from mooncake.engine import TransferEngine
+
+        self.transfer_engine = TransferEngine()
+        self.transfer_engine.initialize(
+            get_local_ip_auto(),
+            "P2PHANDSHAKE",
+            "rdma",
+            self.server_args.mooncake_ib_device,
+        )
+        self.session_id = f"{get_local_ip_auto()}:{self.transfer_engine.get_rpc_port()}"
```
ShangmingCai: Maybe replace this with `get_mooncake_transfer_engine()`, and add the ep usage here: https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/model_executor/model_runner.py#L858-L875

python/sglang/srt/elastic_ep/expert_backup_manager.py
```diff
@@ -0,0 +1,182 @@
...
+    if match:
+        return int(match.group(1))
+    return -1
+
+
+class ExpertBackupManager:
+    def __init__(self, server_args: ServerArgs, port_args: PortArgs):
+        self.server_args = server_args
+        self.port_args = port_args
+        self.model_path = server_args.model_path
```
ShangmingCai: Has this param been used in the backup manager?

python/sglang/srt/elastic_ep/expert_backup_manager.py
```diff
@@ -0,0 +1,182 @@
...
+        self.load_format = server_args.load_format
+        self.model_config = ModelConfig.from_server_args(server_args)
+        self.continuous_buffer = None
+        self.weight_pointer_map = {}
+        self.transfer_engine = None
```
ShangmingCai, line 40: use shared mooncake TE here

python/sglang/srt/elastic_ep/expert_backup_manager.py
```diff
@@ -0,0 +1,182 @@
...
+            "rdma",
+            self.server_args.mooncake_ib_device,
+        )
+        self.session_id = f"{get_local_ip_auto()}:{self.transfer_engine.get_rpc_port()}"
+        server_ptr = self.continuous_buffer.data_ptr()
+        server_len = (
+            self.continuous_buffer.numel() * self.continuous_buffer.element_size()
+        )
+
+        ret_value = self.transfer_engine.register_memory(server_ptr, server_len)
```
ShangmingCai: If `server_ptr` and `server_len` won't change, we should register only once. If not, I think we should deregister it after transfer completion.

python/sglang/srt/elastic_ep/expert_backup_manager.py
```diff
@@ -0,0 +1,182 @@
...
+        if ret_value != 0:
+            raise RuntimeError("Mooncake memory registration failed.")
+
+    def event_loop(self):
+        while True:
+            while True:
```
ShangmingCai: Why double `While True`?

typo?

python/sglang/srt/elastic_ep/expert_backup_client.py
```diff
@@ -0,0 +1,180 @@
...
+    return -1, -1, ""
+
+
+class ExpertBackupClient:
+    def __init__(self, server_args: ServerArgs, model_runner):
+        context = zmq.Context(2)
+        self.send_to_backup_manager = get_zmq_socket(
+            context,
+            zmq.PUSH,
+            f"tcp://127.0.0.1:{10000 + server_args.node_rank * 2}",
```
ShangmingCai: This setup doesn't seem robust to me

python/sglang/srt/elastic_ep/expert_backup_manager.py
```diff
@@ -0,0 +1,182 @@
...
+        self.engine_num = server_args.nnodes
+        self.engine_rank = server_args.node_rank
+        self.expert_num = self.model_config.hf_config.n_routed_experts
+        self.idmn = (self.expert_num // self.engine_num) * self.engine_rank
+        self.idmx = (self.expert_num // self.engine_num) * (self.engine_rank + 1)
+        context = zmq.Context(2)
+        self.recv_from_expert_backup_client = get_zmq_socket(
+            context,
+            zmq.PULL,
+            f"tcp://127.0.0.1:{10000 + server_args.node_rank * 2}",
```
ShangmingCai: What will happen when `dist_init_addr` has been used for multi-node cases?

python/sglang/srt/elastic_ep/expert_backup_client.py
```diff
@@ -0,0 +1,180 @@
...
+    def start_transfer_client(self):
+        from mooncake.engine import TransferEngine
+
+        self.transfer_engine = TransferEngine()
+        self.transfer_engine.initialize(
+            get_local_ip_auto(),
+            "P2PHANDSHAKE",
+            "rdma",
+            self.server_args.mooncake_ib_device,
+        )
```
ShangmingCai: ditto

python/sglang/srt/elastic_ep/expert_backup_client.py
```diff
@@ -0,0 +1,180 @@
...
+                    response = self.recv_list[i].recv_pyobj()
+                except zmq.ZMQError:
+                    continue
+                self.dram_map_list[response._rank] = response._map
+                self.session_id_list[response._rank] = response.session_id
+                self.buffer_size = max(self.buffer_size, response.buffer_size)
+                cnt += 1
+                if cnt == self.engine_num:
+                    self.use_backup = True
+                    self.start_transfer_client()
```
ShangmingCai: This means we repeatedly create transfer_engine and register addr every time. Please figure out what should be only done once, at least transfer_engine should not be initialized over and over again.

python/sglang/srt/managers/io_struct.py
```diff
@@ -1329,6 +1323,19 @@ class SendWeightsToRemoteInstanceReqOutput(BaseReq):
...
 
+@dataclass
+class UpdateExpertBackupReq(BaseReq):
+    pass
+
+
+@dataclass
+class BackupDramReq(BaseReq):
+    _rank: int
+    _map: Dict[str, Any]
```
ShangmingCai: Also, I think these two are not private.


```suggestion
    rank: int
    weight_pointer_map: Dict[str, Any]
```

python/sglang/srt/elastic_ep/expert_backup_manager.py
```diff
@@ -0,0 +1,182 @@
...
+        server_ptr = self.continuous_buffer.data_ptr()
+        server_len = (
+            self.continuous_buffer.numel() * self.continuous_buffer.element_size()
+        )
+
+        ret_value = self.transfer_engine.register_memory(server_ptr, server_len)
+        if ret_value != 0:
+            raise RuntimeError("Mooncake memory registration failed.")
+
+    def event_loop(self):
```
ShangmingCai: Make this a blocking loop or add a interval to avoid occupying the cpu resources.

python/sglang/srt/elastic_ep/expert_backup_client.py
ympcMark: in our design, manager and client's TE will be created only once, maybe we should break the loop after start engine?

python/sglang/srt/elastic_ep/expert_backup_manager.py
ympcMark: yes, i will delete it!

python/sglang/srt/elastic_ep/expert_backup_manager.py
ympcMark: you are right!

python/sglang/srt/managers/io_struct.py
ympcMark: OK

python/sglang/srt/elastic_ep/expert_backup_client.py
ShangmingCai: also at least make `10000` a configurable env var

ympcMark: we find it redundant, so we should lauch TE directly without communication!

python/sglang/srt/elastic_ep/expert_backup_manager.py
ympcMark: it is redundant, and we will skip it and start immediately

python/sglang/srt/elastic_ep/expert_backup_client.py
ympcMark: we also will change it to a blocking loop

docs/advanced_features/server_arguments.md
```diff
@@ -330,6 +330,7 @@ Please consult the documentation below and [server_args.py](https://github.com/s
 | `--deepep-config` | Tuned DeepEP config suitable for your own cluster. It can be either a string with JSON content or a file path. | `None` | Type: str |
 | `--moe-dense-tp-size` | TP size for MoE dense MLP layers. This flag is useful when, with large TP size, there are errors caused by weights in MLP layers having dimension smaller than the min dimension GEMM supports. | `None` | Type: int |
 | `--elastic-ep-backend` | Specify the collective communication backend for elastic EP. Currently supports 'mooncake'. | `none` | `none`, `mooncake` |
+| `--enable-elastic-expert-backup` | Enable elastic expert backup feature. | `False` | bool flag (set to enable) |
```
ch-wan: the explanation here carries zero information

python/sglang/srt/elastic_ep/expert_backup_client.py
```diff
@@ -0,0 +1,169 @@
...
+        self.transfer_engine = get_mooncake_transfer_engine()
+
+        self.params_dict = dict(self.model_runner.model.named_parameters())
+        for name, param in self.params_dict.items():
+            param_data = param.data
+            ret_value = self.transfer_engine.engine.register_memory(
+                param_data.data_ptr(), param_data.numel() * param_data.element_size()
+            )
+            if ret_value != 0:
+                raise RuntimeError("GPU buffer memory registration failed.")
```
ch-wan: Can we avoid engine crash for this resilient feature?

python/sglang/srt/elastic_ep/expert_backup_client.py
```diff
@@ -0,0 +1,169 @@
...
+            self.buffer_size = max(self.buffer_size, response.buffer_size)
+            cnt += 1
+
+        self.use_backup = True
+        self.start_transfer_client()
+
+    def start_transfer_client(self):
+        from sglang.srt.distributed.parallel_state import get_mooncake_transfer_engine
+
+        self.transfer_engine = get_mooncake_transfer_engine()
```
ch-wan, line 88: should we check if the elastic ep backend is mooncake?

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -872,6 +880,7 @@ def init_shared_mooncake_transfer_engine(self):
                 self.server_args.language_only
                 and self.server_args.encoder_transfer_backend == "mooncake"
             )
+            or self.server_args.enable_elastic_expert_backup
```
ch-wan: I feel that we should check `elastic_ep_backend` here instead

docs/advanced_features/server_arguments.md
ympcMark: Yes, we will make it clear to understand

python/sglang/srt/elastic_ep/expert_backup_client.py
ympcMark: according to Shangming's review: https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/model_executor/model_runner.py#L858-L875, maybe it still runs when backend is not mooncake?

python/sglang/srt/model_executor/model_runner.py
ympcMark: yes！

python/sglang/srt/elastic_ep/expert_backup_client.py
ympcMark: yes!

python/sglang/srt/elastic_ep/expert_backup_client.py
ympcMark: we will avoid using backup when failing

python/sglang/srt/model_executor/model_runner.py
ShangmingCai: Does it only work for mooncake ep backend? need to make sure if it is general or we need to check the elastic ep backend == mooncake

python/sglang/srt/elastic_ep/expert_backup_client.py
```diff
@@ -0,0 +1,169 @@
...
+        all_ips = [None] * get_world_size()
+        torch.distributed.all_gather_object(
+            all_ips, local_ip, group=get_world_group().cpu_group
+        )
+        logger.info(f"all_ips: {all_ips}")
+
+        for i in range(self.engine_num):
+            self.recv_list[i] = context.socket(zmq.SUB)
+            self.recv_list[i].connect(
+                f"tcp://{all_ips[i * get_world_size() // server_args.nnodes]}:{PORT_BASE + i * 2 + 1}"
```
ShangmingCai, line 59: Looks a little bit hacky. But you can optimize this in the future.

python/sglang/srt/model_executor/model_runner.py
ympcMark: it is general, precisely we check elastic ep backend is not None

ShangmingCai (approved):
Image

Failed AMD tests (lora, tool call parser) are irrelevant.

PR #17389, opened 2026-01-20
Fix remote weight info nnode>1 and dp>1

JD-ETH:
## Motivation

Cross-Node Scheduler Info Synchronization for Remote Weight Loading
**Problem**
In multi-node enabled setups, each node only knows about its local schedulers' transfer engine info. Remote instances querying /get_remote_instance_transfer_engine_info?rank=N could fail if rank N is on a different node.

In dp>1 case, the dp_controller currently does not propagate the scheduler_info upwards. 

## Solution
Complete refactor of how engine info is shared with http server. instead of piping through 
**ModelRunner -> Scheduler -> Engine -> HTTP Server -> Caller** 
We are simply doing 
**ModelRunner -> BootStrap Server -> Consumer** 

## Modifications

Following PD aggregation server [design](https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/disaggregation/common/conn.py#L237). 
1. Added a `EngineInfoBootstrapServer` to receive per-rank engine information. 
2. `ModelRunner` to register at the server after scheduler subprocess runs. 
3. **completely removing the old path of transfer engine info propagation**. 

## Accuracy Tests
```
# Multi-node TP=4 test
Terminal 1: python test/manual/test_cross_node_scheduler_info_sync.py --test-case tp4_nodes2 --node-rank 0
Terminal 2: python test/manual/test_cross_node_scheduler_info_sync.py --test-case tp4_nodes2 --node-rank 1
Terminal 3: python test/manual/test_cross_node_scheduler_info_sync.py --test-case tp4_nodes2 --test-only

# Single-node DP attention test  
Terminal 1: python test/manual/test_cross_node_scheduler_info_sync.py --test-case dp2_single_node --node-rank 0
Terminal 2: python test/manual/test_cross_node_scheduler_info_sync.py --test-case dp2_single_node --test-only

# Multi-node TP=4 with DP attntetion test
Terminal 1: python test/manual/test_cross_node_scheduler_info_sync.py --test-case dp2_tp2_nodes2 --node-rank 0
Terminal 2: python test/manual/test_cross_node_scheduler_info_sync.py --test-case dp2_tp2_nodes2 --node-rank 1
Terminal 3: python test/manual/test_cross_node_scheduler_info_sync.py --test-case dp2_tp2_nodes2 --test-only
```

## Checklist

## Review Process


python/sglang/srt/entrypoints/engine.py
```diff
@@ -1172,10 +1187,58 @@ def _wait_for_scheduler_ready(
...
+    try:
+        pg = StatelessProcessGroup.create(
+            host=dist_host,
+            port=sync_port,
+            rank=server_args.node_rank,
+            world_size=server_args.nnodes,
+            timeout_seconds=SYNC_TIMEOUT_SECONDS,
+        )
+
+        all_node_infos = pg.all_gather_obj(local_scheduler_infos)
```
amysaq2023: I understand that each node maintains a complete copy of the scheduler information for all other nodes, while only the master node's copy is actively used via http server. I was wondering—would the copies on the other nodes potentially be redundant in this case?

amysaq2023: I was just curious about the time cost of the all-gather operation through the Gloo backend, given that it lies on the boot-up critical path. I was wondering if this might introduce any noticeable latency during startup.

JD-ETH: they are redundant, yes --- but as we are relying on the collectives, this is the most straightforward.

JD-ETH: I have not independently ablated this. But from my observations, more startup time is spent on 
- disk loading (if required) 
- RDMA registration (if required) 

gloo initialization itself should be fast, as there is no nccl involved that does warmup and initialization. I can try measuring this and see if it's observable.

python/sglang/srt/managers/data_parallel_controller.py
```diff
@@ -590,6 +591,7 @@ def run_data_parallel_controller_process(
                 "status": "ready",
                 "max_total_num_tokens": controller.max_total_num_tokens,
                 "max_req_input_len": controller.max_req_input_len,
+                "_dp_scheduler_infos": controller.scheduler_infos,
```
hnyls2002: Explain why using `_dp_scheduler_infos`. It is not dp-scheduler but just collects scheduler info from different dp ranks?

JD-ETH: yes --- previously scheduler_infos from different `dp_rank` are also not synced.

python/sglang/srt/entrypoints/engine.py
```diff
@@ -1172,10 +1187,58 @@ def _wait_for_scheduler_ready(
             raise RuntimeError(
                 "Initialization failed. Please see the error messages above."
             )
-        scheduler_infos.append(data)
+
+        if "_dp_scheduler_infos" in data:
+            scheduler_infos.extend(data["_dp_scheduler_infos"])
```
JD-ETH: they are collected here @amysaq2023

python/sglang/srt/entrypoints/engine.py
JD-ETH: the gathered info for non-lead node ranks is discarded

python/sglang/srt/entrypoints/engine.py
JD-ETH: so each node collects dp_scheduler infos first, if they exist, and send to nnode rank 0 after.

python/sglang/srt/entrypoints/engine.py
JD-ETH: the TCP store trip time is 0.5 sec on my experiment, and the sync is free (<10ms). So this should scale well.

ShangmingCai (commented):
I think the main goal is to make `remote_instance_transfer_engine_info` have all the ranks' info. Is that right?

If so, we can do it in a cleaner way, you can impl a route entry as we did in the PD module:
https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/disaggregation/common/conn.py#L781-L841

Then call a register func in the ModelRunner after transfer engine init, something as we did here:
https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/disaggregation/common/conn.py#L237C9-L237C30

So we don't need to pass this info into the engine and create a temporary group, and this solution will be more robust for future extension.
CC:@amysaq2023, what do you think?

ShangmingCai (commented):
This change is big, maybe I put it wrong earlier. Can we just impl a simple version like this?
```python
@app.post("/register_remote_instance_transfer_engine_info")
@auth_level(AuthLevel.ADMIN_OPTIONAL)
async def register_remote_instance_transfer_engine_info(raw_request: Request):
    # handle _global_state.remote_instance_transfer_engine_info here
```

python/sglang/srt/entrypoints/engine.py
```diff
@@ -622,10 +622,22 @@ def _launch_subprocesses(
...
 
+        # Start the engine info bootstrap server if per-rank info is needed.
+        engine_info_bootstrap = None
+        if (
+            server_args.remote_instance_weight_loader_start_seed_via_transfer_engine
+            and server_args.node_rank == 0
+        ):
+            bootstrap_port = server_args.engine_info_bootstrap_port
+            engine_info_bootstrap = EngineInfoBootstrapServer(
+                host="0.0.0.0", port=bootstrap_port
```
ShangmingCai: Better use `server_args.host` in case people who are working on security make this a CVE.

python/sglang/srt/entrypoints/engine.py
```diff
@@ -622,10 +622,22 @@ def _launch_subprocesses(
             port_args = PortArgs.init_new(server_args)
         logger.info(f"{server_args=}")
 
+        # Start the engine info bootstrap server if per-rank info is needed.
+        engine_info_bootstrap = None
```
ShangmingCai: Use `engine_info_bootstrap_server` or `engine_info_server`, it is a server, not one bootstrap process.

python/sglang/srt/entrypoints/engine_info_bootstrap_server.py
```diff
@@ -0,0 +1,101 @@
...
+from typing import Dict, Optional, Tuple
+
+import uvicorn
+from fastapi import FastAPI, HTTPException
+from fastapi.responses import PlainTextResponse
+
+logger = logging.getLogger(__name__)
+
+
+class EngineInfoBootstrapServer:
```
ShangmingCai, line 26: Need a clean up for this class,
```
def close(self):
```

python/sglang/srt/entrypoints/http_server.py
```diff
@@ -1029,23 +1017,21 @@ async def get_remote_instance_transfer_engine_info(rank: int = None):
...
-            "remote_instance_transfer_engine_info": _global_state.remote_instance_transfer_engine_info[
-                rank
-            ],
-        }
-        return result
-    except Exception as e:
-        logger.error(f"Exception: {e}")
-        return Response(status_code=HTTPStatus.BAD_REQUEST)
+        resp = requests.get(
+            f"http://127.0.0.1:{bootstrap_port}/get_transfer_engine_info",
```
ShangmingCai: Will this local IP work when `--nnode > 1`? Better check whether `--dist-init-addr` is None.

python/sglang/srt/entrypoints/http_server.py
```diff
@@ -1029,23 +1017,21 @@ async def get_remote_instance_transfer_engine_info(rank: int = None):
...
+            params={"rank": rank},
+            timeout=5,
+        )
+        if resp.status_code == 200:
+            return resp.json()
+    except Exception:
+        pass
```
ShangmingCai: Why pass here?

python/sglang/srt/entrypoints/http_server.py
JD-ETH: only the lead node expose the http server.

JD-ETH: so the way it works is, all model_runner report to head node bootstrap server.

JD-ETH: i have structured the port better though.

python/sglang/srt/entrypoints/http_server.py
```diff
@@ -1024,28 +1012,40 @@ async def send_weights_to_remote_instance(
 
 
 @app.get("/get_remote_instance_transfer_engine_info")
-@auth_level(AuthLevel.ADMIN_OPTIONAL)
 async def get_remote_instance_transfer_engine_info(rank: int = None):
```
JD-ETH, line 1020: @amysaq2023  i got the comment that we shouldn't use get_ for the endpoints anymore: https://github.com/sgl-project/sglang/blob/v0.5.7/python/sglang/srt/entrypoints/http_server.py#L525 

kept for backwards compatilbity

amysaq2023: great. Thanks for reminding this.

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -520,6 +520,14 @@ def initialize(self, pre_model_load_memory: float):
...
             )
 
+        # Register transfer engine info with the bootstrap server
+        if (
+            self.server_args.remote_instance_weight_loader_use_transfer_engine()
+            and self.remote_instance_transfer_engine is not None
+            and self.remote_instance_transfer_engine_weight_info is not None
+        ):
+            self._register_to_engine_info_bootstrap()
+
```
amysaq2023, line 528: small suggestion: shall we just merge these two `if condition`?

python/sglang/srt/server_args.py
```diff
@@ -5773,15 +5780,25 @@ def from_cli_args(cls, args: argparse.Namespace):
...
+            scheme
+        )
+
+    @property
+    def engine_info_bootstrap_url(self):
+        return self.url(port=self.engine_info_bootstrap_port)
+
+    @property
+    def disaggregation_bootstrap_url(self):
+        return self.url(port=self.disaggregation_bootstrap_port)
```
ShangmingCai: Does this PR change any disaggregation-related code? Not sure where this property has been used, I think we initiated it through server args.

python/sglang/srt/entrypoints/engine.py
```diff
@@ -113,6 +111,7 @@ class SchedulerInitResult:
     scheduler_infos: List[Dict[str, Any]]
     wait_for_ready: Callable[[], None] = lambda: None
     wait_for_completion: Callable[[], None] = lambda: None
+    engine_info_bootstrap: Optional[Any] = None
```
ShangmingCai: Should this be changed to `engine_info_bootstrap_server` as well?

python/sglang/srt/entrypoints/engine.py
```diff
@@ -195,11 +194,11 @@ def __init__(self, **kwargs):
         self.template_manager = template_manager
         self._scheduler_init_result = scheduler_init_result
         self.port_args = port_args
-        self.remote_instance_transfer_engine_info = (
-            parse_remote_instance_transfer_engine_info_from_scheduler_infos(
-                scheduler_init_result.scheduler_infos
+        # Access transfer engine info if bootstrap server is started.
+        if scheduler_init_result.engine_info_bootstrap is not None:
+            self.remote_instance_transfer_engine_info = (
+                scheduler_init_result.engine_info_bootstrap.transfer_engine_info
```
ShangmingCai: ditto

python/sglang/srt/entrypoints/engine_info_bootstrap_server.py
ShangmingCai, line 1: Maybe rename this file as well

python/sglang/srt/entrypoints/engine.py
```diff
@@ -47,6 +47,7 @@
 import zmq
 
 from sglang.srt.elastic_ep.expert_backup_manager import run_expert_backup_manager
+from sglang.srt.entrypoints.engine_info_bootstrap import EngineInfoBootstrapServer
```
ShangmingCai: Then change here as well

ShangmingCai (commented):
Others LGTM

python/sglang/srt/model_executor/model_runner.py
JD-ETH: good catch!

python/sglang/srt/server_args.py
```diff
@@ -712,6 +712,7 @@ class ServerArgs:
         "transfer_engine", "nccl", "modelexpress"
     ] = "nccl"
     remote_instance_weight_loader_start_seed_via_transfer_engine: bool = False
+    engine_info_bootstrap_port: Optional[int] = None
```
JD-ETH: we shall not set a fixed port as default. Issue is, when launching multiple sgl servers on the same node, the bootstrap server port will conflict. 

The same issue exists for PD ---- if user launches 2 PD servers, say [prefill 2 decode 2 ] x2 on a same node, it will conflict.  @ShangmingCai

python/sglang/srt/server_args.py
```diff
@@ -936,6 +937,18 @@ def _handle_missing_default_values(self):
             self.random_seed = random.randint(0, 1 << 30)
         if self.mm_process_config is None:
             self.mm_process_config = {}
+        if self.engine_info_bootstrap_port is None:
+            if self.nnodes > 1 and self.dist_init_addr:
```
JD-ETH: for nnode>1, reuse existing source port as base

python/sglang/srt/server_args.py
```diff
@@ -936,6 +937,18 @@ def _handle_missing_default_values(self):
...
+                # Multi-node: derive from dist_init_addr port so all nodes
+                # agree on the same bootstrap port regardless of per-node --port.
+                dist_port = int(self.dist_init_addr.rsplit(":", 1)[1])
+                self.engine_info_bootstrap_port = (
+                    dist_port + ENGINE_INFO_BOOTSTRAP_PORT_DELTA
+                )
+            else:
```
JD-ETH: otherwise use it's own port.

python/sglang/srt/server_args.py
ShangmingCai: Why can't we set a default value? We can always explicitly set a different port for the second prefill instances, and we let the router know about the port info.

ShangmingCai (approved):
LGTM

PR #17395, opened 2026-01-20
Fix logprob_start_len handling for prefill-only requests

ch-wan:
## Motivation

The engine's behavior is not clearly defined when users explicitly set `return_logprob` as True but do not set `logprob_start_len` for prefill-only requests. This PR returns logprob of the last logit by default to make it backward-compatible with rerank model service in old versions.

## Checklist

## Review Process


python/sglang/srt/managers/scheduler.py
```diff
@@ -1533,13 +1533,13 @@ def handle_generate_request(
             recv_req.logprob_start_len = -1
 
         if recv_req.logprob_start_len == -1:
-            if req.is_prefill_only:
+            if recv_req.return_logprob:
```
sundar24295s, line 1548: - For prefill-only, there are no output tokens, so we set the range beyond the input tokens.
- If input token logprobs are needed for prefill-only, the caller sets recv_req.logprob_start_len to the desired input token index (e.g., 0, 1, etc.).

ch-wan: What's the default behavior when logprob is True but logprob_start_len is not set? We should provide a default behavior (which is this PR trying to resolve). Otherwise we need an assertion as `logprob=True` plus `logprob_start_len=None` is not defined for prefill-only requests.

sundar24295s: @ch-wan 
- For prefill-only Scoring requests with `logprob_start_len == -1` i.e, return only outpu_token_logprobs, we want to explicity avoid any input logprob computation. This will add unnecessary GPU -> CPU copies and perf degrades.
- We can refer to the definition of `logprob_start_len` [here](https://github.com/sgl-project/sglang/blob/e9d727cb9218f30a7352f2d17614d74ec0577062/python/sglang/srt/managers/io_struct.py#L188C1-L190C62).
```
    # If return logprobs, the start location in the prompt for returning logprobs.
    # By default, this value is "-1", which means it will only return logprobs for output tokens.
    logprob_start_len: Optional[Union[List[int], int]] = None
```

Can you help re-iterate which specific scenario is broken?

ch-wan: 1. Let's focus on the expected engine behavior before discussing performance.
2. The current logic is not forward compatible. In old version (check [this](https://github.com/sgl-project/sglang/blame/94d0f656fb6d37e028c63d953a57d7089d634e2e/python/sglang/srt/managers/scheduler.py#L1262-L1266)), users just need to specify `return_logprob=True` for serving a rerank model. Now the original script does not work and the users need to further specify `logprob_start_len` to obtain the last logit.

Issue #17259, opened 2026-01-17
[Critical Bug] Context corruption in PD disaggregation mode due to metadata buffer index race condition


Simon-Li:
### Checklist

- [x] I searched related issues but found no solution.
- [x] The bug persists in the latest version.
- [x] Issues without environment info and a minimal reproducible demo are hard to resolve and may receive no feedback.
- [x] If this is not a bug report but a general question, please start a discussion at https://github.com/sgl-project/sglang/discussions. Otherwise, it will be closed.
- [x] Please use English. Otherwise, it will be closed.

### Describe the bug

In production environment with Prefill-Decode (PD) disaggregation mode, we are experiencing **critical context corruption issues** where decode servers occasionally receive incorrect first output tokens from prefill servers, leading to completely wrong inference results.

### Reproduction

When running SGLang in PD disaggregation mode under high concurrency, we observe:

  1. **Random context corruption**: Decode server appends wrong first token (from a different request) to `req.output_ids`
  2. **Inconsistent results**: Same prompt produces different (incorrect) outputs across runs
  3. **No error logs**: The bug is silent - no exceptions or warnings in logs
  4. **Unpredictable occurrence**: Happens randomly, more frequently under high load

  **Example scenario**:
  Request A: "Hello" → should generate "Hello world"
  Request B: "Goodbye" → should generate "Goodbye friend"

  But due to the bug:
  Request A receives first token from Request B → generates "Hello friend" (WRONG)

## Environment
0.5.5.post3
2P2D

## Root Cause Analysis

After deep investigation of the codebase, I identified the root cause as a **race condition in `metadata_buffer_index` allocation and reuse**.

  ### The Problem Flow

  The issue occurs in the metadata transfer mechanism between prefill and decode servers:

  **Normal Flow:**
  Prefill Server:
    1. Allocate metadata_buffer_index = X for Request A
    2. Generate first token = tokenA
    3. Write: metadata_buffers.output_ids[X] = tokenA
    4. Send KV cache + metadata to decode server

  Decode Server:
    5. Allocate metadata_buffer_index = Y for Request A
    6. Receive transfer, read: token = metadata_buffers.get_buf(Y)[0]
    7. Append token to req.output_ids

  **Bug Flow (Race Condition):**
  T1: Request A @ Prefill
      - metadata_buffer_index = 5 (allocated)
      - metadata_buffers.output_ids[5] = tokenA (written)
      - KV transfer started (async, takes time)

  T2: Request A @ Decode
[...]

PR #16951, opened 2026-01-12
Add an env var to allow transferring small metadata via TCP for PD

liusy58:
## Motivation

Mooncake's `batch_transfer_sync_write` does not guarantee that data has been written to the remote memory upon return. As a result, the receiver may read initial values (e.g., `output_id=0`)

This PR adds SGLANG_SEND_AUX_TCP to allow transferring small metadata via TCP, ensuring that the data is fully written and visible before it is accessed by the decode rank.
## Checklist

- [ ] Format your code according to the [Format code with pre-commit](https://docs.sglang.io/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.io/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.io/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.io/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.io/developer_guide/contribution_guide.html#benchmark-the-speed).
- [ ] Follow the SGLang code style [guidance](https://docs.sglang.io/developer_guide/contribution_guide.html#code-style-guidance).

## Review Process

1. Ping Merge Oncalls to start the PR flow. See the [PR Merge Process](https://github.com/sgl-project/sglang/blob/main/.github/MAINTAINER.md#pull-request-merge-process).
2. Get approvals from [CODEOWNERS](https://github.com/sgl-project/sglang/blob/main/.github/CODEOWNERS) and other reviewers.
3. Trigger CI tests with [comments](https://docs.sglang.io/developer_guide/contribution_guide.html#how-to-trigger-ci-tests) or contact authorized users to do so.
   - `/tag-run-ci-label`, `/rerun-failed-ci`, `/tag-and-rerun-ci`
4. After green CI and required approvals, ask Merge Oncalls to merge.

PR #17430, opened 2026-01-20
Add bootstrap_room validation to detect metadata corruption in PD disaggregation

Simon-Li:
For #17259

## Summary

Adds proactive detection for context corruption in Prefill-Decode disaggregation mode by validating `bootstrap_room` during metadata transfer.

## Problem

Currently, metadata corruption (where decode server receives wrong first token from a different request) can only be detected through customer reports. This:
  - Severely impacts service quality
  - Makes it difficult to validate whether fixes work
  - Results in silent data corruption reaching users
  
## Motivation
- 100% detection rate for metadata corruption
- Fails fast instead of returning wrong results to users
- Essential for validating effectiveness of #16951  and future fixes
- Zero computation overhead (simple integer comparison)
- Minimal memory cost (64 bytes per metadata buffer slot)
- Production safety best practice for critical data paths

## Solution

Store and validate `bootstrap_room` in metadata buffer:

**Prefill side:**
- Store `bootstrap_room` in metadata buffer during `set_buf()`
- Padded to 64 bytes for RDMA efficiency as suggested

**Decode side:**
- Validate `bootstrap_room` matches before using metadata
- Abort request immediately if mismatch detected
- Log detailed error for debugging

## Accuracy Tests
N/A

## Benchmarking and Profiling
N/A

## Checklist

## Review Process


ShangmingCai:
Use `pre-commit run --all-files` to fix lint.

python/sglang/srt/disaggregation/decode.py
```diff
@@ -737,8 +737,29 @@ def _commit_transfer_to_req(self, decode_req: DecodeRequest) -> None:
...
+                f"Metadata buffer index: {idx}. "
+                f"This indicates metadata buffer index collision."
+            )
+            logger.error(error_msg)
+            prepare_abort(
+                decode_req.req,
+                "Metadata corruption detected - bootstrap_room mismatch",
+                status_code=HTTPStatus.INTERNAL_SERVER_ERROR,
+            )
+            return
```
ShangmingCai: We should not abort immediately when `actual_room == 0` since this req has been served actually, we just need to wait a little time for the metadata, just don't remove it in the transfer queue, and no need to do more handling until it succeeds. If they mismatch, then we do the abort.

python/sglang/srt/disaggregation/utils.py
```diff
@@ -133,6 +133,11 @@ def __init__(
                 (size, hidden_size), dtype=hidden_states_dtype, device=device
             )
 
+            # Request validation: store bootstrap_room to detect metadata corruption
+            # Padded to 64 bytes for RDMA efficiency (8 int64s = 64 bytes)
+            # Only index 0 is used for bootstrap_room, rest is padding
```
ShangmingCai: These have been mentioned before, so they are duplicative
```suggestion
```

python/sglang/srt/disaggregation/utils.py
```diff
@@ -180,9 +188,14 @@ def get_buf(self, idx: int):
...
             self.output_topk_index[idx],
             self.output_hidden_states[idx],
+            self.bootstrap_rooms[idx],
         )
 
     def set_buf(self, req: Req):
+        # Store bootstrap_room for validation on decode side
+        # Use 0 as sentinel if bootstrap_room is None (shouldn't happen in disagg mode)
+        bootstrap_room_value = req.bootstrap_room if req.bootstrap_room is not None else 0
+        self.bootstrap_rooms[req.metadata_buffer_index, 0] = bootstrap_room_value
```
ShangmingCai: Keep the code simple and clean, and also, maybe we can move it to the end to align and improve the readability.
```suggestion
        # Store bootstrap_room for validation on decode side
        self.bootstrap_rooms[req.metadata_buffer_index, 0] = (
            req.bootstrap_room if req.bootstrap_room is not None else 0
        )
```

ShangmingCai (commented):
Looks good to me. Let's check the CI.

ShangmingCai:
I have fixed the bug. Let's try another round of CI.

Simon-Li:
> I have fixed the bug. Let's try another round of CI.

@ShangmingCai Thanks for the fix! The warmup/fake mode bypass makes sense - I hadn't considered that scenario where no actual metadata transfer occurs.

PR #17443, opened 2026-01-20
Fix wait-for-stage jobs running when call-gate fails

alisonshao:
## Summary
- Fix `wait-for-stage-a` and `wait-for-stage-b` jobs running when `call-gate` fails
- When `call-gate` fails, stage-a/b test jobs are skipped, but the wait jobs would still run and either pass (treating skipped as success) or wait forever

**Problem:**
1. `call-gate` fails → all stage-a/b jobs are skipped
2. `wait-for-stage-a` runs, sees `stage-a-test-1` as skipped, treats it as success
3. `wait-for-stage-b` runs, expects 23 matrix jobs but only finds 4 skipped jobs, waits forever

**Fix:**
- Add `call-gate` to `needs` for both wait jobs
- Add condition to skip when `call-gate` fails

**Example failure:** https://github.com/sgl-project/sglang/actions/runs/21186996415?pr=17438

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -191,8 +191,10 @@ jobs:
     needs: [check-changes, call-gate]
     # Only run for PRs (not scheduled) and when not targeting a specific stage
     # Skip if call-gate failed (stage-a jobs will be skipped, nothing to wait for)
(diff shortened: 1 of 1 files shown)


Kangyan-Zhou:
Let's increase the polling time to 2 minutes to reduce GH API call

PR #17461, opened 2026-01-21
Fix assertion tolerance for bf16 precision in triton attention UT

rahulvijayaraghavan:
### Problem

The `test_triton_attention_kernels.py::test_extend_attention` test fails sporadically with an `AssertionError` at the second iteration's `torch.allclose()` check, even though the actual numerical differences between `o_extend` and `o_redundant` are within acceptable bounds for bfloat16 precision.

The failure is masked by setting of fixed seed in setup. If we disable fixed seed as follows, we can see sporadic failures.
```
$ git diff
diff --git a/test/registered/attention/test_triton_attention_kernels.py b/test/registered/attention/test_triton_attention_kernels.py
index 803bb5cfe..676a2c181 100644
--- a/test/registered/attention/test_triton_attention_kernels.py
+++ b/test/registered/attention/test_triton_attention_kernels.py
@@ -166,7 +166,7 @@ class TestTritonAttention(CustomTestCase):

     def setUp(self):
         # Set seeds before each test method
-        self._set_all_seeds(42)
+        pass
```

CUDA
```
passed=0; for i in {1..10}; do echo "===== Run $i/10 ====="; pytest ./test/registered/attention/test_triton_attention_kernels.py::TestTritonAttention::test_extend_attention -v && ((passed++)); done; echo "===== Pass Rate: $passed/10 ($(($passed*100/10))%) ====="
...
===== Pass Rate: 5/10 (50%) =====
```

XPU
```
passed=0; for i in {1..10}; do echo "===== Run $i/10 ====="; pytest ./test/registered/attention/test_triton_attention_kernels.py::TestTritonAttention::test_extend_attention -v && ((passed++)); done; echo "===== Pass Rate: $passed/10 ($(($passed*100/10))%) ====="
...
===== Pass Rate: 6/10 (60%) =====
```

### Root Cause

The assertion used `torch.allclose(o_extend, o_redundant, rtol=1e-2)` with the default `atol=1e-8`. The `torch.allclose` function checks: `|a - b| <= atol + rtol * |b|`.

When tensor values are close to zero, the tolerance becomes approximately `atol` (since `rtol * |b|` ≈ 0). The default `atol=1e-8` is far below bfloat16's precision:

- **bfloat16 epsilon:** ~0.0078 (7.8e-3)
- **Observed max absolute diff:** ~0.001 (1e-3)
- **Default atol:** 1e-8

The observed difference of 0.001 is well within expected bfloat16 rounding error, but the assertion failed because values near zero had a tolerance of only 1e-8.

### Solution

Added `atol=1e-3` to both `torch.allclose()` assertions to properly accommodate bfloat16's limited precision while still catching meaningful numerical differences.

After adjusting tolerance:
```
passed=0; for i in {1..10}; do echo "===== Run $i/10 ====="; pytest ./test/registered/attention/test_triton_attention_kernels.py::TestTritonAttention::test_extend_attention -v && ((passed++)); done; echo "===== Pass Rate: $passed/10 ($(($passed*100/10))%) ====="
...
===== Pass Rate: 10/10 (100%) =====
```


mingfeima (approved):
Good finding!

If we intend to make it more decent, you can follow the pytorch logic:
https://github.com/pytorch/pytorch/blob/main/torch/testing/_comparison.py#L54

and here:
https://github.com/pytorch/pytorch/blob/main/torch/testing/_internal/common_device_type.py

Anyway this is mandatory, in practice, using `atol==rtol==1e-2` or `1e-3` are both acceptable.

PR #17481, opened 2026-01-21
Attempt to fix flaky test_completion_stream IndexError

alisonshao:
## Summary

This PR fixes a race condition in streaming completions where logprobs.tokens could be empty, causing IndexError when clients access tokens[0]. The issue occurs when a streaming chunk is sent after prefill but before any decode batch runs, resulting in an empty output_token_logprobs list. The fix checks if there are actual tokens before building the logprobs object, setting it to None instead of an object with empty tokens when no tokens exist.

Example failure: https://github.com/sgl-project/sglang/actions/runs/21192446524/job/60962023626
https://github.com/sgl-project/sglang/actions/runs/21227973811/job/61080255000?pr=17533
https://github.com/sgl-project/sglang/actions/runs/21230758516/job/61088501185

## Test plan

CI should pass test_completion_stream consistently.

### DIFF
--- a/python/sglang/srt/entrypoints/openai/serving_completions.py
+++ b/python/sglang/srt/entrypoints/openai/serving_completions.py
@@ -237,16 +237,23 @@ class OpenAIServingCompletion(OpenAIServingBase):
                         input_top_logprobs = None
 
                     n_prev_token = n_prev_tokens.get(index, 0)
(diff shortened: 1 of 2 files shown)


alisonshao:
Closing in favor of #17543 which combines fixes from both PRs.

PR #17487, opened 2026-01-21
fix: avoid EOS token exposure by reordering length check

ssssnow:
## Motivation

Fix EOS token exposure caused by max-length check ordering

In the current implementation, the **max-length check** is performed before the **EOS token match check**. This leads to a corner case where the EOS token is generated as the final token exactly when the maximum length is reached.

In this scenario, the request hits the length check first and its finish reason is incorrectly set to `"length"` instead of `"matched_token"`. As a result, the EOS token is not trimmed by `trim_matched_stop` and is unexpectedly streamed to the caller.

### DIFF
--- a/python/sglang/srt/managers/schedule_batch.py
+++ b/python/sglang/srt/managers/schedule_batch.py
@@ -1076,7 +1076,7 @@ class Req(ReqDllmMixin):
 
         if self._check_str_based_finish():
             return
-        
+
         # Check length limit last, so that stop tokens/strings take priority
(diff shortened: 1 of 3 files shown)


hnyls2002 (changes requested):
1. Fix the lint.
2. Add a test for the corner case (in `matched_stop_kit.py`)

hnyls2002 (changes requested):
https://github.com/sgl-project/sglang/actions/runs/21239583979/job/61116293305 CI seems to be broken for most of the cases.

PR #15892, opened 2025-12-26
[HiCache][HA 1/N] Support HiCache storage runtime attach/detach

alphabetc1:
## Motivation
Previously, the HiCache storage backend could only be configured at process startup. Changing the backend meant restarting the whole server, which hurts availability and makes operations clumsy.

This PR adds **runtime attach/detach** so operators can enable, disable, or switch the L3 storage backend **without restarting**. This is especially useful for:

* **Dynamic enable/switch of HiCache storage**
Turn HiCache storage on or off, or switch backends, on the fly based on load, cost, or debugging needs.

* **Fault tolerance (failover)**
Production backends can go bad (timeouts, misconfig, partial outages). With runtime detach, you can quickly stop sending traffic to a broken backend and avoid repeated IO failures impacting the serving path. Then you can attach a healthy backend to restore service, improving resilience and reducing MTTR.

* **Hot upgrade (switchover)**
Backends often need upgrades or migrations (new cluster, protocol changes, config updates). With runtime attach/detach, you can perform a controlled switchover when traffic is low, without restarting the server, making “hot” transitions safer and easier to operate.

## Modifications
The control path is:
1. **HTTP Server** (`python/sglang/srt/entrypoints/http_server.py`)
   - Exposes `PUT /hicache/storage-backend`, `DELETE /hicache/storage-backend`, `GET /hicache/storage-backend`
2. **TokenizerManager** (`python/sglang/srt/managers/tokenizer_communicator_mixin.py`)
   - Sends the request to the Scheduler via `_Communicator`
3. **Scheduler** (`python/sglang/srt/managers/scheduler.py`)
   - Performs a **strict idle check**
   - Calls `tree_cache.attach_storage_backend(...)` / `detach_storage_backend(...)`
4. **HiRadixCache** (`python/sglang/srt/mem_cache/hiradix_cache.py`)
   - Parses `storage_backend_extra_config_json` (supports both backend config and prefetch knobs)
   - Calls `cache_controller.attach_storage_backend(...)` / `detach_storage_backend(...)`
5. **HiCacheController** (`python/sglang/srt/managers/cache_controller.py`)
   - Creates/destroys the storage backend instance (via `StorageBackendFactory`)
   - Starts/stops backend background threads at runtime (prefetch/backup)

On the Scheduler side, add a strict idle-state check: `_is_idle_for_hicache_storage_op()`  
Conditions:
- `_is_no_request()` is `True`
- `waiting_queue` is empty
- `grammar_queue` is empty (if enabled)

[...]

PR #17521, opened 2026-01-21
[HiCache][RFC] SLA-oriented high availability for HiCache storage

alphabetc1:
## Background & Goals

HiCache Storage is currently weak at handling **runtime config updates** and **failure handling**. It doesn’t really support the kind of high availability (HA) we want — things like **seamless switchover** and **automatic failover**.

This patch lays out a **step‑by‑step plan** to make HiCache Storage more HA‑friendly. The roadmap mainly focuses on:

- **Switchover**: planned switches (upgrades, downgrades, maintenance)
- **Failover**: unplanned switches (storage or network failures)

To keep the risk manageable, this will be split into multiple smaller, independent issues.

---

### Switchover (Planned Switch)

For planned operations (upgrades, downgrades, maintenance), we want to be able to switch storage backends smoothly, with as little impact on upstream services as possible.

- Support **dynamic updates** of the storage backend: https://github.com/sgl-project/sglang/pull/15892
  - Status: `GET /hicache/storage-backend`
  - Attach / switch: `PUT /hicache/storage-backend`
  - Detach: `DELETE /hicache/storage-backend`
  - By default, be conservative:
    - Only allow updates when there are **no in‑flight requests**.
    - If there is ongoing traffic, reject the operation and return an error.

- Support **forced updates** of the storage backend: https://github.com/sgl-project/sglang/pull/17521
  - Add a `force` flag so that even under load (with in‑flight requests) and when the storage is failing, we can quickly detach the faulty backend and switch away from it, minimizing impact on upstream traffic.

- **API security**: https://github.com/sgl-project/sglang/pull/15908
  - Use an `admin-api-key` for authentication.
  - Dynamic storage updates are only allowed if:
    - `admin-api-key` is explicitly configured in sglang, and  
    - the client includes this key in the API call.
  - This is to avoid silent or accidental changes in production.

- **Partial rollback**:
  - For multi‑step migrations / switches (e.g., TP / DP / PP phases), if something goes wrong halfway or results don’t match expectations, we should be able to **roll back part of the changes** instead of doing a full reset.

---

### Automatic Failover (Unplanned Failure Handling)

When the underlying storage has problems (e.g., node down, network issues), we want the system to **detect it automatically**, **cut off the bad backend**, and **recover by switching to a healthy one** with minimal impact on callers.

- **Automatic failure handling**: https://github.com/sgl-project/sglang/pull/17521
  - Failure detection:
    - Define how we detect issues: timeouts, specific error codes, and thresholds for consecutive failures.
    - Expose metrics and hook them into alerts, so operators can see failures instead of guessing.
  - Circuit breaking & recovery:
    - When a backend is considered unhealthy, automatically **circuit break** it (logically detach it from HiCache Storage).
    - Try to reconnect to:
      - other healthy instances in the same cluster, or
      - configured standby nodes (if provided).
    - Expose the current backend status (e.g., `healthy` / `circuit_broken` / `recovering`) so it’s observable from the outside.

- **Multi-environment support**: https://github.com/sgl-project/sglang/pull/17521
  - Provide a non‑intrusive way to plug in different storage backends.
  - Try to keep a **unified management layer** for different backends, so logic doesn’t have to be reimplemented per environment.

---

## Architecture / Flow Diagram

TODO


xiezhq-hermann:
deprecated

PR #17531, opened 2026-01-21
perf: precompute lora_triples to avoid repeated zip() creation in loop

ghost:
## Motivation

The `_apply_lora_to_layers` method creates a new `zip()` object on every iteration of the outer loop over `lora_layers` (typically 300+ iterations). This causes unnecessary overhead when applying multiple LoRA adapters.

## Modifications

- Precompute `lora_triples = list(zip(lora_nicknames, lora_paths, strengths))` before the loop
- Use the precomputed `lora_triples` instead of creating `zip()` repeatedly
- Update length check to use `len(lora_triples)` for consistency


chaos-gmi:
## Benchmarking and Profiling

Micro-benchmark results showing loop overhead reduction:

| Scenario | Before | After | Improvement |
|----------|--------|-------|-------------|
| Single LoRA, 301 layers | 0.0996 ms | 0.0481 ms | **51.66%** |
| Dual LoRA, 301 layers | 0.1163 ms | 0.0622 ms | **46.55%** |
| Triple LoRA, 301 layers | 0.1397 ms | 0.0808 ms | **42.20%** |
| Dual LoRA, 600 layers | 0.2332 ms | 0.1278 ms | **45.20%** |

**Average improvement: 46.40%** (loop overhead reduction)

*Note: This micro-benchmark focuses on loop overhead. Real-world impact depends on actual workload in the loop body, but eliminates redundant `zip()` creation for 300+ layer iterations.*

PR #17543, opened 2026-01-22
Fix streaming logprobs IndexError when tokens list is empty

alisonshao:
## Motivation

Flaky `test_completion_stream` failures due to `IndexError: list index out of range` when accessing `logprobs.tokens[0]`.

A race condition causes a streaming chunk to be sent after prefill but before any decode batch runs. In this case `output_token_logprobs` is empty, but the previous code still built a logprobs object — resulting in an empty `tokens` list that causes `IndexError` when clients access `tokens[0]`.

Example failure: https://github.com/sgl-project/sglang/actions/runs/21723230283/job/62658423952

## Modifications

**Server-side (`serving_completions.py`):**
- Only build the logprobs object when there are actual tokens to report (`output_logprobs_slice or input_token_logprobs`). This handles both the race condition case (no tokens yet, `finish_reason` is None) and the finish-chunk case (buffered text flushed with no new tokens).
- Removed the previous guard that only checked for the `finish_reason is not None` case.

**Test-side (`test_openai_server.py`):**
- Added `response.choices[0].text` guard so logprobs assertions only run when the chunk contains a text delta (defense-in-depth).

Supersedes #17481.

## Test plan
- CI `test_completion_stream` should pass consistently without flaky IndexError failures


alisonshao:
https://github.com/sgl-project/sglang/actions/runs/21973792400/job/63493088847

PR #17589, opened 2026-01-22
Fix ds v32 reasoning after tool

### LINKED ISSUE #17551: [Bug] DeepSeek V3.2: All output marked as reasoning_content after tool call result
### Checklist

## Description

When using DeepSeek V3.2 model with `thinking=true` and tool calling, after the model makes a tool call during reasoning and the user sends back the tool result, **all subsequent model output is incorrectly marked as `reasoning_content` instead of normal `content`**.

## Environment

- SGLang version: latest main branch
- Model: DeepSeek-V3.2 (or any model using `--reasoning-parser deepseek-v32 --tool-call-parser deepseekv32`)
- Relevant flags: `--reasoning-parser deepseek-v32 --tool-call-parser deepseekv32`

### First /chat request
```python
    model="deepseek-v3.2",
    messages=[
        {"role": "system", "content": "You are a helpful assistant."},
        {"role": "user", "content": "What's the weather in Tokyo?"}
    ],
    tools=[{
        "type": "function",
        "function": {
            "name": "get_weather",
            "description": "Get weather information",
            "parameters": {
                "type": "object",
                "properties": {
                    "city": {"type": "string"}
                }
            }
        }
    }],
    extra_body={"chat_template_kwargs": {"thinking": True}}
)
# Response contains:
# - reasoning_content: "I need to check the weather..."
# - tool_calls: [{"function": {"name": "get_weather", "arguments": {"city": "Tokyo"}}}]
```

### Second /chat request (with tool result)
```python
    model="deepseek-v3.2",
    messages=[
        {"role": "system", "content": "You are a helpful assistant."},
        {"role": "user", "content": "What's the weather in Tokyo?"},
        {"role": "assistant", "content": "", "reasoning_content": "I need to check the weather...",
         "tool_calls": [{"id": "call_xxx", "type": "function", "function": {"name": "get_weather", "arguments": "{\"city\": \"Tokyo\"}"}}]},
        {"role": "tool", "tool_call_id": "call_xxx", "content": "Sunny, 72°F"}
    ],
    tools=[...],
    extra_body={"chat_template_kwargs": {"thinking": True}}
)
```

## Expected Behavior

The response should contain:
- `reasoning_content`: (optional reasoning about the tool result)
- `content`: "The weather in Tokyo is sunny at 72°F."

## Actual Behavior

The response contains:
- `reasoning_content`: "The weather in Tokyo is sunny at 72°F."  **← ALL output goes here**
- `content`: ""  **← Empty!**

## Root Cause Analysis

The issue is in `python/sglang/srt/entrypoints/openai/encoding_dsv32.py`, lines 240-246:

```python
if tool_call_order == len(assistant_tool_calls):
    prompt += "\n</function_results>"

    if index >= last_user_idx and thinking_mode == "thinking":
        prompt += "\n\n" + thinking_start_token   # Adds <think>
    else:
        prompt += "\n\n" + thinking_end_token     # Adds </think>
```

When the `tool` message is the **last message** (i.e., `index >= last_user_idx`), the code adds `<think>` after the function results. This means:

1. The prompt sent to the model ends with `<think>`
2. The model continues generating from there
3. A **new** `ReasoningParser` is created with `force_reasoning=True`
4. If the model outputs content without `</think>` (e.g., just the answer), **all output is treated as reasoning_content**

## Workaround

Adding a dummy user message after the tool result message fixes the issue:

```python
messages=[
    {"role": "system", "content": "..."},
    {"role": "user", "content": "What's the weather?"},
    {"role": "assistant", ..., "tool_calls": [...]},
    {"role": "tool", "content": "Sunny, 72°F"},
    {"role": "user", "content": "Please continue."}  # ← Workaround
]
```

This works because:
- `last_user_idx` now points to the new user message
- When processing the tool message: `index < last_user_idx` → adds `</think>` instead of `<think>`
- This properly closes the reasoning block before starting fresh

### Reproduction

See above

### Environment

See above

Furkeyfang:
## Motivation
In `encoding_dsv32.py`, the logic was forced to append a `thinking_start_token` (\<think\>) whenever a tool message was the last message in the list. This caused the model to continue generating within a reasoning block, leading to empty content in the API response.
This PR fixes the issue where all subsequent model output was incorrectly treated as `reasoning_content` after a tool result was provided.
Fixes #17551

## Modifications
- Removed the conditional check that appends `thinking_start_token` after the final tool result.
- Consistently append `thinking_end_token` (\</think\>) after the tool result block to ensure the model starts its final response in the correct content area.

## Accuracy Tests

Verified the prompt rendering logic with a new unit test `tests/srt/test_encoding_dsv32.py`. The test confirms that:
- The prompt correctly appends `</think>` after tool results.
- It no longer incorrectly appends `<think>` when the tool message is the last message.
The fix ensures the model's final answer is correctly captured in the `content` field.

## Benchmarking and Profiling

This is a logic-only fix in the template encoding process. It does not impact inference latency or throughput.

## Checklist

## Review Process


test/registered/function_call/test_encoding_dsv32.py
```diff
@@ -0,0 +1,34 @@
+from sglang.srt.entrypoints.openai.encoding_dsv32 import encode_messages
```
Fridge003, line 3: We can move this test to https://github.com/sgl-project/sglang/blob/main/test/registered/function_call
And register it as CI test with `register_cpu_ci(1.0, "default")`

Furkeyfang: Thanks for the suggestion! I have moved the test case to test_registered_function_call and registered it with register_cpu_ci(1.0, "default") as requested.

PR #17597, opened 2026-01-22
[BUGFIX] Fix Qwen3 Embedding model weight loading when weights already have model prefix

ovuruska:
## Summary
- Fixes weight loading for fine-tuned Qwen3 Embedding models where checkpoint weights already contain the "model." prefix
- Adds a check to avoid double-prefixing (model.model.xxx) which causes KeyError during weight loading

## Test plan
- Load a Qwen3 Embedding model that was fine-tuned with ms-swift or similar frameworks
- Verify weights load without "Parameter model.model.xxx not found" errors

### DIFF
--- a/python/sglang/srt/models/qwen3.py
+++ b/python/sglang/srt/models/qwen3.py
@@ -499,7 +499,9 @@ class Qwen3ForCausalLM(nn.Module):
 
         params_dict = dict(self.named_parameters())
         for name, loaded_weight in weights:
-            if "Embedding" in self.config.name_or_path:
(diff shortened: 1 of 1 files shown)


hnyls2002:
@ovuruska The qwen3 load_weights weight-prefix handling was rewritten on main (models/qwen3.py) and now skips re-prefixing already-prefixed names via if not name.startswith("model."), the same guard this PR introduces. Closing as superseded; please reopen if you still hit double-prefixed Embedding weights on current main.

PR #17613, opened 2026-01-23
[Perf] refactor piecewise cuda graph support of Qwen3-Next

zminglei:
## Motivation
Currently in Qwen3Next, we hide all ops inside `Qwen3GatedDeltaNet.forward` in a custom fake op, so that even with piecewise cuda graph, the whole forward of Qwen3GatedDeltaNet is running in eager mode.
Here we refactor `RadixLinearAttention` to follow the same pattern of `RadixAttention` where only the attention op is skipped during piecewise cuda graph capturing, allowing more ops to be captured, improve the prefill performance.

The gsm8k benchmark shows this change could speed up Qwen3-Next with PCG by **14.3%**.

## Modifications
 - Refactored `RadixLinearAttention` to follow the same pattern as `RadixAttention`, wrapping only the attention computation in a custom op (`unified_linear_attention_with_output`)                                                                                                                                         
  - Updated `Qwen3GatedDeltaNet.forward` to allow input projections, split/reshape, norm, and output projections to be piecewise CUDA graph captured                                                                                                                                                                        
  - Fixed `calc_rows_per_block` in `layernorm_gated.py` to use a constant value when piecewise CUDA graph is enabled, avoiding torch.compile guards on dynamic batch dimensions  

## Benchmarking and Profiling
`python3 -m sglang.launch_server --model-path /shared/public/elr-models/Qwen/Qwen3-Next-80B-A3B-Instruct/9c7f2fbe84465e40164a94cc16cd30b6999b0cc7/ --tp 4 --enable-piecewise-cuda-graph`

**Before the change:**
```
python benchmark/gsm8k/bench_sglang.py --data-path /shared/public/data/gsm8k/test.jsonl
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 200/200 [00:12<00:00, 16.23it/s]
Accuracy: 0.950
Invalid: 0.000
Latency: 12.379 s
Output throughput: 2592.343 token/s
```
**After the change:**
```
python benchmark/gsm8k/bench_sglang.py --data-path /shared/public/data/gsm8k/test.jsonl
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 200/200 [00:11<00:00, 18.18it/s]
Accuracy: 0.960
Invalid: 0.000
Latency: 11.041 s
Output throughput: 2963.644 token/s
```
The throughput is improved by **14.3%**.

## Checklist

## Review Process


python/sglang/srt/layers/attention/fla/layernorm_gated.py
```diff
@@ -158,6 +159,14 @@ def _get_sm_count(device: torch.device) -> int:
 
 
 def calc_rows_per_block(M: int, device: torch.device) -> int:
+    # When piecewise cuda graph is enabled, use a constant value to avoid
+    # torch.compile creating guards on the dynamic batch dimension.
+    try:
+        if get_global_server_args().enable_piecewise_cuda_graph:
+            return 4
```
hebiao064: why 4? avoid magic number by declare a constant maybe

python/sglang/srt/layers/attention/fla/layernorm_gated.py
```diff
@@ -158,6 +159,14 @@ def _get_sm_count(device: torch.device) -> int:
...
+    except ValueError:
+        # Global server args not initialized (e.g., in unit tests)
+        pass
     sm_count = _get_sm_count(device)
```
hebiao064, line 181: this is a constant value like 128, why it will affect torch compile

zminglei: This function returns `rows_per_block` which would be consumed by a triton kernel [_layer_norm_fwd_1pass_kernel](https://github.com/sgl-project/sglang/blob/bdaa3de07567afd3295b762df5fc0b09f6318d29/python/sglang/srt/layers/attention/fla/layernorm_gated.py#L212) as a `tl.constexpr`, with different `M` here, it could get different `rows_per_block` and trigger torch recompile.

python/sglang/srt/layers/attention/fla/layernorm_gated.py
zminglei: The 4 is the maximum rows per block value, capped for kernel performance. It appeared at the end line of this function:
    `rows_per_block = min(rows_per_block, 4)`
I've updated it to declare it as a constant for better readability, ptal.

python/sglang/srt/layers/attention/fla/layernorm_gated.py
Chen-0210: I think `M` is a constant during the compilation of a single graph, why would it trigger recompilation

hebiao064 (approved):

python/sglang/srt/layers/attention/fla/layernorm_gated.py
zminglei: It changes when num_tokens change, basically breaks torch compile guards and triggers a lot of recompilations during capturing all tokens, taking forever for the capture to finish.

ispobock (approved):

PR #17157, opened 2026-01-15
[Bugfix] fix server hang when enable --mm-enable-dp-encoder & --enable-dp-attention 

cen121212:
…e-dp-attention parameter

## Motivation

Fix server  hang when both --mm-enable-dp-encoder and --enable-dp-attention are enabled.

## ModificationsCollapse file
‎python/sglang/srt/models/qwen3_vl.py
python/sglang/srt/multimodal/mm_utils.py

## Benchmarking and Profiling

start server command:
```
python -m sglang.launch_server \
    --model-path /***/Qwen3-VL-30B-A3B-Instruct \
    --host *** \
    --port 9596 \
    --tp 2 \
    --max-prefill-tokens 40960 \
    --chunked-prefill-size 40960 \
    --attention-backend ascend \
    --mem-fraction-static 0.8 \
    --mm-attention-backend ascend_attn \
    --max-running-requests 256 \
    --disable-radix-cache \
    --cuda-graph-bs 40 60 80 120 160 200 256 \
    --trust-remote-code \
    --enable-multimodal \
    --sampling-backend ascend \
    --mm-enable-dp-encoder \
    --enable-dp-attention --dp-size 2 2>&1 | tee output.log

```
before this pr
when enable the self.use_data_parallel branch, the server will hang when warm up
```
        if max_patches_per_call == 0 and max_images_per_call == 0:
            # if self.use_data_parallel:
            #     return run_dp_sharded_mrope_vision_model(
            #         self.visual,
            #         pixel_values,
            #         image_grid_thw.tolist(),
            #         rope_type="rope_3d",
            #     )
            # else:
            return self.visual(pixel_values, grid_thw=image_grid_thw)           
```
```
[2026-01-15 13:31:50 DP0 TP0] Load weight begin. avail mem=60.40 GB
[2026-01-15 13:31:50 DP0 TP0] Using ascend_attn as multimodal attention backend.
[2026-01-15 13:31:50 DP1 TP1] Load weight begin. avail mem=60.59 GB
Loading safetensors checkpoint shards:   0% Completed | 0/13 [00:00<?, ?it/s]
[2026-01-15 13:31:50 DP1 TP1] Using ascend_attn as multimodal attention backend.
Loading safetensors checkpoint shards:   8% Completed | 1/13 [00:08<01:40,  8.40s/it]
Loading safetensors checkpoint shards:  15% Completed | 2/13 [00:16<01:32,  8.45s/it]
[...]

PR #17624, opened 2026-01-23
[BUGFIX] Fix dp size > 1 for qwen3 vl model

zju-stu-lizheng:
## Question
In my local tests, the Qwen3-VL service fails to start when both
`--mm-enable-dp-encoder`
and
`--enable-dp-attention`
are enabled.

I came across the related PR: [pr17157](https://github.com/sgl-project/sglang/pull/17157)
After applying this PR and setting tp = dp, enabling both options works without any issues.

However, when using configurations where tp != dp (for example, tp=8 and dp=4), enabling both options leads to precision/accuracy problems[sometimes ima bug].

Environment:
Model: Qwen3-VL
Command-line options: --mm-enable-dp-encoder --enable-dp-attention
Observed: service startup failure (before PR), precision issue (after PR) in tp != dp setups

## Fix bug
This branch is intended to fix the problem mentioned in this issue.

Many thanks to [yizhang2077](mailto:1109276519@qq.com) for identifying and fixing the mrope_positions padding issue.

### DIFF
--- a/python/sglang/srt/model_executor/forward_batch_info.py
+++ b/python/sglang/srt/model_executor/forward_batch_info.py
@@ -860,7 +860,11 @@ class ForwardBatch(ForwardBatchDeepSeekMHAMixin):
             )
 
         if self.mrope_positions is not None:
-            self.mrope_positions = self._pad_tensor_to_size(self.mrope_positions, bs)
+            self.mrope_positions = torch.cat(
+                [self.mrope_positions, 
+                self.mrope_positions.new_zeros(3, num_tokens - self.mrope_positions.shape[1])],
+                dim=1,
+            )
 
         # TODO: check if we need to pad other tensors
         if self.extend_seq_lens is not None:
--- a/python/sglang/srt/model_executor/forward_batch_info.py
+++ b/python/sglang/srt/model_executor/forward_batch_info.py
@@ -860,15 +860,9 @@ class ForwardBatch(ForwardBatchDeepSeekMHAMixin):
             )
 
         if self.mrope_positions is not None:
-            self.mrope_positions = torch.cat(
-                [
-                    self.mrope_positions,
-                    self.mrope_positions.new_zeros(
-                        3, num_tokens - self.mrope_positions.shape[1]
-                    ),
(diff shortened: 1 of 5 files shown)


python/sglang/srt/multimodal/mm_utils.py
```diff
@@ -495,11 +495,19 @@ def run_dp_sharded_mrope_vision_model(
         ```
 
     """
-    tp_size = get_tensor_model_parallel_world_size()
+    from sglang.srt.layers.dp_attention import (
+        get_attention_tp_group,
+        get_attention_tp_rank,
```
yizhang2077, line 500: move it into header

python/sglang/srt/model_executor/forward_batch_info.py
```diff
@@ -860,7 +860,15 @@ def _pad_inputs_to_size(self, model_runner: ModelRunner, num_tokens, bs):
             )
 
         if self.mrope_positions is not None:
-            self.mrope_positions = self._pad_tensor_to_size(self.mrope_positions, bs)
+            self.mrope_positions = torch.cat(
+                [
+                    self.mrope_positions,
```
yizhang2077, line 865: it should be compatible with `self._pad_tensor_to_size`, maybe you could write like 
`self.mrope_positions = self._pad_tensor_to_size(self.mrope_positions.t(), bs).t()`

zju-stu-lizheng: yes, it can refactor as
```
self.mrope_positions = self._pad_tensor_to_size(
                self.mrope_positions.transpose(0, 1), num_tokens
            ).transpose(0, 1)
```

python/sglang/srt/multimodal/mm_utils.py
zju-stu-lizheng: move it into header has some bug:
ImportError: cannot import name 'get_global_server_args' from partially initialized module 'sglang.srt.server_args' (most likely due to a circular import)

yhyang201 (approved):

PR #17684, opened 2026-01-24
[Ray] Add Ray actor support for scheduler process management (DP=1)

### LINKED ISSUE #4678: [Feature] Support Ray Backend for distributed inference
### Checklist

### Motivation

We request Ray for distributed inference orchestration. As detailed in our documentation [here](https://aibrix.readthedocs.io/latest/features/multi-node-inference.html), Ray abstracts away the complexity of distributed programming, making it easier to build and manage multi-node inference workflows.

By treating a Ray cluster as a single inference unit, we can simplify the orchestration layer significantly. At the same time, we continue to leverage Kubernetes for high-level resource management and autoscaling.

This setup combines the best of both worlds:
- Ray handles fine-grained application-level orchestration.
- Kubernetes manages cluster-wide resources and scaling policies.

Now, we have lots of users like to use SGLang to host Deepseek R1 full weights version(https://aibrix.github.io/posts/2025-03-10-deepseek-r1/), we have to use different orchestrator tools for that case. If SGLang can support ray backend as an alternative, that would be great!

Let’s align on whether this approach fits with our broader architecture goals and discuss any implications or alternatives.

### Related resources

_No response_

xyuzh:
This PR introduces an optional `--use-ray` flag enabling SGLang to manage Scheduler processes as Ray actors instead of native Python multiprocessing.

Closes #4678

### Short-term gains
- **Zero-config multi-node**: On a Ray cluster, `--use-ray` replaces manual NCCL/SSH orchestration with declarative placement groups — users just pass `--nnodes` and Ray handles node discovery, GPU assignment, and process placement
- **Operational visibility**: Every scheduler is a named actor visible in the Ray Dashboard with logs, resource usage, and lifecycle state — dramatically simplifies debugging multi-GPU/multi-node deployments
- **No performance cost**: Ray sits on the control plane only; ZMQ data paths are unchanged, so there is zero throughput regression (benchmarks below)

### Long-term gains
- **Ray Serve / Anyscale integration**: Enables first-class SGLang deployment on Ray Serve and Anyscale, including autoscaling, canary rollouts, and multi-model composition
- **Fault tolerance**: Foundation for actor restart policies (automatic recovery on GPU failures)
- **Disaggregated inference**: Pathway for separating prefill/decode into independently scaled actor pools managed by Ray
- **Ray Data pipelines**: Enables streaming large datasets through SGLang for offline batch inference

### Consideration between mp vs Ray
vLLM made Ray optional because it's a heavy dependency for single-node users. We follow the same principle — `--use-ray` is **strictly opt-in** and the default remains multiprocessing. However, for users already on Ray clusters (Ray Serve, Anyscale, KubeRay), this provides native integration rather than fighting the scheduler. vLLM itself still [recommends Ray for multi-node parallelism](https://docs.vllm.ai/en/latest/serving/parallelism_scaling/).

Ray is an optional dependency:
```
pip install sglang           # core only (no ray)
pip install sglang[ray]      # with ray backend
```

## What Works

- `--use-ray` with single GPU or multi-GPU Ray cluster
- TP > 1 with Ray placement groups (`STRICT_PACK` strategy)
- PP (pipeline parallelism) with Ray actors
- Multi-node deployment with per-node placement groups
- No regression for throughput vs multiprocessing mode
- Clean shutdown via `Engine.shutdown()`
- Actor lifecycle management
- Driver scripts safe for CPU-only head nodes (sglang imported only inside actors)

## Current Limitations

| Feature | Status | Notes |
|---------|--------|-------|
| TP=1 | ✅ | Fully supported |
| TP>1 | ✅ | Uses placement groups with `STRICT_PACK` |
| PP | ✅ | Uses ZMQ for activations across stages |
| Multi-node | ✅ | Per-node placement groups, `dist_init_addr` for NCCL |
| EP | ✅ |supported |
| DP | ❌ | Requires `DataParallelControllerActor` implementation |

## Key Changes

- Add `sglang/srt/ray/` module with `SchedulerActor` and `RayEngine` subclass
- `RayEngine` overrides `_launch_scheduler_processes` to use Ray actors instead of `mp.Process`
- Refactor `SchedulerInitResult` from a class hierarchy (base + `MpSchedulerInitResult` + `RaySchedulerInitResult`) into a single flat dataclass with callable fields — simpler, no subclass overhead
- Simplify `Engine.shutdown` — remove redundant guards and atexit indirection
- Add `sglang[ray]` optional dependency in `pyproject.toml`
- Add `ImportError` with install instructions when `--use-ray` is used without Ray installed
- Add manual Ray tests for TP=1, TP=2, PP=2, HTTP server, and error cases

## Design Decisions

- **ZMQ data plane unchanged** — Ray only manages process lifecycle (control plane)
- **No sglang imports at module level** in driver scripts — safe for CPU-only head nodes
- **Topology derived from `server_args.nnodes` + arithmetic** — no cluster discovery needed
- **Placement group required** — `use_ray=True` expects the caller to schedule the Engine onto a placement group; clear error message if missing
- **DP > 1 not yet supported** with Ray (raises `NotImplementedError`) in this initial integration. Would follow up in the next PR for easier reviewing

## Usage

```bash
# Single GPU
    --model meta-llama/Llama-3.2-1B-Instruct \
    --use-ray

# Multi-GPU (TP=4)
    --model meta-llama/Llama-3.1-70B-Instruct \
    --use-ray --tp 4

# Multi-node (TP=4, PP=2, 2 nodes)
    --model meta-llama/Llama-3.1-70B-Instruct \
    --use-ray --tp 4 --pp 2 --nnodes 2
```

### Multiprocessing Mode (baseline)
```
python -m sglang.bench_offline_throughput --model-path TinyLlama/TinyLlama-1.1B-Chat-v1.0 --tp 1 --dataset-name random --random-input-len 256 --random-output-len 128 --num-prompts 30

====== Offline Throughput Benchmark Result =======
Backend:                                 engine    
Successful requests:                     30        
Benchmark duration (s):                  1.56      
Total input tokens:                      4700      
Total generated tokens:                  1984      
Last generation throughput (tok/s):      2014.26   
Request throughput (req/s):              19.29     
Input token throughput (tok/s):          3021.57   
Output token throughput (tok/s):         1275.49   
Total token throughput (tok/s):          4297.05   
```

### Ray Mode
```
python -m sglang.bench_offline_throughput --model-path TinyLlama/TinyLlama-1.1B-Chat-v1.0 --tp 1 --use-ray --dataset-name random --random-input-len 256 --random-output-len 128 --num-prompts 30

====== Offline Throughput Benchmark Result =======
Backend:                                 engine    
Successful requests:                     30        
Benchmark duration (s):                  1.16      
Total input tokens:                      4700      
Total generated tokens:                  1984      
Last generation throughput (tok/s):      2013.75   
Request throughput (req/s):              25.95     
Input token throughput (tok/s):          4065.16   
Output token throughput (tok/s):         1716.01   
Total token throughput (tok/s):          5781.17   
```

### 2-Node: Ray Mode (Qwen3-8B, TP=4, PP=2)
```
====== Offline Throughput Benchmark Result =======
Configuration: backend=ray-actor, model=Qwen/Qwen3-8B, tp=4, pp=2
Backend:                                 engine    
Successful requests:                     1000      
Benchmark duration (s):                  131.84    
Total input tokens:                      383602    
Total generated tokens:                  193474    
Last generation throughput (tok/s):      2657.71   
Request throughput (req/s):              7.59      
Input token throughput (tok/s):          2909.67   
Output token throughput (tok/s):         1467.53   
Total token throughput (tok/s):          4377.20   
```

### 2-Node: Multiprocessing Mode (Qwen3-8B, TP=4, PP=2)
```
====== Offline Throughput Benchmark Result =======
Configuration: backend=mp, model=Qwen/Qwen3-8B, tp=4, pp=2
Backend:                                 engine    
Successful requests:                     1000      
Benchmark duration (s):                  131.73    
Total input tokens:                      383602    
Total generated tokens:                  193474    
Last generation throughput (tok/s):      2666.92   
Request throughput (req/s):              7.59      
Input token throughput (tok/s):          2912.00   
Output token throughput (tok/s):         1468.70   
Total token throughput (tok/s):          4380.70   
```

**Summary**: Ray mode shows no regression compared with the MP mode (both single-node and multi-node)

## Checklist


xyuzh:
> vLLM change from ray worker to mp and gradually deprecate ray, why are we try to add to sglang?

Ray dependency could be optional as well. We could introduce:

pip install sglang          # core only
pip install sglang[ray]   # with ray backend

xyuzh:
> vLLM change from ray worker to mp and gradually deprecate ray, why are we try to add to sglang?

Extra context here, mentioned in [doc](https://docs.vllm.ai/en/latest/serving/parallelism_scaling/)
Ray is still used as the recommended backend for multi-node

Mentioned in [comment](https://github.com/vllm-project/vllm/pull/23691#discussion_r2537048776)

python/sglang/srt/entrypoints/engine.py
```diff
@@ -973,18 +1096,202 @@ def _launch_scheduler_processes(
...
 
-    return scheduler_procs, scheduler_pipe_readers
+    return SchedulerLaunchResult(
+        scheduler_infos=[],
+        _procs=scheduler_procs,
+        _pipe_readers=scheduler_pipe_readers,
+    )
+
+
+def _get_rank0_node_ip(placement_group) -> str:
```
robertnishihara: Can be simpler. No need to create an actor, just run a simple task.

python/sglang/srt/entrypoints/engine.py
```diff
@@ -973,18 +1096,202 @@ def _launch_scheduler_processes(
...
+
+    probe = IPProbe.options(
+        scheduling_strategy=PlacementGroupSchedulingStrategy(
+            placement_group=placement_group,
+            placement_group_bundle_index=0,
+        ),
+    ).remote()
+
+    try:
+        ip = ray.get(probe.get_ip.remote(), timeout=30)
```
robertnishihara: Not so sure about this timeout. In the autoscaling case, where the cluster needs to scale up to create the placement group, it could easily take more than 30s.

Also, if we time out, we need a very clear / helpful error message explaining how to fix it.

python/sglang/srt/entrypoints/engine.py
```diff
@@ -993,38 +1300,75 @@ def _launch_subprocesses(
...
+                # Fallback: discover topology if not pre-computed
+                import ray
+                from sglang.srt.utils.ray_cluster_utils import (
+                    RayClusterInfo,
+                    create_cluster_topology,
+                    create_per_node_placement_groups,
+                    discover_gpu_nodes,
+                )
+
+                os.environ["RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES"] = "1"
```
kouroshHakha: you need to pass this to runtime_env inside ray.init for it to be effective

python/sglang/srt/entrypoints/engine.py
```diff
@@ -898,11 +980,62 @@ def _wait_for_scheduler_ready(
     return scheduler_infos
 
 
-def _launch_scheduler_processes(
+def _launch_schedulers(
```
robertnishihara: Don't rename this? Why?

python/sglang/srt/entrypoints/engine.py
```diff
@@ -973,18 +1096,202 @@ def _launch_scheduler_processes(
...
+    ray_cluster_info=None,
+) -> SchedulerLaunchResult:
+    """
+    Launch scheduler actors using Ray (unified single/multi-node).
+
+    Works for both single-node and multi-node:
+    - Single-node: 1 placement group with all GPUs
+    - Multi-node: N placement groups, one per GPU worker node
+
+    Steps:
```
robertnishihara: I think this can be simplified.
1. We just need "num nodes" and "num_gpus_per_node", right? Then we can create the appropriately sized placement group. Can we get those from `server_args`?
2. If yes, then can we get rid of all the ray cluster utils?
3. You have 6 steps here, but I think I would reorder it like
    1. Create a placement group (using the values from `server_args`)
    2. Wait for placement group creation (have a great error message if it's taking too long)
    3. Create scheduler actors

python/sglang/srt/managers/scheduler_actor.py
```diff
@@ -0,0 +1,130 @@
...
+import logging
+from typing import TYPE_CHECKING, Any, Dict, Optional
+
+if TYPE_CHECKING:
+    from sglang.srt.server_args import PortArgs, ServerArgs
+
+logger = logging.getLogger(__name__)
+
+
+def create_scheduler_actor_class():
```
robertnishihara: Why have a factory? Seems like unnecessary redirection.

python/sglang/srt/utils/ray_cluster_utils.py
```diff
@@ -0,0 +1,162 @@
+# Copyright 2023-2024 SGLang Team
```
robertnishihara: I'd try to eliminate this file if possible :) The entire file that is.

python/sglang/srt/utils/ray_cluster_utils.py
```diff
@@ -0,0 +1,162 @@
...
+    Create one placement group per node based on world size.
+
+    Each PG has gpus_per_node bundles with 1 GPU each.
+    Uses STRICT_PACK to keep all bundles on same node.
+    """
+    from ray.util.placement_group import placement_group
+
+    placement_groups = []
+    for _ in range(topology.nnodes):
+        bundles = [{"GPU": 1} for _ in range(topology.gpus_per_node)]
```
robertnishihara: How do the options compare
1. One placement group with one bundle per node.
2. One placement group per node, each with one bundle.
3. One placement group per node, eeach with `gpus_per_node` bundles.

python/sglang/srt/server_args.py
```diff
@@ -5500,7 +5507,7 @@ def init_new(
         server_args: ServerArgs,
         dp_rank: Optional[int] = None,
         worker_ports: Optional[List[int]] = None,
-    ) -> PortArgs:
+    ) -> "PortArgs":
```
robertnishihara: Undo this change?

python/sglang/srt/entrypoints/engine_actor.py
```diff
@@ -0,0 +1,222 @@
...
+# ==============================================================================
+"""Ray actor wrapper for SGLang Engine.
+
+This module provides a Ray actor that wraps the SGLang Engine, allowing
+sglang to be imported INSIDE the actor (on GPU worker nodes) rather than
+on the head node which may not have GPU access.
+
+Usage:
+    import ray
+    ray.init()
```
robertnishihara: The `ray.init` call is unnecessary.

python/sglang/srt/utils/ray_cluster_utils.py
```diff
@@ -0,0 +1,162 @@
...
+    """Discovered Ray cluster topology for multi-node deployment."""
+
+    nodes: List[RayNodeInfo]
+    total_gpus: int
+    nnodes: int
+    gpus_per_node: int  # Assumes homogeneous
+
+
+@dataclasses.dataclass
+class RayClusterInfo:
```
robertnishihara: Let's not have this class, feels too heavyweight. Instead, we just need to create a placement group of the right shape at the right time (and that just needs to know the number of nodes and GPUs per node.

python/sglang/srt/entrypoints/engine_actor.py
```diff
@@ -0,0 +1,222 @@
...
+    placement_groups = create_per_node_placement_groups(topology)
+
+    logger.info(
+        f"Launching EngineActor on rank 0's node: "
+        f"{topology.nnodes} nodes, {topology.gpus_per_node} GPUs/node, "
+        f"world_size={world_size}"
+    )
+
+    @ray.remote
+    class EngineActor:
```
robertnishihara: Would it be possible to literally do `EngineActor = ray.remote(sglang.Engine)`? That would be awesome / ideal. Or do we need to add some stuff to it?

python/sglang/srt/entrypoints/engine.py
```diff
@@ -973,18 +1096,202 @@ def _launch_scheduler_processes(
...
+
+    topology = ray_cluster_info.topology
+    placement_groups = ray_cluster_info.placement_groups
+    rank0_node_ip = ray_cluster_info.rank0_node_ip
+
+    # Use pre-computed topology/PGs or create them (fallback)
+    if topology is None or placement_groups is None:
+        os.environ["RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES"] = "1"
+        if not ray.is_initialized():
+            ray.init()
```
robertnishihara: Unnecessary

python/sglang/srt/entrypoints/engine.py
```diff
@@ -993,38 +1300,75 @@ def _launch_subprocesses(
...
+                from sglang.srt.utils.ray_cluster_utils import (
+                    RayClusterInfo,
+                    create_cluster_topology,
+                    create_per_node_placement_groups,
+                    discover_gpu_nodes,
+                )
+
+                os.environ["RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES"] = "1"
+                if not ray.is_initialized():
+                    ray.init()
```
robertnishihara: All calls to `ray.init` are unnecessary

python/sglang/srt/entrypoints/engine_actor.py
```diff
@@ -0,0 +1,222 @@
...
+sglang to be imported INSIDE the actor (on GPU worker nodes) rather than
+on the head node which may not have GPU access.
+
+Usage:
+    import ray
+    ray.init()
+
+    from sglang.srt.entrypoints.engine_actor import create_and_launch_engine_actor
+
+    engine_actor = create_and_launch_engine_actor(
```
robertnishihara: Not so sure about this API. Maybe something like

```python
engine = RayEngine(
    model_path=...,
    tp_size=...,
)
```

Or would it make sense for `RayEngine` to be an actor?

python/sglang/srt/entrypoints/engine_actor.py
```diff
@@ -0,0 +1,222 @@
...
+        model_path="meta-llama/Llama-2-7b-hf",
+        tp_size=4,
+        use_ray=True,
```
robertnishihara: Remove this arg because this only works with Ray, right?

python/sglang/srt/entrypoints/engine.py
xyuzh: To distinguish ray/multi-process backend.
At `_launch_schedulers` we would either use subprocess or ray actor for launching scheduler.
Thus launch `_launch_scheduler_processes` is not accurate for ray actor

python/sglang/srt/entrypoints/engine.py
```diff
@@ -973,18 +1096,202 @@ def _launch_scheduler_processes(
...
+    if ray_cluster_info is None:
+        ray_cluster_info = RayClusterInfo()
+
+    topology = ray_cluster_info.topology
+    placement_groups = ray_cluster_info.placement_groups
+    rank0_node_ip = ray_cluster_info.rank0_node_ip
+
+    # Use pre-computed topology/PGs or create them (fallback)
+    if topology is None or placement_groups is None:
+        os.environ["RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES"] = "1"
```
robertnishihara: In many cases, the Ray cluster will already be started. So need to think about how the environment variable is propagated in those cases.

Can we get rid of this environment variable if we use one bundle per placement group (instead of one per GPU)?

Also, in the future, we should try to get rid of this if possible.

python/sglang/srt/entrypoints/engine_actor.py
```diff
@@ -0,0 +1,222 @@
...
+from __future__ import annotations
+
+import logging
+import os
+from typing import Any, Dict, Iterator, List, Optional, Union
+
+logger = logging.getLogger(__name__)
+
+
+def create_and_launch_engine_actor(**engine_kwargs):
```
jonsjlee: Could we add a return type here

```
def create_and_launch_engine_actor(**engine_kwargs) -> ray.actor.ActorProxy[EngineActor]
```

Then do the if TYPE_CHECKING trick

python/sglang/srt/entrypoints/engine_actor.py
jonsjlee: Does this need to be defined inline? I think this would prevent us from accessing the type.

python/sglang/srt/managers/scheduler_actor.py
jonsjlee: +1

If we're only importing Ray within functions (I think that's the goal?), we could have this be:

```
class SchedulerActor:
  ...
```

Then wherever you use it:

```
ray.remote(SchedulerActor).remote(args)
```

Then we can access SchedulerActor for typing.

jonsjlee (changes requested):
Really appreciate the work!

I left some nits - disclaimer I'm not too familiar with this codebase or style. In our codebase, we have strong typing. If possible, could we get as much typing as possible in this API?

python/sglang/srt/entrypoints/engine.py
robertnishihara: I actually think `_launch_scheduler_processes` is still accurate. Ray actors are also processes, it's just a different way of doing process management.

python/sglang/srt/managers/scheduler_actor.py
robertnishihara: Ok, the issue seems to be that `import sglang` can only be run on a GPU node, whereas the Ray driver script may run on a node with CPUs. However, I think that can be handled in the application logic (as opposed to inside of SGLang). 

This is verbose, but another alternative for discussion (which errs on the side of giving more control to the user), in particular the user manually creates placement group (alternative is to it inside SGLang), the benefit is that other things can be colocated in the same placement group of the user wants.

```
import ray
import ray.util.scheduling_strategies
import sglang as sgl

num_nodes = 16
num_gpus_per_node = 8


# This group is used to place the engine actor and one scheduler actor per GPU.
placement_group = ray.util.placement_group(
    name="engine_group",
    bundles=num_nodes * [
        {"CPU": 1},
        {"GPU": num_gpus_per_node},
    ],
)

@ray.remote(
    scheduling_strategy=PlacementGroupSchedulingStrategy(
        placement_group=placement_group,
        placement_group_capture_child_tasks=True
    )
)
class SGLangWrapper:
    def __init__(self):
        import sglang as sgl

        EngineActorClass = ray.remote(sgl.EngineActor)
        self.llm = EngineActorClass.remote(
            model_path="qwen/qwen2.5-0.5b-instruct",
            tp_size=8,
            pp_size=16,
            use_ray=True
        )

    def generate(self, prompts):
        return ray.get(self.llm.generate.remote(prompts))


llm_wrapper = SGLangWrapper.remote()
ray.get(llm_wrapper.generate.remote(["Hello, my name is"]))
```

python/sglang/srt/utils/ray_utils.py
```diff
@@ -0,0 +1,35 @@
...
+    Args:
+        tp_size: Tensor parallelism size.
+        pp_size: Pipeline parallelism size.
+        nnodes: Number of nodes.
+
+    Returns:
+        List of placement groups, one per node.
+    """
+    world_size = tp_size * pp_size
+    gpus_per_node = world_size // nnodes
```
pcmoritz: Are there any higher level checks for / do we know this is divisible?

python/sglang/srt/entrypoints/engine.py
```diff
@@ -94,6 +92,86 @@
...
+    _procs: Optional[List] = None
+    _pipe_readers: Optional[List] = None
+
+    @property
+    def actors(self) -> Optional[List]:
+        """Ray actors (None for mp mode). Exposed for Engine to store."""
+        return self._actors
+
+    @property
+    def event_loop_refs(self) -> Optional[List]:
```
pcmoritz: It doesn't seem that this is still used at the moment, right? If not, it would be better to remove, the less backend specific stuff that is in this class, the better.

xyuzh: It's used at:

```
    def wait_for_completion(self) -> None:
        """Block until all schedulers terminate (used for non-zero rank nodes)."""
        if self._actors is not None:
            import ray

            try:
                ray.get(self._event_loop_refs)
            except Exception as e:
                logger.error(f"Ray scheduler actor terminated with error: {e}")
        elif self._procs is not None:
            for proc in self._procs:
                proc.join()
                logger.error(
                    f"Scheduler or DataParallelController {proc.pid} terminated with {proc.exitcode}"
                )
```

to wait for completion

python/sglang/srt/utils/ray_utils.py
xyuzh: yes at `server_args.py`
```
def check_server_args(self):
        # Check parallel size constraints
        assert (
            self.tp_size * self.pp_size
        ) % self.nnodes == 0, "tp_size must be divisible by number of nodes"
```

python/sglang/srt/ray/scheduler_actor.py
```diff
@@ -0,0 +1,124 @@
...
+            tp_rank,
+            moe_ep_rank,
+            pp_rank,
+            dp_rank,
+        )
+
+        self._tp_rank = tp_rank
+        self._pp_rank = pp_rank
+
+    def get_info(self) -> Dict[str, Any]:
```
pcmoritz, line 101: This seems to be the same as `self.scheduler.get_init_info()` and also not used, should it be removed / deduped?

python/sglang/srt/entrypoints/engine.py
pcmoritz: But that uses the private field `self._event_loop_refs` and not the public property `self.event_loop_refs`, right? It would be better to remove the property, thus making it an implementation detail of the class :)

python/sglang/srt/entrypoints/engine.py
```diff
@@ -94,6 +92,86 @@
...
+    _actors: Optional[List] = None
+    _event_loop_refs: Optional[List] = None
+    _placement_groups: Optional[List] = None
+
+    # mp mode fields
+    _procs: Optional[List] = None
+    _pipe_readers: Optional[List] = None
+
+    @property
+    def actors(self) -> Optional[List]:
```
pcmoritz: I think this public property is also not used

python/sglang/srt/entrypoints/engine.py
```diff
@@ -902,7 +980,63 @@ def _launch_scheduler_processes(
...
+            port_args,
+            placement_groups=placement_groups,
+        )
+    else:
+        return _launch_scheduler_processes_mp(
+            server_args, port_args, run_scheduler_process_func
+        )
+
+
+def _calculate_rank_ranges(server_args: ServerArgs, node_rank: Optional[int] = None):
```
pcmoritz: Instead of passing in server_args and then doing the trick with optional `node_rank`, why not use
`_calculate_rank_ranges(nnodes: int, pp_size: int, tp_size: int, node_rank: int)` as the signature
and therefore make the handling uniform? That would be cleaner I think

pcmoritz: Also it would be good to document what is returned, and add a return type annotation.

python/sglang/srt/ray/scheduler_actor.py
```diff
@@ -0,0 +1,117 @@
...
+import ray
+
+if TYPE_CHECKING:
+    from sglang.srt.server_args import PortArgs, ServerArgs
+
+logger = logging.getLogger(__name__)
+
+
+@ray.remote
+class SchedulerActor:
```
kouroshHakha, line 31: I would make the init signature directly match `Scheduler` so that if there is any signature changes down the line, this can inherit those. 

```
@ray.remote
class SchedulerActor:

    def __init__(self, scheduler_kwargs: dict, dist_init_addr: Optional[str] = None):
        ...

        scheduler_kwargs["server_args"] = dataclasses.replace(
            scheduler_kwargs["server_args"], dist_init_addr=dist_init_addr
        )
        
```

python/sglang/srt/entrypoints/engine.py
```diff
@@ -154,21 +237,30 @@ def __init__(self, **kwargs):
...
-                init_tokenizer_manager_func=self.init_tokenizer_manager_func,
-                run_scheduler_process_func=self.run_scheduler_process_func,
-                run_detokenizer_process_func=self.run_detokenizer_process_func,
-            )
+        (
+            tokenizer_manager,
+            template_manager,
+            scheduler_infos,
+            port_args,
+            scheduler_result,
```
kouroshHakha: this is a bit not clean, I think we can refactor scheduler_infos to include the info that we need for ray stuff, instead of having `_launch_workers` to return both scheduler_info and scheduler_result which already includes scheduler_info. Also the name scheduler_result suggests that this the output of the scheduler on the data plane rather than the control plane information.

python/sglang/srt/entrypoints/engine.py
```diff
@@ -936,18 +1097,157 @@ def _launch_scheduler_processes(
...
+                pass
+        raise RuntimeError(f"Scheduler actor failed to initialize: {e}")
+
+    # Start event loops (non-blocking)
+    event_loop_refs = [actor.run_event_loop.remote() for actor in scheduler_actors]
+
+    return SchedulerLaunchResult(
+        scheduler_infos=scheduler_infos,
+        _actors=scheduler_actors,
+        _event_loop_refs=event_loop_refs,
```
kouroshHakha: why do we need event loop refs in the result object?

python/sglang/srt/entrypoints/engine.py
```diff
@@ -936,18 +1102,157 @@ def _launch_scheduler_processes(
...
+        _placement_groups=placement_groups,
+    )
+
+
+def _launch_workers(
     server_args: ServerArgs,
```
kouroshHakha: this is very invasive, I don't want to create traces of ray_placement_groups in methods that are just acting as pass throughs. 

There is trick we can play here to make things clean. basically until you reach a place to launch the actors you'll be in the driver process, and would have access to global variables of this process.


I would create a context manager that sets a globall variable for placement_groups so that you can retrieve it inside any downstream method (i.e. inside the create_actors method). It would look something like:

```

if use_ray: 
    # this should be at the highest level possible that you can create placement groups
    with setup_ray_pgs(tp_size, pp_size, nnodes):
        Engine(...) # no placement groups passed into Engine
```

the context manager sets a global var like `__ray_placement_groups` that you can retrive by `get_global_placement_groups()` inside any downstream method living in the current process.

python/sglang/srt/entrypoints/engine.py
```diff
@@ -936,18 +1102,157 @@ def _launch_scheduler_processes(
...
+    nnodes = server_args.nnodes
+    gpus_per_node = world_size // nnodes
+
+    logger.info(
+        f"Ray cluster: {nnodes} nodes, "
+        f"Use {gpus_per_node} GPUs/node, world_size={world_size}"
+    )
+
+    # Discover rank 0's node IP for torch.distributed init
+    rank0_node_ip = _get_rank0_node_ip(placement_groups[0])
```
kouroshHakha: why inject instance_id? isn't tp/pp rank enough to identify the actor?

python/sglang/srt/entrypoints/engine.py
```diff
@@ -936,18 +1102,157 @@ def _launch_scheduler_processes(
...
+            "Set dp_size=1 or use_ray=False."
+        )
+
+    world_size = server_args.tp_size * server_args.pp_size
+    nnodes = server_args.nnodes
+    gpus_per_node = world_size // nnodes
+
+    logger.info(
+        f"Ray cluster: {nnodes} nodes, "
+        f"Use {gpus_per_node} GPUs/node, world_size={world_size}"
```
kouroshHakha: avoid this 233 pattern? Create an env var for this offset value?

python/sglang/srt/entrypoints/engine_actor.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def is_ready(self) -> bool:
+        return self.engine is not None
+
+    def generate(self, **kwargs: Any) -> Union[Dict, Iterator[Dict]]:
+        return self.engine.generate(**kwargs)
+
+    def encode(
+        self,
+        prompt: Union[str, List[str], List[Dict], List[List[Dict]]],
```
kouroshHakha: I feel like we have extra layers that don't do much here:

We have EngineActor (not really an actor 1 per cluster on driver) -- InternalEngineActor (1 per cluster on gpu) -- Engine (in the same process as the InternalEngineActor) -- GPU actors

I think we can simplify to Driver -- EngineActor -- GPU actors 

You can pass in the placementgroups via runtime_env so that you don't have to add `placement_group` to signature of classes that are used for both mp and ray.

python/sglang/srt/entrypoints/http_server_actor.py
```diff
@@ -0,0 +1,189 @@
...
+    def get_status(self) -> Dict[str, Union[str, bool]]:
+        """Get server status in a single RPC call."""
+        return {
+            "url": self.get_url(),
+            "node_ip": self.node_ip,
+            "alive": self.is_alive(),
+        }
+
+    def shutdown(self) -> None:
+        """Shutdown the server."""
```
kouroshHakha: having to repeat your design pattern for both engine and http_server suggest that this is not the correct layer to inject this information. It should be lower level in my opinion:

driver creates the node pgs --> Engine actors (driver of gpu work num_gpu=0, num_cpu=0) --> GPU workers (using node pgs)

kouroshHakha (commented):
in your benchmark compaison, could you include the latency numbers? 

Let's add tests with these settings:

- standalone use-ray
- instantiated inside an orchestrator actor to see if the child actors can inherit the placement groups from the parent actors

python/sglang/srt/entrypoints/engine.py
xyuzh: good catch, fixed

python/sglang/srt/entrypoints/engine.py
xyuzh: Blocking until completion / detecting errors. In `wait_for_completion` the code calls `ray.get(self._event_loop_refs)`, which blocks until all scheduler event loops finish. This is the Ray equivalent of calling `proc.join()` on multiprocessing processes

python/sglang/srt/ray/scheduler_actor.py
xyuzh: I don't think using `scheduler_kwargs` to wrap all scheduler params is a good practice, changes of `scheduler` would have to be reflected at `SchedulerActor` regardless of using `scheduler_kwargs`

kouroshHakha: I think we can address potential divergence with good end to end tests on the ray side.

python/sglang/srt/entrypoints/engine.py
```diff
@@ -156,24 +213,31 @@ def __init__(self, **kwargs):
...
-                server_args=server_args,
-                init_tokenizer_manager_func=self.init_tokenizer_manager_func,
-                run_scheduler_process_func=self.run_scheduler_process_func,
-                run_detokenizer_process_func=self.run_detokenizer_process_func,
-            )
+        (
+            tokenizer_manager,
+            template_manager,
+            port_args,
+            scheduler_result,
```
kouroshHakha: let's rename this to `scheduler_init_results` ?

python/sglang/srt/entrypoints/engine.py
```diff
@@ -156,24 +213,31 @@ def __init__(self, **kwargs):
...
+            init_tokenizer_manager_func=self.init_tokenizer_manager_func,
+            run_scheduler_process_func=self.run_scheduler_process_func,
+            run_detokenizer_process_func=self.run_detokenizer_process_func,
         )
         self.tokenizer_manager = tokenizer_manager
         self.template_manager = template_manager
-        self.scheduler_info = scheduler_infos[0]
+        self.scheduler_info = scheduler_result.scheduler_infos[0]
         self.port_args = port_args
+        self._scheduler_result = scheduler_result
```
kouroshHakha: call this `self._scheduler_init_results`

kouroshHakha (commented):
Looks good for sglang team review. Just left a few minor things. 

We need to add the end to end test to this PR + docs before asking for team's review.

python/sglang/srt/ray/scheduler_actor.py
xyuzh: Losing type safety is the major concern, and we have to struct the whole `scheduler_kwargs` as a dict in the downstream which doesn't look clean
@robertnishihara @pcmoritz what you think

python/sglang/srt/entrypoints/engine.py
pcmoritz: To keep the terminology consistent, scheduler_launch_results would be more logical

python/sglang/srt/ray/scheduler_actor.py
pcmoritz: I like the type safety too, but you should decide. If you want to have type safety AND be able to pass through extra args, you could do something like
```
def __init__(
        self,
        server_args: ServerArgs,
        port_args: PortArgs,
        gpu_id: int,
        tp_rank: int,
        moe_ep_rank: int,
        pp_rank: int,
        dp_rank: Optional[int],
        dist_init_addr: Optional[str] = None,
        **extra_args
    )
```
and pass them through.

python/sglang/srt/entrypoints/engine_actor.py
xyuzh: This PR now allows this with:
```
@ray.remote
class EngineActor:
    def __init__(self, **kwargs):
        from sglang import Engine
        self.engine = Engine(**kwargs)

    def generate(self, prompts, sampling_params):
        return [self.engine.generate(prompt=p, sampling_params=sampling_params) for p in prompts]

    def shutdown(self):
        if self.engine:
            self.engine.shutdown()
            self.engine = None

world_size = TP_SIZE * PP_SIZE
gpus_per_node = world_size // NNODES

pg = placement_group(
    bundles=[{"CPU": 1, "GPU": gpus_per_node}] * NNODES,
    strategy="STRICT_SPREAD",
)
ray.get(pg.ready())

engine = EngineActor.options(
    num_cpus=1, num_gpus=0,
    scheduling_strategy=PlacementGroupSchedulingStrategy(
        placement_group=pg, placement_group_bundle_index=0,
    ),
).remote(
    model_path=MODEL_PATH, tp_size=TP_SIZE, pp_size=PP_SIZE,
    nnodes=NNODES, port=PORT, use_ray=True,
)

results = ray.get(engine.generate.remote(
    ["The capital of France is", "What is 2 + 2?"],
    {"max_new_tokens": 64, "temperature": 0.0},
))
```

python/sglang/srt/entrypoints/engine.py
xyuzh: renamed
`scheduler_result ` -> `scheduler_init_results`,
`SchedulerLaunchResult` -> `SchedulerInitResult`

python/sglang/srt/ray/scheduler_actor.py
jonsjlee: I think **extra_args loses the type safety. Maybe 

```
def __init__(
        self,
        server_args: ServerArgs,
        port_args: PortArgs,
        gpu_id: int,
        tp_rank: int,
        moe_ep_rank: int,
        pp_rank: int,
        dp_rank: Optional[int],
        dist_init_addr: Optional[str] = None,
        extra_args: dict[str, obj] | None = None
    )
```

python/sglang/srt/ray/scheduler_actor.py
```diff
@@ -0,0 +1,101 @@
...
+            tp_rank=tp_rank,
+            moe_ep_rank=moe_ep_rank,
+            pp_rank=pp_rank,
+            dp_rank=dp_rank,
+        )
+
+        self._tp_rank = tp_rank
+        self._pp_rank = pp_rank
+
+    def get_info(self) -> Dict[str, Any]:
```
jonsjlee, line 101: Can we decorate these methods with ray.method?

python/sglang/srt/entrypoints/engine.py
```diff
@@ -998,10 +1306,9 @@ def _launch_subprocesses(
     # Allocate ports for inter-process communications
     if port_args is None:
         port_args = PortArgs.init_new(server_args)
-    logger.info(f"{server_args=}")
```
Qiaolin-Yu, line 1075: Why remove this?

python/sglang/srt/entrypoints/engine.py
```diff
@@ -153,27 +223,33 @@ def __init__(self, **kwargs):
                 kwargs["log_level"] = "error"
             server_args = self.server_args_class(**kwargs)
         self.server_args = server_args
-        logger.info(f"{server_args=}")
```
Qiaolin-Yu, line 158: Same

Qiaolin-Yu (changes requested):
Just brainstorm: could we have a ray folder where `rayscheduler` and `rayengine` inherit from the upstream classes (`scheduler`, `engine`, etc.) and selectively override some functions, so we can avoid intrusive changes (so many if/else statements) to the main code path? Also, in the future, I expect you will support more parallelism, it would be even more complicated if we still add all the if/else in the current code path.

python/sglang/srt/entrypoints/engine.py
xyuzh: The logger is called at `__init__` of Engine and before `configure_logger` is called which makes the logger invalid, also it's duplicated hence I removed it here.

I added it back, could address this nit in another PR

python/sglang/srt/entrypoints/engine.py
xyuzh: This one could be kept, added back

xyuzh:
@Qiaolin-Yu I've addressed your feedback — Ray code is now isolated in a dedicated `sglang/srt/ray/` package:
- `SchedulerInitResult` is a polymorphic base class with `MpSchedulerInitResult` and `RaySchedulerInitResult` subclasses (no if/else in the main path)
- All Ray-specific code (`SchedulerActor`, `launch_scheduler_ray_actors`, `RaySchedulerInitResult`) lives in `sglang/srt/ray/`
- Single dispatch point with lazy import in `_launch_scheduler_processes()`

python/sglang/srt/entrypoints/engine.py
```diff
@@ -450,10 +510,36 @@ def rerank(
...
         kill_process_tree(os.getpid(), include_parent=False)
 
+    def _atexit_shutdown(self):
+        """Atexit handler - cleans up Ray actors and child processes."""
+        if self._shutdown_called:
+            return
+        self._shutdown_called = True
+        self._cleanup_processes()
+
+    def shutdown(self):
```
Qiaolin-Yu, line 697: We do not need to change this. We just need to override it in ray folder

python/sglang/srt/entrypoints/engine.py
```diff
@@ -912,7 +1002,70 @@ def _launch_scheduler_processes(
...
+    Returns:
+        SchedulerInitResult that abstracts the mode-specific details.
+        For Ray mode, scheduler_infos is already populated.
+        For mp mode, call result.wait_for_ready() to populate scheduler_infos.
+    """
+    if server_args.use_ray:
+        from sglang.srt.ray.launch import launch_scheduler_ray_actors
+
+        return launch_scheduler_ray_actors(server_args, port_args)
+    else:
```
Qiaolin-Yu: better to have a rayengine class and put all of these logic in the override function

Qiaolin-Yu (changes requested):
By the way, could we add a test for this? Now it could be under `/test/manual`, eventually we may move it to ci.

python/sglang/srt/managers/scheduler.py
```diff
@@ -3084,13 +3147,33 @@ def run_scheduler_process(
...
-    parent_process = psutil.Process().parent()
 
     # Configure the logger
     configure_logger(server_args, prefix=prefix)
     suppress_other_loggers()
 
+    return dp_rank
+
+
+def run_scheduler_process(
```
xyuzh, line 3193: Refactor to wrap the function body in methods for reuse in `scheduler_actors.py`

python/sglang/srt/entrypoints/engine.py
```diff
@@ -94,6 +104,16 @@
...
+class SchedulerInitResult:
+    """Result from launching schedulers."""
+
+    scheduler_infos: List[Dict[str, Any]]
+    wait_for_ready: Callable[[], None] = lambda: None
+    wait_for_completion: Callable[[], None] = lambda: None
+    cleanup: Callable[[], None] = lambda: None
+
+
 def init_tokenizer_manager(
```
xyuzh, line 118: introduce `SchedulerInitResult` for easy override with minimal duplication of code

python/sglang/srt/entrypoints/engine.py
```diff
@@ -931,182 +1143,59 @@ def _wait_for_scheduler_ready(
     return scheduler_infos
 
 
-def _launch_scheduler_processes(
```
xyuzh, line 955: Refactor only changes to wrap function body to reusable methods for avoiding duplication

python/sglang/srt/entrypoints/http_server.py
```diff
@@ -1927,45 +1927,23 @@ def _wait_weights_ready():
     )
 
 
-def launch_server(
+def _setup_and_run_http_server(
```
xyuzh, line 1955: refactor only changes to reuse

python/sglang/srt/entrypoints/http_server.py
```diff
@@ -2058,3 +2020,49 @@ def launch_server(
         if server_args.tokenizer_worker_num > 1:
             multi_tokenizer_args_shm.unlink()
             _global_state.tokenizer_manager.socket_mapping.clear_all_sockets()
+
+
+def launch_server(
```
xyuzh, line 2083: Refactor only changes to wrap function body into reusable methods

python/sglang/srt/managers/scheduler.py
xyuzh, line 3193: Refactor only changes to wrap function body into reusable methods

python/sglang/launch_server.py
```diff
@@ -20,6 +20,16 @@ def run_server(server_args):
     elif server_args.encoder_only:
         from sglang.srt.disaggregation.encode_server import launch_server
 
+        launch_server(server_args)
```
xyuzh: @Qiaolin-Yu This should be the only non-refactoring change to the sglang code

python/sglang/srt/ray/scheduler_actor.py
xyuzh: will do in the follow up PR

python/sglang/srt/entrypoints/engine.py
xyuzh: Now it's refactor only to reuse code

python/sglang/bench_offline_throughput.py
```diff
@@ -323,12 +323,77 @@ def monitor_trace_file(known_files, directory, interval=1):
...
+    return _Proxy()
+
+
 def throughput_test(
     server_args: ServerArgs,
     bench_args: BenchArgs,
 ):
     if bench_args.backend == "engine":
-        backend = Engine(**dataclasses.asdict(server_args))
+        if server_args.use_ray:
```
xyuzh, line 399: supporting:
```
python -m sglang.bench_offline_throughput --model-path TinyLlama/TinyLlama-1.1B-Chat-v1.0 --tp 1 --use-ray --dataset-name random --random-input-len 256 --random-output-len 128 --num-prompts 30
```

python/sglang/srt/entrypoints/engine.py
xyuzh: Previously `scheduler_procs, scheduler_pipe_readers = _launch_scheduler_processes`
`scheduler_procs` was the artifact of mp, use more generic methods to make Engine backend agnostic

python/sglang/srt/entrypoints/http_server.py
```diff
@@ -1932,45 +1932,20 @@ def _wait_weights_ready():
...
-        )
-    )
-
-    # Parse info got from the schedulers
+    """Set global state, configure middleware, and run uvicorn."""
     remote_instance_transfer_engine_info = (
         parse_remote_instance_transfer_engine_info_from_scheduler_infos(scheduler_infos)
     )
 
-    # Set global states
```
Qiaolin-Yu: Why do we need to remove these comments? If you are using vibe coding, could you check these unrelated changes?

python/sglang/srt/ray/engine.py
Qiaolin-Yu, line 1: Previously there were some extra things for the shutdown of ray? But I don't see the override of `shutdown()` here.

python/sglang/srt/entrypoints/engine.py
```diff
@@ -1048,80 +1205,17 @@ def _launch_subprocesses(
...
-    # Launch scheduler processes
-    scheduler_procs, scheduler_pipe_readers = _launch_scheduler_processes(
+) -> Tuple[
+    TokenizerManager,
+    TemplateManager,
+    PortArgs,
+    SchedulerInitResult,
+]:
+    """Module-level launcher used by http_server.py."""
+    return Engine._launch_workers(
```
Qiaolin-Yu: I don't get why we need to add an extra `_launch_workers` here. It seems just duplicated with `_launch_subprocesses`.

Qiaolin-Yu (changes requested):

python/pyproject.toml
```diff
@@ -143,6 +147,7 @@ dev = ["sglang[test]"]
 all = [
   "sglang[diffusion]",
   "sglang[tracing]",
+  "sglang[ray]",
```
Qiaolin-Yu: we do not need to include it in [all]

python/sglang/srt/ray/engine.py
xyuzh: after some investigation, I found the actor processes will automatically terminate when the driver process exits or when handlers go out of scope.
So there is no need to manually call `shutdown` as when the `RayEngine` exits, the actors would be garbage collected automatically.
@robertnishihara @pcmoritz for you to double check

python/sglang/srt/entrypoints/engine.py
xyuzh: `_launch_workers` is a class method meant to be reused by `RayEngine` and `Engine` while `_launch_subprocesses` is used by `http_server.py` 

Separating `_launch_workers` from `_launch_subprocesses` makes the `http_server.py` easier to call different backends with the least code duplication

python/sglang/srt/entrypoints/engine.py
```diff
@@ -160,21 +179,24 @@ def __init__(self, **kwargs):
...
+        ) = self._launch_workers(
+            server_args=server_args,
+            init_tokenizer_manager_func=self.init_tokenizer_manager_func,
+            run_scheduler_process_func=self.run_scheduler_process_func,
+            run_detokenizer_process_func=self.run_detokenizer_process_func,
         )
         self.tokenizer_manager = tokenizer_manager
         self.template_manager = template_manager
-        self.scheduler_info = scheduler_infos[0]
+        self._scheduler_init_result = scheduler_init_result
```
xyuzh, line 196: scheduler_info is included in _scheduler_init_result now

python/sglang/srt/ray/engine.py
xyuzh: Added back manual shutdown by
```
def shutdown(self):
        """Shutdown the engine — kill Ray scheduler actors then local processes."""
        for actor in self._scheduler_init_result.scheduler_actors:
            try:
                ray.kill(actor)
            except Exception:
                pass
        super().shutdown()
```

python/sglang/srt/managers/scheduler.py
```diff
@@ -1106,6 +1106,45 @@ def _abort_on_running_timeout(self):
                     "Request running timeout reached.", HTTPStatus.SERVICE_UNAVAILABLE
                 )
 
+    def get_init_info(self) -> Dict[str, Any]:
```
xyuzh, line 1109: Refactor only, extracted:
- `configure_scheduler()` — the logging/process-title setup (steps
   1). Extracted from the top of run_scheduler_process().
- `Scheduler.get_init_info()` — returns the init handshake dict
  (step 3). Was inline in `run_scheduler_process()` 
- `Scheduler.run_event_loop()` — sets up `schedule_stream` and calls
   `dispatch_event_loop()` (step 4). Was inline in `run_scheduler_process()`.

So `SchedulerActor` could reuse the code above

python/sglang/srt/entrypoints/engine.py
xyuzh: - online http serving: `_launch_subprocesses `-> `_launch_workers` -> `_launch_scheduler_processes`
- offline: `Engine.__init__` -> `_launch_workers` -> `_launch_scheduler_processes`

xyuzh: Ok, now I have removed `_launch_workers` and made `_launch_subprocesses` the class method to be shared between `RayEngine` and `Engine`

python/sglang/srt/entrypoints/engine.py
```diff
@@ -477,6 +499,201 @@ def rerank(
...
+                )
+
+        return SchedulerInitResult(
+            scheduler_infos=scheduler_infos,
+            wait_for_ready=wait_for_ready,
+            wait_for_completion=wait_for_completion,
+        )
+
+    @classmethod
+    def _launch_subprocesses(
```
xyuzh, line 602: Made `_launch_subprocesses` a class method to share between inherited `RayEngine`

python/sglang/srt/entrypoints/http_server.py
hnyls2002: This name should not start with `_xxx`. This usually means internal util functino or private methods.

hnyls2002 (approved):

Qiaolin-Yu (approved):

python/sglang/srt/entrypoints/http_server.py
xyuzh: `_setup_and_run_http_server` is only going to be used for `ray/http_server.py` and `http_server.py`
also `_execute_server_warmup` is going to be reused the same way in `ray/http_server.py`

PR #17687, opened 2026-01-24
Fix flaky streaming logprobs test by handling detokenizer text buffering

Kangyan-Zhou:
## Summary

Fixes the flaky `test_completion_stream` test in `test_openai_server.py`.

**Root cause**: The detokenizer holds back text at word boundaries during streaming (via `find_printable_text()`) to avoid showing incomplete words. On the final chunk when `finish_reason` is set, this buffered text is flushed. However, by then all logprobs have already been sent in previous chunks, causing the final chunk to have text content but empty logprobs - which broke the test.

**Fix**: Return `None` for logprobs when `finish_reason` is set and all logprobs have been sent. This is semantically correct since no new tokens were generated - the text is just previously buffered text being flushed.

## Changes

- `serving_completions.py`: Return `None` for logprobs on final flush chunk
- `serving_chat.py`: Apply same fix to chat completions streaming endpoint
- `test_openai_server.py`: Update test to handle `logprobs=None` on final chunk
- `scheduler_output_processor_mixin.py`: Remove debug logging from previous commit


python/sglang/srt/entrypoints/openai/serving_completions.py
```diff
@@ -242,19 +242,33 @@ async def _generate_completion_stream(
...
+                    ][n_prev_token:]
+                    finish_reason_for_logprobs = content["meta_info"]["finish_reason"]
+
+                    # When finish_reason is set and all logprobs have been sent,
+                    # any remaining text is just buffered text being flushed by the
+                    # detokenizer (it holds back text at word boundaries). Return None
+                    # for logprobs since no new tokens were generated for this text.
+                    if (
+                        len(output_logprobs_slice) == 0
+                        and finish_reason_for_logprobs is not None
```
JustinTong0323, line 259: Input logprobs discarded when `echo=True` and generation finishes on first chunk
When streaming completions with `echo=True` and logprobs enabled, if the generation finishes on the very first chunk with no output tokens, the input logprobs are incorrectly discarded. (may be used in embedding/rerank scenario)


```suggestion
                        and finish_reason_for_logprobs is not None
                        and input_token_logprobs is None
```

PR #17706, opened 2026-01-25
[bugfix] avoid attention padding tokens computation in pcg

Chen-0210:
## Motivation

When PCG is enabled, the attention metadata is initialized with real_num_tokens, but the input tensor still contains padded tokens. Attention backends such as FlashInfer can not handle this well, which can lead to undefined behavior, including nan value, corrupted outputs (repeated !!!!!), result in abnormally long output lengths.

To fix this, exclude the padded tokens and make PCG more robust.

## Modifications
1.  Reverts  PR #21452
2. Slices tensors to real_num_tokens before/after the attention forward call — in `unified_attention_with_output` and `unified_linear_attention_with_output`, `query/key/value and out_cache_loc` are narrowed to exclude padding tokens before calling the backend, then restored out_cache_loc afterward, so attention only ever sees real tokens.

## Accuracy Tests
Tested in CI
## Checklist

## Review Process

3. Get approvals from [CODEOWNERS](https://github.com/sgl-project/sglang/blob/main/.github/CODEOWNERS) and other reviewers.
4. Trigger CI tests with [comments](https://docs.sglang.io/developer_guide/contribution_guide.html#how-to-trigger-ci-tests) or contact authorized users to do so.


python/sglang/srt/models/qwen3_next.py
```diff
@@ -393,7 +393,7 @@ def forward(
         forward_batch: ForwardBatch,
     ):
         if forward_batch.forward_mode.is_extend() and get_forward_context() is not None:
-            output = torch.empty_like(hidden_states)
+            output = torch.zeros_like(hidden_states)
```
hebiao064: is this change necessary? i think the output will be overridden anyway?

python/sglang/srt/layers/radix_attention.py
```diff
@@ -113,9 +113,9 @@ def forward(
 
         if forward_batch.forward_mode.is_extend() and get_forward_context() is not None:
             if self.qk_head_dim != self.v_head_dim:
-                output = q.new_empty((q.shape[0], self.tp_q_head_num * self.v_head_dim))
+                output = q.new_zeros((q.shape[0], self.tp_q_head_num * self.v_head_dim))
```
Oasis-Git: Why we need to set zero for the buffer? If required, maybe we should set it for linear model or next model and keep empty for normal case.

python/sglang/srt/model_executor/piecewise_cuda_graph_runner.py
```diff
@@ -350,6 +301,7 @@ def warmup_torch_compile(self, num_tokens: int):
                 num_token_non_padded=None,
                 global_forward_mode=ForwardMode.EXTEND,
                 lora_ids=None,
+                real_num_tokens=num_tokens,
```
ch-wan: we can use `num_token_non_padded`

python/sglang/srt/model_executor/forward_batch_info.py
```diff
@@ -375,6 +375,9 @@ class ForwardBatch(ForwardBatchDeepSeekMHAMixin):
     # For hidden states before normal
     return_hidden_states_before_norm: bool = False
 
+    # For piecewise cuda graph
+    real_num_tokens: Optional[int] = None
```
ch-wan: it is equivalent to `num_token_non_padded`

Oasis-Git (commented):
Since these changes affect a broad range of functionality, the full unit test suite should be carefully validated before approving this PR.

python/sglang/srt/layers/radix_attention.py
Oasis-Git: For example, for deepseek v2 `out_cache_loc` might be used within cuda graph?

python/sglang/srt/model_executor/forward_batch_info.py
Chen-0210: great! replace it with `num_token_non_padded_cpu`.

python/sglang/srt/layers/radix_attention.py
Chen-0210: zeros is not a must, removed it

Chen-0210: > For example, for deepseek v2 `out_cache_loc` might be used within cuda graph?

What does this mean?

python/sglang/srt/models/qwen3_next.py
Chen-0210: only the non-padding part is overwritten. But it will not have and side effect, drop it

python/sglang/srt/layers/radix_attention.py
Oasis-Git: https://github.com/sgl-project/sglang/blob/81449b4beebf761eab1ad67ad19e3cab4c0109d4/python/sglang/srt/models/deepseek_v2.py#L1777.

For ds v2 model, out_cache_loc will be included in pcg.

Chen-0210: @Oasis-Git  I see, add out_cache_loc back

python/sglang/srt/model_executor/piecewise_cuda_graph_runner.py
```diff
@@ -198,21 +190,6 @@ def __init__(self, model_runner: ModelRunner):
                 if model_runner.is_hybrid_swa
                 else None
             )
-            self.mamba_track_indices = (
```
hebiao064: why we need to remove these code?

Chen-0210: The relevant part is not included in PCG, so it does not require a fixed memory address.

zminglei:
> image
> previously disabled CI `test_qwen3_next_models_pcg` is now enabled and pass.

Seems previously this CI only has a fixed chance to fail, not fail every time. Could we run the test several times, let's say 10 times to verify it passes every time?

Chen-0210:
@zminglei It seems the re-run didn’t take effect

python/sglang/srt/layers/radix_attention.py
```diff
@@ -152,21 +152,23 @@ def unified_attention_with_output(
...
     kwargs = {}
     if q_rope is not None:
-        kwargs["q_rope"] = q_rope
+        kwargs["q_rope"] = q_rope[:real_num_tokens, :]
     if k_rope is not None:
-        kwargs["k_rope"] = k_rope
+        kwargs["k_rope"] = k_rope[:real_num_tokens, :]
     if sinks is not None:
-        kwargs["sinks"] = sinks
+        kwargs["sinks"] = sinks[:real_num_tokens, :]
```
ispobock: Could you check the shape of attention sinks for models like gpt-oss and mimo? There is no token dim.

Chen-0210: Yes, see that, updated to keep sinks unchanged.

Oasis-Git:
@Chen-0210 Could you fix the ci of stage-a?

Oasis-Git (approved):
I think this is a clean change. approved

ispobock:
@Chen-0210 could you fix the test conflict?

Chen-0210:
> @Chen-0210 could you fix the test conflict?

Done

PR #17714, opened 2026-01-25
Add GLM45 tool interruption support

Leoyzen:
# Motivation
This PR adds support for GLM-4.5 models where tool calls can interrupt reasoning blocks without an explicit closing tag. 
Specifically, GLM-4.5 uses the pattern `<|think|>...<|tool|>` where the occurrence of the tool start token acts as an implicit terminator for the reasoning block. Previously, the `BaseReasoningFormatDetector` expected an explicit `think_end_token`, leading to parsing failures or incorrect reasoning text extraction when tool calls were present.
# Modifications
- **`BaseReasoningFormatDetector`**: Added `tool_start_token` support. Updated `detect_and_parse` and `parse_streaming_increment` to split reasoning text at the last occurrence of the tool start token if the explicit end token is missing.
- **`Glm45Detector`**: Implemented a dedicated detector for GLM-4.5 using:
  - `think_start_token`: `<|think|>`
  - `think_end_token`: `<|/think|>`
  - `tool_start_token`: `<|tool|>`
- **`DetectorMap`**: Registered `Glm45Detector` for the `glm45` model identifier in `ReasoningParser`.
- **Unit Tests**: Added `TestGlm45Detector` with 13 comprehensive test methods covering one-shot parsing, streaming with split tokens, and multiple tool calls.
# Accuracy Tests
Verified with unit tests in `test/registered/parser/test_reasoning_parser.py`:
- `test_detect_and_parse_tool_interrupt`: PASS
- `test_streaming_tool_interrupt_split_tokens`: PASS
- `test_detect_and_parse_multiple_tool_calls_rfind`: PASS
- `test_detect_and_parse_truncated_reasoning`: PASS
# Benchmarking and Profiling
No significant impact on inference speed as the overhead of `rfind` on small buffers is negligible.
# Checklist
- [x] Format your code according to the [Format code with pre-commit](https://docs.sglang.io/developer_guide/contribution_guide.html#format-code-with-pre-commit). (Note: pre-commit hooks were executed; verified locally).


python/sglang/srt/parser/reasoning_parser.py
```diff
@@ -47,17 +49,23 @@ def detect_and_parse(self, text: str) -> StreamingParseResult:
         processed_text = text.replace(self.think_start_token, "").strip()
 
         if self.think_end_token not in processed_text:
-            # Assume reasoning was truncated before `</think>` token
+            # Check for tool_start_token interruption
+            if self._in_reasoning and self.tool_start_token is not None and self.tool_start_token in processed_text:
+                # Find the last occurrence of tool_start_token and split there
+                tool_idx = processed_text.find(self.tool_start_token)
```
harvenstar: Comment above says "last occurrence" but find() gives the first. Should be rfind(). test_detect_and_parse_multiple_tool_calls_rfind will fail as-is.

python/sglang/srt/parser/reasoning_parser.py
```diff
@@ -95,12 +103,19 @@ def parse_streaming_increment(self, new_text: str) -> StreamingParseResult:
...
-            return StreamingParseResult(
-                normal_text=normal_text, reasoning_text=reasoning_text.rstrip()
-            )
+            return StreamingParseResult(normal_text=normal_text, reasoning_text=reasoning_text.rstrip())
 
         # Continue with reasoning content
         if self._in_reasoning:
+            # Check for tool_start_token interruption
+            if self.tool_start_token and self.tool_start_token in current_text:
+                tool_idx = current_text.find(self.tool_start_token)
```
harvenstar, line 148: Same thing — rfind().

python/sglang/srt/parser/reasoning_parser.py
```diff
@@ -47,17 +49,23 @@ def detect_and_parse(self, text: str) -> StreamingParseResult:
         processed_text = text.replace(self.think_start_token, "").strip()
 
         if self.think_end_token not in processed_text:
-            # Assume reasoning was truncated before `</think>` token
+            # Check for tool_start_token interruption
+            if self._in_reasoning and self.tool_start_token is not None and self.tool_start_token in processed_text:
```
harvenstar: Should be in_reasoning (local var) instead of self._in_reasoning? — the latter is False when force_reasoning=False, so this branch is unreachable even when <think> is in the text.

python/sglang/srt/parser/reasoning_parser.py
harvenstar, line 78: Comment above says "last occurrence" but find() gives the first. Should be rfind(). test_detect_and_parse_multiple_tool_calls_rfind will fail as-is.

harvenstar (changes requested):

harvenstar (approved):

JustinTong0323 (approved):

PR #17736, opened 2026-01-26
Fix streaming logprobs returning empty tokens mid-stream

Kangyan-Zhou:
## Summary

Fixes the remaining flaky `test_completion_stream` test that PR #17687 didn't fully address.

**Root cause**: PR #17687 fixed the case where empty logprobs were returned on the final chunk (when `finish_reason` was set). However, the detokenizer can also flush buffered text **mid-stream** when `finish_reason` is still `None`, causing the same `IndexError: list index out of range` when the test tries to access `tokens[0]` on an empty list.

**Fix**: Return `logprobs=None` whenever there are no new logprobs to report, regardless of whether `finish_reason` is set.

## Changes

- `serving_completions.py`: Remove the `finish_reason is not None` requirement - return `logprobs=None` when `output_logprobs_slice` is empty
- `serving_chat.py`: Remove the `or finish_reason is None` condition that was causing empty logprobs to be processed mid-stream  
- `utils.py`: Remove debug logging added in previous commit
- `test_openai_server.py`: Remove debug logging and update comments to clarify that `logprobs=None` can happen both mid-stream and on the final chunk

## Comparison

| State | Server Behavior | Test |
|-------|----------------|------|
| Before #17687 | Always returns `Logprobs(tokens=[], ...)` even with empty slice | Assumes logprobs always has tokens → **FLAKY** |
| After #17687 | Returns `None` only when `finish_reason is not None` AND slice is empty | Handles `None` for final chunk only → **STILL FLAKY** |
| This PR | Returns `None` whenever slice is empty (regardless of `finish_reason`) | Handles `None` for any chunk → **CORRECT** |

### DIFF
--- a/python/sglang/srt/entrypoints/openai/serving_chat.py
+++ b/python/sglang/srt/entrypoints/openai/serving_chat.py
@@ -634,11 +634,12 @@ class OpenAIServingChat(OpenAIServingBase):
                     total_output_logprobs = len(
                         content["meta_info"]["output_token_logprobs"]
                     )
-                    # When finish_reason is set and all logprobs have been sent,
-                    # any remaining text is just buffered text being flushed by the
-                    # detokenizer (it holds back text at word boundaries). Return None
-                    # for logprobs since no new tokens were generated for this text.
-                    if n_prev_token < total_output_logprobs or finish_reason is None:
+                    # When all logprobs have been sent, any remaining text is just
+                    # buffered text being flushed by the detokenizer (it holds back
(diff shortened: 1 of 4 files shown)


hnyls2002:
@Kangyan-Zhou The mid-stream empty-logprobs case this PR addresses was fixed on main by #21030, which removed the finish_reason conditions in both serving_chat.py and serving_completions.py and updated test_openai_server.py to accept logprobs=None on any chunk. Closing as superseded; please reopen if you can still reproduce the flaky test on current main.

PR #17744, opened 2026-01-26
Fix OOM in DeepSeek weight loading by deferring dict(weights) materialization

hsuchifeng:
## Summary

- Fix OOM when loading DeepSeek models with streaming/remote weight loading
- Defer `dict(weights)` call until quantization is actually needed
- Return original weights iterator when no FP8 UE8M0 quantization is required

## Problem

The `_maybe_quant_weights_to_fp8_ue8m0` function unconditionally calls `dict(weights)` at the start, which materializes **all** weights into memory even when no quantization is needed (which is the common case when `SGLANG_NVFP4_CKPT_FP8_GEMM_IN_ATTN` is not set).

This causes OOM when:
1. Using streaming/remote weight loading where weights are loaded on-demand
2. Loading large models like DeepSeek-V3 (~300GB) on limited GPU memory

**Before fix:**
```
weights iterator → dict(weights) → 300GB all loaded at once → OOM!
```

**After fix:**
```
weights iterator → check partial_names → empty? → return iterator directly (streaming)
                                       → non-empty? → dict(weights) → quantize
```

## Error Log

```
torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 20.00 MiB. GPU 5 has a total capacity of 95.08 GiB of which 1.12 MiB is free. Including non-PyTorch memory, this process has 95.07 GiB memory in use. Of the allocated memory 83.95 GiB is allocated by PyTorch, and 9.65 GiB is reserved by PyTorch but unallocated. If reserved but unallocated memory is large try setting PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True to avoid fragmentation.  See documentation for Memory Management  (https://pytorch.org/docs/stable/notes/cuda.html#environment-variables)
```

Full stack trace:
```
[2026-01-26 03:41:43 TP4] Scheduler hit an exception: Traceback (most recent call last):
  File "/sgl-workspace/sglang/python/sglang/srt/managers/scheduler.py", line 2915, in run_scheduler_process
    scheduler = Scheduler(
                ^^^^^^^^^^
  File "/sgl-workspace/sglang/python/sglang/srt/managers/scheduler.py", line 333, in __init__
    self.init_model_worker()
  File "/sgl-workspace/sglang/python/sglang/srt/managers/scheduler.py", line 471, in init_model_worker
    self.tp_worker = TpModelWorker(
                     ^^^^^^^^^^^^^^
  File "/sgl-workspace/sglang/python/sglang/srt/managers/tp_worker.py", line 253, in __init__
    self._model_runner = ModelRunner(
                         ^^^^^^^^^^^^
  File "/sgl-workspace/sglang/python/sglang/srt/model_executor/model_runner.py", line 356, in __init__
    self.initialize(min_per_gpu_memory)
  File "/sgl-workspace/sglang/python/sglang/srt/model_executor/model_runner.py", line 433, in initialize
    self.load_model()
  File "/sgl-workspace/sglang/python/sglang/srt/model_executor/model_runner.py", line 896, in load_model
    self.model = self.loader.load_model(
                 ^^^^^^^^^^^^^^^^^^^^^^^
  File "/sgl-workspace/sglang/python/sglang/srt/model_loader/loader.py", line 2381, in load_model
    self._load_model_from_remote_fs(
  File "/sgl-workspace/sglang/python/sglang/srt/model_loader/loader.py", line 2338, in _load_model_from_remote_fs
    model.load_weights(self._get_weights_iterator_fs(client))
  File "/sgl-workspace/sglang/python/sglang/srt/models/deepseek_v2.py", line 3654, in load_weights
    weights = self._maybe_quant_weights_to_fp8_ue8m0(
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/sgl-workspace/sglang/python/sglang/srt/models/deepseek_v2.py", line 3976, in _maybe_quant_weights_to_fp8_ue8m0
    weights_dict = dict(weights)
                   ^^^^^^^^^^^^^
  ...
torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 20.00 MiB. GPU 4 has a total capacity of 95.08 GiB of which 1.12 MiB is free.
```

## Test plan

- [x] Tested with DeepSeek-V3-W4AFP8 model on 8x H20 GPUs with remote weight loading


python/sglang/srt/models/deepseek_common/deepseek_weight_loader.py
```diff
@@ -655,16 +655,24 @@ def _maybe_quant_weights_to_fp8_ue8m0(
...
-                partial_names, desc="quant weights to fp8 ue8m0"
-            ):
-                original_weight = weights_dict[f"{partial_name}.weight"]
-                out_w, out_s = quant_weight_ue8m0(
-                    original_weight, weight_block_size=weight_block_size
-                )
-                weights_dict[f"{partial_name}.weight"] = out_w
-                weights_dict[f"{partial_name}.weight_scale_inv"] = out_s
+        # Early return if no quantization needed - avoid materializing all weights into memory
+        if not partial_names:
```
Fridge003, line 659: We don't need line 660 to line 663, since partial_names won't be empty when nextn/enable_nextn_moe_bf16_cast_to_fp8 are both enabled

Fridge003 (changes requested):

python/sglang/srt/models/deepseek_common/deepseek_weight_loader.py
hsuchifeng: Thanks for the review. Fixed.

Fridge003 (approved):

PR #17754, opened 2026-01-26
Feature:Reserve HTTP server port before model loading to immediately detect port  conflicts instead of failing after several minutes of model loading.

xrwang8:
## Motivation

When users specify a port for the inference server (e.g., `--port 8000`), if that port is already in use, the error only appears after the model finishes loading, which can take several minutes. This wastes time and provides poor user experience.

The goal is to detect port conflicts immediately before model loading, providing fast feedback to users.

## Modifications

- Add `_create_server_socket()` function to bind and reserve the HTTP port early
- Add `_is_valid_ipv6_address()` helper function for IPv6 address detection
- Modify `launch_server()` to bind port before calling `_launch_subprocesses()`
- Set `SO_REUSEADDR` socket option to allow quick restart after crashes (avoids TIME_WAIT issues)
- Set `SO_REUSEPORT` socket option for multi-worker compatibility
- Provide clear, actionable error messages with diagnostic commands

**Key implementation details:**
- Port is bound at the very start of `launch_server()`
- Socket is held during subprocess launch to protect against port being claimed
- Socket is closed before uvicorn starts (SO_REUSEADDR allows immediate rebinding)
- Supports both IPv4 and IPv6 addresse
## Accuracy Tests

N/A - This change does not affect model outputs.

## Checklist

## Review Process


python/sglang/srt/entrypoints/http_server.py
```diff
@@ -1839,15 +1913,27 @@ def launch_server(
...
+            )
         )
-    )
+    except Exception:
+        reserved_socket.close()
+        raise
+
+    # Release the reserved socket so uvicorn can bind to the same port.
+    # The SO_REUSEADDR option we set earlier allows uvicorn to bind immediately.
+    reserved_socket.close()
```
bjacobgordon: Binding the socket directly to uvicorn using the [`fd` option](https://uvicorn.dev/settings/#socket-binding) could be a more conventional move here. I believe uvicorn copies the socket object, so we could defer closing the reserved_socket until after `uvicorn.run` returns. What do you think?

python/sglang/srt/entrypoints/http_server.py
```diff
@@ -1816,6 +1818,78 @@ def _wait_weights_ready():
     )
 
 
+def _is_valid_ipv6_address(address: str) -> bool:
```
bjacobgordon: [This](https://github.com/sgl-project/sglang/blob/8acd4d7d7e6f436601ef3ae51678f130fdc04d25/python/sglang/srt/utils/common.py#L2500) is available. Would it fit this use case despite having a different implementation?

python/sglang/srt/entrypoints/http_server.py
```diff
@@ -1816,6 +1818,78 @@ def _wait_weights_ready():
...
+    """Check if the given address is a valid IPv6 address."""
+    try:
+        socket.inet_pton(socket.AF_INET6, address)
+        return True
+    except OSError:
+        return False
+
+
+def _create_server_socket(host: str, port: int) -> socket.socket:
```
bjacobgordon: [This](https://github.com/sgl-project/sglang/blob/8acd4d7d7e6f436601ef3ae51678f130fdc04d25/python/sglang/utils.py#L377) is available. It doesn't have any provisions for IPv6, but maybe it could updated? Or a sibling method could be created?

bjacobgordon (commented):
Hey, @xrwang8! I'm not a decision-maker here, but I _am_ currently working on refactoring the `http_server` as a whole, so just mentioning a few points here that might help avoid extra debt and cognitive load right out of the gate. Happy to discuss!

python/sglang/srt/entrypoints/http_server.py
bjacobgordon: This function is packaged well enough that we could probably keep it next to [this one](https://github.com/sgl-project/sglang/blob/f8636fbb253a83d268ffa1636eac4e111966376c/python/sglang/utils.py#L377)! This would help keep the focus of "/http_server.py" on just concerns that are unique to the concept of "the sglang http server". What do you think?

python/sglang/srt/entrypoints/http_server.py
```diff
@@ -1816,6 +1819,71 @@ def _wait_weights_ready():
     )
 
 
+def _create_server_socket(host: str, port: int) -> socket.socket:
```
bjacobgordon: This function is packaged well enough that we could probably keep it next to [this one](https://github.com/sgl-project/sglang/blob/f8636fbb253a83d268ffa1636eac4e111966376c/python/sglang/utils.py#L377)! This would help keep the focus of "/http_server.py" on just concerns that are unique to the concept of "the sglang http server". What do you think?

bjacobgordon: The name `_create_server_socket` is a good start! Let's iterate on it:
- `create` doesn't give your implementation enough credit, it does way more than that! We could say `reserve`, but it's the port being reserved, not the socket. I believe `prebind` would more accurately describe what's happening.
- `server` is accurate, but since we're server-side, those are really the only kind of sockets we can mess with! So, the thing to ask ourselves is "what kind of server socket?". I believe the operative term here would be `listening`.

So, the new name could be `_prebind_listening_socket`. What do you think?

bjacobgordon (commented):
@xrwang8 Here's my next round of thoughts!

python/sglang/utils.py
```diff
@@ -395,6 +395,75 @@ def reserve_port(host, start=30000, end=40000):
     raise RuntimeError("No free port available.")
 
 
+def _prebind_listening_socket(host: str, port: int) -> socket.socket:
```
Kangyan-Zhou, line 412: since it's a public function, should we remove the leading _?

Kangyan-Zhou (approved):

python/sglang/utils.py
xrwang8: I think we should keep the underscore - this is a low-level socket binding utility used internally during server initialization. Making it public would add unnecessary API surface area and create backward compatibility constraints for what's essentially an implementation detail.

Kangyan-Zhou:
### Code review

Found 1 issue:

1. `NameError` in `finally` block masks original exception when `tokenizer_worker_num > 1` (bug due to restructured try/finally scope)

The PR moves `_launch_subprocesses()` and the `multi_tokenizer_args_shm = write_data_for_multi_tokenizer(...)` assignment inside the `try` block (line 1896). Previously, both were outside the `try` block, guaranteeing `multi_tokenizer_args_shm` was always defined before the `finally` ran. Now, if any exception is raised before line 1962 (e.g., `_launch_subprocesses()` fails at line 1899) when `tokenizer_worker_num > 1`, the `finally` block at line 2006 will attempt `multi_tokenizer_args_shm.unlink()` on an undefined variable, raising `UnboundLocalError` and masking the original exception.

Fix: initialize `multi_tokenizer_args_shm = None` before the `try` block and guard the cleanup with `if multi_tokenizer_args_shm is not None:`.

https://github.com/sgl-project/sglang/blob/a2da7b9044d8ad1688f729256694e1fcfca8c3ee/python/sglang/srt/entrypoints/http_server.py#L1999-L2008

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #17760, opened 2026-01-26
feat: add EAGLE3 support for MiniCPMV4_0

Lollipop:
## Summary

Add EAGLE3 speculative decoding support for MiniCPM-V-4 model.

- Add `get_embed_and_head()` method to get embedding and lm_head weights for EAGLE3
- Add `set_eagle3_layers_to_capture()` method to set layers for speculative decoding
- Include fallback mechanism to ensure compatibility with different LLM implementations

## Changes

- `python/sglang/srt/models/minicpmv.py`: Added 2 new methods to `MiniCPMV4_0` class


python/sglang/srt/models/minicpmv.py
```diff
@@ -932,6 +932,28 @@ def pad_input_ids(self, input_ids: List[int], image_inputs: MultimodalInputs):
 
         return pattern.pad_input_tokens(input_ids, image_inputs)
 
+    def get_embed_and_head(self):
+        """Get embedding and lm_head weights for EAGLE3 support."""
+        if hasattr(self.llm, "get_embed_and_head"):
```
JustinTong0323: Why do we need a fallback? Could the LLM module `LlamaForCausalLM` be changed to another module?

Lollipop: The fallback was added defensively, but since all LLM modules in sglang that support EAGLE3 already implement this method, it's not necessary.
I will remove the fallback

Issue #17526, opened 2026-01-21
[Tracking] GLM 4.5/4.6/4.7 Blackwell performance optimization


b8zhong:
## Motivation

Focus on GLM 4.7 FP4 (mainly for Blackwell) and latency scenario. 

@ynwang007 @pdasgup @vincentzed 

# Items

- [x] #17166 GLM + `flashinfer_trtllm` MoE backend (10% perf gain), needs Flashinfer 0.6.2 or nightly to be automatically set
- [ ] #16755 Make sure MHA + `topk=1` speculative decoding still sets `trtllm_mha` as attention backend (10% perf gain)
- [x] Check #17237, GLM has `--enable-flashinfer-allreduce-fusion` auto-enabled
- [ ] #17773 Fuse the FP8 KV buffer kernel with Q cast to FP8 under FP8 attention (est. 3% performance gain) <img width="655" height="188" alt="Image" src="https://github.com/user-attachments/assets/43da326b-6dab-4a6a-9243-619544345992" />
- [ ] Improve sgl-kernel `scaled_fp4_quant`, perhaps switch to better maintained CUTLASS backend of `flashinfer.fp4_quantization.fp4_quantize` (very small of E2E but maybe 1-2%)
- [ ] Use better scale layout for `--fp4-gemm-backend` for very low concurrency (maybe backend: `flashinfer_trtllm`) (maybe 5-10%)
- [ ] #18801 Use CuteDSL FP4 GEMM (4% perf gain)
- [ ] Fuse two elementwise in FlashinferFP4MoE (not sure yet what they are) <img width="761" height="172" alt="Image" src="https://github.com/user-attachments/assets/b00e94ab-75ce-49e8-ad89-3de548d3f436" />
- [x] #12787 (General but include because important)  can achieve better performance in single node as well, to replace TRT Allreduce Fusion
- [x] Seems FP8 KV cache is always slower due to extra quantize (?) see if we can mitigate this (52 us vs 47 us). Even at concurrency=1024, it's still not better (output throughput of 4384.482 vs 4893.194) tps. The attention kernel takes less time, as expected. The slowdown is caused by extra quantization and data-shuffling kernels including DeviceGemmFp4GemmSm100, cvt_fp16_to_fp4, float8_copy_kernel_cuda, and _fused_fp8_set_kv_buffer_kernel. These added steps and their launch overhead total about 23us

BF16:
[...]

PR #17773, opened 2026-01-26
[Kernel] Fuse Q FP8 cast into KV buffer kernel for TRTLLM MHA backend

harvenstar:
## Summary                                                                              
  - Fuse the FP8 cast operation for Q tensor into the existing KV buffer kernel for TRTLLM
   MHA backend                                                                            
  - This eliminates a separate `q.to(fp8)` kernel launch, reducing kernel launch overhead 
  and improving memory bandwidth utilization                                              
  - Part of #17526 (Blackwell performance optimization)                                   
                                                                                          
  ## Motivation                                                                           
  In the original implementation, FP8 attention requires two separate operations:         
  1. `_fused_fp8_set_kv_buffer_kernel`: quantize K/V to FP8 and write to paged KV cache   
  2. `q.to(torch.float8_e4m3fn)`: cast Q to FP8 (via PyTorch elementwise kernel)          
                                                                                          
  This PR fuses both operations into a single Triton kernel, eliminating the extra kernel 
  launch overhead.                                                                        
                                                                                          
  ## Changes                                                                              
                                                                                          
  ### Kernel Design (`trtllm_fp8_kv_kernel.py`)                                           
  - Added `_fused_fp8_set_qkv_buffer_kernel` with precise block mapping to avoid empty    
  thread blocks in GQA models:                                                            
    - Grid: `(num_tokens, 2 * Bkv + Bq)` where `Bkv = ceil(num_kv_heads / BLOCK_HEAD)`,   
  `Bq = ceil(num_q_heads / BLOCK_HEAD)`                                                   
    - Block mapping: `[0, Bkv)` → K, `[Bkv, 2*Bkv)` → V, `[2*Bkv, 2*Bkv+Bq)` → Q          
  - Added `_process_q_tensor` helper for Q FP8 cast (equivalent to `q.to(fp8)`)           
  - Q cast produces bitwise-identical results to `q.to(torch.float8_e4m3fn)`              
                                                                                          
  ### Backend Changes (`trtllm_mha_backend.py`)                                           
  - Added `q_fp8_buffer` for pre-allocated Q output buffer                                
  - Added `_inv_scale_cache` to cache pre-computed inverse scale tensors per layer        
  - Added `_ensure_q_fp8_buffer()` for lazy allocation in non-CUDA-graph mode             
  - Added `_get_inv_scale_tensors()` to avoid tensor allocation during CUDA graph capture 
  - Modified `forward_decode` and `forward_extend` to use fused QKV path when FP8         
  attention is enabled                                                                    
                                                                                          
  ### CUDA Graph Compatibility                                                            
  - `q_fp8_buffer` is pre-allocated in `init_cuda_graph_state()`                          
  - Inverse scale tensors are computed once and cached in `_inv_scale_cache`              
  - No tensor allocation occurs during forward pass, ensuring CUDA graph capture works    
  correctly                                                                               
                                                                                          
  ## Performance                                                                          
  Tested on GLM-4.7 with 4xB300, FP8 attention, TP=4:                                     
                                                                                          
  | Metric | Before | After | Improvement |                                               
  |--------|--------|-------|-------------|                                               
  | Output throughput | 990 tok/s | 1019 tok/s | +2.9% |                                  
  | Q+KV kernel time | 4262 us | 1483 us | -65% |                                         
                                                                                          
  Profile comparison (920 calls per decode step on TP0):                                  
  - **Before**: `_fused_fp8_set_kv_buffer_kernel` (1432us) + `q.to(fp8)` elementwise      
  (2830us) = 4262us                                                                       
  - **After**: `_fused_fp8_set_qkv_buffer_kernel` (1483us)                                
                                                                                          
  ### Before                                                                              
  beforechange                                                        
                                                                                          
  ### After                                                                               
   afterchange

### DIFF
--- a/python/sglang/srt/layers/attention/triton_ops/trtllm_fp8_kv_kernel.py
+++ b/python/sglang/srt/layers/attention/triton_ops/trtllm_fp8_kv_kernel.py
@@ -109,7 +109,6 @@ def _process_q_tensor(
         head_offsets = head_idx + tl.arange(0, BLOCK_HEAD)
         dim_offsets = dim_idx + tl.arange(0, BLOCK_DIM)
 
-        # Use real boundaries for masking (avoid Python min() with Triton scalars)
         head_mask = head_offsets < num_q_heads
         dim_mask = dim_offsets < head_dim
         mask = head_mask[:, None] & dim_mask[None, :]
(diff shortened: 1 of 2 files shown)


b8zhong (approved):

python/sglang/srt/layers/attention/triton_ops/trtllm_fp8_kv_kernel.py
```diff
@@ -502,3 +717,93 @@ def _naive_fp8_set_kv_buffer(
...
+        v_3d.stride(0),
+        v_3d.stride(1),
+        v_3d.stride(2),
+        v_cache_stride_page,
+        v_cache_stride_offset,
+        v_cache_stride_head,
+        v_cache_stride_dim,
+        BLOCK_HEAD=BLOCK_HEAD,
+        BLOCK_DIM=BLOCK_DIM,
+    )
```
Fridge003, line 781: Can you please add a kernel test in this file. 
Maybe under __main__ function

python/sglang/srt/layers/attention/triton_ops/trtllm_fp8_kv_kernel.py
```diff
@@ -201,6 +252,170 @@ def _fused_fp8_set_kv_buffer_kernel(
...
+    # Tensor dimensions
+    num_q_heads: tl.constexpr,
+    num_kv_heads: tl.constexpr,
+    head_dim: tl.constexpr,
+    page_size: tl.constexpr,
+    # Number of head blocks for mapping
+    num_kv_head_blocks: tl.constexpr,  # Bkv = ceil(num_kv_heads / BLOCK_HEAD)
+    num_q_head_blocks: tl.constexpr,  # Bq = ceil(num_q_heads / BLOCK_HEAD)
+    # Strides for Q input [num_tokens, num_q_heads, head_dim]
+    q_stride_token: tl.constexpr,
```
Fridge003, line 278: We can assert the input and output ptr of q/k/v shares the same stride, head number and dimension, so as to reduce the number of kernel arguments.

The current code is hard to read

Fridge003 (changes requested):

python/sglang/srt/layers/attention/triton_ops/trtllm_fp8_kv_kernel.py
harvenstar: sure, i will assume the input and output ptr of q/k/v shares the same stride

python/sglang/srt/layers/attention/triton_ops/trtllm_fp8_kv_kernel.py
harvenstar: Done

python/sglang/srt/layers/attention/trtllm_mha_backend.py
```diff
@@ -428,35 +448,77 @@ def get_cuda_graph_seq_len_fill_value(self) -> int:
...
+            q=q,
             k=k,
             v=v,
             k_cache=k_cache,
             v_cache=v_cache,
             cache_loc=cache_loc,
-            k_scale=layer.k_scale,  # May be None
-            v_scale=layer.v_scale,  # May be None
+            q_out=q_out,
+            inv_k_scale=inv_k_scale,  # Pre-computed tensor or None
```
Fridge003: Scale is a float rather than a tensor. We need to handle it as float if it's not None.
In this case we don't need `_get_inv_scale_tensors` and the cache for inv_scale

Fridge003 (changes requested):

b8zhong:
I think this creations maybe need to be outside forward context, right?

Screenshot 2026-02-13 at 12 12 44 PM

b8zhong:
Let's move to https://github.com/sgl-project/sglang/pull/29690

PR #17781, opened 2026-01-26
[sglang] fix mm token padded value overlap with text token id

bixue2010:
## Motivation

Currently mm token padded value could overlaop with text token id range, and in a rare case, when it's overlapped with token id, and accidentially there's an exact token id in the prompt, this will be counted as MM token id, when during the check at

python/sglang/srt/managers/mm_utils.py#L849-L851

```
 raise RuntimeError( 
     f"Insufficient multimodal embedding length: {num_mm_tokens_in_input_ids=} vs {num_mm_tokens_in_embedding=}. This is an internal error" 
 ) 
 ```
, throw exception:
```
num_mm_tokens_in_input_ids=1531 vs num_mm_tokens_in_embedding=1530
```

## Modifications

mm pad

## Accuracy Tests

unit test

## Benchmarking and Profiling

n/a

## Checklist

## Review Process

### DIFF
--- a/python/sglang/srt/managers/schedule_batch.py
+++ b/python/sglang/srt/managers/schedule_batch.py
@@ -43,6 +43,7 @@ import logging
 import re
 import time
 from enum import Enum, auto
+from functools import lru_cache
 from http import HTTPStatus
 from itertools import chain
 from typing import TYPE_CHECKING, Any, List, Optional, Set, Tuple, Union
@@ -95,8 +96,26 @@ if TYPE_CHECKING:
 
 INIT_INCREMENTAL_DETOKENIZATION_OFFSET = 5
 
+# Constant used as the base offset for MM (multimodal) pad values.
+# This ensures pad_values don't overlap with valid text token IDs.
+MM_PAD_SHIFT_VALUE = 1_000_000
 
 logger = logging.getLogger(__name__)
+
+
+@lru_cache(maxsize=1)
+def sanity_check_mm_pad_shift_value(vocab_size: int) -> None:
+    if vocab_size > MM_PAD_SHIFT_VALUE:
+        raise ValueError(
+            f"Model vocab_size ({vocab_size}) exceeds MM_PAD_SHIFT_VALUE ({MM_PAD_SHIFT_VALUE}). "
+            f"MM pad_values may overlap with valid token IDs. "
+            f"Please increase MM_PAD_SHIFT_VALUE in schedule_batch.py."
+        )
+
+
+def _compute_pad_value(hash: int) -> int:
+    """Compute pad value from hash."""
+    return MM_PAD_SHIFT_VALUE + (hash % (1 << 30))
 
 
 class BaseFinishReason:
@@ -262,7 +281,7 @@ class MultimodalDataItem:
             import uuid
 
             self.hash = uuid.uuid4().int
-            self.pad_value = self.hash % (1 << 30)
+            self.pad_value = _compute_pad_value(self.hash)
             return
         if self.hash is None:
             if self.feature is not None:
@@ -271,7 +290,7 @@ class MultimodalDataItem:
                 hashed_feature = self.precomputed_embeddings
             self.hash = hash_feature(hashed_feature)
         assert self.hash is not None
-        self.pad_value = self.hash % (1 << 30)
+        self.pad_value = _compute_pad_value(self.hash)
 
     def is_modality(self, modality: Modality) -> bool:
         return self.modality == modality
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: python/sglang/srt/managers/schedule_batch.py [hunk 1]
```diff
@@ -258,11 +258,22 @@ def set_pad_value(self):
...
+            except Exception:
+                vocab_size = 1_000_000
+        else:
+            vocab_size = 1_000_000
+
         if envs.SGLANG_MM_SKIP_COMPUTE_HASH.get():
             import uuid
 
             self.hash = uuid.uuid4().int
-            self.pad_value = self.hash % (1 << 30)
```
yinghai, line 265: why we constraint the pad_value to be ( 1<< 30) ?

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -258,11 +258,22 @@ def set_pad_value(self):
 
         from sglang.srt.managers.mm_utils import hash_feature
 
+        # Get vocab_size from model config to ensure MM pad_values don't overlap with text tokens
+        # Fallback to 1M if not available (e.g., during early initialization)
```
fzyzcjy: qq: will this cause inconsistent hash and cause later issues

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -258,11 +258,22 @@ def set_pad_value(self):
...
+        server_args = get_global_server_args()
+        if server_args is not None:
+            try:
+                vocab_size = server_args.get_model_config().vocab_size
+            except Exception:
```
fzyzcjy: nit: what Exception will we get, it would be great to not catch such a general exception to avoid hiding bugs

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -258,11 +258,22 @@ def set_pad_value(self):
 
         from sglang.srt.managers.mm_utils import hash_feature
 
+        # Get vocab_size from model config to ensure MM pad_values don't overlap with text tokens
+        # Fallback to 1M if not available (e.g., during early initialization)
+        server_args = get_global_server_args()
+        if server_args is not None:
+            try:
+                vocab_size = server_args.get_model_config().vocab_size
```
fzyzcjy: nit: get_model_config may be a little bit non-lightweight, wondering whether we could do sth more lightweight

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -258,11 +258,22 @@ def set_pad_value(self):
...
+            except Exception:
+                vocab_size = 1_000_000
+        else:
+            vocab_size = 1_000_000
```
fzyzcjy: btw, what if we do sth like:

```
PAD_SHIT_VALUE = 1_000_000

def sanity_check_pad_shift_value():
  assert get real vocab size <= pad_shift_value

call sanity_check_pad_shift_value at e.g. ModelRunner.__init__ or whatever init phase
```

python/sglang/srt/managers/schedule_batch.py
bixue2010: @fzyzcjy  do you know?

fzyzcjy: no... maybe code blame to see who wrote that

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -271,7 +283,7 @@ def set_pad_value(self):
                 hashed_feature = self.precomputed_embeddings
             self.hash = hash_feature(hashed_feature)
         assert self.hash is not None
-        self.pad_value = self.hash % (1 << 30)
+        self.pad_value = MM_PAD_SHIFT_VALUE + (self.hash % (1 << 30))
```
fzyzcjy: btw nit:

def _compute_pad_value(hash: int): return MM_PAD_SHIFT_VALUE + (hash % (1 << 30))

to avoid dup two times, and may help when later we do 64 bit hash or whatever

fzyzcjy (approved):
LGTM w/ a tiny nit

JustinTong0323 (approved):

mickqian (approved):

python/sglang/srt/managers/schedule_batch.py
```diff
@@ -95,10 +95,27 @@
...
 INIT_INCREMENTAL_DETOKENIZATION_OFFSET = 5
 
+# Constant used as the base offset for MM (multimodal) pad values.
+# This ensures pad_values don't overlap with valid text token IDs.
+MM_PAD_SHIFT_VALUE = 1_000_000
 
 logger = logging.getLogger(__name__)
 
 
+def sanity_check_mm_pad_shift_value(vocab_size: int) -> None:
```
mickqian, line 107: something like `@lru_cache(maxsize=1)` could be helpful

python/sglang/srt/managers/schedule_batch.py
liusy58: Hi, @yinghai  I added this to keep it consistent with the existing logic (as seen in line 274), which also applies `% (1 << 30)` to the hash

Issue #15299, opened 2025-12-17
[RFC] Unifying Multi-Vendor Support via a Platform Interface in sglang


yeahdongcn:
### Background

As part of ongoing efforts to extend sglang to additional GPU vendors (including MThreads GPU), we have been carefully reviewing the current multi-vendor implementation. While doing so, we observed that much of the existing codebase relies on scattered `_is_xxx`–style checks to distinguish between different platforms. These conditionals appear in many files and tend to grow as more vendors are added, which increases both cognitive load and long-term maintenance cost.

In contrast, `multimodal_gen` has already introduced a `Platform` abstraction, where platform-specific logic is encapsulated behind a unified interface and accessed via `current_platform`. This approach appears more concise and scalable, and may be a good fit for other parts of sglang as well.

Based on an examination of the srt code path and existing vendor-specific implementations, this RFC proposes extending the `Platform` abstraction into the sglang srt layer.

---

### Proposal

Introduce and consistently use a `Platform` interface within sglang srt, consolidating platform-specific logic that is currently spread across conditional checks. In addition to existing platform properties and methods, the following extensions are proposed.

---

### 1. One-time Platform Initialization (`init_platform`)

Some platforms (e.g. NPU, MUSA) require one-time setup steps such as monkey-patching PyTorch or performing special imports. Today, this logic is typically guarded by ad-hoc conditionals.

We propose adding a `init_platform()` (run-once) hook to the `Platform` interface. After resolving `current_platform`, sglang can explicitly invoke this method to ensure that all required imports and patches are applied exactly once and in a well-defined location.

This makes platform initialization:

* Explicit
* Centralized
* Easier to reason about and maintain

---

### 2. Platform-specific Server Argument Post-processing

Currently, server arguments are post-processed through multiple platform-specific handlers, for example:

```python
self._handle_hpu_backends()
self._handle_cpu_backends()
self._handle_npu_backends()
```

With a richer `Platform` interface, this logic could be moved into the platform implementation itself. For example, sglang would call a single hook such as:

```python
current_platform.postprocess_server_args(args)
```

This avoids proliferating platform checks and keeps backend-specific behavior localized within the corresponding platform implementation.

---

### 3. Platform-specific Op / Kernel Resolution

In files such as `python/sglang/srt/layers/activation.py`, platform checks are currently used at import time to decide which ops to import from `sgl-kernel`.

[...]

PR #17803, opened 2026-01-27
Example for [RFC] Unifying Multi-Vendor Support via a Platform Interface in sglang

yeahdongcn:
## Motivation

To add a naive example for https://github.com/sgl-project/sglang/issues/15299

## Checklist

## Review Process

### DIFF
--- a/python/sglang/srt/platforms/cuda.py
+++ b/python/sglang/srt/platforms/cuda.py
@@ -1,0 +1,119 @@
+# Copyright 2023-2024 SGLang Team
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+"""
+NVIDIA CUDA Platform Implementation
+====================================
+
+This file demonstrates the standard pattern for implementing a Platform.
+Use this as a reference when adding support for new hardware vendors.
+
+THE OP REGISTRY PATTERN:
+------------------------
+Each platform maintains a lazy-loaded dictionary (_ops) that maps operation
+names (strings) to their platform-specific implementations (callables).
+
+    _ops = {
+        "silu_and_mul": sgl_kernel.silu_and_mul,  # CUDA kernel
+        "gelu_and_mul": sgl_kernel.gelu_and_mul,  # CUDA kernel
+        ...
+    }
+
+The registry is populated on first access via _init_ops(), which:
+1. Imports the platform-specific kernel library (sgl_kernel for CUDA)
+2. Returns a dict mapping op names to kernel functions
+
+WHY LAZY LOADING?
+-----------------
+Kernel imports can be expensive (loading .so files, GPU initialization).
+By deferring _init_ops() until an op is actually requested:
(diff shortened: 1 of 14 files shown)


python/sglang/srt/platforms/cuda.py
```diff
@@ -0,0 +1,119 @@
...
+
+        This method is called once, on the first op lookup.
+        All sgl_kernel imports happen here to keep startup fast.
+
+        Returns:
+            Dict mapping op names to sgl_kernel functions.
+        """
+        # Import CUDA kernels from sgl_kernel
+        # These are compiled CUDA extensions providing optimized implementations
+        from sgl_kernel import gelu_and_mul as cuda_gelu_and_mul
```
DarkSharpness: 1. Can we support lazy import here? This could reduce import time, and avoid potential import errors.
2. Can we support out-of-class registration? e.g. register an op out of this file for given platform (similar to the dispatch mechanism of PyTorch)

yeahdongcn: Good questions and thanks for raising them.
I think both are technically feasible, but there are some trade-offs worth calling out.

1. Lazy import

Lazy import is doable. One possible approach would be to introduce an `OpSpec` abstraction and resolve the actual implementation on first access (per-op), which could indeed reduce import time and avoid eager import failures.

That said, my main concern here is maintainability. Approaches that rely heavily on string-based representations (similar to m.def("op_name(...)") in PyTorch) tend to be:

* harder to refactor
* easier to break silently due to typos or mismatches
* less friendly to static analysis and tooling

So while lazy import is possible, it is not a free win, and the complexity cost needs to be justified by a clear import-time or dependency-related benefit.

2. Out-of-class registration

Out-of-class registration is also doable from a design perspective. However, I’m not entirely sure about the concrete user scenario we are optimizing for here.

From a platform-integration point of view, a centralized registration model is often easier to reason about and maintain. It avoids conditional registrations scattered across multiple files (e.g. sglang/srt/layer/...) and gives platform owners a single, explicit place to see what is supported.

That said, if we expect strong third-party extensibility or downstream backends registering ops independently, then out-of-class registration could make sense. I’m open to revisiting this if there’s a concrete use case that benefits from it.

I have pushed a new commit to this PR (trying to address your above questions with AI assisted for duplicated parts and comments). Please take a further look. Thanks.

mingfeima (approved):
Generally this one LGTM. we will run a couple of more test cases offline to double check if we have issues with CPU or XPU platforms.

ZailiWang:
Hi, heterogeneous architectures is a trend to further improve system throughput, e.g. offloading some ops to CPU. Shall we expand the `detect_platform()` functionality to support multiple platform detection? The return value can be a tuple of `Platform` objects.

python/sglang/srt/platforms/interface.py
```diff
@@ -0,0 +1,554 @@
...
+        except (ImportError, AttributeError):
+            # Import failed - return None (will fall back to native impl)
+            return None
+    else:
+        # Already a callable
+        _resolved_ops_cache[key] = impl
+        return impl
+
+
+def list_registered_ops(platform: "PlatformEnum" | None = None) -> list[str]:
```
jianan-gu, line 216: will be errors with `"PlatformEnum"` as its str type not class, so as others in this file.

yeahdongcn:
> > Generally this one LGTM. we will run a couple of more test cases offline to double check if we have issues with CPU or XPU platforms.
> 
> Hi @yeahdongcn , we have conducted a couple of tests for XPU/CPU devices, both are running with errors:
>

Ah, yes. This is just example code (and hasn’t been tested). Its purpose is to demonstrate how the integration could work in `srt`. To complete this work properly, I think the first step is to ensure that the overall approach makes sense for all vendors and that we agree on a unified interface. Once that’s in place, each vendor can collaborate on the corresponding implementation.

PR #17825, opened 2026-01-27
[Fix] VAE tiling/slicing argument handling for diffusers backend

sfiisf:
## Motivation

When using `--backend diffusers`, the VAE tiling/slicing launch argument was not properly handled, even though VAE tiling/slicing was already supported.

## Modifications

- Properly handle the VAE slicing launch argument when using the diffusers backend
- Align backend initialization logic so that both VAE tiling and slicing configurations are correctly applied
- Update the related API usage to ensure compatibility with the current diffusers initialization flow

## Accuracy Tests

This PR does not modify model forward logic or numerical computation. No accuracy impact is expected.

## Benchmarking and Profiling

This change does not affect inference performance unless VAE slicing is explicitly enabled by the user. No performance regression is expected.

## Checklist

## Review Process


mickqian:
@DefTruth could you take a look?

python/sglang/multimodal_gen/runtime/pipelines/diffusers_pipeline.py
```diff
@@ -489,13 +489,19 @@ def _apply_vae_optimizations(self, pipe: Any, server_args: ServerArgs) -> None:
             if hasattr(pipe, "enable_vae_slicing"):
```
DefTruth: I prefer this one. We should always use the non-deprecated APIs in Diffusers. @mickqian What do you think?

```suggestion
            if hasattr(pipe, "vae") and hasattr(pipe.vae, "enable_slicing"):
                pipe.vae.enable_slicing()
                logger.info("Enabled VAE slicing for lower memory usage")
            elif hasattr(pipe, "enable_vae_slicing"):
                pipe.enable_vae_slicing()
                logger.info("Enabled VAE slicing for lower memory usage")
```

python/sglang/multimodal_gen/runtime/pipelines/diffusers_pipeline.py
```diff
@@ -489,13 +489,19 @@ def _apply_vae_optimizations(self, pipe: Any, server_args: ServerArgs) -> None:
...
+                pipe.vae.enable_slicing()
+                logger.info("Enabled VAE slicing for lower memory usage")
 
         # VAE tiling: decode latents tile-by-tile for large images
         # https://huggingface.co/docs/diffusers/optimization/memory#vae-tiling
         if getattr(config, "vae_tiling", False):
             if hasattr(pipe, "enable_vae_tiling"):
                 pipe.enable_vae_tiling()
                 logger.info("Enabled VAE tiling for large image support")
+            elif hasattr(pipe, "vae") and hasattr(pipe.vae, "enable_tiling"):
```
DefTruth: same as above~

DefTruth (changes requested):

python/sglang/multimodal_gen/runtime/pipelines/diffusers_pipeline.py
sfiisf: Updated 👍
Now preferring the non-deprecated `pipe.vae.*` APIs, keeping backward compatibility, and adding warnings to avoid silent no-ops when slicing/tiling is unavailable.

DefTruth (approved):
LGTM~

mickqian (approved):
Could you add a testcase for diffusers pipeline in another PR?

mickqian (approved):

mickqian:
please fix CI

sfiisf:
> please fix CI

Sorry for the delayed follow-up on this.

I took another look at the CI failures. From the logs, it seems the errors are happening during model loading rather than in the code paths modified in this PR.

Specifically:

- One failure reports that many model weights were not initialized from the checkpoint (`qkv_proj`, `gate_up_proj`, etc.), which suggests a mismatch between the model definition and the checkpoint used in CI.
- Another failure shows `VAEArchConfig` missing the `input_channels` attribute, which looks like a configuration/schema mismatch.
- Another failure shows `EncoderConfig` missing the `tokenizer_kwargs` attribute, which looks like a configuration/schema mismatch.

These appear to be related to the model/config versions used in the CI environment rather than the changes introduced in this PR.

Please let me know if you'd like me to try addressing this in this PR (e.g., by aligning the dependency or config versions), or if this should be handled separately in the CI setup.

mickqian:
excellent

PR #17824, opened 2026-01-27
[EPD][VLM] support video/audio input

ZhengWG:
## Motivation

The PR adds video/audio input support with EPD mode, tested on Qwen-series models, addressing the requirements outlined in https://github.com/sgl-project/sglang/issues/15118.

based on https://github.com/sgl-project/sglang/pull/15475

## Modifications

1. Added video/audio processing pipeline
2. Handle multimodal embedding aggregation with `MultiModalEmbeddingData`
3. Support modality-based assignment

## Accuracy Tests

Test with Qwen-series model:

| Model Family | Model Variants | Supported Modalities | Test Status |
|--------------|----------------|---------------------|-------------|
| **Qwen2.5-VL** | `Qwen/Qwen2.5-VL-7B-Instruct` | Image, Video | ✅ Tested |
| **Qwen3-VL** | `Qwen/Qwen3-VL-30B-A3B-Instruct`, `Qwen/Qwen3-VL-235B-A22B-Instruct-FP8` | Image, Video | ✅ Tested |
| **Qwen2-Audio** | `Qwen/Qwen2-Audio-7B-Instruct` | Audio | ✅ Tested |
| **Qwen3-Omni** | `Qwen/Qwen3-Omni-MoE-A22B-Instruct` | Image, Video, Audio | ✅ Tested |

Launch script samle:

```bash
# language
MODEL_PATH=Qwen3-VL-30B-A3B-Instruct  # Qwen2.5-VL-7B-Instruct
CUDA_VISIBLE_DEVICES=4 python3 -m sglang.launch_server --model-path ${MODEL_PATH} --disable-radix-cache \
        --host $HOST_IP --port 8003 --trust-remote-code --tp-size 1 \
        --enable-cache-report --log-level info  --encoder-urls 'http://127.0.0.1:8001' 'http://127.0.0.1:8002' \
        --mem-fraction-static 0.7--chunked-prefill-size 8192--attention-backend fa3 \
        --enable-multimodal --language-only
# encoder
CUDA_VISIBLE_DEVICES=0 python3 -m sglang.launch_server --model-path ${MODEL_PATH} \
        --host $HOST_IP --port 8001 --trust-remote-code --tp-size 1 \
        --enable-cache-report --log-level info  \
        --mem-fraction-static 0.7--chunked-prefill-size 8192--attention-backend fa3 \
        --mm-attention-backend fa3 --encoder-only
[...]

PR #17837, opened 2026-01-27
fix(disaggregation): Support string/dict image_data in native API for EPD mode

flexorRegev:
## Summary

The native `/generate` endpoint fails with images in EPD (Encoder-Prefill-Decode) disaggregated mode because `encode_receiver.py` assumes `image_data` items are `ImageData` objects with `.url` attribute, but the native API passes raw strings or dicts from JSON.

**Error:**
```
AttributeError: 'str' object has no attribute 'url'
```
or
```
AttributeError: 'dict' object has no attribute 'url'
```

## Root Cause

The OpenAI-compatible `/v1/chat/completions` endpoint works because it goes through `conversation.py` which explicitly creates `ImageData` objects. However, the native `/generate` endpoint passes the raw JSON data directly, which can be:
- `str`: Direct URL or base64 data URI
- `dict`: `{"url": "..."}` format

The problematic code in `encode_receiver.py`:
1. `send_encode_request()` (zmq_to_scheduler backend)
2. `recv_mm_data()` (zmq_to_tokenizer and mooncake backends)

Both methods assumed `.url` attribute access which fails for strings and dicts.

## Solution

Added helper functions `_extract_image_url()` and `_extract_image_urls()` that handle all supported image data formats:
- `str`: Direct URL or base64 data URI (return as-is)
- `dict`: Extract from `{"url": "..."}` or `{"image_url": "..."}` keys
- `ImageData`: Access `.url` attribute (existing behavior)

### DIFF
--- a/python/sglang/srt/disaggregation/encode_receiver.py
+++ b/python/sglang/srt/disaggregation/encode_receiver.py
@@ -5,7 +5,7 @@ import random
 import threading
 import uuid
 from enum import IntEnum
-from typing import TYPE_CHECKING, List, Optional
+from typing import TYPE_CHECKING, List, Optional, Union
 
 import aiohttp
 import torch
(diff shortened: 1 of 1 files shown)


ShangmingCai (commented):
CC: @ZhengWG will this affect the ongoing video mm input PR?

flexorRegev:
@ZhengWG Thanks for pointing that out! I've refactored to align with the `_extract_url_data` pattern from #17824 — consolidated the two helper functions into a single `_extract_url_data` that uses `isinstance(item, ImageData)` instead of `hasattr`.

One thing this PR adds on top of #17824's handling is support for `dict` inputs (e.g. `{"url": "..."}` or `{"image_url": "..."}`), which the native `/generate` API can pass from raw JSON. That case isn't covered in #17824 yet — happy to contribute it there as well if useful.

python/sglang/srt/disaggregation/encode_receiver.py
```diff
@@ -28,6 +29,41 @@
...
+            url = item.get("url") or item.get("image_url")
+            if not url:
+                raise ValueError(
+                    f"Dict image data must contain 'url' or 'image_url' key: {item}"
+                )
+            urls.append(url)
+        elif isinstance(item, ImageData):
+            urls.append(item.url)
+        else:
+            raise ValueError(f"Unsupported image data type: {type(item)}")
```
ShangmingCai, line 63: Need to catch this error, or change it to a warning. Or the server will be attacked by this kinds of reqs.

hnyls2002:
@flexorRegev Thanks for the fix and for aligning it with the _extract_url_data pattern. #17824 has since merged with that helper (to_raw_url in encode_receiver.py now tolerates str, dict, and ImageData inputs), and all encode dispatch paths route through it, so the AttributeError this PR fixes no longer reproduces on main. Closing as superseded; please reopen if any EPD backend path still fails with raw string or dict image_data.

PR #17851, opened 2026-01-28
[diffusion] feat: add model offloading support for diffusers backend

sfiisf:
## Motivation

Diffusers natively supports model CPU offloading via `accelerate`, which can significantly reduce GPU memory usage with minimal impact on performance. However, when using `backend=diffusers`, sglang currently does not expose any way to configure this capability.

The goal of this PR is to expose diffusers model CPU offloading through sglang’s launch arguments, allowing users to reduce memory usage when running diffusers pipelines.

## Modifications

- Add a `--model-offload` launch argument for the diffusers backend
- When enabled, apply diffusers model CPU offloading during pipeline initialization
- Keep the default behavior unchanged when the option is not specified

## Checklist

## Review Process

### DIFF
--- a/python/sglang/multimodal_gen/configs/pipeline_configs/base.py
+++ b/python/sglang/multimodal_gen/configs/pipeline_configs/base.py
@@ -213,6 +213,9 @@ class PipelineConfig:
     # Wan2.2 TI2V parameters
     boundary_ratio: float | None = None
 
+    # Offload parameters
+    model_offload: bool = False
+
     # Compilation
     # enable_torch_compile: bool = False
 
@@ -480,6 +483,15 @@ class PipelineConfig:
             type=parse_int_list,
             default=PipelineConfig.dmd_denoising_steps,
             help="Comma-separated list of denoising steps (e.g., '1000,757,522')",
+        )
+
+        # Offload parameters
+        parser.add_argument(
+            f"--{prefix_with_dot}model-offload",
+            action=StoreBoolean,
+            dest=f"{prefix_with_dot.replace('-', '_')}model_offload",
+            default=PipelineConfig.model_offload,
+            help="Enable model offloading to CPU",
         )
 
         # Add VAE configuration arguments
--- a/python/sglang/multimodal_gen/configs/pipeline_configs/base.py
+++ b/python/sglang/multimodal_gen/configs/pipeline_configs/base.py
@@ -213,9 +213,6 @@ class PipelineConfig:
     # Wan2.2 TI2V parameters
     boundary_ratio: float | None = None
 
-    # Offload parameters
-    model_offload: bool = False
-
     # Compilation
     # enable_torch_compile: bool = False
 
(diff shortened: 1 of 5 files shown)


python/sglang/multimodal_gen/configs/pipeline_configs/base.py
```diff
@@ -213,6 +213,9 @@ def postprocess_image(self, image):
     # Wan2.2 TI2V parameters
     boundary_ratio: float | None = None
 
+    # Offload parameters
+    model_offload: bool = False
```
mickqian: this should be a server arg for user to override

sfiisf: Thanks for the feedback! I've moved `model_offload` to ServerArgs.

DefTruth: @mickqian @sfiisf I think we should reuse the flags from server args, such as `--dit-cpu-offload`, `--text-encoder-cpu-offload`, the new flag name `model_offload` will introduce the complexity of code maintenance and the meaning of 'model_offload' is ambiguous.

sfiisf: > @mickqian @sfiisf I think we should reuse the flags from server args, such as `--dit-cpu-offload`, `--text-encoder-cpu-offload`, the new flag name `model_offload` will introduce the complexity of code maintenance and the meaning of 'model_offload' is ambiguous.

Thanks for the suggestion!

The reason I introduced `--model-offload` is that the offload mechanism used in diffusers moves the entire model to CPU when inactive, which is slightly different from the existing flags like `--dit-cpu-offload` or `--text-encoder-cpu-offload`, where the offloading granularity is per-component.

I used `--model-offload` to reflect this difference in behavior. That said, I agree that introducing a new flag may increase maintenance complexity and the naming could indeed be ambiguous.

If you prefer, I can revise the implementation to reuse the existing flags (e.g., aligning with `--dit-cpu-offload` / `--text-encoder-cpu-offload`) and adapt the logic accordingly. Let me know which approach you think would fit better with the current design.

DefTruth: @sfiisf Thanks for the clear explanation. The CPU offload of diffusers does treat text encoder and dit as a whole. I prefer to explicitly state that this is a diffusers backend-related setting, e.g. `--diffusers-cpu-offload` (we've already done this for attn backend, `--diffusers-attention-backend`). @mickqian, what do you think?

sfiisf: While rebasing and looking at the latest `main`, I noticed that a similar change was recently merged which enables the diffusers offload behavior using `any(...)` over the existing flags by the PR https://github.com/sgl-project/sglang/pull/20173

Since that implementation is already in place, my original motivation for introducing `--model-offload` / `--diffusers-cpu-offload` might already be partially covered.

However, my initial intent was to reflect that the diffusers offload mechanism moves the whole model (e.g., text encoder and DiT together), while the existing flags control offloading at a more fine-grained component level. That's why I originally proposed a dedicated flag.

Given the current implementation, do you think it would be better to:

1. keep the current `any(existing_flags)` approach, or
2. introduce an explicit `--diffusers-cpu-offload` flag for clarity of behavior?

Happy to adjust this PR or close it if the current solution is preferred. @DefTruth @mickqian

PR #17862, opened 2026-01-28
Refactor: observability code cleanup

### LINKED ISSUE #17482: [Refactor ideas] Per request time stats tracing
Currently, the code for per request time stats tracing is scattered around the codebase. Everyone adds its own timestamps collection code. The collection code is mixed with core logic in the critical path. This hurts the readability and maintainability.

## Refactor goals
1. A single base class for per request time stats collection. All collection and analysis apis come from that class or file.
2. Remove existing scattered code.
    - Remove [APIServingTimingMixin](https://github.com/sgl-project/sglang/blob/236772c0e1e21a4bd7b0cb38f067690c1be66e72/python/sglang/srt/managers/io_struct.py#L122), RequestTimingMetricsMixin, TimeStats. Unify them into a new single class.
    - Remove ad-hoc collection code in the critical path. such as the code below. https://github.com/sgl-project/sglang/blob/236772c0e1e21a4bd7b0cb38f067690c1be66e72/python/sglang/srt/managers/tokenizer_manager.py#L1189-L1194, https://github.com/sgl-project/sglang/blob/236772c0e1e21a4bd7b0cb38f067690c1be66e72/python/sglang/srt/entrypoints/openai/serving_base.py#L92-L93
3. We only allow the the call of `time.time()` and `time.perf_counter()` in a dedicated folder. Files such as `scheduler.py`, `tokenizer_manager.py` is no allowed to call these functions.
4. For the new class `ReqTimeStatsBase`, it should only contain a list of monotonically increasing timestamps. In the doc string, it documents the lifetime of a request based on these timestamps. People compute their wanted duration by subtracting two timestamps.
5. For functions like `_calculate_timing_metrics`, also move to  `ReqTimeStatsBase` as utility or member functions.
6. We provide two tiers. By default, we collect all timestamps. If users turn on tracing, we call a more advanced industry-standard library to turn on more tracing features.
7. Add structured logging and allow people to analyze the log latter. We prefer logging + async analysis over ad-hoc analysis on the critical path.
8. The tracing logic should be a centralized, separated subpackage. It should not be mixed with core scheduling logic or data flow.
9. We do not need to worry about any backward compatibility in terms of the API, but we still want to cover all features (i.e. with a different API or methods to analyze)

sufeng-buaa:
## Motivation

According to https://github.com/sgl-project/sglang/issues/17482, organize the observability-related code, remove redundant code, and unify the interfaces for time statistics, request latency metrics, and request tracing.

## Modifications

- Move observability-related code to python/sglang/srt/observability
- Remove redundant code, refactor inappropriate code, and correct non-standard naming.
- Timestamps Record
   - In python/sglang/srt/observability/req_time_stats.py, APIServerReqTimeStats, DPControllerReqTimeStats, and SchedulerReqTimeStats are defined to record timestamp information for the tokenizer/gRPC server, dp controller, and scheduler, respectively. A series of set_*_time methods are provided to set timestamps, along with get_* methods to calculate latency.
   - Uniformly use MONOTONIC TIME, and update the time difference between MONOTONIC TIME and REALTIME upon each incoming request, for converting to REALTIME when necessary.
- Request latency metrics
   - Define the base class ReqTimeStatsBase for APIServerReqTimeStats, DPControllerReqTimeStats, and SchedulerReqTimeStats, integrating a metrics collector inside it. Export latency information to the metrics collector within each set_*_time method.
- Request Tracing
   - Define the base class ReqTimeStatsBase for APIServerReqTimeStats, DPControllerReqTimeStats, and SchedulerReqTimeStats, integrating a trace context inside it. Export trace spans within each set_*_time method. Define __getstate__ and __setstate__ of ReqTimeStatsBase to propagate the trace context.
   - Refactor the tracing package and optimize the span structure.
   - Support trace levels, and dynamically adjust trace levels via HTTP API.
   - Support tracing for requests with parallel_sample_num > 1.
   - Support tracing for request retract

## Checklist

## Review Process


ShangmingCai (commented):
The disaggregation part LGTM.

python/sglang/srt/observability/req_time_stats.py
```diff
@@ -0,0 +1,904 @@
...
+SGLANG_TEST_REQUEST_TIME_STATS = get_bool_env_var("SGLANG_TEST_REQUEST_TIME_STATS")
+
+
+logger = logging.getLogger(__name__)
+
+# Reduce system time calls by computing time.time() based on calibrated perf_counter() values.
+global_diff_realtime_monotonic = time.time() - time.perf_counter()
+
+
+def calibrate_time_diff():
```
zhanghaotong, line 51: Using global_diff_realtime_monotonic is a good idea for converting perf_counter values to timestamps. However, when DP > 1, you may have a single tokenization manager but multiple schedulers running across different devices. In that case, you need to be careful: each process/device can have its own monotonic clock offset, so the conversion may be inconsistent across ranks unless those offsets are synchronized or computed per rank.

sufeng-buaa: Monotonic time is typically obtained via the Linux kernel's vDSO (or system call), and the Linux kernel ensures timestamp consistency across different CPUs. Even in extremely rare cases, non-observability functionalities only rely on time retrieved within the same process, while observability features can fully tolerate such minimal timing discrepancies.

zhanghaotong: But if the tokenize manager and the scheduler are running on completely different machines, will the monotonic time they obtain still be consistent?

sufeng-buaa: Fixed. Correct monotonic time during deserialization by propagating it with a diff.

ishandhanani (approved):

python/sglang/srt/managers/io_struct.py
```diff
@@ -1131,6 +1077,9 @@ class BatchStrOutput(
     # Detailed breakdown of cached tokens by source (device/host/storage)
     cached_tokens_details: Optional[List[Optional[Dict[str, Any]]]] = None
 
+    # for observability
```
sherlockwu: nit: For observability

same as other places.

python/sglang/srt/observability/trace.py
```diff
@@ -0,0 +1,698 @@
...
+        if not self.tracing_enable:
+            return
+
+        ts = ts or get_cur_time_ns()
+
+        # End all unclosed thread spans.
+        self.abort()
+
+        if attrs:
+            self.root_span.set_attributes(attrs)
```
sherlockwu, line 430: Is it possible for `root_span` to be None (e.g. after deserialization)?

It looks impossible under current logic -- but probably safer to add an explicit guard / null check

sherlockwu (approved):
Thanks for the cleanup of the metrics and tracing code! LGTM. A few small nits below.

python/sglang/srt/observability/trace.py
sufeng-buaa: ok, I have added it.

sherlockwu: nice!

merrymercy (approved):

PR #17870, opened 2026-01-28
Fix Description for MTP Model update_weights_from_tensor API OOM Issue

ArronHZG:
## Problem Background

Multiple calls to the MTP model's `update_weights_from_tensor` API for parameter updates will result in **OOM due to memory leakage**.

### Root Cause Analysis:

1. SGLang and the RL engine communicate via **inter-process communication (IPC)**, with parameters transmitted through **CUDA IPC**.

2. In `EAGLEWorker/EAGLEWorker2`, `update_weights_from_tensor` sequentially invokes:
        

    - `draft_worker.draft_runner.update_weights_from_tensor()`

    - `target_worker.model_runner.update_weights_from_tensor()`

3. **Each worker deserializes the**  `named_tensors`  **independently**, causing the CUDA IPC handle in `LocalSerializedTensor` to be opened twice (via `cudaIpcOpenMemHandle`) but closed only once (via `cudaIpcCloseMemHandle`), leading to memory leakage.

## Solution

Add a new `unwrap_ipc_tensors` function to perform IPC allocation and deallocation **only once per update process**.

### Function Implementation

```python

def unwrap_ipc_tensors(
    named_tensors: List[Tuple[str, torch.Tensor]], tp_rank: int, device: torch.device
) -> List[Tuple[str, torch.Tensor]]:
    """Unwrap LocalSerializedTensor objects to detach from IPC memory.

    This function should be called only once after deserialization to avoid
    opening multiple CUDA IPC handles for the same tensor.
    """

    result = []
    for name, tensor in named_tensors:
        if isinstance(tensor, LocalSerializedTensor):
            ipc_tensor = tensor.get(tp_rank)
            result.append((name, ipc_tensor.to(device)))
            del ipc_tensor
        else:
            result.append((name, tensor.to(device)))
    return result
```

## Modified Files

1. **`python/sglang/srt/model_executor/model_runner.py`**: Add the `unwrap_ipc_tensors` function.

2. **`eagle_worker.py/eagle_worker_v2.py`**: In `update_weights_from_tensor`, **prioritize** the deallocation logic for deserialized IPC tensors, and unwrap the tensors before passing them to the two workers.

This ensures that each IPC handle of `LocalSerializedTensor` is opened and closed exactly once, preventing OOM issues caused by memory leakage.
## Checklist

## Review Process


yushengsu-thu (commented):
This modification looks reasonable. @ArronHZG, could you please share your failing testing cases and add them as a test case in this PR? Thanks

AniZpZ (approved):

ArronHZG:
@yushengsu-thu Marking ~  could you please review my code again?

san-tian:
We hit a production failure on the same EAGLE v2 CUDA-IPC update path and have some additional TP=8 evidence that may help prioritize this fix.

Environment:
- 8x B200, TP=8
- SGLang downstream commit `9b292aad8b9583cab2a741c621fb750d8dfc939a`
- EAGLE v2 enabled with an MTP draft model
- server had been running for several days

Failure timeline:
- At `20:35:11`, the detokenizer heartbeat stopped immediately after a tensor-weight update entered the scheduler.
- TP ranks 1-7 reached `SchedulerWeightUpdaterManager.update_weights_from_tensor()`'s final Gloo barrier and remained there.
- Two hours later, those ranks failed with:

```text
RuntimeError: .../gloo/transport/tcp/unbound_buffer.cc:78
Timed out waiting 7200000ms for recv operation to complete
```

- A py-spy dump showed TP0 had not reached the barrier. It was still in the EAGLE draft update's IPC unwrap path:

```text
LocalSerializedTensor.get / _unwrap_tensor
  ModelRunner.update_weights_from_tensor (model_runner.py:2168)
  EAGLEWorkerV2.update_weights_from_tensor (eagle_worker_v2.py:1684)
  SchedulerWeightUpdaterManager.update_weights_from_tensor (weight_updater.py:158)
```

- The other seven ranks were already at `torch.distributed.barrier(group=self.tp_cpu_group)`.
- We found no host reboot, kernel OOM, or NVIDIA Xid around the incident.

The deployed code deserializes `named_tensors` once in `EAGLEWorkerV2`, then passes the same `LocalSerializedTensor` objects to both draft and target `ModelRunner.update_weights_from_tensor()` calls, each of which invokes `_unwrap_tensor()` / `LocalSerializedTensor.get()` independently. That matches this PR's double-open analysis.

This does not by itself prove that the leaked IPC handles were the only possible cause, but it does show the failure can become a hard rank-local hang: one rank remains in IPC unwrap while peers are hidden behind the post-update Gloo barrier for the full two-hour process-group timeout.

I can help validate a current-main rebase of this PR on TP=8 B200 with repeated EAGLE tensor updates if useful.

AliceChenyy:
Independent reproduction with **native stack traces** pinpointing the exact hang site, plus a rebased expression of this fix validated on v0.5.17. Posting in the hope it helps this land — we hit the same double-open and can confirm both the mechanism and the fix.

## Context

verl RL weight sync (`update_weights_from_tensor`, `LocalSerializedTensor` path) on **Qwen3.5-35B-A3B with its native MTP head** (EAGLE, steps=3, topk=1), sglang **v0.5.17**, TP=8, single node (reproduced identically on 8×H200 NVL and 8×H20). With speculative rollout enabled, the *first* weight sync hangs forever: static log, 100% GPU (NCCL spin), 7 of 8 scheduler ranks parked in the `tp_cpu_group` barrier.

## What the native stack shows

Python-level py-spy is actively misleading here — it shows the diverged rank at the `eagle_worker_v2.py` call line with **no callee frame**, which reads like a stranded collective. `py-spy dump --native` on the diverged rank shows it is actually blocked inside a **CUDA-IPC shared-storage destructor**:

```
THPVariable_dealloc
c10::TensorImpl::~TensorImpl
THPStorage_dealloc
c10::StorageImpl::release_resources
THPStorage_newSharedCuda(...)::{lambda(void*)#1}::_FUN   <- CUDA-IPC deleter
cudaStreamSynchronize
cuStreamSynchronize                                       <- never returns
update_weights_from_tensor (eagle_worker_v2.py, target-load line)
```

A gdb backtrace of the barrier ranks confirms they are in plain gloo `pthread_cond_wait` — no stranded NCCL collective anywhere (only the watchdog thread).

Two details that support the double-open analysis in this PR:

- **Which rank diverges is nondeterministic** — TP1 on one run, TP3 on the next, same config. That is consistent with the draft load's IPC-opened copies being kept alive by reference cycles and freed by the *cyclic collector* at an arbitrary later point. When that point falls inside the target load, `cudaIpcCloseMemHandle`'s preceding stream sync runs against a handle the target load still has mapped and never returns.
- So beyond the leak this PR describes (opened twice, closed once → OOM over many updates), the same double-open can also manifest as a **hard hang on a single update** — matching the production incident @san-tian reported above. Same root cause, two presentations.

## Rebase note + validated fix expression

This PR predates #31148, which moved weight updating into `WeightUpdater` components; `unwrap_ipc_tensors` no longer exists and `ModelRunner.update_weights_from_tensor` is now `model_runner.weight_updater.update_weights_from_tensor`. The same fix expressed against the current layout — materialise once before either load, pass the resulting tensors to both (`_unwrap_tensor` is a no-op on already-materialised tensors, so both loads work unmodified):

```python
# eagle_worker_v2.EAGLEWorkerV2.update_weights_from_tensor, after deserialize:
if recv_req.load_format != "flattened_bucket":
    from sglang.srt.model_executor.model_runner_components.weight_updater import _unwrap_tensor
    device = torch.get_device_module(self.target_worker.model_runner.device).current_device()
    named_tensors = [
        (name, _unwrap_tensor(tensor, tp_rank=self.ps.tp_rank, device=device))
        for name, tensor in named_tensors
    ]
```

(Both runners share the worker's `ParallelState`, so `self.ps.tp_rank` is the value each `weight_updater` would have used itself. `flattened_bucket` is skipped — that format passes `named_tensors` through as a dict and unwraps downstream.)

Idempotent patch script + how we bake it into the image: https://github.com/AliceChenyy/verl/blob/feat/qwen35-mtp-spec-rollout/recipes/qwen35_mtp_spec/docker/patch_sglang_eagle_ipc.py

## Validation (v0.5.17, config that hung deterministically twice before)

| | before | after |
|---|---|---|
| server ready → first `Decode batch` | never (>17 min, killed) | 58 s |
| weight-sync buckets completed | 0 | 197 |
| spec accept rate | — | 0.68–0.79 |
| 40-step RL soak (fresh weight sync every step) | — | accept rate stable 0.67–0.75, no drift, no leak-driven failure |

One cautionary data point for whoever writes the test: an earlier workaround we tried (routing the update to `tp_worker` only via `disable_draft_model`) removed the hang and left *every* other signal healthy — decode proceeded, training stepped, loss curves normal — while silently disabling speculation (accept rate 0.00, draft weights never delivered). A regression test for this path should assert on **accept rate / accept length after a weight update**, not just on the update returning success.

Happy to share full native dumps / logs if useful.

PR #17876, opened 2026-01-28
support nextn>1 for deep_gemm.fp8_paged_mqa_logits and flash_mla_with…

yzk1989:
support nextn>1 when use deep_gemm.fp8_paged_mqa_logits and flash_mla_with_kvcache


hnyls2002:
@yzk1989 Thanks for the contribution. Support for nextn > 1 with deep_gemm.fp8_paged_mqa_logits and flash_mla_with_kvcache has since landed on main via #24870 (relanded in #27166), and the NSA code this PR patches has been renamed to the dsa/ module, so the change no longer applies to current main. Closing as superseded; please reopen if you hit a remaining case on current main that the merged implementation does not cover.

PR #17892, opened 2026-01-28
[RL] support deepep when enabling return routed experts

ocss884:
## Checklist

## Review Process


yueming-yuan:
information for review: I tested R3 with this PR in Miles, and it works

hnyls2002:
@ocss884 

Please at least write something, even for the simplest summary...

python/sglang/srt/layers/moe/routed_experts_capturer.py
```diff
@@ -187,7 +191,29 @@ def _sync_fwd_experts_buffer_DtoH(
         can_run_graph: bool,
         cuda_graph_batch: int,
     ):
-        if is_dp_attention_enabled():
+        # we need to have the topk at each rank being TP_ATTN_FULL
+        # With Deepep enabled, mlp mode is SCATTERED
+        if get_moe_a2a_backend().is_deepep():
```
hnyls2002, line 196: Not only deepep, but all non-NONE a2a backends are under `SCATTERED` mode (see `_compute_mlp_mode`).

If other backends are not supported for R3, please raise an exception here instead of silently falling through to the DP-attn-only path.

hnyls2002 (changes requested):

python/sglang/srt/layers/moe/routed_experts_capturer.py
hnyls2002: @ocss884

ocss884: Yes, a better way could be directly judging the `LayerScatterModes` of mlp instead of an if-else on whether deepep is on. I will refine this part later for review

PR #17893, opened 2026-01-28
[NVIDIA] Update Triton to 3.6.0

Kh4L:
Update needed for Thor
Full context:
https://github.com/triton-lang/triton/commit/6213a0e8a7c3d4f2e2983b5afdf89abd2004d585


hnyls2002:
Hi @Kh4L, I'm fairly sure this one no longer applies. The Dockerfile has been substantially rewritten since (FLASHINFER_VERSION is now 0.6.12 and the "Fix Triton to use system ptxas for Blackwell" block your diff anchors on has been removed), so the patch no longer applies cleanly. Closing as obsolete; please reopen with a rebased version if Triton 3.6.0 is still needed.

PR #17898, opened 2026-01-28
[Feature] Support Marconi

qimcis:
## Motivation

following on #11214, add Marconi prefix caching for hybrid llms

## Modifications

ref: [doc](https://docs.google.com/document/d/1zjX47VZ65wJPUgmGfI7nG3PuOCBXolaDtcUAVtq5308/edit?tab=t.0)

## Accuracy Tests

```bash
python3 -m sglang.launch_server --model Qwen/Qwen3-Next-80B-A3B-Instruct --tp-size 4 --chunked-prefill-size 64 --max-running-requests 16 --enable-marconi

python3 benchmark/gsm8k/bench_sglang.py --num-questions 1000
```

```text
Accuracy: 0.946
Invalid: 0.000
Latency: 308.085 s
Output throughput: 536.209 token/s
```

## Benchmarking and Profiling

benchmark environment:

`Qwen/Qwen3-Next-80B-A3B-Instruct`, 4xh100
Baseline: `main` at `af2807e14`

```bash
  --model-path Qwen/Qwen3-Next-80B-A3B-Instruct \
  --tp-size 4 \
  --context-length 32768 \
  --schedule-policy lpm \
  --mamba-scheduler-strategy extra_buffer \
  --mamba-ssm-dtype bfloat16 \
  --mamba-full-memory-ratio 0.9 \
  --page-size 1 \
  --enable-metrics
```

marconi run: 

```bash
  --model-path Qwen/Qwen3-Next-80B-A3B-Instruct \
  --tp-size 4 \
  --context-length 32768 \
  --schedule-policy lpm \
  --mamba-scheduler-strategy extra_buffer \
  --mamba-ssm-dtype bfloat16 \
  --mamba-full-memory-ratio 0.9 \
  --page-size 1 \
  --enable-metrics \
  --enable-marconi
```

benchmark:

```bash
  --backend sglang \
  --dataset-name generated-shared-prefix \
  --num-prompts 3072 \
  --gsp-num-groups 192 \
  --gsp-prompts-per-group 16 \
  --gsp-system-prompt-len 8192 \
  --gsp-question-len 512 \
  --gsp-output-len 128 \
  --gsp-num-turns 1 \
  --request-rate inf \
  --seed 7
```

result:

| Variant | Requests | Mean TTFT (ms) | Median TTFT (ms) | P99 TTFT (ms) | Mean E2E Latency (ms) | Request Throughput (req/s) |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `main` | 3072 | 228,919.32 | 219,894.56 | 394,390.84 | 239,085.36 | 7.56 |
| `marconi` | 3072 | 182,469.33 | 166,363.30 | 298,156.68 | 190,285.89 | 10.08 |

deltas:
- Mean TTFT: `20.29%` lower
- Median TTFT: `24.34%` lower
- P99 TTFT: `24.40%` lower
- Mean end-to-end latency: `20.41%` lower
- Request throughput: `33.25%` higher

marconi runtime stats on this run:

- `evict_full_tokens = 9,065,653`
- `evict_mamba_states = 0`
- `token_hit_rate = 52.94%`
- `admission_branches = 3009`
- `track_entry_inserts = 5649`
- `branch_checkpoint_inserts = 0`
- `rehydrate_count = 0`

## Checklist


python/sglang/srt/server_args.py
```diff
@@ -3697,6 +3720,17 @@ def add_cli_args(parser: argparse.ArgumentParser):
             default=ServerArgs.radix_eviction_policy,
             help="The eviction policy of radix trees. 'lru' stands for Least Recently Used, 'lfu' stands for Least Frequently Used, and 'slru' stands for Segmented Least Recently Used.",
         )
+        parser.add_argument(
+            "--enable-marconi",
```
XinyiQiao, line 3724: Instead of introducing a separate flag, can we use the existing radix-eviction-policy, and support marconi as a valid option for hybrid radix cache?

python/sglang/srt/server_args.py
```diff
@@ -3060,6 +3075,14 @@ def _handle_cache_compatibility(self):
                 "and cannot be used at the same time. Please use only one of them."
             )
 
+        if self.enable_marconi:
+            if self.disable_radix_cache:
+                logger.warning("Marconi is disabled because radix cache is disabled.")
```
XinyiQiao, line 3080: Should we just log as warning or throw exception because this config is invalid? If we log as warning, would recommend changing the log line to be :disabling marconi since radix cache is disabled for more clarity

python/sglang/srt/mem_cache/mamba_radix_cache.py
```diff
@@ -705,15 +1406,188 @@ def pretty_print(self) -> None:
...
+        efficiency_scores = []
+        mamba_weight, attn_weight, mlp_weight = self.marconi_eviction_latency_weights
+        for node in nodes:
+            seqlen_total = node.prefix_len
+            seqlen_child = len(node.key)
+            seqlen_parent = max(seqlen_total - seqlen_child, 0)
+
+            mamba_layers = self._marconi_node_mamba_layers(node)
+            flops_savings_mamba = mamba_layers * get_mamba1_flops(
+                seqlen_child, stats.model_dim, stats.ssm_state_size
```
XinyiQiao, line 1495: i saw you're using seqlen_child as the marginal Mamba-layer saving for a node, but this assumes that the immediate parent node is not tombstoned. If the parent node is tombstoned, recomputing this node’s state requires replaying all segments from the nearest live Mamba ancestor, not just this node’s local segment. We've ran some benchmark on this, using seqlen_child as approximation is less problematic on simple uniform workloads, but it can mis-rank nodes on more heterogeneous workload where shared-prefix lengths vary a lot

python/sglang/srt/mem_cache/mamba_radix_cache.py
```diff
@@ -705,15 +1406,188 @@ def pretty_print(self) -> None:
...
+            )
+            flops_savings_attn = stats.num_attn_layers * (
+                get_attn_flops(seqlen_total, stats.model_dim)
+                - get_attn_flops(seqlen_parent, stats.model_dim)
+            )
+            flops_savings_mlp = stats.num_mlp_layers * (
+                get_mlp_flops(seqlen_total, stats.model_dim)
+                - get_mlp_flops(seqlen_parent, stats.model_dim)
+            )
+            total_flops_savings = (
```
XinyiQiao, line 1505: Marconi FLOPS Saving calculation is heavily dependent on hybrid model architecture, eg mamba2 vs GDN, MLP vs MoE do not have the same compute profile. Using one generic formula might not be accurate.

qimcis:
> Hi @qimcis thank you, I've left few comments on the PR, my main questions are around marconi FLOPs saving calculation
> 
>     1. I saw currently you are using seqlen_child to calculate marginal flops saving, but this assume that the immediate parent node is not tombstoned. I've ran some benchmark and saw that this is not working well when the workload contains many competing shared prefixes of various length
> 
>     2. The flops saving calculations depends on hybrid model architecture, how do we want to handle models with different hybrid blocks, such as GDN, or models with MoE layers
> 
> 
> Let me know your thoughts? Would love to discuss further or collaborate.

Per request from @ruipeterpan, I've split this PR into two; #21860 for admission (open for review) and #21862 for eviction (wip). I've addressed most of the changes/fixes both you and Rui pointed out to the best of my knowledge, (addressed in the new prs), with the exception of the two quoted above.

On the eviction branch I refactored the FLOPs model into per-model profiles (refer to code [here](https://github.com/sgl-project/sglang/pull/21862/changes#diff-a10231e62cd7764c092e4bd4a8130a1c0d590d5b748eb7917fe1c22816244ff9), recurrent side is family-specific (for example Mamba2 vs GDN) and the FFN side distinguishes dense MLP vs MoE). For your first comment, I'm investigating this now too, and I agree the current implementation does not behave well under workloads with many competing shared prefixes of different lengths. Please feel free to leave comments on #21862 @XinyiQiao we can continue the conversation there!

PR #17905, opened 2026-01-29
Fix added tokens config with sensible filter

sbeurnier:
## Motivation

Fix the fake added tokens in `added_tokens.json`

Many adapters copy `added_tokens.json` from the base model's tokenizer. In that case, this file contains tokens that are already part of the base vocabulary (token ID < `base_vocab_size`). When SGLang reads this file, it incorrectly counts these as "added tokens", leading to a validation error in lora_manager.py that causes Sglang to either crash during startup or fail the adapter loading gracefully during runtime.

## Modifications

Filter fake added tokens (`lora_config.py`, `lora_manager.py`)

- Added `filter_added_tokens(base_vocab_size)` method to `LoRAConfig` that removes tokens with ID < `base_vocab_size` from `added_tokens_config`
- Called in `LoRAManager.init_lora_shapes()` for startup adapters
- Called in `LoRAManager.load_lora_adapter()` and `load_lora_adapter_from_tensors()` for dynamically loaded adapters
- Updates `lora_added_tokens_size` after filtering

Examples of adapters that fail to load (most on huggingface)
https://huggingface.co/zhanzhongrui/dlcv-qwen-vl-finetune2000/tree/main

## Accuracy Tests

N/A - This is a bug fix that prevents crashes; it does not affect model outputs.

## Benchmarking and Profiling

N/A - No changes to inference speed.

## Checklist

- [x] Provide accuracy and speed benchmark results according to [Test the accuracy] (https://docs.sglang.io/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed] (https://docs.sglang.io/developer_guide/contribution_guide.html#benchmark-the-speed). Acc/speed is unrelated to this change


python/sglang/srt/lora/lora_config.py
```diff
@@ -42,6 +48,103 @@ def __init__(
...
+                f"No safetensors file found in {weights_dir}, "
+                "falling back to declared target_modules"
+            )
+            return None
+
+        modules = set()
+        with safe_open(safetensors_path, framework="pt") as f:
+            for key in f.keys():
+                module_name = get_target_module_name(key, set(self.target_modules))
+                modules.add(module_name)
```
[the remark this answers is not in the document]
sbeurnier: In normal PEFT workflows, target_modules in adapter_config.json and the weights in adapter_model.safetensors are generated together during fine-tuning. The safetensors file only contains weights for modules specified in target_modules, so weights should always be a subset of (or equal to) target_modules.
The only way to have weights for modules NOT in target_modules is manual tampering with the config after training or file mismatch (using config from adapter A with weights from adapter B). In those cases, crashing with a clear error is preferable to silently ignoring weights, which could lead to confusing behavior.

python/sglang/srt/lora/lora_config.py
```diff
@@ -42,6 +48,103 @@ def __init__(
...
+        logger.debug(f"Computed effective target modules from safetensors: {modules}")
+        return modules
+
+    def _get_weights_dir(self) -> Optional[str]:
+        """Get the directory containing adapter weights."""
+        if os.path.isdir(self.path):
+            return self.path
+        # For HF repo IDs, fall back to declared target_modules
+        # (safetensors would need to be downloaded separately)
+        return None
```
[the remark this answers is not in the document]
sbeurnier: good rec

sbeurnier:
Rebased onto latest main and squashed into a single commit. The key change from the previous version: `filter_added_tokens()` is now called **before** `validate_new_adapter()` in all three loading paths, so adapters with fake added tokens (copied from the base tokenizer) pass the upstream validation check that was added in #17680.

@Fridge003 could you take another look when you get a chance? Also needs CI approval to run.

python/sglang/srt/lora/lora_config.py
```diff
@@ -42,6 +45,35 @@ def __init__(
...
+
+        original_count = len(self.added_tokens_config)
+        self.added_tokens_config = {
+            token: token_id
+            for token, token_id in self.added_tokens_config.items()
+            if token_id >= base_vocab_size
+        }
+        self.lora_added_tokens_size = len(self.added_tokens_config)
+
+        filtered_count = original_count - self.lora_added_tokens_size
```
glenliu21: I think we can just inline the `original_count` variable

python/sglang/srt/lora/lora_manager.py
```diff
@@ -489,6 +491,11 @@ def init_lora_shapes(
                 default=0,
             )
 
+        # Filter fake added tokens from startup adapters (ID >= base vocab_size)
+        base_vocab_size = self.base_hf_config.vocab_size
+        for config in self.configs.values():
+            config.filter_added_tokens(base_vocab_size)
```
glenliu21: We can probably inline `base_vocab_size` here too

python/sglang/srt/lora/lora_config.py
```diff
@@ -42,6 +45,35 @@ def __init__(
...
+        if filtered_count > 0:
+            logger.debug(
```
Fridge003: We should throw an error here instead

test/registered/lora/test_lora_config.py
```diff
@@ -0,0 +1,138 @@
+# Copyright 2023-2024 SGLang Team
```
Fridge003: This test is unnecessary

python/sglang/srt/lora/lora_config.py
```diff
@@ -42,6 +45,35 @@ def __init__(
...
+        original_count = len(self.added_tokens_config)
+        self.added_tokens_config = {
+            token: token_id
+            for token, token_id in self.added_tokens_config.items()
+            if token_id >= base_vocab_size
+        }
+        self.lora_added_tokens_size = len(self.added_tokens_config)
+
+        filtered_count = original_count - self.lora_added_tokens_size
+        if filtered_count > 0:
```
Fridge003: We need to raise an error when self.lora_added_tokens_size is not 0 after filtering, since it hasn't been supported

python/sglang/srt/lora/lora_config.py
```diff
@@ -42,6 +45,35 @@ def __init__(
             len(self.added_tokens_config) if self.added_tokens_config is not None else 0
         )
 
+    def filter_added_tokens(self, base_vocab_size: int) -> None:
```
Fridge003: Actually we don't need a new function `filter_added_tokens`. It can be part of __init__ function

Fridge003 (approved):

Fridge003:
Tested with 
```
python3 -m sglang.launch_server --model-path Qwen/Qwen3-VL-8B-Instruct --enable-lora --lora-paths lora0=zhanzhongrui/dlcv-qwen-vl-finetune2000 --max-loras-per-batch 2
```

On main branch, it throws error:
```
RuntimeError: Failed to load LoRA adapter lora0: LoRA serving currently doesn't support adapters that add tokens to the vocabulary
```

With this PR:
```
...
[2026-04-01 04:41:39] INFO:     127.0.0.1:41830 - "POST /v1/chat/completions HTTP/1.1" 200 OK
[2026-04-01 04:41:39] The server is fired up and ready to roll!
```

PR #17948, opened 2026-01-29
Direct model loading from object storage with Runai Model Streamer

noa-neria:
# RunAI Model Streamer Integration

## Overview

This integration adds support for the `runai_streamer` load format, enabling direct model loading from object storage and significantly improving loading performance from high-speed local storage and shared file systems. Runai model streamer is already used for multimodal in SGLang as the default loader.

### 1. Object Storage Support (S3 & GCS)

- **Direct Loading**: Load models directly from S3 (`s3://`) and Google Cloud Storage (`gs://`) without downloading to local disk
- **Minimal Local Storage**: Only metadata files (~100MB) are cached locally; weights are streamed on-demand
- **Automatic Detection**: Object storage URIs are automatically detected and use `runai_streamer` format

**Example:**
```bash
python -m sglang.launch_server --model-path s3://my-bucket/models/llama-3-70b/
```

### 2. Accelerated Local Loading

- **2-3x Faster**: Concurrent streaming significantly speeds up weight loading from SSDs and NVMe drives
- **Shared File Systems**: Improved performance on NFS, Lustre, and GPFS through parallel reads
- **Tunable Concurrency**: Adjust parallelism based on storage characteristics

**Example:**
```bash
  --model-path /mnt/nvme/models/llama-3-70b/ \
  --load-format runai_streamer \
  --model-loader-extra-config '{"concurrency": 16}'
```

### 3. Distributed Loading

- **Model Parallelism**: Streaming significantly speeds up weight loading from object storage by dividing the reading between the processes and broadcasting between devices
 - **Multi-Node **: All nodes stream independently from shared cloud storage, or can be configured to share the reading across nodes

### 4. Multiple Files Loading

- **Multiple Files**: Concurrent reading of multiple safetensors files

### Design Pattern: Single Metadata Download From Object Storage Before Fork

1. **Main Process**: Downloads config and tokenizer files once to local cache (`~/.cache/sglang/model_streamer/`)
2. **Worker Processes**: Read cached metadata and stream weights concurrently
3. **No Coordination**: Avoids file locks, race conditions, and duplicate downloads

### Basic Usage

```bash
# Object storage (auto-detected)
python -m sglang.launch_server --model-path s3://bucket/model/

# Local storage with explicit format
  --model-path /path/to/model/ \
  --load-format runai_streamer
```

### Key Components
1. **ObjectStorageModel** - Manages metadata download and caching
2. **RunaiModelStreamerLoader** - Model loader implementation
3. **runai_safetensors_weights_iterator** - Concurrent weight streaming
4. **set_runai_streamer_env** - Environment variable configuration

## Documentation

Comprehensive documentation available at:
- `docs/advanced_features/object_storage.md` - Full usage guide
- `docs/references/environment_variables.md` - Environment variable reference

## Credits

Based on [Run:ai Model Streamer](https://github.com/run-ai/runai-model-streamer) library.
Adapted from vLLM's implementation with enhancements for SGLang's architecture.

## References

- [Run:ai Model Streamer Documentation](https://github.com/run-ai/runai-model-streamer)


docker/Dockerfile
```diff
@@ -398,6 +398,9 @@ RUN CMAKE_VERSION=3.31.1 \
     && cp -r "${CMAKE_INSTALLER}/share/"* /usr/local/share/ \
     && rm -rf "${CMAKE_INSTALLER}" "${CMAKE_INSTALLER}.tar.gz"
 
+ARG RUNAI_MODEL_STREAMER_VERSION=">=0.15.4"
```
b8zhong: Q: why not add it to `pyproject.toml`?

python/sglang/srt/entrypoints/engine.py
```diff
@@ -151,7 +151,6 @@ def __init__(self, **kwargs):
                 kwargs["log_level"] = "error"
             server_args = self.server_args_class(**kwargs)
         self.server_args = server_args
-        logger.info(f"{server_args=}")
```
b8zhong, line 181: QQ: why delete this line?

python/sglang/srt/model_loader/loader.py
```diff
@@ -2695,6 +2695,288 @@ def _standard_quantization_workflow(
...
+        self._is_distributed = None
+        if load_config.model_loader_extra_config:
+            extra_config = load_config.model_loader_extra_config
+
+            if "distributed" in extra_config and isinstance(
+                extra_config.get("distributed"), bool
+            ):
+                self._is_distributed = extra_config.get("distributed")
+
+    def _maybe_download_from_modelscope(
```
b8zhong: Do we need this function? Since it's only for object storage format, I don't think so right?

python/sglang/srt/model_loader/loader.py
```diff
@@ -2695,6 +2695,288 @@ def _standard_quantization_workflow(
...
+    def _prepare_weights(
+        self, model_name_or_path: str, revision: Optional[str]
+    ) -> Tuple[str, List[str]]:
+        """Prepare weights for the model.
+
+        If the model is not local, it will be downloaded."""
+        from sglang.srt.utils.runai_utils import is_runai_obj_uri, list_safetensors
+
+        model_name_or_path = (
+            self._maybe_download_from_modelscope(model_name_or_path, revision)
```
b8zhong: Same thing here

python/sglang/srt/model_loader/loader.py
```diff
@@ -2695,6 +2695,288 @@ def _standard_quantization_workflow(
...
+        if self._is_distributed is None:
+            self._is_distributed = is_object_storage_path
+        is_local = os.path.isdir(model_name_or_path)
+        safetensors_pattern = "*.safetensors"
+        index_file = SAFE_WEIGHTS_INDEX_NAME
+
+        hf_folder = (
+            model_name_or_path
+            if (is_local or is_object_storage_path)
+            else download_weights_from_hf(
```
b8zhong, line 2955: We could also delete these codes, if the above is true right? Since, it's not pointing to a HF repo (correct me if my understanding not correct)

b8zhong: Thus I think we could just download the entire contents of the remote store.

docker/Dockerfile
noa-neria: Thanks! Added as an optional dependency `sglang[runai] `

python/sglang/srt/model_loader/loader.py
noa-neria: We thought to preserve the fallback to downloading from HF, as this is the standard. 
Preserving the fallback ensures that the integration adds a new capability without breaking the existing core behavior of the engine.

test/registered/runai_streamer/test_runai_model_loader.py
```diff
@@ -0,0 +1,46 @@
+import unittest
+
+import sglang as sgl
+from sglang.srt.environ import temp_set_env
+from sglang.test.ci.ci_register import register_cuda_ci
+
+register_cuda_ci(est_time=600, suite="stage-b-test-small-1-gpu")
```
b8zhong: Could you make sure the time of this test is relatively accurate?

python/sglang/srt/server_args.py
```diff
@@ -702,6 +704,8 @@ def __post_init__(self):
         Orchestrates the handling of various server arguments, ensuring proper configuration and validation.
         """
 
+        self._maybe_download_model_for_runai()
```
b8zhong, line 750: Nit: could you help move it to handle load format area below.

python/sglang/srt/model_loader/loader.py
b8zhong: I see. But if it inherits from BaseModelLoader do we need to implement this function again? 

> We thought to preserve the fallback to downloading from HF, as this is the standard.
Because, it would be annoying to debug if, there was something wrong, then we should just let it error out. (thus it's a rare case and not sure if it should be considered).

python/sglang/srt/model_loader/loader.py
b8zhong: @noa-neria What I mean is that, how could it download from modelscope if it's in a S3 bucket?

noa-neria: @b8zhong  Runai model streamer optimizes download from shared filesystems, SSDs and object storage. Obviously, the streamer can also read from local files.
If the model path is not found, we fallback to downloading from HF to local disk, exactly as the default loader. 

Supporting Modelscope means that if `SGLANG_USE_MODELSCOPE` is set, the fallback is downloading from Modelscope instead of HF.

However, I removed Modelscope from the Runai loader. Let me know if you want us to restore it.

python/sglang/srt/model_loader/loader.py
noa-neria: I don't see this function in `BaseModelLoader` 
It is in the `DefaultModelLoader`

python/sglang/srt/server_args.py
noa-neria: This must be the first call, before any use of `self.model_path`, because when downloading from object storage `self.model_path` is set to a local path.
The configuration files are downloaded to `self.model_path` while the saftetensors files are streamed directly from object storage to memory.

test/registered/runai_streamer/test_runai_model_loader.py
noa-neria: That was just a placeholder until the first ci run, as it is pulling the model from GCS bucket
Now it is set to 120 seconds

PR #17972, opened 2026-01-30
feat: expose consistent_hashing policy in Python router CLI args

bledden:
## Motivation

The consistent hashing routing policy is fully implemented in Rust (`ConsistentHashingPolicy`) and the Python binding enum (`PolicyType.ConsistentHashing`) already includes it, along with the string conversion in `policy_from_str()`. However, it was missing from the argparse `choices` lists for `--policy`, `--prefill-policy`, and `--decode-policy`, so users couldn't actually select it from the CLI.

## Changes

Added `"consistent_hashing"` to the `choices` list for:
- `--policy` (main routing policy)
- `--prefill-policy` (PD mode prefill)
- `--decode-policy` (PD mode decode)

Single file change in `router_args.py`.

## Test plan

- No new logic introduced — just adding a string to existing choices lists
- The policy itself already has comprehensive Rust tests in `consistent_hashing.rs`
- Existing `policy_from_str()` mapping and `PolicyType.ConsistentHashing` enum value are already wired up

Fixes #17802

### DIFF
--- a/sgl-model-gateway/bindings/python/src/sglang_router/router_args.py
+++ b/sgl-model-gateway/bindings/python/src/sglang_router/router_args.py
@@ -246,7 +246,15 @@ class RouterArgs:
             f"--{prefix}policy",
             type=str,
             default=RouterArgs.policy,
-            choices=["random", "round_robin", "cache_aware", "power_of_two", "manual"],
+            choices=[
+                "random",
+                "round_robin",
+                "cache_aware",
+                "power_of_two",
+                "manual",
+                "consistent_hashing",
+                "prefix_hash",
+            ],
             help="Load balancing policy to use. In PD mode, this is used for both prefill and decode unless overridden",
         )
         routing_group.add_argument(
@@ -260,6 +268,8 @@ class RouterArgs:
                 "power_of_two",
                 "manual",
                 "bucket",
+                "consistent_hashing",
+                "prefix_hash",
             ],
             help="Specific policy for prefill nodes in PD mode. If not specified, uses the main policy",
         )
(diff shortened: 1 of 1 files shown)


sgl-model-gateway/bindings/python/src/sglang_router/router_args.py
```diff
@@ -260,14 +267,22 @@ def add_cli_args(
...
             type=str,
             default=None,
-            choices=["random", "round_robin", "cache_aware", "power_of_two", "manual"],
+            choices=[
+                "random",
+                "round_robin",
+                "cache_aware",
+                "power_of_two",
+                "manual",
+                "consistent_hashing",
```
slin1237, line 286: there is also prefix_consistent_hashing
Can we add that too
Thanks for the fix!!

slin1237 (approved):

sgl-model-gateway/bindings/python/src/sglang_router/router_args.py
bledden: Added\! The Rust router maps this policy as `prefix_hash` (via `policy_from_str` in the Python bindings and the Rust factory which accepts both `prefix_hash` and `prefixhash`), so I've added `prefix_hash` to all three choices lists (`--policy`, `--prefill-policy`, `--decode-policy`).

bledden: @slin1237 to be clear I did `prefix_hashing` here since I could not find `prefix_consistent_hashing` in the repo, but if this should be explicitly `prefix_consistent_hashing` please lmk!

PR #18007, opened 2026-01-30
Fix Responses API crash for multimodal models

jhinpan:
## Summary

- Fix `/v1/responses` endpoint crashing with `"input_ids should be a list of lists for batch processing"` for multimodal models (e.g. Kimi-K2.5, LLaVA, Qwen-VL).
- Root cause: for multimodal models, the message processing pipeline returns a **text string** prompt, but `serving_responses.py` always passed it as `input_ids=` to `GenerateReqInput`. This caused the string to be misinterpreted as a batch of single-character elements.
- The fix mirrors the existing pattern in `serving_chat.py:260-266`: check `isinstance(engine_prompt, str)` and use `text=` for strings, `input_ids=` for token ID lists.

## Details

The Chat Completions handler (`serving_chat.py`) correctly dispatches:
```python
if is_multimodal:
    prompt_kwargs = {"text": processed_messages.prompt}
else:
    if isinstance(processed_messages.prompt_ids, str):
        prompt_kwargs = {"text": processed_messages.prompt_ids}
    else:
        prompt_kwargs = {"input_ids": processed_messages.prompt_ids}
```

But the Responses API handler (`serving_responses.py`) unconditionally used:
```python
adapted_request = GenerateReqInput(input_ids=engine_prompt, ...)
```

When `engine_prompt` is a string (multimodal case), `GenerateReqInput._determine_batch_size` sees `input_ids[0]` as a character (not an `int`), sets `is_single=False`, and `_expand_inputs` raises because a string is not a list of lists.

## Test plan

- [x] Tested `/v1/responses` (non-streaming) on Kimi-K2.5 (multimodal) — previously crashed, now returns correct reasoning + output
- [x] Tested `/v1/responses` (streaming) on Kimi-K2.5 — SSE events emitted correctly
- [x] Tested `/v1/chat/completions` still works (no regression)
- [x] Tested through SGLang Model Gateway router proxy — works on both endpoints


JustinTong0323:
Hi! Thanks for tackling this Responses API issue. To reduce reviewer & CI load across the many in-flight Responses-API fixes, the maintainer-track effort has been consolidated into **#25881 — Fix Responses API request handling**, which already incorporates the change this PR proposes (the fix is covered there with a unit test).

Could you take a look at #25881 when you have a moment? If you're happy, please feel free to close this PR — your contribution is credited in the PR body and on the consolidated commit's `Co-authored-by` trailer. AI-assisted PR volume has made it hard for maintainers to review every overlapping Responses-API fix in parallel, so consolidating into one correctness-focused PR is the most efficient path to landing these.

Thanks again for the work here — it directly informed the consolidated fix.

hnyls2002:
Thanks @jhinpan! The multimodal text-vs-input_ids dispatch fix landed on main via the consolidated #25881 - string prompts are now routed to text= and media is forwarded into GenerateReqInput. Closing as superseded; please reopen if I missed a case.

PR #18024, opened 2026-01-31
fix: correct weight loading prefix mapping for Qwen3-VL

Lollipop:
## Summary

Fix Qwen3-VL-8B model producing garbage output due to incorrect weight loading.

Fixes #17887

## Problem

The weight loading code unconditionally copies `embed_tokens.weight` to `lm_head.weight`:

```python
if self.pp_group.is_last_rank and "model.embed_tokens.weight" in name:
    if "lm_head.weight" in params_dict:
        # copies embed_tokens to lm_head unconditionally
```

This is incorrect for models with `tie_word_embeddings=False` (like Qwen3-VL-8B), where `lm_head` has independent weights that should not be overwritten.

| Model | tie_word_embeddings | lm_head weights |
|-------|---------------------|-----------------|
| Qwen3-VL-2B | True | Shared with embed_tokens |
| Qwen3-VL-8B | False | Independent (should NOT be overwritten) |

## Fix

Add a check to only copy when `tie_word_embeddings=True`:

```python
if (
    self.pp_group.is_last_rank
    and "model.embed_tokens.weight" in name
    and self.config.tie_word_embeddings  # <-- added check
):
```


Lollipop:
@JustinTong0323 Thanks for testing! The previous fix was incorrect - I've updated the PR with the correct fix now.

## Root Cause Clarification

The issue only affects models with `tie_word_embeddings=False`. Here's the difference:

| Model | tie_word_embeddings | Affected |
|-------|---------------------|----------|
| Qwen3-VL-2B | True | No |
| Qwen3-VL-4B | True | No |
| **Qwen3-VL-8B** | **False** | **Yes** |

For models with `tie_word_embeddings=True`, `embed_tokens` and `lm_head` share the same weights, so copying is correct.

For Qwen3-VL-8B (`tie_word_embeddings=False`), `lm_head` has its own independent weights. The unconditional copy overwrites these weights with `embed_tokens`, causing garbage output.

## Updated Fix

The new fix adds a check for `self.config.tie_word_embeddings`:

```python
if (
    self.pp_group.is_last_rank
    and "model.embed_tokens.weight" in name
    and self.config.tie_word_embeddings  # <-- only copy when weights are shared
):
```

Could you please test again with **Qwen3-VL-8B** specifically? The 2B/4B models should work fine with or without this fix.

JustinTong0323 (approved):
LGTM

PR #18070, opened 2026-02-01
Feat/add fi selective state update kernel call

shaharmor98:
## Motivation
FlashInfer provides optimized CUDA kernels for Mamba operations. This change allows users to optionally leverage FlashInfer's mamba kernels for potential performance improvements while maintaining backward compatibility with the existing Triton implementation.

This PR adds support for FlashInfer as an alternative backend for Mamba selective state update kernel.
It introduces `--mamba-backend` CLI argument to switch between `triton` (default) and `flashinfer` backends
Add dispatcher module (`ssu_dispatch.py`) that routes `selective_state_update` calls to the configured backend.

## Benchmarking
For the below benchmarking, we used the `nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16` model, with TP=1 on B200.

Below is comparison of performance on Output throughput (tok/s)
```bash
| BS  | w/o FI | w/ FI | Speedup_Factor |
|-----|--------|-------|----------------|
| 4   | 627.54 | 632.12 | 1.007298       |
| 16  | 1412.57| 1452.11| 1.027992       |
| 64  | 3060.60| 3224.70| 1.053617       |
| 128 | 4505.74| 4868.57| 1.080526       |
| 512 | 6740.35| 7593.40| 1.126559       |
```
throughput_plot

Full performance comparison:

```bash
| BS  | Configuration | Output_Throughput | Total_Throughput | Mean_E2E_Latency | Mean_TTFT |
|-----|---------------|-------------------|------------------|------------------|-----------|
| 4   | w/ FI         | 632.12            | 1366.12          | 2938.45          | 105.85    |
| 4   | w/o FI        | 627.54            | 1356.23          | 2960.14          | 100.85    |
| 16  | w/ FI         | 1452.11           | 2988.50          | 4777.32          | 117.76    |
| 16  | w/o FI        | 1412.57           | 2907.14          | 4916.66          | 116.68    |
| 64  | w/ FI         | 3224.70           | 6560.51          | 8755.97          | 219.08    |
| 64  | w/o FI        | 3060.60           | 6226.65          | 9260.55          | 221.39    |
| 128 | w/ FI         | 4868.57           | 9662.67          | 12132.85         | 309.62    |
| 128 | w/o FI        | 4505.74           | 8942.57          | 13133.84         | 311.07    |
| 512 | w/ FI         | 7593.40           | 15306.34         | 31143.92         | 11270.59  |
| 512 | w/o FI        | 6740.35           | 13586.80         | 35087.40         | 12624.50  |
```

## Checklist

## Review Process


python/pyproject.toml
```diff
@@ -21,7 +21,7 @@ dependencies = [
   "anthropic>=0.20.0",
   "av ; sys_platform == 'linux' and (platform_machine == 'aarch64' or platform_machine == 'arm64' or platform_machine == 'armv7l')",
   "blobfile==3.0.0",
-  "build",
+  "build",
```
yizhang2077: remove space

python/sglang/srt/managers/scheduler.py
```diff
@@ -484,6 +490,9 @@ def init_tokenizer(self):
                 reasoning_parser.detector.think_end_token, add_special_tokens=False
             )[0]
 
+    def init_mamba_backend(self) -> None:
+        initialize_mamba_selective_state_update_backend(self.server_args)
```
yizhang2077, line 499: maybe we need to rename it `maybe_xxx`, and models which do not use mamba backend do not need to update backend

python/pyproject.toml
shaharmor98: My bad

python/sglang/srt/managers/scheduler.py
shaharmor98: Hi @yizhang2077  
Thank you for your comment.  
As I see it, we have two alternatives  
The first:  
```code
    def maybe_init_mamba_backend(self) -> None:
        """Initialize the mamba selective-state-update backend only for mamba models."""
        model_config = self.server_args.get_model_config()
        architectures = getattr(model_config.hf_config, "architectures", None) or []
        arch = architectures[0] if architectures else ""
        # Architectures that use the mamba SSU backend (align with model_runner.mamba2_config / mambaish_config)
        mamba_architectures = (
            "Lfm2ForCausalLM",
            "NemotronHForCausalLM",
            "Qwen3NextForCausalLM",
            "KimiLinearForCausalLM",
            "FalconH1ForCausalLM",
        )
        if arch in mamba_architectures:
            initialize_mamba_selective_state_update_backend(self.server_args)
```
This is more "rigorous" testing, which ensures it runs only for mamba backends. However, it includes the `mamba_architecture` list, which is a big downside.

Alternatively, I could just check whether the mamba-backend flag is not the default one, and only if it’s non-default, run the `initialize_mamba_selective_state_update_backend` method.
WDYT?

yizhang2077: I think the second is ok for me

shaharmor98: Hi @yizhang2077  
I'm sorry, but after taking a deeper look, it seems like my proposal won't work.  
`triton` is the default mamba-backend kernel, and in both cases (default and non-default), this value is used when the model is mamba-based.  
So, we cannot skip this method unless:  
1. we choose the first approach presented above (I vote against it)  
2. we keep the code intact. I think it's fine, since even for non-mamba-based models, the only thing that would happen is the following import: `from sglang.srt.layers.attention.mamba.ops.mamba_ssm import (selective_state_update,)`, which I believe is not as expensive as flashinfer's one.  

I think we should stick with 2, but I'm open to other suggestions
WDYT?

yizhang2077: ok, 2 is also fine for me

yizhang2077:
> > > > Could we integrate this into models like `nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16`? We need to make sure it will be used. Or will it be integrated in the following pr?
> > > 
> > > 
> > > Hi @yizhang2077 This PR adds the ability to use this kernel via the CLI param --mamba-backend. Just like the other flags, if someone sets --mamba-backend flashinfer, then the above kernel will be used :) Does that answer your question?
> > 
> > 
> > I mean we have `selective_state_update` in `sglang.srt.layers.attention.mamba.ops`, and we use it in `sglang.srt.layers.attention.mamba.mamba.py`, do we need to replace it into `selective_state_update` in `ssu_dispatch` (backend in your PR) then it will actually work?
> 
> Ah, now I understand what you mean. The answer is no :) Please note: the `selective_state_update` call that is used in `sglang.srt.layers.attention.mamba.mamba.py`, is imported via:
> 
> ```
> from sglang.srt.layers.attention.mamba.ops import (
>     mamba_chunk_scan_combined,
>     selective_state_update,
> )
> ```
> 
> The changes I added in my PR, specifically in: `python/sglang/srt/layers/attention/mamba/ops/__init__.py` mean that when `selective_state_update` is called- it's now dispatched through the `ssu_dispatch`, which, as discussed before, initializes the correct backend based on the CLI param.
> 
> So to sum up, the answer is no - currently the code auto-selects the correct backend without further integration.

oh I see

yizhang2077 (approved):

ispobock (approved):

PR #18082, opened 2026-02-01
feat: Add LoRA CUDA graph support for FlashAttention and FlashInfer backends

sbeurnier:
## Motivation

This PR addresses a performance issue where non-LoRA batches incur unnecessary overhead when LoRA is enabled on the engine. 

**Problem:** When LoRA is enabled on the engine, all batches (including those without LoRA requests) go through LoRA code paths. Non-LoRA batches unnecessarily call `prepare_lora_batch()` and launch LoRA-specific kernels, causing significant performance degradation compared to running with LoRA disabled.

**Solution:** This PR introduces separate CUDA graph capture and replay for LoRA and non-LoRA batches. By capturing both non-LoRA and LoRA-specific graphs and dynamically selecting the appropriate graph at runtime based on whether a batch contains LoRA requests, non-LoRA batches now use the non-LoRA graph path while LoRA batches use the LoRA-specific graph path. This maintains full LoRA functionality while restoring non-LoRA performance to baseline levels.

### Core Changes

1. **Attention Backend Interface (`base_attn_backend.py`)**
   - Added `supports_separate_lora_graphs()` method to allow backends to opt into separate LoRA graph support
   - Added `is_lora` parameter to `init_forward_metadata_capture_cuda_graph()` and `init_forward_metadata_replay_cuda_graph()` methods
   - Added documentation for LoRA-aware CUDA graph metadata handling

2. **Backend Implementations (16 files)**
   - **FlashAttention & FlashInfer backends**: Implemented `supports_separate_lora_graphs()` returning `True` to enable separate LoRA graphs
   - **Other backends** (AIter, CUTLASS MLA, Dual Chunk, FlashInfer MLA, FlashMLA, Hybrid, Hybrid Linear, NSA, TBO, Triton, TRT-LLM MHA/MLA, Wave): Return `False` (default behavior)
   - Updated all backends to accept `is_lora` parameter in CUDA graph methods

3. **CUDA Graph Runner (`cuda_graph_runner.py`)**
   - Modified graph key generation to include LoRA state when backend supports separate graphs
   - Added logic to select correct graph based on whether batch has LoRA requests
   - Enhanced graph lookup to consider LoRA state for proper graph differentiation

4. **Forward Batch Info (`forward_batch_info.py`)**
   - Added tracking of LoRA adapter presence in batches
   - Updated graph key computation to include LoRA state

5. **LoRA Layers (`lora/layers.py`)**
   - Updated to work with LoRA-aware CUDA graph system

6. **Piecewise CUDA Graph Runner (`piecewise_cuda_graph_runner.py`)**
   - Updated to pass `is_lora` parameter to attention backend methods

### Key Design Decisions

- Backends opt-in to separate LoRA graph support via `supports_separate_lora_graphs()`
- Graph keys are extended to `(batch_size, is_lora)` for supporting backends
- Backward compatible: backends without support continue to work with existing behavior

## Accuracy & Speed Tests

I benchmarked the speed using the following command on a standard SGlang server with a flash attention (fa3) backend:

```
  --backend sglang \
  --dataset-name random \
  --num-prompts 200 \
  --request-rate 4 \
  --random-input-len 2048 \
  --random-output-len 1024 \
  --disable-ignore-eos \
  --disable-tqdm
```

Here are the results on 1xH100 running inference for GPT OSS 20B:

| Metric | LoRA Disabled (Baseline) | LoRA Enabled (This PR) | LoRA Enabled (Before PR) | 
|--------|-------------------------|------------------------|-------------------------|
| **TTFT Median (ms)** | 49.12 | 50.32 | 56.5 | 
| **TTFT P99 (ms)** | 55.12 | 55.66 | 63.4 | 
| **P99 Round-trip Latency (ms)** | 9,381 | 9,465 | 9,850 | 
| **Job-level Decode TPS** | 3,504 | 3,507 | 3,329 | 

We also tested the new image in production and saw similar perf improvements of around 15% for TTFT and 5-7% for TPS. We also benchmarked the performance between main and this branch using same command above but with 7 copies of the same adapter (sglang does not deduplicate adapters) to ensure there was no regression for lora traffic:

```
for i in $(seq 1 8); do
  curl -s -X POST http://localhost:30001/load_lora_adapter \
    -H 'Content-Type: application/json' \
    -d "{
      \"lora_name\": \"adapter${i}\",
      \"lora_path\": \"/sbeurnier/adapters/gpt-oss20b\"
    }"
done

  --backend sglang \
  --base-url http://localhost:30001 \
  --dataset-name random \
  --num-prompts 200 \
  --request-rate 4 \
  --random-input-len 2048 \
  --random-output-len 1024 \
  --disable-ignore-eos \
  --disable-tqdm \
  --lora-name \
    adapter1 \
    adapter2 \
    adapter3 \
    adapter4 \
    adapter5 \
    adapter6 \
    adapter7
```
Results show negligible performance differences for lora requests:

| Metric | Before PR | After PR | 
|--------|-----------|----------|
| **TTFT Median (ms)** | 55.94 | 56.17 |
| **TTFT P99 (ms)** | 66.02 | 67.82 | 
| **P99 Round-trip Latency (ms)** | 11293 | 11037 | 
| **Job-level Decode TPS** | 3428.94 | 3427.93 | 

In terms of accuracy, there are no changes to model outputs. This PR only affects CUDA graph capture/replay mechanics and path selection. Even though less kernels are executed for nonlora batches, these kernels were originally filled with 0's and end early, so they only added launch overhead.

Profiling also shows that CPU no longer calls prepare_lora_batch and that the LoRa kernels are no longer executed on the GPU.

Before:
image

image

With PR:
image

image

## Checklist


Fridge003 (changes requested):
Can we avoid adding the `is_lora` argument for every attention backend?
This design is a little invasive

python/sglang/srt/layers/attention/flashattention_backend.py
```diff
@@ -1629,8 +1629,14 @@ def init_forward_metadata_capture_cuda_graph(
         encoder_lens: Optional[torch.Tensor],
         forward_mode: ForwardMode,
         spec_info: Optional[SpecInput],
+        is_lora: bool = False,
     ):
-        """Initialize forward metadata for capturing CUDA graph."""
+        """Initialize forward metadata for capturing CUDA graph.
+
+        Args:
```
glenliu21, line 1636: In general, can we reduce the number of comments and additions to doc-strings that we make? Sometimes I think it can be too much and it can quickly become confusing.

For instance, I think this docstring addition is largely redundant with the docstring additions made in `base_attn_backend.py`.

python/sglang/srt/lora/layers.py
```diff
@@ -43,6 +43,39 @@ def forward(self, x: torch.Tensor):
...
+        # (prepare_lora_batch() was called, so LoRA kernels should be launched)
+        if (
+            hasattr(self.lora_backend, "batch_info")
+            and self.lora_backend.batch_info is not None
+        ):
+            return True
+
+        # If batch_info doesn't exist, we're in a non-LoRA path
+        # (prepare_lora_batch() was NOT called, so skip LoRA entirely)
+        return False
```
glenliu21, line 77: Can we combine some of these checks together to make this function smaller? I also think the comments are largely self-explanatory here.

python/sglang/srt/model_executor/cuda_graph_runner.py
```diff
@@ -378,6 +378,50 @@ def maybe_init_pdmux(self):
...
+    def _supports_separate_lora_graphs(self) -> bool:
+        """Check if the attention backend supports separate LoRA and non-LoRA graphs.
+
+        Backends that support this key metadata by (bs, is_lora) to prevent overwrites
+        when both graph types are captured for the same batch size.
+
+        For non-supported backends, we only capture the LoRA graph and use it
+        for all batches (LoRA kernels no-op when adapters are zeroed).
+        """
+        return self.model_runner.attn_backend.supports_separate_lora_graphs()
```
glenliu21, line 397: I think this function is probably unnecessary, and we can just directly call the inner function.

python/sglang/srt/model_executor/cuda_graph_runner.py
```diff
@@ -680,17 +782,37 @@ def capture_one_batch_size(
...
-        # Attention backend
-        attn_backend.init_forward_metadata_capture_cuda_graph(
-            bs,
-            num_tokens,
-            req_pool_indices,
-            seq_lens,
-            encoder_lens,
-            forward_batch.forward_mode,
-            forward_batch.spec_info,
-        )
```
glenliu21, line 693: Can we use kwargs and splatting to combine the `if` and `else` branches, given that they only differ by `is_lora`?

python/sglang/srt/model_executor/forward_batch_info.py
```diff
@@ -518,12 +518,41 @@ def init_new(
...
         # Init lora information
         if model_runner.server_args.enable_lora:
-            # In the non-LoRA overlap loading case, we fetch LoRA adapters into the memory pool
-            # as a batch, right before running the batch
-            if not model_runner.server_args.enable_lora_overlap_loading:
-                model_runner.lora_manager.fetch_new_loras(set(ret.lora_ids))
+            # Check if batch has LoRA requests (same logic as _has_lora_requests in cuda_graph_runner)
+            has_lora_requests = ret.lora_ids is not None and any(
+                lora_id is not None for lora_id in ret.lora_ids
+            )
```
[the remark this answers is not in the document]
glenliu21: I think having some sort of helper method like this sounds reasonable.

python/sglang/srt/model_executor/cuda_graph_runner.py
sbeurnier: not opposed to this but see fridge's earlier comment about touching a lot of files, this would require modifying every attention backend which is a little invasive.

glenliu21: Ah okay, I see - that's my bad. The current approach is fine then.

python/sglang/srt/model_executor/cuda_graph_runner.py
```diff
@@ -817,22 +939,37 @@ def replay_prepare(
             )
         if forward_batch.forward_mode.is_idle() and forward_batch.spec_info is not None:
             forward_batch.spec_info.custom_mask = buffers.custom_mask
-        # Attention backend
+        # Attention backend - pass is_lora only to backends that support separate LoRA graphs
+        has_lora = self._has_lora_requests(forward_batch)
```
ConnorLi96, line 943: this condition happens 3 times, pls follow **DRY (Don't Repeat Yourself)** principle.

python/sglang/srt/model_executor/forward_batch_info.py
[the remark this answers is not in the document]
ConnorLi96: yes, agree

python/sglang/srt/model_executor/cuda_graph_runner.py
```diff
@@ -378,6 +378,50 @@ def maybe_init_pdmux(self):
...
+        if enable_pdmux:
+            stream_part = (
+                stream_idx if stream_idx is not None else get_current_stream_idx()
+            )
+            base_key = f"{stream_part}_{bs}"
+
+        # Only add LoRA suffix if LoRA is enabled at the server level
+        # This allows separate graphs for non-LoRA batches when LoRA is enabled
+        if enable_lora and has_lora:
+            return f"{base_key}_lora"
```
ConnorLi96, line 422: a little bit hard coding, maybe use `LORA_GRAPH_SUFFIX = "_lora"`

python/sglang/srt/model_executor/cuda_graph_runner.py
```diff
@@ -680,17 +782,37 @@ def capture_one_batch_size(
...
         if lora_ids is not None:
             self.model_runner.lora_manager.prepare_lora_batch(forward_batch)
+        else:
+            # Clear batch_info for non-LoRA graph capture to ensure _should_apply_lora() returns False
+            # This prevents stale batch_info from previous LoRA graph capture from causing
+            # LoRA kernels to be launched during non-LoRA graph capture.
+            if self.model_runner.server_args.enable_lora and hasattr(
+                self.model_runner.lora_manager.lora_backend, "batch_info"
+            ):
+                self.model_runner.lora_manager.lora_backend.batch_info = None
```
ConnorLi96, line 792: does it really work in race condition? if just simple set it as `None`

python/sglang/srt/model_executor/cuda_graph_runner.py
```diff
@@ -503,14 +561,52 @@ def _capture_one_stream(stream_idx: Optional[int] = None):
                     num_tokens=bs * self.num_tokens_per_bs,
                     tp_group=self.model_runner.tp_group,
                 ) as forward:
+                    # Capture graph with LoRA logic (if LoRA is enabled)
```
ConnorLi96, line 564: this `can_run` function is over long, can you move this long logic to lora part?

python/sglang/srt/model_executor/cuda_graph_runner.py
```diff
@@ -680,17 +782,37 @@ def capture_one_batch_size(
...
-            bs,
-            num_tokens,
-            req_pool_indices,
-            seq_lens,
-            encoder_lens,
-            forward_batch.forward_mode,
-            forward_batch.spec_info,
-        )
+        # Attention backend - pass is_lora only to backends that support separate LoRA graphs
+        if self._supports_separate_lora_graphs():
```
ConnorLi96, line 795: Overall, can you add a flag in launch command to disable two cuda graphs? which means that by default we can enable it for better performance, but we can give user a option to disable it if memory is limited especially for big models.

PR #18127, opened 2026-02-02
fix streaming parallel tool calls with tool_choice=auto for qwen25

harvenstar:
Fixes #18001: 

when streaming multiple tool calls with tool_choice="auto", only the first tool call was emitted because parse_streaming_increment mistakenly treated the start of the eot token as a separator, for example separator \n matching the leading \n</tool_call>, which broke parsing of the second tool call. This patch skips separator matching when the candidate match is actually the prefix of an eot token. Impacts the Qwen25 and Trinity detectors.


hnyls2002:
@harvenstar I think this has been superseded: base_format_detector.parse_streaming_increment on main now tracks a used_separator_branch flag and, on a MalformedJSON/JSONDecodeError, falls back to searching for bot_token past inter-object markup. That handles the separator-vs-eot_token collision your is_separator guard targeted. Closing as superseded; reopen if there's a case the fallback still misses.

PR #18047, opened 2026-01-31
[TestFix] rewrite LoRA overlap loading tests

glenliu21:
## Motivation

The existing test implementation is flaky (see https://github.com/sgl-project/sglang/actions/runs/21320657015/job/61370002606).

## Modifications

- Use existing test utils to write new tests

## Accuracy Tests

- LoRA overlap loading tests pass locally

## Checklist

- [x] Format your code according to the [Format code with pre-commit](https://docs.sglang.io/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [x] Add unit tests according to the [Run and add unit tests](https://docs.sglang.io/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [x] Follow the SGLang code style [guidance](https://docs.sglang.io/developer_guide/contribution_guide.html#code-style-guidance).

## Review Process

1. Ping Merge Oncalls to start the PR flow. See the [PR Merge Process](https://github.com/sgl-project/sglang/blob/main/.github/MAINTAINER.md#pull-request-merge-process).
2. Get approvals from [CODEOWNERS](https://github.com/sgl-project/sglang/blob/main/.github/CODEOWNERS) and other reviewers.
3. Trigger CI tests with [comments](https://docs.sglang.io/developer_guide/contribution_guide.html#how-to-trigger-ci-tests) or contact authorized users to do so.
   - `/tag-run-ci-label`, `/rerun-failed-ci`, `/tag-and-rerun-ci`
4. After green CI and required approvals, ask Merge Oncalls to merge.

PR #18133, opened 2026-02-02
Fix HFRunner LoRA adapter contamination between prompts

alisonshao:
## Summary
Fixes LoRA adapter contamination issue in HFRunner where calling `PeftModel.from_pretrained()` multiple times on the same base model causes the warning:
```
UserWarning: Already found a `peft_config` attribute in the model. This will lead to having multiple adapters in the model.
```

## Changes
- Track `self.peft_model` instance variable to persist PeftModel wrapper across batches
- Use `load_adapter()` and `set_adapter()` for subsequent LoRA adapters instead of creating new PeftModel instances
- Use `disable_adapter_layers()` / `enable_adapter_layers()` to switch between LoRA and base model modes

## Testing
Tested on H200 machine with `test/registered/lora/test_lora_overlap_loading.py`:
- The adapter contamination warning is eliminated
- `test_ci_lora_models_batch_splitting` passes consistently

Note: `test_ci_lora_models_multi_batch` still fails due to a separate issue where SRT and HF produce different outputs for the same prompt (even without LoRA adapters). This is a known flakiness issue (see #17683, #18047) and is unrelated to adapter contamination.

## Related
- CI failure example: https://github.com/sgl-project/sglang/actions/runs/13606682814/job/38015694785
- Related flakiness issues: #17683, #18047

### DIFF
--- a/python/sglang/test/runners.py
+++ b/python/sglang/test/runners.py
@@ -272,6 +272,8 @@ class HFRunner:
                 trust_remote_code=self.trust_remote_code,
                 low_cpu_mem_usage=True,
             ).cuda()
+            # Track PeftModel wrapper for LoRA adapter management across batches
+            self.peft_model = None
         elif self.model_type == "embedding":
             if "gme-qwen2-vl" in model_path.lower():
                 self.model = AutoModelForVision2Seq.from_pretrained(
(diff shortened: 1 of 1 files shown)


alisonshao:
fixed by #18140

PR #18152, opened 2026-02-03
Fix Session for multimodal and expose it through Engine

aurickq:


python/sglang/srt/managers/scheduler.py
```diff
@@ -1503,11 +1508,23 @@ def handle_generate_request(
                 self.init_req_max_new_tokens(req)
                 self._add_request_to_queue(req)
                 return
+            session_prefix_len = len(req.origin_input_ids) - len(recv_req.input_ids)
```
mickqian: nit: would a single line of this works for both branches?

mickqian (approved):

python/sglang/srt/managers/scheduler.py
aurickq: done

PR #18157, opened 2026-02-03
[Multimodal] Add multimodal special tokens for video&audio in Qwen_vl

coding-famer:
## Motivation

Update qwen_vl's `MultimodalSpecialTokens` to support video&audio token's regex

## Checklist

### DIFF
--- a/python/sglang/srt/multimodal/processors/qwen_vl.py
+++ b/python/sglang/srt/multimodal/processors/qwen_vl.py
@@ -256,8 +256,16 @@ class QwenVLImageProcessor(SGLangBaseProcessor):
             image_token_regex=re.compile(
                 r"<\|vision_start\|>(?:<\|image_pad\|>)+<\|vision_end\|>"
             ),
+            video_token="<|video_start|><|video_pad|><|video_end|>",
             video_token_id=hf_config.video_token_id,
+            video_token_regex=re.compile(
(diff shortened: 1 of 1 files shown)


hnyls2002:
@coding-famer I'm fairly sure this is no longer needed: MultiModalSpecialTokens on main already defines video_token, audio_token, and the *_regex fields (see multimodal/processors/base_processor.py), with auto-fill from token ids in __post_init__. The qwen_vl.py changes here also appear to have syntax issues (missing commas between kwargs). Closing as obsolete; reopen if there's a specific video/audio regex case main still doesn't cover.

JingwenGu0829:
@hnyls2002 Hi. I think the fields exist, but the auto-filled regex only matches one <|video_pad|> / <|audio_pad|>, not one full processor-expanded media span. As a result, one expanded video is parsed as multiple video occurrences and mismatches the single video_data item.

We need to handle this because we are working on supporting the video and audio input in the Miles and MCore training contract, which uses processor-expanded IDs together with the processed media tensors, so SGLang rollout needs to accept the same expanded IDs plus the media. 

Qwen2.5-VL, Qwen3-VL, and Qwen3-Omni have different expanded layouts, so we’ll add the corresponding grouped regexes and a regression test in a follow-up. 

All in all I think Sglang currently has indeed a incorrect parsing contract in multimodal input. But we will rework on the matching correct regex instead of following this PR.

PR #18172, opened 2026-02-03
[NPU]Support model Trinity-mini for Npu, accuracy 90%

McZyWu:
## Motivation

Previously, model trinity mini is not supported for npu.

## Modifications

As follows
custom_routing_function was used because the trinity mini model (afmoe.py) has its unique routing function.
## Accuracy Tests
90%, surpass GPU.
image
image
image
image
image


python/sglang/srt/hardware_backend/npu/moe/topk.py
```diff
@@ -25,8 +25,13 @@ def fused_topk_npu(
     use_grouped_topk = topk_config.use_grouped_topk
     renormalize = topk_config.renormalize
     correction_bias = topk_config.correction_bias
+    custom_routing_function = topk_config.custom_routing_function
```
ping1jing2: please explain why we need this variable here in the PR description

python/sglang/srt/hardware_backend/npu/quantization/fused_moe_method_npu.py
```diff
@@ -381,6 +381,13 @@ def fused_moe_npu(
         expanded_src_to_dst_row=expanded_row_idx,
         export_for_source_row=topk_ids,
     )
+
+    if (
+        moe_runner_config.routed_scaling_factor is not None
+        and abs(moe_runner_config.routed_scaling_factor - 1.0) > 1e-2
```
ping1jing2: what's the meaning of magic number `1e-2`?

python/sglang/srt/layers/moe/fused_moe_native.py
```diff
@@ -97,11 +97,24 @@ def moe_forward_native(
...
+        )
+    else:
+        final_out = (
+            new_x.view(*topk_ids.shape, -1)
+            .type(topk_weights.dtype)
+            .mul_(topk_weights.unsqueeze(dim=-1))
+            .sum(dim=1)
+            .type(new_x.dtype)
+            * moe_runner_config.routed_scaling_factor
+        )
```
ping1jing2: it might be better to write below code:
```python
    if (
        moe_runner_config.routed_scaling_factor is not None
        and abs(moe_runner_config.routed_scaling_factor - 1.0) >= 1e-5
    ):
        final_out.mul_(moe_runner_config.routed_scaling_factor)
```

python/sglang/srt/models/afmoe.py
```diff
@@ -58,7 +57,19 @@
...
+_is_cuda = is_cuda()
+
+if _is_cuda:
+    from sglang.srt.layers.moe.fused_moe_triton import fused_moe
+elif _is_npu:
+    from sglang.srt.hardware_backend.npu.quantization.fused_moe_method_npu import (
+        fused_moe_npu,
+    )
+else:
+    from sglang.srt.layers.moe.fused_moe_native import moe_forward_native
```
ping1jing2: please explain the below questions in your description:
1. why do we need to import moe_forward_native here?
2. could you use `from sglang...fused_moe_method_npu import fused_moe_npu as fused_moe` to avoid using `if-elif-else`

python/sglang/srt/models/afmoe.py
```diff
@@ -255,6 +273,10 @@ def pack_params(self) -> None:
         for data, param in zip(w2s, w2):
             param.data = data
         self.w2 = self.w2.view(len(w2), *w2s[0].shape)
+        if not _is_cuda and not _is_npu:
```
ping1jing2: please use `if _is_xxx` instead of `if not _is_xxx`, it's hard to understand the code logic

python/sglang/srt/models/afmoe.py
```diff
@@ -266,16 +288,27 @@ def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
...
+                hidden_states,
+                w1=self.w1,
+                w2=self.w2,
+                topk_output=topk_output,
+                moe_runner_config=MoeRunnerConfig(
+                    inplace=True,
+                    routed_scaling_factor=self.route_scale,
+                ),
+            )
+        else:
```
ping1jing2: actually, i still don't know in which scenarios this branch will be called.

python/sglang/srt/models/afmoe.py
```diff
@@ -314,8 +347,8 @@ def __init__(
         self.kv_size = self.num_kv_heads * self.head_dim
         self.scaling = self.head_dim**-0.5
 
-        rope_theta = config.rope_parameters["rope_theta"]
-        rope_scaling = config.rope_parameters
+        rope_theta = config.rope_theta
+        rope_scaling = config.rope_scaling
```
ping1jing2: why do we need these change? bugfix or not

ping1jing2 (changes requested):

python/sglang/srt/layers/moe/fused_moe_native.py
```diff
@@ -97,11 +99,20 @@ def moe_forward_native(
...
+            .type(new_x.dtype)
+        )
+    else:
+        final_out = (
+            new_x.view(*topk_ids.shape, -1)
+            .type(topk_weights.dtype)
+            .mul_(topk_weights.unsqueeze(dim=-1))
+            .sum(dim=1)
+            .type(new_x.dtype) * moe_runner_config.routed_scaling_factor
+        )
```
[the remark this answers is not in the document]
ping1jing2: make sense

python/sglang/srt/hardware_backend/npu/moe/topk.py
McZyWu: Ok. Thanks

python/sglang/srt/models/afmoe.py
McZyWu: bugfix, for transformer 5.3.0

python/sglang/srt/models/afmoe.py
```diff
@@ -227,16 +236,29 @@ def __init__(
...
+                top_k=self.top_k,
+                renormalize=False,
+                use_grouped_topk=self.use_grouped_topk,
+                num_expert_group=self.n_group if self.use_grouped_topk else None,
+                topk_group=self.topk_group if self.use_grouped_topk else None,
+                custom_routing_function=custom_routing_fn,
+                correction_bias=self.expert_bias,
+                scoring_func=self.score_func,
+                routed_scaling_factor=self.route_scale,
+            )
```
iforgetmyname: revert to using the default topk settings
there's no performance requirements on this model

McZyWu: Adopted. Thanks

ping1jing2 (approved):

ping1jing2:
I merged it as all GPU CIs passed

PR #18205, opened 2026-02-04
feat(kv-events): Add medium field to KV event types for storage tier tracking

ishandhanani:
## Summary

Add `medium` field to `BlockStored` and `BlockRemoved` event classes in `kv_events.py` to indicate which storage tier the block resides in. This enables KV-aware routers to track SGLang KV blocks across all hi-cache storage tiers (L1/GPU, L2/Host, L3/Storage).

### Changes
- Add `MEDIUM_GPU`, `MEDIUM_CPU_TIER1`, `MEDIUM_CPU_TIER2` constants to `kv_events.py`
- Add optional `medium: Optional[str]` field to `BlockStored` and `BlockRemoved` classes
- Update `radix_cache.py` to emit events with `medium="GPU"` for L1 cache operations

### Compatibility
- Field is optional (`None` by default) for backward compatibility
- Uses vLLM-compatible values: `"GPU"` (L1), `"CPU_TIER1"` (L2), `"CPU_TIER2"` (L3)
- Events serialize correctly with msgpack

## Test Plan

- [x] All existing unit tests pass (`test_radix_cache_unit.py` - 35 tests)
- [x] Verified msgpack serialization/deserialization works correctly
- [x] Verified events are emitted with `medium="GPU"` via ZMQ subscriber
- [x] Backward compatibility maintained (field is optional)


python/sglang/srt/disaggregation/kv_events.py
```diff
@@ -56,16 +56,24 @@ class KVCacheEvent(
...
+MEDIUM_CPU_TIER2 = "CPU_TIER2"
+
+
 class BlockStored(KVCacheEvent):
     block_hashes: list[int]
     parent_block_hash: Optional[int]
     token_ids: list[int]
     block_size: int
     lora_id: Optional[int]
+    medium: Optional[str] = None
```
stmatengss, line 70: Why not use an enum for medium?

stmatengss: I should be better to extend the medium type.

stmatengss (approved):

python/sglang/srt/disaggregation/kv_events.py
```diff
@@ -56,16 +56,24 @@ class KVCacheEvent(
     """Base class for all KV cache-related events"""
 
 
+# Medium values for storage tiers (compatible with vLLM)
+MEDIUM_GPU = "GPU"
+MEDIUM_CPU_TIER1 = "CPU_TIER1"
```
doujiang24: Does `CPU_TIER1` means HiRadixCache? And `CPU_TIER2` means remote cpu memory pool?

python/sglang/srt/disaggregation/kv_events.py
ishandhanani: yea as you saw - might be less friendly for serialization...

python/sglang/srt/disaggregation/kv_events.py
ishandhanani: It's a bit confusing actually.

- `CPU_TIER1` is actually `L2` storage. Its the `host_to_device` and `device_to_host` buffers in `HiRadixCache`. Pinned memory
- `CPU_TIER2` is actually `L3` storage. This is the remote storage backend

 I'm not sure why VLLM does it this way. I am ok with changing this btw

doujiang24: How about `MEDIUM_CPU = "CPU_PINNED"`? As discussed in slack.
And let's the backend storage to emit the kvevents, then, no need `MEDIUM_CPU_TIER2` here?

ishandhanani: Sure - changed to `CPU_PINNED` and for now have removed the L3 events

doujiang24 (approved):
Great, LGTM

PR #18211, opened 2026-02-04
[CI] Fix torchaudio/torchvision CUDA version mismatch

alisonshao:
## Summary
- Add `torchaudio` and `torchvision` to `[tool.uv.sources]` config in `pyproject.toml`
- Ensures PyTorch ecosystem packages are resolved from the same package index based on platform
- Fixes CUDA version mismatch where torch gets cu128 from PyPI while torchaudio/torchvision get cu129 from extra-index-url

## Root Cause
The existing `[tool.uv.sources]` config only specified source overrides for `torch`, but not for `torchaudio` and `torchvision`. This caused inconsistent CUDA versions:
- `torch` from PyPI (cu128)
- `torchaudio`/`torchvision` from extra-index-url (cu129)

The CUDA version mismatch led to circular import errors when loading multimodal processors that require torchaudio.

## Example Failure
https://github.com/sgl-project/sglang/actions/runs/21652299684/job/62419467816

Error: `AttributeError: partially initialized module 'torchaudio' has no attribute 'lib' (most likely due to a circular import)`

## Test plan
- [ ] CI passes with torchaudio-dependent VLM tests (TestMiniCPMo26Server)
- [ ] Verify `torch.version.cuda` matches torchaudio's CUDA version in CI environment


alisonshao:
## Test Results (Latest - 3 commits)

Tested on H200 GPU with `pip install -e python[dev]`:

```
torch version: 2.9.1+cu128
torch CUDA: 12.8
torchaudio version: 2.9.1+cu128
torchvision version: 0.24.1+cu128
sglang version: 0.0.0.dev1+gef3b24662

All imports successful!
```

✅ All PyTorch packages have matching CUDA versions (cu128)
✅ No circular import error
✅ sglang imports correctly

scripts/ci/cuda/ci_install_dependency.sh
```diff
@@ -275,5 +280,48 @@ fi
 $PIP_CMD list
 python3 -c "import torch; print(torch.version.cuda)"
 
-# Prepare the CI runner (cleanup HuggingFace cache, etc.)
+# Clean up stale HuggingFace cache artifacts BEFORE pre-caching
```
Kangyan-Zhou: Let's remove this change?

python/pyproject.toml
```diff
@@ -93,9 +93,16 @@ url = "https://download.pytorch.org/whl/cu129"
 explicit = true
 
 [tool.uv.sources]
+# Use pytorch.org wheel index for all torch packages to ensure matching CUDA versions
+# PyPI has mismatched CUDA versions (torch cu128, torchaudio cu129)
 torch = [
-  { index = "pypi", marker = "platform_machine == 'x86_64'"},
-  { index = "torch-cu129", marker = "platform_machine == 'aarch64'"},
+  { index = "torch-cu129" },
```
hnyls2002: Why remove platform configs? cc @Fridge003

scripts/ci/cuda/ci_install_dependency.sh
```diff
@@ -128,6 +128,22 @@ echo "Installing python extras: [${EXTRAS}]"
...
 $PIP_CMD install -e "python[${EXTRAS}]" --extra-index-url https://download.pytorch.org/whl/${CU_VERSION} $PIP_INSTALL_SUFFIX
 
+# Fix CUDA version mismatch between torch and torchaudio.
+# PyPI's torch 2.9.1 bundles cu128 but torchaudio from pytorch.org/cu129 uses cu129.
+# This mismatch causes torchaudio's C extension to fail loading, producing:
+#   "partially initialized module 'torchaudio' has no attribute 'lib'"
+# We cannot replace torch with cu129 (breaks sgl_kernel ABI), so instead we reinstall
+# torchaudio/torchvision from an index matching torch's CUDA version.
+TORCH_CUDA_VER=$(python3 -c "import torch; v=torch.version.cuda; parts=v.split('.'); print(f'cu{parts[0]}{parts[1]}')")
+echo "Detected torch CUDA version: ${TORCH_CUDA_VER}"
```
hnyls2002, line 138: Why should we also modify `pyproject.toml`? If this PR is to fix the CI failure, is the modification of `ci_install_dependency.sh` enough?

python/pyproject.toml
Fridge003: We shouldn't remove platform configs....

alisonshao: Reverted — restored the original platform-specific markers. The pyproject.toml change has been dropped entirely since the CI fix is self-contained in ci_install_dependency.sh.

scripts/ci/cuda/ci_install_dependency.sh
alisonshao: reverted the pyproject.toml change.

python/pyproject.toml
alisonshao: Reverted — platform configs are restored.

scripts/ci/cuda/ci_install_dependency.sh
alisonshao: Removed — this change is no longer in the PR.

hnyls2002 (approved):

hnyls2002:
The previous try has that error: https://github.com/sgl-project/sglang/actions/runs/21916574771/job/63288147727#step:5:1940

which indicates that `python-dev` may not be installed correctly. Is that flaky?

PR #18233, opened 2026-02-04
Support Qwen3 MoE context parallel

Shunkangz:
## Motivation

Context parallelism is essential in long context LLM inference. It splits a long input sequence across multiple GPUs so attention can be computed in parallel, drastically reducing latency, which enables practical million-token context windows.
In this PR, we add support for the context-parallel form of Qwen3-MoE. With this update, context parallelism can now be enabled during the prefill phase under various parallel configurations.
As for attention layer, users can use CP, TP and CP + TP.
As for moe layer, users can use CP, TP, CP + TP, CP + EP.  

## Modifications

In this implementation, we allocate a full-sequence KV cache on each CP rank. This approach simplifies both KV cache management and reuse by replicating the KV cache across all CP ranks. Before performing the attention computation, we use an allgather operation to collect the KV cache from all ranks, and then apply the FlashAttention backend for the calculation.

## Accuracy Tests

Command
`sglang serve --model-path /home/scratch.trt_llm_data/llm-models/Qwen3/Qwen3-30B-A3B-FP8/ --trust-remote-code --model-loader-extra-config '{"enable_multithread_load": true, "num_threads": 64}' --tp=4 --moe-dp-size=2 --ep-size=2 --attn-cp-size=2 --enable-prefill-context-parallel --cuda-graph-max-bs=32 --max-running-requests=32`
Results
`
Accuracy: 0.785
Invalid: 0.000
Latency: 43.704 s
Output throughput: 1027.630 token/s`

H200 Qwen3-235B
Screenshot 2026-03-10 at 16 05 11

## Checklist

## Review Process


python/sglang/srt/layers/utils/cp_utils.py
```diff
@@ -0,0 +1,400 @@
+# temp NSA debugging environ
```
ShangmingCai: This line should be removed in the final version.

python/sglang/srt/layers/communicator.py
```diff
@@ -450,11 +455,11 @@ def prepare_attn(
                 ) and hasattr(self.input_layernorm, "forward_with_allreduce_fusion"):
                     hidden_states, residual = (
                         self.input_layernorm.forward_with_allreduce_fusion(
-                            hidden_states, residual
+                            hidden_states, residual, use_attn_tp_group=True
                         )
                     )
                 else:
-                    hidden_states = tensor_model_parallel_all_reduce(hidden_states)
+                    # hidden_states = tensor_model_parallel_all_reduce(hidden_states)
```
ShangmingCai: Will removing this line affect other models?

python/sglang/srt/layers/flashinfer_comm_fusion.py
```diff
@@ -43,6 +47,7 @@ def __init__(self):
         self.max_token_num = None
         self.hidden_dim = None
         self.dtype = None
+        self.group = None
```
ShangmingCai: Has this been initialized and used?

python/sglang/srt/layers/utils/cp_utils.py
```diff
@@ -0,0 +1,400 @@
...
+        torch.split(positions, forward_batch.attn_cp_metadata.split_list, dim=-1)
+    )
+    positions = torch.cat(
+        [position_id_list[i] for i in forward_batch.attn_cp_metadata.zigzag_index],
+        dim=-1,
+    )
+    return positions
+
+
+def cp_all_gather_reorganazied_into_tensor(
```
ShangmingCai: typo? `reorganazied` -> `reorganized`

python/sglang/srt/layers/utils/cp_utils.py
```diff
@@ -0,0 +1,400 @@
...
+from typing import TYPE_CHECKING, List
+
+import torch
+import torch.nn.functional as F
+
+from sglang.srt.layers.dp_attention import get_attention_cp_group
+from sglang.srt.server_args import get_global_server_args
+
+if TYPE_CHECKING:
+    pass
```
ShangmingCai:
```suggestion
```

python/sglang/srt/layers/utils/cp_utils.py
```diff
@@ -0,0 +1,400 @@
...
+    |   | dp_atten_tp2: block2, block5 |
+    |   | dp_atten_tp3: block3, block4 |
+    |   +-------------------------+
+
+    Why zigzag rearrange?
+    - Attention calculations must follow causal attention principles.
+    - Simply slicing by rank order can lead to computational load imbalance:
+        * First rank may focus on fewer historical key-value tokens (less computation)
+        * Last rank may focus on more tokens (more computation)
+    - To mitigate uneven load, the input hissenstate needs to be sliced by cp_size*2 and rearranged.
```
ShangmingCai: hissenstate -> hidden state

python/sglang/srt/layers/communicator.py
```diff
@@ -297,15 +302,15 @@ def _compute_layer_input_mode(cls, context: _LayerModeComputationContext):
...
-            )
+            # When using specialized MoE backends (a2a or FP4), they handle token dispatch/combine
+            if (
+                not get_moe_a2a_backend().is_none()
+                or should_use_flashinfer_cutlass_moe_fp4_allgather()
+            ):
+                return ScatterMode.SCATTERED
+
+            # Without CP, use FULL mode (gather across all DP groups)
+            return ScatterMode.FULL
```
ShangmingCai: Should we check `attn_cp_size` here? What does this comment mean? To me, the logic didn't change at all. So maybe these change are not needed.

Shunkangz: I change the logic back. I think that we do not need to check `attn_cp_size` because the context parallel is orthogonal to tp communication.

python/sglang/srt/layers/communicator.py
Shunkangz: Thank you for heads-up. This change is introduced by [PR](https://github.com/sgl-project/sglang/commit/17b0affbdf3b532f816a644194168e004cf4fe4d). We use the different ways to fix the bug when the code does not enable allreduce fusion. Let me unify them together.

ShangmingCai:
Need to fix lint.
```bash
--- a/python/sglang/srt/layers/layernorm.py
+++ b/python/sglang/srt/layers/layernorm.py
@@ -315,12 +315,11 @@ class RMSNorm(MultiPlatformOp):
         """
         if residual is not None:
             from sglang.srt.distributed import (
-                get_tensor_model_parallel_world_size,
-                tensor_model_parallel_all_reduce,
-                tensor_model_parallel_fused_allreduce_rmsnorm,
                 get_attn_tensor_model_parallel_world_size,
                 get_moe_expert_parallel_world_size,
                 get_moe_tensor_parallel_world_size,
+                tensor_model_parallel_all_reduce,
+                tensor_model_parallel_fused_allreduce_rmsnorm,
             )
             from sglang.srt.layers.flashinfer_comm_fusion import (
                 flashinfer_allreduce_residual_rmsnorm,
Error: Process completed with exit code 1.
```

python/sglang/srt/layers/utils/cp_utils.py
```diff
@@ -0,0 +1,392 @@
...
+    actual_seq_q_next = split_list[cp_size * 2 - cp_rank - 1]
+    # Flash Attention expects cache_seqlens to have shape (batch_size,), not scalar
+    kv_len_prev_tensor = torch.tensor([kv_len_prev], device="cuda", dtype=torch.int32)
+    kv_len_next_tensor = torch.tensor([kv_len_next], device="cuda", dtype=torch.int32)
+    actual_seq_q_prev_tensor = torch.tensor(
+        [actual_seq_q_prev], device="cuda", dtype=torch.int32
+    )
+    actual_seq_q_next_tensor = torch.tensor(
+        [actual_seq_q_next], device="cuda", dtype=torch.int32
+    )
```
ShangmingCai, line 441: Maybe a stupid question: should we assign a device for these params? Or could they be placed in the cpu mem side?

python/sglang/srt/server_args.py
```diff
@@ -157,6 +157,8 @@
 
 NSA_PREFILL_CP_SPLIT_CHOICES = ["in-seq-split", "round-robin-split"]
 
+PREFILL_CP_SPLIT_CHOICES = ["in-seq-split"]
```
ShangmingCai, line 163: Can these two global params be united as one?

python/sglang/srt/server_args.py
```diff
@@ -4882,6 +4897,18 @@ def add_cli_args(parser: argparse.ArgumentParser):
...
+            action="store_true",
+            help="Enable context parallelism used in the prefill phase",
+        )
+        parser.add_argument(
+            "--prefill-cp-mode",
+            type=str,
+            default=ServerArgs.prefill_cp_mode,
+            choices=PREFILL_CP_SPLIT_CHOICES,
+            help="Token splitting mode for the prefill phase under context parallelism. Optional values: 'in-seq-split' (default)",
+        )
```
ShangmingCai, line 5556: Maybe we can unite these params, and deprecated those two with `nsa_` label.

Shunkangz: We discussed this before. The status is that `nsa_` did some tricky constrains which are not required for general cp. So the conclusion is that we rewrite everything and leave the old code unchanged for now.

python/sglang/srt/server_args.py
Shunkangz: As I mentioned in another reply.

python/sglang/srt/layers/utils/cp_utils.py
Shunkangz: We do not use it in our case for now and we add it only for compatibility of NSA.

python/sglang/srt/server_args.py
ShangmingCai: Sure. This can be done step by step in the future when we achieve a general design.

python/sglang/test/attention/test_flashattn_backend.py
```diff
@@ -359,116 +447,116 @@ class TestUpdateDraftDecodeSetExpandMetadata(CustomTestCase):
     This is to align with the current allocation logic. It does not affect the correctness.
     """
 
-    def test_draft_decode_set_expand_metadata(self):
```
Fridge003, line 362: Why commenting out these tests?

python/sglang/test/attention/test_flashattn_backend.py
```diff
@@ -5,13 +5,11 @@
...
 from sglang.srt.layers.attention.flashattention_backend import (
     FlashAttentionBackend,
-    draft_decode_set_expand_metadata,
 )
 from sglang.srt.layers.attention.torch_native_backend import TorchNativeAttnBackend
 from sglang.srt.layers.radix_attention import RadixAttention
 from sglang.srt.mem_cache.memory_pool import MHATokenToKVPool
 from sglang.srt.model_executor.forward_batch_info import ForwardBatch, ForwardMode
-from sglang.srt.model_executor.model_runner import ServerArgs
 from sglang.test.test_utils import CustomTestCase
```
Fridge003, line 16: If this test can pass, can we move it under register folder and register it as pr test

test/registered/8-gpu-models/test_qwen3_235b.py
```diff
@@ -69,6 +70,42 @@ def test_qwen3_235b_fp8_all_variants(self):
             ),
         )
 
+    def test_qwen3_235b_fp8_cp(self):
```
Fridge003, line 73: Can you please post result of this subtest (H200/B200)

python/sglang/srt/server_args.py
```diff
@@ -5146,6 +4897,18 @@ def add_cli_args(parser: argparse.ArgumentParser):
             help="Token splitting mode for the prefill phase of DeepSeek v3.2 under context parallelism. Optional values: 'round-robin-split'(default), 'in-seq-split'  "
             "'round-robin-split' distributes tokens across ranks based on token_idx %% cp_size. It supports multi-batch prefill, fused MoE, and FP8 KV cache.",
         )
+        parser.add_argument(
+            "--enable-prefill-context-parallel",
```
Fridge003, line 5546: Maybe we can deprecate the nsa cp related arguments? (`--nsa-prefill-cp-mode` and `--enable-nsa-prefill-context-parallel`) The new arguments can be routed to the prior nsa cp logics.
Otherwise it's a bit confusing for two different sets of cp arguments to coexist.

python/sglang/srt/server_args.py
```diff
@@ -2339,6 +2174,10 @@ def _handle_context_parallelism(self):
                     self.ep_size * self.moe_dp_size == self.tp_size
                 ), "ep_size * moe_dp_size must be equal to tp_size"
 
+            assert (
```
Fridge003, line 2572: Why adding aiter assertion?

python/sglang/test/attention/test_flashattn_backend.py
```diff
@@ -328,29 +362,83 @@ def _run_attention_test(self, mode, q_len, prefix_len=0, page_size=1):
...
-        self._run_attention_test(ForwardMode.EXTEND, q_len=self.seq_len, page_size=64)
+        output = self.backend.forward_extend(q, k, v, layer, forward_batch)
+        # else:
+        #     expected_shape = (self.batch_size, self.num_heads * self.head_dim)
+        #     output = self.backend.forward_decode(q, k, v, layer, forward_batch)
+
+        output_ref = self._run_reference_forward(
+            mode, q, k, v, layer, forward_batch, expected_shape
+        )
+        print(f"output.shape: {output.shape}")
```
Fridge003: Remove all the print commands in the test file

python/sglang/srt/layers/utils/cp_utils.py
```diff
@@ -0,0 +1,392 @@
+from dataclasses import dataclass
```
Fridge003, line 1: Looks like some of the utils are migrated from `python/sglang/srt/layers/attention/nsa/utils.py`. Can we delete them in `python/sglang/srt/layers/attention/nsa/utils.py` to avoid duplication?

python/sglang/srt/server_args.py
```diff
@@ -157,6 +157,8 @@
 
 NSA_PREFILL_CP_SPLIT_CHOICES = ["in-seq-split", "round-robin-split"]
```
Fridge003, line 161: We might replace this one with `PREFILL_CP_SPLIT_CHOICES = ["in-seq-split", "round-robin-split"]`

python/sglang/srt/server_args.py
```diff
@@ -157,6 +157,8 @@
...
 
+PREFILL_CP_SPLIT_CHOICES = ["in-seq-split"]
+
```
Fridge003, line 164: Do we need to make a whitelist that records all the models supporting prefill CP?
Seems every time we support CP for a new model, its modeling logic needs to be manually changed. So we need to notify the users that not all models support CP

python/sglang/srt/layers/attention/flashattention_backend.py
```diff
@@ -747,7 +749,15 @@ def forward_extend(
     ):
         if k is not None:
             assert v is not None
-            if save_kv_cache:
+
+            # DEBUG: Print KV cache info for both CP and non-CP cases
```
Fridge003: Remove debug message

python/sglang/srt/server_args.py
```diff
@@ -2324,6 +2155,10 @@ def _handle_context_parallelism(self):
             ), "tp_size must be divisible by dp_size * attn_cp_size"
             assert self.pp_size == 1, "PP is not supported with context parallelism"
 
+            assert (
```
Fridge003, line 2553: Why adding aiter assertion?

python/sglang/srt/layers/attention/flashattention_backend.py
```diff
@@ -764,6 +774,38 @@ def forward_extend(
                         k,
                         k_rope,
                     )
+            # When enable context parallelism, we need to all gather the kv cache from other ranks
+            # and save the kv cache to the memory pool.
+            if is_cp_mode:
```
Fridge003, line 779: Looks like this piece of code is unrelated to concrete attention backend. Can we move it to `cp_utils.py`, so it can be reused between different attention backends.

python/sglang/srt/layers/attention/flashattention_backend.py
```diff
@@ -878,25 +921,89 @@ def forward_extend(
...
-                causal=False if use_cascade_attn else causal,
-                window_size=window_size,
-                softcap=layer.logit_cap,
-                k_descale=k_descale,
-                v_descale=v_descale,
-                return_softmax_lse=use_cascade_attn,
-                num_splits=self.num_splits,
-                **kwargs,
-            )
+            if (
```
Fridge003, line 898: Is it possible to wrap the cp branch (q_prev, q_next operations, chunking logics, blabla) into a function (or maybe context manager?) Maybe we can reuse these logics in other attention backends

python/sglang/srt/models/qwen3_moe.py
```diff
@@ -303,13 +311,9 @@ def forward_normal(
...
-        if (
-            self.tp_size > 1
-            and not should_allreduce_fusion
-            and not use_reduce_scatter
-            and not should_use_flashinfer_cutlass_moe_fp4_allgather()
-        ):
-            final_hidden_states = tensor_model_parallel_all_reduce(final_hidden_states)
+
+        if self.ep_size > 1 and not should_allreduce_fusion:
+            final_hidden_states = moe_expert_parallel_all_reduce(final_hidden_states)
```
Fridge003: Why removing the conditions:
```
  and not use_reduce_scatter
  and not should_use_flashinfer_cutlass_moe_fp4_allgather()
```

python/sglang/srt/managers/schedule_policy.py
```diff
@@ -740,6 +742,9 @@ def add_one_req(
         if self.nsa_prefill_cp_in_seq_split and len(self.can_run_list) >= 1:
             return AddReqResult.OTHER
 
+        if self.prefill_context_parallel_enabled and len(self.can_run_list) >= 1:
```
Fridge003: Is the current CP not runnable on bs >= 1？

python/sglang/srt/managers/schedule_policy.py
```diff
@@ -740,6 +742,9 @@ def add_one_req(
         if self.nsa_prefill_cp_in_seq_split and len(self.can_run_list) >= 1:
```
Fridge003: Can be combined with line 745

test/registered/8-gpu-models/test_qwen3_235b.py
```diff
@@ -18,9 +18,10 @@
 class TestQwen3235BFP8(unittest.TestCase):
```
Fridge003, line 18: Can we add a small per pr test with smaller model, like Qwen3-30B-A3B?

python/sglang/srt/layers/communicator.py
```diff
@@ -452,11 +456,11 @@ def prepare_attn(
                 ) and hasattr(self.input_layernorm, "forward_with_allreduce_fusion"):
                     hidden_states, residual = (
                         self.input_layernorm.forward_with_allreduce_fusion(
-                            hidden_states, residual
+                            hidden_states, residual, use_attn_tp_group=True
                         )
                     )
                 else:
-                    hidden_states = tensor_model_parallel_all_reduce(hidden_states)
+                    hidden_states = moe_tensor_model_parallel_all_reduce(hidden_states)
```
Fridge003, line 467: Why changing to `moe_tensor_model_parallel_all_reduce` ?

Fridge003 (changes requested):

python/sglang/srt/server_args.py
Shunkangz: Agreed. As we talked before, let's deprecate these in the following PR.

python/sglang/srt/server_args.py
Shunkangz: Due to the change of FlashInfer API, it cannot construct the fused allreduce based on sub comm group. This will have an impact on performance. So I will fix this after the PR merged.

python/sglang/srt/layers/utils/cp_utils.py
Shunkangz: Let's refactor this in the next PR.

python/sglang/srt/server_args.py
Shunkangz: In our implementation, we only support `in-seq-split`, because it is more general. The `round-robin-split` will only have performance benefit on sparse attention. But I think that we can refactor it with nas related utils together.

python/sglang/srt/server_args.py
Shunkangz: No problem. I prefer to add this in the next PR.

python/sglang/srt/server_args.py
Shunkangz: As I mentioned early, the AMD team add a new fusion optimization where they do not use flashInfer. In this case, the context parallel does not support it.

python/sglang/srt/models/qwen3_moe.py
Shunkangz: Due to the logic change by the previous merged PR, the allreduce is called in communicator.

python/sglang/srt/managers/schedule_policy.py
Shunkangz: To be honest, we do not test it. Because the priority of multi-batch support is lower than performance optimization.

test/registered/8-gpu-models/test_qwen3_235b.py
Shunkangz: So you want to keep the 30B and 235B test at the same time. Is it correct?

python/sglang/srt/layers/communicator.py
Shunkangz: In the communicator, we target communication for tensor parallel. This is the collective communication for pre-attention, the allreduce should reduce the results of MoE output.

test/registered/8-gpu-models/test_qwen3_235b.py
Shunkangz: Okey. I will also run the 235B model.

python/sglang/test/attention/test_flashattn_backend.py
Shunkangz: Let me add them back.

python/sglang/srt/layers/utils/cp_utils.py
Fridge003: @Shunkangz Please update the refactor task in tracking issue

test/registered/4-gpu-models/test_qwen3_30b.py
```diff
@@ -0,0 +1,65 @@
+import unittest
+
+from sglang.test.accuracy_test_runner import AccuracyTestParams
+from sglang.test.ci.ci_register import register_cuda_ci
+from sglang.test.performance_test_runner import PerformanceTestParams
+from sglang.test.run_combined_tests import run_combined_tests
+from sglang.test.test_utils import ModelLaunchSettings
+
+# Runs on 8-GPU systems via nightly suite
+register_cuda_ci(est_time=1200, suite="nightly-8-gpu-common", nightly=True)
```
Fridge003: We can move this test to pr test:
```
register_cuda_ci(est_time=xxx, suite="stage-c-test-4-gpu-h100")
```

and rewrite it in the style of pr test (something like e2e gsm8k test, should be much shorter than this one)

test/registered/4-gpu-models/test_qwen3_30b.py
```diff
@@ -0,0 +1,76 @@
...
+from sglang.test.run_eval import run_eval
+from sglang.test.test_utils import (
+    DEFAULT_TIMEOUT_FOR_SERVER_LAUNCH,
+    DEFAULT_URL_FOR_TEST,
+    CustomTestCase,
+    kill_process_tree,
+    popen_launch_server,
+)
+
+register_cuda_ci(est_time=1000, suite="stage-c-test-4-gpu-h100")
```
Fridge003:
```suggestion
register_cuda_ci(est_time=300, suite="stage-c-test-4-gpu-h100")
```

Fridge003 (approved):

PR #18225, opened 2026-02-04
fix: add SGLANG_IS_IN_CI env var to release-docs workflow

zwang86:
## Summary
- Add `SGLANG_IS_IN_CI: true` environment variable to `release-docs.yml` workflow
- This fixes the speculative decoding notebook CI failure where `_derive_context_length()` raised a ValueError

## Root Cause
The documentation CI was failing because `is_in_ci()` returned `False` (since `SGLANG_IS_IN_CI` was not set), causing the context length validation to fail when the draft model's context_length (2048) was less than the target model's context_length (8192).

Other CI workflows (`pr-test.yml`, `nightly-test-nvidia.yml`, etc.) already set this environment variable. This fix aligns `release-docs.yml` with them.

Using git blame on [python/sglang/srt/configs/model_config.py](https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/configs/model_config.py) (lines 361-363):
```
eb19ccadae (Liangsheng Yin 2025-08-21) or is_in_ci()  # FIXME: fix this special case
```
The is_in_ci() check was introduced in PR #9388 (eb19ccadae) on 2025-08-21, titled [bug] fix errors related to context length in SD. This PR added the special case handling for CI environments, but the release-docs.yml workflow was not updated to set the required environment variable.

## Test Plan
- [ ] Documentation CI should pass after this change

Fixes #18220

Issue #18220, opened 2026-02-04
[Bug] Speculative Decoding Document failed.


zhaochenyang20:
### Checklist

- [x] If this is not a feature request but a general question, please start a discussion at https://github.com/sgl-project/sglang/discussions. Otherwise, it will be closed.
- [x] Please use English. Otherwise, it will be closed.

### Motivation

https://github.com/sgl-project/sglang/actions/runs/21657515497/job/62435300224#step:5:1360

Documentation CI has failed. Need to debug which PR introduced this bug and fix this CI. Basically speaking, just running notebook `docs/advanced_features/speculative_decoding.ipynb` and you can see the bug. I think just a minimal bug fix is needed.

Please also explain how you fix this error and which PR introduced this. May use git blame.

### Related resources

Detailed error as:

```bash
[2026-02-04 03:54:42] Scheduler hit an exception: Traceback (most recent call last):
  File "/actions-runner/_work/sglang/sglang/python/sglang/srt/managers/scheduler.py", line 3049, in run_scheduler_process
    scheduler = Scheduler(
  File "/actions-runner/_work/sglang/sglang/python/sglang/srt/managers/scheduler.py", line 350, in __init__
    self.init_model_worker()
  File "/actions-runner/_work/sglang/sglang/python/sglang/srt/managers/scheduler.py", line 540, in init_model_worker
    self.maybe_init_draft_worker()
  File "/actions-runner/_work/sglang/sglang/python/sglang/srt/managers/scheduler.py", line 536, in maybe_init_draft_worker
    self.draft_worker = DraftWorkerClass(**draft_worker_kwargs)
  File "/actions-runner/_work/sglang/sglang/python/sglang/srt/speculative/eagle_worker.py", line 140, in __init__
    super().__init__(
  File "/actions-runner/_work/sglang/sglang/python/sglang/srt/managers/tp_worker.py", line 241, in __init__
    self._init_model_config()
  File "/actions-runner/_work/sglang/sglang/python/sglang/srt/managers/tp_worker.py", line 307, in _init_model_config
    self.model_config = ModelConfig.from_server_args(
  File "/actions-runner/_work/sglang/sglang/python/sglang/srt/configs/model_config.py", line 249, in from_server_args
    return ModelConfig(
  File "/actions-runner/_work/sglang/sglang/python/sglang/srt/configs/model_config.py", line 198, in __init__
    self._derive_context_length(context_length)
[...]

PR #18234, opened 2026-02-04
fix: allow draft models to override context length

JackZeng0208:
## Motivation
See: https://github.com/sgl-project/sglang/issues/18220
Purpose: allow draft models to override context length  

The speculative decoding documentation CI was failing because draft models with smaller context lengths (ex: EAGLE with 2048) cannot use target model's context length (ex: like 8192) without setting environment variables.

## Modifications

This fix adds `is_draft_model` to the bypass condition in `_derive_context_length()` in `sglang/srt/configs/model_config.py`, allowing draft models to automatically override context length. The fix will solve the problem fundamentally (not just for the doc itself). This fix is safe because code already handles draft models safely right after that(https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/configs/model_config.py#L367)

Comparing with PR #18225 (PR #18226) and PR #18228
- For PR #18225: changing `SGLANG_IS_IN_CI` will affect other codebase significantly
- For PR #18228: only works for doc using `doc_patch.py`
- Basically, both PRs are trying to bypass validation

## Checklist

## Review Process


mmangkad:
Hi @JackZeng0208, `is_draft_model` shouldn’t be made an unconditional bypass. PR #9388 intentionally made this opt‑in because forcing longer context on shorter drafts can trigger IMA. #10787 only makes the override safe _when explicitly enabled_ (it bumps `max_position_embeddings`), it doesn’t justify always‑on behavior. This effectively undoes #9388’s safety guard for draft models.

Concrete example: target `meta-llama/Meta-Llama-3-8B-Instruct` (8192) + draft `lmsys/sglang-EAGLE-LLaMA3-Instruct-8B` (2048). If we always bypass, the draft gets forced to 8192 and its rope cache/attention metadata are still sized for 2048, which is exactly how you hit device‑side asserts (scatter/gather index out of bounds) as described in #9388.

For #18220 the right fix is docs‑scoped opt‑in (`SGLANG_ALLOW_OVERWRITE_LONGER_CONTEXT_LEN=1` during docs runs) so we keep the global guard intact.

PR #18178, opened 2026-02-03
Enhance Kimi K2.5 support with improved reasoning parser and function call detector

cynial:
## Motivation

Kimi K2.5 models exhibit special behaviors that required dedicated handling:

- Reasoning Mode Issue: When deciding to invoke a tool while in reasoning mode (`<think>...</think>`), the model may output tool call markers directly without first outputting the `</think>` closing tag.
- UUID Format Support: Support UUID-style call IDs (e.g., `call_6b0e7c24f0c44f0d996b59e5`). This ensures compatibility for scenarios that mix models (e.g., using Model A in the first round and Model Kimi-K2.5 in the second round).
- Streaming Robustness: The original streaming parser had issues with special token leakage and incomplete state management.

Related Issues: https://github.com/sgl-project/sglang/issues/18086

### Reasoning Parser
1. New KimiK2Detector Class (inherits from Qwen3Detector)
Force exit reasoning mode, return content before marker as reasoning_text, content from marker onwards as normal_text

### Function Call Detector 
1. UUID Format Support
Extended regex patterns to support both standard and UUID formats
3. Function Name Inference
New `_infer_function_name_from_args()` method for UUID format
4. Robust State Management
Added proper state tracking and reset methods
5. Special Token Leakage Prevention
Comprehensive token filtering to prevent special tokens from leaking to normal_text
6. Improved Streaming State Machine
Refactored `parse_streaming_increment()` with clearer state transitions
7. Hyphenated Function Name Support
Updated regex to support hyphenated function names (common in MCP tools)
Example: `mcp__portal__search-documents`

## Checklist

- [ ] Format your code according to the [Format code with pre-commit](https://docs.sglang.io/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.io/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.io/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.io/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.io/developer_guide/contribution_guide.html#benchmark-the-speed).
- [ ] Follow the SGLang code style [guidance](https://docs.sglang.io/developer_guide/contribution_guide.html#code-style-guidance).

## Review Process

[...]

PR #18246, opened 2026-02-04
[Grammar] Add strict reasoning grammar for think_end token missing (#18086)

eraser00:
## Motivation

As a bug reported in #18086 , some thinking models like kimi-k2-thinking or QWen3-thinking might generate an incomplete reasoning content even under the enforced thinking mode.

A reasoning output like:
```
"message": {
  "role": "assistant",
  "content": "",
  "reasoning_content": " The search results did not directly return the DOI information for this paper. Let me search more sp
ecifically to verify if the DOI provided by the user is correct. <|tool_calls_section_begin|> <|tool_call_begin|> functions.search:1 <|tool_call_argument_begin|> {\"queries\": [\"10.11897/SP.J.1016.2025.01313\", \"Chinese Journal of Computers DOI 10.11897 SP.J.1016\", \"A Survey of UAV Semantic Security Research DOI 10.11897\"]} <|tool_call_end|> <|tool_calls_section_end|>",
  "tool_calls": null
}
```
breaks both the reasoning parser and the tool call parser, and interrupts the multi-round agent loop frequently.

We believe the root cause of these problems is the totally grammar-free reasoning processing. The reasoning parser mismatch with the implementations of the reasoning grammar.

We try to handle all these mess-up issues by introducing a `StrictReasonerGrammar` to slightly constrain the reasoning content format.

## Modifications

We add a `StrictReasonerGrammar` to enable special tokens filtering during the reasoning content generate.

By carefully configuring the special token filter, both the reasoning content format and the tool call format can be achieved with 100% correctness.

Unlike existing solutions (for example: [PR-18199](https://github.com/sgl-project/sglang/pull/18199) and [PR-18178](https://github.com/sgl-project/sglang/pull/18178) ), we choose to address this problem at the Reasoning Grammar level rather than modifying the tool call parser one by one.

Unit testing and accuracy benchmarking results are underway; We just want to get useful feedback ASAP, so we post this initial PR.

## Checklist

## Review Process


eraser00:
@yhyang201 
Could you please help to run the CI on this PR?

eraser00:
> . The problem doesn't appear to be completely fixed; rather, it occurs at a lower frequency compared to before.

Any bad cases examples？I think the </think> token missing problems should be fixed.

dongfangzan:
> > . The problem doesn't appear to be completely fixed; rather, it occurs at a lower frequency compared to before.
> 
> Any bad cases examples？I think the token missing problem should be fixed when using the new reasoning parser `kimi_k2-strict` . If that is not the case, there might be some unknown bugs in this initial patch.

@eraser00 My bad — looks like I was routed to the wrong model. Should be good now.

PR #18250, opened 2026-02-04
Fix long prompt KV allocation by falling back to torch native APIs when exceeding Triton tensor limit

ch-wan:
## Motivation

When processing very long prompts, the number of extend tokens can exceed `TRITON_MAX_TENSOR_NUMEL`, causing the Triton-based `alloc_extend_kernel` to fail. This limits the maximum prompt length that SGLang can handle with paged KV cache allocation.

## Modifications

- **Added `alloc_extend_naive` function** (`python/sglang/srt/mem_cache/allocator.py`): A pure PyTorch fallback implementation for KV cache allocation that handles the same logic as `alloc_extend_kernel` but without Triton tensor size constraints.
- **Added conditional dispatch in `alloc_extend`**: When `extend_num_tokens < tl.core.TRITON_MAX_TENSOR_NUMEL`, the existing Triton kernel is used. Otherwise, the new `alloc_extend_naive` fallback is invoked.
- **Clamped `seen_max_num_extend_tokens_next_power_of_2`**: The tracked maximum is now capped at `TRITON_MAX_TENSOR_NUMEL` to avoid allocating oversized Triton buffers.

## Accuracy Tests

N/A — This change does not modify model forward logic or kernel computation. The fallback produces the same KV index mappings as the Triton kernel.

## Benchmarking and Profiling

N/A — The fallback path is only triggered for very long prompts that exceed Triton limits. For normal-length prompts, the existing Triton kernel continues to be used with no performance impact.

## Checklist

## Review Process

### DIFF
--- a/python/sglang/srt/mem_cache/allocator.py
+++ b/python/sglang/srt/mem_cache/allocator.py
@@ -423,16 +423,28 @@ class PagedTokenToKVPoolAllocator(BaseTokenToKVPoolAllocator):
         out_indices = torch.empty(
             (extend_num_tokens,), dtype=torch.int64, device=self.device
         )
-        alloc_extend_kernel[(bs,)](
-            prefix_lens,
-            seq_lens,
-            last_loc,
-            self.free_pages,
-            out_indices,
-            next_power_of_2(bs),
-            self.page_size,
-            self.seen_max_num_extend_tokens_next_power_of_2,
-        )
+
+        if extend_num_tokens < tl.core.TRITON_MAX_TENSOR_NUMEL:
+            alloc_extend_kernel[(bs,)](
+                prefix_lens,
+                seq_lens,
+                last_loc,
+                self.free_pages,
+                out_indices,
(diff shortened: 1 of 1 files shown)


python/sglang/srt/mem_cache/allocator.py
```diff
@@ -423,16 +423,28 @@ def alloc_extend(
...
         )
-        alloc_extend_kernel[(bs,)](
-            prefix_lens,
-            seq_lens,
-            last_loc,
-            self.free_pages,
-            out_indices,
-            next_power_of_2(bs),
-            self.page_size,
-            self.seen_max_num_extend_tokens_next_power_of_2,
```
fzyzcjy, line 434: FYI: iirc the seen_max_num_extend_tokens_next_power_of_2 is used to avoid many recompiles (do not remember clearly to be honest) 

(p.s. ping me on slack if not reply on github - I get hundreds of messages in github inbox)

PR #18274, opened 2026-02-05
Fix test_return_routed_experts to use response-level sglext

alisonshao:
## Summary
- Fix `test_return_routed_experts.py` to access `sglext` at response level instead of `sgl_ext` in choices

PR #17648 moved the `sgl_ext` field from `choices[0]` to the response level and renamed it to `sglext` (no underscore). The test was not updated accordingly, causing CI failures.

## Test plan
- [ ] CI passes for `stage-c-test-large-4-gpu` suite

🔗 [Failure example](https://github.com/sgl-project/sglang/actions/runs/21653054987/job/62421861711)


ishandhanani:
Apologies @Kangyan-Zhou and @alisonshao - I should have caught this. Thank you for the fix

PR #18306, opened 2026-02-05
[Feature] Implement update_weights_from_disk for SGLang-D (Diffusion …

dreamyang-liu:
## Motivation

Implement the `update_weights_from_disk` interface for SGLang-D (diffusion engine) to enable dynamic weight updates for RL workflows and iterative model fine-tuning without restarting the server.

This feature is essential for:
- **RL training loops**: Update policy model weights after each training iteration
- **Iterative fine-tuning**: Hot-swap checkpoints during experimentation
- **A/B testing**: Switch between model versions without downtime

Mirrors the existing LLM engine's `update_weights_from_disk` functionality (see `sglang/srt/managers/scheduler_update_weights_mixin.py`).

Closes #18078

### Core Implementation (`gpu_worker.py`)
- Add `update_weights_from_disk()` method with:
  - Support for all `nn.Module` components by default (transformer, vae, text_encoder, etc.)
  - **Layerwise offload handling**: Properly disable offload before update (load weights from CPU to GPU), then re-enable after (sync new weights back to CPU)
  - **DTensor support**: Handle distributed tensor parameters for tensor parallelism using `distribute_tensor()`
  - **Atomic updates with rollback**: If any module fails to update, rollback all previously updated modules
  - **TeaCache state reset**: Clear cached states after weight updates

### Request Handling
- Add `UpdateWeightsFromDiskReq` dataclass in new `io_struct.py`
- Add scheduler handler for update_weights requests via ZMQ
- Add `/update_weights_from_disk` HTTP endpoint in `http_server.py`

### Tests (`test_update_weights_from_disk.py`)
- Basic API tests: same model reload, flush_cache options, specific modules
- Layerwise offload integration tests
- End-to-end tests: verify generation works after weight update

## Accuracy Tests

N/A - This feature only reloads weights from disk without modifying model forward logic. The same weights produce identical outputs.

Tested the images generated before weight update and after the weight update are identical. 

Tested with
- Qwen/Qwen-Image
- Wan-AI/Wan2.1-T2V-1.3B-Diffusers
- black-forest-labs/FLUX.2-klein-4B
- zai-org/GLM-Image (weight update succeed, the image generation is having some issue on my machine, but it's not related to this change, other models tested that update weight can start inference and produce same result)

on 8 * A100 80G machine.

Logging output example:
```
Updating 4 modules: text_encoder <- /sgl-workspace/sglang/glm-image/text_encoder, vae <- /sgl-workspace/sglang/glm-image/vae, vision_language_encoder <- /sgl-workspace/sglang/glm-image/vision_language_encoder, transformer <- /sgl-workspace/sglang/glm-image/transformer
```
This log shows which modules are being updated and their corresponding weight directories, making it easy to verify the weight loading source for each module.

Example request
```
curl -X POST http://localhost:30010/update_weights_from_disk \
  -d '{"model_path": "/sgl-workspace/sglang/glm-image"}'
```

## Benchmarking and Profiling

N/A - This is an operational API for weight management, not a performance-critical path. Weight loading time depends on model size and storage I/O.

## Checklist


python/sglang/multimodal_gen/test/server/test_update_weights_from_disk.py
zhaochenyang20, line 1: I strongly think that not all the tests will be verified in our per-commit CI. Please find out how to add this test in the CI run suite, since this is an important PR.

zhaochenyang20 (changes requested):

zhaochenyang20 (changes requested):
One critical thing I wanna mention is that, do not copy yourself. I think a lot of code for `update_weights_from_disk` in LLM can be reused in Diffusion. Share these functions and do not copy these lines, which introduce quite a difficulty for maintenance.

If we can not reuse it, please also explain why. And, provide your running results of your tests, with the speed.

dreamyang-liu:
> One critical thing I wanna mention is that, do not copy yourself. I think a lot of code for `update_weights_from_disk` in LLM can be reused in Diffusion. Share these functions and do not copy these lines, which introduce quite a difficulty for maintenance.
> 
> If we can not reuse it, please also explain why. And, provide your running results of your tests, with the speed.

Thanks for the comments and suggestions! I agree we don't want to reinvent the wheel, but following reasons prevent us from reusing more from LLM engine's implementation.

## 1. Missing `load_weights` Interface in DiT Models

- LLM's `load_weights_and_postprocess` requires models to implement `load_weights()` method
- DiT models (Flux, HunyuanVideo, etc.) don't have this interface - they use `param_names_mapping` instead

```python
# LLM: sglang/srt/model_loader/loader.py:683-684
def load_weights_and_postprocess(model, weights, target_device):
    model.load_weights(weights)  # Requires load_weights() method

# Diffusion DiT models have no load_weights() - check:
# sglang/multimodal_gen/runtime/models/dits/flux.py
# sglang/multimodal_gen/runtime/models/dits/hunyuanvideo.py
```
Unless we implement this interface in all diffusion models, it's hard to reuse the DefaultModelLoader

## 2. Different Distributed Mechanisms

- LLM: Manual sharding via `narrow()` in `weight_loader`, no DTensor
- Diffusion: FSDP2/DTensor with automatic `distribute_tensor()`

```python
# LLM: sglang/srt/layers/linear.py:370-417
def weight_loader(self, param, loaded_weight):
    shard_size = param_data.shape[output_dim]
    start_idx = self.tp_rank * shard_size
    loaded_weight = loaded_weight.narrow(output_dim, start_idx, shard_size)
    param_data.copy_(loaded_weight)

# Diffusion: sglang/multimodal_gen/runtime/loader/fsdp_load.py:264-270
if hasattr(meta_sharded_param, "device_mesh"):
    sharded_tensor = distribute_tensor(
        full_tensor, meta_sharded_param.device_mesh, meta_sharded_param.placements
    )
```

## 3. LLM Weight Loader Does Not Handle DTensor

- LLM's `weight_loader` assumes regular `torch.Tensor`, no DTensor checks
- Diffusion explicitly handles DTensor in weight loading

```python
# Diffusion: sglang/multimodal_gen/runtime/managers/gpu_worker.py:408-415
if DTensor is not None and isinstance(param, DTensor):
    distributed_weight = distribute_tensor(
        loaded_weight.to(param.device, param.dtype),
        param.device_mesh, param.placements,
    )
    param._local_tensor.copy_(distributed_weight._local_tensor)
```

## 4. Different Weight Mapping Strategies

- LLM: Complex `stacked_params_mapping` for qkv merging, quantization
- Diffusion: Simple `param_names_mapping` for name translation

```python
# LLM: sglang/srt/models/qwen2.py:556-561
stacked_params_mapping = [
    ("qkv_proj", "q_proj", "q"),
    ("qkv_proj", "k_proj", "k"),
    ("qkv_proj", "v_proj", "v"),
]

# Diffusion: uses model.param_names_mapping attribute
# sglang/multimodal_gen/runtime/loader/fsdp_load.py:118
param_names_mapping_fn = get_param_names_mapping(model.param_names_mapping)
```

## 5. Different Model Structures

- LLM: Single `self.model`
- Diffusion: Multi-module pipeline (transformer, vae, text_encoder, etc.)

```python
# LLM: sglang/srt/model_executor/model_runner.py:1085
model = model_load_weights(self.model, iter)

# Diffusion: sglang/multimodal_gen/runtime/managers/gpu_worker.py:379-385
for name in module_names:
    module = self.pipeline.get_module(name)  # Multiple modules
    if module is not None and isinstance(module, torch.nn.Module):
        modules_to_update.append((name, module))
```

Btw, I do agree if you think this function too long, I can try to separate if you think that would make it much easier for maintain. Please let me know your thoughts.

dreamyang-liu:
## Test 

### Unit Test
Test with single A100 80G with model `black-forest-labs/FLUX.2-klein-4B`
```
============================= test session starts ==============================
platform linux -- Python 3.12.3, pytest-9.0.2, pluggy-1.6.0 -- /usr/bin/python
cachedir: .pytest_cache
rootdir: /sgl-workspace/sglang/python
configfile: pyproject.toml
plugins: anyio-4.12.1, typeguard-4.4.4
collecting ... collected 9 items

python/sglang/multimodal_gen/test/server/test_update_weights_from_disk.py::TestUpdateWeightsFromDisk::test_get_model_info PASSED [ 11%]
python/sglang/multimodal_gen/test/server/test_update_weights_from_disk.py::TestUpdateWeightsFromDisk::test_update_weights_same_model PASSED [ 22%]
python/sglang/multimodal_gen/test/server/test_update_weights_from_disk.py::TestUpdateWeightsFromDisk::test_update_weights_with_flush_cache PASSED [ 33%]
python/sglang/multimodal_gen/test/server/test_update_weights_from_disk.py::TestUpdateWeightsFromDisk::test_update_weights_without_flush_cache PASSED [ 44%]
python/sglang/multimodal_gen/test/server/test_update_weights_from_disk.py::TestUpdateWeightsFromDisk::test_update_weights_nonexistent_model PASSED [ 55%]
python/sglang/multimodal_gen/test/server/test_update_weights_from_disk.py::TestUpdateWeightsFromDisk::test_update_weights_missing_model_path PASSED [ 66%]
python/sglang/multimodal_gen/test/server/test_update_weights_from_disk.py::TestUpdateWeightsFromDisk::test_update_weights_specific_modules PASSED [ 77%]
python/sglang/multimodal_gen/test/server/test_update_weights_from_disk.py::TestUpdateWeightsFromDiskWithOffload::test_update_weights_with_offload_enabled PASSED [ 88%]
python/sglang/multimodal_gen/test/server/test_update_weights_from_disk.py::TestUpdateWeightsEndToEnd::test_generation_after_weight_update PASSED [100%]

=============================== warnings summary ===============================
<frozen importlib._bootstrap>:488
  <frozen importlib._bootstrap>:488: DeprecationWarning: builtin type SwigPyPacked has no __module__ attribute

<frozen importlib._bootstrap>:488
  <frozen importlib._bootstrap>:488: DeprecationWarning: builtin type SwigPyObject has no __module__ attribute

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
================== 9 passed, 2 warnings in 187.21s (0:03:07) ===================
```

### Speed Test:
```
============================================================
>>> update_weights_from_disk took 3.65 seconds <<< (black-forest-labs/FLUX.2-klein-4B)
============================================================

============================================================
>>> update_weights_from_disk took 12.45 seconds <<< (Qwen/Qwen-Image)
============================================================
```

python/sglang/multimodal_gen/runtime/entrypoints/http_server.py
```diff
@@ -92,6 +92,90 @@ async def health_generate():
...
+async def get_model_info(request: Request):
+    """Get information about the current model."""
+    server_args: ServerArgs = request.app.state.server_args
+    return {
+        "model_path": server_args.model_path,
+    }
+
+
+# Weight update router for RL workflows
+weight_update_router = APIRouter()
```
zhaochenyang20: I am just pondering what `APIRouter()` is used for. Could the `health_router` be shared? Should we have a duplicate router here?

python/sglang/multimodal_gen/runtime/entrypoints/http_server.py
```diff
@@ -92,6 +92,90 @@ async def health_generate():
...
+        - load_format (str, optional): Format of the weights to load (default: "auto")
+        - flush_cache (bool, optional): Whether to flush cache after update (default: True)
+        - target_modules (list[str], optional): List of module names to update.
+            Default: updates ALL nn.Module components (transformer, vae, text_encoder, etc.)
+            Examples: ["transformer"] to update only transformer
+
+    Returns:
+        - success (bool): Whether the update was successful
+        - message (str): Status message
+    """
```
zhaochenyang20: I think this doc string is not dedicated enough 😂 Could you change this documentation according to:

https://docs.sglang.io/advanced_features/sglang_for_rl.html#update-weights-from-disk

python/sglang/multimodal_gen/runtime/entrypoints/http_server.py
```diff
@@ -92,6 +92,90 @@ async def health_generate():
...
+    from sglang.multimodal_gen.runtime.managers.io_struct import (
+        UpdateWeightsFromDiskReq,
+    )
```
zhaochenyang20: import it at the beggining of files.

python/sglang/multimodal_gen/runtime/entrypoints/http_server.py
```diff
@@ -92,6 +92,90 @@ async def health_generate():
...
+            return ORJSONResponse(
+                {"success": False, "message": "Unknown response format"},
+                status_code=500,
+            )
+
+    except Exception as e:
+        return ORJSONResponse(
+            {"success": False, "message": f"Error: {str(e)}"},
+            status_code=500,
+        )
```
zhaochenyang20: 1. Have a large try except block is a bad coding pattern. Please make sure your error catch range is small.

python/sglang/multimodal_gen/runtime/entrypoints/http_server.py
```diff
@@ -92,6 +92,90 @@ async def health_generate():
...
+            model_path=model_path,
+            load_format=body.get("load_format", "auto"),
+            flush_cache=body.get("flush_cache", True),
+            target_modules=body.get("target_modules"),
+        )
+
+        response = await async_scheduler_client.forward(req)
+
+        # Handle response
+        if hasattr(response, "output") and response.output:
```
zhaochenyang20: 2. Strongly do not use `hasattr` and `getattr`. Which makes the code messy.

python/sglang/multimodal_gen/runtime/entrypoints/http_server.py
```diff
@@ -92,6 +92,90 @@ async def health_generate():
     return {"status": "ok"}
 
 
+@health_router.get("/get_model_info")
+async def get_model_info(request: Request):
+    """Get information about the current model."""
+    server_args: ServerArgs = request.app.state.server_args
+    return {
+        "model_path": server_args.model_path,
+    }
```
zhaochenyang20: Even though this code change is small. But, please detach this API in a separate PR. Make small PRs for easy code review.

python/sglang/multimodal_gen/runtime/entrypoints/http_server.py
```diff
@@ -222,6 +306,7 @@ def create_app(server_args: ServerArgs):
     app = FastAPI(lifespan=lifespan)
 
     app.include_router(health_router)
+    app.include_router(weight_update_router)
     app.include_router(vertex_router)
```
zhaochenyang20, line 211: well. Seems alreay has so many routers. Just pondering why the router can not be shared? 😂

python/sglang/multimodal_gen/runtime/managers/gpu_worker.py
```diff
@@ -296,6 +296,319 @@ def list_loras(self) -> OutputBatch:
...
+    }
+
+    # Default modules to update for RL workflows (typically only transformer is trained)
+    _DEFAULT_TARGET_MODULES = [
+        "transformer",
+        "transformer_2",
+        "video_dit",
+        "video_dit_2",
+        "audio_dit",
+    ]
```
zhaochenyang20: 1. Are you sure these hyperparameters are right as new models come out? Could we read them from the model config?
2. Make Hyperparameters at the beginning of the files.

python/sglang/multimodal_gen/runtime/managers/gpu_worker.py
```diff
@@ -296,6 +296,319 @@ def list_loras(self) -> OutputBatch:
...
+            model_path: Path to the new model weights (HuggingFace model path or local directory).
+            load_format: Format of the weights to load (default: "auto").
+            flush_cache: Whether to reset cache state after updating weights (default: True).
+            target_modules: List of module names to update. If None or ["all"], updates all
+                           nn.Module components. Specify a list like ["transformer"] to update
+                           only specific modules.
+
+        Returns:
+            Tuple of (success: bool, message: str).
+        """
```
zhaochenyang20: change these comments as consistent as https://github.com/sgl-project/sglang/pull/18306/changes#r2772203606

python/sglang/multimodal_gen/runtime/managers/gpu_worker.py
```diff
@@ -296,6 +296,319 @@ def list_loras(self) -> OutputBatch:
...
+        import gc
+        import os
+
+        from sglang.multimodal_gen.runtime.loader.utils import _list_safetensors_files
+        from sglang.multimodal_gen.runtime.loader.weight_utils import (
+            safetensors_weights_iterator,
+        )
+        from sglang.multimodal_gen.runtime.utils.hf_diffusers_utils import (
+            maybe_download_model,
+        )
```
zhaochenyang20: never import inside function

python/sglang/multimodal_gen/runtime/managers/gpu_worker.py
```diff
@@ -296,6 +296,319 @@ def list_loras(self) -> OutputBatch:
...
+            module = self.pipeline.get_module(name)
+            if module is not None and isinstance(module, torch.nn.Module):
+                modules_to_update.append((name, module))
+
+        # For DiffusersPipeline, also check diffusers_pipe attributes
+        diffusers_pipe = self.pipeline.get_module("diffusers_pipeline")
+        if diffusers_pipe is not None and not modules_to_update:
+            for name in module_names:
+                if hasattr(diffusers_pipe, name):
+                    module = getattr(diffusers_pipe, name)
```
zhaochenyang20: do not use `hasattr` and `getattr` as possible.

python/sglang/multimodal_gen/runtime/managers/gpu_worker.py
```diff
@@ -296,6 +296,319 @@ def list_loras(self) -> OutputBatch:
...
+
+        # Log which modules will be updated from which directories
+        logger.info(
+            f"Updating {len(module_weights_map)} modules: "
+            + ", ".join(
+                f"{name} <- {path}" for name, path in module_weights_map.items()
+            )
+        )
+
+        # Phase 2: Update all modules
```
zhaochenyang20: these `Phase` seems to be too vibe coding 😂

zhaochenyang20 (changes requested):
Try to clean up the code and make it compact. thanks

python/sglang/multimodal_gen/runtime/entrypoints/http_server.py
dreamyang-liu: Yeah, I also kinda feel this can be shared, I just merged it. It's a two-way door decision and can be changed easily.

dreamyang-liu:
> Try to clean up the code and make it compact. thanks

@zhaochenyang20 Thanks for the suggestions! I just made some new some refactor and clean up, please check the revision when you get time. Thanks!

docs/advanced_features/sglang_for_rl.md
```diff
@@ -106,6 +106,23 @@ This path trades some I/O overhead for simplicity and flexibility. It integrates
 
 **Python Engine API:** `engine.update_weights_from_disk(model_path, load_format=None)`
 
+**Diffusion engine (SGLang-D):** The diffusion engine exposes the same `POST /update_weights_from_disk` endpoint. The update is all-or-nothing with automatic rollback on failure. When layerwise offload (`--dit-layerwise-offload`) is enabled, offload is temporarily disabled during the update, which causes a temporary GPU memory peak before returning to normal offloaded memory usage.
```
zhaochenyang20: > The update is all-or-nothing with automatic rollback on failure. When layerwise offload (`--dit-layerwise-offload`) is enabled, offload is temporarily disabled during the update, which causes a temporary GPU memory peak before returning to normal offloaded memory usage.


What does this mean?

python/sglang/multimodal_gen/runtime/entrypoints/http_server.py
zhaochenyang20: I think a better way is to discuss this with the developer team. cc @mickqian why there are so many routers that seems can be shared?

docs/advanced_features/sglang_for_rl.md
zhaochenyang20: Make this part detailed rather detailed. You can directly up these up.

* **All-or-nothing**: if any module fails to load, all previously updated
  modules are rolled back to the original weights.
* **Rollback failures propagate**: if rollback itself fails, the exception is
  *not* caught so the caller knows the model is in an inconsistent state.
  This matches the LLM engine behaviour.
* **Offload-aware**: layerwise offload is temporarily disabled during the
  update so that weight tensors are fully materialised on the target device.
  This is necessary because the offload manager replaces parameters with
  small placeholders; the real tensors must be restored before copying.
* **DTensor-aware**: parameters that have been distributed via
  ``torch.distributed.tensor`` are updated through ``distribute_tensor``
  so that each shard is correctly placed.

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
```diff
@@ -0,0 +1,335 @@
...
+* **Rollback failures propagate**: if rollback itself fails, the exception is
+  *not* caught so the caller knows the model is in an inconsistent state.
+  This matches the LLM engine behaviour.
+* **Offload-aware**: layerwise offload is temporarily disabled during the
+  update so that weight tensors are fully materialised on the target device.
+  This is necessary because the offload manager replaces parameters with
+  small placeholders; the real tensors must be restored before copying.
+* **DTensor-aware**: parameters that have been distributed via
+  ``torch.distributed.tensor`` are updated through ``distribute_tensor``
+  so that each shard is correctly placed.
```
zhaochenyang20: Nice designs

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
```diff
@@ -0,0 +1,335 @@
...
+        original_path = maybe_download_model(self._original_model_path)
+        for name in updated_modules:
+            module = self.pipeline.get_module(name)
+            if module is None:
+                continue
+            weights_dir = find_weights_dir(original_path, name)
+            if weights_dir is None:
+                continue
+            weights_iter = _get_weights_iter(weights_dir)
+            load_weights_into_model(weights_iter, dict(module.named_parameters()))
```
zhaochenyang20: `rollback` seems to be important. Do you have any idea if we can have a test for it? How to mock the update to fail? We can adds test to `rollback` in following PRs. No need for it right now.

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
```diff
@@ -0,0 +1,335 @@
...
+# Module-level utility functions
+# ---------------------------------------------------------------------------
+
+
+def find_weights_dir(local_path: str, module_name: str) -> str | None:
+    """Locate the safetensors directory for *module_name* under *local_path*.
+
+    Tries ``<local_path>/<module_name>/`` first, then falls back to
+    *local_path* itself if it directly contains safetensors files (common
+    for RL checkpoints that save weights in a flat directory).
```
zhaochenyang20: This comment is quite weird. I think in LLM's update weights from disk, we do not have any of this kind of function `find_weights_dir`. Could you share the same design pattern with LLMs?

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
```diff
@@ -0,0 +1,335 @@
...
+    dir_path = os.path.join(local_path, module_name)
+    if os.path.exists(dir_path):
+        return dir_path
+    if _list_safetensors_files(local_path):
+        return local_path
+    return None
+
+
+def _get_weights_iter(weights_dir: str):
+    """Return a ``(name, tensor)`` iterator over safetensors in *weights_dir*."""
```
zhaochenyang20: this doc string is too messy:

do not have ` or * in your doc string. Python files are not markdown.

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
```diff
@@ -0,0 +1,335 @@
...
+
+
+def _validate_weight_files(
+    local_model_path: str,
+    modules_to_update: list[tuple[str, torch.nn.Module]],
+) -> tuple[dict[str, str], list[str]]:
+    """Check that every module has a weights directory with safetensors files.
+
+    Returns:
+        ``(weights_map, missing)`` where *weights_map* maps module name to its
```
zhaochenyang20: same to all:

https://github.com/sgl-project/sglang/pull/18306/changes#r2776305670

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
```diff
@@ -0,0 +1,335 @@
...
+    """
+    disabled: list[torch.nn.Module] = []
+    for _, module in modules_to_update:
+        if (
+            isinstance(module, OffloadableDiTMixin)
+            and module.layerwise_offload_managers
+        ):
+            module.disable_offload()
+            disabled.append(module)
+    return disabled
```
zhaochenyang20: Why should we explicitly disable offload, but not just dynamicly check whether some parts are offloaded and just load it back? And, I do not why it seems that in LLMs, we do not care about wether some parts are offloaded or not?

python/sglang/multimodal_gen/runtime/managers/io_struct.py
```diff
@@ -0,0 +1,15 @@
...
+from typing import List, Optional
+
+
+@dataclass
+class UpdateWeightsFromDiskReq:
+    """Request to update model weights from disk for diffusion models."""
+
+    model_path: str
+    flush_cache: bool = True
+    target_modules: Optional[List[str]] = None
```
zhaochenyang20: These codes seems right, but I do not know why other requests do not need a `io_struct`? Could we follow what other requests do?

zhaochenyang20: In other words, I strongly think there is no need to add a new file.

zhaochenyang20 (changes requested):

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
dreamyang-liu: This is because for diffusion models, they store weights of different modules in different folders. 

See example in https://huggingface.co/zai-org/GLM-Image/tree/main, it has 
- text_encoders
- vae
- transformer
- vision_language_encoder 

Each of these folders contains the safetensor for the module. For LLM, most likely it only have one folder that contains all tensors.

python/sglang/multimodal_gen/runtime/managers/io_struct.py
dreamyang-liu: I agree that having a new file looks weird, especially only one data class here. My point would be 

1. Reusing the LLM's `UpdateWeightFromDiskReqInput` seems not very compatible given a lot features we don't support yet in diffusion, and we have parameter that LLM might not use like `target_modules`.  If we think adding target_modules to [UpdateWeightFromDiskReqInput](https://github.com/sgl-project/sglang/blob/f1ff697494c59bd0a4d298d5a1411f9eed563982/python/sglang/srt/managers/io_struct.py#L1259) is acceptable, I'm ok reuse LLM one.
2. Other request that in diffusion are in openai folders but this doesn't seem like to be an openai API.

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
dreamyang-liu: Testing Approach:

Scenario 1: Corrupted New Model Detection
Pull a diffusion model to a local folder, delete a tensor file from one of the module directories, then attempt to load it. The implementation should detect the corruption and rollback to the previous working model.

Scenario 2: Rollback Failure Handling
Corrupt the currently loaded model's checkpoint (delete partial checkpoint files), then trigger a reload. Expected behavior:
1. Load fails due to corruption
2. System attempts rollback
3. Rollback also fails (no valid fallback state)
4. System enters error state with appropriate error handling

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
dreamyang-liu: The latest change checks at runtime whether the module has active offload managers and routes accordingly, see [_load_weights_into_module](https://github.com/dreamyang-liu/sglang/tree/feat/diffusion-update-weights-from-disk/python/sglang/multimodal_gen/runtime/loader/weights_updater.py#L301) which calls [_get_offload_managers](https://github.com/dreamyang-liu/sglang/tree/feat/diffusion-update-weights-from-disk/python/sglang/multimodal_gen/runtime/loader/weights_updater.py#L294). We don't disable offload.

The reason we can't just "load it back to GPU and update normally" is that the diffusion offload manager replaces each parameter with a [torch.empty((1,)) placeholder](https://github.com/dreamyang-liu/sglang/tree/feat/diffusion-update-weights-from-disk/python/sglang/multimodal_gen/runtime/utils/layerwise_offload.py#L126) to free GPU memory, while real weights live in consolidated CPU buffers. Loading all layers back simultaneously would defeat the purpose of offloading (the model doesn't fit in GPU memory). Instead we write directly into the CPU buffers via [update_cpu_weights()](https://github.com/dreamyang-liu/sglang/tree/feat/diffusion-update-weights-from-disk/python/sglang/multimodal_gen/runtime/utils/layerwise_offload.py#L280).

The LLM engine doesn't need this because its offloader preserves original parameter shapes — it [moves params to CPU](https://github.com/dreamyang-liu/sglang/tree/feat/diffusion-update-weights-from-disk/python/sglang/srt/utils/offloader.py#L97) but keeps shapes intact, using [functional_call()](python/sglang/srt/utils/offloader.py#L144) to pass them through forward. So standard param.data.copy_() just works. The diffusion offloader uses a different design (consolidated flat buffers + placeholders).

python/sglang/multimodal_gen/runtime/managers/io_struct.py
dreamyang-liu: I just move it to `WeightsUpdater` for now, given it's the most related file for now.

zhaochenyang20: Thanks. Quite a good design pattern problem. I will ask mick and yuhao for decision. At the same time, could you help to rebase this PR. And, welcome to review this PR:

https://github.com/radixark/miles/pull/544

@yuhao318 @mickqian

python/sglang/multimodal_gen/runtime/entrypoints/http_server.py
```diff
@@ -92,6 +95,40 @@ async def health_generate():
     return {"status": "ok"}
 
 
+@health_router.post("/update_weights_from_disk")
+async def update_weights_from_disk(request: Request):
```
mickqian: could we have this in a separate file, for all the upcoming post-training related apis?

dreamyang-liu: Yeah, we can do that, @mickqian  have you foreseen how many or what APIs that we might need ? If there're a lot coming, I think we can create a new folder for separation.

Also, could you also provide your comment on the `UpdateWeightFromDiskReqInput` for diffusion in above comment ? https://github.com/sgl-project/sglang/pull/18306#discussion_r2776318325

Actually I'm thinking we can create a folder `post_training/` and have `post_training/weights_api.py` and `post_training/utils.py` (for `UpdateWeightFromDiskReqInput`and future requests) and import them into `http_server.py`

docs/advanced_features/sglang_for_rl.md
```diff
@@ -106,6 +106,27 @@ This path trades some I/O overhead for simplicity and flexibility. It integrates
 
 **Python Engine API:** `engine.update_weights_from_disk(model_path, load_format=None)`
 
+**Diffusion engine (SGLang-D):** The diffusion engine exposes the same `POST /update_weights_from_disk` endpoint with the following behavior:
```
mickqian: nit: `SGLang-Diffusion`

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
```diff
@@ -0,0 +1,356 @@
+"""
```
mickqian, line 1: I'm afraid this is probably not a component loader (since it's in the loader directory), I'll draft a pr to collect component loaders to a single folder

python/sglang/multimodal_gen/runtime/managers/io_struct.py
mickqian: for http-endpoint-usage only types, it should stay in the `entrypoints/openai` folder
actually the name `openai` is a bit obsolete, should be something like `http_server`

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
dreamyang-liu: Sounds good — I'll keep it in loader/ for now and we can move it in your follow-up PR to avoid conflicts.

zhaochenyang20: Current design looks good to me 😂

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
```diff
@@ -0,0 +1,346 @@
...
+        for name in names:
+            module = self.pipeline.get_module(name)
+            if module is not None and isinstance(module, torch.nn.Module):
+                result.append((name, module))
+
+        # Fallback for DiffusersPipeline: modules live on the diffusers pipe,
+        # not in self.pipeline.modules.
+        if not result:
+            diffusers_pipe = self.pipeline.get_module("diffusers_pipeline")
+            if diffusers_pipe is not None and hasattr(diffusers_pipe, "components"):
```
zhaochenyang20: remove this `hasattr`. Maybe we can use `if diffusers_pipe.components is not None`?

docs/advanced_features/sglang_for_rl.md
```diff
@@ -106,6 +106,27 @@ This path trades some I/O overhead for simplicity and flexibility. It integrates
...
 
+**Diffusion engine (SGLang-Diffusion):** The diffusion engine exposes the same `POST /update_weights_from_disk` endpoint with the following behavior:
+
+- **All-or-nothing with rollback:** if any module fails to load, all previously updated modules are rolled back to the original weights by reloading from the original model path. No partial updates are left behind. If rollback itself fails, the exception propagates so the caller knows the model is in an inconsistent state.
+- **Offload-aware:** when layerwise offload (`--dit-layerwise-offload`) is enabled, the diffusion offload manager replaces GPU parameters with small `torch.empty((1,))` placeholders while real weights live in consolidated pinned CPU buffers. A naive `param.data.copy_()` would fail with a shape mismatch. Instead, the updater dynamically detects active offload managers and writes new weights directly into their CPU buffers, bypassing the placeholders entirely. For any layer that happens to be prefetched on GPU at update time, the live GPU tensor is also updated so the change takes effect immediately. This requires no extra GPU memory and does not disturb the offload state.
```
zhaochenyang20, line 112: Since I am not familiar with the offloading settings of SGLang Diffusion, could Mick Yuhao help to check whether this makes sense? @mickqian @yhyang201 thanks! I will also check with the diffusion team.

python/sglang/multimodal_gen/runtime/entrypoints/post_training/weights_api.py
zhaochenyang20, line 1: Put it under `post_training` makes it rather organized. Nice done.

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
zhaochenyang20: The design is quite good. I think we need a test for the rollback in following PRs. Also, I think we should raise an error to logs right now that states "the weight update is wrong, the parameter will roll back to original weights and are not updated."

zhaochenyang20 (changes requested):

docs/advanced_features/sglang_for_rl.md
mickqian: this narrative is correct

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
```diff
@@ -0,0 +1,346 @@
...
+        load_weights_into_model(weights_iter, dict(module.named_parameters()))
+
+
+def load_weights_into_model(weights_iter, model_params: dict) -> None:
+    """Copy weights from weights_iter into model_params in-place.
+
+    Handles DTensor parameters by re-distributing the loaded weight
+    according to the existing device mesh and placements.
+
+    Raises:
```
mickqian: these comments are a bit redundant, could be simplified

docs/advanced_features/sglang_for_rl.md
zhaochenyang20: @RubiaCx Could you also find someone to reivew this.

zhaochenyang20: Thanks. I understand your nice design. Just have one comment that:

> For any layer that happens to be prefetched on GPU at update time, the live GPU tensor is also updated so the change takes effect immediately. This requires no extra GPU memory and does not disturb the offload state.

If a layer is currently prefetched on GPU during update time, is it under the current inference workload? In other words, hot-refitting to LLM is still under discussion in RL. Someone states that we can hot-refit an LLM when it's decoding, just need to flash the kv cache, but someone says we can only refit a model when it's not inferencing. To me, I think for diffusion, we should assert that when `update_weights_from_disk` is called, the model should not be making any inference work. In other words, we do not allow hot refitting, which is only used in async RL.

zhaochenyang20 (changes requested):

python/sglang/multimodal_gen/runtime/entrypoints/http_server.py
zhaochenyang20: Following our current design pattern, I think multiple routers are acceptable. But I raised a question here.


https://github.com/sgl-project/sglang/issues/18568

zhaochenyang20 (changes requested):

python/sglang/multimodal_gen/runtime/entrypoints/post_training/weights_api.py
```diff
@@ -0,0 +1,45 @@
+"""Post-training APIs: weight updates and related operations."""
```
zhaochenyang20: > """Post-training APIs: weight updates and related operations."""

Our post training shall further have other APIs. Please modify this.

python/sglang/multimodal_gen/runtime/entrypoints/post_training/io_struct.py
zhaochenyang20, line 1: As we have created this `post_training` subfolder, I think we can name this file to `python/sglang/multimodal_gen/runtime/entrypoints/post_training/io_struct.py`. Following LLM's design pattern.

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
```diff
@@ -0,0 +1,351 @@
...
+  fail with a shape mismatch.  Instead, the updater dynamically detects
+  active offload managers and writes new weights directly into their CPU
+  buffers via update_cpu_weights(), bypassing the placeholders entirely.
+  For any layer that happens to be prefetched on GPU at update time, the
+  live GPU tensor is also updated so the change takes effect immediately.
+  This requires no extra GPU memory and does not disturb the offload state.
+
+- DTensor-aware: parameters that have been distributed via
+  torch.distributed.tensor are updated through distribute_tensor
+  so that each shard is correctly placed on the right device mesh.
```
zhaochenyang20, line 38: Make sure that these design patterns are consistent across all the files. I mean, check whether your docs are the same as these comments.

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
```diff
@@ -0,0 +1,351 @@
...
+from sglang.multimodal_gen.runtime.pipelines.diffusers_pipeline import DiffusersPipeline
+from sglang.multimodal_gen.runtime.utils.hf_diffusers_utils import maybe_download_model
+from sglang.multimodal_gen.runtime.utils.layerwise_offload import OffloadableDiTMixin
+from sglang.multimodal_gen.runtime.utils.logging_utils import init_logger
+
+try:
+    from torch.distributed.tensor import DTensor, distribute_tensor
+except ImportError:
+    DTensor = None
+    distribute_tensor = None
```
zhaochenyang20: Why this import can fail? If it failed, what will your API do?

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
```diff
@@ -0,0 +1,351 @@
...
+        model_path: str,
+        original_model_path: str,
+        flush_cache: bool = True,
+        target_modules: list[str] | None = None,
+    ) -> tuple[bool, str]:
+        """Update model weights from disk without restarting the server.
+
+        Args:
+            model_path: HF repo id or local path to the new weights.
+            original_model_path: Path to the currently loaded weights (used
```
zhaochenyang20: I do not know how LLM's API is handling the roll back logics. But, we do not need a specific `original_model_path` argument to handle the roll back.

```python
    def update_weights_from_disk(
        self,
        model_path: str,
        load_format: Optional[str] = None,
    ):
        """Update the weights from disk inplace without re-launching the engine.

        This method allows updating the model weights from disk without restarting
        the engine. It can be used to load a different model or update weights with
        new training.
        """
```

We should change this accordingly.

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
```diff
@@ -0,0 +1,351 @@
...
+                for rollback on failure).
+            flush_cache: If True, reset TeaCache state after a successful
+                update so that stale cached residuals are not reused.
+            target_modules: Explicit list of module names to update.  None
+                or ["all"] updates every nn.Module in the pipeline.
```
zhaochenyang20: Do not use `"all"`, which seems to be weird 😂 Only let `None` to refer to all the modules.

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
```diff
@@ -0,0 +1,351 @@
...
+
+        Returns:
+            (success, message) tuple.
```
zhaochenyang20: I think `success` is a bad parameters naming. Could we make it like, "state"?

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
```diff
@@ -0,0 +1,351 @@
...
+        """
+        tic = time.perf_counter()
+        self._original_model_path = original_model_path
```
zhaochenyang20: As I said, I highly suspect this is not needed.

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
```diff
@@ -0,0 +1,351 @@
...
+        success, message = self._apply_weights(modules_to_update, weights_map)
+
+        gc.collect()
+        torch.cuda.empty_cache()
+
+        if success and flush_cache:
+            for _, module in modules_to_update:
+                _reset_cache_state(module)
+
+        elapsed = time.perf_counter() - tic
```
zhaochenyang20: No need to calculate `elapsed` in normal use case. this calculation is only used for debugging and in the unit test.

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
```diff
@@ -0,0 +1,351 @@
...
+        return success, message
+
+    # ------------------------------------------------------------------
+    # Private helpers
+    # ------------------------------------------------------------------
+
+    def _collect_modules(
+        self, target_modules: list[str] | None
+    ) -> list[tuple[str, torch.nn.Module]]:
+        """Resolve target_modules to (name, module) pairs.
```
zhaochenyang20, line 225: Why should we name this "resolve", it should be better for "convert"?

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
```diff
@@ -0,0 +1,351 @@
...
+
+        Raises:
+            ValueError: If target_modules contains names not found in the pipeline.
+        """
+        components = get_updatable_modules(self.pipeline)
+
+        if target_modules is None or target_modules == ["all"]:
```
zhaochenyang20: remove the `["all"]`.

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
```diff
@@ -0,0 +1,351 @@
...
+        original_path = maybe_download_model(self._original_model_path)
+        for name in updated_modules:
+            module = self.pipeline.get_module(name)
+            if module is None:
+                continue
+            weights_dir = find_weights_dir(original_path, name)
+            if weights_dir is None:
+                continue
+            weights_iter = _get_weights_iter(weights_dir)
+            _load_weights_into_module(module, weights_iter)
```
zhaochenyang20, line 293: As I said, I think we should have other methods to `_rollback`.

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
```diff
@@ -0,0 +1,351 @@
...
+    <local_path>/<module_name>/ first, then falls back to local_path
+    itself if it directly contains safetensors files (common for RL
+    checkpoints that save weights in a flat directory).
+    """
+    dir_path = os.path.join(local_path, module_name)
+    if os.path.exists(dir_path):
+        return dir_path
+    if _list_safetensors_files(local_path):
+        return local_path
+    return None
```
zhaochenyang20: This seems to be quite a general method. Put it elsewhere. Like in a general `utils.py`.

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
```diff
@@ -0,0 +1,351 @@
...
+
+    For offloaded modules, updates CPU buffers directly via
+    update_cpu_weights(); non-offloaded parameters use in-place copy.
+    """
+    offload_managers = _get_offload_managers(module)
+    if offload_managers:
+        weight_dict = dict(weights_iter)
+        offloaded_names: set[str] = set()
+        for manager in offload_managers:
+            offloaded_names |= manager.update_cpu_weights(weight_dict)
```
zhaochenyang20: we seldom use `|=`. Could you use: `offloaded_names.update(manager.update_cpu_weights(weight_dict))`?

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
```diff
@@ -0,0 +1,351 @@
...
+                loaded_weight.to(param.dtype),
+                param.device_mesh,
+                param.placements,
+            )
+            param._local_tensor.copy_(distributed_weight._local_tensor)
+        else:
+            param.data.copy_(loaded_weight.to(param.dtype))
+
+
+def _reset_cache_state(module: torch.nn.Module) -> None:
```
zhaochenyang20: Too short a funcition. ->  No need to be a function.

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
```diff
@@ -0,0 +1,351 @@
...
+        else:
+            missing.append(module_name)
+    return weights_map, missing
+
+
+def _get_offload_managers(module: torch.nn.Module) -> list:
+    """Return active offload managers for the given module, if any."""
+    if isinstance(module, OffloadableDiTMixin) and module.layerwise_offload_managers:
+        return [m for m in module.layerwise_offload_managers if m.enabled]
+    return []
```
zhaochenyang20: No need to be a function.

python/sglang/multimodal_gen/runtime/utils/layerwise_offload.py
```diff
@@ -276,6 +276,60 @@ def sync_all_layers_to_cpu(self) -> None:
         for layer_idx in list(self._gpu_layers):
             self.sync_layer_to_cpu(layer_idx)
 
+    @torch.compiler.disable
+    def update_cpu_weights(self, weight_dict: Dict[str, torch.Tensor]) -> Set[str]:
+        """Update consolidated CPU buffers with new weights.
```
zhaochenyang20, line 283: Adds this here:

> when layerwise offload (`--dit-layerwise-offload`) is enabled, the diffusion offload manager replaces GPU parameters with small `torch.empty((1,))` placeholders while real weights live in consolidated pinned CPU buffers. A naive `param.data.copy_()` would fail with a shape mismatch. Instead, the updater dynamically detects active offload managers and writes new weights directly into their CPU buffers, bypassing the placeholders entirely. For any layer that happens to be prefetched on GPU at update time, the live GPU tensor is also updated so the change takes effect immediately. This requires no extra GPU memory and does not disturb the offload state.

python/sglang/multimodal_gen/runtime/utils/layerwise_offload.py
```diff
@@ -276,6 +276,60 @@ def sync_all_layers_to_cpu(self) -> None:
...
+            weight_dict: Mapping of parameter name to new weight tensor.
+
+        Returns:
+            Set of parameter names that were successfully updated.
+
+        Raises:
+            ValueError: If a weight's shape does not match the recorded
+                metadata (i.e. the real shape, not the placeholder shape).
+        """
+        if not self.enabled:
```
zhaochenyang20, line 304: If not `enabled`, we should not return an empty `set()`. Which feels like no weights are successfully updated. Instead, just assert that it should be enabled in this function. And, check whether `--dit-layerwise-offload` is enabled before actaully calling this function.

python/sglang/multimodal_gen/runtime/utils/layerwise_offload.py
```diff
@@ -276,6 +276,60 @@ def sync_all_layers_to_cpu(self) -> None:
...
+                    f"loaded={tuple(loaded_weight.shape)}"
+                )
+
+            dtype = meta["dtype"]
+            offset = meta["offset"]
+            numel = meta["numel"]
+            cpu_buffer = self._consolidated_cpu_weights[layer_idx][dtype]
+            cpu_buffer[offset : offset + numel].copy_(
+                loaded_weight.to(dtype=dtype).flatten()
+            )
```
zhaochenyang20, line 330: Nice done. But we make unit tests to check whether we can successfully update weights or roll back when `--dit-layerwise-offload` is enabled.

zhaochenyang20 (changes requested):

docs/advanced_features/sglang_for_rl.md
dreamyang-liu: Actually No — a prefetched layer is not under active inference during update_weights_from_disk. The current architecture already guarantees this.

The diffusion scheduler is a single-threaded event loop that processes one complete request per iteration. `execute_forward` runs the entire denoise loop synchronously before returning, so there are never any in-flight requests when the next item (e.g. a weight update) is dispatched. A layer in _gpu_layers at update time is simply a stale resident from the previous inference — no CUDA kernel is reading it.

This differs from the LLM scheduler, which is also single-threaded but uses continuous batching: one iteration = one decode step for all requests in the running batch. Requests span hundreds of iterations, so a weight update may arrive while the running batch is non-empty with KV caches computed from old weights. That's why the LLM side needs RWLock + drain + KV cache flush, it's due to request lifetimes cross iteration boundaries.

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
dreamyang-liu: success here is a boolean and match the semantic. 
LLM side also use this naming, why you think it's bad ? 

https://github.com/sgl-project/sglang/blob/2cc235e7952e6565a06ac2e179e0e1dffdd9a2c9/python/sglang/srt/managers/tokenizer_manager.py#L1347-L1378


https://github.com/sgl-project/sglang/blob/2cc235e7952e6565a06ac2e179e0e1dffdd9a2c9/python/sglang/srt/managers/scheduler_update_weights_mixin.py#L46-L57

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
dreamyang-liu: I'm fine putting it inline given only one case it's handling. But if we have more caches need to clean up, i still think we need to extract it as a function

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
dreamyang-liu: Could you clarify what you mean by "other methods to _rollback"?

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
dreamyang-liu: DTensor is only needed when multi-gpu used. This try-catch would allow it still work when using an old version of pytorch under single gpu mode. 

Given current repo require `torch==2.9.1`, probably we can remove it.

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
dreamyang-liu: The function isn't converting between formats — it takes a list of module name strings and resolves them against the pipeline to locate the actual (name, nn.Module) pairs, raising if any name can't be found.

python/sglang/multimodal_gen/runtime/utils/layerwise_offload.py
dreamyang-liu: If `--dit-layerwise-offload` not enabled, this shouldn't be called at all given no offload manager will be created. The only possible scenario is using this offload manager is enable `--dit-layerwise-offload` on a non-CUDA accelerator. 

I can change it return None, assertion to fail might not follow the pattern in other interfaces in the class. Other interfaces just return and don't do anything. 

To be honest, I think we should move this condition out of the class, so that if not qualified for offloading, we shouldn't create offload manager at all.


```
self.enabled = bool(enabled and torch.cuda.is_available())
```

https://github.com/sgl-project/sglang/blob/4f7422f7bada1f1d83b2cadc3ae3e609820f0e41/python/sglang/multimodal_gen/runtime/utils/layerwise_offload.py#L43C9-L43C67


https://github.com/sgl-project/sglang/blob/4f7422f7bada1f1d83b2cadc3ae3e609820f0e41/python/sglang/multimodal_gen/runtime/managers/gpu_worker.py#L125C29-L142

python/sglang/multimodal_gen/test/server/test_update_weights_from_disk.py
```diff
@@ -0,0 +1,319 @@
+"""
+Tests for update_weights_from_disk API in SGLang-D (diffusion engine).
+
+This tests the ability to dynamically update model weights without restarting the server,
+which is critical for RL workflows and iterative fine-tuning scenarios.
+"""
```
zhaochenyang20: Give a detailed documentation of how these 9 tests are organized.  What are your three test cases covered with? What do these three classes mean to do?



How to simulate the actual RL scenarios. Please refer to this file:

https://github.com/sgl-project/sglang/blob/main/test/registered/rl/test_update_weights_from_distributed.py

zhaochenyang20: I made it myself.

zhaochenyang20 (changes requested):

zhaochenyang20:
> > 1. Overall Feedback & Docstrings
> >    I have thoroughly reviewed this unit test suite. It is exceptionally well-structured and demonstrates great attention to detail. Huge thanks to Menyang for the solid work. I have added documentation comments (docstrings) from my side. To be completely candid, these were generated via "vibe coding," but I have manually verified and audited them multiple times to ensure accuracy.
> >    
> >    2. Streamlining the Test Suite
> >       I have a strong intuition regarding Class 3: TestUpdateWeightsEndToEnd. If we can rigorously guarantee that the weight update itself is correct at the parameter level, this end-to-end generation test becomes redundant. I propose removing this class, which would leave us with a leaner suite of 2 main classes and 8 test cases.
> >    3. Realism in RL Scenarios
> >       The more critical issue is: How do we ensure the update is truly correct? Currently, the test starts a server with FLUX.2-klein-4B and then "updates" it using the exact same model path. This is a bit of a "circular" test and doesn't reflect real-world RL workflows where a model evolves. A more robust and logical approach—similar to our [distributed RL test](https://github.com/sgl-project/sglang/blob/main/test/registered/rl/test_update_weights_from_distributed.py)—would be to initialize the server with a Base model (e.g., FLUX.2-klein-base-4B) and then trigger an update to the Instruct model (FLUX.2-klein-4B). This correctly simulates the iterative refinement process in RL.
> >    4. Strict Weight Verification Strategy
> >       Changing the models will make the tests more complex, necessitating a strict verification mechanism. While my previous get_weights_by_name implementation works, it isn't the most efficient. I’m considering two directions:
> > 
> > Option A (The "Steady" Path): Use a get_weights_by_name style interface to pull specific layer parameters from the running SGLang server and compare them against the target model weights.
> > Option B (The "Cool" Path): 1. Load the target Instruct model using a standard Hugging Face diffusers pipeline. 2. Calculate a SHA-256 hash of its parameter tensors. 3. Release the HF pipeline memory. 4. Start the SGLang server with the Base model, perform the update, and then expose an interface to calculate/retrieve the SHA-256 hash of the internal weights. 5. Compare the two hashes.
> > While Option B is more elegant, we must ensure the parameters are bit-wise equal, which is a stricter condition than standard PyTorch tensor equality. I’m open to either approach, though Option A might be more predictable in the short term.
> 
> Thanks @zhaochenyang20 for the feedback.
> 
> For 2, we can remove it cause it could be included in other tests.
> 
> For 3, I think we can adopt the same approach to use `instruct` model to update `base` model for simulating the RL update
> 
> For 4, I feel like we can use generation result to verify whether it is correctly updated.
> 
> (1) Generate image A with black-forest-labs/FLUX.2-klein-base-4B (fixed prompt + seed). (2) Swap toblack-forest-labs/FLUX.2-klein-4B . (3) Generate image B with same prompt + seed → assert A ≠ B. (4) Swap back to black-forest-labs/FLUX.2-klein-4B . (5) Generate image C with same prompt + seed → assert A == C.
> 
> It can also verify the updated weight is not corrupted and can function well.

1. Unreliability of Generation Results: In rigorous, large-scale systems, relying solely on generation outcomes to validate parameter updates is insufficient. Issues like disk write errors or silent failures during the loading process can occur. These subtle bugs might not cause the generation to fail outright, but they compromise the model's integrity.

2. Granularity of Weight Updates: In real-world RL or fine-tuning scenarios, weight updates are often incremental, and the parameter deltas can be extremely small. Such minor differences are difficult to capture through a single generation. It is entirely possible for two slightly different model versions to produce bit-for-bit identical outputs for the same prompt and seed.

3. The Necessity of Strict Assertion: Our goal is to ensure that "what is loaded is exactly what was intended." By verifying the Hash of the parameters or performing direct Tensor comparisons, we can guarantee atomicity and accuracy at the infrastructure level. Relying on Model Generated Content (MGC) as a proxy is simply too high-level and lacks the precision needed for system-level unit tests.

python/sglang/multimodal_gen/runtime/loader/weight_utils.py
```diff
@@ -335,3 +336,23 @@ def maybe_remap_kv_scale_name(name: str, params_dict: dict) -> str | None:
...
+
+def compute_weights_checksum(
+    named_params: Iterable[tuple[str, torch.Tensor]],
+) -> str:
+    """Compute SHA-256 checksum over (name, tensor) pairs.
+
+    Parameters are sorted by name so the digest is deterministic
+    regardless of iteration order.  Raw bytes are hashed directly
+    (no dtype conversion) for speed and fidelity.
+    """
```
zhaochenyang20, line 350: add comments that this is a helper function for checking the correctness of weight refiting.

python/sglang/multimodal_gen/runtime/loader/weight_utils.py
```diff
@@ -335,3 +336,23 @@ def maybe_remap_kv_scale_name(name: str, params_dict: dict) -> str | None:
 
     # If there were no matches, return the untouched param name
     return name
+
+
+def compute_weights_checksum(
+    named_params: Iterable[tuple[str, torch.Tensor]],
+) -> str:
+    """Compute SHA-256 checksum over (name, tensor) pairs.
```
zhaochenyang20: Compute a unique SHA-256 value for a group of (name, tensor) paris.

zhaochenyang20 (changes requested):

python/sglang/multimodal_gen/runtime/managers/gpu_worker.py
```diff
@@ -341,6 +349,48 @@ def list_loras(self) -> OutputBatch:
...
+            return {"error": "Pipeline is not initialized"}
+
+        all_modules = get_updatable_modules(self.pipeline)
+        names = module_names if module_names is not None else list(all_modules.keys())
+
+        checksums: dict[str, str] = {}
+        for name in names:
+            module = all_modules.get(name)
+            if module is None:
+                checksums[name] = "not_found"
```
zhaochenyang20, line 417: In this sense, there should raise an error? rather than assign it to `"not_found"`?

zhaochenyang20 (changes requested):

python/sglang/multimodal_gen/runtime/managers/gpu_worker.py
dreamyang-liu: Intentionally not raising here, a typo in the user's module_names list shouldn't bring down the server. In this case, it's a client side issue, and assign it to not_found can help client identify it.

zhaochenyang20:
I changed this test:

```bash
  • test_update_weights_specific_modules

    Randomly selects a subset of pipeline modules as target_modules, then asserts:
    (1) updated modules' checksums match the update model's disk checksums;
    (2) non-updated modules' checksums are unchanged (same before and after).
```

python/sglang/multimodal_gen/test/server/test_update_weights_from_disk.py
```diff
@@ -420,22 +457,67 @@ def test_update_weights_missing_model_path(self, diffusion_server_no_offload):
...
+                assert after_checksums.get(name) == disk_cs, (
+                    f"Updated module '{name}': checksum should match update model disk\n"
+                    f"  disk: {disk_cs}\n  gpu:  {after_checksums.get(name)}"
+                )
+            else:
+                assert after_checksums.get(name) == before_checksums.get(name), (
+                    f"Non-updated module '{name}': checksum must be unchanged\n"
+                    f"  before: {before_checksums.get(name)}\n"
+                    f"  after:  {after_checksums.get(name)}"
+                )
```
dreamyang-liu: Might not be a blocker, just a note. From my observation, base model and instruct model only have difference in transformer module. If accidentally update `vae` or `text_encoder`, they might still have same checksum. Not sure what we want to verify here.

zhaochenyang20: Oh. Great pointing out.

python/sglang/multimodal_gen/runtime/entrypoints/post_training/weights_api.py
```diff
@@ -0,0 +1,62 @@
...
+    if not model_path:
+        return ORJSONResponse(
+            {"success": False, "message": "model_path is required"},
+            status_code=400,
+        )
+
+    req = UpdateWeightFromDiskReqInput(
+        model_path=model_path,
+        flush_cache=body.get("flush_cache", True),
+        target_modules=body.get("target_modules"),
```
zhaochenyang20, line 29: should it be `body.get("target_modules", None),`?

python/sglang/multimodal_gen/runtime/loader/weight_utils.py
```diff
@@ -335,3 +336,28 @@ def maybe_remap_kv_scale_name(name: str, params_dict: dict) -> str | None:
...
+
+def compute_weights_checksum(
+    named_params: Iterable[tuple[str, torch.Tensor]],
+) -> str:
+    """Compute a SHA-256 checksum for a set of (name, tensor) pairs.
+
+    Helper function for verifying the correctness of weight refitting
+    (update_weights_from_disk). After a refit, callers can compare the
+    checksum of the in-GPU model weights against the checksum of the
+    on-disk tensors to confirm they match exactly.
```
zhaochenyang20: No need to restrict it to `update_weights_from_disk`. We will add more refit methods.

zhaochenyang20 (changes requested):

python/sglang/multimodal_gen/runtime/loader/utils.py
```diff
@@ -145,6 +145,18 @@ def _list_safetensors_files(model_path: str) -> list[str]:
...
+def find_weights_dir(local_path: str, module_name: str) -> str | None:
+    """Locate the safetensors directory for module_name under local_path.
+
+    Diffusion models store weights in per-module subdirectories (e.g.
+    transformer/, vae/, text_encoder/).
+    """
+    dir_path = os.path.join(local_path, module_name)
+    if os.path.exists(dir_path):
+        return dir_path
+    return None
```
zhaochenyang20: Better to use `Pathlib` instead of `os.path`. `from Pathlib import Path`.

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
zhaochenyang20, line 1: This file will not only contain `update_weights_from_disk`. It's good right now. But we will suport more.

python/sglang/multimodal_gen/runtime/entrypoints/post_training/weights_api.py
dreamyang-liu: dict.get() returns None by default when the key is missing

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
```diff
@@ -0,0 +1,322 @@
...
+            for _, module in modules_to_update:
+                if isinstance(module, TeaCacheMixin):
+                    module.reset_teacache_state()
+
+        logger.info(message)
+        return success, message
+
+    # ------------------------------------------------------------------
+    # Private helpers
+    # ------------------------------------------------------------------
```
zhaochenyang20: remove thise

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
```diff
@@ -0,0 +1,322 @@
...
+            )
+        if isinstance(param, DTensor):
+            distributed_weight = distribute_tensor(
+                loaded_weight.to(param.dtype),
+                param.device_mesh,
+                param.placements,
+            )
+            param._local_tensor.copy_(distributed_weight._local_tensor)
+        else:
+            param.data.copy_(loaded_weight.to(param.dtype))
```
zhaochenyang20: Put these function above `WeightsUpdater`

zhaochenyang20 (changes requested):

dreamyang-liu:
StrEnum is not available before python 3.11, I will refactor it to make it compatible.

zhaochenyang20 (approved):

python/sglang/multimodal_gen/runtime/loader/utils.py
```diff
@@ -148,6 +149,18 @@ def _list_safetensors_files(model_path: str) -> list[str]:
     return sorted(glob.glob(os.path.join(str(model_path), "*.safetensors")))
 
 
+def find_weights_dir(local_path: str, module_name: str) -> Path | None:
```
mickqian: look for https://github.com/sgl-project/sglang/blob/b21390f8f396f734a0d516009fa2c87f2b5e645b/python/sglang/multimodal_gen/runtime/pipelines_core/composed_pipeline_base.py#L290 as well

do we really need this helper? it's a one-liner

python/sglang/multimodal_gen/runtime/loader/weights_updater.py
```diff
@@ -0,0 +1,313 @@
...
+    def __init__(self, pipeline):
+        self.pipeline = pipeline
+
+    def update_weights_from_disk(
+        self,
+        model_path: str,
+        flush_cache: bool = True,
+        target_modules: list[str] | None = None,
+    ) -> tuple[bool, str]:
+        """Update model weights from disk without restarting the server.
```
mickqian: remove excessive docstrings about Args and Returns

python/sglang/multimodal_gen/runtime/loader/utils.py
dreamyang-liu: During init, panicking on missing paths is fine — we shouldn't launch a broken server. But during weight updates, we want graceful handling. Returning `None` makes the "not found" case explicit, letting callers decide
 whether to panic or skip. A bare `os.path.join()` would silently propagate invalid paths until they cause unexpected failures downstream.

zhaochenyang20: I will use `pathlib.Path` to refactor.

mickqian (approved):

python/sglang/multimodal_gen/runtime/loader/weight_utils.py
```diff
@@ -336,3 +337,23 @@ def maybe_remap_kv_scale_name(name: str, params_dict: dict) -> str | None:
...
+    compare the checksum of the in-GPU model weights against the checksum
+    of the on-disk tensors or the tensors in the training engine.
+    """
+    hasher = hashlib.sha256()
+    for name, tensor in sorted(named_params, key=lambda x: x[0]):
+        hasher.update(name.encode())
+        t = tensor.detach()
+        # DTensor doesn't support .numpy(); extract the local tensor.
+        if isinstance(t, DTensor):
+            t = t._local_tensor
```
Rockdu, line 357: Since some of the DTensors may be sharded across the devices and local_tensor is only the tensor on current device, do we need all-gather or some hash value merging logics here?

dreamyang-liu: Good question, let me check.

PR #18308, opened 2026-02-05
fix the assert error when set mtp and disable-cuda-graph

yzk1989:
my test commond is 
python3 -m  sglang.launch_server \
    --model-path /upfs/models/deepseek-ai/DeepSeek-V3.2 \
    --disaggregation-ib-device mlx5_0,mlx5_1,mlx5_2,mlx5_3,mlx5_4,mlx5_5,mlx5_6,mlx5_7 \
    --disaggregation-transfer-backend mooncake \
    --dist-init-addr $d0:5757 \
    --nnodes 1 \
    --node-rank $di \
    --tp 8 \
    --ep 8 \
    --dp 8 \
    --enable-dp-attention \
    --decode-log-interval 1 \
    --enable-metrics \
    --enable-metrics-for-all-schedulers \
    --host 0.0.0.0 \
    --port 8000 \
    --trust-remote-code \
    --moe-dense-tp-size 1 \
    --enable-dp-lm-head \
    --enable-cache-report \
    --watchdog-timeout 1000000 \
    --deepep-mode low_latency \
    --mem-fraction-static 0.8 \
    --max-running-requests 64 \
    --context-length 3000 \
    --moe-a2a-backend deepep \
    --prefill-round-robin-balance \
    --page-size 64 \
    --log-level debug \
    --speculative-num-steps 3 \
    --speculative-eagle-topk 1 \
    --speculative-attention-mode decode \
    --disable-cuda-graph \
    --load-balance-method round_robin

when I set mtp and disable-cuda-graph,there will be an error 
image
,so I fix it


hnyls2002:
@yzk1989 The code this patches (speculative/eagle_worker.py, the Spec V1 worker) was removed from main when Spec V1 was retired (#27964 and related). The num_tokens_per_batch path you fix no longer exists in the V2 worker. Closing as obsolete; if the same assert reproduces on the current spec V2 path, please reopen with a fresh repro.

Issue #18313, opened 2026-02-05
[Feature] Change  new allreduce_fusion api from Flashinfer


Fridge003:
### Checklist

- [ ] If this is not a feature request but a general question, please start a discussion at https://github.com/sgl-project/sglang/discussions. Otherwise, it will be closed.
- [ ] Please use English. Otherwise, it will be closed.

### Motivation

Currently SGLang calls allreduce fusion through `flashinfer.comm.trtllm_allreduce_fusion` API [here](https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/layers/flashinfer_comm_fusion.py#L194)

This API will be replaced by `flashinfer.comm.allreduce_fusion`(Reference: https://github.com/flashinfer-ai/flashinfer/blob/f84ac1c97e2e1a4391150ca73971b973ee569e5a/flashinfer/comm/allreduce.py#L447) in the future.  So we need to change the usage of all reduce fusion

PR #18322, opened 2026-02-05
[Feature] Change new allreduce_fusion api from Flashinfer

ghost:
Issue: https://github.com/sgl-project/sglang/issues/18313

switched the flashinfer fused allreduce path to the new allreduce_fusion api. 

cc @Fridge003

### DIFF
--- a/python/sglang/srt/layers/flashinfer_comm_fusion.py
+++ b/python/sglang/srt/layers/flashinfer_comm_fusion.py
@@ -40,6 +40,7 @@ class FlashInferWorkspaceManager:
         hidden_dim: int,
         group=None,
         use_fp32_lamport: bool = False,
+        dtype: torch.dtype = torch.bfloat16,
     ):
         """Initialize workspace"""
         if self.initialized and self.world_size == world_size:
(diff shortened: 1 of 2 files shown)


python/sglang/srt/layers/flashinfer_comm_fusion.py
mmangkad: Please guard for unified API availability (`allreduce_fusion` / `create_allreduce_fusion_workspace`) to avoid runtime crashes on older FlashInfer.
```suggestion
        if hasattr(comm, "allreduce_fusion") and hasattr(
            comm, "create_allreduce_fusion_workspace"
        ):
            _flashinfer_comm = comm
        else:
            logger.warning(
                "flashinfer.comm unified allreduce_fusion API is not available, "
                "falling back to standard implementation"
            )
```

mmangkad: Need workspace size validation (`token_num`/`hidden_dim`/`dtype`/`use_oneshot`). Reusing blindly risks undersized buffers.

PR #18369, opened 2026-02-06
[Bug]Fix PP-safe v_head_dim usage in Triton backend

niuheben:
## Accuracy Tests

Tested inference correctness with Qwen3-32B model under pipeline parallel (PP) and tensor parallel (TP) settings on H2O cluster.  
All outputs matched expected results when using the Triton attention backend.

## Benchmarking and Profiling

Tested with the following multi-GPU configuration on H2O nodes:

```bash
# Prefill server (H2O node 0)
HIP_VISIBLE_DEVICES=0,1,2,3 python3 -m sglang.launch_server \
  --model-path /models/qwen3/Qwen3-32B \
  --port 30003 \
  --tp-size 1 \
  --pp-size 4 \
  --context-len 4096 \
  --attention-backend triton

# Decode server (H2O node 1)
HIP_VISIBLE_DEVICES=4,5,6,7 python3 -m sglang.launch_server \
  --model-path /models/qwen3/Qwen3-32B \
  --port 30001 \
  --tp-size 4 \
  --context-len 4096 \
  --attention-backend triton \
  --cuda-graph-max-bs 32

# Router
python -m sglang_router.launch_router \
  --pd-disaggregation \
  --prefill http://127.0.0.1:30003 \
  --decode http://127.0.0.1:30001 \
  --host 0.0.0.0 \
  --port 8000

### DIFF
--- a/python/sglang/srt/layers/attention/triton_backend.py
+++ b/python/sglang/srt/layers/attention/triton_backend.py
@@ -99,9 +99,9 @@ class TritonAttnBackend(AttentionBackend):
             # For hybrid linear models, layer_id = 0 may not be full attention
             self.v_head_dim = model_runner.token_to_kv_pool.get_v_head_dim()
         else:
-            self.v_head_dim = model_runner.token_to_kv_pool.get_value_buffer(0).shape[
-                -1
-            ]
+            # PP-safe: v_head_dim is a global metadata, do not rely on layer 0 buffer which might be non-existent or have different shape due to PP partition. 
+            # Instead, get it from model config which is consistent across layers.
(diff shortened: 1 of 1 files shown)


hnyls2002:
@niuheben I'm fairly sure this is fixed on main: triton_backend.py no longer reads get_value_buffer(0).shape[-1] for v_head_dim and instead uses model_config.v_head_dim / token_to_kv_pool.get_v_head_dim() (lines 146, 157), which is PP-safe. Closing as fixed-elsewhere; reopen if a PP partition still trips this.

PR #18417, opened 2026-02-07
[CI] fix: notebook ci may not working

### LINKED ISSUE #18430: [Bug] /tag-and-rerun-ci does not work as expected
### Checklist

- [ ] Issues without environment info and a minimal reproducible demo are hard to resolve and may receive no feedback.
- [ ] If this is not a bug report but a general question, please start a discussion at https://github.com/sgl-project/sglang/discussions. Otherwise, it will be closed.
- [ ] Please use English. Otherwise, it will be closed.

### Describe the bug
When we comment `/tag-and-rerun-ci` (which we do for most PRs), we expect it to add the `run-ci` label and trigger all required CI runs. However, this does not always happen in certain cases, such as the notebook workflow (even when the PR modifies files under `docs/**`).

### Reproduction
Steps to reproduce:

1. Modify an `.ipynb` file under `docs/` and open a PR.
2. Comment `/tag-and-rerun-ci` in the PR.
3. Observe that nothing happens: the notebook workflow is not triggered. As a result, this PR can be merged with changes under `docs/` but without running CI, which may later break the notebook CI.

Example PR: https://github.com/sgl-project/sglang/pull/18439

The same issue occurs with `/tag-and-run-ci`, but manually adding the `run-ci` label always works as expected. 
It is unclear whether other workflows are also affected by this issue.

As a reference, the current implementation for handling slash commands(`slash-command-hadler.yml`) uses `GITHUB_TOKEN`. However, it seems that `GITHUB_TOKEN` cannot trigger new workflows to run. See: https://docs.github.com/en/actions/concepts/security/github_token#when-github_token-triggers-workflow-runs

I’m not sure if this is the root cause of the issue, but it might be related.

### Environment

CI bug. not related to env

alphabetc1:
## Motivation
trying to fix https://github.com/sgl-project/sglang/issues/18430

In the previous code, the `Execute Notebooks` workflow could be effectively skipped due to a combination of factors:

- When a PR is created (`opened`), the notebook workflow is **not** triggered because `pull_request.types` did not include `opened`.
- If the PR does not receive any further pushes (`synchronize`) and no other triggering events occur (such as adding a label),
- Then **no notebook workflow run is ever created for that head SHA**.
- However, `/tag-and-rerun-ci` can only rerun jobs if there is already a workflow run record (either `skipped` or `failure`) for that head SHA.
- As a result, the entire notebook CI effectively gets skipped.

Adding `opened` alone only solves part of the problem: the workflow run will at least **exist**. But even if `run-all-notebooks` is skipped, the overall workflow often still ends in `success` instead of `skipped`. In that case, `handle_rerun_failed_ci` will not pick it up as a failed/skipped target to rerun. To fully address this, we also need to adjust both `call-gate` and the finish job (following the pattern used in the `PR Test `workflow).

### how to reproduce
1. Modify files under docs/** and submit a PR.
2. Tag the PR with /tag-and-rerun-ci.
3. After the run-ci label is added, the notebook CI is still not triggered and does not even appear in the Checks list.
For example: 
https://github.com/sgl-project/sglang/pull/18418

## Benchmarking and Profiling
See https://github.com/sgl-project/sglang/pull/18416.
After this fix, any non-markdown changes will correctly trigger the notebook CI once it is tagged with `/tag-and-rerun-ci`.

## Checklist

## Review Process

### DIFF
--- a/.github/workflows/execute-notebook.yml
+++ b/.github/workflows/execute-notebook.yml
@@ -3,10 +3,13 @@ name: Execute Notebooks
 on:
   pull_request:
     branches: [ main ]
-    types: [synchronize, labeled]
+    types: [opened, synchronize, reopened, labeled]
     paths:
       - "python/sglang/**"
       - "docs/**"
+    paths-ignore:
+      - "python/sglang/**/*.md"
+      - "docs/**/*.md"
   workflow_dispatch:
 
 
@@ -18,9 +21,16 @@ env:
   SGLANG_IS_IN_CI: true
 
 jobs:
+  call-gate:
+    # Align with PR Test: fail fast if PR doesn't have run-ci label.
+    # This makes /tag-and-rerun-ci work by rerunning this failed workflow.
+    uses: ./.github/workflows/pr-gate.yml
+    secrets: inherit
+
   run-all-notebooks:
+    needs: [call-gate]
     runs-on: 1-gpu-runner
-    if: github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'run-ci')
+    if: github.event_name != 'pull_request' || needs.call-gate.result == 'success'
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


zhaochenyang20:
> After this fix, a new PR that modifies docs/** will correctly trigger the notebook CI once it is tagged with /tag-and-rerun-ci.

Not that, as I mentioned, any non-markdown changes should trigger this CI. If there is a single line of code change, even in Python with just a blank line, this CI should also be triggered.

PR #18435, opened 2026-02-08
Allow custom status codes to be returned for scheduler aborts.

harrisonlimh:
## Motivation
Different organizations have different preferences in handling throttling and server overloaded behaviors. In it, they may prefer using different status codes. 

This change users to configure the status coed baed on environment variables.  

## Modifications
* Allow users to define and return custom status codes for scheduler aborts. 
* Leave the current codes as default for backward compatibility 
* Clarify error message 
 

## Checklist

## Review Process


JustinTong0323 (approved):

test/registered/scheduler/test_abort.py
```diff
@@ -251,6 +254,19 @@ def setUpClass(cls):
                 ],
             )
 
+        if cls._timed_out_request_status_code is None:
```
hnyls2002: nits: Unify this logic. When the status code is not set for this test, you can set the class var to 503 and then use the same code under the `else` branch

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -2162,6 +2164,10 @@ def force_exit_handler(self):
     def _handle_abort_req(self, recv_obj: AbortReq):
         if is_health_check_generate_req(recv_obj):
             return
+        # State for the target request is already cleaned up.
+        if recv_obj.rid not in self.rid_to_state:
+            return
+
```
hnyls2002, line 2144: Is this dead code? the original impl does not contain these lines. Do you consider it as a fix or a guard?

hnyls2002 (changes requested):

test/registered/scheduler/test_abort.py
harrisonlimh: Done!

python/sglang/srt/managers/tokenizer_manager.py
harrisonlimh: Sorry, I should have shared more context. 

We started noticing KeyError that happens sometimes after throttling. This is a guard to mitigate it for short term. 
 
```
[2026-01-30 22:17:00] TokenizerManager hit an exception: Traceback (most recent call last):
  File "/sgl-workspace/sglang/python/sglang/srt/managers/tokenizer_manager.py", line 2471, in print_exception_wrapper
    await func()
  File "/sgl-workspace/sglang/python/sglang/srt/managers/tokenizer_manager.py", line 1542, in handle_loop
    self._result_dispatcher(recv_obj)
  File "/sgl-workspace/sglang/python/sglang/utils.py", line 507, in __call__
    return fn(obj)
           ^^^^^^^
  File "/sgl-workspace/sglang/python/sglang/srt/managers/tokenizer_manager.py", line 2040, in _handle_abort_req
    state = self.rid_to_state[recv_obj.rid]
            ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
KeyError: '2y19aYWWApO9jNsP55KvCQ'
```

python/sglang/srt/managers/tokenizer_manager.py
```diff
@@ -1168,6 +1168,8 @@ async def _wait_one_response(
                     ) in (
                         HTTPStatus.SERVICE_UNAVAILABLE,
                         HTTPStatus.INTERNAL_SERVER_ERROR,
+                        envs.SGLANG_ABORTED_REQUEST_STATUS_CODE.get(),
+                        envs.SGLANG_TIMED_OUT_REQUEST_STATUS_CODE.get(),
                     ):
```
JustinTong0323, line 1194: Using status code to classify the finish type is not a good design and we need a fix. Maybe we could use a more dedicated way to do so, like adding a field to finish_reason and pop here?

python/sglang/srt/managers/tokenizer_manager.py
hnyls2002: If that is a guard, then there are extra potential issues. Could you help locate that issue?

hnyls2002 (approved):

PR #18447, opened 2026-02-08
Fix modelopt FP8 create weights

danielafrimi:
`ModelOptFp8LinearMethod.create_weights() `was missing `input_size` and `output_size` parameters, causing type error

in commit `80b2b3207`, `LinearMethodBase.create_weights()` was updated to include `input_size` and `output_size` parameters. ModelOptFp8LinearMethod was never updated to match.

Fix 2: NemotronH is missing s`elf.quant_config `attribute
KV cache defaulted to bfloat16 instead of FP8, even when `hf_quant_config.json` specified kv_cache_quant_algo: "FP8". `NemotronHForCausalLM` didn't store `self.quant_config = quant_config`, so `model_runner.py` couldn't read `kv_cache_quant_algo` from the model.


b8zhong (approved):

python/sglang/srt/layers/quantization/modelopt_quant.py
```diff
@@ -461,6 +461,8 @@ def create_weights(
         layer: torch.nn.Module,
         input_size_per_partition: int,
         output_partition_sizes: List[int],
+        input_size: int,
```
Fridge003: Can we set these two arguments optional, so this change doesn't break other fp8 models?

Fridge003 (approved):

PR #18451, opened 2026-02-08
[AMD] Use aiter_dsv3_router_gemm kernel if number of experts <= 256.

amd-mvarjoka:
This change modified the router kernel selection with ROCm backend for models using DeepSeek model architecture. The Aiter kernel aiter_dsv3_router_gemm has been optimized for the DeepSeek model, but it is at the moment better to use the default kernel for the Kimi K2 model. The condition is modified to select the Aiter kernel for DeepSeek models and default branch for the Kimi K2 models using the number of experts. This is faster for Kimi K2 models. The model accuracy was also checked with a lm_eval task.

Below is performance numbers on an Instinct MI350X node with -tp 8:

Original:
```
Mean TPOT (ms):                          41.28     
Median TPOT (ms):                        40.77     
P99 TPOT (ms):                           44.12  
```

with the PR:
```
Mean TPOT (ms):                          32.87     
Median TPOT (ms):                        33.54     
P99 TPOT (ms):                           36.26     
```

There is lm_eval gsm8k test results for the original instruct model (moonshotai/Kimi-K2-Instruct):
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9492|±  |0.0060|
|     |       |strict-match    |     5|exact_match|↑  |0.9469|±  |0.0062|
```

And the AMD's MXFP4 quantized model (amd/Kimi-K2-Instruct-0905-MXFP4):
```
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9538|±  |0.0058|
|     |       |strict-match    |     5|exact_match|↑  |0.9538|±  |0.0058|
```


yichiche (approved):
Considering DSv3 has 256 experts and Kimi-k2 has 384, this modification should be fine. LGTM.

PR #18454, opened 2026-02-08
[Diff]: support SGLANG_TORCH_PROFILER_DIR environment variable for profiler log directory

Johnsonms:
## Summary
- Modified SGLDiffusionProfiler to check SGLANG_TORCH_PROFILER_DIR env var when log_dir is not explicitly specified
- If log_dir parameter is None (default), check environment variable first, then fall back to ./logs
- If log_dir is explicitly provided by caller, use it directly and skip environment check
- This allows users to control profiler output location via environment variable without code changes

## Changes
- Updated `python/sglang/multimodal_gen/runtime/utils/profiler.py` to add environment variable support

## Testing
Tested with `export SGLANG_TORCH_PROFILER_DIR=/scratch/johnson/profile/` and verified profiler traces are saved to the specified directory.


BBuf (approved):

Johnsonms:
> Fix lint

Done, Thanks @BBuf !

PR #18467, opened 2026-02-09
VLM: support passing  --mm-process-config for all models

edwingao28:
## Motivation

Fixes #14672.

The --mm-process-config CLI flag is parsed and stored in ServerArgs.mm_process_config, but never passed to the HuggingFace processor's __call__() for any model except Qwen VL (which had its own manual extraction). This means users setting e.g. --mm-process-config '{"image":{"max_pixels":5000000}}' saw no effect on InternVL, LLaVA,MiniCPM, or any other VLM.

A prior attempt (PR #14968) tried fixing this by flattening per-modality configs into a single top-level kwargs dict via kwargs.update(self.image_config). As noted by @JustinTong0323, this causes a parameter collision: when both image and video configs share a key (e.g. max_pixels), the later update() silently overwrites the earlier one.

This PR uses HuggingFace's built-in per-modality kwargs routing (images_kwargs, videos_kwargs, audio_kwargs) to keep each modality's config separate, directly resolving the collision issue.
## Modifications
- server_args.py: Add type validation for mm_process_config (top-level must be dict, each modality value must be dict). Placed before the dummy-model early return so validation runs unconditionally.
- base_processor.py: Extract image_config/video_config/audio_config in __init__(). Inject them as images_kwargs/videos_kwargs/audio_kwargs in process_mm_data() using setdefault().update() to preserve existing defaults (e.g. truncation=False for audio).
- qwen_vl.py: Remove duplicated config extraction (now inherited from base class). self.video_config usage in preprocess_video() still works via inheritance.
- ernie45_vl.py: Add images_kwargs/videos_kwargs injection to its overridden process_mm_data() (vision model, no audio).
- midashenglm.py: Add audio_kwargs config injection to its overridden process_mm_data() (audio-only model). Changed audio_kwargs init from = {} to setdefault() to support merging.
- multimodal_language_models.md: Remove "only qwen_vl supports this config" caveat.
- test/registered/core/test_mm_process_config.py: New test file covering server-args validation, per-modality kwargs injection, no-collision verification, and override processor regression tests.

## Checklist

## Review Process


python/sglang/srt/multimodal/processors/ernie45_vl.py
```diff
@@ -289,8 +289,12 @@ def process_mm_data(
         """
         if images:
             kwargs["images"] = images
+            if self.image_config:
```
yuan-luo, line 295: The fix is solid. Just one minor question, why this code segment doesn't appear in qwen_vl processor?

edwingao28: Qwen-VL doesn’t need that block in its own file because it does not override process_mm_data; it inherits the base path where images_kwargs/videos_kwargs/audio_kwargs are already injected (python/sglang/srt/multimodal/processors/base_processor.py).
  
Ernie4.5-VL overrides process_mm_data for custom postprocessing, so the we need to add same injection explicity otherwise it would bypass base injection.

edwingao28: @yuan-luo Do you think it would be good for us to merge this PR? I’d love your feedback!

anigrigoryan760-eng (approved):

anigrigoryan760-eng (approved):

edwingao28:
@JustinTong0323 @mickqian @yhyang201 @yuan-luo  This PR is ready for merge, CI failures are all unrelated flakes (XPU deepseek_ocr, client disconnects, cascade gates)

python/sglang/srt/server_args.py
```diff
@@ -745,6 +745,21 @@ def __post_init__(self):
         # Normalize load balancing defaults early (before dummy-model short-circuit).
         self._handle_load_balance_method()
 
+        # Validate mm_process_config before dummy-model early return.
```
mickqian, line 761: extract this to `_handle_multimodal()`

edwingao28: resolved @mickqian

mickqian (approved):

edwingao28:
Hey @mickqian could you take a look when you have a chance, current failed ci seems to be pre existing CI issues after the fix.

PR #18468, opened 2026-02-09
Tiny rename for spec related fileds.

hnyls2002:
The original `per_batch` is misleading and should be `per_req`.


hnyls2002:
It is just some string replacement, and basic tests are passed.

PR #18472, opened 2026-02-09
Fix Eagle draft metadata after MLP sync padding

Aphoh:
## Motivation

  Fix a correctness bug in EAGLE speculative draft decode when CUDA graphs are disabled and MLP sync padding is active.
  In this path, attention metadata can be initialized before prepare_mlp_sync_batch pads the batch, causing metadata/query shape mismatch and potential illegal memory access or corrupted outputs.

  ## Modifications

  - Updated python/sglang/srt/model_executor/model_runner.py in ModelRunner._forward_raw.
  - After forward_batch.prepare_mlp_sync_batch(self), added a guarded attention metadata re-init:
      - Condition: skip_attn_backend_init is True and _original_batch_size exists (padding happened).
      - Action: forward_batch.attn_backend.init_forward_metadata(forward_batch).

  This keeps metadata aligned with padded draft inputs in non-CUDA-graph EAGLE draft flows.

  ## Accuracy Tests

  - Not run in this change set.

  ## Benchmarking and Profiling

  - Not run in this change set.
  - Change is gated and only triggers when:
      - MLP sync path is active,
      - attention init was skipped,
      - and batch padding actually occurred.

  ## Checklist

  - [x] Format your code according to the Format code with pre-commit (https://docs.sglang.io/developer_guide/contribution_guide.html#format-code-with-pre-commit).
  - [ ] Add unit tests according to the Run and add unit tests (https://docs.sglang.io/developer_guide/contribution_guide.html#run-and-add-unit-tests).
  - [ ] Update documentation according to Write documentations (https://docs.sglang.io/developer_guide/contribution_guide.html#write-documentations).
  - [ ] Provide accuracy and speed benchmark results according to Test the accuracy (https://docs.sglang.io/developer_guide/contribution_guide.html#test-the-accuracy) and Benchmark the speed (https://
    docs.sglang.io/developer_guide/contribution_guide.html#benchmark-the-speed).
  - [x] Follow the SGLang code style guidance (https://docs.sglang.io/developer_guide/contribution_guide.html#code-style-guidance).

### DIFF
--- a/python/sglang/srt/model_executor/model_runner.py
+++ b/python/sglang/srt/model_executor/model_runner.py
@@ -2403,6 +2403,14 @@ class ModelRunner(ModelRunnerKVCacheMixin):
         # For MLP sync
         if forward_batch.global_num_tokens_cpu is not None:
             forward_batch.prepare_mlp_sync_batch(self)
+            # If MLP sync padded this batch and attention metadata was initialized
+            # earlier (e.g. EAGLE draft path with skip_attn_backend_init=True),
+            # re-init metadata so it matches the padded batch shape.
+            if (
+                skip_attn_backend_init
+                and getattr(forward_batch, "_original_batch_size", None) is not None
+            ):
+                forward_batch.attn_backend.init_forward_metadata(forward_batch)
         else:
             forward_batch.prepare_attn_tp_scatter_input(self)
 


Qiaolin-Yu (commented):
I think ideally we shouldn't init this twice. Instead we should fix the issue inside the attention backend.

PR #18489, opened 2026-02-09
[MODEL] Adding Support for Qwen3.5 Models

zju-stu-lizheng:
## Purpose

This PR adds model support for the upcoming Qwen3.5 models, including both dense and MoE variants.

Special thanks to @cao1zhg, @yizhang2077, and @attack204 for their review, and to @hnyls2002 and @mickqian from the SGLang team for their valuable review and support.

Reference HF implementation - [huggingface/transformers#43830](https://github.com/huggingface/transformers/pull/43830)

## Main Changes
1. **Model Support**: Added support for two new model classes:
   - `Qwen3_5MoeForConditionalGeneration`
   - `Qwen3_5ForConditionalGeneration`

2. **Attention Layer**: Introduced a new linear attention layer `Qwen3_5GatedDeltaNet` to optimize the `fix_qkvz` reordering operation.

## Launch Server

Hardware setup: 8 × H200 GPUs

### Without MTP 

```
TP_SIZE=8

    --model $vlmoe \
    --tp-size ${TP_SIZE} \
    --enable-multimodal \
    --max-mamba-cache-size 128 \
    --max-running-requests 128 \
    --chunked-prefill-size 2048 \
    --mamba-ssm-dtype bfloat16
```

### With MTP

Add the following arguments to the above command:
```
    --speculative-algo NEXTN    \
    --speculative-num-steps 3     \
    --speculative-eagle-topk 1     \
```

Check our cookbook for detailed deployment instruction: https://cookbook.sglang.io/autoregressive/Qwen/Qwen3.5


python/sglang/srt/configs/qwen3_5.py
```diff
@@ -0,0 +1,113 @@
...
+class Qwen3_5TextConfig(Qwen3NextConfig):
+    model_type = "qwen3_5_text"
+    base_config_key = "text_config"
+
+    def __init__(
+        self,
+        **kwargs,
+    ):
+        super().__init__(**kwargs)
+        if self.rope_scaling is None:
```
ovowei: Unlike qwen3next, qwen3_5config does not have the rope_scaling field but has the rope_parameters field. My test found that setting self.rope_scaling={} overwrites rope_parameters. Cause the File "sglang/python/sglang/SRT/the layers/rotary_embedding py", line 3063, in get_rope 
raise ValueError("Unknown RoPE scaling type") 
ValueError: Unknown RoPE scaling type error. The fix is to remove these two lines

Huixxi:
Emm, may I ask why qwen3.5's moe sparse block reuse qwen2's not qwen3's? @zju-stu-lizheng

Cppowboy:
Which tool call paser should be used for qwen3.5 series model?

PR #18491, opened 2026-02-09
Add cu130 extra for simplified CUDA 13 installation

mmangkad:
## Motivation

Currently, installing SGLang on CUDA 13 requires a 3-step manual process. With this change, installation is now simplified to:

```
Command                           Behavior
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
uv pip install -e python          Installs torch with cu129 (default)
uv pip install -e python[cu130]   Installs torch with cu130 + correct sgl-kernel wheel (and arch) automatically
```

Verified on both x86_64 and aarch64 machines.

Note: This only works with uv. Without uv, `pip install` still pulls torch CPU from PyPI and still requires manually adding `--extra-index-url` - thus the slight change in the Method 1 name.

cc @Fridge003 what do you think?

## Checklist

## Review Process

4. Trigger CI tests with [comments](https://docs.sglang.io/developer_guide/contribution_guide.html#how-to-trigger-ci-tests) or contact authorized users to do so.

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -69,13 +69,12 @@ dependencies = [
   "torchao==0.9.0",
   "torchaudio==2.9.1",
   "torchcodec==0.8.0 ; sys_platform != 'linux' or (sys_platform == 'linux' and platform_machine != 'aarch64' and platform_machine != 'arm64' and platform_machine != 'armv7l')", # torchcodec does not exist in those systems. If not provided, transformer will use torchvision instead by default.
-  "torchvision",
+  "torchvision==0.24.1",
   "tqdm",
   "transformers==4.57.1",
   "uvicorn",
   "uvloop",
   "xgrammar==0.1.27",
-
   "grpcio==1.75.1", # keep it align with compile_proto.py
   "grpcio-tools==1.75.1", # keep it align with compile_proto.py
   "grpcio-reflection==1.75.1", # required by srt/entrypoints/grpc_server.py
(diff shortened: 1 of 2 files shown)


python/pyproject.toml
```diff
@@ -92,14 +91,42 @@ name = "torch-cu129"
...
+  { url = "https://github.com/sgl-project/whl/releases/download/v0.3.21/sgl_kernel-0.3.21+cu130-cp310-abi3-manylinux2014_x86_64.whl", extra = "cu130", marker = "platform_machine == 'x86_64'" },
+  { url = "https://github.com/sgl-project/whl/releases/download/v0.3.21/sgl_kernel-0.3.21+cu130-cp310-abi3-manylinux2014_aarch64.whl", extra = "cu130", marker = "platform_machine == 'aarch64'" }
 ]
 
 [project.optional-dependencies]
+cu130 = [
+  "torch",
+  "torchaudio",
+  "torchvision",
+  "sgl-kernel"
```
Fridge003, line 125: Also add "cuda-python" (it defaults to be the latest cuda 13 version)

python/pyproject.toml
```diff
@@ -92,14 +91,42 @@ name = "torch-cu129"
...
+]
+
+sgl-kernel = [
+  { url = "https://github.com/sgl-project/whl/releases/download/v0.3.21/sgl_kernel-0.3.21+cu130-cp310-abi3-manylinux2014_x86_64.whl", extra = "cu130", marker = "platform_machine == 'x86_64'" },
+  { url = "https://github.com/sgl-project/whl/releases/download/v0.3.21/sgl_kernel-0.3.21+cu130-cp310-abi3-manylinux2014_aarch64.whl", extra = "cu130", marker = "platform_machine == 'aarch64'" }
 ]
 
 [project.optional-dependencies]
+cu130 = [
+  "torch",
```
Fridge003, line 122: Pin it to 2.9.1, since the latest stable torch version is 2.10

python/pyproject.toml
```diff
@@ -69,13 +69,12 @@ dependencies = [
   "torchao==0.9.0",
   "torchaudio==2.9.1",
   "torchcodec==0.8.0 ; sys_platform != 'linux' or (sys_platform == 'linux' and platform_machine != 'aarch64' and platform_machine != 'arm64' and platform_machine != 'armv7l')", # torchcodec does not exist in those systems. If not provided, transformer will use torchvision instead by default.
-  "torchvision",
+  "torchvision==0.24.1",
```
Fridge003, line 72: Is it OK to fix this torchvision version?

python/pyproject.toml
```diff
@@ -92,14 +91,42 @@ name = "torch-cu129"
...
+
 [tool.uv.sources]
 torch = [
-  { index = "pypi", marker = "platform_machine == 'x86_64'"},
-  { index = "torch-cu129", marker = "platform_machine == 'aarch64'"},
+  { index = "torch-cu130", extra = "cu130" },
+  { index = "torch-cu129", marker = "extra != 'cu130'" }
+]
+
+torchaudio = [
```
Fridge003, line 105: Can we remove torchaudio and torchvision here? Since they can be installed along with torch

python/pyproject.toml
mmangkad: Not necessary - torch version is inherited from the main dependencies. The `[cu130]` extra doesn't need to repeat it.

python/pyproject.toml
mmangkad: Yes, it should be.

python/pyproject.toml
mmangkad: Still needed. I tested without these and they install CPU versions. The explicit sources ensure they use the correct cuda versions.

python/pyproject.toml
mmangkad: Creates version conflict with base `cuda-python==12.9` from the main dependencies. uv cannot resolve both versions. This would need unpinning the default.

mmangkad: Ok what I tried earlier is pinning it to `13.1` like in the Dockerfile, which created the conflict. Unpinned works but still resolves to `12.9` from main deps, so it doesn't add value to include it.

Issue #18514, opened 2026-02-10
[Bug] DeepGEMM shape mismatch error with FP8 MTP layer under PD disagg setting for nvfp4 models


wenscarl:
### Checklist

- [ ] I searched related issues but found no solution.
- [ ] The bug persists in the latest version.
- [ ] Issues without environment info and a minimal reproducible demo are hard to resolve and may receive no feedback.
- [ ] If this is not a bug report but a general question, please start a discussion at https://github.com/sgl-project/sglang/discussions. Otherwise, it will be closed.
- [ ] Please use English. Otherwise, it will be closed.

### Describe the bug

Under PD disaggregation with NVFP4 quantization, in some low-latency deployments the decode node may run without expert parallelism. An example configuration is shown below:
```
envs:
    SGLANG_NVFP4_CKPT_FP8_NEXTN_MOE: "1"
    SGLANG_ENABLE_SPEC_V2: "1"
    SGLANG_DEEPEP_NUM_MAX_DISPATCH_TOKENS_PER_RANK: "768"   
 decode:
      disaggregation-mode: "decode"
      served-model-name: "deepseek-ai/DeepSeek-R1"
      prefill-round-robin-balance: true
      trust-remote-code: true
      disable-radix-cache: true
      kv-cache-dtype: "fp8_e4m3"
      attention-backend: "trtllm_mla"
      quantization: "modelopt_fp4"
      moe-runner-backend: "flashinfer_trtllm"
      disaggregation-bootstrap-port: 30001
      stream-interval: 10
      watchdog-timeout: 1000000
      context-length: 10000
      mem-fraction-static: 0.8
      chunked-prefill-size: 8192
      cuda-graph-max-bs: 256
      disable-shared-experts-fusion: true
      scheduler-recv-interval: 10
      tensor-parallel-size: 4
      expert-parallel-size: 1
      data-parallel-size: 1
      speculative-algorithm: "EAGLE"
      speculative-num-steps: 2
      speculative-eagle-topk: 1
      speculative-num-draft-tokens: 3
```
With SGLANG_NVFP4_CKPT_FP8_NEXTN_MOE=1, the MTP layer is expected to run in FP8. However, execution fails with the following runtime error during MoE execution:
```
  File "/sgl-workspace/sglang/python/sglang/srt/layers/moe/ep_moe/layer.py", line 180, in forward_impl
    return super().forward_impl(
           ^^^^^^^^^^^^^^^^^^^^^
[...]

PR #18515, opened 2026-02-10
Fix Fp8 MTP layer a2a backend without EP.

wenscarl:
## Motivation

Context: https://github.com/sgl-project/sglang/issues/18514

## Checklist

## Review Process

### DIFF
--- a/python/sglang/srt/server_args.py
+++ b/python/sglang/srt/server_args.py
@@ -1308,20 +1308,10 @@ class ServerArgs:
                 ):
                     if envs.SGLANG_NVFP4_CKPT_FP8_NEXTN_MOE.get():
                         self.speculative_moe_runner_backend = "deep_gemm"
-                        # deepep only supports expert parallelism
-                        if self.ep_size > 1:
-                            self.speculative_moe_a2a_backend = "deepep"
-                            logger.info(
-                                "Use deep_gemm moe runner and deepep a2a backend for bf16 nextn layer in deepseek fp4 checkpoint."
-                            )
-                        else:
-                            self.speculative_moe_a2a_backend = "none"
-                            logger.warning(
-                                "Expert parallelism is not enabled (ep_size=1), falling back to 'none' a2a backend instead of 'deepep' for speculative MoE."
-                            )
-                            logger.info(
-                                "Use deep_gemm moe runner and none a2a backend for bf16 nextn layer in deepseek fp4 checkpoint."
-                            )
+                        self.speculative_moe_a2a_backend = "deepep"
+                        logger.info(
+                            "Use deep_gemm moe runner and deepep a2a backend for bf16 nextn layer in deepseek fp4 checkpoint."
+                        )
                     else:
                         self.speculative_moe_runner_backend = "triton"
                         self.speculative_moe_a2a_backend = "none"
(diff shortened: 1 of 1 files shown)


python/sglang/srt/server_args.py
```diff
@@ -1308,10 +1308,20 @@ def _handle_model_specific_adjustments(self):
                 ):
                     if envs.SGLANG_NVFP4_CKPT_FP8_NEXTN_MOE.get():
                         self.speculative_moe_runner_backend = "deep_gemm"
-                        self.speculative_moe_a2a_backend = "deepep"
-                        logger.info(
-                            "Use deep_gemm moe runner and deepep a2a backend for bf16 nextn layer in deepseek fp4 checkpoint."
-                        )
+                        # deepep only supports expert parallelism
+                        if self.ep_size > 1:
+                            self.speculative_moe_a2a_backend = "deepep"
```
nvpohanh: @wenscarl What if users already set this in their server_args? Do we override it?

python/sglang/srt/server_args.py
```diff
@@ -1308,10 +1308,20 @@ def _handle_model_specific_adjustments(self):
...
+                            logger.info(
+                                "Use deep_gemm moe runner and deepep a2a backend for bf16 nextn layer in deepseek fp4 checkpoint."
+                            )
+                        else:
+                            self.speculative_moe_a2a_backend = "none"
+                            logger.warning(
```
nvpohanh: "warning" usually means that users must take action to correct it and follow the recommended flow. But here the warning message isn't clear about what users should change to suppress the  warning.

python/sglang/srt/server_args.py
```diff
@@ -1308,10 +1308,20 @@ def _handle_model_specific_adjustments(self):
...
+                                "Expert parallelism is not enabled (ep_size=1), falling back to 'none' a2a backend instead of 'deepep' for speculative MoE."
+                            )
+                            logger.info(
```
nvpohanh: so when `self.speculative_moe_a2a_backend = "none"`, the `SGLANG_NVFP4_CKPT_FP8_NEXTN_MOE` basically has no effect? In other words, draft model will run in BF16 instead of FP8 when TP8 is used and SGLANG_NVFP4_CKPT_FP8_NEXTN_MOE is set to 1?

Fridge003 (commented):
@nvpohanh @wenscarl 
Seems a better way is to throw error when the user sets `SGLANG_NVFP4_CKPT_FP8_NEXTN_MOE=1`, but not setting 
```
self.speculative_moe_runner_backend = "deep_gemm"
self.speculative_moe_a2a_backend = "deepep"
self.ep_size > 1
```

If tp is runnable under fp8 mtp moe, we can mitigate this assertion

nvpohanh (approved):

python/sglang/srt/server_args.py
wenscarl: We don't override it but just error out.

python/sglang/srt/server_args.py
wenscarl: Fixed by erroring out if the config is not supported yet.

Fridge003 (approved):

PR #18516, opened 2026-02-10
[Diffusion] Fix Ulysses SP for QwenImageEdit: shard text/cond/noisy tokens correctly

Jasen2201:
### Motivation

Running `QwenImageEdit` (or `QwenImageEditPlus`) with `ulysses_degree >= 2` crashes or produces corrupted output. The root cause is that only the noisy latent tokens are sharded across SP ranks, while **condition image tokens**, **text embeddings**, and their corresponding **RoPE** and **modulate_index** remain un-sharded. This creates shape mismatches in attention and modulation, leading to either a runtime crash or visually broken outputs (mosaic artifacts in the lower half of the generated image).

### Problem Analysis

In the Ulysses SP paradigm, an all-to-all exchange maps `[B, S_local, H, D] → [B, S_global, H_local, D]`. This assumes **all** sequence tokens entering the DiT are already partitioned across SP ranks. For `QwenImageEdit` I2I tasks, the DiT input is `torch.cat([noisy_local, cond_image_local], dim=1)`, plus text tokens fed separately.

**Shape flow comparison (per-rank, 768x1024, ulysses=2):**

image

### Test

**Model:** `Qwen-Image-Edit-2511` (QwenImageEditPlus)  
**Resolution:** 768 x 1024  
**Config:** `ulysses_degree=2`, `num_gpus=2`, `guidance_scale=1`, `seed=42`  

**Test script:**

```python

def main():
        model_path="<path-to>/Qwen-Image-Edit-2511",
        num_gpus=2,
        ulysses_degree=2,
        tp_size=1,
        enable_torch_compile=False,
        dit_cpu_offload=False,
        text_encoder_cpu_offload=False,
        vae_cpu_offload=False,
        image_encoder_cpu_offload=False,
        image_encoder_precision="bf16",
        vae_precision="bf16",
        use_fsdp_inference=False,
    )

    image = generator.generate(
        sampling_params_kwargs=dict(
            prompt="make the clothes to red",
            image_path="<path-to>/768x1024.png",
            output_path="output/",
            save_output=True,
            height=1024,
            width=768,
            num_inference_steps=40,
            guidance_scale=1,
            seed=42,
        )
    )
    generator.shutdown()

if __name__ == "__main__":
    main()
```

**Input image** :
768x1024

**Results image** (prompt: "make the clothes to red"):
make_the_clothes_to_red_20260210-024537_10fd71fb

### DIFF
--- a/python/sglang/multimodal_gen/configs/pipeline_configs/qwen_image.py
+++ b/python/sglang/multimodal_gen/configs/pipeline_configs/qwen_image.py
@@ -184,7 +184,14 @@ class QwenImagePipelineConfig(ImagePipelineConfig):
                 )
             ]
         ] * batch_size
-        txt_seq_lens = [prompt_embeds[0].shape[1]]
+
+        # Use original txt seq len (before SP sharding) for RoPE generation
+        # so each rank gets the correct global position embeddings
+        original_txt_seq_len = getattr(batch, "original_txt_seq_len", None)
+        if original_txt_seq_len is not None:
+            txt_seq_lens = [original_txt_seq_len]
+        else:
+            txt_seq_lens = [prompt_embeds[0].shape[1]]
 
         freqs_cis = self.get_freqs_cis(
             img_shapes, txt_seq_lens, rotary_emb, device, dtype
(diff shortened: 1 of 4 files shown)


hnyls2002:
Hi @Jasen2201, I believe this was largely superseded by #20679, which added _shard_qwen_edit_img_cache_for_sp to shard both the noisy and condition image RoPE caches for Ulysses SP. The text-token sharding part wasn't taken (main keeps txt_cache unsharded by design). Closing as superseded; please reopen if the current code still misses a case you hit.

PR #18518, opened 2026-02-10
Fix: redact API secrets in ServerArgs repr and request header logs

pockers21:
## Motivation

This PR mitigates potential secret exposure in logs, based on #18256.

- `ServerArgs` repr may include API secrets (`api_key`, `admin_api_key`).
- Request logging may include raw credential-like header values when headers are whitelisted.

## Modifications

- Hide `api_key` and `admin_api_key` from `ServerArgs` repr output.
- Mask sensitive request header values before logging (e.g., `authorization`, `proxy-authorization`, `x-api-key`, `api-key`).
- Add regression tests for:
  - `ServerArgs` repr secret redaction.

## Accuracy Tests

N/A. This PR does not change model forward/inference output logic.

## Benchmarking and Profiling

N/A. This PR does not change performance-critical inference paths.

## Checklist

- [ ] Update documentation according to [Write documentations](https://docs.sglang.io/developer_guide/contribution_guide.html#write-documentations). (Not required for this internal logging/security hardening change.)


JustinTong0323 (approved):

sethforprivacy:
Still reproducible on current HEAD (`0.0.0.dev1+gd91c3682b`) — `ServerArgs` has
no `__repr__` override, no `repr=False`, and no redaction anywhere in the class,
so the startup banner still prints `api_key='…'` verbatim.

Production data point, in case it helps justify a merge: we run this across an
8-node DGX Spark fleet (4× TP=2 clusters). Every boot writes the configured key
set into the launcher's captured log at the ambient umask, so the keys sit
world-readable on disk and persist across restarts. We just finished a fleet
sweep to chmod and redact them after the fact — this PR is the thing that would
have prevented it.

The multi-key case is the sharp edge. `--api-key` is a single string, so a
comma-separated set (we serve six shared keys this way, cf. #36666) lands in the
banner as **one line containing every key at once** — one leaked log line
compromises the entire key set, not one credential.

One review note on the approach, from having shipped the equivalent patch for
vLLM internally:

`dataclasses.field(repr=False)` removes the field from the repr entirely, which
also removes the operator's ability to tell from the log whether auth is
configured at all — and, in the multi-key case, whether the intended N keys
actually parsed. A count-preserving marker keeps that diagnostic while exposing
nothing:

```
api_key='<redacted:6 value(s)>'
```

That distinction earned its keep for us: repeating `--api-key` overwrites rather
than appends, so the count in the log is the only cheap confirmation that a
single flag carried the whole set. Not a blocker on this PR — dropping the field
is strictly better than printing it, and the current test would just need to
assert the marker instead of `assertNotIn("api_key=", …)`.

Happy to send the count-preserving variant as a follow-up if maintainers prefer
to land the redaction as-is first.

pockers21:
> Still reproducible on current HEAD (`0.0.0.dev1+gd91c3682b`) — `ServerArgs` has no `__repr__` override, no `repr=False`, and no redaction anywhere in the class, so the startup banner still prints `api_key='…'` verbatim.
> 
> Production data point, in case it helps justify a merge: we run this across an 8-node DGX Spark fleet (4× TP=2 clusters). Every boot writes the configured key set into the launcher's captured log at the ambient umask, so the keys sit world-readable on disk and persist across restarts. We just finished a fleet sweep to chmod and redact them after the fact — this PR is the thing that would have prevented it.
> 
> The multi-key case is the sharp edge. `--api-key` is a single string, so a comma-separated set (we serve six shared keys this way, cf. #36666) lands in the banner as **one line containing every key at once** — one leaked log line compromises the entire key set, not one credential.
> 
> One review note on the approach, from having shipped the equivalent patch for vLLM internally:
> 
> `dataclasses.field(repr=False)` removes the field from the repr entirely, which also removes the operator's ability to tell from the log whether auth is configured at all — and, in the multi-key case, whether the intended N keys actually parsed. A count-preserving marker keeps that diagnostic while exposing nothing:
> 
> ```
> api_key='<redacted:6 value(s)>'
> ```
> 
> That distinction earned its keep for us: repeating `--api-key` overwrites rather than appends, so the count in the log is the only cheap confirmation that a single flag carried the whole set. Not a blocker on this PR — dropping the field is strictly better than printing it, and the current test would just need to assert the marker instead of `assertNotIn("api_key=", …)`.
> 
> Happy to send the count-preserving variant as a follow-up if maintainers prefer to land the redaction as-is first.

Thanks @sethforprivacy — really appreciate the production details and the multi-key suggestion.

This PR is quite old now, and the `ServerArgs` implementation has changed a lot since then. #39449 is handling the same credential leak on the current code paths.

The count-preserving marker is a good idea. @avifenesh, could you consider adding it to #39449?

avifenesh:
@pockers21 @sethforprivacy done in #39449 (commit 3f78acaf53): the record-side readbacks now publish `<redacted:1 value>` / `<redacted:N values>`, counting the values the parsed field holds, so the line still says whether auth is configured and how many keys the flag fanned out into. On current main `--api-key a,b,c` parses as one string, so it reads `1 value` there until #36666 turns the field into a list; the launch-command echo keeps the plain marker because the argv does not know how a value parsed. Details and tests on #39449.

PR #18567, opened 2026-02-10
Fix the IPV6 issue in get_zmq_socket_host function

hhzguo:
## Motivation

When run EPD on node with IPV6 address, the zmq for IPV6 has the following issues:
1.  ipv6 address is not handled correctly in get_zmq_socket_host function
2. encode server doesn't handle IPV6 correctly which causes the communication problems between encode server and prefill server

## Modifications

1. python/sglang/srt/utils/common.py
     get_zmq_socket_host
2. python/sglang/srt/disaggregation/encode_server.py
    MMEncoder:_send
3. Add test_zmq unittest

## Accuracy Tests

image

## Benchmarking and Profiling
1 encoder server on node1, 1 prefill server with 4 B200 GPU on node2, 1 decode server with 4 B200 GPU on node2

python -m sglang.bench_serving \
    --dataset-name  mmmu \
    --port 28000 \
    --backend vllm-chat \
    --request-rate 20 \
    --model /data03/models/Qwen3-VL-235B-A22B-Instruct \
    --num-prompts 800 \
    --image-count 500 \
    --max-concurrency 200 \
    --image-resolution 1080p

image
## Checklist

## Review Process

4. Get approvals from [CODEOWNERS](https://github.com/sgl-project/sglang/blob/main/.github/CODEOWNERS) and other reviewers.
5. Trigger CI tests with [comments](https://docs.sglang.io/developer_guide/contribution_guide.html#how-to-trigger-ci-tests) or contact authorized users to do so.
6. After green CI and required approvals, ask Merge Oncalls to merge.


python/sglang/srt/utils/common.py
```diff
@@ -1456,12 +1454,39 @@ def get_zmq_socket_on_host(
...
+        except zmq.ZMQError as e:
+            last_error = e
+            if e.errno != zmq.EADDRINUSE:
+                raise
+            new_socket.close(linger=0)
+            if retry == retry_count:
+                break
+
+            time.sleep(delay)
+    raise RuntimeError("Failed to bind socket after retries") from last_error
```
ShangmingCai, line 1501: Please don't change this, I think we have a util called `maybe_wrap_ipv6_address`, can you try to solve this problem through that util?

hhzguo: Shangming, sure, I will take a look at the util lib, of course we should use the available functions if they already there. Thank you for let me know and I will do it and test.

python/sglang/srt/utils/common.py
VanyaGlazunov, line 1448: Hello, @ShangmingCai and @hhzguo I encountered what is described by the contributor recently and I have a suggestion. Maybe we can use this logic in get_zmq_socket_on_host too?

test/manual/test_zmq.py
```diff
@@ -0,0 +1,78 @@
+import socket
+import unittest
+
+import zmq
+
+from sglang.srt.utils import get_zmq_socket_on_host
+
+
+def get_routable_ip_strings():
```
hnyls2002, line 9: Why not put it inside `registered`, this is only cpu UT.

hhzguo:
[hnyls2002] checked in big PR which covers the changers in this PR

PR #18582, opened 2026-02-11
Add subprocess liveness monitor to detect scheduler crashes

### LINKED ISSUE #18421: [Bug] Scheduler subprocess crash due to NCCL timeout remains undetected, causing zombie service state
### Checklist

### Describe the bug

When the SGLang scheduler subprocess crashes due to NCCL timeout (or other fatal errors that trigger
 process termination at the C++ level), the main HTTP server process continues running but cannot  process inference requests. While health checks eventually detect the failure, there is a significant detection delay (default 20 seconds) during which:
  - The service appears operational and accepts new requests
  - All new inference requests hang indefinitely without error responses
  - Orchestration systems (Kubernetes, load balancers) cannot detect the failure
  - No automatic recovery is triggered

This delayed failure detection creates a critical "zombie service" window where the system silently
fails, resulting in poor user experience and prolonged downtime before automated recovery can begin.

## Root Cause Analysis

The issue stems from three architectural gaps:

### 1. NCCL Timeout Terminates Process via Unhandled C++ Exception

When NCCL operations timeout, PyTorch's C++ code terminates the scheduler process
 before Python exception handlers can execute:

```
  [Rank 5] Watchdog caught collective operation timeout: WorkNCCL(SeqNum=549519, OpType=ALLREDUCE,
  NumelIn=83886080, NumelOut=83886080, Timeout(ms)=600000) ran for 600084 milliseconds before timing out.

  [rank5]:[E205 16:38:41.072031170 ProcessGroupNCCL.cpp:758] [Rank 5] To avoid data inconsistency, we are
  taking the entire process down.

  Fatal Python error: Aborted
```

  **What happens internally** (based on PyTorch source code):
  1. NCCL watchdog detects timeout
  2. PyTorch logs: "To avoid data inconsistency, we are taking the entire process down"
  3. `std::rethrow_exception()` is called ([ProcessGroupNCCL.cpp:765](https://github.com/pytorch/pytorch/blob/main/torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp#L765))
  4. Unhandled C++ exception triggers `std::terminate()`
  5. Process terminates immediately

  The well-designed exception handler in `scheduler.py:2962-2965` correctly sends SIGQUIT to the parent process for normal Python exceptions:

  ```python
  except Exception:
      traceback = get_exception_traceback()
      logger.error(f"Scheduler hit an exception: {traceback}")
      parent_process.send_signal(signal.SIGQUIT)  # This works for Python exceptions
```

**However**, when NCCL timeout occurs, the C++ exception handling terminates the process before Python code can execute:
  - Process terminates via `std::terminate()` (triggered by unhandled C++ exception)
  - Bypasses all Python exception handlers
  - Never reaches the `except Exception:` block
  - Never sends SIGQUIT to parent process

This is not a bug in SGLang's exception handling design—it's a fundamental limitation when subprocess death is triggered by C++-level exception handling rather than Python exceptions.

### 2. No Subprocess Liveness Monitoring

  Unlike the disaggregated prefill/decode setup (#13657), standard deployments lack a monitoring thread that checks `scheduler_proc.is_alive()`. The main process relies entirely on signal-based notification (SIGQUIT), which fails when the subprocess is killed via std::abort().

 ### 3. Health Checks Only Verify Message Reception

The health check endpoint (`http_server.py:452-528`) only validates `last_receive_tstamp` from the request queue, not the actual scheduler process state. When the scheduler crashes, health checks wait for `HEALTH_CHECK_TIMEOUT` (default 20 seconds) before failing, creating a significant detection gap.

###  Production Consequences:
  - Detection Delay: Health checks take 20 seconds (by default) to fail after scheduler crash
  - Silent Failures During Window: New requests accepted but never processed during detection period
  - Delayed Recovery: Kubernetes/orchestration systems wait for health check failures before restarting
  pods
  - Cumulative Downtime: 20s detection + container restart time = prolonged unavailability
  - Users experience hanging requests with no error messages
  - Difficult to diagnose without manual inspection of process states

### Reproduction

While challenging to reproduce deterministically, this occurs in production scenarios with:

  1. Multi-GPU distributed inference deployments
  2. Any fatal error that triggers std::abort() and bypasses Python exception handling
  3. NCCL timeout thresholds set too low for workload
  4. Network instability or high inter-GPU communication latency

Related Issues:
  - #17941 - Pipeline parallelism hangs with NCCL timeout (distributed communication context)
  - #12779 - Disaggregated prefill node death detection (subprocess monitoring precedent)
  - #12731, #13796 - Health check limitations and bypasses

### Environment

H20
0.5.8

### PR DESCRIPTION
**Pull request by Simon-Li**
## Summary

Adds a subprocess liveness watchdog to detect scheduler crashes that bypass Python exception handlers (e.g., NCCL timeout causing C++ `std::terminate()`).

**Problem**: When a scheduler subprocess crashes at the C++ level, `std::terminate()` runs before Python exception handlers, leaving the main process unaware. The service continues accepting requests but cannot process them ("zombie service").

**Solution**: A daemon thread in `Engine` polls `proc.is_alive()` every 1 second. On detecting an abnormal exit (`exitcode != 0`), it sends `SIGQUIT` to trigger existing cleanup infrastructure.

```
Engine (Main Process)
  │
  └── SubprocessWatchdog (daemon thread)
        polls is_alive() every 1s
        on crash (exitcode != 0): os.kill(SIGQUIT)
        │
        ├── scheduler_0
        ├── scheduler_1
        └── detokenizer
```

## Changes

| File | Change |
|------|--------|
| `python/sglang/srt/utils/watchdog.py` | +63 lines: `SubprocessWatchdog` class |
| `python/sglang/srt/entrypoints/engine.py` | +24 lines: create/store/stop watchdog |
| `python/sglang/srt/entrypoints/http_server.py` | +1 line: unpack return value |
| `python/sglang/srt/ray/http_server.py` | +1 line: unpack return value |
| `test/registered/unit/utils/test_subprocess_monitor.py` | +213 lines: 6 test cases |

**Note**: Zero changes to `tokenizer_manager.py`.

## Design Decisions

1. **Placed in `watchdog.py`**: Reuses the same daemon thread + SIGQUIT pattern as existing `WatchdogRaw`
2. **Owned by `Engine`**: All process handles available in `_launch_subprocesses()`, no cross-module passing needed
3. **Direct SIGQUIT**: Leverages existing signal handling infrastructure, thread-safe

## Edge Cases Handled

- Normal exit (`exitcode == 0`) does NOT trigger SIGQUIT
- `shutdown()` stops watchdog before killing process tree (prevents false positives)
- Empty process list handled gracefully
- Multi-node: non-zero rank nodes get `None` watchdog
- Ray backend: only monitors detokenizer (schedulers are actors)

## Test Plan

- [x] `test_healthy_processes_no_callback` - healthy processes don't trigger
- [x] `test_crashed_process_triggers_sigquit` - delayed crash detected
- [x] `test_immediate_crash_detection` - immediate crash detected  
- [x] `test_multiple_processes_one_crashes` - one crash in multi-process triggers
- [x] `test_empty_processes_list` - empty list handled
- [x] `test_normal_exit_no_sigquit` - normal exit (exitcode=0) doesn't trigger

Fixes #18421

### CODE REVIEW

File: test/registered/unit/utils/test_subprocess_watchdog.py [hunk 1]
```diff
@@ -0,0 +1,152 @@
...
+def crashing_worker():
+    os._exit(1)
+
+
+def slow_crash_worker(delay: float = 0.5):
+    time.sleep(delay)
+    os._exit(42)
+
+
+class TestSubprocessWatchdog(unittest.TestCase):
```
hnyls2002: Register with `CustomTestcase`

hnyls2002 (changes requested):

python/sglang/srt/entrypoints/engine.py
```diff
@@ -606,11 +612,17 @@ def _launch_subprocesses(
         run_scheduler_process_func: Callable,
         run_detokenizer_process_func: Callable,
         port_args: Optional[PortArgs] = None,
-    ) -> Tuple[TokenizerManager, TemplateManager, PortArgs, SchedulerInitResult]:
+    ) -> Tuple[
+        TokenizerManager,
+        TemplateManager,
+        PortArgs,
+        SchedulerInitResult,
+        Optional["SubprocessWatchdog"],
```
hnyls2002: from future import annotation instead of import this with `"`

python/sglang/srt/entrypoints/http_server.py
```diff
@@ -2166,7 +2166,7 @@ def launch_server(
     2. Inter-process communication is done through IPC (each process uses a different port) via the ZMQ library.
     """
     # Launch subprocesses
-    tokenizer_manager, template_manager, port_args, scheduler_init_result = (
+    tokenizer_manager, template_manager, port_args, scheduler_init_result, _ = (
```
hnyls2002: Why discard the reference of watch dog? When the server receives a `SIGQUIT`, I think you also need to stop the watch dog.

python/sglang/srt/entrypoints/engine.py
```diff
@@ -111,6 +112,7 @@ class SchedulerInitResult:
     """Result from launching schedulers."""
 
     scheduler_infos: List[Dict[str, Any]]
+    scheduler_procs: List = dataclasses.field(default_factory=list)
```
hnyls2002: This should not be kept in the init result. It is only used in the watch dog building steps.

python/sglang/srt/entrypoints/engine.py
```diff
@@ -687,15 +701,32 @@ def _launch_subprocesses(
             "max_req_input_len"
         ]
 
+        # Set up subprocess liveness watchdog to detect crashes
+        # (e.g., NCCL timeout causing C++ std::terminate() before Python can handle it)
+        processes = list(scheduler_init_result.scheduler_procs)
+        names = [f"scheduler_{i}" for i in range(len(processes))]
+        if detoken_proc is not None:
```
hnyls2002: This is never None.

hnyls2002 (changes requested):

hnyls2002 (approved):

PR #18596, opened 2026-02-11
fix: skip PDL launch attribute during CUDA graph capture

alisonshao:
## Summary
- Skip `cudaLaunchAttributeProgrammaticStreamSerialization` when the stream is in CUDA graph capture mode
- Check `cudaStreamIsCapturing()` before setting the PDL attribute in `LaunchKernel::enable_pdl()`

## Problem
On SM120 (RTX 5090/Blackwell), CUDA graph capture fails with `CUDA error: invalid resource handle` at `kvcache.cuh:196` during `init_device_graphs()`. The PDL stream serialization launch attribute is incompatible with graph capture mode.

```
RuntimeError: Runtime check failed at .../kvcache.cuh:196: CUDA error: invalid resource handle
Exception: Capture cuda graph failed
```

Failure: https://github.com/sgl-project/sglang/actions/runs/21894982918/job/63209014248

## Fix
PDL is only applied during normal kernel execution, not during CUDA graph capture. This preserves the PDL performance benefit for regular inference while allowing graph capture to succeed.

## Test plan
- [ ] Verify CUDA graph capture succeeds on SM120 (5090) runners
- [ ] Verify no performance regression on SM90 (H100/H200) runners

### DIFF
--- a/python/sglang/jit_kernel/include/sgl_kernel/utils.cuh
+++ b/python/sglang/jit_kernel/include/sgl_kernel/utils.cuh
@@ -112,10 +112,17 @@ struct LaunchKernel {
 
   auto enable_pdl(bool enabled = true) -> LaunchKernel& {
     if (enabled) {
-      m_attrs[0].id = cudaLaunchAttributeProgrammaticStreamSerialization;
-      m_attrs[0].val.programmaticStreamSerializationAllowed = true;
-      m_config.numAttrs = 1;
-      m_config.attrs = m_attrs;
+      // Skip PDL during CUDA graph capture - the ProgrammaticStreamSerialization
+      // launch attribute is incompatible with graph capture on SM120 (Blackwell)
+      // and causes "CUDA error: invalid resource handle".
+      cudaStreamCaptureStatus capture_status = cudaStreamCaptureStatusNone;
+      cudaStreamIsCapturing(m_config.stream, &capture_status);
+      if (capture_status == cudaStreamCaptureStatusNone) {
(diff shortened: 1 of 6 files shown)


DarkSharpness:
Calling runtime API `cudaStreamIsCapturing` might be time consuming. We need double check about the host-side overhead.

BTW, we may only apply this fix on 5090 if the issue only exists on SM120 target.

python/sglang/jit_kernel/include/sgl_kernel/utils.cuh
```diff
@@ -112,10 +112,17 @@ struct LaunchKernel {
...
-      m_attrs[0].id = cudaLaunchAttributeProgrammaticStreamSerialization;
-      m_attrs[0].val.programmaticStreamSerializationAllowed = true;
-      m_config.numAttrs = 1;
-      m_config.attrs = m_attrs;
+      // Skip PDL during CUDA graph capture - the ProgrammaticStreamSerialization
+      // launch attribute is incompatible with graph capture on SM120 (Blackwell)
+      // and causes "CUDA error: invalid resource handle".
+      cudaStreamCaptureStatus capture_status = cudaStreamCaptureStatusNone;
+      cudaStreamIsCapturing(m_config.stream, &capture_status);
+      if (capture_status == cudaStreamCaptureStatusNone) {
```
DarkSharpness, line 120: If we have disabled PDL on SM120 in Python, then we should not change the C++ code, which completely. Just keep the original behavior on other devices (B200, H200) which have no compatibility issue.

PR #18614, opened 2026-02-11
Automatically resolve NVFP4 MoE backend to `flashinfer_trtllm` instead of `flashinfer_cutlass`

b8zhong:
## Motivation

Generally the performance of `trtllm-gen` in most workloads is better, and there is a lot of hardcoding in server args for every architecture (Qwen/Q3N/Deepseek/GLM+), thus I think it'd just be best to change it for new architectures. And now, it also supports a variety of routing types and dtypes, so it seems safe to change from a compatibility perspective. If a model doesn't explicitly support the routing type detection and needs it in the modelling file, we can also change that when needed.

## Modifications

Change the default


b8zhong:
Looks done in https://github.com/sgl-project/sglang/pull/21776

PR #18627, opened 2026-02-11
fix(model-config): guard rope scaling factor access

mondaylord:
# Motivation
Some MLA model configs provide `rope_scaling` with a non-default rope type but without `factor`.

Current code directly reads `rope_scaling["factor"]` once rope type is non-default, which can raise a runtime `KeyError` during model config initialization.

This PR adds a guard so scaling adjustment is only applied when `factor` exists.

# Modifications
- `python/sglang/srt/configs/model_config.py`
  - Change condition from `if rope_type != "default":` to `if rope_type != "default" and "factor" in rope_scaling:`.
  - Behavior is unchanged when `factor` exists; it safely skips scaling adjustment otherwise.

# Testing
- Verified diff and logic locally.

# Checklist
- [ ] Format your code according to the Format code with pre-commit.
- [ ] Add unit tests according to the Run and add unit tests.
- [ ] Update documentation according to Write documentations.
- [ ] Provide accuracy and speed benchmark results.
- [ ] Follow the SGLang code style guidance.

### DIFF
--- a/python/sglang/srt/configs/model_config.py
+++ b/python/sglang/srt/configs/model_config.py
@@ -450,7 +450,7 @@ class ModelConfig:
                     or rope_scaling.get("type")
                     or "default"
                 )
-                if rope_type != "default":
+                if rope_type != "default" and "factor" in rope_scaling:
                     mscale_all_dim = rope_scaling.get("mscale_all_dim", False)
                     scaling_factor = rope_scaling["factor"]
                     mscale = yarn_get_mscale(scaling_factor, float(mscale_all_dim))


hnyls2002:
Thanks @mondaylord. I'm fairly sure this is now handled on main: the MLA rope-scaling logic was extracted into compute_mla_mscale_scaling (model_config.py), which uses rope_scaling.get("factor", 1.0) and warns when factor is missing instead of raising. Closing as superseded; reopen if there's a config path I missed.

PR #18630, opened 2026-02-11
[FEAT] Add Anthropic compatible API endpoint

JustinTong0323:
## Summary
- Add Anthropic-compatible `/v1/messages` and `/v1/messages/count_tokens` endpoints that translate between Anthropic Messages API format and
SGLang's existing OpenAI-compatible chat completion infrastructure
- Supports non-streaming, streaming (SSE), tool use, system messages, and all standard Anthropic request parameters
- Enables tools like Claude Code to use SGLang-served models as drop-in Anthropic API replacements via `ANTHROPIC_BASE_URL`

## Details

**Architecture**: A translation layer (`AnthropicServing`) that delegates to `OpenAIServingChat` internally. Anthropic requests are converted to
`ChatCompletionRequest`, processed through existing infrastructure, and responses are converted back to Anthropic format.

**New files**:
- `python/sglang/srt/entrypoints/anthropic/protocol.py` — Pydantic models for Anthropic Messages API
- `python/sglang/srt/entrypoints/anthropic/serving.py` — Core handler with request/response conversion and streaming state machine
- `test/registered/openai_server/basic/test_anthropic_server.py` — 19 basic API tests
- `test/registered/openai_server/function_call/test_anthropic_tool_use.py` — 10 tool use tests
- `test/manual/vlm/test_anthropic_vision.py`  — visual understanding test

**Modified files**:
- `python/sglang/srt/entrypoints/http_server.py` — Register endpoints and initialize handler

## Test plan
- [x] 19 basic tests passing (non-streaming, streaming, system messages, content blocks, error handling, count tokens)
- [x] 10 tool use tests passing (tool format, tool_choice auto/any/specific, multi-turn, streaming tool calls, event sequence)
- [x] Manual integration test with cherry-studio anthropic client
- [x] Manual integration test with Claude Code via set `ANTHROPIC_BASE_URL`

Note: This PR is vibed by claude code, I believe he is more familiar with it's own api then me 🤣


jhinpan (commented):
## Comprehensive Review: Anthropic-Compatible API Endpoints (PR #18630)

Thanks for this well-structured contribution! The translation-layer approach of delegating to `OpenAIServingChat` is clean and avoids duplicating business logic. However, this review identified several issues that need attention before merge, organized by severity.

---

### 🔴 Critical Bugs (5)

**C1. Image URL conversion is broken** (`serving.py:128`)
Raw base64 data is passed directly as a URL. OpenAI format expects a data URI: `data:<media_type>;base64,<data>`. All image inputs will silently fail.
```python
# Current (broken):
"url": block.source.get("data", ""),
# Fix:
media_type = block.source.get("media_type", "image/png")
data = block.source.get("data", "")
"url": f"data:{media_type};base64,{data}",
```

**C2. `tool_result` uses `id` instead of `tool_use_id`** (`protocol.py:40`)
The Anthropic spec defines `tool_use_id` for tool_result blocks, but `AnthropicContentBlock` only has `id`. SDK clients sending spec-compliant `tool_use_id` will have it silently ignored, breaking tool-use round-trips. The serving.py code at line 151 reads `block.id` which won't be populated from `tool_use_id` in the JSON payload.

**C3. `tool_result.content` with list content produces garbled output** (`serving.py:147`)
When `block.content` is a `list[dict]` (Anthropic allows content blocks inside tool_result), `str()` produces Python repr like `[{'type': 'text', 'text': '...'}]` instead of extracting text. Should iterate the list and concatenate text block values.

**C4. Non-streaming response `id` uses OpenAI `chatcmpl-*` format** (`serving.py:594`)
`id=response.id` passes through the OpenAI ID. Anthropic IDs must be `msg_*`. Streaming correctly generates `msg_*` IDs (line 361), but non-streaming doesn't. Fix: use `f"msg_{uuid.uuid4().hex}"` as in the streaming path.

**C5. Missing `thinking` content block type and request parameter** (`protocol.py:35, 94-108`)
Extended thinking is a major Anthropic API feature. The `type` literal only allows `text|image|tool_use|tool_result`, missing `thinking` and `redacted_thinking`. No `thinking` request parameter exists. While not strictly required for an MVP, SDK clients using thinking blocks will get validation errors — this should at least be documented as a known limitation.

---

### 🟠 Important Issues (8)

**I1. `stop_sequence` stop_reason is never emitted** (`serving.py:46-50`)
Both natural EOS and stop-sequence stops map to `"end_turn"`. The code doesn't check `matched_stop` from the OpenAI response. Clients relying on `stop_reason == "stop_sequence"` will never see it.

**I2. Streaming errors are silently swallowed** (`serving.py:410-414`)
If the OpenAI stream emits an error, parsing as `ChatCompletionStreamResponse` fails and the `continue` silently skips it. The stream ends without an error event or `message_stop`. Should emit an Anthropic `error` event.

**I3. Empty responses produce no content blocks** (streaming state machine)
If the model emits only a finish_reason with no text/tool deltas, the stream emits `message_start → message_delta → message_stop` with zero content blocks, violating the Anthropic event schema which expects at least one content block.

**I4. Missing `"none"` tool_choice type** (`protocol.py:74`)
The Anthropic spec defines `"none"` to disallow all tool use. Only `auto|any|tool` are accepted.

**I5. Missing `disable_parallel_tool_use`** on `AnthropicToolChoice`)
This is a commonly-used Anthropic API field with no equivalent. Should be documented as unsupported at minimum.

**I6. System text blocks concatenated without separator** (`serving.py:105`)
`["Hello", "World"]` becomes `"HelloWorld"`. Should use `"\n"` as separator.

**I7. Tight coupling to `OpenAIServingChat` private methods** (`serving.py:248, 260, 269, 351, 652`)
Five underscore-prefixed private methods are called. Any internal refactor of `OpenAIServingChat` will silently break this. Consider making these methods a stable internal API (e.g., remove underscore prefix and document them) or add a comment acknowledging the coupling.

**I8. Exception details exposed in error responses** (`serving.py:77, 277, 323, 669`)
`message=str(e)` can leak internal details (stack frames, module paths) to clients. Use generic messages for 500 errors; only expose specifics for 400 errors.

---

### 🟡 Minor Issues (7)

**M1.** `x-api-key` header is never validated — Anthropic clients using only `x-api-key` (the standard Anthropic auth pattern) without `Authorization: Bearer` will be rejected. Tests misleadingly send both headers.

**M2.** `AnthropicUsage` is missing `cache_creation_input_tokens` and `cache_read_input_tokens` in the `message_delta` usage tracking (they're defined in the model but never populated from the OpenAI response).

**M3.** `response.choices[0]` at serving.py:564 can `IndexError` on empty choices (defensive check recommended).

**M4.** `AnthropicError.type` should be a `Literal` enum of known error types (`invalid_request_error`, `authentication_error`, `permission_error`, `not_found_error`, `rate_limit_error`, `api_error`, `overloaded_error`).

**M5.** `input_schema` validator mutates input (adds `type: "object"`) rather than rejecting invalid schemas.

**M6.** `message_delta` usage includes both `input_tokens` and `output_tokens` — Anthropic spec says `message_delta` should only have `output_tokens` (input_tokens goes in `message_start`).

**M7.** No `ping` event generation — useful for keep-alive on long-running streams.

---

### 🟢 Test Coverage Gaps

1. **No streaming error test** — no test for error events during streaming
2. **No auth failure test** — wrong/missing API key not tested  
3. **No `tool_result` with list content test** — only string content tested
4. **No `stop_sequence` stop_reason assertion** — `test_stop_sequences` only checks status 200
5. **No concurrent request test**
6. **`_parse_sse_events` silently swallows parse errors** — could mask test failures
7. **Several tool use tests have conditional assertions** (e.g. `if len(tool_use_starts) > 0:`) that pass even if no tool use occurred

---

### ✅ What's Done Well

- **Clean architecture** — Translation-layer pattern is correct; no duplicated business logic
- **SSE event format** — Correct `event: <type>\ndata: <json>\n\n` formatting
- **Content block lifecycle** — start→deltas→stop ordering is mostly correct for both text and tool_use
- **tool_choice mapping** — `auto→auto`, `any→required`, `tool→function` is correct
- **`stream_options` injection** — Properly enables usage tracking for streaming
- **Streaming arg reconstruction test** — `test_tool_use_streaming_args_parsing` is excellent
- **Comparative token count test** — Smart approach that's model-independent
- **No route conflicts** — `/v1/messages` and `/v1/messages/count_tokens` are clean additions
- **Test isolation** — Each test class manages its own server lifecycle

---

### Recommendation

**Request changes.** The 5 critical bugs (especially C1-C4) will produce incorrect behavior for common API usage patterns. C1 breaks all image inputs. C2 breaks spec-compliant tool use round-trips. C3 produces garbled tool results. C4 violates the ID format contract. These should be fixed before merge. The important issues (I1-I3 especially) should also be addressed to prevent silent failures in production.

python/sglang/srt/entrypoints/anthropic/serving.py
```diff
@@ -0,0 +1,674 @@
...
+            for block in msg.content:
+                if block.type == "text" and block.text:
+                    content_parts.append({"type": "text", "text": block.text})
+
+                elif block.type == "image" and block.source:
+                    content_parts.append(
+                        {
+                            "type": "image_url",
+                            "image_url": {
+                                "url": block.source.get("data", ""),
```
jhinpan: 🔴 **Bug: Image URL conversion is broken**

Anthropic's image format uses `source.type = "base64"` with `source.media_type` and `source.data`. OpenAI expects a data URI. This passes raw base64 as a URL, which will fail for all image inputs.

Suggested fix:
```python
media_type = block.source.get("media_type", "image/png")
data = block.source.get("data", "")
content_parts.append({
    "type": "image_url",
    "image_url": {
        "url": f"data:{media_type};base64,{data}",
    },
})
```

python/sglang/srt/entrypoints/anthropic/protocol.py
```diff
@@ -0,0 +1,172 @@
...
+
+class AnthropicContentBlock(BaseModel):
+    """Content block in message"""
+
+    type: Literal["text", "image", "tool_use", "tool_result"]
+    text: Optional[str] = None
+    # For image content
+    source: Optional[dict[str, Any]] = None
+    # For tool use/result
+    id: Optional[str] = None
```
jhinpan, line 42: 🔴 **Bug: `tool_result` uses `id` instead of `tool_use_id`**

The Anthropic spec defines `tool_use_id` for tool_result blocks, not `id`. SDK clients sending spec-compliant `tool_use_id` will have it silently ignored via Pydantic's `model_config`, breaking tool-use round-trips.

Suggested fix: add a `tool_use_id` field and use it in serving.py's tool_result handling:
```python
tool_use_id: Optional[str] = None
```
Then in serving.py line 151, use `block.tool_use_id or block.id or ""`.

python/sglang/srt/entrypoints/anthropic/serving.py
```diff
@@ -0,0 +1,674 @@
...
+                            "name": block.name or "",
+                            "arguments": json.dumps(block.input or {}),
+                        },
+                    }
+                    tool_calls.append(tool_call)
+
+                elif block.type == "tool_result":
+                    # Tool results from user become separate tool messages
+                    if msg.role == "user":
+                        tool_content = str(block.content) if block.content else ""
```
jhinpan: 🔴 **Bug: `str()` on list content produces Python repr**

When `block.content` is a `list[dict]` (Anthropic allows content blocks inside tool_result), `str()` produces `[{'type': 'text', 'text': '...'}]` instead of extracting text.

Suggested fix:
```python
if isinstance(block.content, list):
    tool_content = "\n".join(
        item.get("text", "") for item in block.content
        if isinstance(item, dict) and item.get("type") == "text"
    )
else:
    tool_content = str(block.content) if block.content else ""
```

python/sglang/srt/entrypoints/anthropic/serving.py
```diff
@@ -0,0 +1,674 @@
...
+                        name=tool_call.function.name,
+                        input=tool_input,
+                    )
+                )
+
+        # Map stop reason
+        stop_reason = STOP_REASON_MAP.get(choice.finish_reason or "stop", "end_turn")
+
+        return AnthropicMessagesResponse(
+            id=response.id,
```
jhinpan: 🔴 **Bug: Non-streaming response uses OpenAI `chatcmpl-*` ID format**

`id=response.id` passes through the OpenAI ID. Anthropic IDs must be `msg_*` format. The streaming path correctly generates `msg_*` IDs (line 361), but non-streaming doesn't.

Suggested fix:
```python
id=f"msg_{uuid.uuid4().hex}",
```

python/sglang/srt/entrypoints/anthropic/serving.py
```diff
@@ -0,0 +1,674 @@
...
+    ToolChoiceFuncName,
+)
+
+if TYPE_CHECKING:
+    from sglang.srt.entrypoints.openai.serving_chat import OpenAIServingChat
+
+logger = logging.getLogger(__name__)
+
+# Map OpenAI finish reasons to Anthropic stop reasons
+STOP_REASON_MAP = {
```
jhinpan, line 46: 🟠 **`stop_sequence` stop_reason never emitted**

Both natural EOS and stop-sequence stops from OpenAI return `finish_reason: "stop"`. This map sends both to `"end_turn"`. To properly distinguish, check if `stop_sequences` were in the request AND if the OpenAI response's `matched_stop` is non-null:

```python
# In _convert_response / streaming [DONE] handler:
if request_had_stop_sequences and response has matched_stop:
    stop_reason = "stop_sequence"
```

python/sglang/srt/entrypoints/anthropic/serving.py
```diff
@@ -0,0 +1,674 @@
...
+                yield _wrap_sse_event(
+                    stop_msg.model_dump_json(exclude_none=True),
+                    "message_stop",
+                )
+                continue
+
+            # Parse the OpenAI chunk
+            try:
+                chunk = ChatCompletionStreamResponse.model_validate_json(data_str)
+            except Exception:
```
jhinpan, line 428: 🟠 **Streaming errors silently swallowed**

If the OpenAI stream emits an error (or malformed JSON), this `continue` silently skips it. The stream ends without any error indication or `message_stop` event. Clients will see an incomplete stream.

Should emit an Anthropic `error` event:
```python
except Exception:
    logger.debug("Failed to parse stream chunk: %s", data_str)
    error_event = AnthropicStreamEvent(
        type="error",
        error=AnthropicError(type="api_error", message="Stream processing error"),
    )
    yield _wrap_sse_event(error_event.model_dump_json(exclude_none=True), "error")
    continue
```

python/sglang/srt/entrypoints/anthropic/serving.py
```diff
@@ -0,0 +1,674 @@
...
+        if anthropic_request.system:
+            if isinstance(anthropic_request.system, str):
+                openai_messages.append(
+                    {"role": "system", "content": anthropic_request.system}
+                )
+            else:
+                system_text = ""
+                for block in anthropic_request.system:
+                    if block.type == "text" and block.text:
+                        system_text += block.text
```
jhinpan: 🟠 **System text blocks concatenated without separator**

Multiple system text blocks are concatenated directly: `["Hello", "World"]` becomes `"HelloWorld"`. Should use `"\n"` as separator:
```python
system_parts = []
for block in anthropic_request.system:
    if block.type == "text" and block.text:
        system_parts.append(block.text)
system_text = "\n".join(system_parts)
```

python/sglang/srt/entrypoints/anthropic/serving.py
```diff
@@ -0,0 +1,674 @@
...
+            # Get response from OpenAI handler
+            response = await self.openai_serving_chat._handle_non_streaming_request(
+                adapted_request, processed_request, raw_request
+            )
+        except Exception as e:
+            logger.exception("Error processing Anthropic request: %s", e)
+            return self._error_response(
+                status_code=500,
+                error_type="internal_error",
+                message=str(e),
```
jhinpan: 🟠 **Exception details exposed to clients**

`message=str(e)` at 500 error paths (lines 77, 277, 323, 669) can leak internal details like module paths, stack frames, or config values. Use a generic message for 500 errors:
```python
message="Internal server error",
```
The full exception is already logged via `logger.exception()`.

python/sglang/srt/entrypoints/anthropic/protocol.py
```diff
@@ -0,0 +1,172 @@
...
+    input_tokens: int
+    output_tokens: int
+    cache_creation_input_tokens: Optional[int] = None
+    cache_read_input_tokens: Optional[int] = None
+
+
+class AnthropicContentBlock(BaseModel):
+    """Content block in message"""
+
+    type: Literal["text", "image", "tool_use", "tool_result"]
```
jhinpan: 🟠 **Missing `thinking` and `redacted_thinking` content block types**

Extended thinking is a key Anthropic API feature. Clients using thinking blocks will get validation errors. At minimum, add these to the `type` literal:
```python
type: Literal["text", "image", "tool_use", "tool_result", "thinking", "redacted_thinking"]
```
And add corresponding fields:
```python
thinking: Optional[str] = None
signature: Optional[str] = None
```

python/sglang/srt/entrypoints/anthropic/protocol.py
```diff
@@ -0,0 +1,172 @@
...
+            raise ValueError("input_schema must be a dictionary")
+        if "type" not in v:
+            v["type"] = "object"
+        return v
+
+
+class AnthropicToolChoice(BaseModel):
+    """Tool Choice definition"""
+
+    type: Literal["auto", "any", "tool"]
```
jhinpan: 🟠 **Missing `"none"` tool_choice type**

The Anthropic spec defines `"none"` to disallow all tool use. Currently only `auto|any|tool` are accepted.
```python
type: Literal["auto", "any", "tool", "none"]
```

python/sglang/srt/entrypoints/anthropic/serving.py
```diff
@@ -0,0 +1,674 @@
...
+                yield _wrap_sse_event(
+                    delta_event.model_dump_json(exclude_none=True),
+                    "content_block_delta",
+                )
+
+    def _convert_response(
+        self, response: ChatCompletionResponse
+    ) -> AnthropicMessagesResponse:
+        """Convert an OpenAI ChatCompletionResponse to an Anthropic Messages response."""
+        choice = response.choices[0]
```
jhinpan, line 596: 🟡 **Potential `IndexError` on empty choices**

`response.choices[0]` will throw `IndexError` if `choices` is empty. Add a defensive check:
```python
if not response.choices:
    return AnthropicMessagesResponse(
        content=[AnthropicContentBlock(type="text", text="")],
        model=response.model,
        stop_reason="end_turn",
        usage=AnthropicUsage(input_tokens=0, output_tokens=0),
    )
```

test/registered/openai_server/basic/test_anthropic_server.py
```diff
@@ -0,0 +1,492 @@
...
+    @classmethod
+    def tearDownClass(cls):
+        kill_process_tree(cls.process.pid)
+
+    def _make_request(self, payload, stream=False):
+        """Send a request to the /v1/messages endpoint."""
+        headers = {
+            "Content-Type": "application/json",
+            "Authorization": f"Bearer {self.api_key}",
+            "x-api-key": self.api_key,
```
jhinpan: 🟡 **`x-api-key` header is decorative**

The server's auth middleware only checks `Authorization` header. The `x-api-key` header here is never validated. This is misleading — Anthropic SDK clients using only `x-api-key` will be rejected.

Either:
1. Add `x-api-key` support to the auth middleware, or
2. Remove it from tests to avoid false confidence

test/registered/openai_server/function_call/test_anthropic_tool_use.py
```diff
@@ -0,0 +1,551 @@
...
+
+        # Check for tool use content block events
+        block_starts = [e for e in events if e["type"] == "content_block_start"]
+        tool_use_starts = [
+            e
+            for e in block_starts
+            if e.get("content_block", {}).get("type") == "tool_use"
+        ]
+
+        if len(tool_use_starts) > 0:
```
jhinpan: 🟡 **Conditional assertion means test can pass without tool use**

`if len(tool_use_starts) > 0:` means this test passes even if the model never produces a tool call. Consider using `tool_choice: {"type": "any"}` to force tool use (like `test_tool_use_streaming_event_sequence` does), or at minimum log a warning.

Fridge003 (approved):

PR #18634, opened 2026-02-11
[Fix] Add lora tied lm head support (for Qwen2.5, Gemma, etc model need)

yushengsu-thu:
## Motivation

Added test ci: `324` lines
Modified code in sgl: `87` lines

When `tie_word_embeddings=True` (e.g., Qwen2.5, Gemma), the model's `lm_head` is the same Python object as `embed_tokens`. PyTorch's `named_modules()` deduplicates by object identity, so `lm_head` never appears as a separate module — LoRA cannot wrap it. This causes LoRA adapters that target `lm_head` to silently fail or produce incorrect results.

Additionally, PEFT adapters may use shorthand strings like `"all-linear"` or `"all"` for `target_modules`, which SGLang previously did not handle, leading to crashes during adapter loading. PEFT also renames `lm_head` to `unembed_tokens` internally in some configurations, which was not recognized by SGLang's weight loader.

### `python/sglang/srt/lora/lora_manager.py`
- **Untie `lm_head` for LoRA wrapping**: When `lm_head` is the same object as `embed_tokens`, create a new `ParallelLMHead` that shares the base weight tensor (no extra GPU memory) so that `named_modules()` yields it as an independent module.
- **Handle PEFT shorthand `target_modules`**: Gracefully handle `"all-linear"` and `"all"` strings by requiring the user to specify `--lora-target-modules` at server startup. Raise clear error messages for unrecognized string values.

### `python/sglang/srt/lora/lora.py`
- **Remap `unembed_tokens` to `lm_head`**: PEFT internally renames `lm_head` to `unembed_tokens` in some adapter configs. The weight loader now remaps this key so the weight is loaded into the correct buffer.
- **Allow loading when `normalized_target_modules` is empty**: When `target_modules` is a shorthand like `"all-linear"`, the normalized set is empty. Allow `embed_tokens`/`lm_head` weights to be loaded in this case, deferring to `--lora-target-modules` for module selection.

### `python/sglang/srt/lora/utils.py`
- **Handle string `target_modules` in `get_normalized_target_modules()`**: Return an empty set for PEFT shorthands so callers can fall back to CLI config.
- **Add `unembed_tokens` → `lm_head` mapping** to `params_mapping`.

### `test/registered/lora/test_lora_tied_lm_head.py` (new)
- Programmatically creates a LoRA adapter with `lm_head` in `target_modules` on a model with `tie_word_embeddings=True` (`Qwen/Qwen2.5-0.5B`).
- **`test_tied_lm_head_lora_no_nan`**: Verifies SGLang does not produce NaN values.
- **`test_tied_lm_head_lora_differs_from_base`**: Confirms LoRA output differs from the base model (i.e., `lm_head` LoRA is actually applied).
- **`test_tied_lm_head_lora_hf_sgl_logprob_match`**: Compares prefill and decode logprobs between HuggingFace+PEFT and SGLang, ensuring numerical consistency within threshold.

## Accuracy Tests

The new test `test_lora_tied_lm_head.py` validates:
1. No NaN values in output logprobs
2. LoRA adapter produces different output from the base model
3. SGLang logprobs match HuggingFace+PEFT logprobs (max diff < 2e-1)

Tested with `Qwen/Qwen2.5-0.5B` (tie_word_embeddings=True) + triton LoRA backend.

## Benchmarking and Profiling

No impact on inference speed — the untied `lm_head` shares the same weight tensor as `embed_tokens`, adding zero GPU memory overhead. The change only affects the module graph structure during initialization.

## Checklist


python/sglang/srt/lora/utils.py
```diff
@@ -98,12 +98,22 @@ def get_hidden_dim(
...
+
+    Also handles PEFT shorthand strings like "all-linear" or "all" by returning
+    an empty set (the caller should rely on --lora-target-modules from the CLI
+    to determine the concrete module set).
     """
+    # Handle PEFT shorthand strings — these cannot be resolved to concrete
+    # module names without inspecting the base model, so we return an empty
+    # set and let the caller fall back to the CLI --lora-target-modules.
+    if isinstance(target_modules, str):
+        return set()
```
Fridge003: Rather than empty set, can we return {"all"}, which is less confusing

test/registered/lora/test_lora_tied_lm_head.py
```diff
@@ -0,0 +1,328 @@
...
+            if diff > 1e-6:
+                any_differ = True
+
+        self.assertTrue(
+            any_differ,
+            "LoRA output logprobs are identical to base model. "
+            "lm_head LoRA may not be applied correctly.",
+        )
+
+    def test_tied_lm_head_lora_hf_sgl_logprob_match(self):
```
Fridge003, line 150: We can only keep this subtest, since it includes the functionality of the two subtests above

yushengsu-thu: fixed

Fridge003 (approved):

PR #18643, opened 2026-02-11
Register tensors with symmetric memory for qwen

nvcastet:
## Motivation

Perform add op in-place so that the MoE output still belongs to the symmetric memory pool (when symm mem is enabled).

## Checklist

## Review Process

### DIFF
--- a/python/sglang/srt/models/qwen2_moe.py
+++ b/python/sglang/srt/models/qwen2_moe.py
@@ -324,6 +324,10 @@ class Qwen2MoeSparseMoeBlock(nn.Module):
             final_hidden_states = self._forward_router_experts(hidden_states)
 
         if shared_output is not None:
+            # In-place add is required to keep final_hidden_states in the
+            # symmetric memory pool (when --enable-symm-mem is used).
+            # An out-of-place add would allocate a new tensor outside symm
+            # memory, breaking subsequent symmetric collective operations.
             final_hidden_states += shared_output
         if self.tp_size > 1 and not use_reduce_scatter:
             final_hidden_states = tensor_model_parallel_all_reduce(final_hidden_states)
--- a/python/sglang/srt/models/qwen2_moe.py
+++ b/python/sglang/srt/models/qwen2_moe.py
@@ -324,7 +324,7 @@ class Qwen2MoeSparseMoeBlock(nn.Module):
             final_hidden_states = self._forward_router_experts(hidden_states)
 
         if shared_output is not None:
-            final_hidden_states = final_hidden_states + shared_output
+            final_hidden_states += shared_output
         if self.tp_size > 1 and not use_reduce_scatter:
             final_hidden_states = tensor_model_parallel_all_reduce(final_hidden_states)
 


python/sglang/srt/models/qwen2_moe.py
```diff
@@ -324,7 +323,7 @@ def forward(
             final_hidden_states = self._forward_router_experts(hidden_states)
 
         if shared_output is not None:
-            final_hidden_states = final_hidden_states + shared_output
+            final_hidden_states += shared_output
```
hlu1, line 331: This is extremely error prone. How can the user know that `shared_output` is actually symmetric memory and in-place add here is needed?

nvcastet: > This is extremely error prone.

I agree.

> How can the user know that shared_output is actually symmetric memory and in-place add here is needed?

Here `final_hidden_states` is symmetrically allocated/registered but if we perform an out-of-place operation, we would need to add a symm memory context manager to register the output tensor.

The better solution involves using torch compiler to dynamically trace buffers used by symmetric ops and automatically register them (in this case it would have registered the output of the out-of-place add op).
See proposal at work: https://github.com/pytorch/pytorch/issues/162859#issuecomment-3518237652

For now, to have some introspection on what gets registered or not, I added a debug mode (https://github.com/sgl-project/sglang/pull/18569). Once merged, we can add unit tests to symm memory configs to fail if new comm tensors are not registered.

hlu1:
Please add perf benchmark numbers

python/sglang/srt/models/qwen2_moe.py
hlu1: I remember that in the early implementation the symmetric memory backed tensors are tagged and that tag has been removed later. Do you remember why?

nvcastet: 1- The tag was another source of errors: developer adds the context manager but forgot to tag the buffer that was supposed to be in symm memory
2- Tagging was done by adding an attribute to the tensor object which was not preserved if a tensor slice happens. Alternative implementation was doing a look up in the pytorch memory snapshot but it was not accepted since memory snapshot is expensive on the CPU (debug mode at https://github.com/sgl-project/sglang/pull/18569 is using the snapshot)

nvcastet:
## Symmetric Memory Benchmark: Qwen3-Next-80B-A3B-Thinking-FP8 on 4x GB200

### TP4 Configuration

**Server (symm-mem enabled):**
```bash
python -m sglang.launch_server --model Qwen/Qwen3-Next-80B-A3B-Thinking-FP8 --tp 4 --max-running-requests 1024 --chunked-prefill-size 8192 --mem-fraction-static 0.8 --disable-radix-cache --mamba-ssm-dtype float32 --fp8-gemm-backend flashinfer_trtllm --attention-backend trtllm_mha --cuda-graph-max-bs=1024 --enable-symm-mem
```

**Server (symm-mem disabled):** same as above without `--enable-symm-mem`

**Client:**
```bash
python3 -m sglang.bench_one_batch_server --model-path Qwen/Qwen3-Next-80B-A3B-Thinking-FP8 --dataset-name random --input-len 1024 --output-len 1024 --batch-size 1024 512 128 --base-url http://127.0.0.1:30000
```

| Batch Size | Metric | Symm-Mem Disabled | Symm-Mem Enabled | Gain |
|---|---|---|---|---|
| 1024 | Prefill (tok/s) | 58,472.54 | 63,989.35 | **+9.4%** |
| 1024 | Decode (tok/s) | 31,369.96 | 32,934.25 | **+5.0%** |
| 512 | Prefill (tok/s) | 62,012.44 | 63,895.43 | **+3.0%** |
| 512 | Decode (tok/s) | 24,305.91 | 24,859.08 | **+2.3%** |
| 128 | Prefill (tok/s) | 61,414.28 | 62,912.60 | **+2.4%** |
| 128 | Decode (tok/s) | 11,316.08 | 11,375.96 | **+0.5%** |

---

### DEP4 Configuration

**Server (symm-mem enabled):**
```bash
python3 -m sglang.launch_server --model Qwen/Qwen3-Next-80B-A3B-Thinking-FP8 --disable-radix-cache --attention-backend trtllm_mha --tp-size 4 --ep 4 --cuda-graph-max-bs 512 --enable-dp-attention --dp 4 --stream-interval 10 --mem-fraction-static 0.9 --max-running-requests 2048 --enable-dp-lm-head --mamba-ssm-dtype float32 --fp8-gemm-backend flashinfer_trtllm --chunked-prefill-size 8192 --enable-symm-mem
```

**Server (symm-mem disabled):** same as above without `--enable-symm-mem`

**Client:**
```bash
python3 -m sglang.bench_one_batch_server --model-path Qwen/Qwen3-Next-80B-A3B-Thinking-FP8 --dataset-name random --input-len 1024 --output-len 1024 --batch-size 2048 --base-url http://127.0.0.1:30000
```

| Batch Size | Metric | Symm-Mem Disabled | Symm-Mem Enabled | Gain |
|---|---|---|---|---|
| 2048 | Prefill (tok/s) | 43,284.91 | 60,495.37 | **+39.8%** |
| 2048 | Decode (tok/s) | 11,903.58 | 12,384.14 | **+4.0%** |

---

python/sglang/srt/models/qwen2_moe.py
hlu1: Could you please add a comment here on why in-place add is a must so the developer is aware of the intricacies of symmetric memory? Thanks!

hlu1 (approved):

python/sglang/srt/models/qwen2_moe.py
nvcastet: Yes let me do that.

ispobock (approved):

PR #18684, opened 2026-02-12
[AMD] Add MoE weights and scales padding

mqhc2020:
## Motivation

Right now, Aiter MoE requires weights and scales to align with a fixed number. Since some models have intermediate sizes that don't fit this rule, we need to add extra padding to the weights so they can be processed by the Fused MoE.

## Modifications

Add padding for the weights. Below listed are the models and configurations that has been verified with:
1. Qwen/Qwen3-235B-A22B-Instruct-2507-FP8 : TP=8
2. amd/Qwen3-235B-A22B-Instruct-2507-MXFP4: TP=8
3. zai-org/GLM-4.7: TP=8

## Accuracy Tests

We run GSM8K accuracy test for models below:

Qwen/Qwen3-235B-A22B-Instruct-2507-FP8 (TP8):  

Accuracy: 0.980
Invalid: 0.000
Latency: 10.525 s
Output throughput: 2663.037 token/s

amd/Qwen3-235B-A22B-Instruct-2507-MXFP4 (TP8):

Accuracy: 0.940
Invalid: 0.000
Latency: 23.853 s
Output throughput: 7488.441 token/s
zai-org/GLM-4.7 (TP8):

(better than VLLM TP4 on model card)

Accuracy: 0.975
Invalid: 0.000
Latency: 7.349 s
Output throughput: 2683.189 token/s

## Benchmarking and Profiling

Baseline benchmark for amd/Qwen3-235B-A22B-Instruct-2507-mxfp4 (TP4) and Qwen/Qwen3-235B-A22B-Instruct-2507-FP8 (TP8):
image

## Checklist

## Review Process


python/sglang/srt/model_executor/model_runner.py
```diff
@@ -732,7 +721,9 @@ def check_quantized_moe_compatibility(self):
                     f"moe_intermediate_size {moe_intermediate_size} must be divisible by moe_tp_size ({moe_tp_size}) which is tp_size ({self.tp_size}) divided by moe_ep_size ({self.moe_ep_size})."
                 )
 
-            if (moe_intermediate_size // moe_tp_size) % weight_block_size_n != 0:
+            if (
+                moe_intermediate_size // moe_tp_size
+            ) % weight_block_size_n != 0 and not _is_hip:
```
sogalin: Hello @mqhc2020,
With this modification, it looks like we are going to ignore all checks in SGL when _is_hip is True?

python/sglang/srt/layers/quantization/fp8.py
```diff
@@ -739,27 +716,42 @@ def create_weights(
...
+                # Required by column parallel or enabling merged weights
+                if intermediate_size_per_partition % block_n != 0:
                     raise ValueError(
-                        f"The input_size of down's weight = "
+                        f"The output_size of gate's and up's weight = "
                         f"{intermediate_size_per_partition} is not divisible by "
-                        f"weight quantization block_k = {block_k}."
+                        f"weight quantization block_n = {block_n}."
                     )
+                if tp_size > 1:
```
yichiche, line 810: Do aiter support TP > 1 case if using block_quant?

mqhc2020: Yes, I think the sanity check should be kept for aiter MoE too.

yichiche (approved):
LGTM

mqhc2020:
I found a problem myself, running Grok 1. Trying to fix that.

sogalin (approved):

kkHuang-amd (commented):
Could we add a utile function to do padding and remove the padding_size declaration in the "python/sglang/srt/layers/moe/fused_moe_triton/fused_moe.py", the place is weird.

mqhc2020:
> @mqhc2020 , Is accuracy expected 0.835 with amd/Qwen3-235B-A22B-Instruct-2507-mxfp4 (TP4)?
> 
> Accuracy: 0.835 Invalid: 0.000 Latency: 13.401 s Output throughput: 2009.483 token/s

@sogalin  this patch involves various models. I can't guarantee the accuracy of the models can be totally fine after the 128-aligned problem is fixed by this PR. That might be another task later.

python/sglang/srt/layers/quantization/quark/schemes/quark_w4a4_mxfp4_moe.py
```diff
@@ -99,7 +110,7 @@ def create_weights(
             torch.empty(
                 num_experts,
                 hidden_size,
-                intermediate_size_per_partition // 2,
+                w2_down_dim,
```
fxmarty-amd, line 113: There is no issue with padding here? Won't we load data into wrong columns?

fxmarty-amd, line 113: Also: why is `w2_weight` padded but not `w2_weight_scale`?

python/sglang/srt/layers/moe/utils.py
```diff
@@ -341,3 +336,30 @@ class RoutingMethodType(IntEnum):
...
+
+    if is_aiter_moe:
+        padding_size = get_moe_padding_size(True)
+        align_aiter = lambda n: ((n + padding_size - 1) // padding_size) * padding_size
+        is_padded = (w2_down_dim % padding_size) > 0
+        if is_padded:
+            w2_down_dim = align_aiter(w2_down_dim)
+        # packed
+        if hasattr(torch, "float4_e2m1fn_x2") and is_packed:
+            w13_up_dim *= 2
```
fxmarty-amd: Why not `w2_down_dim` as well here? Is this for fp4 packing on one byte or for `gate_proj`/`up_proj` packing only?

python/sglang/srt/layers/quantization/quark/schemes/quark_w4a4_mxfp4_moe.py
fxmarty-amd: Ah no this is handled in `python/sglang/srt/layers/moe/fused_moe_triton/layer.py`.

Personally, I would prefer to have padding handled in `process_weights_after_loading` (e.g. it makes it easier for online quantization), but both are fine though.

mqhc2020: Yes, loading happens in layer.py. It basically copy tensors with 128-aligned inter_dim and some GPUs have "dummy" tensors, which won't contribute to the logits.

python/sglang/srt/layers/moe/utils.py
mqhc2020: The alignment unit is "byte" and two fp4s are packed into one byte.

python/sglang/srt/layers/quantization/quark/schemes/quark_w4a4_mxfp4_moe.py
mqhc2020: Yes you are right and we are on it. This should have been fixed in latest commit.

fxmarty-amd (approved):
LGTM.

python/sglang/srt/layers/moe/utils.py
```diff
@@ -341,3 +344,30 @@ class RoutingMethodType(IntEnum):
     TopK = (5,)
     # Unspecified
     Unspecified = 6
+
+
+def get_moe_padding_size(is_aiter_moe):
+    if is_aiter_moe:
+        return 128
```
BowenBao, line 351: Hi @mqhc2020 , I recently looked at something similar on vllm side https://github.com/vllm-project/vllm/pull/34285, it was mentioned that [256 pad size might have better performance](https://github.com/vllm-project/vllm/pull/34285#discussion_r2855710524). Could you check if this is the case too?

HaiShaw:
@saienduri please help to look at time out issue from new runner? cc @bingxche @yctseng0211 
@BowenBao will merge now. @mqhc2020 please try out padding 256 and file another PR if that is better.

PR #18702, opened 2026-02-12
fix(function_call): forward unknown tool calls in streaming path

soaringk:
## Motivation

PR #12226 added `SGLANG_FORWARD_UNKNOWN_TOOLS` support for `parse_base_json` (non-streaming) and several detector-specific streaming paths (`qwen3_coder_detector`, `gpt_oss_detector`, `pythonic_detector`), but missed `BaseFormatDetector.parse_streaming_increment` — the streaming path used by `Qwen25Detector` (and any detector delegating to the base class).

When `SGLANG_FORWARD_UNKNOWN_TOOLS=false`, the streaming path also behaved inconsistently with non-streaming: only the tool name was discarded while remaining argument chunks leaked as normal text fragments. Non-streaming skips the entire tool call block as a unit.

## Modifications

Three changes in `BaseFormatDetector`:

1. **New `_discarding_unknown_tool` state flag**: When an unknown tool name is detected and forwarding is off, the parser enters "discard mode" — silently consuming tokens until the `eot_token` is found. This ensures the entire tool call block is skipped, consistent with `parse_base_json`.

2. **Tool name validation** (streaming): Gate the state-reset behind `SGLANG_FORWARD_UNKNOWN_TOOLS`. When the env var is True, log a warning but continue processing. When False, enter discard mode.

3. **Tool name sending** (streaming): Accept the function name when `SGLANG_FORWARD_UNKNOWN_TOOLS` is True, even if it's not in `_tool_indices`.

Default behavior (env var False) is unchanged — unknown tool call blocks are skipped entirely, matching non-streaming behavior.

## Checklist


ShangmingCai:
please fix lint first.

`pre-commit run --all-files`

PR #18709, opened 2026-02-12
[diffusion][CI]: Add individual component accuracy CI for diffusion models

Ratish1:
## Motivation

This PR introduces component-level accuracy testing for the diffusion runtime in `sglang.multimodal_gen`. Related: #12987

This PR adds a dedicated component-accuracy framework for validating SGLang runtime components against their reference Hugging Face implementations.

For individual component accuracy, the reference side is:
- **diffusers** for `transformer` and `vae`
- **transformers** for `text_encoder`

That is the core value of this PR. It gives us a targeted correctness layer between raw checkpoint loading and full
serving tests.

This is especially important for diffusion runtime components because they are often not raw HF modules. In practice they may differ in:

- parameter naming
- fused versus split tensor layouts
- tensor-parallel sharding
- runtime-specific initialization and loading paths

A component-level parity framework is therefore necessary to validate the actual runtime module implementation instead of relying only on end-to-end pipeline behavior.

### Add a component-accuracy framework for diffusion runtime modules

This PR adds a dedicated accuracy flow for comparing runtime-loaded SGLang components against Hugging Face reference components.

The framework supports the main component families used by the diffusion runtime:

- transformer
- VAE
- text encoder

The comparison flow is:

1. resolve the concrete component source for a testcase
2. build runtime-style `ServerArgs`
3. initialize the required distributed/runtime state
4. load the SGLang component through the actual runtime loader path
5. load the corresponding Hugging Face reference component
6. align reference weights into the runtime component
7. run both components
8. compare outputs with cosine similarity

This gives us a stable component-level correctness signal that can run inside CI.

### Use the real SGLang runtime loader path

The SGLang side of the comparison is loaded through the actual runtime component loader stack.

That means the framework exercises the same runtime-side loading logic used by the codebase rather than constructing fake
test-only modules. This is important because the goal is to validate the real runtime implementation, including:

- runtime model class resolution
- distributed/model-parallel setup
- component-specific loader logic
- runtime-specific tensor layouts

### Compare against Hugging Face reference implementations

The reference side is loaded from the corresponding Hugging Face implementation for each component family:

- `transformer` and `vae` use **diffusers**
- `text_encoder` uses **transformers**

This is intentionally a component-level reference path, so that the framework can compare one runtime component at a time and localize issues much more precisely than end-to-end tests allow.

### Add weight-alignment logic for runtime parity testing

A core part of this PR is the weight-alignment path used before comparison.

This exists because the runtime component and the Hugging Face reference component do not always expose weights in the same form. The framework therefore handles:

- parameter name remapping
- known fused tensor layouts such as QKV and gate/up projections
- tensor-parallel-aware copying into sharded runtime parameters

This allows the SGLang runtime component to be compared in the layout it actually uses in the codebase, while still
grounding the comparison in the Hugging Face reference weights.

### Add forward adapters for component-level comparison

The PR introduces native hook/input adapters that build deterministic synthetic inputs for component comparisons.

These adapters exist so that the framework can produce a valid shared input bundle for both:

- the SGLang runtime component
- the Hugging Face reference component

This is needed because the two sides do not always expose identical forward signatures. The adapters normalize those differences so the framework can run a clean parity comparison.

### Add staged low-memory execution for selected 1-GPU cases

Some large 1-GPU cases cannot safely keep both the runtime component and the reference component resident at the same time.

For these cases, the framework supports staged execution:

- run the SGLang side first
- move the output to CPU
- release the first stage
- run the reference side
- compare CPU outputs

This keeps the framework usable for memory-constrained cases without changing the comparison contract.

### Add distributed-aware runtime setup and cleanup

The framework initializes the same distributed/model-parallel context needed for the selected testcase topology,
including 1-GPU and 2-GPU accuracy runs.

It also performs explicit cleanup after each case so that:

- runtime state does not leak between cases
- multi-GPU runs exit cleanly
- failures are easier to localize per testcase

### Add testcase policy for thresholds and unsupported boundaries

The framework also includes testcase policy for:

- threshold overrides where distributed/runtime variance is acceptable
- unsupported reference-loading boundaries that should be skipped rather than reported as false failures

This keeps the reported results meaningful and avoids mixing together:

- real runtime correctness issues
- unsupported comparison boundaries
- expected minor variance in multi-GPU parity runs

In short, this PR adds the first component-accuracy framework for the diffusion runtime and grounds it against the
corresponding Hugging Face reference implementations.

## Checklist

## Review Process


python/sglang/multimodal_gen/test/server/component_accuracy.py
```diff
@@ -0,0 +1,846 @@
...
+
+def _load_reference_component(
+    comp_path: str,
+    local_root: str,
+    component: ComponentType,
+    hub_id: str,
+    pipeline_config,
+    subfolder: str,
+) -> nn.Module:
+    if component == ComponentType.VAE and "wan" in hub_id.lower():
```
BBuf, line 162: I think this WAN VAE reference path is not loading checkpoint weights. Here we instantiate `AutoencoderKLWan(...)` directly, and then later use it as the source for weight transfer into the SGLang module. That means this accuracy test may end up comparing two freshly initialized models instead of the real checkpoint. Could we load the WAN VAE weights explicitly here before running the comparison?

python/sglang/multimodal_gen/test/server/component_accuracy.py
```diff
@@ -0,0 +1,846 @@
...
+        spec = COMPONENT_SPECS[component]
+        if library != spec.reference_library:
+            logger.warning(
+                "Overriding library '%s' with '%s' for component '%s'.",
+                library,
+                spec.reference_library,
+                component.value,
+            )
+            library = spec.reference_library
+        hub_id = case.server_args.model_path
```
BBuf, line 464: This harness seems to ignore component path overrides from the testcase. We always resolve components from `case.server_args.model_path`, but some cases use overrides like `--transformer-path` or `--vae-path`. As written, those variants may still be testing the default component instead of the overridden one. Could we plumb `component_paths` through here so the accuracy test matches the real runtime configuration?

BBuf (commented):
`python/sglang/multimodal_gen/test/server/component_accuracy.py` file is very long, can we simplify it?

python/sglang/multimodal_gen/test/server/component_accuracy.py
Ratish1: fixed

python/sglang/multimodal_gen/test/server/component_accuracy.py
Ratish1: yes, correct. I fixed this

BBuf (approved):
LGTM now.

BBuf:
We should try to keep CI runtime short — ideally each test completes within a few minutes.

BBuf (commented):
Could the naming style be improved? The `_a` and `_b` suffixes don't make the distinction immediately clear, and it's also hard to tell the difference just from reading the code.

BBuf (approved):

PR #18733, opened 2026-02-12
Add DeepSeek V32 PD disaggregation test

ShangmingCai:
## Checklist

## Review Process

### DIFF
--- a/test/registered/distributed/test_disaggregation_deepseek_v32.py
+++ b/test/registered/distributed/test_disaggregation_deepseek_v32.py
@@ -11,7 +11,7 @@ from sglang.test.test_utils import (
     popen_launch_pd_server,
 )
 
-register_cuda_ci(est_time=500, suite="stage-c-test-8-gpu-h200")
+register_cuda_ci(est_time=800, suite="stage-c-test-8-gpu-h200")
 
 FULL_DEEPSEEK_V32_MODEL_PATH = "deepseek-ai/DeepSeek-V3.2-Exp"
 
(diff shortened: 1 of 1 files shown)


test/registered/distributed/test_disaggregation_deepseek_v32.py
```diff
@@ -0,0 +1,102 @@
...
+            "--trust-remote-code",
+            "--disaggregation-mode",
+            "decode",
+            "--tp",
+            "4",
+            "--base-gpu-id",
+            "4",
+            "--cpu-offload-gb",
+            "128",
+            "--disable-cuda-graph",
```
Fridge003: Is `--disable-cuda-graph` necessary? Since in practice we usually enable it for decode node

ShangmingCai: The truth is that we don't have enough GPU mem to run 1 prefill and 1 decode DPSK V32 instance with only 8 H200 GPU, LOL, I am just trying everything I think that might reduce the memory footprint.

ShangmingCai: Seems like "--cpu-offload-gb" cannot help me to get it through, it will report  `RuntimeError: Expected all tensors to be on the same device, but got mat2 is on cpu, different from other tensors on cuda:5 (when checking argument in method wrapper_CUDA_bmm)`.

Do we have a smaller model with the same architecture as DPSK V32?

Fridge003: Maybe `nvidia/DeepSeek-V3.2-NVFP4` ? But that's for b200 and we only have 4-gpu b200 runners... 

One solution would be offload this test to nightly-8-gpu-b200 test

ShangmingCai: Should I put it in the run_suite_nightly.py, or just register it?

Seems like b200 8 gpu does not support isolated runs yet.

PR #18752, opened 2026-02-12
[Fix] Skip signal handler registration when not on main thread

xyuzh:
## Motivation

When `launch_server` is called from a background thread (e.g., inside a Ray actor that runs the server in a `threading.Thread`), Python raises `ValueError` because `signal.signal()` can only be called from the main thread.

## Changes

Guard the SIGQUIT handler registration in `_set_envs_and_config` with a `threading.main_thread()` check. When running on a non-main thread, signal registration is skipped entirely — this is safe because Ray actors have their own lifecycle management for error handling and cleanup.


python/sglang/srt/entrypoints/engine.py
```diff
@@ -853,25 +853,27 @@ def _set_envs_and_config(server_args: ServerArgs):
...
             )
 
-    if server_args.custom_sigquit_handler is None:
-        # Register the signal handler.
-        # The child processes will send SIGQUIT to this process when any error happens
-        # This process then clean up the whole process tree
-        # Note: This sigquit handler is used in the launch phase, and may be replaced by
-        # the running_phase_sigquit_handler in the tokenizer manager after the grpc server is launched.
-        def launch_phase_sigquit_handler(signum, frame):
+    # Signal handlers can only be registered from the main thread.
```
Kangyan-Zhou, line 856: Can you follow tokenizer_manager.py:1461-1468 to add a similar else branch with a warning?

xyuzh: Added

Kangyan-Zhou (approved):

Kangyan-Zhou:
> @Kangyan-Zhou is there any follow up on this? Thanks for reviewing!

Sry for the delayed reply, just merged the PR

PR #18764, opened 2026-02-13
[diffusion] Add dynamic batching v0

qimcis:
## Motivation

#18594 

## Modifications

Added dynamic batching (with max batch size + delay) to the diffusion scheduler. Across the tested text-to-image models, dynamic batching gave up to +29.6% higher throughput, -22.4% lower mean latency, and -31.8% lower P99 latency. 

Currently does not support i2i/edits/i2v, just prompt-only diffusion requests (t2i, t2v). 

## Benchmarking and Profiling

1x NVIDIA H100 80GB
Results (64 prompts):

| Model                              | Throughput Baseline (req/s) | Throughput Dynamic (req/s) | Δ (%)  | Mean Lat Baseline (s) | Mean Lat Dynamic (s) | Δ (%)  | P99 Baseline (s) | P99 Dynamic (s) | Δ (%)  | Duration Baseline (s) | Duration Dynamic (s) | Δ (%)  | Peak Mem Baseline (MB) | Peak Mem Dynamic (MB) | Δ (%)  |
|-------------------------------------|--------------------------|-----------------------------|--------|-------------------|-----------------------|--------|--------------|------------------|--------|-------------------|-----------------------|--------|---------------------|------------------------|--------|
| Tongyi-MAI/Z-Image-Turbo            | 0.377                    | 0.395                       | +4.7%  | 10.351            | 9.960                 | -3.8%  | 13.267       | 10.358           | -21.9% | 169.581           | 161.984               | -4.5%  | 20388.42            | 22215.95               | +9.0%  |
| black-forest-labs/FLUX.1-dev        | 0.496                    | 0.538                       | +8.4%  | 7.877             | 7.437                 | -5.6%  | 10.087       | 9.244            | -8.4%  | 129.055           | 119.062               | -7.7%  | 32839.30            | 34676.05               | +5.6%  |
| Qwen/Qwen-Image-2512                | 0.488                    | 0.633                       | +29.6% | 8.002             | 6.212                 | -22.4% | 10.190       | 6.954            | -31.8% | 131.120           | 101.147               | -22.9% | 57282.23            | 59880.09               | +4.5%  |

For example, on Qwen Image: 

Dynamic Batching:
```
sglang serve --model-path Qwen/Qwen-Image-2512 --backend diffusers --num-gpus 1 --host 127.0.0.1 --port 30200 --log-level info --dynamic-batch-max-size 4 --dynamic-batch-delay-ms 5 --dit-layerwise-offload true
```
```
python -m sglang.multimodal_gen.benchmarks.bench_serving --base-url http://127.0.0.1:30200 --model Qwen/Qwen-Image-2512 --dataset vbench --dataset-path /path/to/prompts --task text-to-image --width 512 --height 512 --num-prompts 64 --max-concurrency 4 --request-rate inf --disable-tqdm
```
Baseline:
```
sglang serve --model-path Qwen/Qwen-Image-2512 --backend diffusers --num-gpus 1 --host 127.0.0.1 --port 30300 --log-level info --dit-layerwise-offload true
```
```
python -m sglang.multimodal_gen.benchmarks.bench_serving --base-url http://127.0.0.1:30300 --model Qwen/Qwen-Image-2512 --dataset vbench --dataset-path /path/to/prompts --task text-to-image --width 512 --height 512 --num-prompts 64 --max-concurrency 4 --request-rate inf --disable-tqdm
```

## Checklist

- [x] Follow the SGLang code style [guidance](https://docs.sglang.io/developer_guide/contribution_guide.html#code-style-guidance)


python/sglang/multimodal_gen/runtime/managers/scheduler.py
```diff
@@ -42,6 +45,28 @@
 
 MINIMUM_PICTURE_BASE64_FOR_WARMUP = "data:image/jpg;base64,iVBORw0KGgoAAAANSUhEUgAAACAAAAAgCAYAAABzenr0AAAACXBIWXMAAA7EAAAOxAGVKw4bAAAAbUlEQVRYhe3VsQ2AMAxE0Y/lIgNQULD/OqyCMgCihCKSG4yRuKuiNH6JLsoEbMACOGBcua9HOR7Y6w6swBwMy0qLTpkeI77qdEBpBFAHBBDAGH8WrwJKI4AAegUCfAKgEgpQDvh3CR3oQCuav58qlAw73kKCSgAAAABJRU5ErkJggg=="
 
+_DYNAMIC_BATCH_SIGNATURE_EXCLUDED_FIELDS = {
```
mickqian: should we make it an attribute to the fields of `Req`, instead of maintaining a list of hard-coded names here?

python/sglang/multimodal_gen/runtime/server_args.py
```diff
@@ -1206,6 +1220,11 @@ def _validate_parallelism(self):
                 f"({self.ring_degree} * {self.ulysses_degree} = {self.ring_degree * self.ulysses_degree})"
             )
 
+        if self.dynamic_batch_max_size < 1:
```
mickqian: this doesn't fall in the category of `parallelism`, isn't it

python/sglang/multimodal_gen/runtime/managers/scheduler.py
```diff
@@ -99,8 +124,15 @@ def __init__(
             GetWeightsChecksumReqInput: self._handle_get_weights_checksum,
         }
 
-        # FIFO, new reqs are appended
-        self.waiting_queue: deque[tuple[bytes, Req]] = deque()
+        # FIFO queue entries: (identity, request, enqueue_ts_s)
+        self.waiting_queue: deque[tuple[bytes | None, Any, float]] = deque()
+        self._dynamic_batch_max_size = max(1, server_args.dynamic_batch_max_size)
+        self._dynamic_batch_delay_s = max(
+            0.0, server_args.dynamic_batch_delay_ms / 1000.0
```
mickqian: `dynamic_batch_delay_ms` is validated and > 0 for sure, so `max` is needless

python/sglang/multimodal_gen/runtime/managers/scheduler.py
```diff
@@ -99,8 +124,15 @@ def __init__(
             GetWeightsChecksumReqInput: self._handle_get_weights_checksum,
         }
 
-        # FIFO, new reqs are appended
-        self.waiting_queue: deque[tuple[bytes, Req]] = deque()
+        # FIFO queue entries: (identity, request, enqueue_ts_s)
+        self.waiting_queue: deque[tuple[bytes | None, Any, float]] = deque()
+        self._dynamic_batch_max_size = max(1, server_args.dynamic_batch_max_size)
```
mickqian: ditto

python/sglang/multimodal_gen/runtime/managers/scheduler.py
```diff
@@ -358,18 +695,27 @@ def event_loop(self) -> None:
             # 2: execute, make sure a reply is always sent
             items = self.get_next_batch_to_run()
             if not items:
+                if self.waiting_queue and self._dynamic_batch_delay_s > 0:
```
mickqian: condition could be simplified

python/sglang/multimodal_gen/runtime/managers/scheduler.py
qimcis: yes this would be better i agree

qimcis:
> Thanks for your great work! Is there any reason or concerns on not supporting i2i/edits requests? Since recently I need to deploy Qwen-Image-Edit on my production environment, I'd be happy to collaborate and contribute on the batch inference of img edit requests.

Valid question! I initially opened this pr as a first implementation of inference batching for sgl diffusion, so I wanted minimal complexity (hence why I focused only on the most common case, t2i, as requests only differ by prompt and seed). I figured that image requests would be more complex to batch. I'd be happy to try to implement this tomorrow though!

The main challenge I could think of off the top of my head is that for i2i/edit batching is that different models handle image conditioning differently (some concatenate spatial conditioning (image/mask latents) with the noisy latent along the channel dimension, others inject conditioning through cross-attention, etc etc) so the batching logic would likely need to be model-specific rather than one general solution. My intuition could be wrong though, lmk what you think @SYChen123

SYChen123:
> Valid question! I initially opened this pr as a first implementation of inference batching for sgl diffusion, so I wanted minimal complexity (hence why I focused only on the most common case, t2i, as requests only differ by prompt and seed). I figured that image requests would be more complex to batch. I'd be happy to try to implement this tomorrow though!
> 
> The main challenge I could think of off the top of my head is that for i2i/edit batching is that different models handle image conditioning differently (some concatenate spatial conditioning (image/mask latents) with the noisy latent along the channel dimension, others inject conditioning through cross-attention, etc etc) so the batching logic would likely need to be model-specific rather than one general solution. My intuition could be wrong though, lmk what you think @SYChen123

@qimcis Hi qimcis. I think from the perspective of design, how the models handle image conditioning should not affect the scheduling logic. For any types of model, we just batch the requests based on perhaps the number of image and text tokens and pass the batch to executor. However, I am new to diffusion so I am not sure if there're any details that I missed. If so feel free to correct me!

Another doubts I have is that will a model simultaneously execute different types of reqs? For example, a model is able to execute t2i and i2i requests simultaneously. Not sure will such case happens. In such case, shall we batch t2i and i2i requests separately? Since i2i request usually takes longer time to generate and will slow down t2i requests if in 1 batch.

niehen6174:
Thanks for the great work on adding dynamic batching support for diffusion models.

I noticed that the current evaluation is primarily based on 512×512 resolution. At this scale, many diffusion workloads tend to be more memory-bound, where batching optimizations can show clearer benefits.

I’m wondering if it might also be helpful to include benchmarks at higher resolutions, such as 1024×1024, which are more commonly used in practice. At these larger resolutions, workloads often become more compute-bound, and the impact of dynamic batching may differ.

From some quick experiments on my side, I observed that for Qwen-Image at 1024×1024, the workload still appears to be largely memory-bound, and the benefit from batching is relatively limited. I’ve shared some test results here for reference:
https://github.com/sgl-project/sglang/pull/22183#issuecomment-4224352520

It might be useful to run similar tests to better understand how the optimization behaves under this setting.

JoeyTPChou:
I can't tell how much I'm excited for the dynamic batch feature! Do you know any ETA for this PR?

qimcis:
> I can't tell how much I'm excited for the dynamic batch feature! Do you know any ETA for this PR?

Thank you for your interest! Hopefully within the next couple of days

qimcis:
> Thanks for the great work on adding dynamic batching support for diffusion models.
> 
> I noticed that the current evaluation is primarily based on 512×512 resolution. At this scale, many diffusion workloads tend to be more memory-bound, where batching optimizations can show clearer benefits.
> 
> I’m wondering if it might also be helpful to include benchmarks at higher resolutions, such as 1024×1024, which are more commonly used in practice. At these larger resolutions, workloads often become more compute-bound, and the impact of dynamic batching may differ.
> 
> From some quick experiments on my side, I observed that for Qwen-Image at 1024×1024, the workload still appears to be largely memory-bound, and the benefit from batching is relatively limited. I’ve shared some test results here for reference: [#22183 (comment)](https://github.com/sgl-project/sglang/pull/22183#issuecomment-4224352520)
> 
> It might be useful to run similar tests to better understand how the optimization behaves under this setting.

Thanks for suggesting this. I tested at 1024x1024 for select models (results below). As expected, the 1024x1024 results behaved differently from the lower res runs. In this case the issue is not just whether the scheduler can form full batches. At 1024x1024, some models already keep the GPU busy per request, or scale activation/memory work almost linearly with batch size. In those cases, even when batches are full it gives little or no useful throughput gain, but increases memory pressure and queueing latency.

The results below show that dynamic batching is model- and resolution-dependent. Models like SANA (that have more headroom for batching), performed much better than models whose per-request workload already saturated the GPU.

("cap" is maximum batch size allowed)

all on 1xh200 sxm
P95 includes queueing; concurrency is shown in each cell.
selected results below:
| model | res | baseline | dynamic | qps delta | P95 delta | peak memory delta |
|---|---:|---|---|---:|---:|---:|
| Qwen/Qwen-Image-2512 | 512x512 | cap1 C16: 0.1149 qps, 140.08s P95, 55.6 GiB | cap8 C16: 0.3416 qps, 47.85s P95, 73.3 GiB | +197.3% | -65.8% | +17.7 GiB |
| Qwen/Qwen-Image-2512 | 1024x1024 | cap1 C16: 0.0774 qps, 206.87s P95, 63.3 GiB | cap2 C16: 0.0810 qps, 197.64s P95, 73.5 GiB | +4.7% | -4.5% | +10.2 GiB |
| black-forest-labs/FLUX.1-dev | 512x512 | cap1 C16: 0.4529 qps, 35.39s P95, 33.8 GiB | cap8 C16: 0.5121 qps, 31.27s P95, 47.5 GiB | +13.1% | -11.6% | +13.7 GiB |
| black-forest-labs/FLUX.1-dev | 1024x1024 | cap1 C16: 0.1465 qps, 109.24s P95, 40.2 GiB | cap8 C16: 0.1536 qps, 104.21s P95, 96.4 GiB | +4.8% | -4.6% | +56.2 GiB |
| black-forest-labs/FLUX.2-klein-4B | 1024x1024 | cap1 C16: 2.4498 qps, 6.60s P95, 19.7 GiB | cap8 C24: 2.7422 qps, 9.12s P95, 50.7 GiB | +11.9% | +38.3% | +31.0 GiB |
| Efficient-Large-Model/SANA1.5_1.6B_1024px_diffusers | 1024x1024 | cap1 C16: 1.3342 qps, 12.18s P95, 12.0 GiB | cap4 C16: 1.8960 qps, 8.38s P95, 22.0 GiB | +42.1% | -31.2% | +10.0 GiB |
| Efficient-Large-Model/SANA1.5_1.6B_1024px_diffusers | 1024x1024 | cap1 C16: 1.3342 qps, 12.18s P95, 12.0 GiB | cap8 C24: 1.9816 qps, 13.41s P95, 35.4 GiB | +48.5% | +10.1% | +23.4 GiB |
| Efficient-Large-Model/SANA1.5_4.8B_1024px_diffusers | 1024x1024 | cap1 C16: 0.5136 qps, 31.67s P95, 17.9 GiB | cap4 C16: 0.7026 qps, 22.54s P95, 27.9 GiB | +36.8% | -28.9% | +10.0 GiB |
| Efficient-Large-Model/SANA1.5_4.8B_1024px_diffusers | 1024x1024 | cap1 C16: 0.5136 qps, 31.67s P95, 17.9 GiB | cap8 C24: 0.7425 qps, 33.81s P95, 41.3 GiB | +44.6% | +6.8% | +23.4 GiB |
| Tongyi-MAI/Z-Image-Turbo | 1024x1024 | cap1 C16: 1.0370 qps, 15.55s P95, 28.5 GiB | cap8 C24: 1.0576 qps, 22.77s P95, 77.3 GiB | +2.0% | +46.4% | +48.8 GiB |

python/sglang/multimodal_gen/runtime/managers/scheduler.py
```diff
@@ -414,28 +963,26 @@ def recv_reqs(self) -> List[tuple[bytes, Any]]:
...
+
+                    try:
+                        identity, payload = parts[0], parts[-1]
+                        reqs = pickle.loads(payload) if len(parts) > 2 else []
+                    except (pickle.UnpicklingError, IndexError, EOFError):
+                        continue
+
+                    if not isinstance(reqs, list):
+                        reqs = [reqs]
+                    recv_reqs.extend((identity, req) for req in reqs)
```
mickqian: Could we keep client-submitted `list[Req]` groups atomic here and use a separate internal representation for scheduler-created dynamic batches? after #23759, `list[Req]` can be a single multi-output request expecting one reply; flattening it here may split that request into multiple replies to the same socket (request) and bypass the grouped/stage-wise path

mickqian (commented):
left one scheduler semantics suggestion

python/sglang/multimodal_gen/configs/pipeline_configs/base.py
```diff
@@ -468,6 +499,87 @@ def get_text_encoder_attention_mask(
...
+            "text attention mask length does not match postprocessed prompt embeddings. "
+            "Postprocess functions that trim, pack, or otherwise change text sequence "
+            "length must return TextConditioningOutput with an embedding-aligned mask."
+        )
+
+    @staticmethod
+    def seq_lens_from_text_conditioning_mask(mask: "torch.Tensor") -> list[int]:
+        if mask.ndim != 2:
+            raise ValueError("text conditioning mask must have shape [batch, seq]")
+        return [int(x) for x in mask.to(torch.int64).sum(dim=1).tolist()]
```
mickqian: this method could be made faster

python/sglang/multimodal_gen/configs/pipeline_configs/flux.py
```diff
@@ -175,6 +218,12 @@ def post_denoising_loop(self, latents, batch):
         return latents
 
     def prepare_pos_cond_kwargs(self, batch, device, rotary_emb, dtype):
+        txt_seq_lens = self.require_text_seq_lens(
```
mickqian, line 231: add more documentation here

python/sglang/multimodal_gen/configs/pipeline_configs/qwen_image.py
```diff
@@ -305,35 +289,87 @@ def _prepare_cond_kwargs(self, batch, prompt_embeds, rotary_emb, device, dtype):
...
+        self,
+        batch,
+        encoder_index: int,
+        txt_seq_lens: list[int],
+        text_seq_len: int,
+        batch_size: int,
+        *,
+        negative: bool = False,
+    ):
+        if all(seq_len == text_seq_len for seq_len in txt_seq_lens):
```
mickqian, line 392: also add some docs here, illustrating:
1. what do we need this
2. what does it return
3. how do we build this mask

python/sglang/multimodal_gen/configs/pipeline_configs/qwen_image.py
```diff
@@ -305,35 +289,87 @@ def _prepare_cond_kwargs(self, batch, prompt_embeds, rotary_emb, device, dtype):
...
+        if masks_by_encoder is not None and encoder_index < len(masks_by_encoder):
+            mask = masks_by_encoder[encoder_index]
+            if mask.shape != (batch_size, text_seq_len):
+                raise ValueError(
+                    "QwenImage text conditioning mask has shape "
+                    f"{tuple(mask.shape)}, expected {(batch_size, text_seq_len)}."
+                )
+            return mask
+
+        positions = torch.arange(text_seq_len, device=batch.prompt_embeds[0].device)
```
mickqian, line 408: maybe cache a global positions in the future

python/sglang/multimodal_gen/configs/pipeline_configs/qwen_image.py
```diff
@@ -305,35 +289,87 @@ def _prepare_cond_kwargs(self, batch, prompt_embeds, rotary_emb, device, dtype):
                 )
             ]
         ] * batch_size
-        txt_seq_lens = [prompt_embeds[0].shape[1]]
+        txt_seq_lens = self.require_text_seq_lens(
+            batch, 0, negative=negative, expected_batch_size=batch_size
+        )
+        encoder_hidden_states_mask = self._prepare_encoder_hidden_states_mask(
```
mickqian: the mask solution to deal with variable-length input is fine for a v0, but since this solution introduce overhead, in the long run, we should consider let each model declares its (text-conditioning) layout as either of:
1. mask
2. fixed-length (like flux Klein)
3. PackedVarLen with ragged format (the most ideal one)

qimcis: ok sounds good, i can address this in a follow-up pr

python/sglang/multimodal_gen/runtime/managers/scheduler.py
```diff
@@ -56,6 +63,24 @@
 # _combine_cfg_parallel's all-reduce.
 DEFAULT_PLACEHOLDER_PROMPT = "warmup"
 
+_MAX_RECV_REQS_PER_POLL = 1024
+_BATCH_METRICS_LOG_INTERVAL = 5
+
+
+@dataclasses.dataclass
+class _BatchMetricsWindow:
```
mickqian: could we avoid class name prefixed with `_`? and this seems like belong to `schedule_batch.py`

python/sglang/multimodal_gen/runtime/managers/scheduler.py
```diff
@@ -285,15 +543,322 @@ def return_result(
...
+            )
+            if split.metrics is not None:
+                split.metrics.request_id = req.request_id
+            outputs.append(split)
+            start = end
+
+        return outputs
+
+    def _dynamic_batching_enabled(self) -> bool:
+        return self._batching_max_size > 1
```
mickqian: how do we decide if a model supports batching now? could you point it out for me

python/sglang/multimodal_gen/runtime/managers/dynamic_batch_admission.py
```diff
@@ -0,0 +1,396 @@
...
+            hints.append(f"{key!r} (did you mean {matches[0]!r}?)")
+        else:
+            hints.append(repr(key))
+    raise ValueError(
+        f"batching config rule from {source} contains unknown key(s): "
+        f"{', '.join(hints)}"
+    )
+
+
+def _resolution_key(req: Req) -> str | None:
```
mickqian: this is better be made a method of `Req`

docs/diffusion/performance/dynamic_batching.md
```diff
@@ -0,0 +1,114 @@
+# Dynamic Batching
+
+Dynamic batching is an opt-in mode for `sglang.multimodal_gen` that merges compatible queued diffusion requests into one native pipeline batch. It is separate from LLM continuous batching and tokenizer batching.
```
mickqian: nit: `SGLang-Diffusion`

python/sglang/multimodal_gen/configs/pipeline_configs/base.py
```diff
@@ -329,6 +368,41 @@ def get_latent_dtype(self, prompt_dtype: torch.dtype) -> torch.dtype:
...
+        return self.task_type in (ModelTaskType.T2I, ModelTaskType.T2V)
+
+    def estimate_request_cost(self, batch) -> float:
+        """Return the relative cost used for batching admission caps.
+
+        This is compared with `max_cost` from the batching config; it is not a
+        memory estimate. The default cost is latent tokens times frames times
+        outputs; pipelines can override it for model-specific admission.
+        """
+        latent_tokens = float(getattr(batch, "n_tokens", None) or 0)
```
mickqian: avoid getattr if we are sure about the static type

mickqian (approved):
awesome. this could be merged as a decent v0, but still needs further polishing in the future

desenSunUBW:
Hi, do you consider adding the support for mixed resolution batching strategy?

PR #18802, opened 2026-02-13
Fix DataParallelController and BatchTokenizedGenerateReqInput API

akhoroshev:
If you start a server with a DP size of 2 (or more) and send a BatchTokenizedGenerateReqInput request this request will be broadcast to every Scheduler of each data parallel worker.


python/sglang/srt/managers/data_parallel_controller.py
akhoroshev, line 224: The current callback (send_control_message) for BatchTokenizedGenerateReqInput/BatchTokenizedEmbeddingReqInput is incorrect

akhoroshev:
fixed later in https://github.com/sgl-project/sglang/commit/95327458ee361a81c5d28ff242538387eeebcb3f

PR #18806, opened 2026-02-13
[diffusion] feat: add rollout `log_prob` with flow-matching SDE/CPS support

MikukuOvO:
## Motivation

This PR adds rollout `log_prob` support for diffusion flow-matching pipelines.  
Previously, rollout paths did not expose consistent `log_prob` signals for flow-matching variants (especially SDE/CPS), which limited downstream training/evaluation workflows that depend on likelihood-based objectives.

## Modifications

- Add rollout `log_prob` computation in the diffusion rollout path.
- Add flow-matching rollout support for SDE mode.
- Add flow-matching rollout support for CPS mode.
- Expose `log_prob` in rollout outputs with consistent shape/semantics across modes.
- Keep backward compatibility for existing rollout callers when `log_prob` is not requested.
- Add/extend related tests and validations for shape/dtype, mode coverage (SDE/CPS), and regression behavior.

## Accuracy Tests

All tests were run with fixed random seeds.  
Under the same seed and configuration, rollout outputs are deterministic: `latent` and `log_prob` are consistent across repeated runs.

### Flux

- Sampling modes tested: `no rollout`, `CPS`, `SDE`.
- Verified deterministic consistency of both `latent` and `log_prob` under fixed seed.
- Figure: sampling_comparison_figure_v2

### Qwen

- Sampling modes tested: `no rollout`, `CPS`, `SDE`.
- Verified deterministic consistency of both `latent` and `log_prob` under fixed seed.
- Figure: sampling_comparison_figure

### Z-Image

- Sampling modes tested: `no rollout`, `CPS`, `SDE`.
- Verified deterministic consistency of both `latent` and `log_prob` under fixed seed.
- Figure: sampling_comparison_figure_v3_compressed

## Benchmarking and Profiling

- No dedicated speed benchmark/profiling numbers are included in this PR yet.
- Functional focus of this PR is rollout `log_prob` correctness and mode coverage.

## Checklist

## Review Process


python/sglang/multimodal_gen/runtime/pipelines_core/stages/denoising.py
```diff
@@ -1085,13 +1119,25 @@ def forward(
...
-                        latents = self.scheduler.step(
-                            model_output=noise_pred,
-                            timestep=t_device,
-                            sample=latents,
-                            **extra_step_kwargs,
-                            return_dict=False,
-                        )[0]
+                        if rollout_enabled:
+                            latents, step_log_prob, _, _ = (
+                                self.scheduler.sde_step_with_logprob(
```
Rockdu: We should expose noise_level args to inference interface, which is a critical RL argument

python/sglang/multimodal_gen/runtime/pipelines/patches/flow_matching_with_logprob.py
```diff
@@ -0,0 +1,115 @@
...
+
+    sigma = self.sigmas[step_indices].to(sample.device).to(sample.dtype)
+    sigma_prev = self.sigmas[prev_step_indices].to(sample.device).to(sample.dtype)
+    sigma = sigma.view(-1, *([1] * (sample.ndim - 1)))
+    sigma_prev = sigma_prev.view(-1, *([1] * (sample.ndim - 1)))
+    sigma_max = self.sigmas[min(1, len(self.sigmas) - 1)].item()
+    dt = sigma_prev - sigma
+
+    if sde_type == "sde":
+        denom_sigma = 1 - torch.where(sigma == 1, torch.as_tensor(sigma_max), sigma)
```
dreamyang-liu: sigma is on gpu and torch.as_tensor(sigma_max) is on cpu, will it cause any issue ?

python/sglang/multimodal_gen/runtime/pipelines/patches/flow_matching_with_logprob.py
```diff
@@ -0,0 +1,115 @@
...
+    batch_size = sample.shape[0]
+    timestep_tensor = _as_timestep_tensor(timestep, batch_size, sample.device)
+    step_indices = torch.tensor(
+        [self.index_for_timestep(t.item()) for t in timestep_tensor],
+        device=sample.device,
+        dtype=torch.long,
+    )
+    prev_step_indices = (step_indices + 1).clamp_max(len(self.sigmas) - 1)
+
+    sigma = self.sigmas[step_indices].to(sample.device).to(sample.dtype)
```
dreamyang-liu: It's currently working but it highly depends on this line
https://github.com/sgl-project/sglang/blob/main/python/sglang/multimodal_gen/runtime/models/schedulers/scheduling_flow_match_euler_discrete.py#L409

Might be good to make sure step_indices and self.sigma on same device.

python/sglang/multimodal_gen/runtime/pipelines/patches/flow_matching_with_logprob.py
```diff
@@ -0,0 +1,115 @@
...
+
+    if ts.ndim == 0:
+        ts = ts.view(1)
+    else:
+        ts = ts.view(-1)
+
+    if ts.numel() == 1 and batch_size > 1:
+        ts = ts.repeat(batch_size)
+    elif ts.numel() != batch_size:
+        ts = ts[:1].repeat(batch_size)
```
dreamyang-liu: In which case it will come to this branch ? Do we really need it ?

python/sglang/multimodal_gen/runtime/pipelines/patches/flow_matching_with_logprob.py
```diff
@@ -0,0 +1,115 @@
...
+"""Flow-matching rollout step utilities for log-prob computation."""
+
+import math
+from typing import Any, Optional, Union
+
+import torch
+from diffusers.utils.torch_utils import randn_tensor
+
+
+def _as_timestep_tensor(
```
zhaochenyang20: Add comments for exaplanation to this funciton/

python/sglang/multimodal_gen/runtime/pipelines/patches/flow_matching_with_logprob.py
zhaochenyang20: Nice catch

python/sglang/multimodal_gen/runtime/pipelines/patches/flow_matching_with_logprob.py
```diff
@@ -0,0 +1,115 @@
...
+    sample: torch.FloatTensor,
+    noise_level: float = 0.7,
+    prev_sample: Optional[torch.FloatTensor] = None,
+    generator: Optional[Union[torch.Generator, list[torch.Generator]]] = None,
+    sde_type: str = "sde",
+):
+    """One rollout step with log-prob estimate.
+
+    Supports the two variants used by flow_grpo patches: `sde` and `cps`.
+    """
```
zhaochenyang20: 1. do not use "`" in your python comment.
2. make this comments clearer.

python/sglang/multimodal_gen/runtime/pipelines/patches/flow_matching_with_logprob.py
```diff
@@ -0,0 +1,115 @@
...
+    if prev_sample is not None:
+        prev_sample = prev_sample.float()
+
+    batch_size = sample.shape[0]
+    timestep_tensor = _as_timestep_tensor(timestep, batch_size, sample.device)
+    step_indices = torch.tensor(
+        [self.index_for_timestep(t.item()) for t in timestep_tensor],
+        device=sample.device,
+        dtype=torch.long,
+    )
```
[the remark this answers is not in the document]
zhaochenyang20: This suggestion is reasonable. As we said in our code style guide:

Minimize Synchronization: Strictly avoid frequent calls to .item(), .cpu(), or .tolist() within the model inference path.

https://github.com/zhaochenyang20/sglang-diffusion-routing/issues/32

python/sglang/multimodal_gen/runtime/pipelines_core/stages/denoising.py
```diff
@@ -110,6 +113,10 @@ def __init__(
             self._maybe_enable_torch_compile(transformer)
 
         self.scheduler = scheduler
+        if not hasattr(self.scheduler, "sde_step_with_logprob"):
+            self.scheduler.sde_step_with_logprob = sde_step_with_logprob.__get__(
+                self.scheduler, type(self.scheduler)
+            )
```
zhaochenyang20: Avoid using `has_attr`, as we said in:

https://github.com/zhaochenyang20/sglang-diffusion-routing/issues/32

Avoid Dynamic Attributes: Minimize the use of getattr or setattr. Code should be explicit for better traceability. https://github.com/THUDM/slime/pull/1141#discussion_r2654651150 https://github.com/THUDM/slime/pull/1141#discussion_r2658247124

python/sglang/multimodal_gen/runtime/pipelines_core/stages/denoising.py
```diff
@@ -998,6 +1012,29 @@ def forward(
         # Initialize lists for ODE trajectory
         trajectory_timesteps: list[torch.Tensor] = []
         trajectory_latents: list[torch.Tensor] = []
+        trajectory_log_probs: list[torch.Tensor] = []
+        rollout_enabled = bool(batch.rollout)
+        rollout_sde_type = getattr(batch, "rollout_sde_type", None)
```
zhaochenyang20: avoid using `get_attr`

zhaochenyang20 (changes requested):
1. Fix the lint.
2. Could you add unit test to SDE and CPS. These APIs are important.

zhaochenyang20:
unit test could be like:

https://github.com/sgl-project/sglang/issues/19164#issuecomment-3954951773

python/sglang/multimodal_gen/runtime/pipelines/patches/flow_matching_with_logprob.py
MikukuOvO: Good catch. You are right that torch.as_tensor(sigma_max) can create a CPU tensor while sigma is on GPU. I fixed this by keeping sigma_max as a tensor on the same device/dtype as sample/sigma, and removed the .item() host sync path as well.

python/sglang/multimodal_gen/runtime/pipelines/patches/flow_matching_with_logprob.py
MikukuOvO: Good point. I kept index tensors on self.sigmas.device (required for indexing), and cast sigma values to sample.device for computation. This makes device handling explicit and avoids hidden assumptions.

python/sglang/multimodal_gen/runtime/pipelines/patches/flow_matching_with_logprob.py
MikukuOvO: Thanks for the catch. I removed that fallback branch (ts = ts[:1].repeat(batch_size)) to avoid silently masking invalid timestep shapes.
Now _as_timestep_tensor only broadcasts the scalar case (numel() == 1) and no longer applies that forced fallback for mismatched lengths.

python/sglang/multimodal_gen/runtime/pipelines/patches/flow_matching_with_logprob.py
MikukuOvO: Added!

python/sglang/multimodal_gen/runtime/pipelines/patches/flow_matching_with_logprob.py
MikukuOvO: Updated. I removed backticks in the docstring and clarified the behavior for both rollout modes: SDE uses Gaussian transition log_prob, while CPS uses the simplified CPS objective from the rollout patch.

python/sglang/multimodal_gen/runtime/pipelines_core/stages/denoising.py
MikukuOvO: Fixed!

python/sglang/multimodal_gen/runtime/pipelines_core/stages/denoising.py
MikukuOvO: Fixed!

python/sglang/multimodal_gen/runtime/pipelines_core/schedule_batch.py
```diff
@@ -132,6 +132,7 @@ class Req:
 
     trajectory_timesteps: list[torch.Tensor] | None = None
     trajectory_latents: torch.Tensor | None = None
+    trajectory_log_probs: torch.Tensor | None = None
```
dreamyang-liu: In python/sglang/multimodal_gen/runtime/pipelines_core/stages/denoising.py

We have `trajectory_log_probs: list[torch.Tensor] = []` but here it's torch.Tensor, it this correct ?

python/sglang/multimodal_gen/runtime/pipelines/patches/flow_matching_with_logprob.py
```diff
@@ -0,0 +1,120 @@
...
+            )
+            prev_sample = prev_sample_mean + std_dev_t * variance_noise
+
+        # Keep the same simplified cps objective used in the original patch.
+        log_prob = -((prev_sample.detach() - prev_sample_mean) ** 2)
+    else:
+        raise ValueError(f"Unsupported sde_type: {sde_type}")
+
+    log_prob = log_prob.mean(dim=tuple(range(1, log_prob.ndim)))
+    return prev_sample.to(sample_dtype), log_prob, prev_sample_mean, std_dev_t
```
dreamyang-liu: Do we need to return `prev_sample_mean, std_dev_t` ? 

Seems we don't use it `latents, step_log_prob, _, _ `

python/sglang/multimodal_gen/runtime/pipelines_core/schedule_batch.py
dreamyang-liu: NVM, seems it's a issue with trajectory_timesteps

python/sglang/multimodal_gen/runtime/pipelines_core/stages/denoising.py
```diff
@@ -998,6 +1011,29 @@ def forward(
...
+        if rollout_sde_type not in ("sde", "cps"):
+            logger.warning(
+                "Unknown rollout_sde_type '%s', using default 'sde'.",
+                rollout_sde_type,
+            )
+            rollout_sde_type = "sde"
+
+        rollout_noise_level = batch.rollout_noise_level
+
+        if rollout_enabled and not hasattr(self.scheduler, "sde_step_with_logprob"):
```
dreamyang-liu: Do we really need it ? 
We already set during __init__
```
self.scheduler.sde_step_with_logprob = sde_step_with_logprob.__get__(
            self.scheduler, type(self.scheduler)
        )
```

python/sglang/multimodal_gen/runtime/pipelines/patches/flow_matching_with_logprob.py
```diff
@@ -0,0 +1,120 @@
...
+    sigma_prev = self.sigmas[prev_step_indices].to(sample.device).to(sample.dtype)
+    sigma = sigma.view(-1, *([1] * (sample.ndim - 1)))
+    sigma_prev = sigma_prev.view(-1, *([1] * (sample.ndim - 1)))
+    sigma_max = self.sigmas[min(1, len(self.sigmas) - 1)].to(
+        device=sample.device, dtype=sample.dtype
+    )
+    dt = sigma_prev - sigma
+
+    if sde_type == "sde":
+        denom_sigma = 1 - torch.where(sigma == 1, sigma_max, sigma)
```
dreamyang-liu: For sigma == 1, can we use `torch.isclose` given sigma is float tensor

Like 

```
denom_sigma = 1 - torch.where(
            torch.isclose(sigma, sigma.new_tensor(1.0)), sigma_max, sigma
        )
```

python/sglang/multimodal_gen/runtime/pipelines/patches/flow_matching_with_logprob.py
```diff
@@ -0,0 +1,120 @@
...
+    prev_step_indices = prev_step_indices.to(device=self.sigmas.device)
+
+    sigma = self.sigmas[step_indices].to(sample.device).to(sample.dtype)
+    sigma_prev = self.sigmas[prev_step_indices].to(sample.device).to(sample.dtype)
+    sigma = sigma.view(-1, *([1] * (sample.ndim - 1)))
+    sigma_prev = sigma_prev.view(-1, *([1] * (sample.ndim - 1)))
+    sigma_max = self.sigmas[min(1, len(self.sigmas) - 1)].to(
+        device=sample.device, dtype=sample.dtype
+    )
+    dt = sigma_prev - sigma
```
dreamyang-liu: Seems that there's a roundtrip for step_indices (sample -> sigma) but sigma also move to sample device later. Can we simplify this ?

python/sglang/multimodal_gen/runtime/pipelines/patches/flow_matching_with_logprob.py
```diff
@@ -0,0 +1,120 @@
...
+    if prev_sample is not None:
+        prev_sample = prev_sample.float()
+
+    batch_size = sample.shape[0]
+    timestep_tensor = _as_timestep_tensor(timestep, batch_size, sample.device)
+    step_indices = torch.tensor(
+        [self.index_for_timestep(t.to(self.timesteps.device)) for t in timestep_tensor],
+        device=sample.device,
+        dtype=torch.long,
+    )
```
zhaochenyang20: This is concerning:

Minimize Synchronization: Strictly avoid frequent calls to .item(), .cpu(), or .tolist() within the model inference path.

`index_for_timestep` internally calls `.item()` to move a GPU tensor to the CPU and performs element-wise comparisons using `==`, which triggers synchronization. This list comprehension is invoked for every sample in the batch, causing a GPU-CPU round trip each time.

The timestep-to-index mapping should be pre-calculated outside the denoising loop, or the step index should be passed in directly to avoid runtime lookups.

python/sglang/multimodal_gen/runtime/pipelines_core/stages/denoising.py
```diff
@@ -110,6 +113,9 @@ def __init__(
             self._maybe_enable_torch_compile(transformer)
 
         self.scheduler = scheduler
+        self.scheduler.sde_step_with_logprob = sde_step_with_logprob.__get__(
+            self.scheduler, type(self.scheduler)
+        )
```
zhaochenyang20: Avoid Dynamic Attributes

In `__init__`, an external function is bound as a method to the scheduler using `__get__`, which is a classic case of monkey-patching. Our code style guide explicitly mandates: "Minimize the use of getattr or setattr. Code should be explicit for better traceability."

zhaochenyang20 (changes requested):

python/sglang/multimodal_gen/runtime/pipelines_core/stages/denoising.py
zhaochenyang20: Since the attribute is already set during initialization, this hasattr check will never be triggered.

python/sglang/multimodal_gen/runtime/pipelines/patches/flow_matching_with_logprob.py
zhaochenyang20: This is important

python/sglang/multimodal_gen/runtime/pipelines/patches/flow_matching_with_logprob.py
zhaochenyang20: `prev_sample_mean` and `std_dev_t` are never used. Returning unused values increases cognitive load and violates the Principle of Least Interface.

zhaochenyang20 (changes requested):

python/sglang/multimodal_gen/runtime/pipelines_core/stages/denoising.py
```diff
@@ -998,6 +1011,29 @@ def forward(
...
+                logger.warning("rollout_sde_type is not set, defaulting to 'sde'.")
+            rollout_sde_type = "sde"
+        else:
+            rollout_sde_type = str(rollout_sde_type).strip().lower()
+        if rollout_sde_type not in ("sde", "cps"):
+            logger.warning(
+                "Unknown rollout_sde_type '%s', using default 'sde'.",
+                rollout_sde_type,
+            )
+            rollout_sde_type = "sde"
```
zhaochenyang20: `SamplingParams` already defines `default="sde"` and the CLI arguments are restricted to `choices=["sde", "cps"]`. The subsequent `None` checks, empty string checks, `.strip()`, and `.lower()` calls are textbook examples of over-defensive programming.

Following our code style guide’s "Do not overprotect" principle: if upstream logic already guarantees valid values, do not repeat the validation downstream. Furthermore, if validation is strictly necessary, the code should raise an exception rather than performing a silently fallback.

zhaochenyang20 (changes requested):

python/sglang/multimodal_gen/runtime/pipelines/patches/flow_matching_with_logprob.py
```diff
@@ -0,0 +1,120 @@
...
+            variance_noise = randn_tensor(
+                model_output.shape,
+                generator=generator,
+                device=model_output.device,
+                dtype=model_output.dtype,
+            )
+            prev_sample = prev_sample_mean + std_dev_t * variance_noise
+
+        # Keep the same simplified cps objective used in the original patch.
+        log_prob = -((prev_sample.detach() - prev_sample_mean) ** 2)
```
zhaochenyang20: This is merely a negative MSE without the normalization constant; it is not a mathematically valid log-probability. Although the comment mentions a "simplified CPS objective," the name `log_prob` is misleading. In RLHF scenarios, if upstream code assumes that `log_prob` carries the formal semantics of a log-probability (e.g., for calculating importance sampling ratios), the CPS mode will yield incorrect results.

python/sglang/multimodal_gen/runtime/pipelines_core/stages/denoising.py
```diff
@@ -731,6 +742,8 @@ def _post_denoising_loop(
         if trajectory_tensor is not None and trajectory_timesteps_tensor is not None:
             batch.trajectory_timesteps = trajectory_timesteps_tensor.cpu()
             batch.trajectory_latents = trajectory_tensor.cpu()
+        if trajectory_log_probs_tensor is not None:
```
zhaochenyang20: No sure wether this works in SP. We can leave it to the future.

zhaochenyang20 (changes requested):

python/sglang/multimodal_gen/runtime/pipelines_core/schedule_batch.py
MikukuOvO: Thanks for catching this. You’re right — trajectory_timesteps is stacked and stored as a tensor in post-processing, so I’ve updated the type annotations in Req/OutputBatch to torch.Tensor | None for consistency.

python/sglang/multimodal_gen/runtime/pipelines/patches/flow_matching_with_logprob.py
MikukuOvO: Good catch. prev_sample_mean and std_dev_t were unused at call sites. I simplified sde_step_with_logprob to return only (prev_sample, log_prob) and updated denoising accordingly.

python/sglang/multimodal_gen/runtime/pipelines_core/stages/denoising.py
MikukuOvO: Good point. This check is redundant because sde_step_with_logprob is already bound in __init__. I removed the hasattr guard.

python/sglang/multimodal_gen/runtime/pipelines/patches/flow_matching_with_logprob.py
MikukuOvO: Updated. Replaced the sigma == 1 check with torch.isclose(sigma, sigma.new_tensor(1.0)) to avoid exact float equality.

python/sglang/multimodal_gen/runtime/pipelines/patches/flow_matching_with_logprob.py
MikukuOvO: Good point. I simplified this by creating step_indices directly on self.sigmas.device and removing the unnecessary device roundtrip before indexing sigmas.

python/sglang/multimodal_gen/runtime/pipelines_core/stages/denoising.py
MikukuOvO: Good point. I removed the dynamic method binding and switched to an explicit function call (sde_step_with_logprob(self.scheduler, ...)) for better traceability.

python/sglang/multimodal_gen/runtime/pipelines_core/stages/denoising.py
MikukuOvO: Agreed. I removed the downstream normalization/fallback logic and now use rollout_sde_type as provided by upstream. Invalid values will fail explicitly via ValueError in sde_step_with_logprob instead of silently falling back.

python/sglang/multimodal_gen/runtime/pipelines/patches/flow_matching_with_logprob.py
MikukuOvO: Updated. I moved timestep->index mapping out of the inner inference path: step indices are now precomputed once before the denoising loop and passed directly into sde_step_with_logprob (as a required step_index). This removes per-sample index_for_timestep lookups and avoids repeated sync in rollout.

python/sglang/multimodal_gen/runtime/pipelines/patches/flow_matching_with_logprob.py
MikukuOvO: Fixed!

MikukuOvO:
Thanks for the reviews! I've gone through all your comments and pushed the fixes.

zhaochenyang20:
This is my verification commands:

1. Install the changes:

```bash
cd python 
uv pip install -e ".[diffusion]"
```

2. With Python. Note that only python API can get `log_probs`.

```python
from sglang import DiffGenerator

  gen = DiffGenerator.from_pretrained("Wan-AI/Wan2.1-T2V-1.3B-Diffusers")

  # Mode 1: No Rollout (baseline)
  result_baseline = gen.generate(sampling_params_kwargs={
      "prompt": "A curious raccoon in a forest",
      "rollout": False, "save_output": True,
  })

  # Mode 2: SDE Rollout
  result_sde = gen.generate(sampling_params_kwargs={
      "prompt": "A curious raccoon in a forest",
      "rollout": True, "rollout_sde_type": "sde",
      "rollout_noise_level": 0.7, "return_trajectory_latents": True, "save_output": True,
  })

  # Mode 3: CPS Rollout
  result_cps = gen.generate(sampling_params_kwargs={
      "prompt": "A curious raccoon in a forest",
      "rollout": True, "rollout_sde_type": "cps",
      "rollout_noise_level": 0.7, "return_trajectory_latents": True, "save_output": True,
  })

  results = []
  for i in range(2):
      r = gen.generate(sampling_params_kwargs={
          "prompt": "A curious raccoon in a forest",
          "rollout": True, "rollout_sde_type": "sde",
          "rollout_noise_level": 0.7, "return_trajectory_latents": True, "seed": 42,
      })
      results.append(r)

  import torch
  print(f"Log_probs match: {torch.allclose(results[0].trajectory_log_probs, results[1].trajectory_log_probs)}")
  print(f"Latents match: {torch.allclose(results[0].trajectory_latents, results[1].trajectory_latents)}")
```

3. With Server

```bash
sglang serve --model-path Wan-AI/Wan2.1-T2V-1.3B-Diffusers --num-gpus 1
```

```bash
  # No Rollout                                                                                                                                         
  curl http://localhost:30000/v1/images/generations \             
    -d '{                                                         
      "prompt": "A curious raccoon in a forest",
      "model": "Wan-AI/Wan2.1-T2V-1.3B-Diffusers"
    }'

  # SDE Rollout
  curl http://localhost:30000/v1/images/generations \
    -d '{
      "prompt": "A curious raccoon in a forest",
      "model": "Wan-AI/Wan2.1-T2V-1.3B-Diffusers",
      "rollout": true,
      "rollout_sde_type": "sde",
      "rollout_noise_level": 0.7
    }'

  # CPS Rollout
  curl http://localhost:30000/v1/images/generations \
    -d '{
      "prompt": "A curious raccoon in a forest",
      "model": "Wan-AI/Wan2.1-T2V-1.3B-Diffusers",
      "rollout": true,
      "rollout_sde_type": "cps",
      "rollout_noise_level": 0.7
    }'
```

Rockdu (approved):

yhyang201:
@mickqian Nvidia CI passed and PR is approved, ready for merge

python/sglang/multimodal_gen/runtime/pipelines/patches/flow_matching_with_logprob.py
```diff
@@ -0,0 +1,105 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Flow-matching rollout step utilities for log-prob computation."""
```
zhaochenyang20: If we adapt from other open-source diffusion workflows, we shall add acknowledgment here.

python/sglang/multimodal_gen/runtime/pipelines_core/stages/denoising.py
```diff
@@ -1085,13 +1107,25 @@ def forward(
...
+                            )
+                            trajectory_log_probs.append(step_log_prob)
+                        else:
+                            latents = self.scheduler.step(
+                                model_output=noise_pred,
+                                timestep=t_device,
+                                sample=latents,
+                                **extra_step_kwargs,
+                                return_dict=False,
+                            )[0]
```
zhaochenyang20: It's a little bit unclear to me that `sde_step_with_logprob` vs `self.scheduler.step` in the input parameter. The most strange thing is that `sde_step_with_logprob` takes `self.scheduler` as a parameter while `self.scheduler.step` is an object method of scheduler. Could we share the same design pattern for parameters like:

1. change `sde_step_with_logprob` to `self.scheduler.sde_step_with_logprob`
2. Or, only have one entrypoint `self.scheduler.step`, but pass in `step_index=rollout_step_indices[i], generator=batch.generator, sde_type=rollout_sde_type, noise_level=rollout_noise_level,` as `kwargs`?

In deed I am not so sure about the process of SDE and CPS. Shall ask for help on design from BBuf, mick and Yuhao.

python/sglang/multimodal_gen/runtime/pipelines_core/stages/denoising.py
```diff
@@ -1085,13 +1107,25 @@ def forward(
...
                         # Compute the previous noisy sample
-                        latents = self.scheduler.step(
-                            model_output=noise_pred,
-                            timestep=t_device,
-                            sample=latents,
-                            **extra_step_kwargs,
-                            return_dict=False,
-                        )[0]
+                        if rollout_enabled:
+                            latents, step_log_prob = sde_step_with_logprob(
```
alphabetc1: This rollout path bypasses `self.scheduler.step(...)` and directly computes the next sample from `sigmas`.

That seems not equivalent for multi-step schedulers like `FlowUniPCMultistepScheduler`, because their `step()` also updates internal state such as `last_sample`, `model_outputs`, `timestep_list`, and `lower_order_nums`.

Is this by design?

python/sglang/multimodal_gen/runtime/pipelines_core/stages/denoising.py
alphabetc1: It looks like SP isn’t supported. Could we add a TODO or FIXME?

python/sglang/multimodal_gen/configs/sample/sampling_params.py
```diff
@@ -159,6 +159,9 @@ class SamplingParams:
     # Misc
     save_output: bool = True
     return_frames: bool = False
+    rollout: bool = False
+    rollout_sde_type: str = "sde"
```
alphabetc1, line 163: can we validate `rollout_sde_type` in `_validate` so the request fails early

hnyls2002:
Hi @MikukuOvO, the flow-matching rollout log_prob support (incl. SDE/CPS, noise level, and per-step indices) has since landed on main under multimodal_gen with an expanded set of SamplingParams fields and unit tests. The implementation diverged from the patch-module approach here, so I think this is now superseded. Closing; please reopen if a piece is still missing.

PR #18815, opened 2026-02-13
Add timeout abort kits for normal / eagle.

hnyls2002:
The supplementary tests following #18760


hnyls2002:
`test_abort.py` and `test_eagle_infer_b.py` passed.

PR #18843, opened 2026-02-14
[Feature] Add SLRU eviction policy & fix RadixCache hit_count bug

liubiyongge:
## Motivation

In RAG (Retrieval-Augmented Generation) and batch processing scenarios, the system often faces **"scan" workloads**: large volumes of one-time requests (e.g., retrieving many documents). Under the default `LRUStrategy`, these one-time requests flush valuable hot data (e.g., shared System Prompts, multi-turn dialog history) out of the KV cache. This leads to frequent re-computation (Cache Miss) and high tail latency.

While implementing **SLRU (Segmented LRU)** to solve this, I discovered a critical bug in `RadixCache`: the `node.hit_count` was **never incremented** during `match_prefix` or `insert` operations. This rendered any frequency-based policies (like the existing LFU) ineffective.

## Modifications

1.  **New Strategy (`sglang/srt/mem_cache/eviction_policy.py`):**
    -   Implemented `SLRUStrategy` (Segmented LRU).
    -   **Probationary Segment:** Items with `hit_count < k` (default 2). Evicted first.
    -   **Protected Segment:** Items with `hit_count >= k`. Protected from scan noise.

2.  **Critical Bug Fix (`sglang/srt/managers/router/radix_cache.py`):**
    -   Added logic to increment `node.hit_count` in `match_prefix` (on cache hit) and `insert` (on new node creation). This fixes the underlying mechanism for LFU/SLRU.

3.  **Bug Fix (`sglang/srt/server_args.py`):**
    -   Fixed a typo: `add_radix_eviction_policy_choices` -> `radix_eviction_policy`. This fixes an `AttributeError` when accessing `ServerArgs.radix_eviction_policy`.
    -   Registered `'slru'` as a valid choice for the CLI argument `--radix-eviction-policy`.

## Accuracy Tests

This PR affects the **cache eviction policy** and **memory management** logic. It does not modify the model architecture, kernel implementation, or arithmetic precision. Therefore, the model output accuracy remains unaffected.

## Benchmarking and Profiling

I performed a "Scan Attack" benchmark to verify **Scan Resistance**.
- **Hardware:** NVIDIA A10 (24GB)
- **Constraint:** `mem_fraction_static=0.78` (Simulating memory pressure with ~11k token capacity).
- **Workload:**
    - **Hot Data:** 1 VIP prompt (~1800 tokens), accessed 10 times (Warmup).
    - **Scan Noise:** 2000 unique requests (~800 tokens each) sent sequentially to flush the cache.

**Results:**

| Metric | LRU (Baseline) | SLRU (Ours) | Improvement |
| :--- | :--- | :--- | :--- |
| **Warmup Latency** | 72.18 ms | 71.84 ms | ~0% |
| **After Scan Latency** | **549.75 ms** (Evicted) | **73.06 ms** (Survived) | **7.6x Faster** |
| **Cache Status** | Miss (Recompute) | **Hit (Protected)** | ✅ Success |

The results demonstrate that `SLRU` effectively isolates hot data from scan traffic, maintaining consistent low latency (TTFT).

## Checklist

## Review Process


python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -655,6 +659,7 @@ def _match_prefix_helper(self, node: TreeNode, key: RadixKey):
         while len(key) > 0 and child_key in node.children.keys():
             child = node.children[child_key]
             child.last_access_time = access_time
+            child.hit_count += 1
```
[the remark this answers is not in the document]
liubiyongge: Image

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -298,6 +299,9 @@ def __init__(self, params: CacheInitParams):
             self.eviction_strategy: EvictionStrategy = FILOStrategy()
         elif self.eviction_policy == "priority":
             self.eviction_strategy: EvictionStrategy = PriorityStrategy()
+        elif self.eviction_policy == "slru":
+            self.eviction_strategy: EvictionStrategy = SLRUStrategy()
```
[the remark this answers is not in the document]
liubiyongge: Done. I have updated the CLI help message in server_args.py and the ValueError message in radix_cache.py. Both now explicitly list 'slru' as a supported option

python/sglang/srt/mem_cache/radix_cache.py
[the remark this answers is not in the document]
liubiyongge: Fixed. I have updated radix_cache.py to align with the C++ implementation:

Inheritance: _split_node now correctly inherits hit_count from the child node.

Traversal: _insert_helper now increments hit_count for all existing nodes along the path.

Initialization: New nodes created during insertion are initialized with hit_count = 1.

hzh0425 (approved):
Looks good

Could you add a ci accuracy test ?

python/sglang/srt/mem_cache/radix_cache.py
ispobock: For partial match (`prefix_len < len(child.key)`), only the new node should increase hit_count. Another node does not need. Could you check this?

liubiyongge: Thanks for catching this edge case! You are absolutely right. 

I have updated the logic in `_match_prefix_helper`. Now, `child.hit_count += 1` is strictly executed only when there is a full match (`prefix_len == len(child.key)`). 

For partial matches, we no longer incorrectly inflate the hit count of the existing node. This ensures that when a node splits, the unhit suffix node won't be unfairly protected by the SLRU policy.

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -709,6 +716,7 @@ def _insert_helper(self, node: TreeNode, key: RadixKey, value, priority: int = 0
         while len(key) > 0 and child_key in node.children.keys():
             node = node.children[child_key]
             node.last_access_time = access_time
+            node.hit_count += 1
```
hzh0425: hi  @liubiyongge , Should we also consider partial match cases here? Only increment the hit count for the new node.

python/sglang/srt/mem_cache/radix_cache.py
```diff
@@ -658,10 +662,12 @@ def _match_prefix_helper(self, node: TreeNode, key: RadixKey):
             prefix_len = self.key_match_fn(child.key, key)
             if prefix_len < len(child.key):
                 new_node = self._split_node(child.key, child, prefix_len)
+                new_node.hit_count += 1
```
hzh0425: in the cache_finished_req/unfinised_req,  the scheduler already calls match_prefix before insertion, which increments hit_count for the same nodes in _match_prefix_helper and _insert_helper. This results in 2 increments per request on matched nodes

 **How hiradix_cache.py handles this:**
hiradix_cache.py picks only one increment path — the insert path — and keeps _match_prefix_helper free of any hit_count side effects

Fix solution: 

- [ ] Remove hit_count increments from _match_prefix_helper, keep only the one in _insert_helper. （ref hiradix_cache.py）
- [ ] maybe skip the hit count update for chunked requests


cc @liubiyongge  @ispobock

hzh0425 (changes requested):

python/sglang/srt/mem_cache/radix_cache.py
liubiyongge: Fixed. Added logic to increment node.hit_count for exact matches and new_node.hit_count for partial matches.

python/sglang/srt/mem_cache/radix_cache.py
liubiyongge: Fixed.

ispobock (approved):

hzh0425 (approved):

PR #18879, opened 2026-02-16
Change dump output format to dict with value and metadata

fzyzcjy:


fzyzcjy:
@gemini-code-assist review

PR #18883, opened 2026-02-16
[Diffusion] Fix LoRA weight snapshot aliasing in unmerge

ChangyiYang:
## Summary
Fix LoRA unmerge state corruption in diffusion runtime by taking an immutable snapshot of base weights.

## Problem
In `BaseLayerWithLoRA`, the backup weight (`cpu_weight`) was initialized with:
```python
base_layer.weight.to("cpu")
```
This may share storage with the original parameter when the weight is already on CPU.  
As a result, LoRA merge operations could mutate the backup tensor itself, so later `unmerge_lora_weights()` no longer restored true base weights.

## Root Cause
`cpu_weight` was not guaranteed to be an independent copy.  
`unmerge` relied on a backup that could already be polluted by previous merge cycles.

## Fix
Update backup initialization to always create an immutable snapshot:
```python
self.cpu_weight = base_layer.weight.detach().to("cpu").clone()
```
File:
- `python/sglang/multimodal_gen/runtime/layers/lora/linear.py`

## User-Visible Impact
- `set_lora(default)` after merge/unmerge cycles now restores expected generation quality.
- Prevents progressive model state drift across LoRA API operations (`merge_lora_weights`, `unmerge_lora_weights`, `set_lora`).

## Validation
- Reproduced previously failing behavior in diffusion LoRA workflow.
- After fix, output quality remains stable across LoRA state transitions in:
  - `test_diffusion_perf[zimage_image_t2i_multi_lora]` flow (perf, unmerge/merge/set_lora, switch, multi-lora steps).


python/sglang/multimodal_gen/runtime/layers/lora/linear.py
```diff
@@ -48,7 +48,11 @@ def __init__(
         self.base_layer: nn.Module = base_layer
 
         self.merged: bool = False
-        self.cpu_weight = base_layer.weight.to("cpu")
+        # Keep an immutable snapshot of the original base weight.
```
mickqian: could you simplify the comment a bit?

mickqian (approved):

python/sglang/multimodal_gen/runtime/layers/lora/linear.py
ChangyiYang: Done!

PR #18887, opened 2026-02-16
Fix CI: add flashinfer --download-cubin to install dependencies

alisonshao:
## Summary
- Add `python3 -m flashinfer --download-cubin` step to `scripts/ci/cuda/ci_install_dependency.sh` after flashinfer-jit-cache installation
- Fixes `TypeError: 'NoneType' object is not callable` for `trtllm_fp4_block_scale_moe` on B200 CI runner
- Mirrors the existing pattern in the Dockerfile (line 220)

failure example: https://github.com/sgl-project/sglang/actions/runs/22018874131/job/63624809610?pr=14105#logs

## Root Cause
The B200 (sm_100) CI runner fails on `test_deepseek_v3_fp4_mtp_small` because `flashinfer.fused_moe.trtllm_fp4_block_scale_moe` cannot be imported. The import is silently caught and set to `None`, which later causes a `TypeError` at runtime.

The Dockerfile already runs `python3 -m flashinfer --download-cubin` to download precompiled CUDA binaries for the current GPU architecture. The CI install script was missing this step, so the cubins needed for `flashinfer.fused_moe` were not available on B200/sm_100.

## Test plan
- [ ] Verify `stage-b-test-4-gpu-b200` CI passes, specifically `test_deepseek_v3_fp4_mtp_small`

### DIFF
--- a/scripts/ci/cuda/ci_download_flashinfer_cubin.sh
+++ b/scripts/ci/cuda/ci_download_flashinfer_cubin.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+# Download flashinfer cubins if the local set is incomplete.
+#
+# The flashinfer-cubin pip package may not include cubins for newer architectures
+# (e.g. sm_100, sm_120) due to PyPI size limits. This script checks the local
+# cubin status against the flashinfer artifact repository and downloads any
+# missing files.
+set -euxo pipefail
+
+CUBIN_STATUS=$(FLASHINFER_LOGGING_LEVEL=warning python3 -c "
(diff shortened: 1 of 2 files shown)


mmangkad (changes requested):
I think this is redundant and likely not the fix. `python[dev]` already installs `flashinfer-cubin`, which already contains precompiled cubins and is FlashInfer’s preferred cubin path, so unconditional `flashinfer download-cubin` just adds extra download time. I’ve also seen recent runs of this step pass including one of my PRs, so this failure seems PR-related.

Issue #18872, opened 2026-02-15
Document & auto-enable FP8 block-wise CUTLASS GEMM for SM120


b8zhong:
It seems a better SM120 compatible implementation of FP8 blockwise GEMM is implemented here, compared to Triton:

https://github.com/sgl-project/sglang/pull/9969

Help document it here. We should also enable it by default on SM120 (through auto-detection), with `--fp8-gemm-backend`.

https://docs.sglang.io/advanced_features/server_arguments.html#kernel-backends-attention-sampling-grammar-gemm

SwekeR-463:
is this up for grabs? would love to start my contris towards sgl.

Arush04:
Hi @b8zhong is this open? would like to contribute

PR #9969, opened 2025-09-03
CUTLASS fp8 blockwise gemm support of sm120

jianyingzhu:
## Motivation

This pull request adds support for the SM120 architecture in the Cutlass FP8 blockwise GEMM kernel, enabling improved performance on the latest hardware.

## Modifications

Introduced an SM120-specific branch in `fp8_blockwise_gemm_kernel.cu` to leverage new features and optimizations of Cutlass.

## Accuracy Tests

image

## Benchmarking and Profiling

Benchmarking on the Qwen3-32B model shows a 55% throughput increase compared to the Triton FP8 blockwise GEMM kernel, from 1585 tok/s to 2473 tok/s.

MLP up of cutlass kernel of: 30ms
image
MLP up of Triton kernel: 60ms
Pasted Graphic 4

## Checklist

- [x] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [x] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.ai/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.ai/developer_guide/contribution_guide.html#benchmark-the-speed).

PR #18901, opened 2026-02-16
Document & auto-enable FP8 block-wise CUTLASS GEMM for SM120

AlanPonnachan:
## Motivation

#18872 . 
This PR follows up on the integration of the SM120-compatible FP8 block-wise GEMM kernel (introduced in #9969). currently, users on SM120 (Blackwell) hardware may default to suboptimal backends (like Triton) because the server arguments do not automatically select the new kernel.

This PR ensures that `sglang` automatically detects SM120 hardware and defaults the `--fp8-gemm-backend` to `cutlass`, ensuring users get the best performance out of the box. It also documents this support in the server arguments reference.

## Modifications

- **`python/sglang/srt/server_args.py`**:
  - Added `_handle_fp8_gemm_backend` method in `ServerArgs`.
  - Logic: If `fp8_gemm_runner_backend` is `auto` and `is_sm120_supported()` is True, set backend to `cutlass`.
  - Updated the CLI help string for `--fp8-gemm-backend` to explicit mention SM120 support in the `cutlass` option.

- **`docs/advanced_features/server_arguments.md`**:
  - Updated the **Kernel Backends** table to reflect that `cutlass` is the optimal choice for Hopper, Blackwell, and SM120 GPUs.

- **`test/registered/core/test_server_args.py`**:
  - Added a new test class `TestServerArgsFP8` to verify:
    - Auto-detection sets backend to `cutlass` on SM120.
    - Backend remains `auto` on non-SM120 hardware.
    - Manual overrides (e.g., explicitly setting `triton`) are respected.

## Checklist


python/sglang/srt/server_args.py
```diff
@@ -2085,6 +2088,12 @@ def _handle_context_parallelism(self):
                     self.ep_size * self.moe_dp_size == self.tp_size
                 ), "ep_size * moe_dp_size must be equal to tp_size"
 
+    def _handle_fp8_gemm_backend(self):
+        if self.fp8_gemm_runner_backend == "auto":
+            if is_sm120_supported():
+                self.fp8_gemm_runner_backend = "cutlass"
+                logger.info("Using cutlass FP8 GEMM backend for SM120.")
+
     def _handle_data_parallelism(self):
```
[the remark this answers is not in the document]
AlanPonnachan: We are intentionally limiting this change to SM120. For SM90 (Hopper) and SM100 (Blackwell), the auto setting is designed to prefer deep_gemm if it is installed (as noted in the CLI help text)?

 Forcing cutlass here would override that preference and prevent users from utilizing DeepGEMM by default. We want to preserve the existing optimization logic for those architectures.

test/registered/core/test_server_args.py
```diff
@@ -266,5 +266,38 @@ def test_init_new_with_malformed_ipv6_address_wrong_separator(
...
+    def test_fp8_gemm_backend_manual_override(self, mock_is_sm120):
+
+        mock_is_sm120.return_value = True
+
+        args = ServerArgs(model_path="dummy", fp8_gemm_runner_backend="triton")
+        args._handle_fp8_gemm_backend()
+
+        self.assertEqual(args.fp8_gemm_runner_backend, "triton")
+
+
```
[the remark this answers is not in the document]
AlanPonnachan: The existing test_fp8_gemm_backend_auto_detection_non_sm120 case covers the scenario where the hardware is not SM120 (implicitly covering SM90/100), ensuring no regression in their default behavior.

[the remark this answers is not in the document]
b8zhong: @AlanPonnachan Delete this test

python/sglang/srt/server_args.py
```diff
@@ -2085,6 +2088,12 @@ def _handle_context_parallelism(self):
                     self.ep_size * self.moe_dp_size == self.tp_size
                 ), "ep_size * moe_dp_size must be equal to tp_size"
 
+    def _handle_fp8_gemm_backend(self):
```
b8zhong: Please delete this function. There should be somewhere else to handle this.

python/sglang/srt/server_args.py
```diff
@@ -749,6 +749,9 @@ def __post_init__(self):
         # Handle context parallelism.
         self._handle_context_parallelism()
 
+        # Handle FP8 GEMM backend
+        self._handle_fp8_gemm_backend()
```
b8zhong: Same thing below

b8zhong (changes requested):

test/registered/core/test_fp8_utils.py
b8zhong, line 1: I mean, you could delete this test alltogether, we don't need to test for setting default.

python/sglang/srt/layers/quantization/fp8_utils.py
```diff
@@ -254,13 +255,16 @@ def _dispatch_explicit_backend(backend: Fp8GemmRunnerBackend) -> Callable:
...
     """Auto-select the best backend based on hardware capabilities."""
     # Priority order for auto selection:
-    # 1. DeepGEMM (if enabled and available)
-    # 2. FlashInfer TRTLLM (if Blackwell GPU and FlashInfer available)
-    # 3. CUTLASS (if Hopper+ GPU and CUDA 12.0+)
-    # 4. AITER (if AMD GPU with AITER enabled)
-    # 5. Triton (fallback)
-
-    if deep_gemm_wrapper.ENABLE_JIT_DEEPGEMM:
+    # 1. CUTLASS (if SM120) - Prioritized for SM120
```
b8zhong, line 258: Nit: could you please put it on line 4 instead? It doesn't need to be at the very top.

python/sglang/srt/layers/quantization/fp8_utils.py
```diff
@@ -168,7 +169,7 @@ def is_aiter(self) -> bool:
 
 def _check_cutlass_block_fp8_hardware_support() -> bool:
     """Return True if CUTLASS block FP8 is supported (Hopper or newer with CUDA 12.0+)."""
-    return is_sm90_supported() or is_blackwell_supported()
+    return is_sm90_supported() or is_blackwell_supported() or is_sm120_supported()
```
b8zhong, line 172: Blackwell supported covers SM120? I don't think it's needed to add this line

b8zhong (changes requested):

AlanPonnachan:
Got it, thanks for the feedback.
I’ll make future PR descriptions more explicit. The final implementation and decisions were done by me.


please let me know if you’d prefer a new PR or a reopen.

PR #18908, opened 2026-02-16
Fix symm-mem deadlock during autotune and CUDA graph capture

alisonshao:
## Summary
- Fix `TestDeepseekV3FP4SymmetricMemory` timeout in `stage-c-test-4-gpu-b200` caused by NCCL deadlocks when `--enable-symm-mem` is enabled
- Three-part fix for sequential NCCL collective deadlock points during server init:
  1. **Disable symm_mem during FlashInfer autotune** — different ranks may profile different kernel configs, causing NCCL collective divergence
  2. **Disable symm_mem during CUDA graph capture** (both regular and piecewise) — warmup forward passes may trigger NCCL collectives that deadlock across ranks
  3. **Synchronize NCCL allocator JIT compilation before prealloc** — the NCCL allocator C extension takes ~90s to JIT compile and different ranks finish at different times. Without a barrier, the first rank to finish calls `ncclMemAlloc`/`ncclCommWindowRegister` (collective ops) while other ranks are still compiling. Also keep the pre-allocated tensor alive to prevent `ncclMemFree` (another collective) from being called.
- Same disable pattern as the DeepGEMM compilation deadlock fix (commit 3a64844a1)

MoE forward passes wrap operations in `use_symmetric_memory()` which triggers NCCL collective operations (`ncclMemAlloc`/`ncclCommWindowRegister`). During warmup phases (autotune profiling, CUDA graph capture), different ranks may execute different kernel configurations or batch sizes, causing these collectives to deadlock. The `prealloc_symmetric_memory_pool` additionally requires synchronization because the NCCL allocator's JIT compilation has variable latency across ranks.

Example failures:
- Autotune deadlock: https://github.com/sgl-project/sglang/actions/runs/22079202405/attempts/2
- CUDA graph capture deadlock: https://github.com/sgl-project/sglang/actions/runs/22121740777/job/63943145435
- Prealloc hang (JIT compile desync): https://github.com/sgl-project/sglang/actions/runs/22154429047

## Test plan
- [ ] Verify `stage-c-test-4-gpu-b200` CI passes with `TestDeepseekV3FP4SymmetricMemory` completing without timeout

### DIFF
--- a/python/sglang/srt/model_executor/model_runner.py
+++ b/python/sglang/srt/model_executor/model_runner.py
@@ -2660,13 +2660,28 @@ class ModelRunner(ModelRunnerKVCacheMixin):
         ):
             return
 
+        # JIT-compile the NCCL allocator before any collective memory
+        # operations. The compilation takes ~90s and different ranks may
+        # finish at different times. Without this barrier, the first rank
+        # to finish would call ncclMemAlloc/ncclCommWindowRegister
+        # (collective ops) while other ranks are still compiling.
+        from sglang.srt.distributed.device_communicators.pynccl_allocator import (
+            get_nccl_mem_pool,
+        )
+
+        get_nccl_mem_pool()
+        torch.distributed.barrier()
+
         # Memory allocation is tied to a cuda stream, use the forward stream
         with torch.get_device_module(self.device).stream(self.forward_stream):
             logger.info(
                 f"Pre-allocating symmetric memory pool with {envs.SGLANG_SYMM_MEM_PREALLOC_GB_SIZE.get()} GiB"
             )
             with use_symmetric_memory(get_tp_group()):
-                torch.empty(
+                # Keep the tensor alive to prevent ncclMemFree from being
+                # called (ncclMemFree is a collective that could deadlock
+                # if ranks free at different times).
+                self._symm_mem_prealloc = torch.empty(
                     (envs.SGLANG_SYMM_MEM_PREALLOC_GB_SIZE.get() * 1024 * 1024 * 1024,),
                     dtype=torch.uint8,
                     device=self.device,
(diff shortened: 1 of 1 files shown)


alisonshao:
only disabling symm_mem during the autotune warmup phase; the actual serving still uses it:                       
server startup:                                                                                                                                                              
    1. kernel_warmup() → _flashinfer_autotune()  ← symm_mem disabled         
    2. init_device_graphs()                                                                                                                                                    
    3. prealloc_symmetric_memory_pool()           ← symm_mem enabled                                                                                                           
    4. Server ready, serving requests             ← symm_mem enabled

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -2641,13 +2660,28 @@ def prealloc_symmetric_memory_pool(self):
...
+        torch.distributed.barrier()
+
         # Memory allocation is tied to a cuda stream, use the forward stream
         with torch.get_device_module(self.device).stream(self.forward_stream):
             logger.info(
                 f"Pre-allocating symmetric memory pool with {envs.SGLANG_SYMM_MEM_PREALLOC_GB_SIZE.get()} GiB"
             )
             with use_symmetric_memory(get_tp_group()):
-                torch.empty(
+                # Keep the tensor alive to prevent ncclMemFree from being
```
nvcastet, line 2681: ncclMemFree should never be called as long as the symmetric memory pool object does not go out of scope which cannot happen since it is a global variable.
pytorch mempools never release memory back to allocator while pool is alive.

nvcastet:
@Fridge003 and @alisonshao https://github.com/sgl-project/sglang/pull/18987 should fix the hang by cleaning up leftover lock files.

PR #18434, opened 2026-02-08
[Fix] Kimi K2.5 support pp

lw9527:
## Motivation
fix [18433](https://github.com/sgl-project/sglang/issues/18433)

## Checklist

- [ ] Format your code according to the [Format code with pre-commit](https://docs.sglang.io/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.io/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.io/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.io/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.io/developer_guide/contribution_guide.html#benchmark-the-speed).
- [ ] Follow the SGLang code style [guidance](https://docs.sglang.io/developer_guide/contribution_guide.html#code-style-guidance).

## Review Process

1. Ping Merge Oncalls to start the PR flow. See the [PR Merge Process](https://github.com/sgl-project/sglang/blob/main/.github/MAINTAINER.md#pull-request-merge-process).
2. Get approvals from [CODEOWNERS](https://github.com/sgl-project/sglang/blob/main/.github/CODEOWNERS) and other reviewers.
3. Trigger CI tests with [comments](https://docs.sglang.io/developer_guide/contribution_guide.html#how-to-trigger-ci-tests) or contact authorized users to do so.
   - `/tag-run-ci-label`, `/rerun-failed-ci`, `/tag-and-rerun-ci`
4. After green CI and required approvals, ask Merge Oncalls to merge.

PR #18925, opened 2026-02-17
[Fix] Enable Pipeline Parallelism support for Kimi K2.5

ieBoytsov:
## Motivation

The PR adds a support to use pipeline parallelism with Kimi k2.5 model. This is helpful for those who use the model for multi-node deployment on H100 GPUs. One common setup is 8TP + 2PP. Such a set up works in vllm but I discovered it didn't work in sglang.

Below is the command I used to start sglang on 2 nodes.
```
  --model-path $MODEL_PATH \
  --trust-remote-code \
  --dist-init-addr $MASTER_IP:$MASTER_PORT \
  --nnodes 2 \
  --node-rank $RANK \
  --tp-size 8 \
  --pp-size 2 \
  --tool-call-parser kimi_k2 \
  --reasoning-parser kimi_k2
```

Before my fix I caught `AssertionError: Pipeline Parallel is not compatible with this model.` with this command. Moreover, this assertion appeared not in the beginning but after about one hour of trying to set up the server (loading weigths, etc). 

## Modifications

1. **python/sglang/srt/models/kimi_k25.py**
   - Added `PPProxyTensors` import from `forward_batch_info`
   - Added `pp_proxy_tensors` parameter to the `forward()` method signature
   - Passed `pp_proxy_tensors` through to `general_mm_embed_routine()`

2. **python/sglang/srt/managers/mm_utils.py**
   - Fixed `general_mm_embed_routine()` to pass `input_ids` parameter to the language model instead of `None`
   - This ensures non-first PP ranks have access to `input_ids` for device inference

After the fix the server successfully starts and I tested it with a few requests. Let me know if you need more details.

## Checklist

## Review Process

### DIFF
--- a/python/sglang/srt/managers/mm_utils.py
+++ b/python/sglang/srt/managers/mm_utils.py
@@ -1132,7 +1132,7 @@ def general_mm_embed_routine(
         input_embeds = None
 
     hidden_states = language_model(
-        input_ids=None,
+        input_ids=input_ids,
         forward_batch=forward_batch,
         input_embeds=input_embeds,
         **kwargs,
--- a/python/sglang/srt/models/kimi_k25.py
+++ b/python/sglang/srt/models/kimi_k25.py
@@ -30,7 +30,7 @@ from sglang.srt.managers.schedule_batch import (
     MultimodalDataItem,
     MultimodalInputs,
 )
-from sglang.srt.model_executor.forward_batch_info import ForwardBatch
+from sglang.srt.model_executor.forward_batch_info import ForwardBatch, PPProxyTensors
 from sglang.srt.model_loader.weight_utils import default_weight_loader
 from sglang.srt.models.deepseek_v2 import DeepseekV3ForCausalLM
 from sglang.srt.models.kimi_vl_moonvit import MLP2
@@ -705,9 +705,11 @@ class KimiK25ForConditionalGeneration(nn.Module):
         input_ids: torch.Tensor,
         positions: torch.Tensor,
         forward_batch: ForwardBatch,
+        input_embeds: torch.Tensor = None,
         get_embedding: bool = False,
+        pp_proxy_tensors: Optional[PPProxyTensors] = None,
     ):
-        hidden_states = general_mm_embed_routine(
+        return general_mm_embed_routine(
             input_ids=input_ids,
             forward_batch=forward_batch,
             language_model=self.language_model,
@@ -715,9 +717,8 @@ class KimiK25ForConditionalGeneration(nn.Module):
                 Modality.IMAGE: self.get_image_feature,
             },
             positions=positions,
+            pp_proxy_tensors=pp_proxy_tensors,
         )
-
-        return hidden_states
 
     def load_weights(self, weights: Iterable[Tuple[str, torch.Tensor]]):
         """Load weights for the model, separating vision and language weights"""
--- a/test/registered/8-gpu-models/test_kimi_k25_pp.py
+++ b/test/registered/8-gpu-models/test_kimi_k25_pp.py
@@ -1,0 +1,42 @@
+import unittest
+
+from sglang.test.accuracy_test_runner import AccuracyTestParams
+from sglang.test.ci.ci_register import register_cuda_ci
+from sglang.test.run_combined_tests import run_combined_tests
+from sglang.test.test_utils import ModelLaunchSettings
+
+register_cuda_ci(est_time=1800, suite="nightly-8-gpu-common", nightly=True)
+
+KIMI_K25_MODEL_PATH = "moonshotai/Kimi-K2.5"
+
+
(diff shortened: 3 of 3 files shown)


python/sglang/srt/models/kimi_k25.py
```diff
@@ -705,20 +705,21 @@ def forward(
...
+        return general_mm_embed_routine(
             input_ids=input_ids,
             forward_batch=forward_batch,
             language_model=self.language_model,
             data_embedding_funcs={
                 Modality.IMAGE: self.get_image_feature,
             },
             positions=positions,
+            pp_proxy_tensors=pp_proxy_tensors,
         )
```
ShangmingCai, line 721: This change doesn't seem correct to me. pp logic are few lines below, can you check those lines?

ieBoytsov: Hi @ShangmingCai I don't see below the logic about pp (in the kimi_k25.py). I thought maybe you are referring to the pattern like in [qwen2_5_vl.py](https://github.com/ieBoytsov/sglang/blob/86fcb88d54dab35a4b2fefea278f9401c596d1f7/python/sglang/srt/models/qwen2_5_vl.py#L777) ?. In that model indeed after `general_mm_embed_routine` there is a check 
```
if self.pp_group.is_last_rank:
    ...
```
As per my understanding there is no such code for kimi_k25.py because this logic is already handled inside DeepseekV3ForCausalLM.forward() [here](https://github.com/ieBoytsov/sglang/blob/86fcb88d54dab35a4b2fefea278f9401c596d1f7/python/sglang/srt/models/deepseek_v2.py#L2926).

So, my understanding is that given `DeepseekV3ForCausalLM` already contains this logic, we simply pass pp_proxy_tensors and let DeepSeek handle it.

Let me know what you think and whether I am missing something.

ShangmingCai: Will check it when I am back from vacation.
cc: @whybeyoung I remember you verified that k2.5 can run with PP. Do you have time to check on this PR? I am currently on vacation right now, and I don't have a machine with enough GPU memory to verify temporarily.

ieBoytsov: Thank you!

yuan-luo: This explanation about deepseek llm makes sense. The 'pp_proxy_tensors' will be passed to language model along with kwargs inside `general_mm_embed_routine`.
@ieBoytsov Could you add a test for this feature enabling? Thanks.

ieBoytsov: Hi @yuan-luo thanks! Added test in a separate commit, please have a look.

ieBoytsov: Hi @ShangmingCai could you please check PP for kimi if you already back from a vacation? : ) I also added a test.

python/sglang/srt/models/kimi_k25.py
```diff
@@ -705,20 +705,21 @@ def forward(
         input_ids: torch.Tensor,
         positions: torch.Tensor,
         forward_batch: ForwardBatch,
+        input_embeds: torch.Tensor = None,
```
ShangmingCai, line 708: What is this line for?

test/registered/8-gpu-models/test_kimi_k25_pp.py
```diff
@@ -0,0 +1,42 @@
+import unittest
+
+from sglang.test.accuracy_test_runner import AccuracyTestParams
+from sglang.test.ci.ci_register import register_cuda_ci
+from sglang.test.run_combined_tests import run_combined_tests
+from sglang.test.test_utils import ModelLaunchSettings
+
+register_cuda_ci(est_time=1800, suite="nightly-8-gpu-common", nightly=True)
```
ShangmingCai, line 8: CC: @Fridge003 Is it appropriate to put k25 pp test in this CI?

Or should we not register it and move it to the manual folder?

python/sglang/srt/managers/mm_utils.py
```diff
@@ -1132,7 +1132,7 @@ def general_mm_embed_routine(
         input_embeds = None
 
     hidden_states = language_model(
-        input_ids=None,
+        input_ids=input_ids,
```
ShangmingCai, line 1135: @ieBoytsov Can you check this PR: https://github.com/sgl-project/sglang/pull/18434?
I think there are some similar efforts, and 18434 does not change this line, which also works.

ShangmingCai:
To ensure minimum impact on DeepSeek V2 models, we have decided to merge #18434. Thanks for your contribution, will add you as a co-author for that PR.

If the accuracy issue pops up, please reopen this PR, and we can reconsider other suggested changes in this PR.

PR #18389, opened 2026-02-07
Nsa trtllm mla sparse fp8 support with Deepseek v3.2 NVFP4

rainj-me:
## Motivation

#17655 

- support Deepseek v3.2 NVFP4 with trtllm mla sparse fp8 attention backend

## Modifications

- update the nsa backend to support trtllm sparse fp8 attention backend
- update the deepseek v2 to make sure the cos_sin_cache pass to trtllm kernels

### GSM8K
```bash
python3 benchmark/gsm8k/bench_sglang.py --num-shots 8 --num-questions 200 --parallel 100 --port 30000
100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 200/200 [00:07<00:00, 28.50it/s]
Accuracy: 0.960
Invalid: 0.000
Latency: 7.268 s
Output throughput: 2611.031 token/s
```

### GPQA

```bash
 python3 -m sglang.test.run_eval --port 30000 --eval-name gpqa --num-examples 198 --max-tokens 128000 --repeat 8 --top-p 0.95 --temperature 1.0 --thinking-mode deepseek-v3
ChatCompletionSampler initialized with self.system_message=None self.temperature=1.0 self.max_tokens=128000 self.reasoning_effort=None self.extra_body={'chat_template_kwargs': {'thinking': True}}
ChatCompletionSampler initialized with self.system_message=None self.temperature=1.0 self.max_tokens=128000 self.reasoning_effort=None self.extra_body={'chat_template_kwargs': {'thinking': True}}
ChatCompletionSampler initialized with self.system_message=None self.temperature=1.0 self.max_tokens=128000 self.reasoning_effort=None self.extra_body={'chat_template_kwargs': {'thinking': True}}
ChatCompletionSampler initialized with self.system_message=None self.temperature=1.0 self.max_tokens=128000 self.reasoning_effort=None self.extra_body={'chat_template_kwargs': {'thinking': True}}
ChatCompletionSampler initialized with self.system_message=None self.temperature=1.0 self.max_tokens=128000 self.reasoning_effort=None self.extra_body={'chat_template_kwargs': {'thinking': True}}
ChatCompletionSampler initialized with self.system_message=None self.temperature=1.0 self.max_tokens=128000 self.reasoning_effort=None self.extra_body={'chat_template_kwargs': {'thinking': True}}
[...]

PR #18931, opened 2026-02-17
Fix NSA FP8 KV cache path for both-trtllm MHA one-shot

mmangkad:
## Motivation

This is a follow-up to #18389 for the NSA FP8 KV cache path case when both prefill and decode backends are trtllm.

In this setup, the MHA one-shot prefix path could still enter paged dequant and hit a layout mismatch assertion (compact trtllm KV layout vs packed dequant expectation). For the both-trtllm configuration, this dequant step is unnecessary.

cc @rainj-me @Fridge003 

## Modifications

- Updated the FP8 branch condition in forward_mha.py to skip paged dequant when both `nsa_prefill_backend` and `nsa_decode_backend` are `trtllm`.
- In the both-trtllm case, reuse the existing direct KV fetch path.
- No changes to NSA routing semantics or dequant kernel implementation.

## Accuracy Tests

This is an example of a case that previously crashed with `AssertionError: dim_quant: 576 != 656`.

```shell
SGLANG_ENABLE_SPEC_V2=1 python -m sglang.launch_server \
  --model-path nvidia/DeepSeek-V3.2-NVFP4 \
  --tensor-parallel-size 4 \
  --attention-backend nsa \
  --nsa-prefill-backend trtllm \
  --nsa-decode-backend trtllm \
  --moe-runner-backend flashinfer_trtllm \
  --quantization modelopt_fp4 \
  --kv-cache-dtype fp8_e4m3 \
  --reasoning-parser deepseek-v3 \
  --speculative-num-steps 3 \
  --speculative-eagle-topk 1 \
  --model-loader-extra-config '{"enable_multithread_load": true,"num_threads": 96}'
```

```shell
python benchmark/gsm8k/bench_sglang.py --num-shots 20 --num-questions 1319 --parallel 48

100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1319/1319 [02:45<00:00,  7.96it/s]
Accuracy: 0.955
Invalid: 0.000
Latency: 165.688 s
Output throughput: 786.323 token/s
```

## Checklist

## Review Process


python/sglang/srt/layers/attention/nsa_backend.py
```diff
@@ -1435,7 +1435,10 @@ def forward_extend(
                     )
                     assert page_table_1_flattened is not None
                     kv_cache = dequantize_k_cache_paged(
```
Fridge003, line 1437: For trtllm kernel, can we just skip the cslling of `dequantize_k_cache_paged` and directly use the original fp8 kv cache instead?

mmangkad: @Fridge003 in `trtllm` mode we never enter the `flashmla_sparse` dequant branch: `forward_decode` goes straight to`_forward_trtllm`, and `forward_extend` goes to `_forward_trtllm` (MLA path) or `_forward_standard_mha` (MHA one-shot), so `dequantize_k_cache_paged` is bypassed either way

rainj-me:
> > > > @mmangkad From my understanding, the MHA_ONE_SHOT supported attn backends are "fa3", "flashinfer", "flashmla" per code https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/models/deepseek_common/attention_backend_handler.py#L64-L69 . Since the trtllmla attn backend is not in the list, if we find there is MHA_ONE_SHOT seq in the forward_batch, there probably some other bugs involved.
> > > 
> > > 
> > > @rainj-me my read is that this goes through the NSA-specific dispatcher: [`handle_attention_nsa`](https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/models/deepseek_common/attention_backend_handler.py#L144-L155) checks `backend.use_mha` and returns `MHA_ONE_SHOT` when true, and `backend.use_mha` is decided in [`set_nsa_prefill_impl`](https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/layers/attention/nsa_backend.py#L2021-L2076).
> > 
> > 
> > Then the fix should be only support MHA instead of MHA_ONE_SHOT with [`handle_attention_nsa`](https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/models/deepseek_common/attention_backend_handler.py#L144-L155) . The reason to avoid changing the dequant logic is because we should keep it only for triton and flashmla util trtllm-mla support the sparse fp8 kv cache.
> 
> Thanks - I thought NSA and `trtllm_mla` are different dispatch paths, and NSA can intentionally go through `MHA_ONE_SHOT` via `use_mha`, so I kept this PR scoped to the FP8 layout mismatch fix. As I mentioned to Fridge, `trtllm` kernel paths still bypass this dequant path anyway. Given that, what would you prefer here: keep this scoped fix, or change NSA routing semantics in this PR?

Fix the kv cache dequant to use 576 kv cache dim may lead to revert in future if trtllm support sparse kv cache. Base on this, I believe change the NSA routing should be a simple and sufficient fix. The only difference for MHA_ONE_SHOT and MHA is the forward_batch.mha_one_shot flag.

rainj-me (approved):

Fridge003 (approved):

PR #18938, opened 2026-02-17
[Sarvam] Add inference support for Sarvam MoE LLMs

rakesh-sarvam:
Adds inference support for two Sarvam MoE models:
Sarvam 30B MoE -- GQA attention with QK norm, fused QK-norm-RoPE kernel, sparse MoE with shared experts, and HF checkpoint weight remapping.
Sarvam 105B MoE -- MLA (Multi-head Latent Attention) with weight absorption (kv_b_proj split into w_kc/w_vc BMMs), FP8 BMM support, multi-backend attention dispatch (FA3, FlashMLA, CutlassMLA, etc.), and optional fp8 MHA prefill path for full-head attention during prefill.
Both models support dual-stream CUDA overlap for shared/routed experts, forward_prepare/forward_core split, chunked prefill, and TP/PP/EP/DP parallelism.


adarshxs (approved):
fixes were applied, and model is to be released soon. please add the model in the doc matrix. LGTM

python/sglang/srt/batch_overlap/two_batch_overlap.py
```diff
@@ -79,7 +79,7 @@ def compute_split_seq_index(
     extend_lens: Optional[Sequence[int]],
     token_num_per_seq: Optional[int],
 ) -> Optional[int]:
-    if forward_mode == ForwardMode.EXTEND:
+    if forward_mode == ForwardMode.EXTEND or forward_mode == ForwardMode.MIXED:
```
adarshxs: i'm guessing this is unrelated to the model support. Can we not change this? affects a lot of downstream models.

Fridge003: @rakesh-sarvam yes can we remove this

rakesh-sarvam: done, reverted the change

python/sglang/srt/models/sarvam_moe.py
```diff
@@ -0,0 +1,1195 @@
...
+_is_cublas_ge_129 = is_nvidia_cublas_version_ge_12_9()
+
+if _is_cuda:
+    try:
+        from sgl_kernel import bmm_fp8, merge_state_v2
+
+        from sglang.srt.layers.quantization.fp8_kernel import per_tensor_quant_mla_fp8
+
+        _has_fp8_support = True
+        _has_concat_mla_k = True
```
adarshxs, line 83: `_has_concat_mla_k ` what's the use?

python/sglang/srt/models/sarvam_moe.py
```diff
@@ -0,0 +1,1195 @@
...
+        merge_state_v2 = None
+        per_tensor_quant_mla_fp8 = None
+else:
+    _has_fp8_support = False
+    _has_concat_mla_k = False
+    bmm_fp8 = None
+    merge_state_v2 = None
+    per_tensor_quant_mla_fp8 = None
+
+_MHA_DEBUG_PRINTED = False
```
adarshxs: `_MHA_DEBUG_PRINTED` same

python/sglang/srt/models/sarvam_moe.py
```diff
@@ -0,0 +1,1220 @@
...
+        self._swap_sarvam_blocks(quant_config)
+
+    @staticmethod
+    def _remap_config(config: PretrainedConfig) -> None:
+        if not hasattr(config, "n_routed_experts"):
+            config.n_routed_experts = config.num_experts
+        if not hasattr(config, "n_shared_experts"):
+            config.n_shared_experts = getattr(config, "num_shared_experts", None)
+
+        config.n_shared_experts = None
```
adarshxs: what's the intent here?

rakesh-sarvam: We set n_shared_experts only for schema compatibility, then force it to None to skip DeepSeek’s shared-expert init path (it is incompatible with ModelOpt FP8 and expects weight_block_size).

python/sglang/srt/models/sarvam_moe.py
```diff
@@ -0,0 +1,1220 @@
...
+
+    def __init__(
+        self,
+        config: PretrainedConfig,
+        quant_config: Optional[QuantizationConfig] = None,
+        prefix: str = "",
+    ) -> None:
+        self._remap_config(config)
+        super().__init__(config, quant_config, prefix)
+        self._swap_sarvam_blocks(quant_config)
```
adarshxs: i'm guessing the init here will be quite memory expensive? Is that why u have the OOM try except? can we handle this better?

rakesh-sarvam: correct, this init path is memory heavy because we first instantiate deepseek blocks, then replace them with sarvam blocks. The OOM try/retry is guarding that transient allocation spike during replacement.
the clean fix is a full sarvam native model definition (no deepseek init + swap)

python/sglang/srt/models/sarvam_moe.py
adarshxs: then I think we dont need to check for that attribute since we set it to None anyways. 

Can we do a native implementation init? No deepseek init + swap? that way you won't have to deal with this schema incompatibility

python/sglang/srt/models/sarvam_moe.py
adarshxs: > the clean fix is a full sarvam native model definition (no deepseek init + swap)

I think this is the way to go.

python/sglang/srt/models/sarvam_moe.py
rakesh-sarvam: yes with the pure sarvam definition we wouldn't require these hotfixes

adarshxs: you can refer glm 4 moe lite: https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/models/glm4_moe_lite.py

i think it does something similar

Fridge003 (approved):

Issue #18201, opened 2026-02-03
[Bug] Fails to start with dual-stack because warmup treats host as URL


GuyStone:
### Checklist

- [x] I searched related issues but found no solution.
- [x] The bug persists in the latest version.
- [x] Issues without environment info and a minimal reproducible demo are hard to resolve and may receive no feedback.
- [x] If this is not a bug report but a general question, please start a discussion at https://github.com/sgl-project/sglang/discussions. Otherwise, it will be closed.
- [x] Please use English. Otherwise, it will be closed.

### Describe the bug

In environments with a dual stack configuration (such as a Kubernetes pod on a [standard dual-stack enabled cluster](https://kubernetes.io/docs/concepts/services-networking/dual-stack/), like GKE), an `--host` empty is generally expected to bind the server to all available addresses.

With SGLang, this would be declaring `--host=''` empty, however, SGLang will **fail to start unless warmup is disabled** with `--skip-server-warmup`, because the warmup logic uses the `--host` value as a URL.

Ideally we'd want to support dual-stack with the warmup before accepting requests. Why does dual stack matter? If a server is running in single-stack mode on a dual-stack node, clients may experience higher latency due to failed IPv6 connection attempt before falling back to IPv4. This is most noticeable for cross-region requests.
```
20:10:55.232459 eth0  In  IP6 <SRC_IPV6.SRCPORT> > <DST_IPV6.DSTPORT>: Flags [S], seq 2151374032, win 43400, options [mss 1400,sackOK,TS val 460105967 ecr 0,nop,wscale 9], length 0
20:10:55.232475 eth0  Out IP6 <DST_IPV6.DSTPORT > SRC_IPV6.SRCPORT>: Flags [R.], seq 0, ack 2151374033, win 0, length 0
``` 

#### Command (empty host)
```
python -m sglang.launch_server --model-path Qwen/Qwen3-0.6B --host ''
```
#### Output
```
[2026-02-03 20:24:45] INFO:     Started server process [1495710]
[2026-02-03 20:24:45] INFO:     Waiting for application startup.
[2026-02-03 20:24:45] Using default chat sampling params from model generation config: {'repetition_penalty': 1.0, 'temperature': 0.6, 'top_k': 20, 'top_p': 0.95}
[2026-02-03 20:24:45] Using default chat sampling params from model generation config: {'repetition_penalty': 1.0, 'temperature': 0.6, 'top_k': 20, 'top_p': 0.95}
[...]

PR #18973, opened 2026-02-18
Add SSL/TLS support for HTTP and gRPC servers

Kangyan-Zhou:
## Summary
- Add `--ssl-keyfile`, `--ssl-certfile`, `--ssl-ca-certs`, and `--ssl-keyfile-password` CLI flags to enable HTTPS/TLS termination for both HTTP (uvicorn) and gRPC servers
- Add `--enable-ssl-refresh` flag for automatic certificate hot-reloading when cert/key files change on disk (e.g., rotated by cert-manager, Let's Encrypt, or a sidecar)
- Internal health-check and warmup requests respect SSL configuration via a new `ssl_verify()` helper on `ServerArgs`
- `url()` now rewrites wildcard bind addresses (`0.0.0.0` → `127.0.0.1`, `::` → `::1`) so internal requests use loopback addresses compatible with certificate SANs

co authored by @GuyStone 

## Motivation
Users deploying SGLang in production environments need TLS encryption for compliance and security. This PR brings SSL support inline with vLLM's `--ssl-*` flags, making it straightforward to enable HTTPS without a reverse proxy. The auto-refresh capability avoids server restarts when certificates are rotated, which is critical for zero-downtime deployments with short-lived certificates.

### SSL/TLS support
- **`server_args.py`**: New SSL dataclass fields, early validation (`_handle_ssl_validation`) for consistent args and file existence, `ssl_verify()` helper, scheme-aware `url()` with loopback rewrite
- **`http_server.py`**: SSL params passed to `uvicorn.run()` (both single and multi-worker), warmup requests use `verify=ssl_verify`
- **`http_server_engine.py`**: Health-check and `_make_request` use `ssl_verify()` and `url()`
- **`grpc_server.py`**: TLS credentials with error handling for file reads, credential creation, and port binding; rejects `--ssl-keyfile-password` (unsupported by gRPC)

### SSL certificate auto-refresh (`--enable-ssl-refresh`)
- **`ssl_utils.py`** (new): `SSLCertRefresher` class using `watchfiles.awatch()` (Rust-based async file watcher) to monitor cert/key/CA files and reload the `SSLContext` in-place when changes are detected. Errors are logged but don't crash the server.
- **`http_server.py`**: When refresh is enabled, uses `uvicorn.Config`/`Server` API with `asyncio.run()` so the refresher runs on the same event loop as the server.
- **`grpc_server.py`**: When refresh is enabled, uses `grpc.dynamic_ssl_server_credentials` with an mtime-based fetcher callback. Mtimes are only updated after successful reload to ensure retries on failure.
- **`pyproject.toml`**: Added `watchfiles` dependency (already a transitive dep of uvicorn's reload feature)
- **`server_args.py`**: New `enable_ssl_refresh` field, `--enable-ssl-refresh` CLI flag, validation that SSL cert/key are provided when refresh is enabled

### Tests
- **`test_server_args.py`**: 21 unit tests covering SSL defaults, validation errors, file existence checks, URL scheme/loopback, `ssl_verify()` return values, CLI parsing, IPv6+SSL, and `--enable-ssl-refresh` flag parsing/validation
- **`test_ssl_cert_refresher.py`** (new): 6 unit tests covering cert/key reload on change, CA reload on change, stop cancels tasks, task count with/without CA, and error resilience

## Error handling
- Early file-existence validation prevents late failures after model loading
- gRPC file reads wrapped with actionable `ValueError` messages
- `grpc.ssl_server_credentials()` failures caught with guidance about key/cert mismatch
- `add_secure_port` return value checked for bind failures
- Orphan SSL args (`--ssl-ca-certs` or `--ssl-keyfile-password` without cert/key) rejected
- `--enable-ssl-refresh` without cert/key rejected with clear error message
- Warning logged when certificate verification is disabled (SSL without `--ssl-ca-certs`)
- Warning logged when `--enable-ssl-refresh` is used with multi-worker mode (not supported, gracefully ignored)
- SSL refresh errors (bad cert files) are logged but don't crash — server continues with previous certificates

## Test plan
- [x] Unit tests: 21 tests in `TestSSLArgs` all pass
- [x] Unit tests: 6 tests in `TestSSLCertRefresher` all pass
- [x] E2E: HTTPS server with self-signed certs on H200 GPU (Qwen2.5-0.5B-Instruct)
  - [x] Server starts and completes warmup over HTTPS
  - [x] `/health`, `/v1/models`, `/v1/chat/completions` work via `curl --cacert`
  - [x] Plain HTTP connection to HTTPS port is rejected
  - [x] HTTPS without CA cert fails verification
- [x] E2E: Plain HTTP server still works (no regression)
- [x] Validation tests: all error paths produce correct messages
- [x] Pre-commit hooks pass

### DIFF
--- a/python/sglang/srt/server_args.py
+++ b/python/sglang/srt/server_args.py
@@ -5400,7 +5400,7 @@ class ServerArgs:
         scheme = "https" if self.ssl_certfile else "http"
         # When binding to all interfaces, use loopback for internal requests.
         host = self.host
-        if host == "0.0.0.0":
+        if not host or host == "0.0.0.0":
             host = "127.0.0.1"
         elif host == "::":
             host = "::1"
--- a/python/sglang/srt/server_args.py
+++ b/python/sglang/srt/server_args.py
@@ -314,6 +314,7 @@ class ServerArgs:
     ssl_certfile: Optional[str] = None
     ssl_ca_certs: Optional[str] = None
     ssl_keyfile_password: Optional[str] = None
+    enable_ssl_refresh: bool = False
 
     # Quantization and data type
     dtype: str = "auto"
@@ -868,6 +869,11 @@ class ServerArgs:
             raise ValueError(
                 f"SSL CA certificates file not found: '{self.ssl_ca_certs}'. "
                 f"Please check the --ssl-ca-certs path."
+            )
+        if self.enable_ssl_refresh and not (self.ssl_certfile and self.ssl_keyfile):
+            raise ValueError(
+                "--enable-ssl-refresh requires --ssl-certfile and --ssl-keyfile "
+                "to be specified."
             )
 
     def _handle_deprecated_args(self):
@@ -3294,6 +3300,13 @@ class ServerArgs:
             type=str,
             default=ServerArgs.ssl_keyfile_password,
             help="The password to decrypt the SSL keyfile.",
+        )
+        parser.add_argument(
+            "--enable-ssl-refresh",
+            action="store_true",
+            default=ServerArgs.enable_ssl_refresh,
+            help="Enable automatic SSL certificate hot-reloading when cert/key "
+            "files change on disk. Requires --ssl-certfile and --ssl-keyfile.",
         )
 
         # Quantization and data type
--- a/python/sglang/srt/server_args.py
+++ b/python/sglang/srt/server_args.py
@@ -309,6 +309,12 @@ class ServerArgs:
     nccl_port: Optional[int] = None
     checkpoint_engine_wait_weights_before_ready: bool = False
 
+    # SSL/TLS
+    ssl_keyfile: Optional[str] = None
+    ssl_certfile: Optional[str] = None
+    ssl_ca_certs: Optional[str] = None
+    ssl_keyfile_password: Optional[str] = None
+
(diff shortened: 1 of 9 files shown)


python/sglang/srt/server_args.py
```diff
@@ -5165,10 +5249,43 @@ def from_cli_args(cls, args: argparse.Namespace):
...
+        host = self.host
+        if host == "0.0.0.0":
+            host = "127.0.0.1"
+        elif host == "::":
+            host = "::1"
+        if is_valid_ipv6_address(host):
+            return f"{scheme}://[{host}]:{self.port}"
         else:
-            return f"http://{self.host}:{self.port}"
+            return f"{scheme}://{host}:{self.port}"
```
GuyStone, line 5410: Might be slightly out of scope for the SSL/TLS work, but since we’re already touching/changing `url()`, could we include a fix/support for dual stack / empty host? 🙏 

- Issue: https://github.com/sgl-project/sglang/issues/18201
- Basic example fix: https://github.com/sgl-project/sglang/pull/18202

Kangyan-Zhou: Cool I'll implement this fix in the same PR

PR #18974, opened 2026-02-18
Enable to test bench_one_batch with image inputs for multimodal

kalyank007:
Enable to test bench_one_batch with image inputs for multimodal

---
### CI States

Latest PR Test (Base): :x: [Run #32824524353](https://github.com/sgl-project/sglang/actions/runs/32824524353)
Latest PR Test (Extra): :x: [Run #32824524196](https://github.com/sgl-project/sglang/actions/runs/32824524196)
Latest PR Test (AMD ROCm 7.2): :x: [Run #32824524319](https://github.com/sgl-project/sglang/actions/runs/32824524319)

### DIFF
--- a/python/sglang/benchmark/one_batch.py
+++ b/python/sglang/benchmark/one_batch.py
@@ -17,7 +17,13 @@ python -m sglang.benchmark.one_batch --model-path meta-llama/Meta-Llama-3-8B-Ins
 ## run with CUDA profiler (nsys):
 nsys profile --force-overwrite=true -o bench_one_batch python -m sglang.benchmark.one_batch --model-path meta-llama/Meta-Llama-3-8B-Instruct --batch-size 1 --input-len 256 --profile --profile-activities CUDA_PROFILER
 # Usage (correctness test):
-python -m sglang.benchmark.one_batch --model-path TinyLlama/TinyLlama-1.1B-Chat-v0.4 --correct
+python -m sglang.benchmark.one_batch --model-path TinyLlama/TinyLlama-1.1B-Chat-v0.4 --correctness-test
+
+# Usage (multimodal/vision model testing):
+## Latency test with images:
+python -m sglang.benchmark.one_batch --model-path meta-llama/Llama-4-Scout-17B-Vision-Instruct --image-test --image-resolution 1080p --image-count 2 --batch-size 1 --input-len 512 --output-len 128
+## Correctness test with images:
+python -m sglang.benchmark.one_batch --model-path meta-llama/Llama-4-Scout-17B-Vision-Instruct --correctness-test --image-test --image-urls https://raw.githubusercontent.com/sgl-project/sgl-test-files/refs/heads/main/images/man_ironing_on_back_of_suv.png
 
 ## Reference output (of the correctness test above, can be gpu dependent):
 input_ids=[[1, 450, 7483, 310, 3444, 338], [1, 450, 7483, 310, 278, 3303, 13187, 290, 338], [1, 20628, 338, 263, 6575, 1460, 2462, 322, 306, 763]]
@@ -61,8 +67,10 @@ from types import SimpleNamespace
 from typing import Optional, Tuple
 
 import numpy as np
+import requests
 import torch
 import torch.distributed as dist
+from PIL import Image
 
 from sglang.srt.configs.model_config import ModelConfig
 from sglang.srt.distributed.parallel_state import (
@@ -71,11 +79,19 @@ from sglang.srt.distributed.parallel_state import (
 )
 from sglang.srt.distributed.parallel_state_wrapper import ParallelState
 from sglang.srt.entrypoints.engine import _set_envs_and_config
+from sglang.srt.environ import envs
 from sglang.srt.layers.dp_attention import compute_dp_attention_world_info
 from sglang.srt.layers.moe import initialize_moe_config
 from sglang.srt.layers.quantization.fp4_utils import initialize_fp4_gemm_config
 from sglang.srt.layers.quantization.fp8_utils import initialize_fp8_gemm_config
-from sglang.srt.managers.schedule_batch import Req, ScheduleBatch
+from sglang.srt.managers.schedule_batch import (
+    Modality,
+    MultimodalDataItem,
+    MultimodalInputFormat,
+    MultimodalInputs,
+    Req,
+    ScheduleBatch,
+)
 from sglang.srt.managers.scheduler_components.dp_attn import prepare_mlp_sync_batch_raw
 from sglang.srt.mem_cache.base_prefix_cache import EvictParams
 from sglang.srt.model_executor.cuda_graph_config import Phase
@@ -211,6 +227,13 @@ class BenchArgs:
     profile_prefix: str = "profile"
     profile_start_step: Optional[int] = None
     profile_steps: Optional[int] = None
+    # Multimodal support
+    image_test: bool = False
+    image_urls: Tuple[str] = ()
+    image_count: int = 1
+    image_resolution: str = "1080p"
+    image_content: str = "random"
+    image_placeholder: str = "<|image|>"
 
     @staticmethod
     def add_cli_args(parser: argparse.ArgumentParser):
@@ -279,6 +302,51 @@ class BenchArgs:
             type=int,
             default=None,
             help="Number of decode steps to profile starting from profile-start-step. If not specified, profiles only one step.",
+        )
+        parser.add_argument(
+            "--image-test",
+            action="store_true",
+            help="Enable multimodal (vision) support for the benchmark.",
+        )
+        parser.add_argument(
+            "--image-urls",
+            type=str,
+            nargs="+",
+            default=[],
+            help="List of image URLs to use for multimodal testing. Can be local file paths or HTTP URLs.",
+        )
+        parser.add_argument(
+            "--image-count",
+            "--num-images",
+            dest="image_count",
+            type=int,
+            default=BenchArgs.image_count,
+            help="Number of images per request for generated synthetic image inputs.",
+        )
+        parser.add_argument(
+            "--image-resolution",
+            type=str,
+            default=BenchArgs.image_resolution,
+            help=(
+                "Resolution for generated synthetic images. "
+                "Supports presets 4k/1080p/720p/360p or custom 'heightxwidth' (e.g., 1080x1920)."
+            ),
+        )
+        parser.add_argument(
+            "--image-content",
+            type=str,
+            default=BenchArgs.image_content,
+            choices=["random", "blank"],
+            help="Content for generated synthetic images. Supports random and blank.",
+        )
+        parser.add_argument(
+            "--image-placeholder",
+            type=str,
+            default=BenchArgs.image_placeholder,
+            help=(
+                "Image placeholder token used in the prompt for multimodal models. "
+                "Defaults to '<|image|>' (Llama-4). Use '<image>' for LLaVA-style models."
+            ),
         )
 
     @classmethod
@@ -352,6 +420,16 @@ def load_model(server_args, port_args, gpu_id, tp_rank):
         model_runner.alloc_memory_pool()
         model_runner.init_attention_backends()
         model_runner.init_cuda_graphs()
+
+    # Initialize multimodal embedding cache for vision models
+    if model_config.is_multimodal:
+        from sglang.srt.managers.mm_utils import init_mm_embedding_cache
+
+        # Initialize with a reasonable cache size (100MB by default)
+        cache_size_mb = envs.SGLANG_VLM_CACHE_SIZE_MB.get()
+        init_mm_embedding_cache(max_size=cache_size_mb * 1024 * 1024)
+        rank_print(f"Initialized multimodal embedding cache with size {cache_size_mb}MB")
+
     rank_print(f"max_total_num_tokens={model_runner.max_total_num_tokens}")
     tokenizer = get_tokenizer(
         server_args.tokenizer_path,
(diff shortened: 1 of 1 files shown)


python/sglang/bench_one_batch.py
```diff
@@ -269,12 +386,240 @@ def load_model(server_args, port_args, gpu_id, tp_rank):
...
+        from transformers import AutoProcessor
+
+        # Load the processor for the model
+        processor = AutoProcessor.from_pretrained(
+            model_runner.model_config.model_path, trust_remote_code=True
+        )
+
+        # Create a dummy text with image placeholders for processor
+        # Llama-4 uses <|image|> as the placeholder token
+        image_token_str = "<|image|>"
```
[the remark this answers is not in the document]
kalyank007: Done

python/sglang/bench_one_batch.py
```diff
@@ -269,12 +386,240 @@ def load_model(server_args, port_args, gpu_id, tp_rank):
...
+        print()
+        print("For reliable multimodal benchmarking, consider using:")
+        print("1. examples/multimodal_bench_wrapper.py (uses Engine API)")
+        print("2. test/srt/test_vlm_input_format.py for working examples")
+        print()
+        import traceback
+
+        traceback.print_exc()
+        print("=" * 80)
+        return None
```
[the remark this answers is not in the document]
kalyank007: Done

python/sglang/bench_one_batch.py
```diff
@@ -281,12 +348,26 @@ def load_model(server_args, port_args, gpu_id, tp_rank):
...
+        )
+
     if server_args.tp_size > 1:
         dist.barrier()
     return model_runner, tokenizer
 
 
-def prepare_inputs_for_correctness_test(bench_args, tokenizer, custom_prompts):
+def prepare_inputs_for_correctness_test(
+    bench_args, tokenizer, custom_prompts, images=None, model_runner=None
```
polisettyvarma, line 414: new args are not used, why is it added ?

python/sglang/bench_one_batch.py
```diff
@@ -320,6 +387,18 @@ def load_model(server_args, port_args, gpu_id, tp_rank):
         tokenizer_mode=server_args.tokenizer_mode,
         trust_remote_code=server_args.trust_remote_code,
     )
+
+    # Initialize multimodal embedding cache for vision models
+    if model_config.is_multimodal:
+        from sglang.srt.managers.mm_utils import init_mm_embedding_cache
+
+        # Initialize with a reasonable cache size (100MB by default)
+        cache_size_mb = int(os.environ.get("SGLANG_VLM_CACHE_SIZE_MB", "100"))
```
polisettyvarma, line 396: from sglang.srt.environ import envs
use envs.SGLANG_VLM_CACHE_SIZE_MB.get()

python/sglang/bench_one_batch.py
```diff
@@ -351,6 +431,275 @@ def prepare_inputs_for_correctness_test(bench_args, tokenizer, custom_prompts):
...
+            return None
+
+        # Find image token positions in input_ids
+        image_token_positions = [
+            idx for idx, token_id in enumerate(input_ids) if token_id == image_token_id
+        ]
+
+        if not image_token_positions:
+            print(f"WARNING: Image token {image_token_id} not found in input_ids")
+            print(f"  Input IDs (first 20): {input_ids[:20]}")
```
polisettyvarma, line 605: why/how will this print help ?

python/sglang/bench_one_batch.py
```diff
@@ -351,6 +431,275 @@ def prepare_inputs_for_correctness_test(bench_args, tokenizer, custom_prompts):
...
+
+        local_rank_print(
+            f"✓ Created MultimodalInputs with {len(images)} image(s), offsets={offsets}, im_token_id={image_token_id}"
+        )
+
+        return mm_inputs
+
+    except Exception as e:
+        # See comment above: use local_rank_print to avoid per-rank noisy output.
+        verbose = os.environ.get("SGLANG_BENCH_MULTIMODAL_VERBOSE", "0") not in (
```
polisettyvarma, line 661: we are printing on rank0, why this new one ?
is it huge print output ?

python/sglang/bench_one_batch.py
```diff
@@ -351,6 +431,275 @@ def prepare_inputs_for_correctness_test(bench_args, tokenizer, custom_prompts):
...
+            traceback.print_exc()
+            local_rank_print("=" * 80)
+        return None
+
+    # For multimodal, use image description prompt with image tokens
+    if bench_args.image_test and images:
+        # For multimodal models, we need to include image placeholder tokens in the prompt
+        image_token_str = bench_args.image_placeholder
+        image_placeholders = " ".join([image_token_str] * len(images))
+        prompts = [f"{image_placeholders} Describe the images in detail."]
```
polisettyvarma, line 701: we need to discuss about this text

python/sglang/bench_one_batch.py
```diff
@@ -392,18 +765,107 @@ def prepare_extend_inputs_for_correctness_test(
...
+
+        # Encode the prompt to get the token IDs
+        encoded = tokenizer.encode(prompt_template)
+
+        # Find where the image token appears in the encoded sequence
+        # Encode just the image token to find its ID
+        image_token_ids = tokenizer.encode(image_token, add_special_tokens=False)
+
+        print(f"image_token_str='{image_token}', image_token_ids={image_token_ids}")
+        print(f"Full prompt: '{prompt_template}'")
```
polisettyvarma, line 794: do we need to print ?

python/sglang/bench_one_batch.py
```diff
@@ -392,18 +765,107 @@ def prepare_extend_inputs_for_correctness_test(
...
+            image_token_id = image_token_ids[0]
+            # Find all positions of the image token
+            positions = [
+                idx
+                for idx, token_id in enumerate(encoded)
+                if token_id == image_token_id
+            ]
+            print(f"Found image_token_id {image_token_id} at positions: {positions}")
+            if positions:
+                # For simplicity, use the first image token position
```
polisettyvarma, line 807: will position be same for different resolution images in the same run ?

python/sglang/bench_one_batch.py
```diff
@@ -392,18 +765,107 @@ def prepare_extend_inputs_for_correctness_test(
...
+
+        if not image_token_positions:
+            print(f"WARNING: Could not find image token in encoded sequence!")
+            print(f"  This may cause issues with multimodal processing")
+
+        # If the encoded prompt is shorter than input_len, pad with random tokens
+        if len(encoded) < input_len:
+            # Pad with random tokens (excluding special tokens)
+            padding_tokens = np.random.randint(
+                100, 10000, input_len - len(encoded), dtype=np.int32
```
polisettyvarma, line 819: how do we know which numbers to use ?

python/sglang/bench_one_batch.py
```diff
@@ -392,18 +765,107 @@ def prepare_extend_inputs_for_correctness_test(
...
+            ).tolist()
+            encoded = encoded + padding_tokens
+        elif len(encoded) > input_len:
+            # Truncate if too long
+            encoded = encoded[:input_len]
+
+        # Create the same input for all requests in the batch
```
polisettyvarma, line 826: meaning ?

python/sglang/bench_one_batch.py
```diff
@@ -392,18 +765,107 @@ def prepare_extend_inputs_for_correctness_test(
...
+            else np.random.randint(0, 10000, (batch_size, input_len), dtype=np.int32)
+        )
+
     sampling_params = SamplingParams(
         temperature=0,
         max_new_tokens=BenchArgs.output_len,
     )
 
+    # For multimodal batch processing, create multimodal inputs once and share across all requests
+    # This ensures consistent pad_value usage and avoids duplication
```
polisettyvarma, line 842: can we change variable name based on this ?

python/sglang/bench_one_batch.py
```diff
@@ -436,6 +903,13 @@ def is_tree_cache(self) -> bool:
     def evict(self, params: EvictParams):
         pass
 
+    def available_and_evictable_str(self) -> str:
+        available_size = self.token_to_kv_pool_allocator.available_size()
+        return f"Available tokens: {available_size} ({available_size=} + evictable_size=0)\n"
+
+    def pretty_print(self):
+        pass
```
polisettyvarma, line 911: without this , we observe error ?

python/sglang/bench_one_batch.py
```diff
@@ -605,10 +1079,33 @@ def correctness_test(
...
+                height,
+                rank_print,
+                image_content=bench_args.image_content,
+            )
+
     # Prepare inputs
     custom_prompts = _read_prompts_from_file(bench_args.prompt_filename, rank_print)
     input_ids, reqs = prepare_inputs_for_correctness_test(
-        bench_args, tokenizer, custom_prompts
+        bench_args, tokenizer, custom_prompts, images=images, model_runner=model_runner
```
polisettyvarma, line 1108: can we remove new args if not used

python/sglang/bench_one_batch.py
```diff
@@ -640,8 +1137,9 @@ def correctness_test(
 
     # Print output texts
     for i in range(len(reqs)):
-        rank_print(f"========== Prompt {i} ==========")
-        rank_print(tokenizer.decode(output_ids[i]), "\n")
+        rank_print(f"========== Prompt {i} ===========")
+        output_text = tokenizer.decode(output_ids[i])
+        rank_print(output_text, "\n")
```
polisettyvarma, line 1142: what's the advantage ?

python/sglang/bench_one_batch.py
```diff
@@ -815,9 +1313,37 @@ def latency_test(
...
+                f"Generating {bench_args.image_count} synthetic image(s) at {width}x{height} for latency test..."
+            )
+            images = generate_random_images(
+                bench_args.image_count,
+                width,
+                height,
+                rank_print,
+                image_content=bench_args.image_content,
+            )
+
```
polisettyvarma, line 1338: seems repeated, create function and reuse ?

python/sglang/bench_one_batch.py
```diff
@@ -907,6 +1441,11 @@ def latency_test(
 def main(server_args, bench_args):
     server_args.cuda_graph_max_bs = max(bench_args.batch_size)
 
+    # bench_one_batch --image-test implies multimodal execution.
+    # Keep explicit user choice intact if --enable-multimodal is provided.
+    if bench_args.image_test and server_args.enable_multimodal is None:
+        server_args.enable_multimodal = True
```
polisettyvarma, line 1447: should we set this or raise error or warn user ?

polisettyvarma (changes requested):

python/sglang/bench_one_batch.py
mingfeima: @kalyank007

mingfeima:
@kalyank007 please change according the comments.

python/sglang/bench_one_batch.py
kalyank007: Re-structured and fixed the issue

python/sglang/bench_one_batch.py
kalyank007: Done

python/sglang/bench_one_batch.py
kalyank007: Re-structured and removed this part

python/sglang/bench_one_batch.py
kalyank007: These prints help diagnose image token lookup failures — they show which token ID was searched for and the actual encoded token IDs so users can identify mismatches (e.g. wrong placeholder string or tokenizer encoding the placeholder as multiple sub-tokens).

python/sglang/bench_one_batch.py
kalyank007: verbose and local_rank_print are defined at the end of the try block — if an exception fires early (e.g. during processor loading), those names won't be in scope in the except. Could hoist the verbose check and local_rank_print definition before the try to avoid the duplication entirely.

python/sglang/bench_one_batch.py
kalyank007: All args are now used — images and model_runner were previously unused because the function body was accidentally truncated (the multimodal logic had fallen into process_images_to_multimodal_inputs as unreachable dead code). That's been fixed and both args are now exercised in the restored body.

python/sglang/bench_one_batch.py
kalyank007: Yes — ScheduleBatch.init_new calls alloc_token_slots from mem_cache/common.py, which calls tree_cache.available_and_evictable_str() and tree_cache.pretty_print() on OOM. Without these stubs on TreeCacheNamespace, any allocation failure would raise AttributeError instead of the intended RuntimeError with a useful message.

python/sglang/bench_one_batch.py
kalyank007: In TP>1 mode each rank runs this function in its own process. Using rank_print (which is a no-op on non-zero ranks) ensures the decoded output is printed exactly once instead of once per GPU, avoiding duplicate/interleaved output in the terminal.

python/sglang/bench_one_batch.py
kalyank007: Done

python/sglang/bench_one_batch.py
kalyank007: Setting it silently is intentional as a convenience — --image-test already implies multimodal, so requiring users to also pass --enable-multimodal would be redundant. A warning could make sense if users pass --image-test together with an explicit --enable-multimodal False, but the current guard (is None) already preserves any explicit user choice. An error would be overly strict for a benchmark script.

python/sglang/bench_one_batch.py
```diff
@@ -1,22 +1,1544 @@
-"""Deprecated import path for ``sglang.benchmark.one_batch``.
+"""
```
polisettyvarma: file moved to python/sglang/benchmark/one_batch.py
add changes there

polisettyvarma (changes requested):

PR #18995, opened 2026-02-19
[nit] Remove redundant server_args log in Engine.__init__

xyuzh:
## Summary
- Remove duplicate `logger.info(f"{server_args=}")` in `Engine.__init__` since the same log is already emitted in `_launch_subprocesses`, which is called immediately after.

## Test plan
- No behavioral change; only removes a redundant log statement.


Qiaolin-Yu (commented):
Why do we need to remove this?

Qiaolin-Yu:
ahh ic, make sense to me.

Qiaolin-Yu (approved):

slin1237 (approved):

PR #19009, opened 2026-02-19
Fix disagg PD bootstrap and KV transfer metrics

Kangyan-Zhou:
## Summary
- **Fix disagg PD bootstrap and KV transfer metrics tracking**: Add `bootstrap_done_time` to `TimeStats` and correctly compute `bootstrap_duration` and `alloc_waiting_duration` on both prefill and decode paths (replacing the previous `# TODO: correct set them` placeholder). Add per-request KV transfer metrics (latency, size, speed) in the disagg prefill inflight queue processing. Add missing `log_prefill_stats` call in the disagg prefill batch result path to match the non-disagg path.
- **Convert KV transfer metrics from Gauge to Histogram**: Convert 5 KV transfer metrics (`kv_transfer_speed_gbps`, `kv_transfer_latency_ms`, `kv_transfer_bootstrap_ms`, `kv_transfer_alloc_ms`, `kv_transfer_total_mb`) from Gauge to Histogram with appropriate bucket ranges for per-request observability. Remove the corresponding fields from `SchedulerStats` and the scheduler mixin (no longer needed in the Gauge pipeline).
- **Add bootstrap sub-phase breakdown to per-request time stats log**: The per-request log from `enable_request_time_stats_logging` now shows `bootstrap_queue_duration` broken down into bootstrap (handshake wait) and alloc_wait (metadata buffer allocation wait) sub-phases for both prefill and decode disaggregation modes. Previously the entire duration was opaque, making it hard to diagnose PD disagg latency issues.
- **Refactor bootstrap and KV transfer metrics into observability layer**: Add `set_bootstrap_done_time()` and `compute_and_observe_kv_transfer_metrics()` methods to `SchedulerReqTimeStats` to encapsulate timestamp setting and metrics computation in the observability layer, keeping disaggregation core files (`prefill.py`, `decode.py`) focused on orchestration.

## Test plan
- [ ] Verify Histogram metrics are correctly reported in Prometheus on a PD disaggregation setup
- [ ] Confirm `bootstrap_duration` and `alloc_waiting_duration` have reasonable values under normal load
- [ ] Confirm aborted requests are excluded from KV transfer metrics
- [ ] Verify per-request time stats log shows sub-phase breakdown (e.g. `bootstrap_queue_duration(6399.36ms) = bootstrap(5800.00ms) + alloc_wait(599.36ms)`)

### DIFF
--- a/python/sglang/srt/disaggregation/prefill.py
+++ b/python/sglang/srt/disaggregation/prefill.py
@@ -20,6 +20,7 @@ Life cycle of a request in the prefill server
 from __future__ import annotations
 
 import logging
+import time
 from collections import deque
 from http import HTTPStatus
 from typing import TYPE_CHECKING, List, Optional
@@ -42,7 +43,12 @@ from sglang.srt.disaggregation.utils import (
     poll_and_all_reduce_attn_cp_tp_group,
     prepare_abort,
 )
-from sglang.srt.managers.schedule_batch import FINISH_LENGTH, Req, ScheduleBatch
+from sglang.srt.managers.schedule_batch import (
+    FINISH_ABORT,
+    FINISH_LENGTH,
+    Req,
+    ScheduleBatch,
+)
 from sglang.srt.mem_cache.common import release_kv_cache
 from sglang.srt.mem_cache.memory_pool import HybridLinearKVPool, NSATokenToKVPool
 from sglang.srt.mem_cache.swa_memory_pool import SWAKVPool
@@ -299,6 +305,8 @@ class PrefillBootstrapQueue:
                 continue
 
             # KV.WaitingForInput - init here
+            if req.time_stats.bootstrap_done_time == 0.0:
+                req.time_stats.bootstrap_done_time = time.perf_counter()
             num_kv_indices = len(req.origin_input_ids)
             if self.req_to_metadata_buffer_idx_allocator.available_size() == 0:
                 break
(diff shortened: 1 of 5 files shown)


ShangmingCai (approved):
LGTM

python/sglang/srt/disaggregation/prefill.py
```diff
@@ -299,6 +305,8 @@ def pop_bootstrapped(
                 continue
 
             # KV.WaitingForInput - init here
+            if req.time_stats.bootstrap_done_time == 0.0:
```
sufeng-buaa: Could you change it to save the timestamp by implementing `set_bootstrap_done_time()` in `time_stats`? This will make it easier to extend metrics and tracing in the future.

python/sglang/srt/disaggregation/prefill.py
```diff
@@ -597,6 +605,65 @@ def process_disagg_prefill_inflight_queue(
         for req in done_reqs:
             req.time_stats.set_completion_time()
 
+        if self.enable_metrics:
```
sufeng-buaa: If possible, please implement this in one of the files under the observability directory—for example, by adding it to `set_completion_time()`. These core files are becoming increasingly bloated, so we should keep them as clean and concise as possible.

Kangyan-Zhou: resolved both comments, PTAL, thanks!

sufeng-buaa (approved):

python/sglang/srt/disaggregation/prefill.py
```diff
@@ -597,6 +603,27 @@ def process_disagg_prefill_inflight_queue(
         for req in done_reqs:
             req.time_stats.set_completion_time()
 
+        page_size = self.token_to_kv_pool_allocator.page_size
+        kv_item_lens = (
+            self.disagg_prefill_bootstrap_queue.kv_manager.kv_args.kv_item_lens
+        )
+        bytes_per_page_all_layers = sum(kv_item_lens)
```
merrymercy, line 610: please cache these constants as much as possible

python/sglang/srt/disaggregation/prefill.py
```diff
@@ -597,6 +603,27 @@ def process_disagg_prefill_inflight_queue(
...
+
+        for req in done_reqs:
```
merrymercy, line 612: only do this when enable_metrics

PR #19035, opened 2026-02-20
Fix: Reject requests with a duplicate request ID which can cause server crash/hang

pyc96:
## Motivation

SGLang currently allows duplicate request ID which will cause server crashes and potential data corruptions as tokenizerManager, scheduler, and detokenizer all maintain internal states keyed by rid. While the client should enforce the ID uniqueness but in reality the requests might get retried by other components. So it would be ideal to strictly enforce this in sglang server.

## Reproduce

```
python -m sglang.launch_server     --model-path {MODEL}   --tp-size 8     --host 0.0.0.0     --trust-remote-code     --port 8000
```

- If 2 requests are very closed to each other, the server will complain `Received output for rid='x' but the state was deleted in TokenizerManager.` The second request will hang
- If the first request started to decode, the server will crash: 
```
[2026-02-20 01:04:15] DetokenizerManager hit an exception: Traceback (most recent call last):
  File "/home/pyc_google_com/dev/sglang/python/sglang/srt/managers/detokenizer_manager.py", line 309, in _decode_batch_token_id_output
    s = self.decode_status[rid]
        ~~~~~~~~~~~~~~~~~~^^^^^
KeyError: 'x'

During handling of the above exception, another exception occurred:

  File "/home/pyc_google_com/dev/sglang/python/sglang/srt/managers/detokenizer_manager.py", line 311, in _decode_batch_token_id_output
    raise RuntimeError(
RuntimeError: Decode status not found for request x. It may be due to the request being evicted from the decode status due to memory pressure. Please increase the maximum number of requests by setting the SGLANG_DETOKENIZER_MAX_STATES environment variable to a bigger value than the default value. The current value is 65536. For more details, see: https://github.com/sgl-project/sglang/issues/2812
```
- It could also cause scheduler KeyError if one of the request is queued.

Client side script: https://gist.github.com/pyc96/617cbf41db11c08c5e206e532f0b3adb

## Modifications

Checks rid in tokenizerManager before sending the request scheduler. Adding the check in tokenizer since it's the first component that maintains rid mapping.

## Checklist

## Review Process


Muqi1029:
Maybe it would be better to redesign how SGLang manages its internal `rid`. These internal `rid`s are used across multiple components, such as `TokenizerManager` and `DetokenizerManager`. In a well-designed system, internal state should not be modified or relied upon by external requests.

While this PR resolves the duplicate `rid` issue, it does so by rejecting the second request with the same `rid`. That behavior doesn’t seem ideal or fully aligned with a robust system design.
CC @hnyls2002

hnyls2002:
@Muqi1029 Do you have better ideas? I think duplicate RID should be aborted and current solution is more like a workaround

Muqi1029:
Yes, as I said "Record original rid for potential bug report", this should also be used for tracing, etc.

PR #19040, opened 2026-02-20
feat: add Glm4MoeLiteConfig and fix enable_a2a_moe for GLM-4.7-Flash

lujangus:
## Summary

GLM-4.7-Flash uses architecture type `glm4_moe_lite`, which Transformers does not natively register. This causes SGLang to fail during server argument parsing when loading the model config via `AutoConfig`.

Two issues are fixed:

### 1. Missing config registration (`Glm4MoeLiteConfig`)

- Add `python/sglang/srt/configs/glm4_moe_lite.py` with `Glm4MoeLiteConfig` extending `Glm4MoeConfig`
- Adds the MLA attention parameters present in the model's `config.json`: `q_lora_rank`, `kv_lora_rank`, `qk_nope_head_dim`, `qk_rope_head_dim`, `v_head_dim`
- Sets `model_type = "glm4_moe_lite"` to match the HuggingFace model card
- Registers the class in `configs/__init__.py` and `hf_transformers_utils._CONFIG_REGISTRY` so `AutoConfig` resolution works without `trust_remote_code`

### 2. Missing `enable_a2a_moe` attribute (`AttributeError` at runtime)

`Glm4MoeLiteModel` extends `DeepseekV2Model` but calls `nn.Module.__init__` directly, bypassing `DeepseekV2Model.__init__`. The inherited `DeepseekV2Model.forward` accesses `self.enable_a2a_moe`, causing an `AttributeError` at runtime.

Fix: explicitly set `self.enable_a2a_moe = False` in `Glm4MoeLiteModel.__init__`. GLM-4.7-Flash does not use DeepEP or Mooncake all-to-all MoE dispatch, so `False` is the correct value.

## Test plan

- [ ] Launch SGLang server with `--model-path <GLM-4.7-Flash>` and verify it starts without `AttributeError` or `AutoConfig` resolution errors
- [ ] Run a basic generation request and confirm correct output
- [ ] Verify `enable_a2a_moe = False` does not affect models that do use all-to-all dispatch (DeepSeek-V3, etc.) — the fix is scoped to `Glm4MoeLiteModel.__init__` only


hnyls2002:
Thanks @lujangus! Both issues are resolved on main via #21851 and #26088 - `Glm4MoeLiteModel` is now a standalone `nn.Module` at `python/sglang/srt/models/glm4_moe_lite.py:753` that sets `_enable_a2a_moe` itself at `:296`, and the pinned `transformers==5.12.1` registers `glm4_moe_lite` natively, so no local config class is needed. Closing as obsolete - please reopen if I've missed something.

PR #19047, opened 2026-02-20
fix(remote): add device default and distributed streaming for RunAI S3 loading

smb2love:
## Motivation

When using `--load-format remote --model-path s3://...`, two issues prevent correct and efficient model loading:

1. **`create_remote_connector()` requires `device` but 3 call-sites omit it** — causes `TypeError` at startup for any S3/Redis remote path (`ModelConfig._maybe_pull_model_tokenizer_from_remote`, `get_config`, `get_tokenizer`).

2. **RunAI Model Streamer defaults to `device="cpu", is_distributed=False`** — on a multi-GPU node (TP=8), every rank reads the full model from S3 independently (8x I/O), and weights land on CPU causing a large async H2D copy backlog.

### `python/sglang/srt/connector/__init__.py` (+1 line)
- Add `device: str = "cpu"` default to `create_remote_connector()`. Only `instance://` uses `device`; S3 and Redis ignore it.

### `python/sglang/srt/model_loader/weight_utils.py` (+28 / -4 lines)
- **`runai_safetensors_weights_iterator()`**: read `SGLANG_RUNAI_STREAMER_DEVICE` (default `"auto"` → `"cuda"`) and `SGLANG_RUNAI_STREAMER_DISTRIBUTED` (default `"auto"` → `True` when distributed+GPU), pass to `streamer.stream_file()`.
- **`set_runai_streamer_env()`**: add `AWS_ENDPOINT_URL_S3` as fallback for `RUNAI_STREAMER_S3_ENDPOINT`.

With `device="cuda"`, `should_async_load()` returns `False` for GPU tensors, so the CPU async-copy backlog is also eliminated.

## Accuracy Tests

No changes to model forward code. Weight loading path only.

## Benchmarking and Profiling

Tested on 8×GPU node with DeepSeek-V3-0324 (600 GB) from S3:
- Before: 8× full S3 reads, ~45 min, CPU OOM risk
- After: 1× distributed read to GPU, ~6 min

## Checklist

- [x] Format code with pre-commit
- [x] Follow SGLang code style guidance
- [ ] Add unit tests (no existing infra for remote/RunAI path; tested manually)
- [ ] Update documentation

### DIFF
--- a/python/sglang/srt/model_loader/weight_utils.py
+++ b/python/sglang/srt/model_loader/weight_utils.py
@@ -1101,8 +1101,30 @@ def composed_weight_loader(
 def runai_safetensors_weights_iterator(
     hf_weights_files: List[str],
 ) -> Generator[Tuple[str, torch.Tensor], None, None]:
-    """Iterate over the weights in the model safetensor files."""
+    """Iterate over the weights in the model safetensor files.
+
+    Reads env vars to configure RunAI Model Streamer:
+    - SGLANG_RUNAI_STREAMER_DEVICE: "cpu"|"cuda"|"auto" (default "auto")
+    - SGLANG_RUNAI_STREAMER_DISTRIBUTED: "0"|"1"|"auto" (default "auto")
+    """
     from runai_model_streamer import SafetensorsStreamer
+
+    device = os.getenv("SGLANG_RUNAI_STREAMER_DEVICE", "auto")
+    if device == "auto":
+        device = "cuda" if torch.cuda.is_available() else "cpu"
+
+    want_distributed = (
+        device != "cpu"
+        and torch.distributed.is_initialized()
+        and torch.distributed.get_world_size() > 1
+    )
+    dist_env = os.getenv("SGLANG_RUNAI_STREAMER_DISTRIBUTED", "auto").strip().lower()
+    if dist_env == "auto":
+        is_distributed = want_distributed
+    else:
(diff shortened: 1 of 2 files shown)


smb2love:
@merrymercy @Fridge003 @ispobock  Could you please take a look at my PR?

python/sglang/srt/model_loader/weight_utils.py
```diff
@@ -1101,9 +1101,31 @@ def composed_loader(param: torch.Tensor, loaded_weight: torch.Tensor) -> None:
...
-    """Iterate over the weights in the model safetensor files."""
+    """Iterate over the weights in the model safetensor files.
+
+    Reads env vars to configure RunAI Model Streamer:
+    - SGLANG_RUNAI_STREAMER_DEVICE: "cpu"|"cuda"|"auto" (default "auto")
+    - SGLANG_RUNAI_STREAMER_DISTRIBUTED: "0"|"1"|"auto" (default "auto")
+    """
     from runai_model_streamer import SafetensorsStreamer
 
+    device = os.getenv("SGLANG_RUNAI_STREAMER_DEVICE", "auto")
```
Fridge003, line 1112: Please register environ in https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/environ.py, and update the document in https://github.com/sgl-project/sglang/blob/main/docs/references/environment_variables.md

Fridge003 (changes requested):

PR #19066, opened 2026-02-20
 [diffusion] model: optimize Qwen2.5-VL ViT and text encoder

Jasen2201:
## Summary

Replace HuggingFace's `Qwen2_5_VisionTransformerPretrainedModel` with a custom `Qwen2_5_VisionTransformer` for Qwen2.5-VL ViT, and extend `torch.compile` to cover the text encoder (ViT + LLM encode).

**Custom ViT (`qwen2_5vl.py`):**
- Unified attention backend (`flash_attention_2` / `sdpa` / `eager`) with automatic fallback: uses `flash_attn_varlen_func` when available, falls back to SDPA with `.contiguous()` fix for AMD MIOpen precision issue
- `VisionRotaryEmbeddingWithCache`: pre-computed frequency cache, defers cos/sin to indexing time
- `@lru_cache` on `_get_rope_by_thw`: caches window index + RoPE per (t, h, w) grid
- Fused `gate_up_proj` via `MergedColumnParallelLinear` in ViT MLP
- `RMSNorm` (with platform-specific fused kernels) replaces `Qwen2RMSNorm`
- `torch.compile` compatibility: removed `pin_memory` in `_invert_permutation`, use `torch.tensor_split` instead of `torch.split`

**AMD GPU precision fix:**
- Root cause: HF's SDPA path passes non-contiguous tensors (from `torch.split`) to `F.scaled_dot_product_attention`, causing incorrect results in MIOpen backend. Single-layer error is small but accumulates exponentially through 32 ViT layers (cosine_sim drops to ~0.55).
- Fix: SDPA path applies `.contiguous()` on qkv slices; FlashAttention varlen path is unaffected (uses contiguous `[seq, heads, dim]` format natively).

### Without `torch.compile`
image

### With `torch.compile`

image

## Test Plan

- [x] Unit test: custom ViT vs HF ViT numerical consistency (3 cases, atol=1e-5)
- [x] End-to-end: 40-step image generation produces correct output
- [x] Profiling traces generated for both compile/no-compile modes


test/unit/test_qwen25_vit.py
```diff
@@ -0,0 +1,180 @@
+"""
```
mickqian, line 1: could you move it under the `unit` folder?

python/sglang/multimodal_gen/runtime/models/encoders/qwen2_5vl.py
```diff
@@ -62,21 +64,573 @@
...
     eager_attention_forward,
 )
 
 logger = logging.getLogger(__name__)
 
 
+# ====================== Vision Transformer Components ======================
+
+
+def _vit_flash_attn_varlen(
```
yuan-luo: Maybe it's better put this code segment into vision.py?

Jasen2201: > Maybe it's better put this code segment into vision.py?

Agreed — done in the latest push. Moved get_vit_attn_backend, vit_attention_forward into vision.py as shared ViT utilities.

python/sglang/multimodal_gen/runtime/pipelines_core/stages/image_encoding.py
```diff
@@ -64,6 +66,42 @@ def __init__(
...
+            is_fsdp = False
+
+        if is_fsdp:
+            logger.info(
+                f"Text encoder ({self.text_encoder.__class__.__name__}) is sharded with FSDP. "
+                "Skipping torch.compile."
+            )
+            return
+
+        mode = os.environ.get("SGLANG_TORCH_COMPILE_MODE", "max-autotune-no-cudagraphs")
```
mickqian: use `envs`

mickqian: use `envs`

mickqian (commented):
nicely done

test/unit/test_qwen25_vit.py
Jasen2201: > could you move it under the `unit` folder?

Done

python/sglang/multimodal_gen/runtime/pipelines_core/stages/image_encoding.py
Jasen2201: > use `envs`

Done, I have also modified the similar usage in denoising.py and model_specific_stages/mova.py.

Issue #19063, opened 2026-02-20
[Bug] [Diffusion] Text to image model, --prompt-file-path feature not working


sushildubey171:
### Checklist

- [x] I searched related issues but found no solution.
- [x] The bug persists in the latest version.
- [ ] Issues without environment info and a minimal reproducible demo are hard to resolve and may receive no feedback.
- [ ] If this is not a bug report but a general question, please start a discussion at https://github.com/sgl-project/sglang/discussions. Otherwise, it will be closed.
- [x] Please use English. Otherwise, it will be closed.

### Describe the bug

while running sglang diffusion text to image model, when multiple prompts are passed through text file, it doesn't work.

```
Error:
Traceback (most recent call last):
  File "/root/miniforge3/envs/py312_env/bin/sglang", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/sushildu/sglang/python/sglang/cli/main.py", line 42, in main
    args.func(args, extra_argv)
  File "/home/sushildu/sglang/python/sglang/cli/generate.py", line 29, in generate
    generate_cmd(parsed_args)
  File "/home/sushildu/sglang/python/sglang/multimodal_gen/runtime/entrypoints/cli/generate.py", line 122, in generate_cmd
    results = generator.generate(sampling_params_kwargs=sampling_params_kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sushildu/sglang/python/sglang/multimodal_gen/runtime/entrypoints/diffusion_generator.py", line 165, in generate
    sampling_params = SamplingParams.from_user_sampling_params_args(
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sushildu/sglang/python/sglang/multimodal_gen/configs/sample/sampling_params.py", line 506, in from_user_sampling_params_args
    sampling_params._adjust(server_args)
  File "/home/sushildu/sglang/python/sglang/multimodal_gen/configs/sample/sampling_params.py", line 343, in _adjust
    raise TypeError(f"`prompt` must be a string, but got {type(self.prompt)}")
TypeError: `prompt` must be a string, but got <class 'NoneType'>
 ```

### Reproduction

[...]

PR #19075, opened 2026-02-20
[Diffusion][BUG] Fix reading multiple prompts from prompt file

sushildubey171:
## Motivation

This PR fixes issue https://github.com/sgl-project/sglang/issues/19063.

When running SGLang diffusion, users can provide multiple prompts via a file using the --prompt-file-path argument. However, the current implementation fails to handle multiple prompts correctly. Only the last prompt in the file gets used, and earlier prompts are overwritten.

## Modifications

Fixed request preparation logic where earlier prompts were incorrectly overwritten by the last prompt in the prompt file.
Updated the code to correctly substitute each prompt and generate the appropriate output filename for every request.

## Benchmarking and Profiling
Baseline
```
================= Serving Benchmark Result =================
Task:                                    text-to-image
Model:                                   black-forest-labs/FLUX.1-dev
Dataset:                                 vbench
--------------------------------------------------
Benchmark duration (s):                  361.59
Request rate:                            inf
Max request concurrency:                 1
Successful requests:                     20/20
--------------------------------------------------
Request throughput (req/s):              0.06
Latency Mean (s):                        18.0795
Latency Median (s):                      16.6147
Latency P99 (s):                         40.2891
--------------------------------------------------
Peak Memory Max (MB):                    22570.00
Peak Memory Mean (MB):                   22569.90
Peak Memory Median (MB):                 22570.00
============================================================
```
After fix:
```
================= Serving Benchmark Result =================
Task:                                    text-to-image
Model:                                   black-forest-labs/FLUX.1-dev
Dataset:                                 vbench
--------------------------------------------------
Benchmark duration (s):                  343.16
Request rate:                            inf
Max request concurrency:                 1
Successful requests:                     20/20
--------------------------------------------------
Request throughput (req/s):              0.06
Latency Mean (s):                        17.1577
Latency Median (s):                      16.5860
Latency P99 (s):                         25.7898
--------------------------------------------------
Peak Memory Max (MB):                    22570.00
Peak Memory Mean (MB):                   22569.90
Peak Memory Median (MB):                 22570.00
============================================================
```

## Checklist

## Review Process


python/sglang/multimodal_gen/runtime/entrypoints/diffusion_generator.py
```diff
@@ -162,15 +163,21 @@ def generate(
         """
         # 1. prepare requests
         prompts = self._resolve_prompts(sampling_params_kwargs.get("prompt"))
-        sampling_params = SamplingParams.from_user_sampling_params_args(
+        sampling_params_kwargs["prompt"] = prompts[0]
```
ping1jing2: In my opinion, this solution is a bit hacky ^_^

python/sglang/multimodal_gen/runtime/entrypoints/diffusion_generator.py
```diff
@@ -162,15 +163,21 @@ def generate(
...
 
         requests: list[Req] = []
         for p in prompts:
-            sampling_params.prompt = p
+            sampling_params = dataclasses.replace(
+                sampling_params_orig,
+                prompt=p,
+                output_file_name=None,
+            )
+            sampling_params._set_output_file_name()
```
ping1jing2, line 183: I think we don't need to change these code.

ping1jing2:
To be honest, i don't understand the meaning of `Only the last prompt in the file gets used, and earlier prompts are overwritten.`

let's review the workflow in your scenario:
1. `prompts = self._resolve_prompts(sampling_params_kwargs.get("prompt"))` : Collect prompts from prompt file when `prompt_file_path` is set, but I don't find any issue in this function.
2. `sampling_params: SamplingParams = model_info.sampling_param_cls(**kwargs)` : create a default SamplingParams instance. Notice: `SamplingParams.prompt` is None even if `--prompt` is set.
3. `sampling_params._merge_with_user_params(user_sampling_params)`: update sampling_params with user params. However, since we don't set `--prompt`, its value remains None.
4. `sampling_params._adjust(server_args)`: update sampling_params with server_args. Actually i am confused about the difference between user_sampling_params and server_args
5. `req = prepare_request(...)` prepare requests for inference.

Actually, i think it might be better to move `raise TypeError` from `_adjust` to `prepare_request` according to SRP. The purpose of `_adjust` is to update sampling_params and no reason  to check prompt here. `prepare_request` is the final stage where the prompt needs to be checked.

python/sglang/multimodal_gen/runtime/entrypoints/diffusion_generator.py
sushildubey171: Hi @ping1jing2, thanks for your review, I can remove this line and move `raise TypeError` from `_adjust` to `prepare_request`

python/sglang/multimodal_gen/runtime/entrypoints/diffusion_generator.py
sushildubey171: `dataclasses.replace` is needed to prevent a shared-object bug when there are multiple prompts in a file.

What happens without it:

We create `req` object for each prompt in a loop.
But all `req` objects in `requests` share the same `sampling_params` reference.
When the loop sets `sampling_params.prompt = p`, it mutates that same object.
So earlier requests can get overwritten and end up using the last prompt.

output_file_name bug: `_set_output_file_name()` is called once in `__post_init__` using the original prompt. So every output file gets the same name.

Let me know if you want me to move this block of code in `prepare_request`

sushildubey171: @ping1jing2, could you please review the proposal if its ok ?

ping1jing2: /gemini wdyt

ping1jing2: ok, make sense

python/sglang/multimodal_gen/runtime/entrypoints/diffusion_generator.py
ping1jing2: > Hi [@ping1jing2](https://github.com/ping1jing2), thanks for your review, I can remove this line and move `raise TypeError` from `_adjust` to `prepare_request`

ok, please change it

python/sglang/multimodal_gen/runtime/entrypoints/diffusion_generator.py
ping1jing2: I am so sorry for replying late due to work. However, I still have some concerns about this change, so I've written a simple test code below. Based on my test results, there’s no need to change this code – please let me know if I’m wrong.

```shell
# python test.py
sampling_params.prompt='a'
req.sampling_params.prompt='a'
sampling_params.prompt='b'
req.sampling_params.prompt='b'
sampling_params.prompt='c'
req.sampling_params.prompt='c'

```



<details>

<summary>My personal simple test code</summary>

```ruby
from typing import Any

@dataclass
class SamplingParams:
    prompt: str | list[str] | None = None

@dataclass(init=False)
class Req:
    sampling_params: SamplingParams | None = None
    def __init__(self, **kwargs):
        # Initialize dataclass fields
        for name, field in self.__class__.__dataclass_fields__.items():
            if name in kwargs:
                object.__setattr__(self, name, kwargs.pop(name))
            elif field.default is not MISSING:
                object.__setattr__(self, name, field.default)
            elif field.default_factory is not MISSING:
                object.__setattr__(self, name, field.default_factory())

        for name, value in kwargs.items():
            setattr(self, name, value)
    def __getattr__(self, name: str) -> Any:
        """
        Delegate attribute access to sampling_params if not found in Req.
        This is only called when the attribute is not found in the instance.
        """
        if name == "sampling_params":
            raise AttributeError(
                f"'{type(self).__name__}' object has no attribute '{name}'"
            )

        sampling_params = object.__getattribute__(self, "sampling_params")
        if sampling_params is not None and hasattr(sampling_params, name):
            return getattr(sampling_params, name)

        raise AttributeError(
            f"'{type(self).__name__}' object has no attribute '{name}'"
        )

    def __setattr__(self, name: str, value: Any) -> None:
        """
        Smart attribute setting:
        1. If field exists in Req, set it in Req
        2. Else if field exists in sampling_params, set it in sampling_params
        3. Else set it in Req (for dynamic attributes)
        """
        if name == "sampling_params":
            object.__setattr__(self, name, value)
            return

        if name in self.__class__.__dataclass_fields__:
            object.__setattr__(self, name, value)
            return

        try:
            sampling_params = object.__getattribute__(self, "sampling_params")
        except AttributeError:
            sampling_params = None

        if sampling_params is not None and hasattr(sampling_params, name):
            setattr(sampling_params, name, value)
            return

        if sampling_params is None and name in SAMPLING_PARAMS_FIELDS:
            new_sp = SamplingParams()
            object.__setattr__(self, "sampling_params", new_sp)
            setattr(new_sp, name, value)
            return

        object.__setattr__(self, name, value)


if __name__ == "__main__":
    sampling_params = SamplingParams()
    prompts = ['a', 'b', 'c']

    requests: list[Req] = []
    for p in prompts:
        sampling_params.prompt = p
        print(f'{sampling_params.prompt=}')
        req = Req(
        )
        print(f'{req.sampling_params.prompt=}')
        requests.append(req)


```

</details>

ping1jing2:
> I am so sorry for replying late due to work. However, I still have some concerns about this change, so I've written a simple test code below. Based on my test results, there’s no need to change this code – please let me know if I’m wrong.
> 
> ```shell
> # python test.py
> sampling_params.prompt='a'
> req.sampling_params.prompt='a'
> sampling_params.prompt='b'
> req.sampling_params.prompt='b'
> sampling_params.prompt='c'
> req.sampling_params.prompt='c'
> ```
> 
> My personal simple test code

@sushildubey171 please take a look at this, thanks

sushildubey171:
```
if __name__ == "__main__":
    sampling_params = SamplingParams()
    prompts = ['a', 'b', 'c']

    requests: list[Req] = []
    for p in prompts:
        sampling_params.prompt = p
        print(f'{sampling_params.prompt=}')
        req = Req(
        )
        print(f'{req.sampling_params.prompt=}')
        requests.append(req)
           
    # add this line to reproduce
    for request_idx, req in enumerate(requests):
        print(f'{req.sampling_params.prompt=}')
```

```
# python test.py
sampling_params.prompt='a'
req.sampling_params.prompt='a'
sampling_params.prompt='b'
req.sampling_params.prompt='b'
sampling_params.prompt='c'
req.sampling_params.prompt='c'
req.sampling_params.prompt='c'
req.sampling_params.prompt='c'
req.sampling_params.prompt='c'
```

@ping1jing2, yes your test case is incomplete, please see my output.  you need that for loop to reproduce.

ping1jing2:
> @ping1jing2, yes your test case is incomplete, please see my output. you need that for loop to reproduce.

thank you. got it.

ping1jing2 (approved):

Issue #10177, opened 2025-09-08
[Refactor] Benchmark Scripts Refactor


hnyls2002:
### Motivation

Current benchmark scripts (`bench_serving.py`) are too complex and may be hard to maintain and add new functions. Also, different benchmark scripts(`bench_serving`, `bench_one_batch`, `bench_one_batch_serving`, etc.) share some code, but not a good abstraction for now.

---

## Analysis & Phased Refactoring Plan

I did a thorough analysis of the current benchmark code duplication and also reviewed PR #10409. Here's a summary and a concrete phased plan.

### Current State

`bench_serving.py` is **3415 lines** — a monolith containing 5 distinct responsibilities:

| Responsibility | Approx Lines | Examples |
|---|---|---|
| Backend request functions | ~700 | `async_request_sglang_generate`, `async_request_openai_completions`, `async_request_trt_llm`, etc. |
| Dataset loading | ~800 | `sample_sharegpt_requests`, `sample_random_requests`, `sample_image_requests`, `get_dataset`, etc. |
| Benchmark orchestration | ~500 | `benchmark()`, `run_benchmark()`, warmup, profiling |
| Metrics & result reporting | ~200 | `BenchmarkMetrics`, `calculate_metrics`, result printing/saving |
| CLI arg parsing | ~400 | argparse definitions in `__main__` |

Multiple other files import from this monolith to reuse dataset/utility functions:

```python
# bench_offline_throughput.py
from sglang.bench_serving import DatasetRow, get_dataset, get_tokenizer, sample_random_requests, set_ulimit

# bench_one_batch_server_internal.py  
from sglang.bench_serving import get_processor, get_tokenizer, sample_mmmu_requests, sample_random_requests

# benchmark/hicache/bench_serving.py
from sglang.bench_serving import get_tokenizer, remove_prefix, set_ulimit
```

Key duplication:
- `bench_offline_throughput.py` (476 lines) duplicates **~130 lines** of dataset-related CLI arg definitions from `bench_serving.py`
- `benchmark/hicache/bench_serving.py` (1029 lines) **copy-pastes** `RequestFuncInput`/`RequestFuncOutput` and `async_request_openai_completions` with modifications
- `bench_one_batch_server_internal.py` (891 lines) has its own `BenchArgs` with overlapping fields

### PR #10409 Review

PR #10409 created a `python/sglang/benchmark/` package with good design decisions worth adopting:

**Good ideas to keep:**
[...]

PR #19077, opened 2026-02-20
[Refactor] Benchmark Phase 1: extract utils and datasets from bench_serving

Ratish1:
## Motivation

 This PR implements Phase 1 of issue #10177 by extracting benchmark utility and dataset logic from `python/sglang/bench_serving.py `into a new package structure under `python/sglang/benchmark/`.

## Modifications

  - Added `python/sglang/benchmark/utils.py` for benchmark utility helpers.
  - Added `python/sglang/benchmark/datasets/` with dataset-specific modules:
    - `common.py`
    - `sharegpt.py`
    - `random.py`
    - `custom.py`
    - `openai_dataset.py`
    - `image.py`
    - `mmmu.py`
    - `mooncake.py`
    - `generated_shared_prefix.py`
    - `__init__.py` with loader mapping and `get_dataset(...)`.
  - Refactored `python/sglang/bench_serving.py` to import/re-export moved variables and remove moved implementations.
  -  Tested python3 -m sglang.bench_serving --help and python3 -m sglang.bench_offline_throughput --help

## Checklist

## Review Process


python/sglang/benchmark/datasets/sharegpt.py
```diff
@@ -0,0 +1,109 @@
...
+import numpy as np
+from transformers import PreTrainedTokenizerBase
+
+from sglang.benchmark.utils import (
+    download_and_cache_hf_file,
+    is_file_valid_json,
+    remove_suffix,
+)
+
+from .common import ASSISTANT_SUFFIX, SHAREGPT_FILENAME, SHAREGPT_REPO_ID, DatasetRow
```
hnyls2002: Avoid using relative imports.

hnyls2002 (changes requested):

hnyls2002 (commented):
Thanks for working on this! The code movement is accurate and backward compatibility is properly maintained via re-exports. A few items to address:

## Design

### 1. (Suggestion) Phase 1 should not introduce class hierarchy — keep the original if-elif `get_dataset()` logic

**File**: `python/sglang/benchmark/datasets/__init__.py`

Phase 1's goal is pure code movement ("extract utils and datasets from bench_serving"), but this PR also introduces `BaseDatasetLoader` ABC + 9 `*DatasetLoader` classes + `DATASET_MAPPING` dict, rewriting `get_dataset()` dispatch logic. The original was a simple if-elif chain — this design change is out of scope for Phase 1.

Additionally, the current Loader classes are all parameter-unpacking boilerplate (extract fields from `args` and forward to `sample_*` functions), adding an abstraction layer without clear benefit.

## Type Hints

### 2. (Suggestion) `get_processor` return type is wrong

**File**: `python/sglang/benchmark/utils.py`, L70-72

Return type is annotated as `Union[PreTrainedTokenizer, PreTrainedTokenizerFast]` but it actually returns an `AutoProcessor` instance. Looks like it was copied from `get_tokenizer` without updating the return type.

### 3. (Suggestion) `DatasetRow.prompt` type is too narrow

**File**: `python/sglang/benchmark/datasets/common.py`, L18

`prompt` is typed as `str`, but it can also be `List[int]` (random dataset, `return_text=False`), `List[str]` (generated-shared-prefix multi-turn), or `List[Dict]` (openai dataset messages). Should be `Any` (already imported).

### 4. (Suggestion) `_gen_random_image_data_uri` return type syntax error

**File**: `python/sglang/benchmark/datasets/image.py`, L179-181

`(Image, str, int)` is not valid Python type hint syntax, and `Image` refers to the PIL module, not `PIL.Image.Image`. Should be `Tuple[Image.Image, str, int]`.

## Nits

### 5. Redundant `list()` call

**File**: `python/sglang/benchmark/datasets/image.py`, L211

`sum()` can iterate tuples directly — `sum(list(images_bytes))` -> `sum(images_bytes)`.

### 6. `get_dataset()` random fallback needs a comment

**File**: `python/sglang/benchmark/datasets/__init__.py`, L163

The `startswith("random")` fallback is not obvious. A comment explaining why it exists (matching original behavior where any `random-*` variant other than `random-ids` falls back to `random_sample=True`) would help readability.

### 7. `MooncakeDatasetLoader` returns `List[Dict]` instead of `List[DatasetRow]`

**File**: `python/sglang/benchmark/datasets/__init__.py`, L108-114

All other loaders return `List[DatasetRow]`, but Mooncake returns `List[Dict]` (raw trace data), breaking the `BaseDatasetLoader` contract. If keeping the class hierarchy, at least add a docstring explaining this difference.

python/sglang/benchmark/datasets/__init__.py
```diff
@@ -0,0 +1,194 @@
...
+)
+from .image import create_mm_data_row, parse_image_resolution, sample_image_requests
+from .mmmu import sample_mmmu_requests
+from .mooncake import get_mooncake_request_over_time, load_mooncake_requests
+from .openai_dataset import sample_openai_requests
+from .random import compute_random_lens, sample_random_requests
+from .sharegpt import sample_sharegpt_requests
+
+
+class ShareGPTDatasetLoader(BaseDatasetLoader):
```
hnyls2002: `datasets/__init__.py`: Phase 1 should be pure code movement. Keep the original if-elif `get_dataset()` logic instead of introducing `BaseDatasetLoader` / `*DatasetLoader` classes / `DATASET_MAPPING`. The class hierarchy can be done in a follow-up PR (consider vLLM's `BenchmarkDataset` pattern where `sample(**kwargs)` takes dataset-specific params).

python/sglang/benchmark/utils.py
```diff
@@ -0,0 +1,159 @@
...
+    ):
+        pretrained_model_name_or_path = get_model(pretrained_model_name_or_path)
+    return AutoTokenizer.from_pretrained(
+        pretrained_model_name_or_path, trust_remote_code=True
+    )
+
+
+def get_processor(
+    pretrained_model_name_or_path: str,
+) -> Union[PreTrainedTokenizer, PreTrainedTokenizerFast]:
```
hnyls2002: `get_processor` return type should be `AutoProcessor`, not `Union[PreTrainedTokenizer, PreTrainedTokenizerFast]`

hnyls2002 (changes requested):

hnyls2002:
The first step should only contain splitting the old `bench_serving.py` into different files based on the dataset. We can discuss the following designs later.

@Ratish1

python/sglang/benchmark/datasets/sharegpt.py
Ratish1: Done

python/sglang/benchmark/datasets/__init__.py
Ratish1: Done, I will make the follow up PR once this gets merged. will look into vLLM's  pattern

python/sglang/benchmark/utils.py
Ratish1: correct,  I changed it now. Thanks

hnyls2002 (approved):

PR #19076, opened 2026-02-20
[Fix] Quick fix for int32 overflow in Mooncakes' send_kvcache_slice

YAMY1234:
## Motivation

Fix overflow regression in https://github.com/sgl-project/sglang/commit/afae4c7178f57a60f45c0685f41e1ac42ae4b607#diff-3be47cd55ac1c34f7566da855340b2ac3871526c03906247d9e7fe8609e6b7aeL467-L468

## Checklist

- [ ] Format your code according to the [Format code with pre-commit](https://docs.sglang.io/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.io/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.io/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.io/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.io/developer_guide/contribution_guide.html#benchmark-the-speed).
- [ ] Follow the SGLang code style [guidance](https://docs.sglang.io/developer_guide/contribution_guide.html#code-style-guidance).

## Review Process

1. Ping Merge Oncalls to start the PR flow. See the [PR Merge Process](https://github.com/sgl-project/sglang/blob/main/.github/MAINTAINER.md#pull-request-merge-process).
2. Get approvals from [CODEOWNERS](https://github.com/sgl-project/sglang/blob/main/.github/CODEOWNERS) and other reviewers.
3. Trigger CI tests with [comments](https://docs.sglang.io/developer_guide/contribution_guide.html#how-to-trigger-ci-tests) or contact authorized users to do so.
   - `/tag-run-ci-label`, `/rerun-failed-ci`, `/tag-and-rerun-ci`
4. After green CI and required approvals, ask Merge Oncalls to merge.

PR #19086, opened 2026-02-20
[Fix][Qwen3.5] Fix KV cache slice transfer for GQA models with replicated KV heads

YAMY1234:
## Motivation

Merge with https://github.com/sgl-project/sglang/pull/19076

PD disaggregation with heterogeneous `attn_tp_size` (e.g., prefill TP4 with decode TP4/DP4/dp_attention where decode `attn_tp=1`) produces completely wrong outputs for non-MLA GQA models. This affects models like Qwen3.5-397B-A17B-FP8 which has only 2 KV heads — fewer than `tp_size`.

With GQA replication (`num_kv_heads < tp_size`), per-rank `kv_head_num = max(1, total // tp)`. The existing code back-computes total heads as `kv_head_num * attn_tp_size`, which overestimates when heads are replicated:

```
Qwen3.5: total_kv_heads=2, prefill attn_tp=4
  kv_head_num = max(1, 2//4) = 1 (replicated)
  computed total = 1 * 4 = 4  (wrong, actual = 2)
  dst_heads_per_rank = 4 // 1 = 4  (wrong, should be 2)
```

This causes each prefill rank to write KV data to incorrect offsets in the decode buffer, corrupting the transferred KV cache.

## Modifications

**`base/conn.py`**: Add `total_kv_head_num` field to `KVArgs` to carry the un-sharded total KV head count.

**`prefill.py`**: Set `total_kv_head_num` from `model_config.get_total_num_kv_heads()` during KV manager initialization.

**`mooncake/conn.py`**:
`send_kvcache_slice`: Use `total_kv_head_num` for correct head distribution. Introduce `src_replication` factor to map replicated ranks to unique head indices via `rank // replication`.

## Accuracy Tests

Model: Qwen3.5-397B-A17B-FP8, Benchmark: GSM8K (1319 examples, 5-shot), Hardware: GB200 (4 GPUs/node)

| Config | Prefill | Decode | Before Fix | After Fix |
|--------|---------|--------|------------|-----------|
| AGG + TP4/DP4/dp_attn (baseline, no disagg) | — | TP4/DP4/dp_attn, 1 node | 0.970 (GSM8k Full)| — |
| 1P1D homogeneous TP (no dp_attn) | TP4, 1 node | TP4, 1 node | 0.875 (GPQA) | — |
| 1P1D + TP4/DP4/dp_attn | TP4, 1 node | TP4/DP4/dp_attn, 1 node | **0.010** | **0.976** |
| 1P1D + TP8/DP8/dp_attn | TP4, 1 node | TP8/DP8/dp_attn, 2 nodes | **0.010** | **0.976** |

Config for 1P1D + TP4/DP4/dp_attn:
```
prefill:
  tensor-parallel-size: 4
  disaggregation-mode: "prefill"
  disaggregation-decode-tp: 4
  disaggregation-decode-dp: 4

decode:
  tp-size: 4
  dp-size: 4
  ep-size: 1
  enable-dp-attention: true
  disaggregation-mode: "decode"
```

Config for 1P1D + TP8/DP8/dp_attn:
```
prefill:
  tensor-parallel-size: 4
  disaggregation-mode: "prefill"
  disaggregation-decode-tp: 8
  disaggregation-decode-dp: 8

decode:
  tp-size: 8
  dp-size: 8
  ep-size: 1
  enable-dp-attention: true
  disaggregation-mode: "decode"
```

## Checklist

### DIFF
--- a/python/sglang/srt/disaggregation/mooncake/conn.py
+++ b/python/sglang/srt/disaggregation/mooncake/conn.py
@@ -441,9 +441,7 @@ class MooncakeKVManager(CommonKVManager):
             src_head_start_offset = 0
             num_heads_to_send = src_heads_per_rank
             unique_head_idx = local_tp_rank_in_group // src_replication
-            dst_head_start_offset = (
-                unique_head_idx * src_heads_per_rank
-            ) % dst_heads_per_rank
+            dst_head_start_offset = unique_head_idx * src_heads_per_rank
         else:
             # Send KVCache from 1 prefill instance to multiple decode instances
             src_head_start_offset = (
--- a/python/sglang/srt/disaggregation/mooncake/conn.py
+++ b/python/sglang/srt/disaggregation/mooncake/conn.py
@@ -442,6 +442,8 @@ class MooncakeKVManager(CommonKVManager):
             num_heads_to_send = src_heads_per_rank
             unique_head_idx = local_tp_rank_in_group // src_replication
             dst_head_start_offset = unique_head_idx * src_heads_per_rank
+            # Ensure the head slice fits within the decode buffer
+            assert dst_head_start_offset + num_heads_to_send <= dst_heads_per_rank
         else:
             # Send KVCache from 1 prefill instance to multiple decode instances
             src_head_start_offset = (
--- a/python/sglang/srt/disaggregation/mooncake/conn.py
+++ b/python/sglang/srt/disaggregation/mooncake/conn.py
@@ -441,7 +441,9 @@ class MooncakeKVManager(CommonKVManager):
             src_head_start_offset = 0
             num_heads_to_send = src_heads_per_rank
             unique_head_idx = local_tp_rank_in_group // src_replication
-            dst_head_start_offset = (unique_head_idx * src_heads_per_rank) % dst_heads_per_rank
+            dst_head_start_offset = (
+                unique_head_idx * src_heads_per_rank
+            ) % dst_heads_per_rank
         else:
             # Send KVCache from 1 prefill instance to multiple decode instances
             src_head_start_offset = (
--- a/python/sglang/srt/disaggregation/mooncake/conn.py
+++ b/python/sglang/srt/disaggregation/mooncake/conn.py
@@ -441,7 +441,7 @@ class MooncakeKVManager(CommonKVManager):
             src_head_start_offset = 0
             num_heads_to_send = src_heads_per_rank
             unique_head_idx = local_tp_rank_in_group // src_replication
-            dst_head_start_offset = unique_head_idx * src_heads_per_rank
+            dst_head_start_offset = (unique_head_idx * src_heads_per_rank) % dst_heads_per_rank
         else:
             # Send KVCache from 1 prefill instance to multiple decode instances
             src_head_start_offset = (
--- a/python/sglang/srt/disaggregation/mooncake/conn.py
+++ b/python/sglang/srt/disaggregation/mooncake/conn.py
@@ -420,10 +420,8 @@ class MooncakeKVManager(CommonKVManager):
         dst_tp_rank_in_group = dst_tp_rank % dst_attn_tp_size
         page_size = self.kv_args.page_size
 
-        # Use total (un-sharded) KV head count for correct head distribution.
-        # With GQA replication (num_kv_heads < tp_size), per-rank kv_head_num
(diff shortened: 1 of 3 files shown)


ShangmingCai:
@YAMY1234 Thanks for the clarification. I understand the change now.

python/sglang/srt/disaggregation/mooncake/conn.py
```diff
@@ -418,22 +418,32 @@ def send_kvcache_slice(
...
-        src_heads_per_rank = num_kv_heads
-        dst_heads_per_rank = num_kv_heads * self.attn_tp_size // dst_attn_tp_size
+        # Use total KV head count (not per-rank) for correct head distribution.
+        # Per-rank kv_head_num is max(1, total//tp) which loses info when total < tp.
+        total_kv_heads = getattr(self.kv_args, "total_kv_head_num", 0)
+        if total_kv_heads <= 0:
+            total_kv_heads = self.kv_args.kv_head_num * self.attn_tp_size
+
+        src_heads_per_rank = max(1, total_kv_heads // self.attn_tp_size)
+        dst_heads_per_rank = max(1, total_kv_heads // dst_attn_tp_size)
```
ShangmingCai, line 430: Is it possible that we don't pass `total_kv_head_num` in, just use `self.kv_args.kv_head_num * self.attn_tp_size` for all cases?

ShangmingCai: OK, I think the problem is that `self.kv_args.kv_head_num * self.attn_tp_size` won't represent the correct total heads under this case since `self.attn_tp_size` does not stand for the real attn tp size anymore, it contains multiple duplicates.

python/sglang/srt/disaggregation/mooncake/conn.py
```diff
@@ -418,22 +418,32 @@ def send_kvcache_slice(
...
         # Determine slicing parameters based on TP configuration
         if self.attn_tp_size > dst_attn_tp_size:
             # Send KVCache from multiple prefill instances to 1 decode instance
             src_head_start_offset = 0
             num_heads_to_send = src_heads_per_rank
-            dst_head_start_offset = local_tp_rank_in_group * src_heads_per_rank
+            unique_head_idx = local_tp_rank_in_group // src_replication
+            dst_head_start_offset = (
+                unique_head_idx * src_heads_per_rank
+            ) % dst_heads_per_rank
```
ShangmingCai, line 446: `% dst_heads_per_rank` should be removed since it is not possible under such cases?

ShangmingCai (commented):
Others LGTM

python/sglang/srt/disaggregation/mooncake/conn.py
YAMY1234: Yes, this was a suggestion from Gemini code review BTW😂(the third commit). You're right that mathematically it can't exceed dst_heads_per_rank. It was added as a defensive guard against potential out-of-bounds writes. Happy to remove it if you prefer — the logic is correct either way.

YAMY1234: @ShangmingCai I removed it. After thinking twice, the modulo could silently mask offset bugs by wrapping to a wrong-but-valid position. This can led to hard-to-debug data corruption.

Also added an assert instead so any out-of-range offset will fail fast with a clear error message.

Thanks for taking a second look🙏

YAMY1234:
> One last question: if we don't enable dp attention, what is the benefit of duplicating the KV heads? I am still confused about the computation logic here, although it won't affect the correctness of this PR. Just curious about whether this setup is a valid use case.

Great question, I thought through and here is my understanding:
 there are two different kinds of KV head "duplication":

1. dp_attention replication (this is the 'benefit' you talked about): When dp_attention is enabled, each rank holds full KV heads for independent attention computation (or partial dp attention, each rank group holds theirs). This is a parallelism strategy choice

1. GQA replication (this is what happened in qwen3.5 situation): When total_kv_heads < tp_size, TP requires each rank to hold at least 1 KV head, so the framework replicates them automatically (QKVParallelLinear). This is purely model-architecture driven, Qwen3.5 has only 2 KV heads, so TP4 always triggers it **"regardless of dp_attention"**. There's no "benefit"; it's just how TP handles models with very few KV heads.

ShangmingCai (approved):

PR #19101, opened 2026-02-21
[Nvidia] 3/N ModelOpt to Compressed-Tensors bridge: FP8 PCPT + FP8 block

wenscarl:
## Summary
Adds a ModelOpt → Compressed-Tensors (CT) loading path for FP8 per-channel per-token (PCPT) and FP8 block checkpoints, built on the ModelOpt package layout introduced in [#20963](https://github.com/sgl-project/sglang/pull/20963).

[#20963](https://github.com/sgl-project/sglang/pull/20963) is a structural refactor (split modelopt_quant into layers/quantization/modelopt/ with schemes, utils, and import wiring). This PR extends that tree with new routing and bridge behavior for additional ModelOpt FP8 recipes served via CompressedTensorsConfig / CompressedTensorsW8A8Fp8.

## What’s included (high level)
modelopt_scheme: scheme enum + detection for recipes such as FP8_PER_CHANNEL_PER_TOKEN and FP8_BLOCK (and related fallbacks).
modelopt_ct_bridge: conversion from ModelOpt HF / hf_quant_config style configs into compressed-tensors config (_modelopt_bridge flag, block vs PCPT paths).
weight_utils: scheme-based quant config selection; dict normalization for HF quantization_config; optional FP8 block checkpoint weight iterator (e.g. scale key / shape fixes); integration with ModelOptFp8Config when recipe stays native FP8.
loader: optional block checkpoint iterator when is_modelopt_fp8_block_checkpoint applies.
compressed_tensors / compressed_tensors_w8a8_fp8 / fp8_utils: bridge-aware behavior (e.g. KV / attention when _modelopt_bridge, channel scale loading, per-channel weight scale layout for kernels).
Docs: note on FP8 block GEMM backend env vars (alongside existing quantization table).
Tests: test_modelopt_ct_bridge, test_modelopt_scheme (as added on the branch).
Relation to [#20963](https://github.com/sgl-project/sglang/pull/20963)
Depends on / assumes the refactor in #20963 (modelopt package, schemes/, import paths).
Not a second refactor of the same scope as #20963; this PR is feature + bridge work on that foundation.
## How to run (examples)
PCPT (ModelOpt recipe → CT bridge) — checkpoint must describe FP8 per-channel per-token (e.g. matching FP8_PER_CHANNEL_PER_TOKEN in config / quant_cfg):

  --model-path <model_or_hub_id> \
  --quantization modelopt
FP8 block — checkpoint with block-style FP8 (e.g. FP8_BLOCK / fp8_pb_wo / block sizes in quant section); optional backend steering via env (see docs/references/environment_variables.md):

export SGLANG_SUPPORT_CUTLASS_BLOCK_FP8=1   # example
  --model-path <model_or_hub_id> \
  --quantization modelopt
Test plan

pytest test/srt/test_modelopt_scheme.py test/srt/test_modelopt_ct_bridge.py

Smoke load + short generation for one
PCPT
and one
block
ModelOpt-exported model

TP > 1 smoke if scales are sharded
Related
Builds on [#20963](https://github.com/sgl-project/sglang/pull/20963) — ModelOpt quantization refactor
```
  --num-shots 8 \
  --num-questions 1316 \
  --parallel 1316

Original:
python -m sglang.launch_server --model-path --tp-size 4 

Accuracy: 0.798
Invalid: 0.001
Latency: 10.652s
Output throughput: 12587.253 token/s

Fp8 per-tensor per-channel:
python -m sglang.launch_server --model-path /path/to/fp8_ptpc_ckpt/ --tp-size 4 --quantization modelopt

Accuracy: 0.790
Invalid: 0.002
Latency: 10.960s
Output throughput: 11961.431 token/s

Fp8 per-block-weight-only:
SGLANG_ENABLE_FLASHINFER_FP8_GEMM=1 python -m sglang.launch_server --model-path /path/to/fp8_pbwo_ckpt/ --tp-size 4  --quantization modelopt(deepgemm has a issue)
Accuracy: 0.784
Invalid: 0.000
Latency: 12.401s
Output throughput: 10653.914 token/s
```


python/sglang/srt/layers/quantization/compressed_tensors/modelopt_ct_bridge.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+def _get_quant_section(config: Dict[str, Any]) -> Dict[str, Any]:
+    """Get quantization section from flat or nested config."""
+    if "quantization" in config:
+        return config["quantization"]
+    return config
+
+
+def modelopt_config_to_compressed_tensors_config(
```
juhi10071998, line 56: Hi @wenscarl,

Since `modelopt_ct_bridge` is designed to be generic and support conversion for multiple quantization schemes/recipes, I suggest we use a more specific function name for each scheme. The current name doesn't clearly indicate that it handles `FP8 per-channel per-token` quantization specifically - it could also support FP4 or other schemes.

This same naming consideration applies to the function (`modelopt_config_to_compressed_tensors_config_block`) below as well. We might want to choose names that are reflective of the specific scheme config conversion they are designed for IMO, we could discuss further

cc: @Edwardf0t1

juhi10071998:
Hi @wenscarl, another thing I'd like to discuss is the design choice for handling the weight transformations when using the ckpt from ModelOpt through the CT's implementation of consuming ckpt in vLLM.

My concern is why we need to make the weight tensor shape / dtype changes at numerous places parameter.py, linear.py, compressed_tensors_w8a8_fp8.py, fp8_utils.py, loader.py.. I understand some of these may be acting like a safety net/ pass through rather then actual shape changes. 

I was wondering if instead of doing these changes at these paths we make the weight change at the source so we don't have to do at all these weight loading paths.. I am relatively new to the codebase so please forgive me if this is not making sense, happy to chat more if you prefer.


```
`Implementation: Single Normalization in Base Parameter
Add a method to the base parameter class that normalizes the incoming tensor by comparing it to self.data.shape:


# parameter.py — in BasevLLMParameter
def _normalize_loaded_weight(self, loaded_weight: torch.Tensor) -> torch.Tensor:
    """
    Normalize loaded_weight shape/dtype to match this parameter.
    Called at the start of every load method — no name heuristics needed.
    """
    param_data = self.data

    # 4D → 2D squeeze (ModelOpt block scales: (32,1,112,1) → (32,112))
    if loaded_weight.dim() == 4 and param_data.dim() == 2:
        loaded_weight = loaded_weight.squeeze().contiguous()

    # 1D → 2D (ModelOpt PCPT: (4096,) → (4096,1))
    # Only when param expects (N,1) — derived from param shape, not from name
    if (loaded_weight.dim() == 1
            and param_data.dim() == 2
            and param_data.shape[1] == 1):
        loaded_weight = loaded_weight.reshape(-1, 1).contiguous()

    # dtype: float32 for scale tensors
    # Derived from param dtype, not from name
    if (param_data.dtype == torch.float32
            and loaded_weight.dtype != torch.float32):
        loaded_weight = loaded_weight.to(torch.float32)

    return loaded_weight
Then call it at the top of every load method:


def load_column_parallel_weight(self, loaded_weight, ...):
    loaded_weight = self._normalize_loaded_weight(loaded_weight)  # ← one line
    # ... rest of the method unchanged
    assert param_data.shape == loaded_weight.shape   # now always passes
    param_data.copy_(loaded_weight)

def load_merged_column_weight(self, loaded_weight, **kwargs):
    loaded_weight = self._normalize_loaded_weight(loaded_weight)  # ← same
    # ...

def load_qkv_weight(self, loaded_weight, ...):
    loaded_weight = self._normalize_loaded_weight(loaded_weight)  # ← same
    # ...`
```

cc- @Edwardf0t1

PR #19102, opened 2026-02-21
Introduce CUDA graph debug mode with breakable CUDA graph

cctry:
Introduce **Breakable CUDA Graph** — a lightweight mechanism to insert graph breaks into CUDA graph capture. Marked operations run eagerly between captured graph segments, while the rest stays graph-captured.

## CUDA graph debug mode

```bash
# Debug mode: all ops run eagerly through graph capture/replay path
python -m sglang.launch_server --model meta-llama/Llama-3-8B --debug-cuda-graph
```

```python
# Selective graph breaks in model code
from sglang.srt.model_executor.breakable_cuda_graph.breakable_cuda_graph import non_graph

@non_graph(enable=True)
def my_dynamic_op(x):
    return some_incompatible_op(x)
```

## Breakable CUDA graph
when enable `SGLANG_USE_BREAKABLE_CUDA_GRAPH`, the decode graph is breakable. The overhead is minimal if no graph break inserted.


python/sglang/srt/model_executor/breakable_cuda_graph/breakable_cuda_graph.py
```diff
@@ -0,0 +1,270 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+import logging
+from contextvars import ContextVar
+from typing import Any, Callable, NamedTuple, Self
+
+import torch
+from cuda.bindings import runtime as rt
```
ch-wan: Should we add try-except here?

python/sglang/srt/model_executor/breakable_cuda_graph/cuda_utils.py
```diff
@@ -0,0 +1,38 @@
+# Copyright 2023-2024 SGLang Team
```
ch-wan:
```suggestion
# Copyright 2023-2026 SGLang Team
```

python/sglang/srt/model_executor/breakable_cuda_graph/breakable_cuda_graph.py
```diff
@@ -0,0 +1,270 @@
+# Copyright 2023-2024 SGLang Team
```
ch-wan:
```suggestion
# Copyright 2023-2026 SGLang Team
```

ch-wan (commented):
Left some minor comments. Also, can we also apply `--debug-cuda-graph` to piecewise cuda graph?

python/sglang/srt/server_args.py
```diff
@@ -3060,6 +3061,12 @@ def _handle_environment_variables(self):
         envs.SGLANG_ENABLE_DETERMINISTIC_INFERENCE.set(
             "1" if self.enable_deterministic_inference else "0"
         )
+        if self.debug_cuda_graph:
```
BBuf, line 3609: `--debug-cuda-graph` currently enables breakable CUDA graph on all platforms, but `BreakableCUDAGraph` is only imported on non-HIP builds. On ROCm this will fail with a `NameError` in `_capture_graph()` / `_create_device_graph()` instead of gracefully rejecting or disabling the flag. Could we either guard this CLI/env path or make the feature explicitly CUDA-only?

python/sglang/srt/model_executor/breakable_cuda_graph/breakable_cuda_graph.py
```diff
@@ -0,0 +1,274 @@
...
+        def wrapper(*args, **kwargs):
+            stream = get_current_stream()
+            if not _is_capturing(stream.cuda_stream):
+                return inner(*args, **kwargs)
+            last_graph = _end_capture_segment(stream)
+            logger.debug(f"Break graph due to function: {inner.__name__}")
+            # run the function once to allocate the output tensor captured by later graphs
+            output = inner(*args, **kwargs)  # TODO: use weak ref tensor to save memory
+
+            def f():
```
BBuf: It's not a good name

BBuf: And this only writes replay results back when both `output` and `new_out` are plain tensors. However, the wrapped `run_once_fn` returns structured outputs such as `LogitsProcessorOutput` / `PPProxyTensors`, while replay later still reads from `self.output_buffers[...]`. In those cases the eager callback produces a fresh object but the original captured buffers are never updated, so debug-mode replay can observe stale capture-time outputs. We probably need recursive in-place writeback for structured outputs before using this on the full runner.

PR #19103, opened 2026-02-21
KV Cache FP8 Downcast: JIT Migration and Optimization to Near-HBM-Limited Performance on H200

Johnsonms:
## Summary

This PR does two separate things:

1. **AOT-to-JIT migration**
   - Move `downcast_fp8`, the fused KV-cache BF16/FP16 to FP8 E4M3 downcast kernel, from the AOT `sgl-kernel` extension into the SGLang JIT kernel framework.
   - This is primarily an integration and build-simplification change.
   - It preserves the original kernel semantics and is not counted as the source of the performance improvement.

2. **JIT-kernel performance optimization**
   - Starting from the migrated **JIT scalar baseline**, optimize the kernel for H200 memory bandwidth.
   - The main optimizations are vectorized memory access, a 2D grid with fixed 256-thread blocks, and `__restrict__` pointer qualifiers.

The key point is that **AOT -> JIT alone does not materially improve performance**. The performance gain comes from the optimizations applied after the JIT baseline.

### Final result on H200

| Metric | JIT Scalar Baseline | Optimized JIT Kernel |
|---|---:|---:|
| Peak effective bandwidth | ~917 GB/s | ~4.2 TB/s |
| Peak H200 nominal-bandwidth utilization | ~19% | ~87% |
| Active-warp threshold ratio, `h=32,d=128` | `0.5 / 3.55 = 0.14x` | `16 / 3.55 = 4.51x` |
| Bytes-adjusted Little's-Law coverage, `h=32,d=128` | `(0.5 * 2 B) / (3.55 * 16 B) = 0.018x` | `(16 * 16 B) / (3.55 * 16 B) = 4.51x` |
| End-to-end latency | baseline | up to 61x lower |
| Additional large-shape gain from `__restrict__` | N/A | up to +17% |

The final kernel moves `downcast_fp8` from a scalar, under-exposed memory kernel into a vectorized, grid-parallel JIT kernel that is close to the practical H200 HBM bandwidth ceiling.

---

## Scope and Baseline Terminology

To avoid attributing performance gains to the wrong change, this PR uses the following terminology:

| Term | Meaning | Used for Performance Attribution? |
|---|---|---:|
| **Legacy AOT scalar kernel** | The old `sgl-kernel` implementation before this PR. | No. Used as a legacy reference and migration-equivalence check. |
| **JIT scalar baseline** | The direct JIT port before performance optimizations. It keeps the same scalar memory-access pattern and loop-based work assignment. | Yes. This is the performance baseline. |
| **Optimized JIT kernel** | The final JIT kernel after vectorization, 2D-grid parallelism, and `__restrict__`. | Yes. This is the optimized result. |

The AOT-to-JIT migration by itself is treated as performance-neutral. It is valuable because it reduces build complexity and moves this lightweight kernel into the JIT framework, but the interview-relevant performance story starts from the **JIT scalar baseline**.

---

## Motivation

`downcast_fp8` converts KV-cache tensors from BF16/FP16 to FP8 E4M3. It performs scaling, clamping, and FP8 conversion in a single GPU pass.

The previous implementation lived in the AOT `sgl-kernel` build path. Migrating it to the JIT framework reduces build complexity and aligns with the broader direction of moving lightweight SGLang kernels into JIT.

After migration, profiling showed that the scalar JIT baseline still had the same performance bottlenecks as the old scalar implementation:

```text
1. Low bytes per memory instruction.
2. Low exposed warp-level parallelism.
3. Conservative compiler code generation due to pointer aliasing assumptions.
```

This PR therefore separates the engineering migration from the performance work:

```text
AOT scalar kernel
      |
      |  migration only; no meaningful performance win counted here
      v
JIT scalar baseline
      |
      |  performance optimization starts here
      v
Optimized JIT kernel
```

Related context: #17865

---

### 1. JIT migration

Added the JIT implementation and wrapper:

- `python/sglang/jit_kernel/csrc/elementwise/cast.cuh`
  - CUDA implementation of the fused FP8 downcast kernel.
  - Supports BF16 and FP16 input.
  - Uses TVM FFI `TensorMatcher` validation.
- `python/sglang/jit_kernel/cast.py`
  - Python JIT wrapper.
  - Caches the compiled module per dtype via `@cache_once`.
  - Registers the CUDA wrapper with `load_jit`.
- `python/sglang/jit_kernel/tests/test_cast.py`
  - Correctness tests for BF16 and FP16.
- `python/sglang/jit_kernel/benchmark/bench_cast.py`
  - Latency and bandwidth benchmark.
- `python/sglang/jit_kernel/include/sgl_kernel/type.cuh`
  - Adds FP8 E4M3 dtype trait and platform-dependent clamp value.

Removed the old AOT implementation and registration path from `sgl-kernel`.

## Accuracy Tests
` python -m pytest python/sglang/jit_kernel/tests/test_cast.py -v`
image

## Benchmarking and Profiling
`python python/sglang/jit_kernel/benchmark/bench_cast.py`
image

### 2. JIT baseline optimization

Optimized the migrated JIT kernel in three stages:

| Stage | Change | Main Bottleneck Removed |
|---|---|---|
| Stage 1 | Vectorized 128-bit memory access | Low bytes per memory instruction |
| Stage 2 | 2D grid over `head * dim` with fixed 256-thread blocks | Low exposed warp-level parallelism |
| Stage 3 | `__restrict__` pointer qualifiers | Conservative pointer-aliasing assumptions |

---

#### Problem

The JIT scalar baseline performs scalar BF16/FP16 loads and FP8 stores. The accesses are already coalesced at the warp level, so the main issue is not transaction over-fetch. The issue is that each thread moves too little useful data per memory instruction:

```text
scalar load:     2 B/thread
vectorized load: 16 B/thread
```

This causes low bytes per memory instruction and a high dynamic memory-instruction count.

#### Change

Use 128-bit vectorized accesses:

```text
kVecSize = 16 / sizeof(T)
```

For BF16/FP16:

```text
kVecSize = 16 / 2 = 8 elements/thread
```

Each thread loads one vector of K, one vector of V, converts the elements, and stores vectorized FP8 output.

#### Result

image

#### Observations
- Kernel latency reduced by **4.0×–8.8×** across tested shapes
- Peak effective bandwidth increased by **~4.0×** (`917 → 3646 GB/s`), with the full range improving from **15–917 GB/s** to **99–3646 GB/s**
- Gains come from higher bytes/instruction, fewer load/store instructions, and more bytes in flight
- Nsight Compute then showed the next bottleneck: low occupancy (`~0.5 warps/block`), addressed in Stage 2

The kernel is still below the Little's-Law saturation threshold because the launch geometry exposes only `0.5` active warps/token for `h=32,d=128`.

---

#### Problem

After vectorization, the remaining bottleneck is low exposed warp-level parallelism.

The scalar/loop-based launch ties block size to `dim / kVecSize`:

```text
block = dim / kVecSize
```

For `dim = 128` and `kVecSize = 8`:

```text
block = 128 / 8 = 16 threads = 0.5 warps/block
```

Although the total work is over `head * dim`, that work is handled by an inner loop rather than being exposed through the CUDA grid.

#### Change

Flatten `head * dim` into a vector space and expose it through `grid.y`:

```cpp
const int num_vecs = head * dim / kVecSize;

dim3 grid(input_num_tokens, ceil_div(num_vecs, 256));
dim3 block(256);
```

For `head = 32`, `dim = 128`, and `kVecSize = 8`:

```text
num_vecs = 32 * 128 / 8 = 512
grid.y   = ceil(512 / 256) = 2
block    = 256 threads = 8 warps
```

So each token exposes:

```text
active warps/token = 2 * 8 = 16
```

#### Result

image

#### Observations

- Additional **1.4×–6.8×** speedup over the vectorized implementation
- Effective bandwidth reaches approximately **4.2 TB/s**
- Largest gains occur on smaller workloads where low occupancy was the dominant bottleneck
- The optimized launch exposes `head * dim` parallelism directly to the CUDA grid instead of processing it through an inner loop

Together with vectorization, the bytes-adjusted Little's-Law coverage moves from:

```text
0.018x -> 4.51x
```

This is the main reason the kernel moves into a near-HBM-bound regime.

---

#### Problem

Once memory width and exposed parallelism are fixed, remaining inefficiencies come from generated code.

Without `__restrict__`, the compiler cannot prove that these pointers reference independent memory regions:

```text
cache_k, cache_v, k_scale, v_scale, output_k, output_v, loc
```

This can lead to redundant reloads of values such as `k_scale[0]` and `v_scale[0]`.

#### Change

Add `__restrict__` to kernel pointer arguments:

```cpp
const T* __restrict__ cache_k,
const T* __restrict__ cache_v,
const float* __restrict__ k_scale,
const float* __restrict__ v_scale,
fp8_e4m3_t* __restrict__ output_k,
fp8_e4m3_t* __restrict__ output_v,
const int64_t* __restrict__ loc
```

#### Result

`__restrict__` does not increase Little's-Law coverage. Instead, it helps the compiler keep invariant scale values in registers and generate better code. The gain is small on tiny latency-bound shapes but reaches up to +17% on large bandwidth-bound shapes.

image

| Shape (sl, h, d) | Before | After | Δ |
|----------|----------|----------|----------|
| 64, 8, 128 | 290 GB/s | 290 GB/s | — |
| 64, 32, 128 | 948 GB/s | 945 GB/s | — |
| 256, 32, 128 | 2353 GB/s | 2422 GB/s | +3% |
| 1024, 32, 128 | 3678 GB/s | 4179 GB/s | +14% |
| 1024, 32, 256 | 3641 GB/s | 4017 GB/s | +10% |
| 2048, 32, 128 | 3456 GB/s | 4028 GB/s | +17% |
| 2048, 32, 256 | 3401 GB/s | 3935 GB/s | +16% |

### Observations

- No measurable impact on very small shapes
- Consistent gains on medium and large workloads
- Peak bandwidth improved from **3678 GB/s** to **4179 GB/s**
- Maximum observed improvement was approximately **17%**

Small workloads remain latency-bound, while larger workloads become bandwidth-bound and benefit from stronger aliasing guarantees.

---

## End-to-End Results

image

| Config | Scalar | Final | Speedup |
|----------|----------|----------|----------|
| sl=64, h=32, d=128 | 100.6 µs | 1.7 µs | 61× |
| sl=256, h=32, d=256 | 102.3 µs | 3.7 µs | 28× |
| sl=1024, h=32, d=128 | 103.0 µs | 6.0 µs | 17× |
| sl=2048, h=32, d=256 | 217.9 µs | 25.5 µs | 9× |

---

## Final Outcome

This PR should be read as:

```text
Part 1: Engineering migration
  AOT scalar kernel -> JIT scalar baseline
  Purpose: reduce build complexity and align with JIT kernel framework
  Performance attribution: not counted

Part 2: Performance optimization
  JIT scalar baseline -> optimized JIT kernel
  Purpose: move the kernel from below memory-saturation threshold to near-HBM-bound
  Performance attribution: counted
```

Final optimized JIT result:

Compared to the original implementation:
image

- Up to 61x lower kernel latency versus the JIT scalar baseline.
- Peak effective bandwidth improves from about 917 GB/s to about 4.2 TB/s.
- H200 nominal-bandwidth utilization improves from about 19% to about 87%.
- For `h=32,d=128`, active-warp threshold ratio changes from `0.14x` to `4.51x`.
- Bytes-adjusted Little's-Law coverage changes from `0.018x` to `4.51x`.
- `__restrict__` adds up to 17% improvement on larger bandwidth-bound shapes.

The core performance insight is:

```text
The migration made the kernel easier to integrate.
The optimization made the kernel fast.
```

---

## Kernel Semantics

The JIT kernel preserves the original `downcast_fp8` semantics:

```text
Input:
  k, v      : [input_num_tokens, head, dim], BF16/FP16
  k_scale  : [1], FP32
  v_scale  : [1], FP32
  loc      : [input_num_tokens], INT64

Output:
  k_out, v_out : [output_num_tokens, head, dim], UINT8 storage for FP8 E4M3
```

For each token and each KV element:

```text
out = cast_to_fp8_e4m3(clamp(input / scale, -FP8_MAX, FP8_MAX))
```

Destination indexing is preserved through:

```text
out_seq_idx = loc[token_idx]
out_offset  = (out_seq_idx * mult + offset) * head * dim
```

The platform-specific FP8 E4M3 clamp range is:

```text
CUDA E4M3:       FP8_MAX = 448.0
ROCm FNUZ:       FP8_MAX = 224.0
ROCm E4M3:       FP8_MAX = 448.0
```

---

## Benchmark Methodology

The benchmark should make the migration-vs-optimization split explicit:

```text
Legacy AOT scalar kernel    # sanity reference only
JIT scalar baseline         # performance baseline
JIT + vectorized access
JIT + vectorized access + 2D grid
JIT + vectorized access + 2D grid + __restrict__
```

Speedups are reported against the **JIT scalar baseline**, not against the act of migrating from AOT to JIT.

Effective bandwidth uses a first-order K/V traffic model:

```text
BW_eff = bytes_moved / latency
```

For BF16/FP16 input and FP8 output:

```text
bytes_moved ~= 2 * input_num_tokens * head * dim * (sizeof(input_dtype) + sizeof(fp8))
```

Since BF16/FP16 is 2 bytes and FP8 is 1 byte:

```text
bytes_moved ~= 2 * input_num_tokens * head * dim * 3 B
            = 6 * input_num_tokens * head * dim B
```

The leading factor of 2 accounts for K and V. Scale and `loc` reads are small relative to K/V tensor traffic and are ignored in this first-order model.

---

### Table 1. Optimization-stage ledger

| Stage | Baseline / Variant | Bottleneck Removed | Model-Side Effect | Observed Result |
|---|---|---|---|---:|
| 0 | Legacy AOT -> JIT scalar baseline | N/A | Integration change only | No material performance gain counted |
| 1 | JIT + vectorized access | Low memory instruction width | 2 B/thread -> 16 B/thread | Peak BW ~917 -> 3646 GB/s |
| 2 | JIT + 2D grid + 256-thread blocks | Low exposed warp-level parallelism | Active warps/token 0.5 -> 16 for `h=32,d=128` | Reaches ~4.2 TB/s peak |
| 3 | JIT + `__restrict__` | Pointer aliasing / redundant reloads | Better codegen; no MLP-ratio change | Up to +17% on large shapes |
| Final | Optimized JIT kernel | Bandwidth-bound / near-HBM-limited | LL coverage 0.018x -> 4.51x | Up to 61x lower latency |

### Table 2. H200 bandwidth utilization

H200 nominal HBM bandwidth is 4.8 TB/s. This table uses that number only as a reference ceiling.

| Variant | Peak Effective Bandwidth | Utilization vs 4.8 TB/s Nominal HBM |
|---|---:|---:|
| JIT scalar baseline | ~917 GB/s | ~19% |
| JIT + vectorized access | 3646 GB/s | ~76% |
| Optimized JIT kernel | 4179 GB/s | ~87% |

### Table 3. Little's-Law saturation coverage by optimization stage

This is the key benchmark-side metric for this PR.

Little's Law gives the intuition:

```text
required_bytes_in_flight ~= target_bandwidth * memory_latency
```

For bandwidth-oriented kernels, it is not enough to reduce arithmetic. The kernel must expose enough useful memory work to the memory system. I report two related ratios:

```text
active_warp_threshold_ratio = active_warps_per_token / active_warp_threshold

bytes_adjusted_ll_coverage =
    (active_warps_per_token * bytes_per_thread)
  / (active_warp_threshold * 16 B)
```

For the representative `head = 32, dim = 128` case:

```text
kVecSize = 16 / sizeof(BF16/FP16) = 8
active_warp_threshold ~= 3.55 active warps/token
```

| Variant | Bytes/thread | Active warps/token | Active-warp threshold ratio | Bytes-adjusted LL coverage | Interpretation |
|---|---:|---:|---:|---:|---|
| JIT scalar baseline | 2 B | 0.5 | `0.5 / 3.55 = 0.14x` | `(0.5 * 2) / (3.55 * 16) = 0.018x` | Far below threshold. Both memory width and exposed parallelism are insufficient. |
| JIT + vectorized access | 16 B | 0.5 | `0.5 / 3.55 = 0.14x` | `(0.5 * 16) / (3.55 * 16) = 0.14x` | Memory width is fixed, but launch geometry is still below threshold. |
| JIT + vectorized access + 2D grid | 16 B | 16 | `16 / 3.55 = 4.51x` | `(16 * 16) / (3.55 * 16) = 4.51x` | Crosses the modeled saturation threshold. The kernel can now become HBM-bound. |
| JIT + vectorized access + 2D grid + `__restrict__` | 16 B | 16 | `4.51x` | `4.51x` | Little's-Law coverage is unchanged; improvement comes from compiler code generation. |

Interpretation:

```text
JIT scalar baseline coverage = 0.018x
vectorized-only coverage     = 0.14x
2D-grid coverage             = 4.51x
```

This metric is not a direct speedup predictor. It is a bottleneck classifier:

```text
LL coverage < 1x:
  The kernel likely does not expose enough useful memory work.

LL coverage > 1x and bandwidth utilization is high:
  The kernel is likely near bandwidth-bound.

LL coverage > 1x but bandwidth utilization is still low:
  Look for codegen, instruction mix, register pressure, memory transaction efficiency, or aliasing issues.
```

For this PR, the coverage ratio explains the optimization sequence:

```text
0.018x  -> scalar baseline is far below saturation
0.14x   -> vectorization fixes memory width but not exposed parallelism
4.51x   -> 2D grid crosses the saturation threshold
4.51x   -> __restrict__ does not change MLP; it improves generated code
```

### Table 4. End-to-end latency

| Config | JIT Scalar Baseline | Optimized JIT Kernel | Speedup |
|---|---:|---:|---:|
| `sl=64, h=32, d=128` | 100.6 us | 1.7 us | 61x |
| `sl=256, h=32, d=256` | 102.3 us | 3.7 us | 28x |
| `sl=1024, h=32, d=128` | 103.0 us | 6.0 us | 17x |
| `sl=2048, h=32, d=256` | 217.9 us | 25.5 us | 9x |

The speedup is largest on smaller shapes because the baseline suffers from both low memory width and low exposed parallelism. Larger shapes become more bandwidth-bound, so the final speedup is smaller but the achieved bandwidth is much closer to the hardware limit.

### Table 5. `__restrict__` impact

After vectorized access and 2D-grid parallelism, the kernel is already bandwidth-oriented. `__restrict__` does not change Little's-Law coverage; it improves compiler code generation by giving the compiler stronger aliasing information.

| Shape `(sl, h, d)` | Before `__restrict__` | After `__restrict__` | Delta |
|---|---:|---:|---:|
| `(64, 8, 128)` | 290 GB/s | 290 GB/s | 0% |
| `(64, 32, 128)` | 948 GB/s | 945 GB/s | 0% |
| `(256, 32, 128)` | 2353 GB/s | 2422 GB/s | +3% |
| `(1024, 32, 128)` | 3678 GB/s | 4179 GB/s | +14% |
| `(1024, 32, 256)` | 3641 GB/s | 4017 GB/s | +10% |
| `(2048, 32, 128)` | 3456 GB/s | 4028 GB/s | +17% |
| `(2048, 32, 256)` | 3401 GB/s | 3935 GB/s | +16% |

---

## Correctness and Validation

Correctness was validated for both BF16 and FP16 inputs:

```bash
python -m pytest python/sglang/jit_kernel/tests/test_cast.py -v
```

Performance was benchmarked with:

```bash
python python/sglang/jit_kernel/benchmark/bench_cast.py
```

The benchmark reports latency and effective bandwidth using the same first-order K/V traffic model:

```text
total_bytes = input_num_tokens * head * dim * 6
```

---

## Risk and Compatibility

The optimized JIT kernel preserves the original behavior:

- Same input and output tensor semantics.
- Same FP8 E4M3 clamping behavior.
- Same scale handling.
- Same destination indexing through `loc`, `mult`, and `offset`.
- Same BF16/FP16 input support.

Implementation considerations:

- The vectorized kernel assumes the target KV-cache layout is compatible with 128-bit vectorized accesses.
- The tested head dimensions are multiples of the BF16/FP16 vector size.
- `__restrict__` assumes that input tensors, output tensors, scale tensors, and `loc` do not alias. This matches the intended API usage.
- The JIT wrapper compiles and caches the module per dtype, so compilation overhead is paid once per dtype.

---

<details>
<summary>Appendix A: Detailed Little's-Law Derivation</summary>

## A.1 First-order bandwidth model

For one token, one head, and one dimension element, the kernel reads one K element, reads one V element, writes one K FP8 element, and writes one V FP8 element.

For BF16/FP16 input and FP8 output:

```text
K input read:    2 B
V input read:    2 B
K output write:  1 B
V output write:  1 B
Total:           6 B per element
```

Therefore:

```text
bytes_moved ~= input_num_tokens * head * dim * 6
```

## A.2 Vectorization ratio

Before vectorization:

```text
bytes/thread = sizeof(BF16/FP16) = 2 B
```

After vectorization:

```text
kVecSize = 16 / sizeof(T) = 8
bytes/thread = 8 * 2 B = 16 B
```

Therefore:

```text
bytes/thread ratio = 16 B / 2 B = 8x
```

## A.3 Active-warp ratio for `head = 32, dim = 128`

Before the 2D-grid rewrite:

```text
block = dim / kVecSize
      = 128 / 8
      = 16 threads
      = 0.5 warps
```

After the 2D-grid rewrite:

```text
num_vecs = head * dim / kVecSize
         = 32 * 128 / 8
         = 512

grid.y = ceil(num_vecs / 256)
       = ceil(512 / 256)
       = 2

block = 256 threads = 8 warps
```

Therefore:

```text
active warps/token before = 0.5
active warps/token after  = 2 * 8 = 16
active-warp ratio         = 16 / 0.5 = 32x
```

Using the active-warp threshold from the memory-saturation model:

```text
active_warp_threshold ~= 3.55 active warps/token
```

The threshold ratio changes from:

```text
0.5 / 3.55 = 0.14x
```

to:

```text
16 / 3.55 = 4.51x
```

This isolates the Stage-2 launch-geometry effect.

## A.4 Bytes-adjusted Little's-Law coverage

To include both launch geometry and memory width, use:

```text
bytes_adjusted_ll_coverage =
    (active_warps_per_token * bytes_per_thread)
  / (active_warp_threshold * 16 B)
```

Then:

```text
JIT scalar baseline = (0.5 * 2 B)  / (3.55 * 16 B) = 0.018x
vectorized only     = (0.5 * 16 B) / (3.55 * 16 B) = 0.14x
2D-grid optimized   = (16  * 16 B) / (3.55 * 16 B) = 4.51x
```

This is a bottleneck-removal metric, not a direct speedup predictor.

## A.5 Combined exposure proxy

Combining Stage 1 and Stage 2:

```text
exposure_ratio ~= bytes/thread ratio * active-warp ratio
               ~= 8x * 32x
               ~= 256x
```

This explains why the optimized kernel can move from below-threshold execution to a near-HBM-bound regime.

</details>

<details>
<summary>Appendix B: Whole-PR Performance Ledger</summary>

## Table 12. Whole-PR performance ledger

| ID | Area | Baseline State | Change | Model Variable Improved | Representative Ratio | Measured Evidence | Note |
|---:|---|---|---|---|---:|---|---|
| 1 | Integration | Legacy AOT implementation | Move `downcast_fp8` to JIT | Build complexity / modularity | N/A | JIT wrapper and CUDA kernel added | Not counted as speedup |
| 2 | Performance baseline | Direct scalar JIT port | Define JIT scalar baseline | Clean attribution | N/A | AOT -> JIT alone has no material gain | Speedups start here |
| 3 | Correctness | Existing AOT semantics | Add BF16/FP16 tests | Functional equivalence | N/A | `test_cast.py` | Preserves scale, clamp, `loc`, `mult`, `offset` |
| 4 | Benchmarking | No dedicated staged accounting | Add latency/BW benchmark | Reproducibility | N/A | `bench_cast.py` | Reports bandwidth via K/V traffic model |
| 5 | Memory width | Scalar BF16/FP16 access | 128-bit vectorized access | Bytes per memory instruction | 2 B/thread -> 16 B/thread = 8x | Peak BW ~917 -> 3646 GB/s | Next bottleneck: exposed parallelism |
| 6 | Work assignment | `head * dim` hidden behind loop | One vector per thread | Grid-visible work | Loop work -> grid work | Stage-2 gains over vectorized version | Removes loop-based under-exposure |
| 7 | Launch geometry | `block = dim / kVecSize` | `block = 256` | Warps/block | 0.5 -> 8 for `d=128` | Effective BW reaches ~4.2 TB/s | Occupancy bottleneck removed |
| 8 | Grid geometry | Low active warps/token | `grid.y = ceil(num_vecs / 256)` | Active warps/token | 0.5 -> 16 for `h=32,d=128` | 32x active-warp increase | Crosses saturation threshold |
| 9 | Active-warp threshold | Below modeled threshold | 2D-grid rewrite | Threshold ratio | `0.14x -> 4.51x` | Kernel moves into HBM-bound regime | Isolates launch-geometry effect |
| 10 | LL coverage | Low bytes/thread + low active warps | Vectorization + 2D grid | Bytes-adjusted coverage | `0.018x -> 4.51x` | Explains memory-saturation transition | Main benchmark-side metric |
| 11 | Compiler aliasing | Pointers may alias | Add `__restrict__` | Reload elimination / register retention | N/A | Up to +17% large-shape gain | No coverage-ratio change |
| 12 | FP8 clamp | Platform-specific behavior | Add `kFP8E4M3Max` trait | Portability correctness | N/A | CUDA 448, ROCm FNUZ 224, ROCm E4M3 448 | Keeps CUDA H200 behavior |
| 13 | End-to-end latency | JIT scalar baseline | Optimized JIT kernel | Overall latency | Up to 61x | 100.6 us -> 1.7 us for `sl=64,h=32,d=128` | Smaller shapes benefit most |
| 14 | HBM utilization | Far below H200 bandwidth | Near-HBM-limited kernel | Bandwidth utilization | ~19% -> ~87% | ~917 GB/s -> 4179 GB/s | Close to practical ceiling |

</details>

---

## References

- NVIDIA H200 GPU nominal memory bandwidth: https://www.nvidia.com/en-us/data-center/h200/

## Checklist

## Review Process


python/sglang/jit_kernel/csrc/elementwise/cast.cuh
```diff
@@ -15,8 +27,8 @@ struct ConvertToFP8<__nv_bfloat16> {
 };
 
 template <>
-struct ConvertToFP8<half> {
-  static __device__ __nv_fp8_storage_t convert_to_fp8(half value) {
+struct ConvertToFP8<__half> {
```
DarkSharpness: Can we move these type casting related utilites to `<sgl_kernel/type.cuh>`?

Johnsonms: Good suggestion! will do right now

Johnsonms: Done, Thanks @DarkSharpness

sgl-kernel/csrc/common_extension.cc
```diff
@@ -96,11 +96,6 @@ TORCH_LIBRARY_FRAGMENT(sgl_kernel, m) {
       "                 Tensor cos_sin_cache, bool is_neox) -> ()");
   m.impl("rotary_embedding", torch::kCUDA, &rotary_embedding);
 
-  m.def(
```
BBuf, line 93: This kernel has a negligible impact on the sgl-kernel binary size, so we don't need to remove this operator here.

python/sglang/jit_kernel/csrc/elementwise/cast_scalar.cuh
```diff
@@ -0,0 +1,123 @@
...
+    tvm::ffi::TensorView k_out,
+    tvm::ffi::TensorView v_out,
+    tvm::ffi::TensorView k_scale,
+    tvm::ffi::TensorView v_scale,
+    tvm::ffi::TensorView loc,
+    int64_t mult,
+    int64_t offset) {
+  using namespace host;
+
+  auto input_sl = SymbolicSize{"input_sl"};
```
BBuf: I think `input_num_tokens` would be better.

python/sglang/jit_kernel/csrc/elementwise/cast_scalar.cuh
```diff
@@ -0,0 +1,123 @@
...
+
+  TensorMatcher({input_sl, head, dim}).with_dtype<T>().with_device(device).verify(k);
+  TensorMatcher({input_sl, head, dim}).with_dtype<T>().with_device(device).verify(v);
+  TensorMatcher({out_sl, head, dim}).with_dtype<uint8_t>().with_device(device).verify(k_out);
+  TensorMatcher({out_sl, head, dim}).with_dtype<uint8_t>().with_device(device).verify(v_out);
+  TensorMatcher({1}).with_dtype<float>().with_device(device).verify(k_scale);
+  TensorMatcher({1}).with_dtype<float>().with_device(device).verify(v_scale);
+  TensorMatcher({input_sl}).with_dtype<int64_t>().with_device(device).verify(loc);
+
+  const int isl = static_cast<int>(input_sl.unwrap());
```
BBuf:
```suggestion
  const int input_num_tokens = static_cast<int>(input_sl.unwrap());
```

python/sglang/jit_kernel/csrc/elementwise/cast_scalar.cuh
```diff
@@ -0,0 +1,123 @@
...
+  const int h = static_cast<int>(head.unwrap());
+  const int d = static_cast<int>(dim.unwrap());
+
+  const int vec_size = 8;
+  dim3 grid(isl);
+  dim3 block(std::min(d / vec_size, 1024));
+
+  const T max_fp8 = static_cast<T>(448.0f);
```
BBuf: This variable can be moved to a more generic file, and we need to distinguish its values for CUDA and AMD, referencing the relevant code in sgl-kernel.

python/sglang/jit_kernel/csrc/elementwise/cast_scalar.cuh
```diff
@@ -0,0 +1,123 @@
...
+  const T min_fp8 = static_cast<T>(-448.0f);
```
BBuf: Same as above.

python/sglang/jit_kernel/benchmark/bench_cast.py
```diff
@@ -0,0 +1,135 @@
...
+from sglang.jit_kernel.utils import cache_once, load_jit, make_cpp_args
+
+DEVICE = DEFAULT_DEVICE
+DTYPE = torch.bfloat16
+
+
+# ── Scalar baseline kernel ─────────────────────────────────────────────────────
+
+
+@cache_once
```
BBuf: Why not use `https://github.com/sgl-project/sglang/pull/19103/changes#diff-22b2a4e98024dd02ae25715814d3229712587aa97645c18171d6fc4a16d9a56bR24` directly.

sgl-kernel/csrc/common_extension.cc
Johnsonms: Got the rule. got them recovered, that'd be easy for apple to apple comparing btw aot and jit. So I removed  cast_scalar.cuh file, as which is just be baseline of tests

python/sglang/jit_kernel/csrc/elementwise/cast_scalar.cuh
Johnsonms: Removed this file, as we have aot part recovered

python/sglang/jit_kernel/csrc/elementwise/cast_scalar.cuh
Johnsonms: Sure, my fault, Changed. Thanks @BBuf

python/sglang/jit_kernel/csrc/elementwise/cast_scalar.cuh
Johnsonms: fixed, Thanks

python/sglang/jit_kernel/include/sgl_kernel/type.cuh
```diff
@@ -73,3 +77,65 @@ SGL_DEVICE To cast(const From& value) {
...
+constexpr float kFP8E4M3Max = 448.0f;
+#endif  // HIP_FP8_TYPE_FNUZ
+#endif  // USE_ROCM
+
+// ---------------------------------------------------------------------------
+// FP8 cast helpers (used by elementwise kernels)
+// ---------------------------------------------------------------------------
+
+template <typename T>
+struct ConvertToFP8 {
```
DarkSharpness: Can we unify these type casting traits into `dtype_trait`? Using some unified cast primitives can be easier to understand. You may register new traits for fp8 with `SGL_REGISTER_DTYPE_TRAIT` macro.

python/sglang/jit_kernel/cast.py
```diff
@@ -0,0 +1,52 @@
+from __future__ import annotations
```
DarkSharpness, line 1: Where do we use this kernel in practice?

python/sglang/jit_kernel/csrc/elementwise/cast.cuh
```diff
@@ -0,0 +1,137 @@
...
+
+constexpr int kBlockSize = 256;
+
+template <typename T>
+__global__ void fused_downcast_kernel(
+    const T* cache_k,
+    const T* cache_v,
+    const float* k_scale,
+    const float* v_scale,
+    __nv_fp8_storage_t* output_k,
```
DarkSharpness: Do not use `__nv_fp8_storage_t`. Try use `fp8_e4m3_t`

Johnsonms: Yes, you are right, one for storage, another for computation

python/sglang/jit_kernel/include/sgl_kernel/type.cuh
Johnsonms: That's good idea, I unified them. Cc @BBuf

python/sglang/jit_kernel/cast.py
Johnsonms: So far, I did not find where this kernel is used

python/sglang/jit_kernel/benchmark/utils.py
```diff
@@ -38,5 +44,5 @@ def run_benchmark(
         Tuple of (median_us, max_us, min_us)
     """
     quantiles = quantiles or DEFAULT_QUANTILES
-    ms, min_ms, max_ms = triton.testing.do_bench_cudagraph(fn, quantiles=quantiles)
+    ms, min_ms, max_ms = triton.testing.do_bench(fn, quantiles=quantiles)
```
DarkSharpness, line 42: This will break backward compatibility. Please revert this change. If your test is not compatibility with the cuda-graph, you can add an argument like `use_cuda_graph` to this helper function and enable it by default so that it won't break any compatibility.

python/sglang/jit_kernel/include/sgl_kernel/type.cuh
```diff
@@ -59,6 +63,19 @@ SGL_REGISTER_DTYPE_TRAIT(
 SGL_REGISTER_DTYPE_TRAIT(
     bf16x2_t, void, SGL_REGISTER_TYPE_END; SGL_REGISTER_FROM_FUNCTION(fp32x2_t, __float22bfloat162_rn););
 
+#ifndef USE_ROCM
+SGL_REGISTER_DTYPE_TRAIT(
+    fp8_e4m3_t, fp8x2_e4m3_t, SGL_REGISTER_TYPE_END; SGL_DEVICE static self_t from(const bf16_t& x) {
+      self_t r;
```
DarkSharpness: According to https://docs.nvidia.com/cuda/cuda-math-api/cuda_math_api/group__CUDA__MATH__FP8__MISC.html, there's a default constructor of fp8 type from `__half` and `__nv_bfloat16`. In this case I guess we don't need to register `from` function. The default fallback using `static_cast` might be ok. You may give it a try.

python/sglang/jit_kernel/include/sgl_kernel/type.cuh
```diff
@@ -1,6 +1,10 @@
 #pragma once
 #include <sgl_kernel/utils.cuh>
 
+#include <cuda_bf16.h>
+#include <cuda_fp16.h>
+#include <cuda_fp8.h>
```
DarkSharpness: This may break AMD compatibility. Remove this. Related headers are already included in `sgl_kernel/utils.cuh` I guess.

python/sglang/jit_kernel/include/sgl_kernel/type.cuh
Johnsonms: Verified and worked, added more testing case. Thanks MAN!

python/sglang/jit_kernel/include/sgl_kernel/type.cuh
Johnsonms: You are right, let's get them scoped

python/sglang/jit_kernel/benchmark/utils.py
Johnsonms: Got it. Thanks

DarkSharpness (approved):

BBuf (approved):

PR #17213, opened 2026-01-16
refactor context parallel state 

dongjiyingdjy:
## Motivation

Context parallelism is essential in long context LLM inference. It splits a long input sequence across multiple GPUs so attention can be computed in parallel, drastically reducing latency, which enables practical million-token context windows.
Previously, DeepSeek-V3.2 already supported CP and could use it together with DP. We aim to support the combination of CP, DP, and TP, and make it easier to apply to other models. To achieve this, we first refactored the original implementation.

## Modifications

- Add _ATTN_CP, _ATTN_TP and _MOE_CP as new GroupCoordinator;
- Add attn_cp_size and moe_cp_size as new server_agrs;
- Updated the relevant interfaces and processing logic in Scheduler.

## Accuracy Tests

```python
============================================================
DeepSeek-V3.2-Exp CP Single Node Results Summary
Dataset: gsm8k
Baseline: 0.935
============================================================

Model 1: deepseek-ai/DeepSeek-V3.2-Exp
  Accuracy: PASS
  Score: 0.980

Model 2: deepseek-ai/DeepSeek-V3.2-Exp
  Accuracy: PASS
  Score: 0.980

============================================================
OVERALL: ALL TESTS PASSED
============================================================
```
## Checklist

- [x] Format your code according to the [Format code with pre-commit](https://docs.sglang.io/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [x] Add unit tests according to the [Run and add unit tests](https://docs.sglang.io/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [x] Update documentation according to [Write documentations](https://docs.sglang.io/developer_guide/contribution_guide.html#write-documentations).
- [x] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.io/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.io/developer_guide/contribution_guide.html#benchmark-the-speed).
- [x] Follow the SGLang code style [guidance](https://docs.sglang.io/developer_guide/contribution_guide.html#code-style-guidance).

## Review Process

1. Ping Merge Oncalls to start the PR flow. See the [PR Merge Process](https://github.com/sgl-project/sglang/blob/main/.github/MAINTAINER.md#pull-request-merge-process).
2. Get approvals from [CODEOWNERS](https://github.com/sgl-project/sglang/blob/main/.github/CODEOWNERS) and other reviewers.
[...]

PR #19119, opened 2026-02-21
[DSv32] Support CP + P/D

vladnosiv:
## Motivation

After PR https://github.com/sgl-project/sglang/pull/17213 DeepSeek-V3.2 with MLA + Context Parallelism in P/D disaggregation could stall or fail due to inconsistent KV transfer state across CP/TP participants.
**UPD**: there were problems under load before refactoring

Observed bugs:
* bootstrap registration collisions - multiple CP ranks overwriting bootstrap since they have attn_tp_rank = 0
* status desynchronization across CP/TP-attn participants

This PR makes rank ownership and status propagation deterministic so KV transfer works consistently across TP only, CP-enabled, and mixed TP-attn x CP configurations.

## Modifications

* Authoritative prefill registration for MLA+CP: 
  * In prefill disaggregation mode with MLA backend and `attn_cp_size > 1`, only CP rank 0 registers to bootstrap per attention-TP shard
  * Non-authoritative CP ranks are treated as explicit no-op participants
* Non-authoritative prefill senders start from `WaitingForInput` (instead of bootstrapping), matching their no-op role
* In Mooncake KV manager, when a rank has no transfer chunks but receives `is_last=True`, request state is forced to `Success`
* Added `poll_and_all_reduce_attn_groups` in prefill flow for all-reduce over CP/TP groups

## Accuracy Tests

Setup:
* Nvidia Dynamo
* 1 Prefill: CP8 EP1
* 1 Decode: DP8 EP8

Bench (adapted to openai-compatible api: https://github.com/sgl-project/sglang/pull/19231)

```
python benchmark/gsm8k/bench_sglang.py --port 8000 --num-questions 500 --num-shots 48 --parallel 100
```

Results:
```
Accuracy: 0.960
Invalid: 0.000
Latency: 53.732 s
```


python/sglang/srt/disaggregation/prefill.py
```diff
@@ -62,6 +63,29 @@
 logger = logging.getLogger(__name__)
 
 
+def poll_and_all_reduce_attn_groups(
+    pollers,
+    attn_cp_cpu_group: ProcessGroup,
+    attn_tp_cpu_group: Optional[ProcessGroup] = None,
+):
```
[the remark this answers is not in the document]
vladnosiv: fixed

python/sglang/srt/disaggregation/mooncake/conn.py
```diff
@@ -1075,6 +1075,8 @@ def add_transfer_request(
             # This means that the current rank is a dummy rank for this request,
             # and it has already been marked as success, so there is no need to
             # add further chunks into the transfer queue.
+            if is_last and bootstrap_room in self.request_status:
+                self.update_status(bootstrap_room, KVPoll.Success)
```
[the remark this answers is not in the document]
vladnosiv: added a comment in the code for clarity

whybeyoung:
@xu-yfei  can you take alook ?

python/sglang/srt/disaggregation/common/conn.py
```diff
@@ -78,8 +80,18 @@ def __init__(
...
         self.attn_dp_size = get_attention_dp_size()
         self.attn_dp_rank = get_attention_dp_rank()
+        # For MLA + CP prefill, a single authoritative CP rank per attn-TP rank
+        # is enough because KV is already rebuilt to full sequence on each CP rank.
+        self.should_register_prefill = not (
+            disaggregation_mode == DisaggregationMode.PREFILL
+            and is_mla_backend
+            and self.attn_cp_size > 1
+            and self.attn_cp_rank != 0
+        )
```
ShangmingCai, line 99: The logic here is a bit convoluted.
Maybe we should change the logic from "should_register_" to "skip_register_"?

python/sglang/srt/disaggregation/mooncake/conn.py
```diff
@@ -1070,7 +1073,11 @@ def add_transfer_request(
         if bootstrap_room not in self.transfer_infos:
             # This means that the current rank is a dummy rank for this request,
             # and it has already been marked as success, so there is no need to
-            # add further chunks into the transfer queue.
+            # add further chunks into the transfer queue. We still guard on
+            # request_status membership to avoid recreating terminal status for
+            # requests that have already been cleared on this rank.
+            if is_last and bootstrap_room in self.request_status:
```
ShangmingCai: Does it mean `kv_chunk.is_last`?

python/sglang/srt/disaggregation/common/conn.py
ShangmingCai, line 99: The logic here is a bit convoluted.
Maybe we should change the logic from "should_register_" to "skip_register_"?

vladnosiv: done

python/sglang/srt/disaggregation/mooncake/conn.py
vladnosiv: yep, I renamed this argument for readability

python/sglang/srt/disaggregation/common/conn.py
vladnosiv: duplicate comment

python/sglang/srt/disaggregation/common/conn.py
```diff
@@ -293,9 +305,17 @@ def __init__(
...
         self.curr_idx = 0
-        self.kv_mgr.update_status(self.bootstrap_room, KVPoll.Bootstrapping)
+        initial_status = KVPoll.Bootstrapping
+        if (
+            self.kv_mgr.disaggregation_mode == DisaggregationMode.PREFILL
+            and self.kv_mgr.skip_register_prefill
+        ):
+            # Non-authoritative CP ranks are dummy participants.
+            initial_status = KVPoll.WaitingForInput
+        self.kv_mgr.update_status(self.bootstrap_room, initial_status)
```
ShangmingCai, line 361: For kvsender, `self.kv_mgr.disaggregation_mode == DisaggregationMode.PREFILL` is True for sure.

vladnosiv: simplified

vladnosiv:
Cherry-picked to https://github.com/sgl-project/sglang/pull/19504

PR #19163, opened 2026-02-22
[Feature] Stronger transformers modeling backend with TP, PP, MoE, VLMs, and torch compile

adarshxs:
## Motivation

Adds a generic modeling backend that uses HF transformers models directly via AutoModel.from_config(), enabling any model with a tp_plan, pp_plan and custom attention support to run on SGLang without a dedicated model implementation.

## Modifications

 Architecture

  Mixin-based design that composes capabilities:
  • `TransformersBase` - core class: meta-device init, recursive module replacement (Linear -> TP, RMSNorm -> fused), attention instance creation, PP support, weight loading via
    AutoWeightsLoader
  • `CausalMixin` - LM head + logits processor
  • `EmbeddingMixin` - pooling for embedding models
  • `MoEMixin` - auto-detects experts modules and replaces with TransformersFusedMoE (custom-op backed, fused kernels, EPLB recording)
  • `MultiModalMixin` - vision/audio/video encoder dispatch, M-RoPE, token_type_ids propagation

  Concrete classes cover all combinations: TransformersForCausalLM, TransformersMoEForCausalLM, TransformersMultiModalForCausalLM, TransformersMultiModalMoEForCausalLM, plus embedding
  and sequence classification variants.

## Testing
Conducted the following tests. Infrastructure = (1xH100, 4xA100):

| Model                               | Config                                                                 | Status |
|--------------------------------------|------------------------------------------------------------------------|--------|
| Qwen3-0.6B                           | TP=1, TP=2, TP=1 + Torch Compile, TP=2 + Torch Compile               | PASS   |
| Qwen3-30B-A3B (MoE, 128 experts)     | TP=1, TP=2                                                            | PASS   |
| Gemma3-4B-IT (VLM)                   | TP=4, TP=4 + Torch Compile, text+image                                | PASS   |
| Qwen3-VL-2B (VLM)                    | TP=1, TP=1 + Torch Compile, text                                       | PASS   |

| #  | Category                     | Model                         | Architecture                      | Config                         |
|----|------------------------------|--------------------------------|-----------------------------------|--------------------------------|
| 1  | Non-MoE LLM Tests            | Qwen3-0.6B                    | TransformersForCausalLM           | TP=1                           |
| 2  | Non-MoE LLM Tests            | Qwen3-0.6B                    | TransformersForCausalLM           | TP=2                           |
| 3  | Non-MoE LLM Tests            | Qwen3-0.6B                    | TransformersForCausalLM           | TP=1+compile                   |
| 4  | Non-MoE LLM Tests            | Qwen3-0.6B                    | TransformersForCausalLM           | TP=2+compile                   |
| 5  | Non-MoE LLM Tests            | Qwen2.5-7B-Instruct           | TransformersForCausalLM           | Single GPU                     |
| 6  | Non-MoE LLM Tests            | Qwen2.5-7B-Instruct           | TransformersForCausalLM           | TP=2                           |
| 7  | Non-MoE LLM Tests            | Qwen2.5-7B-Instruct           | TransformersForCausalLM           | TP=4                           |
| 8  | MoE LLM Tests                | Qwen3-30B-A3B (128 experts)   | TransformersMoEForCausalLM        | TP=1                           |
| 9  | MoE LLM Tests                | Qwen3-30B-A3B (128 experts)   | TransformersMoEForCausalLM        | TP=2                           |
| 10 | MoE LLM Tests                | Qwen3-30B-A3B (128 experts)   | TransformersMoEForCausalLM        | EP=2 (tp=2, ep=2)              |
| 11 | MoE LLM Tests                | Qwen3-30B-A3B (128 experts)   | TransformersMoEForCausalLM        | TP=4                           |
| 12 | MoE LLM Tests                | Qwen3-30B-A3B (128 experts)   | TransformersMoEForCausalLM        | EP=4 (tp=4, ep=4)              |
| 13 | VLM Tests                    | Qwen2.5-VL-7B-Instruct        | TransformersMultiModalForCausalLM | Single GPU, text+image         |
| 14 | VLM Tests                    | Qwen3-VL-2B                   | TransformersMultiModalForCausalLM | TP=1, text                     |
| 15 | VLM Tests                    | Qwen3-VL-2B                   | TransformersMultiModalForCausalLM | TP=1+compile, text             |
| 16 | VLM Tests                    | Gemma3-4B-IT                  | TransformersForCausalLM           | Single GPU, text               |
| 17 | VLM Tests                    | Gemma3-4B-IT                  | TransformersMultiModalForCausalLM | Single GPU, text               |
| 18 | VLM Tests                    | Gemma3-4B-IT                  | TransformersMultiModalForCausalLM | TP=4                           |
| 19 | VLM Tests                    | Gemma3-4B-IT                  | TransformersMultiModalForCausalLM | TP=4+compile                   |
| 20 | VLM Tests                    | Gemma3-4B-IT                  | TransformersMultiModalForCausalLM | text+image                     |
| 21 | VLM Tests                    | Gemma3-4B-IT                  | TransformersMultiModalForCausalLM | Single GPU, image              |
| 22 | Pipeline Parallel Tests      | Qwen2.5-7B-Instruct           | TransformersForCausalLM           | PP=2                           |
| 23 | Pipeline Parallel Tests      | Qwen2.5-7B-Instruct           | TransformersForCausalLM           | PP=4                           |
| 24 | Data Parallel Tests          | Qwen2.5-7B-Instruct           | TransformersForCausalLM           | DP=2                           |
| 25 | Data Parallel Tests          | Qwen2.5-7B-Instruct           | TransformersForCausalLM           | DP=4                           |

## Checklist

## Review Process


python/sglang/srt/model_executor/model_runner.py
```diff
@@ -1654,6 +1654,13 @@ def can_run_piecewise_cuda_graph(self):
                 "Disable piecewise CUDA graph because piecewise_cuda_graph has conflict with torch compile",
             )
             return False
+        if hasattr(self.model, "_can_torch_compile"):
```
JustinTong0323: It seems you are assuming _can_torch_compile is only for `transformers backend`?

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -1905,6 +1912,14 @@ def _dummy_run(self, batch_size: int, run_ctx=None):
 
         if self.server_args.enable_torch_compile:
             set_torch_compile_config()
+            if hasattr(self.model, "_can_torch_compile"):
```
JustinTong0323: Also here, we'd better distinguish the semantic of "_can_torch_compile" and "transformers backend"

python/sglang/srt/model_executor/model_runner.py
adarshxs: yeah good catch. fixing

python/sglang/srt/multimodal/processors/transformers_auto.py
```diff
@@ -0,0 +1,215 @@
...
+            kwargs["videos"] = videos
+        return self._processor(text=text, return_tensors="pt", **kwargs)
+
+    def _build_mm_items(
+        self, processor_output: dict, input_ids: torch.Tensor
+    ) -> list[MultimodalDataItem]:
+        """Extract MultimodalDataItem objects from the HF processor output."""
+        items = self.collect_mm_items_from_processor_output(processor_output)
+
+        mm_token_id = self.mm_tokens.image_token_id
```
JustinTong0323: This offset calculation is problematic when dealing with mixed modalities, such as combined image and audio.

adarshxs: fixed

yuan-luo:
Could you specify the usage commands for these new features and paste the accuracy result?

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -1891,6 +1891,14 @@ def _dummy_run(self, batch_size: int, run_ctx=None):
 
         if self.server_args.enable_torch_compile:
             set_torch_compile_config()
+            if self.server_args.model_impl.lower() == ModelImpl.TRANSFORMERS:
```
yuan-luo: There are too many layer nested if. Could we refactor here?

python/sglang/srt/model_loader/utils.py
```diff
@@ -42,32 +112,58 @@ def resolve_transformers_arch(model_config: ModelConfig, architectures: list[str
...
+                "Skipping remote model compatibility checks.",
+                arch,
+                e,
             )
-            for name, module in sorted(auto_map.items(), key=lambda x: x[0])
-        }
         model_module = getattr(transformers, arch, None)
         if model_module is None:
-            if "AutoModel" not in auto_map:
+            if "AutoModel" not in auto_modules:
```
yuan-luo: Likewise. Please refactor the nested if.

python/sglang/srt/model_executor/model_runner.py
adarshxs: will do

adarshxs: done

adarshxs:
adding

yuan-luo (approved):

yuan-luo:
image
It seems the main has some regression, could you rebase the latest main and run CI again?

test/registered/models/test_transformers_backend_eval.py
```diff
@@ -0,0 +1,36 @@
...
+
+register_cuda_ci(est_time=180, suite="stage-b-test-small-1-gpu")
+
+
+class TestTransformersBackendEval(DefaultServerBase):
+    model = "HuggingFaceTB/SmolLM3-3B"
+    gsm8k_num_questions = 30
+    gsm8k_accuracy_thres = 0.5
+    gsm8k_parallel = 30
+    other_args = ["--model-impl", "transformers"]
```
yuan-luo: Please add at least one advanced launch server parameter.

python/sglang/srt/models/utils.py
```diff
@@ -104,6 +113,161 @@ def apply_dict(self, values: dict[str, Any]) -> dict[str, Any]:
         }
 
 
+class AutoWeightsLoader:
+    ROTARY_EMBEDS_UNUSED_WEIGHTS = [
+        "rotary_pos_emb.inv_freq",
```
JustinTong0323, line 120: **`[Critical]` `itertools.groupby` requires sorted input but this doesn't guarantee it.**

`itertools.groupby` only groups *consecutive* elements with the same key. If weights arrive in non-sorted order (common with sharded safetensors files), the same prefix produces multiple separate groups. For modules with custom `load_weights` (like `TransformersFusedMoE`), this means the method gets called multiple times with partial weight subsets, potentially causing incorrect loading.

**Suggested fix:** Sort by prefix before grouping, or use `defaultdict(list)` to collect all weights for the same prefix:
```python
def _groupby_prefix(self, weights):
    weights_by_parts = sorted(
        ((name.split('.', 1), data) for name, data in weights),
        key=lambda x: x[0][0],
    )
    for prefix, group in itertools.groupby(weights_by_parts, key=lambda x: x[0][0]):
        yield prefix, (
            ('' if len(parts) == 1 else parts[1], data)
            for parts, data in group
        )
```
Note: sorting forces materialization into memory — `defaultdict(list)` is an alternative.

python/sglang/srt/models/transformers.py
```diff
@@ -152,138 +514,1081 @@ def __init__(
...
+                        )
+                    except Exception as e:
+                        logger.warning(
+                            "Failed to load dynamic module from auto_map[%s]: %s.",
+                            key,
+                            e,
+                        )
+                    break
+
+        supports_backend = (
```
JustinTong0323, line 600: **`[Critical]` `break` is outside the `try` block — fallback key never attempted on failure.**

If loading from `"AutoModel"` fails, the `except` catches the error and logs a warning, but then the `break` executes unconditionally, skipping `"AutoModelForCausalLM"`. The `break` should be inside the `try` block:

```python
for key in ("AutoModel", "AutoModelForCausalLM"):
    if key in auto_map:
        try:
            model_cls = get_class_from_dynamic_module(
                auto_map[key],
                getattr(config, "_name_or_path", ""),
            )
            break  # Only break on success
        except Exception as e:
            logger.warning(
                "Failed to load dynamic module from auto_map[%s]: %s.",
                key, e,
            )
```

python/sglang/srt/models/transformers.py
```diff
@@ -83,63 +272,236 @@ def sglang_flash_attention_forward(
...
-        input_size=linear.in_features,
-        output_size=linear.out_features,
-        bias=linear.bias is not None,
-        quant_config=quant_config,
+        recorder = get_global_expert_distribution_recorder()
+        if recorder is not None and recorder.recording:
+            recorder.record(self.experts.layer_id, topk_ids)
+    except Exception:
+        pass
+    topk_output = StandardTopKOutput(
```
JustinTong0323, line 482: **`[Critical]` Bare `except Exception: pass` silently swallows all errors.**

This catches every non-system exception (RuntimeError, ValueError, MemoryError, etc.) during EPLB recording and discards them. If expert distribution recording silently fails, EPLB will make scheduling decisions based on stale/missing data with no diagnostic logs.

**Suggested fix:** Narrow to `except ImportError` (the only expected failure) and let the rest propagate:

```python
try:
    from sglang.srt.eplb.expert_distribution import (
        get_global_expert_distribution_recorder,
    )
except ImportError:
    pass
else:
    recorder = get_global_expert_distribution_recorder()
    if recorder is not None and recorder.recording:
        recorder.record(self.experts.layer_id, topk_ids)
```

python/sglang/srt/models/transformers.py
```diff
@@ -152,138 +514,1081 @@ def __init__(
...
+
+                    base_cfg = AutoConfig.for_model(base_type)
+                    base_tp = getattr(base_cfg, "base_model_tp_plan", None)
+                    if base_tp:
+                        plan.update(
+                            {f"{child_name}.{k}": v for k, v in base_tp.items()}
+                        )
+                except Exception:
+                    pass
+        return plan
```
JustinTong0323, line 736: **`[Critical]` Another bare `except Exception: pass` — TP plan inference failure is completely hidden.**

If `AutoConfig.for_model(base_type)` raises any error (network error, invalid config, corrupted cache), it's silently swallowed. The function returns an empty plan, causing confusing downstream errors like "does not support tensor parallel yet!" with no indication that the root cause was a failed config lookup.

```python
try:
    from transformers import AutoConfig
    base_cfg = AutoConfig.for_model(base_type)
    base_tp = getattr(base_cfg, "base_model_tp_plan", None)
    if base_tp:
        plan.update(...)
except (ImportError, ValueError, KeyError) as e:
    logger.debug(
        "Could not infer TP plan from base model type '%s': %s", base_type, e
    )
```

python/sglang/srt/managers/multimodal_processor.py
```diff
@@ -43,12 +43,27 @@ def import_processors(package_name: str, overwrite: bool = False):
...
 
+    if model_impl in {"auto", "transformers"}:
+        from sglang.srt.multimodal.processors.transformers_auto import (
+            TransformersAutoMultimodalProcessor,
+        )
+
+        return TransformersAutoMultimodalProcessor(
+            hf_config, server_args, processor, transport_mode, **kwargs
+        )
+
```
JustinTong0323, line 79: **`[High]` Fallback to `TransformersAutoMultimodalProcessor` may be incorrect when `model_impl='auto'`.**

When `model_impl` is `"auto"`, the model may have resolved to the native sglang backend (not transformers). In that case, falling back to `TransformersAutoMultimodalProcessor` instead of raising the previous `ValueError` could produce incorrect multimodal input formatting.

**Suggested fix:** Only fall back when the resolved backend is actually transformers:
```python
if model_impl == "transformers" or (
    model_impl == "auto" and model_config.model_impl == ModelImpl.TRANSFORMERS
):
```

JustinTong0323 (commented):
## PR Review Summary

Reviewed with specialized agents focusing on: code quality, silent failures, test coverage, and code simplification. Overall this is a well-structured feature addition with a clean mixin-based architecture — but there are several issues worth addressing before merge.

**Critical: 3** | **High: 5** | **Simplification suggestions: 5**

See inline comments for details on each issue.

### Test Coverage

The PR adds ~2074 lines of implementation but only 43 lines of tests (a single GSM8K eval). Key untested areas include:
- `_is_moe_model` / `_get_transformers_backend_arch` routing logic (pure functions, trivially unit-testable)
- `AutoWeightsLoader` weight dispatch
- `TransformersFusedMoE` expert replacement
- `MultiModalMixin` + `TransformersAutoMultimodalProcessor`
- Pipeline parallel layer slicing
- Embedding/classification model classes

### Simplification Opportunities
1. **Consolidate duplicated helpers:** `_first_attr` (in `transformers_auto.py`) and `_getattr_first` (in `transformers.py`) are identical. Same for `_uses_mrope` / `_uses_mrope_positions`.
2. **Extract tensor-list flattening helper** from `_to_tensor_output` — same logic appears twice in the same method.
3. **Simplify `replace_rms_norm_class` Gemma detection** — instantiating a dummy module to check zero weights is fragile; checking class name would be simpler.
4. **Split `TransformersBase.__init__`** (~100+ lines) into focused setup methods for readability.

### Strengths
- Clean mixin-based architecture composing TP, PP, MoE, VLM capabilities
- Elegant `WeightsMapper` composition via `__or__` and `__init_subclass__`
- Comprehensive manual testing across 25 configurations
- Proper `hidden_states.clone()` in MoE custom op to prevent mutation

python/sglang/srt/models/transformers.py
```diff
@@ -83,63 +272,236 @@ def sglang_flash_attention_forward(
...
+                    param = param_dict[direct_name]
+                    weight_loader = getattr(
+                        param, "weight_loader", default_weight_loader
+                    )
+                    try:
+                        weight_loader(param, loaded_weight)
+                    except TypeError:
+                        default_weight_loader(param, loaded_weight)
+                    loaded.add(name)
+        return loaded
```
JustinTong0323, line 465: **`[High]` Unmatched MoE weights are silently dropped — no log, no warning.**

When a weight name from the checkpoint matches neither the expert mapping nor a direct parameter name, it's silently dropped. Corrupted checkpoints or renamed parameters in new transformers versions would cause the model to run with uninitialized (random) weights for some experts, producing garbage output with no diagnostic information.

**Suggested fix:** Add a warning for unmatched weights:
```python
if not matched:
    direct_name = name if name in param_dict else f"experts.{name}"
    if direct_name in param_dict:
        ...
    else:
        logger.warning(
            "MoE weight '%s' (layer %s) could not be matched to any parameter. "
            "This weight will NOT be loaded.", name, self.layer_name,
        )
```

python/sglang/srt/models/transformers.py
```diff
@@ -152,138 +514,1081 @@ def __init__(
...
+            )
+            feature = self._cast_mm_value(
+                item.feature,
+                dtype=target_dtype,
+                device=target_device,
+            )
+            try:
+                result = encoder(feature, **kwargs)
+            except TypeError:
+                kwargs[feature_kwarg] = feature
```
JustinTong0323, line 1455: **`[High]` `except TypeError` used as control flow to probe encoder signature — hides real bugs.**

`TypeError` can also be raised by bugs *inside* the encoder function (wrong tensor types, None values, incompatible dtypes). If the first call fails due to an internal encoder bug, the second call will likely produce a different, more confusing error that obscures the root cause.

**Suggested fix:** Use `inspect.signature` at initialization time instead:
```python
sig = inspect.signature(encoder)
if feature_kwarg in sig.parameters or any(
    p.kind == inspect.Parameter.VAR_KEYWORD for p in sig.parameters.values()
):
    kwargs[feature_kwarg] = feature
    result = encoder(**kwargs)
else:
    result = encoder(feature, **kwargs)
```

python/sglang/srt/model_loader/utils.py
```diff
@@ -27,9 +27,79 @@ def set_default_torch_dtype(dtype: torch.dtype):
...
+    )
+
+
+def _get_transformers_backend_arch(
+    model_config: ModelConfig, architectures: list[str]
+) -> str:
+    is_pooling = not model_config.is_generation
+    is_multimodal = model_config.is_multimodal or (
+        model_config.hf_config is not model_config.hf_text_config
+    )
```
JustinTong0323, line 80: **`[Medium]` `_is_moe_model` has an overly aggressive catch-all that can false-positive.**

After checking for `bool`, `int/float`, `list/tuple/set/dict`, and `str`, the final `if value is not None: return True` catch-all means any non-None value of an unexpected type (custom object, enum, callable) triggers MoE detection. This could falsely classify a non-MoE model as MoE if a config attribute happens to share a name with one of the expert attributes.

Consider removing the catch-all or restricting it to expected types only.

PR #14726, opened 2025-12-09
Use bootstrap server to sync prefill dp rank

changhuaixin:
## Motivation

As described in Feature #10174 and #13052, use bootstrap server to store prefill dp rank, thus allow prefill server to use load balance methods.

## Modifications

The bootstrap process is modified as follows:
image

The cases when DPC assigns prefill dp rank is handled now, enabling us to support load balance methods other than follow-bootstrap-room.

Prefill server always register its dp rank to bootstrap server now. Decode server checks where data_parallel_rank is assigned in GenerateReqInput.

## Accuracy Tests

None

## Evaluating communication overhead

It costs less than 2ms on average to put and get bootstrap info from bootstrap server.
```
[2025-12-31 18:01:40 DP2 TP2 EP2] Prefill DP rank registration stats: 0.000962s
[2025-12-31 18:01:40 DP1 TP1 EP1] Prefill DP rank registration stats: 0.001020s
[2025-12-31 18:01:40 DP1 TP1 EP1] Prefill DP rank registration stats: 0.001160s
[2025-12-31 18:01:42 DP3 TP3 EP3] Prefill DP rank registration stats: 0.001908s
[2025-12-31 18:01:42 DP0 TP0 EP0] Prefill DP rank registration stats: 0.001397s
[2025-12-31 18:01:42 DP2 TP2 EP2] Prefill DP rank registration stats: 0.001447s
[2025-12-31 18:01:44 DP3 TP3 EP3] Prefill DP rank registration stats: 0.003155s

```

## Benchmarking and Profiling

I have used Qwen3-235B to verify its effectiveness. To reduce TTFT is our primary goal and ISL/OSL is set to 3500/50. In this case the maximum throughput for prefill server is around 3.2. Thus request rate with 1/2/2.6/3.2/4/inf is evaluated.

| request rate |  original mean TTFT |  optimized mean TTFT | optimized ratio |
|----------|----------|----------|----------|
| 1   |  1860.30   | 1432.17   | 23.01%   |
| 2   | 3896.73   | 2669.39   | 31.49%   |
| 2.6   | 5976.88   | 3854.59   | 35.50%   |
| 3.2   | 10446.72   | 4816.37   | 53.89%  |
| 4   |  15599.93   | 12293.58   | 21.19%   |
| inf  |  37264.81   | 37184.89   | 0.21%   |

Also throughput has increased from 3.09 to 3.13 by 2%. In other cases, it might be higher.

Without this patch and request rate 3.2
```
[...]

PR #19168, opened 2026-02-23
[PD-Disagg] Support query dp rank from bootstrap server.

hnyls2002:
## Motivation

Based on [PR #14726](https://github.com/sgl-project/sglang/pull/14726) by @changhuaixin. Allows prefill servers to use load balance methods other than `follow_bootstrap_room` (e.g., `round_robin`) in PD-disaggregation mode, which significantly reduces TTFT under load (see benchmarks in #14726).

When prefill uses non-deterministic load balancing, the decode server cannot infer which prefill DP rank processed a request. This PR adds per-request dp_rank synchronization through the bootstrap server.

## Modifications

- **Bootstrap server**: New `POST /register_dp_rank` and `POST /query_dp_ranks` endpoints with TTL-based cleanup. Parallel info now includes `follow_bootstrap_room` flag so decode knows the prefill's LB strategy.
- **Prefill sender**: Registers dp_rank per request when not using `follow_bootstrap_room`.
- **Decode side**: Resolves dp_rank before creating the KV receiver. Requests whose dp_rank is not yet available are held in a `pending_reqs` queue and batch-resolved each scheduler loop.
- **Connection-level init**: Extracted `ensure_parallel_info()` on `CommonKVManager` to decouple parallel info fetching from receiver creation, shared by both `DecodePreallocQueue` and `CommonKVReceiver`.
- **Naming cleanup**: Removed redundant `target_dp_group` alias; unified to `prefill_dp_rank`.

## Design difference from #14726

dp_rank is resolved **before** receiver creation, not inside `init()`. This means:
- No half-initialized receiver state or `should_notify_dp_rank` flag
- `init()` signature unchanged — **zero changes to mooncake/nixl/fake backends**
- `pop_preallocated` loop has no dp_rank awareness
- When using `follow_bootstrap_room`, dp_rank is computed locally (`bootstrap_room % dp_size`) with no HTTP overhead — the original PR always queries the bootstrap server


python/sglang/srt/disaggregation/common/conn.py
```diff
@@ -518,9 +559,14 @@ def __init__(self, host: str, port: int):
         self.attn_tp_size = None
         self.dp_size = None
```
ShangmingCai: We have this line, which seems like the `dp_size` in
`def __init__(self, host: str, port: int, dp_size: int = 1):` is never used.
Is this left for future refactoration?

python/sglang/srt/managers/disagg_service.py
```diff
@@ -28,6 +28,7 @@ def start_disagg_service(
         bootstrap_server: BaseKVBootstrapServer = kv_bootstrap_server_class(
             host=server_args.host,
             port=server_args.disaggregation_bootstrap_port,
+            dp_size=server_args.dp_size,
```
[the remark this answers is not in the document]
ShangmingCai: KVBootstrapServer does not need dp size info.

We get this in the init of kv manager (per rank) by:
```python
register_to_bootstrap
```

python/sglang/srt/disaggregation/common/conn.py
```diff
@@ -93,12 +96,41 @@ def __init__(
...
+                f"Page size mismatch: prefill server has page_size={page_size}, "
+                f"but decode server has page_size={self.kv_args.page_size}. "
+                f"Both servers must use the same --page-size value."
+            )
+
+        self.prefill_attn_tp_size_table[bootstrap_addr] = tp_size
+        self.prefill_dp_size_table[bootstrap_addr] = dp_size
+        self.prefill_pp_size_table[bootstrap_addr] = pp_size
+        self.prefill_page_size_table[bootstrap_addr] = page_size
+        self.follow_bootstrap_room_table[bootstrap_addr] = follow_bootstrap_room
```
ShangmingCai, line 127: Might need to del this in `_handle_node_failure` as well

python/sglang/srt/disaggregation/common/conn.py
hnyls2002: Typo, fixed.

python/sglang/srt/disaggregation/common/conn.py
hnyls2002: done

ShangmingCai (approved):
Image

CI looks good. Failed amd tests are irrelevant.

PR #19169, opened 2026-02-23
[Qwen3.5] Raise Exception when radix_cache and extra_buffer are enabled at the same time

YAMY1234:
## Motivation

Two issues with mamba cache handling when `disable_radix_cache=True`:

1. **Invalid extra_buffer usage**: `extra_buffer` (ping-pong track buffer for radix cache overlap scheduling) serves no purpose when radix cache is disabled and speculative decoding is not in use — `no_buffer` already supports overlap scheduling in this case. Allowing this wastes 2 mamba slots per request and will cause OOM at high concurrency.

2. **Ratio calculation bug**: When `extra_buffer` is used with `disable_radix_cache=True` (possible with speculative decoding), each request allocates track buffer slot(s) from the mamba pool. But the ratio was hardcoded to `1`, not accounting for the extra track buffer slots (`additional_ratio`), which could lead to mamba pool overflow.

## Modifications

- `server_args.py`: Raise `ValueError` when `extra_buffer` + `disable_radix_cache` + no speculative decoding, guiding users to use `--mamba-scheduler-strategy no_buffer` instead.
- `model_runner_kv_cache_mixin.py`: Fix ratio from `1` to `1 + additional_ratio` in the `disable_radix_cache` branch, consistent with the `else` branch. This correctly accounts for the track buffer slots allocated by `extra_buffer`.

## Benchmarking and Profiling

Verified with Qwen3.5 on PD disagg (tp4/tep4/dep4) with concurrency 8x32x128x256x512x1024.
Before the fix (crash after concurrency>=128)
```
Maximum request concurrency: 128
[Sending] All 384 requests sent. Waiting for completions...
slurmstepd: error: *** STEP 998582.10 ON lyris0127 CANCELLED AT 2026-02-22T04:10:02 DUE TO TIME LIMIT ***


  File "/sgl-workspace/sglang/python/sglang/srt/mem_cache/memory_pool.py", line 515, in alloc
    mid is not None
AssertionError: Not enough space for mamba cache, try to increase --mamba-full-memory-ratio or --max-mamba-cache-size. mid=None, self.mamba_pool.size=60, self.mamba_pool.available_size()=0, len(reqs)=1
```
After the fix (remains stable across different concurrencies)
```
+ set +x
2026-02-22 15:10:21
Completed benchmark with concurrency: 1024
-----------------------------------------
SA-Bench complete. Results in /logs/sa-bench_isl_1024_osl_1024
```

## Checklist

## Review Process

### DIFF
--- a/python/sglang/srt/model_executor/model_runner_kv_cache_mixin.py
+++ b/python/sglang/srt/model_executor/model_runner_kv_cache_mixin.py
@@ -360,7 +360,7 @@ class ModelRunnerKVCacheMixin:
                 else:
                     additional_ratio = MAMBA_CACHE_V2_ADDITIONAL_RATIO_OVERLAP
             if self.server_args.disable_radix_cache:
-                ratio = 1 + additional_ratio
+                ratio = 1
             else:
                 ratio = MAMBA_CACHE_SIZE_MAX_RUNNING_REQUESTS_RATIO + additional_ratio
             max_num_reqs = min(
--- a/python/sglang/srt/model_executor/model_runner_kv_cache_mixin.py
+++ b/python/sglang/srt/model_executor/model_runner_kv_cache_mixin.py
@@ -360,7 +360,7 @@ class ModelRunnerKVCacheMixin:
                 else:
                     additional_ratio = MAMBA_CACHE_V2_ADDITIONAL_RATIO_OVERLAP
             if self.server_args.disable_radix_cache:
-                ratio = 1
+                ratio = 1 + additional_ratio
             else:
                 ratio = MAMBA_CACHE_SIZE_MAX_RUNNING_REQUESTS_RATIO + additional_ratio
             max_num_reqs = min(
--- a/python/sglang/srt/server_args.py
+++ b/python/sglang/srt/server_args.py
@@ -1726,6 +1726,15 @@ class ServerArgs:
             ), f"mamba extra_buffer is not supported for {model_arch} model"
 
         if self.enable_mamba_extra_buffer():  # extra_buffer
+            if self.disable_radix_cache:
+                logger.warning(
+                    "mamba extra_buffer is not needed when radix cache is disabled. "
+                    "Overlap scheduling is already supported with no_buffer + disable_radix_cache. "
+                    "Falling back to mamba_scheduler_strategy='no_buffer'."
+                )
+                self.mamba_scheduler_strategy = "no_buffer"
+                return
+
             assert (
                 is_cuda()
             ), "Mamba extra_buffer is only supported on CUDA devices with FLA backend"
--- a/python/sglang/srt/server_args.py
+++ b/python/sglang/srt/server_args.py
@@ -1726,10 +1726,9 @@ class ServerArgs:
             ), f"mamba extra_buffer is not supported for {model_arch} model"
 
         if self.enable_mamba_extra_buffer():  # extra_buffer
-            if self.disable_radix_cache and self.speculative_algorithm is None:
+            if self.disable_radix_cache:
                 raise ValueError(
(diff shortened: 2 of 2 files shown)


python/sglang/srt/model_executor/model_runner_kv_cache_mixin.py
```diff
@@ -360,7 +360,7 @@ def init_memory_pool(self: ModelRunner, total_gpu_memory: int):
                 else:
                     additional_ratio = MAMBA_CACHE_V2_ADDITIONAL_RATIO_OVERLAP
             if self.server_args.disable_radix_cache:
-                ratio = 1
+                ratio = 1 + additional_ratio
```
hlu1: I don't think this is necessary. Why do you even need to enable extra_buffer when radix cache is not on? The fix should be not allow extra_buffer=on and radix_cache=off at the same time.

YAMY1234: Emmm my understanding is extra_buffer is for mamba overlap scheduling (ping-pong track buffer), which is different from the buffer for radix_cache

YAMY1234: Oh I get what you mean. When disable_radix_cache, no_buffer can also support overlap schedule

hlu1: Can you change the PR to raise an exception when we have extra_buffer=on and radix_cache=off at the same time? Another option is to silently turn off extra_buffer. There are pros and cons with either approach.

YAMY1234: Makes sense. https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/server_args.py#L1747-L1752
Just double checked when disable_radix_cache=True + no_buffer, the code does not enter the elif not self.disable_radix_cache so self.disable_overlap_schedule = True is never executed — overlap schedule remains enabled by default. 

I think we can raise an exception for this case and make some obvious suggestions to turn of extra_buffer when radix_cache is not enabled. I will change this PR.

YAMY1234: @hlu1 I just thought through,

I think the ratio fix (`1` → `1 + additional_ratio`) is still necessary — with speculative decoding (e.g. MTP) + `extra_buffer`, each request still allocates 1 track buffer slot from the mamba pool (for mamba state checkpoint tracking), but the original `ratio = 1` doesn't account for it, which could cause mamba pool overflow.

But the guard you suggested is also needed — without spec decoding, `extra_buffer` serves no purpose when radix cache is off, since `no_buffer` already supports overlap scheduling. So we should reject that invalid combination upfront.

YAMY1234: Updated with both changes:
1. `server_args.py`: Raise `ValueError` when `extra_buffer` + `disable_radix_cache` + no spec decoding, guiding users to `--mamba-scheduler-strategy no_buffer`.
2. `model_runner_kv_cache_mixin.py`: Fix `ratio = 1` → `ratio = 1 + additional_ratio` to correctly size the mamba pool when `extra_buffer` track buffers are in use.

YAMY1234: Removed the additional_ratio and raise this issue upstream.

MTP itself does not require an extra buffer. The additional ratio (MAMBA_CACHE_V2_ADDITIONAL_RATIO_OVERLAP) is only needed when radix cache is enabled. If radix cache is disabled, this extra buffer is still unnecessary, so it also makes sense to remove additional_ratio in this case.

hlu1 (approved):

ispobock (approved):

PR #19171, opened 2026-02-23
[Session] Add `streaming` mode with `SessionAwareCache` fast path

aurickq:
- Add `streaming` option to `open_session` and plumb it through engine/session creation.
- Introduce `SessionAwareCache` to keep KV ownership in per-session slots and bypass radix prefix matching on streaming turns after the first request.
- Keep only the first streaming turn's prompt cacheable in radix and skip prompt/output cache insertion for later turns to avoid radix update overhead in latency-sensitive flows.
- Restrict streaming sessions to append-only behavior (`replace` / `drop_previous_output` / non-zero `offset` are rejected).
- Add session timeout support and periodic reap logic in scheduler (`timeout` on session + `reap_timed_out_sessions`).
- Extend memory/runtime checks to account for session-held KV and req slots outside active batches.
- Add tests for streaming-session correctness and latency behavior.

- Follow-up PRs will address the open concerns raised in review, including: (1) compatibility guard/assert for `streaming` with speculative decoding under KV over-allocation, and (2) remaining cleanup/perf/maintainability refinements around the session cache wrapper path.


python/sglang/srt/mem_cache/swa_radix_cache.py
```diff
@@ -476,6 +476,10 @@ def cache_finished_req(self, req: Req, is_insert: bool = True) -> None:
             # In EAGLE chunked prefill case, the prefix_indices included one unmatched token (kv_indices[actual_kv_len:])
             # Here we -1 to make sure the kv of the unmatched token can be freed correctly to avoid memory leak
             old_prefix_len -= 1
+        if req.session and req.session.streaming:
+            # For streaming sessions, only the prompt of the first request is cached, so we need to use
+            # cache_protected_len instead of prefix_indices length.
+            old_prefix_len = req.cache_protected_len
```
aurickq: @ispobock not sure why it doesn't use `old_prefix_len = req.cache_protected_len` in all cases? seems simpler but not sure if i am missing some corner case

python/sglang/srt/managers/scheduler_runtime_checker_mixin.py
```diff
@@ -315,6 +315,14 @@ def check_tree_cache(self: Scheduler):
...
 
     def self_check_during_idle(self: Scheduler):
+        if any(s.streaming for s in self.sessions.values()):
+            # Disable if there are any streaming sessions open. This is because:
+            # - We want to avoid blocking new streaming session requests (which
+            #   are latency-sensitive) on the idle memory check.
+            # - In between streaming session requests, we hold onto kv memory but
+            #   the request is not part of the scheduler's running batch. The
+            #   memory checker doesn't currently handle this case.
+            return
```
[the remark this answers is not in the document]
aurickq: Hmm what about the other cases where it returns for disaggregation?

python/sglang/srt/mem_cache/swa_radix_cache.py
ispobock: Actually we should use cache_protected_len. Just this change https://github.com/sgl-project/sglang/pull/13714 wasn't applied to swa radix cache.

aurickq: I think I am not familiar enough with this code path to do it myself :) so I will leave it to someone else

python/sglang/srt/managers/scheduler_runtime_checker_mixin.py
```diff
@@ -315,6 +315,14 @@ def check_tree_cache(self: Scheduler):
             self.tree_cache.sanity_check()
 
     def self_check_during_idle(self: Scheduler):
+        if any(s.streaming for s in self.sessions.values()):
```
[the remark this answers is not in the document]
aurickq: should be fine

Qiaolin-Yu (approved):

hnyls2002:
@aurickq 
Added a latency test in CI to prevent performance regression:

```bash
  SUMMARY  (4 sessions x 300 turns)
Mode                 Avg (all)    Avg (last 10)    Speedup (all)    Speedup (last 10)
-----------------  -----------  ---------------  ---------------  -------------------
no_session              98.5ms          153.7ms            1.00x                1.00x
regular_session         97.4ms          150.7ms            1.01x                1.02x
streaming_session       70.4ms           72.9ms            1.40x                2.11x
```

python/sglang/srt/mem_cache/common.py
```diff
@@ -478,6 +478,13 @@ def release_kv_cache(req: Req, tree_cache: BasePrefixCache, is_insert: bool = Tr
 
     tree_cache.cache_finished_req(req, is_insert=is_insert)
 
+    # FIXME: SessionAwareCache.cache_finished_req sets req_pool_idx = None to
+    # transfer KV ownership to the SessionSlot, so we skip the remaining
+    # cleanup (overalloc free + pool slot free). This means over-allocated
+    # tokens from speculative decoding are NOT freed between turns.
+    if req.req_pool_idx is None:
```
ispobock, line 485: May need a test for spec case?

ispobock (approved):

python/sglang/srt/mem_cache/common.py
hnyls2002: Spec is not supported for this streaming API. Currently, we do not support trimming the trailing over-allocated tokens.

PR #19177, opened 2026-02-23
Fix: Cuda Graph + HiCache + Speculative Decoding Working Together were giving Cuda Illegal memory access error.  

PratikNarola1:
Change 1: HiCache Loading Guard for CUDA Graphs

  Files: scheduler.py, schedule_batch.py, forward_batch_info.py, cache_controller.py, hiradix_cache.py

  What: Propagated hicache_consumer_index through the full batch pipeline (ScheduleBatch → ModelWorkerBatch → ForwardBatch) and added checks in all CUDA graph runners to disable graphs when consumer_index >= 0.
  Also added is_loading_in_progress() to catch async loads still running from a previous batch.

  Why: When HiCache loads KV cache from host memory, it allocates new device pages and copies data asynchronously on a separate CUDA stream. During this window, page indices in req_to_token are being updated. If a
  CUDA graph replays and reads half-updated page tables → illegal memory access.

  Why wasn't it done already: The upstream code literally had a TODO comment: # todo (zhiqiang): disable cuda graph execution if hicache loading triggered. The original hicache_consumer_index field existed but was only set during prefill batches, not decode batches. We extended it to decode batches AND added the is_loading_in_progress() check.

  Why other models don't need this: Models without --enable-hierarchical-cache always have consumer_index = -1, so this check is a no-op.

  ---
  Change 2: Memory Relocation Version Tracking

  Files: cache_controller.py, base_prefix_cache.py, radix_cache.py, hiradix_cache.py, allocator.py

  What: Added memory_relocation_version counter to HiCacheController. Incremented only when HiCache actually relocates memory (host→device load, or device eviction). Separate from kv_free_version which increments on every normal free.

  Why: We needed to distinguish "normal KV cache free/alloc" (happens every decode step, safe for CUDA graphs) from "HiCache memory relocation" (changes logical page index mapping). This distinction is what allows
  the last_loc recomputation in eagle_info.py to know when to refresh stale indices.

  Why wasn't it done already: sglang had no version tracking for HiCache relocations at all. The EAGLE code path assumed page indices are stable between operations — which is true without HiCache. HiCache + EAGLE is a very new combination.

  Why other models don't need this: RadixCache (non-hierarchical) returns get_memory_relocation_version() = 0 always. Only HiRadixCache has a real counter.

  ---
  Change 3: last_loc Bounds Clamping + Version-Aware Recomputation

  File: eagle_info.py

  What: In prepare_for_verify():
  1. last_loc.clamp(0, total_slots) — prevents out-of-bounds access, no GPU sync
  2. Check memory_relocation_version before and after alloc_paged_token_slots_extend(). If version changed (HiCache relocated during allocation), recompute last_loc from fresh req_to_token data.

  Why: last_loc reads req_to_token[req_pool_indices, prefix_lens-1] — the last KV cache slot per request. If HiCache evicts/relocates pages between computing last_loc and using it, the value could point to a freed slot. The allocation itself can trigger eviction (need more device pages → evict old ones).

  Why wasn't it done already: Without HiCache, page indices in req_to_token are stable between operations. HiCache introduces asynchronous memory relocation that can invalidate indices mid-operation. This race condition is specific to the alloc → use gap in EAGLE's verify path.

  ---
  Change 4: Draft Attention Backend → FlashInfer

  File: glm5.sh (config only: --speculative-draft-attention-backend flashinfer)

  What: Use FlashInfer instead of NSA for the EAGLE draft model's attention. Target model still uses NSA.

  Why: GLM-5's NSA backend creates NativeSparseAttnMultiStepBackend (2 NSA sub-backends for 3-step draft). The CUDA graph replay path for this multi-step NSA configuration has a latent bug causing illegal memory access. This bug was always present but masked because CUDA graphs were always disabled — the original code used a kv_free_version check that incremented on every free(), so can_run() always returned False. When
  we fixed CUDA graphs to actually run, we exposed this bug.

  Rather than debugging the NSA CUDA graph kernel (deep in the NSA attention implementation), we switch the draft model to FlashInferMLAMultiStepDraftBackend which:
  - Is automatically created for MLA models (GLM-5 uses MLA architecture)
  - Supports page_size=64 and topk=1 (both used by GLM-5)
  - Has a battle-tested CUDA graph implementation
  - Shares the same MLA KV cache format, so draft and target are compatible

  Why wasn't it done already: The --speculative-draft-attention-backend flag exists but is rarely used. Most deployments use the same backend for draft and target. The NSA + EAGLE CUDA graph path was effectively
  dead code — it existed in the codebase but was never reached because can_run() always returned False.

### DIFF
--- a/python/sglang/srt/managers/cache_controller.py
+++ b/python/sglang/srt/managers/cache_controller.py
@@ -96,6 +96,13 @@ class LayerDoneCounter:
         self.producer_index = -1
         self.consumer_index = -1
 
+    def is_loading_in_progress(self) -> bool:
+        """Check if any loading operation is currently in progress."""
+        if self.producer_index < 0:
+            return False
+        # Check if the last producer's finish event has NOT completed
+        return not self.events[self.producer_index].finish_event.query()
+
 
 class CacheOperation:
 
@@ -300,6 +307,10 @@ class HiCacheController:
         self.layer_num = self.mem_pool_device.layer_num
         self.layer_done_counter = LayerDoneCounter(self.layer_num)
         self.mem_pool_device.register_layer_transfer_counter(self.layer_done_counter)
+
+        # Track memory relocations for CUDA graph invalidation
+        # Incremented each time load() allocates new device memory
+        self.memory_relocation_version: int = 0
 
         if write_policy not in [
             "write_through",
@@ -715,6 +726,13 @@ class HiCacheController:
         device_indices = self.mem_pool_device_allocator.alloc(len(host_indices))
         if device_indices is None:
             return None
+
+        # Memory relocated from host to device - increment relocation version
+        self.memory_relocation_version += 1
+        logger.info(
+            f"[HiCache] Load: memory_relocation_version={self.memory_relocation_version}"
+        )
+
         self.load_queue.append(
             CacheOperation(host_indices, device_indices, node_id, priority)
         )
@@ -782,8 +800,25 @@ class HiCacheController:
         )
         return producer_id
 
+    def is_loading_in_progress(self) -> bool:
+        """Check if any hicache loading is currently in progress."""
+        return self.layer_done_counter.is_loading_in_progress()
+
+    def get_memory_relocation_version(self) -> int:
+        """Get current memory relocation version for CUDA graph invalidation.
+
+        Only incremented on actual HiCache memory relocations (load/evict),
+        NOT on every normal free/alloc cycle.
+        """
+        return self.memory_relocation_version
+
     def evict_device(self, device_indices: torch.Tensor) -> int:
         self.mem_pool_device_allocator.free(device_indices)
+        # Memory relocated from device to host - increment relocation version
+        self.memory_relocation_version += 1
+        logger.info(
+            f"[HiCache] Device eviction: memory_relocation_version={self.memory_relocation_version}"
+        )
         return len(device_indices)
 
     def evict_host(self, host_indices: torch.Tensor, backup_only: bool = True) -> int:
--- a/python/sglang/srt/managers/cache_controller.py
+++ b/python/sglang/srt/managers/cache_controller.py
@@ -729,9 +729,6 @@ class HiCacheController:
 
         # Memory relocated from host to device - increment relocation version
         self.memory_relocation_version += 1
-        logger.info(
-            f"[HiCache] Load: memory_relocation_version={self.memory_relocation_version}"
-        )
 
         self.load_queue.append(
             CacheOperation(host_indices, device_indices, node_id, priority)
@@ -816,9 +813,6 @@ class HiCacheController:
         self.mem_pool_device_allocator.free(device_indices)
         # Memory relocated from device to host - increment relocation version
         self.memory_relocation_version += 1
-        logger.info(
-            f"[HiCache] Device eviction: memory_relocation_version={self.memory_relocation_version}"
-        )
         return len(device_indices)
 
     def evict_host(self, host_indices: torch.Tensor, backup_only: bool = True) -> int:
--- a/python/sglang/srt/managers/schedule_batch.py
+++ b/python/sglang/srt/managers/schedule_batch.py
@@ -1435,6 +1435,8 @@ class ScheduleBatch(ScheduleBatchDisaggregationDecodeMixin):
 
     # hicache pointer for synchronizing data loading from CPU to GPU
     hicache_consumer_index: int = -1
+    # Memory relocation version - for invalidating CUDA graphs after HiCache relocates memory
+    hicache_memory_version: int = 0
 
     # Diffusion LLM
     dllm_config: Optional[DllmConfig] = None
@@ -2367,6 +2369,15 @@ class ScheduleBatch(ScheduleBatchDisaggregationDecodeMixin):
         if self.spec_info:
             self.spec_info.merge_batch(other.spec_info)
 
+        # Propagate HiCache fields - use max version (most conservative for CUDA graph invalidation)
+        self.hicache_memory_version = max(
+            self.hicache_memory_version, other.hicache_memory_version
+        )
+        if other.hicache_consumer_index >= 0:
+            self.hicache_consumer_index = max(
+                self.hicache_consumer_index, other.hicache_consumer_index
+            )
+
     def get_model_worker_batch(
         self, seq_lens_cpu_cache: Optional[torch.Tensor] = None
     ) -> ModelWorkerBatch:
@@ -2425,6 +2436,7 @@ class ScheduleBatch(ScheduleBatchDisaggregationDecodeMixin):
             spec_algorithm=self.spec_algorithm,
             spec_info=self.spec_info,
             hicache_consumer_index=self.hicache_consumer_index,
+            hicache_memory_version=self.hicache_memory_version,
             capture_hidden_mode=(
                 CaptureHiddenMode.FULL
                 if self.return_hidden_states
@@ -2618,6 +2630,7 @@ class ModelWorkerBatch:
     # If set, the output of the batch contains the hidden states of the run.
     capture_hidden_mode: CaptureHiddenMode = None
     hicache_consumer_index: int = -1
+    hicache_memory_version: int = 0
 
     # For matryoshka embeddings
     dimensions: Optional[list[int]] = None
--- a/python/sglang/srt/managers/scheduler.py
+++ b/python/sglang/srt/managers/scheduler.py
@@ -2578,11 +2578,21 @@ class Scheduler(
             self.spec_algorithm,
             chunked_req=self.chunked_req,
         )
+        # Always set memory relocation version for CUDA graph invalidation
+        # This tracks all KV cache free operations (request completions, evictions)
+        new_batch.hicache_memory_version = (
+            self.tree_cache.get_memory_relocation_version()
+        )
+
         self.max_prefill_bs = max(self.max_prefill_bs, len(can_run_list))
         if self.enable_hierarchical_cache:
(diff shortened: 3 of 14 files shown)


chenkaiyue:
> Yes @chenkaiyue I had faced same error as well. and yes it should fix it. Could you try out my branch or cherry pick commits and see if it fixes your errors?

@PratikNarola1 OK~ Did you perform stress testing and long-duration testing? I found that this error might only appear once every few hours.

chenkaiyue:
I cherry pick this pr to v0.5.9,  it happend again
`Search for `cudaErrorIllegalAddress' in https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html for more information.
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1
Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.

Exception raised from c10_cuda_check_implementation at /pytorch/c10/cuda/CUDAException.cpp:44 (most recent call first):
frame #0: c10::Error::Error(c10::SourceLocation, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) + 0x80 (0x7f5438d91b80 in /usr/local/lib/python3.12/dist-packages/torch/lib/libc10.so)
frame #1: <unknown function> + 0x11fb7 (0x7f54b3166fb7 in /usr/local/lib/python3.12/dist-packages/torch/lib/libc10_cuda.so)
frame #2: c10d::ProcessGroupNCCL::WorkNCCL::finishedGPUExecutionInternal() const + 0x50 (0x7f5439c92fe0 in /usr/local/lib/python3.12/dist-packages/torch/lib/libtorch_cuda.so)
frame #3: c10d::ProcessGroupNCCL::WorkNCCL::isCompleted() + 0x68 (0x7f5439ca26b8 in /usr/local/lib/python3.12/dist-packages/torch/lib/libtorch_cuda.so)
frame #4: c10d::ProcessGroupNCCL::Watchdog::runLoop() + 0x969 (0x7f5439ca68b9 in /usr/local/lib/python3.12/dist-packages/torch/lib/libtorch_cuda.so)
frame #5: c10d::ProcessGroupNCCL::Watchdog::run() + 0xdf (0x7f5439ca882f in /usr/local/lib/python3.12/dist-packages/torch/lib/libtorch_cuda.so)
frame #6: <unknown function> + 0xecdb4 (0x7f5632bb1db4 in /usr/lib/x86_64-linux-gnu/libstdc++.so.6)
frame #7: <unknown function> + 0x9caa4 (0x7f563575baa4 in /usr/lib/x86_64-linux-gnu/libc.so.6)
frame #8: <unknown function> + 0x129c6c (0x7f56357e8c6c in /usr/lib/x86_64-linux-gnu/libc.so.6)

terminate called after throwing an instance of 'c10::DistBackendError'
I0303 23:04:30.833293 792429 real_client.cpp:1807] Time taken for batch_get_into: 1702us, read store: 0us, with memory key count: 256, offload key count: 0
  what():  [PG ID 2 PG GUID 3 Rank 1] Process group watchdog thread terminated with exception: CUDA error: an illegal memory access was encountered
Search for `cudaErrorIllegalAddress' in https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html for more information.
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1
Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.

Exception raised from c10_cuda_check_implementation at /pytorch/c10/cuda/CUDAException.cpp:44 (most recent call first):
frame #0: c10::Error::Error(c10::SourceLocation, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) + 0x80 (0x7f5438d91b80 in /usr/local/lib/python3.12/dist-packages/torch/lib/libc10.so)
frame #1: <unknown function> + 0x11fb7 (0x7f54b3166fb7 in /usr/local/lib/python3.12/dist-packages/torch/lib/libc10_cuda.so)
frame #2: c10d::ProcessGroupNCCL::WorkNCCL::finishedGPUExecutionInternal() const + 0x50 (0x7f5439c92fe0 in /usr/local/lib/python3.12/dist-packages/torch/lib/libtorch_cuda.so)
frame #3: c10d::ProcessGroupNCCL::WorkNCCL::isCompleted() + 0x68 (0x7f5439ca26b8 in /usr/local/lib/python3.12/dist-packages/torch/lib/libtorch_cuda.so)
frame #4: c10d::ProcessGroupNCCL::Watchdog::runLoop() + 0x969 (0x7f5439ca68b9 in /usr/local/lib/python3.12/dist-packages/torch/lib/libtorch_cuda.so)
frame #5: c10d::ProcessGroupNCCL::Watchdog::run() + 0xdf (0x7f5439ca882f in /usr/local/lib/python3.12/dist-packages/torch/lib/libtorch_cuda.so)
frame #6: <unknown function> + 0xecdb4 (0x7f5632bb1db4 in /usr/lib/x86_64-linux-gnu/libstdc++.so.6)
frame #7: <unknown function> + 0x9caa4 (0x7f563575baa4 in /usr/lib/x86_64-linux-gnu/libc.so.6)
frame #8: <unknown function> + 0x129c6c (0x7f56357e8c6c in /usr/lib/x86_64-linux-gnu/libc.so.6)

Exception raised from run at /pytorch/torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp:2063 (most recent call first):
frame #0: c10::Error::Error(c10::SourceLocation, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) + 0x80 (0x7f5438d91b80 in /usr/local/lib/python3.12/dist-packages/torch/lib/libc10.so)
frame #1: <unknown function> + 0xe69b51 (0x7f5439c7eb51 in /usr/local/lib/python3.12/dist-packages/torch/lib/libtorch_cuda.so)
frame #2: <unknown function> + 0x951271 (0x7f5439766271 in /usr/local/lib/python3.12/dist-packages/torch/lib/libtorch_cuda.so)
frame #3: <unknown function> + 0xecdb4 (0x7f5632bb1db4 in /usr/lib/x86_64-linux-gnu/libstdc++.so.6)
frame #4: <unknown function> + 0x9caa4 (0x7f563575baa4 in /usr/lib/x86_64-linux-gnu/libc.so.6)
frame #5: <unknown function> + 0x129c6c (0x7f56357e8c6c in /usr/lib/x86_64-linux-gnu/libc.so.6)

Fatal Python error: Aborted

Thread 0x00007ef3cbfff6c0 (most recent call first):
  File "/sgl-workspace/sglang/python/sglang/srt/mem_cache/storage/mooncake_store/mooncake_store.py", line 783 in _put_batch_zero_copy_impl
  File "/sgl-workspace/sglang/python/sglang/srt/mem_cache/storage/mooncake_store/mooncake_store.py", line 599 in batch_set_v1
  File "/sgl-workspace/sglang/python/sglang/srt/managers/cache_controller.py", line 1045 in _page_set_zero_copy
  File "/sgl-workspace/sglang/python/sglang/srt/managers/cache_controller.py", line 1075 in _page_backup
  File "/sgl-workspace/sglang/python/sglang/srt/managers/cache_controller.py", line 1101 in backup_thread_func
  File "/usr/lib/python3.12/threading.py", line 1010 in run
  File "/usr/lib/python3.12/threading.py", line 1073 in _bootstrap_inner
  File "/usr/lib/python3.12/threading.py", line 1030 in _bootstrap

Thread 0x00007eef1bffe6c0 (most recent call first):
  File "/sgl-workspace/sglang/python/sglang/srt/mem_cache/storage/mooncake_store/mooncake_store.py", line 788 in _get_batch_zero_copy_impl
  File "/sgl-workspace/sglang/python/sglang/srt/mem_cache/storage/mooncake_store/mooncake_store.py", line 547 in batch_get_v1
  File "/sgl-workspace/sglang/python/sglang/srt/managers/cache_controller.py", line 823 in _page_get_zero_copy
  File "/sgl-workspace/sglang/python/sglang/srt/managers/cache_controller.py", line 870 in _page_transfer
  File "/sgl-workspace/sglang/python/sglang/srt/managers/cache_controller.py", line 891 in prefetch_io_aux_func`

xiezhq-hermann:
Thanks for the PR! A few points to help move this forward:

1. Race condition framing. Could you clarify the specific race you're describing for "page indices in req_to_token"? My understanding is that a HiCache load doesn't mutate existing page-table entries — it allocates new device pages, and the indices are written into req_to_token synchronously in prepare_for_extend before the forward pass is dispatched. The async work on the HiCache stream is the KV-value copy into those already-allocated slots. The wait_event is used during prefill only, therefore should not interference with decoding cuda graph replay.

2. Please split HiCache + EAGLE into its own PR. Changes 2 and 3 (memory relocation version tracking + last_loc clamp/recompute) are a distinct problem from the CUDA-graph guard in Change 1, and mixing them makes review and future git-archaeology harder. A focused PR also gives us room to properly discuss the last_loc fix — the clamp especially needs its own justification, since clamping an index that "shouldn't" be out of bounds risks masking the underlying bug rather than fixing it.

3. Same as above — let's keep the EAGLE-path fixes isolated so we can reason about the alloc-triggered-eviction race on its own terms.

4. Draft attention backend switch should be a separate PR too. Change 4 is a config-level workaround for a latent NSA multi-step CUDA-graph bug, not a fix for the HiCache issue. Please pull it out and, ideally, file a tracking issue for the NSA bug with a repro so it doesn't get silently buried.

xiezhq-hermann:
deprecated

Issue #4097, opened 2025-03-05
[Bug] -inf in top_logprobs breaks FastAPI endpoint


hhschu:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

### Describe the bug

The top log probabilities returned by the engine can be as low as -Infinity, as shown in this example:

```json
{
  "text": "Yes",
  "meta_info": {
    "id": "3125813d8a5b4751870955c586437e29",
    "finish_reason": { "type": "stop", "matched": 128009 },
    "prompt_tokens": 142,
    "input_token_logprobs": [],
    "output_token_logprobs": [
      [-0.3755025565624237, 87575, null],
      [0.0, 82, null],
      [0.0, 128009, null]
    ],
    "input_top_logprobs": [],
    "output_top_logprobs": [
      [
        [-0.3755025565624237, 87575, null],
        [-1.410658836364746, 9642, null],
        [-3.209486961364746, 56, null],
        [-3.568861961364746, 2822, null],
        [-7.584486961364746, 45, null]
      ],
      [
        [0.0, 82, null],
        [-Infinity, 2, null],
        [-Infinity, 0, null],
        [-Infinity, 3, null],
        [-Infinity, 1, null]
      ],
      [
        [0.0, 128009, null],
        [-Infinity, 2, null],
        [-Infinity, 0, null],
        [-Infinity, 3, null],
        [-Infinity, 1, null]
      ]
    ],
    "completion_tokens": 3,
    "cached_tokens": 0
  }
}
```

Because [starlette by default doesn't allow nan of inf in the response](https://github.com/encode/starlette/blob/49158732f8a96744889f5a7644a42ae08131bd15/starlette/responses.py#L188), the -Infinity will trigger a "ValueError: Out of range float values are not JSON compliant" error.

[...]

PR #19183, opened 2026-02-23
[fix] use ORJSONResponse for non-stream /generate

wang-zhuoran:
## Motivation

The non-stream `/generate` endpoint can return payloads containing non-finite values
(e.g. `-inf` in `meta_info.top_logprobs`). Returning the raw dict on this path may
lead to serialization inconsistencies.

## Modifications

- Set `response_class=ORJSONResponse` for `/generate`
- Return `ORJSONResponse(content=ret)` on the non-stream path
- Add a regression unit test for non-stream generate response serialization

## Accuracy Tests

N/A (HTTP response serialization fix only)

## Benchmarking and Profiling

N/A (no kernel/model/inference speed changes)

## Checklist

- [x] Follow the SGLang code style guidance.

Additional tests run:
- `PYTHONPATH=python python -m unittest discover -s python/sglang/test -p test_http_server_generate_response.py -v`
- `PYTHONPATH=python python -m pytest -q python/sglang/test/test_http_server_generate_response.py`

Note:
Related to #4097

Related PR: #6600 (closed)
This PR adds a regression test for the non-stream `/generate` response serialization path.

### DIFF
--- a/test/registered/sampling/test_http_server_generate_response.py
+++ b/test/registered/sampling/test_http_server_generate_response.py
@@ -1,0 +1,39 @@
+import types
+import unittest
+from unittest.mock import patch
+
+from sglang.srt.entrypoints import http_server
+from sglang.test.ci.ci_register import register_amd_ci, register_cuda_ci
+
+register_cuda_ci(est_time=1, suite="stage-b-test-small-1-gpu")
+register_amd_ci(est_time=1, suite="stage-b-test-small-1-gpu-amd")
+
+
+class _FakeTokenizerManager:
+    def generate_request(self, _obj, _request):
+        async def _gen():
+            yield {
+                "text": "ok",
+                "meta_info": {
+                    "top_logprobs": [[float("-inf"), -1, None]],
+                },
+            }
+
+        return _gen()
+
+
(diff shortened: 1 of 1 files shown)


test/registered/sampling/test_http_server_generate_response.py
```diff
@@ -0,0 +1,33 @@
+import types
```
hnyls2002, line 1: Add this file to `test/registered`, put it together with the logprob tests. And register is to CI

hnyls2002 (changes requested):

test/registered/sampling/test_http_server_generate_response.py
wang-zhuoran: Thanks for the review. I updated this PR to follow your suggestion.

Changes made:
1. Moved the test into `test/registered/sampling/`
2. Registered it to CI (`register_cuda_ci` + `register_amd_ci`)
3. Kept this PR as test-only (no `http_server.py` logic change), since upstream `main` already contains the runtime fix

The test now verifies non-stream `/generate` JSON serialization for non-finite logprob values (expects `null`, not `-Infinity`).

Could you please take another look?

PR #19190, opened 2026-02-23
[SGL] sync patch: Remove sync points, prefill cudagraph for DP, disable cache reset in mem check

YazhiGao:
## Summary
- **Remove CPU-GPU sync points** in logits_processor, schedule_batch, and memory_pool: use `pin_memory=True` + `.to(device, non_blocking=True)` instead of direct device tensor construction, use `output_size` in `repeat_interleave` to avoid implicit sync, build mamba track indices on GPU without scalar extraction, use GPU zero-expand instead of `index_fill_`
- **Add `all_extend_in_batch` field** in schedule_batch, forward_batch_info, and ModelWorkerBatch to support prefill CUDA graph capture with DP attention
- Minor: connector `device=None` default, `empty_cache=False` in weight update logging

## Test plan
- [ ] Verify no regressions on existing CI tests
- [ ] Test with DP attention enabled

### DIFF
--- a/python/sglang/srt/mem_cache/memory_pool.py
+++ b/python/sglang/srt/mem_cache/memory_pool.py
@@ -337,10 +337,18 @@ class MambaPool:
 
         select_index = self.free_slots[:need_size]
         self.free_slots = self.free_slots[need_size:]
-        # clear at alloc time, fill allocated slots with zeros
+        # clear at alloc time — expand a scalar GPU zero to the right shape, no CPU-GPU sync
         for i in range(len(self.mamba_cache.conv)):
-            self.mamba_cache.conv[i][:, select_index] = 0
-        self.mamba_cache.temporal[:, select_index] = 0
+            t = self.mamba_cache.conv[i]
+            z = torch.zeros(1, dtype=t.dtype, device=t.device).expand(
+                t.shape[0], need_size, *t.shape[2:]
+            )
+            t[:, select_index] = z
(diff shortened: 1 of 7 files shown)


Qiaolin-Yu (approved):

python/sglang/srt/mem_cache/memory_pool.py
```diff
@@ -337,10 +337,18 @@ def alloc(self, need_size: int) -> Optional[torch.Tensor]:
 
         select_index = self.free_slots[:need_size]
         self.free_slots = self.free_slots[need_size:]
-        # clear at alloc time, fill allocated slots with zeros
+        # clear at alloc time — expand a scalar GPU zero to the right shape, no CPU-GPU sync
         for i in range(len(self.mamba_cache.conv)):
-            self.mamba_cache.conv[i][:, select_index] = 0
-        self.mamba_cache.temporal[:, select_index] = 0
+            t = self.mamba_cache.conv[i]
+            z = torch.zeros(1, dtype=t.dtype, device=t.device).expand(
```
ispobock, line 343: Can we use `t[:, select_index].zero_()` to make it simpler ?

YazhiGao: `t[:, select_index]` is not a view, im not sure it will be correct syntax sugar to evaluate to index_put_

ispobock (approved):

hnyls2002 (approved):

PR #17220, opened 2026-01-16
Evict swa kv cache during decoding

ispobock:
## Motivation
- Before: SWARadixCache only inserts new tokens into the radix tree when (chunked) prefill finishes or a whole request finishes, so out of window tokens can’t be evicted during decoding.
- Issue: Long generations waste memory and limit swa’s benefit.
- Now: Only cache the last sliding_window_size tokens for finished requests (for multi-turn prefix reuse). Out of window tokens can be evicted early without hurting cache hit rate.

## Modifications

- Unify SWA eviction for SWARadixCache and SWAChunkCache. Since these tokens are not inserted into RadixCache during decoding, move the eviction logic from the cache implementation into the schedule batch.
- Update cache_finished_req to consider the evicted tokens.

## Benchmarking and Profiling
```
python3 -m sglang.launch_server --model-path XiaomiMiMo/MiMo-V2-Flash --tp-size 4 --trust-remote-code --mem-fraction-static 0.7 --model-loader-extra-config '{"enable_multithread_load": "true","num_threads": 64}' --attention-backend fa3 --page-size 64 --swa-full-tokens-ratio 0.2

python3 benchmark/hicache/bench_multiturn.py --model-path XiaomiMiMo/MiMo-V2-Flash --disable-random-sample \
--output-length 1 --request-length 300 \
--num-clients 10 --num-rounds 10 --max-parallel 10 --request-rate 16 \
--ready-queue-policy random --disable-auto-run --enable-round-barrier --output-length 3000
```

main:
```
[2026-01-19 02:58:09 TP0] Decode batch, #running-req: 10, #full token: 330688, full token usage: 0.37, #swa token: 48384, swa token usage: 0.27, cuda graph: True, gen throughput (token/s): 836.43, #queue-req: 0, 
[2026-01-19 02:58:10 TP0] Decode batch, #running-req: 10, #full token: 331072, full token usage: 0.37, #swa token: 48768, swa token usage: 0.27, cuda graph: True, gen throughput (token/s): 836.62, #queue-req: 0, 
[2026-01-19 02:58:10 TP0] Decode batch, #running-req: 10, #full token: 331392, full token usage: 0.37, #swa token: 49088, swa token usage: 0.27, cuda graph: True, gen throughput (token/s): 837.37, #queue-req: 0, 
[...]

PR #19199, opened 2026-02-24
[SWA] Remove GPU sync points from SWA memory free path

YazhiGao:
## Summary
- Remove CPU-GPU synchronization points in the SWA memory pool free path
- Replace `swa_indices[swa_indices > 0]` boolean mask gather with `swa_evicted_count` CPU int slicing
- Replace `torch.unique()` page dedup in `free_group_end()` with `_page_select()` built from known entry lengths
- Add `page_select` parameter to `PagedTokenToKVPoolAllocator.free()` as sync-free alternative to `torch.unique()`
- Fix SWA memory leak in `_insert_helper` non-tombstone branch (was freeing sentinel index 0)

## Motivation
Each `torch.unique()` and variable-size boolean mask forces a GPU-to-CPU sync to determine output tensor shape. In the SWA free path this happens per-batch, adding latency on the critical scheduling path. Since the scheduler already tracks `swa_evicted_seqlen` (a CPU int), we can slice off already-evicted tokens without any GPU round-trip.

## Test plan
- [ ] Existing SWA unit tests pass
- [ ] Monitor `swa_available_size()` over extended runs for leaks
- [ ] Verify no regression on non-SWA models

### DIFF
--- a/python/sglang/srt/mem_cache/allocator.py
+++ b/python/sglang/srt/mem_cache/allocator.py
@@ -499,12 +499,25 @@ class PagedTokenToKVPoolAllocator(BaseTokenToKVPoolAllocator):
         self.free_pages = self.free_pages[num_new_pages:]
         return out_indices
 
-    def free(self, free_index: torch.Tensor):
+    def free(self, free_index: torch.Tensor, page_select: torch.Tensor = None):
+        """Free token slots back to the pool.
+
+        Args:
+            free_index: token-level indices to free.
+            page_select: optional pre-computed selection indices into free_index
+                that pick one representative per page. When multiple entries with
+                sub-page tails are batched, the caller can build this via cumsum
+                over per-entry strides to avoid the GPU sync from torch.unique.
+                If None, uses torch.unique (correct but causes a GPU sync).
+        """
         if free_index.numel() == 0:
             return
 
         if self.is_not_in_free_group:
-            free_page_indices = torch.unique(free_index // self.page_size)
+            if page_select is not None:
+                free_page_indices = (free_index // self.page_size)[page_select]
+            else:
+                free_page_indices = torch.unique(free_index // self.page_size)
             if self.need_sort:
                 self.release_pages = torch.cat((free_page_indices, self.release_pages))
             else:
--- a/python/sglang/srt/mem_cache/swa_memory_pool.py
+++ b/python/sglang/srt/mem_cache/swa_memory_pool.py
@@ -414,26 +414,159 @@ class SWATokenToKVPoolAllocator(BaseTokenToKVPoolAllocator):
 
         return alloc_full_indices
 
-    def free(self, free_index: torch.Tensor):
+    def free(self, free_index: torch.Tensor, swa_evicted_count: int = 0):
+        """Free full-attention and SWA KV slots.
+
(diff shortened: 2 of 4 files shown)


python/sglang/srt/mem_cache/allocator.py
```diff
@@ -499,12 +499,25 @@ def alloc_decode(
...
+                over per-entry strides to avoid the GPU sync from torch.unique.
+                If None, uses torch.unique (correct but causes a GPU sync).
+        """
         if free_index.numel() == 0:
             return
 
         if self.is_not_in_free_group:
-            free_page_indices = torch.unique(free_index // self.page_size)
+            if page_select is not None:
+                free_page_indices = (free_index // self.page_size)[page_select]
```
YazhiGao, line 518: @hanming-lu  unique is actually very easier with jagged select. this pr solves more on the [swa_indice > 0]. part......

python/sglang/srt/mem_cache/swa_memory_pool.py
```diff
@@ -414,26 +414,159 @@ def alloc_decode(
...
+        determine the output tensor shape.
+        """
         if free_index.numel() == 0:
             return
 
         # NOTE: the API is not idempotent.
         if self.is_not_in_free_group:
             self.full_attn_allocator.free(free_index)
-            self.free_swa(free_index)
+            self.free_swa(free_index[swa_evicted_count:])
```
ispobock, line 443: In this way, we need to be very careful about the `swa_evicted_count` calculation. I'm not sure if it's necessary. Did you see any performance gain it can bring?

python/sglang/srt/mem_cache/swa_radix_cache.py
```diff
@@ -447,7 +447,9 @@ def cache_finished_req(self, req: Req, is_insert: bool = True) -> None:
             kv_indices = self.req_to_token_pool.req_to_token[
                 req.req_pool_idx, :kv_committed_len
             ]
-            self.token_to_kv_pool_allocator.free(kv_indices)
+            self.token_to_kv_pool_allocator.free(
+                kv_indices, swa_evicted_count=req.swa_evicted_seqlen
```
ispobock: Actually `swa_evicted_seqlen` is introduced in https://github.com/sgl-project/sglang/pull/17220 to identify that [cache_protected_len, swa_evicted_seqlen] are evicted during decoding. For kv indices in [0,cache_protected_len], the eviction is managed by the tree cache and may not be evicted.

python/sglang/srt/mem_cache/swa_memory_pool.py
YazhiGao: w/o this change overlap should always be fake? schedule stream kernel is not free and i find the better modle kernel u have it is harder for the schedule stream kernel to schedule timely and they always sync .there are tons of cpu-gpu in cache layer that prevent the cpu to move ahead of GPU to by one full cpu iter. yeah i know the complexity, this is the only way i know to solve it sync free :(

YazhiGao: im actually curious, do u have a OSS model SWA trace for me to compare? im also shocked why this is not a problem on SGL main.

PR #19202, opened 2026-02-24
refactor: extract device-to-backend mapping into get_default_distributed_backend

sywangyi:
## Summary

- Extract device-to-distributed-backend mapping into `get_default_distributed_backend()` in `parallel_state.py`, replacing duplicated if-elif chains
- Fix `encode_server.py` to pass the correct backend based on device type instead of relying on the default `"nccl"`
- Refactor `model_runner.py` to use the shared helper (mooncake override kept in place)
- Remove silent nccl-to-gloo fallback from `init_distributed_environment` — callers are now responsible for passing the correct backend

## Motivation

`MMEncoder` in `encode_server.py` crashes when running on CPU (e.g. `CUDA_VISIBLE_DEVICES=""`) because it always defaults to `"nccl"`. Rather than adding a band-aid fallback inside `init_distributed_environment`, this PR centralizes the device→backend mapping so all callers get it right.

## Test plan

- [ ] Verify `MMEncoder` initializes correctly with `CUDA_VISIBLE_DEVICES=""`
- [ ] Verify normal GPU serving path is unaffected


hnyls2002:
@sywangyi, could you please check whether the following commits solve your issue?

ShangmingCai (approved):
The logic is straightforward, LGTM.

But I think we need to make sure it passes full CI before merging.

PR #19205, opened 2026-02-24
[PD-Disagg] Deduplicate common KVManager methods into CommonKVManager

hnyls2002:
## Summary

- Lift `check_status`, `update_status`, and `record_failure` from mooncake/mori/nixl KVManager into `CommonKVManager`, removing identical implementations from all three backends
- Move `failure_records` and `failure_lock` fields to `CommonKVManager.__init__` (previously duplicated in mooncake and mori; nixl now also gets them for free via inheritance)


ShangmingCai (approved):
LGTM

PR #16662, opened 2026-01-07
[diffusion] Enable Cache‑DiT config for diffusers backend

qimcis:
## Motivation
Addressing #16642, enabling Cache‑DiT acceleration for any diffusers pipeline in SGLang by allowing a cache‑dit config file to be passed through the diffusers backend.

## Modifications
- `--cache-dit-config` server arg in `python/sglang/multimodal_gen/runtime/server_args.py`
- cache‑dit when using diffusers pipelines, including config loading and module resolution: `python/sglang/multimodal_gen/runtime/pipelines/diffusers_pipeline.py`.
- fixed cli `--diffusers-kwargs` path to avoid `SamplingParams` init errors: `python/sglang/multimodal_gen/configs/sample/sampling_params.py`.

## Benchmarking and Profiling
2x RTX Pro 6000 WS

- Cache‑DiT config:
```bash
cat > /tmp/cache_dit_config.yaml <<'EOF'
cache_config:
  max_warmup_steps: 8
  warmup_interval: 2
  max_cached_steps: -1
  max_continuous_cached_steps: 2
  Fn_compute_blocks: 1
  Bn_compute_blocks: 0
  residual_diff_threshold: 0.12
  enable_taylorseer: true
  taylorseer_order: 1
parallelism_config:
  ulysses_size: 2
  parallel_kwargs:
    attention_backend: native
    extra_parallel_modules: ["text_encoder", "vae"]
EOF
```

## Test 1

- Baseline:
  - 43.79s, peak 67.80 GB
  - Command:
```bash
CUDA_VISIBLE_DEVICES=0,1 sglang generate \
  --model-path Qwen/Qwen-Image \
  --backend diffusers \
  --num-gpus 2 \
  --sp-degree 2 \
  --ulysses-degree 2 \
  --tp-size 1 \
  --prompt 'A coffee shop entrance features a chalkboard sign reading "Qwen Coffee $2 per cup," with a neon light beside it displaying "通义千问". Next to it hangs a poster showing a beautiful Chinese woman, and beneath the poster is written "π≈3.1415926-53589793-23846264-33832795-02384197". Ultra HD, 4K, cinematic composition.' \
  --negative-prompt " " \
  --width 1664 \
  --height 928 \
  --num-inference-steps 50 \
  --seed 42 \
  --save-output \
  --output-path outputs \
  --output-file-name baseline-2gpu-a.png
```
baseline-2gpu-a

- Cache‑DiT:
[...]

PR #19213, opened 2026-02-24
[diffusion] Add cache-dit CI tests

qimcis:
## Motivation

Add CI tests for changes in #16662 , and a CI test for cache-dit on native diffusion engine as well

## Modifications

Add 1 gpu test, 2 gpu test and baselines

## Checklist


qimcis:
output:
585abc22-97c5-48b8-aee0-db3718a84713

DefTruth:
@qimcis @mickqian Any updates? I think it would be better if we could add some tests for the cache to help SGLD avoid errors related to cache (e.g., #19955, #19965) introduced by new commits.

python/sglang/multimodal_gen/test/server/configs/cache_dit_config_1gpu.yaml
```diff
@@ -0,0 +1,10 @@
+cache_config:
```
mickqian, line 1: gather to a `configs` folder?

qimcis:
I updated this PR, instead of focusing only on cache-dit diffusers and the configs, I've also added a native SGLD 2-gpu sp-only cache-dit regression case, which uses the native SGLD cache-dit path with SP enabled and TP disabled, covering the class of issue from #19955 / #19965 around cache-dit parallelism config construction

In addition to this, i kept the 1-GPU diffusers cache-dit config coverage that i added previously (including one scm case)

Perf baselines have been regenerated onto current main too (2x h100 SXM)

Outputs for the 1gpu tests look like:

qwen_image_t2i_cache_dit_config_diffusers_1gpu


For the 2gpu test with wan, the output is not as important, as it's meant to measure simply that the execution path does not crash and produces a valid generation

Can you take a look to make sure things look ok before I get mick to review again? @DefTruth Thank you!

python/sglang/multimodal_gen/test/server/gpu_cases.py
```diff
@@ -521,6 +553,17 @@
         ),
         T2V_sampling_params,
     ),
+    DiffusionTestCase(
```
mickqian, line 551: these tests are heavy, could we try to make it lightweight

mickqian (approved):

mickqian:
CI is failing

python/sglang/multimodal_gen/test/server/gpu_cases.py
qimcis: i removed a redundant test, and made some changes, is this ok? passes locally for me @mickqian

PR #18345, opened 2026-02-06
refactor FAKE transfer backend and remove --disaggregation-decode-enable-fake-auto parameter

Estrella-xx:
## Motivation

When the --disaggregation-decode-enable-fake-auto parameter is enabled, the fake transfer backend is used. Therefore, the fake transfer backend is refactored so that a specific fake transfer backend can be explicitly specified, enabling requests to be sent directly to the decode stage.

previous discussion #17811

## Modifications

1.Add the FakeKVManager class.
2.Remove the --disaggregation-decode-enable-fake-auto parameter and replace it with the --disaggregation-transfer-backend parameter to enable this functionality. When the value of --disaggregation-transfer-backend is set to fake, requests can be sent directly to the decode stage.

## Accuracy Tests
1.NPU
```
MODEL_PATH=/home/weights/Qwen3-32B/
python3 -m sglang.launch_server --model-path ${MODEL_PATH} \
--tp 2 \
--dp-size 2 \
--trust-remote-code \
--device npu \
--disaggregation-mode decode \
--attention-backend ascend \
--disaggregation-transfer-backend fake \
--load-balance-method follow_bootstrap_room \
--mem-fraction-static 0.8 \
--base-gpu-id 12 \
--host 127.0.0.1 \
--port 8008 \
```
```
curl http://localhost:8008/generate \
  -H "Content-Type: application/json" \
  -d '{
    "text": "The capital of France is",
    "sampling_params": {
      "temperature": 0,
      "max_new_tokens": 128
    }
  }' 
```
image

2.GPU
```
MODEL_PATH=/home/weight/Qwen3-32B/
python3 -m sglang.launch_server --model-path ${MODEL_PATH} \
--tp 2 \
--dp-size 2 \
--trust-remote-code \
--disaggregation-mode decode \
--attention-backend fa3 \
--piecewise-cuda-graph-max-tokens 4096 \
--chunked-prefill-size 8192 \
--disaggregation-transfer-backend fake \
--load-balance-method follow_bootstrap_room \
--mem-fraction-static 0.8 \
--base-gpu-id 4 \
--host 127.0.0.1 \
--port 8001 \
```
```
curl http://localhost:8001/generate \
  -H "Content-Type: application/json" \
  -d '{
    "text": "The capital of France is",
    "sampling_params": {
      "temperature": 0,
      "max_new_tokens": 128
    }
  }' 
```
image

## Checklist

[...]

PR #19215, opened 2026-02-24
[PD-Disagg] Support `FAKE` transfer backend for more cases

hnyls2002:
Prev #18345 refactor the `--disaggregation-decode-enable-fake-auto` into `FAKE` transfer backend. But the former implementation only works when dp attention and `follow_bootstrap_room` are enabled.

This PR enables more cases.


ShangmingCai (approved):
LGTM

PR #19225, opened 2026-02-24
[diffusion] Support stable-diffusion-3-medium-diffusers with sglang backend

gxlvera:
## Overview
This PR supports 

1. stable-diffusion-3-medium
2. stable-diffusion-3.5-medium
3. stable-diffusion-3.5-large

as sglang-native (using sglang as backend instead of diffusers).

Note: 
1. this PR hasn't supported TP
2. it doesn't use sglang USPAttention so not yet supported SP.

Run with cli:
```
SGLANG_USE_MODELSCOPE=true sglang generate   --model-path stabilityai/stable-diffusion-3-medium-diffusers  --prompt "Epic anime artwork of a wizard atop a mountain at night casting a cosmic spell into the dark sky that says Stable Diffusion 3 made out of colorful energy"   --height 1024   --width 1024   --num-inference-steps 20   --guidance-scale 7.0   --num-gpus 1 --backend sglang
```
Epic_anime_artwork_of_a_wizard_atop_a_mountain_at_night_casting_a_cosmic_spell_into_the_dark_sky_tha_20260224-064924_e044671b

## Checklist

## Review Process


python/sglang/multimodal_gen/runtime/models/registry.py
```diff
@@ -35,6 +35,7 @@
 _IMAGE_ENCODER_MODELS: dict[str, tuple] = {
     # "HunyuanVideoTransformer3DModel": ("image_encoder", "hunyuanvideo", "HunyuanVideoImageEncoder"),
     "CLIPVisionModelWithProjection": ("encoders", "clip", "CLIPVisionModel"),
+    "CLIPTextModelWithProjection": ("encoders", "clip", "CLIPTextModelWithProjection"),
```
gxlvera: This change might affect other models, I am not fully sure if it's intended. The reason I changed the mapping is that, we need to use CLIPTextModelWithProjection for correctly generating the pooled prompt embedding for SD3. Maybe we should implement it in a SD3-specific way?

python/sglang/multimodal_gen/registry.py
```diff
@@ -652,6 +658,11 @@ def _register_configs():
         pipeline_config_cls=QwenImageLayeredPipelineConfig,
         hf_model_paths=["Qwen/Qwen-Image-Layered"],
     )
+    register_configs(
+        sampling_param_cls=StableDiffusion3SamplingParams,
+        pipeline_config_cls=StableDiffusion3PipelineConfig,
+        hf_model_paths=["stabilityai/stable-diffusion-3-medium-diffusers"],
```
dreamyang-liu: We might need model_detectors here.
```
model_detectors=[lambda hf_id: "stablediffusion3" in hf_id.lower()],
```

python/sglang/multimodal_gen/runtime/models/encoders/clip.py
```diff
@@ -590,6 +590,51 @@ def load_weights(self, weights: Iterable[tuple[str, torch.Tensor]]) -> set[str]:
...
+            config.hidden_size, config.projection_dim, bias=False
+        )
+
+    def forward(
+        self,
+        input_ids: torch.Tensor | None,
+        position_ids: torch.Tensor | None = None,
+        attention_mask: torch.Tensor | None = None,
+        inputs_embeds: torch.Tensor | None = None,
+        output_hidden_states: bool | None = None,
```
dreamyang-liu, line 617: I know we have such signature `bool | None = None` somewhere else, but I want to understand what is the implication of output_hidden_states as None

python/sglang/multimodal_gen/runtime/models/dits/stablediffusion3.py
```diff
@@ -0,0 +1,219 @@
...
+            skip_layers (`list` of `int`, *optional*):
+                A list of layer indices to skip during the forward pass.
+
+        Returns:
+            If `return_dict` is True, an [`~models.transformer_2d.Transformer2DModelOutput`] is returned, otherwise a
+            `tuple` where the first element is the sample tensor.
+        """
+        if joint_attention_kwargs is not None:
+            joint_attention_kwargs = joint_attention_kwargs.copy()
+            lora_scale = joint_attention_kwargs.pop("scale", 1.0)
```
dreamyang-liu: Seems never used ?

python/sglang/multimodal_gen/runtime/models/dits/stablediffusion3.py
```diff
@@ -0,0 +1,219 @@
...
+            return_dict (`bool`, *optional*, defaults to `True`):
+                Whether or not to return a [`~models.transformer_2d.Transformer2DModelOutput`] instead of a plain
+                tuple.
+            skip_layers (`list` of `int`, *optional*):
+                A list of layer indices to skip during the forward pass.
+
+        Returns:
+            If `return_dict` is True, an [`~models.transformer_2d.Transformer2DModelOutput`] is returned, otherwise a
+            `tuple` where the first element is the sample tensor.
+        """
```
dreamyang-liu: These doc string are too verbose, and should place directly after the function signature

python/sglang/multimodal_gen/runtime/models/dits/stablediffusion3.py
```diff
@@ -0,0 +1,219 @@
...
+        num_layers = getattr(arch_config, "num_layers", 18)
+        attention_head_dim = getattr(arch_config, "attention_head_dim", 64)
+        num_attention_heads = getattr(arch_config, "num_attention_heads", 18)
+        joint_attention_dim = getattr(arch_config, "joint_attention_dim", 4096)
+        caption_projection_dim = getattr(arch_config, "caption_projection_dim", 1152)
+        pooled_projection_dim = getattr(arch_config, "pooled_projection_dim", 2048)
+        out_channels = getattr(arch_config, "out_channels", 16)
+        pos_embed_max_size = getattr(arch_config, "pos_embed_max_size", 96)
+        dual_attention_layers = getattr(arch_config, "dual_attention_layers", ())
+        qk_norm = getattr(arch_config, "qk_norm", None)
```
dreamyang-liu: Is these getattr make sense ? Should we make sure the ArchConfig have these field and default value ?

python/sglang/multimodal_gen/runtime/loader/component_loaders/vae_loader.py
```diff
@@ -152,3 +159,32 @@ def load_customized(
...
+
+    def _select_vae_weights(
+        self, safetensors_list, component_model_path, pipeline_config
+    ):
+        """Select appropriate VAE weights based on pipeline configuration."""
+        # Check if pipeline has specific VAE weight selection strategy
+        if (
+            hasattr(pipeline_config, "vae_precision")
+            and hasattr(pipeline_config, "use_precision_specific_weights")
+            and hasattr(pipeline_config, "vae_model_name")
```
dreamyang-liu: avoid hasattr

python/sglang/multimodal_gen/runtime/loader/component_loaders/text_encoder_loader.py
```diff
@@ -202,6 +216,28 @@ def is_not_first_encoder(module_name):
...
+    def _extract_encoder_index(component_name: str) -> int:
+        """
+        Map text encoder component names to zero-based indices.
+
+        Examples:
+        - text_encoder -> 0
+        - text_encoder_2 -> 1
+        - text_encoder_3 -> 2
+        """
+        match = re.search(r"_(\d+)$", component_name)
```
dreamyang-liu, line 256: This regex is a bit too relax, if passing unknown_1, it will also pass

python/sglang/multimodal_gen/configs/pipeline_configs/stablediffusion3.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+    Keep raw hidden states and select pre-final layer in TextEncodingStage
+    to match sglang-2 SD3 implementation.
+    """
+    assert outputs.hidden_states is not None
+    return outputs.hidden_states
+
+
+def t5_postprocess_text(outputs: BaseEncoderOutput, _text_inputs) -> torch.Tensor:
+    return outputs.last_hidden_state
```
dreamyang-liu, line 41: Reconsider do we really need these function

python/sglang/multimodal_gen/configs/pipeline_configs/stablediffusion3.py
```diff
@@ -0,0 +1,136 @@
...
+    SpatialImagePipelineConfig,
+)
+
+
+def t5_preprocess_text(prompt: str) -> str:
+    return prompt
+
+
+def clip_preprocess_text(prompt: str) -> str:
+    return prompt
```
dreamyang-liu: These functions seem do nothing.

python/sglang/multimodal_gen/configs/models/vaes/stablediffusion3.py
```diff
@@ -0,0 +1,125 @@
...
+            arch_config.block_out_channels = tuple(
+                config_dict.get("block_out_channels", [128, 256, 512, 512])
+            )
+            arch_config.layers_per_block = config_dict.get("layers_per_block", 2)
+            arch_config.norm_num_groups = config_dict.get("norm_num_groups", 32)
+            arch_config.attention_head_dim = config_dict.get("attention_head_dim", 8)
+
+        return cls(arch_config=arch_config)
+
+    def get_vae_scale_factor(self) -> int:
```
dreamyang-liu: Too short for a function

python/sglang/multimodal_gen/configs/models/vaes/stablediffusion3.py
```diff
@@ -0,0 +1,125 @@
...
+
+        # Set model class name for SD3 VAE
+        self.update_model_arch({"_class_name": "AutoencoderKL"})
+
+        # Update blend frames for image VAE
+        self.blend_num_frames = 0
+
+    def post_init(self):
+        """Additional initialization after model loading."""
+        pass
```
dreamyang-liu: Why we have this if we don't do anything ?

python/sglang/multimodal_gen/configs/models/vaes/stablediffusion3.py
```diff
@@ -0,0 +1,125 @@
...
+
+    # Compression ratios
+    spatial_compression_ratio: int = 8  # 8x spatial compression
+    temporal_compression_ratio: int = 1  # No temporal compression
+
+    # Model architecture parameters
+    in_channels: int = 3  # RGB input
+    out_channels: int = 3  # RGB output
+    latent_channels: int = 16  # SD3 transformer input channels
+    sample_size: int = 128  # Standard input size
```
dreamyang-liu: These per parameter comments not necessary

dreamyang-liu (dismissed):

python/sglang/multimodal_gen/registry.py
gxlvera: good point, i have added it

python/sglang/multimodal_gen/runtime/models/encoders/clip.py
gxlvera: when output_hidden_states = None, when calling self.text_model(), output_hidden_states will be set to self.config.output_hidden_states, you can check the codes here:
https://github.com/sgl-project/sglang/blob/0fd44ff342bb1e1ed7a110b5164034a288cc67a0/python/sglang/multimodal_gen/runtime/models/encoders/clip.py#L453-L457

python/sglang/multimodal_gen/runtime/models/registry.py
gxlvera: I might delete this, I will check how other sglang-supported models use clip. Maybe we only need to add a layer after calling the ClipTextModel for the projection head. I don't think there's the need to implement a CLIPTextModelWithProjection by myself.

python/sglang/multimodal_gen/runtime/models/dits/stablediffusion3.py
gxlvera: you're right, getatrr shouldn't be used 😂my bad.. will update later

python/sglang/multimodal_gen/runtime/models/dits/stablediffusion3.py
```diff
@@ -0,0 +1,219 @@
...
+        self.proj_out = nn.Linear(
+            self.inner_dim, patch_size * patch_size * self.out_channels, bias=True
+        )
+
+        self.gradient_checkpointing = False
+
+    def forward(
+        self,
+        hidden_states: torch.Tensor,
+        encoder_hidden_states: torch.Tensor | List[torch.Tensor] = [],
```
dreamyang-liu: Using empty list is a bit strange here, can we use None ?

gxlvera:
> please provide a performance comparison report according to `docs/diffusion/contributing.md`
> 
> 1. **Baseline**: run the benchmark (for a single generation task)
>    ```shell
>    $ sglang generate --model-path <model> --prompt "A benchmark prompt" --perf-dump-path baseline.json
>    ```
> 2. **New**: run the same benchmark, without modifying any server_args or sampling_params
>    ```shell
>    $ sglang generate --model-path <model> --prompt "A benchmark prompt" --perf-dump-path new.json
>    ```
> 3. **Compare**: run the compare script, which will print a Markdown table to the console
>    ```shell
>    $ python python/sglang/multimodal_gen/benchmarks/compare_perf.py baseline.json new.json [new2.json ...]
>    ### Performance Comparison Report
>    ...
>    ```
> 4. **Paste**: paste the table into the PR description

Ok I will do that, thanks!

python/sglang/multimodal_gen/runtime/models/dits/stablediffusion3.py
gxlvera: lora_scale is indeed not used, and the codes will never trigger the warining "                    "Passing `scale` via `joint_attention_kwargs` when not using the PEFT backend is ineffective."
I have changed it to be
```
if joint_attention_kwargs is not None:
            joint_attention_kwargs = joint_attention_kwargs.copy()
            if joint_attention_kwargs.pop("scale", None) is not None:
                logger.warning(
                    "Passing `scale` via `joint_attention_kwargs` when not using the PEFT backend is ineffective."
                )
```

python/sglang/multimodal_gen/runtime/models/dits/stablediffusion3.py
gxlvera: fixed

python/sglang/multimodal_gen/runtime/models/dits/stablediffusion3.py
gxlvera: I have abandoned the usage of getattr, and I added some missing attribute in sd3 dit config file so that the values of those attribute to be set correctly in sd3 dit

python/sglang/multimodal_gen/runtime/loader/component_loaders/vae_loader.py
gxlvera: fixed

python/sglang/multimodal_gen/runtime/loader/component_loaders/text_encoder_loader.py
gxlvera: you are right, but the previous codes also don't catch that. I am not quite sure if making it stricter like checking the component_name prefix would cause some problems to other model, I think we can leave it as it is for now?

python/sglang/multimodal_gen/configs/pipeline_configs/stablediffusion3.py
gxlvera: We need these postprocess functions because encoder outputs are pipeline-specific and even sometimes encoder-specific, take SD3 as an example). Keeping postprocess as callback function lets TextEncodingStage stay generic and avoids hardcoding per-model and per-encoder branches in TextEncodingStage

python/sglang/multimodal_gen/configs/pipeline_configs/stablediffusion3.py
gxlvera: We need these postprocess functions because encoder outputs are pipeline-specific and even sometimes encoder-specific, take SD3 as an example). Keeping postprocess as callback function lets TextEncodingStage stay generic and avoids hardcoding per-model and per-encoder branches in TextEncodingStage

python/sglang/multimodal_gen/configs/models/vaes/stablediffusion3.py
gxlvera: 🤔️I don't think this is a big issue?

python/sglang/multimodal_gen/configs/models/vaes/stablediffusion3.py
gxlvera: good catch thanks

python/sglang/multimodal_gen/runtime/pipelines_core/stages/text_encoding.py
```diff
@@ -246,6 +253,19 @@ def encode_text(
...
+            if is_stable_diffusion3:
+                tok_kwargs["padding"] = "max_length"
+                if i != 2:
+                    tok_kwargs["max_length"] = (
+                        getattr(encoder_config, "text_len", None)
+                        or getattr(encoder_config, "max_position_embeddings", None)
+                        or 77
+                    )
+                else:
+                    tok_kwargs["max_length"] = 256
```
gxlvera: hardcoding here is not elegant, I will fix that

python/sglang/multimodal_gen/runtime/models/dits/stablediffusion3.py
gxlvera: right! None is better

python/sglang/multimodal_gen/configs/models/vaes/stablediffusion3.py
zhaochenyang20: done

python/sglang/multimodal_gen/configs/models/vaes/stablediffusion3.py
zhaochenyang20: As I searched across the repo. Other models have this short funciton. So it's good.

zhaochenyang20 (changes requested):

python/sglang/multimodal_gen/configs/pipeline_configs/stablediffusion3.py
zhaochenyang20: These functions are also in `python/sglang/multimodal_gen/configs/pipeline_configs/hunyuan.py`, but `flux.py` directly import hunyuan's function. So I will do the same thing following Flux.

python/sglang/multimodal_gen/configs/pipeline_configs/stablediffusion3.py
```diff
@@ -0,0 +1,137 @@
...
```
zhaochenyang20: In some sense, I found that these functions are quite weird still. As we also have this funciton in hunyuan model:

```python
def clip_preprocess_text(prompt: str) -> str:
    return prompt
```

`python/sglang/multimodal_gen/configs/pipeline_configs/hunyuan.py`. In SD3 PR, we should not chage this function. I think we should found out why there is a function like this. And could we remove this function.

zhaochenyang20 (changes requested):

python/sglang/multimodal_gen/configs/pipeline_configs/stablediffusion3.py
zhaochenyang20: As I list here:

https://github.com/sgl-project/sglang/issues/19525

python/sglang/multimodal_gen/configs/pipeline_configs/stablediffusion3.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+def clip_postprocess_text(outputs: BaseEncoderOutput, _text_inputs) -> torch.Tensor:
+    """
+    Keep raw hidden states and select pre-final layer in TextEncodingStage
+    to match sglang-2 SD3 implementation.
```
zhaochenyang20: what is `sglang-2`?

python/sglang/multimodal_gen/configs/pipeline_configs/stablediffusion3.py
zhaochenyang20: https://github.com/sgl-project/sglang/pull/19225/changes#r2866541381

The same as

zhaochenyang20 (changes requested):

python/sglang/multimodal_gen/configs/pipeline_configs/stablediffusion3.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+def t5_postprocess_text(outputs: BaseEncoderOutput, _text_inputs) -> torch.Tensor:
+    return outputs.last_hidden_state
+
+
+@dataclass
+class StableDiffusion3PipelineConfig(SpatialImagePipelineConfig):
+    """
+    Configuration for SD3 text/image generation.
```
zhaochenyang20: `Configuration for SD3 text/image generation.`

What does this `text` mean? SD3 can do a text generation?

python/sglang/multimodal_gen/runtime/loader/component_loaders/text_encoder_loader.py
zhaochenyang20: Yes. I think we are good right now. But please raise an issue for this case, require others to do refactor. LIke what I do for this:

https://github.com/sgl-project/sglang/issues/19352

python/sglang/multimodal_gen/runtime/loader/utils.py
```diff
@@ -155,9 +155,7 @@ def __exit__(self, exc_type, exc_value, traceback):
 
 def _normalize_component_type(module_type: str) -> str:
     """Normalize module types like 'text_encoder_2' -> 'text_encoder'."""
-    if module_type.endswith("_2"):
-        return module_type[:-2]
-    return module_type
```
zhaochenyang20, line 186: Is this change really equivalent?

python/sglang/multimodal_gen/runtime/pipelines_core/stages/text_encoding.py
zhaochenyang20: refactor this in this PR

zhaochenyang20 (changes requested):

python/sglang/multimodal_gen/runtime/pipelines_core/stages/denoising.py
```diff
@@ -1388,6 +1389,76 @@ def _predict_noise_with_cfg(
...
+                    target_dtype=target_dtype,
+                    guidance=guidance,
+                    **image_kwargs,
+                    **pos_cond_kwargs,
+                )
+                noise_pred_cond = server_args.pipeline_config.slice_noise_pred(
+                    noise_pred_cond, latents
+                )
+            return noise_pred_cond
+
```
zhaochenyang20: Do not make such specific logits for SD3 in general classes.

python/sglang/multimodal_gen/runtime/pipelines_core/stages/text_encoding.py
```diff
@@ -246,6 +252,19 @@ def encode_text(
...
```
zhaochenyang20: Do not invasively change this general class for a specific model

zhaochenyang20: Also, do not use`getattr`

zhaochenyang20: There are several issues with this code:

- Use of getattr: This violates our coding standards/conventions.
- Magic Numbers: The values 77 and 256 are used without context or named constants.
- Fragile Logic: Distinguishing encoder types using the index i != 2 is brittle; the logic will break if the encoder sequence is reordered.
- Redundancy: The implementation duplicates values already defined in the config, indicating an inconsistent architectural design.

python/sglang/multimodal_gen/runtime/models/dits/stablediffusion3.py
```diff
@@ -0,0 +1,203 @@
...
+            shape=(
+                hidden_states.shape[0],
+                height,
+                width,
+                patch_size,
+                patch_size,
+                self.out_channels,
+            )
+        )
+        hidden_states = torch.einsum("nhwpqc->nchpwq", hidden_states)
```
zhaochenyang20: Caught by claude:

`torch.einsum` is relatively inefficient for simple permutation operations, and its readability is inferior to `permute` or `rearrange`. This can be directly implemented using `hidden_states.permute(0, 5, 1, 3, 2, 4).reshape(...)` or `einops.rearrange`.

python/sglang/multimodal_gen/configs/pipeline_configs/stablediffusion3.py
```diff
@@ -0,0 +1,136 @@
...
+            clip_postprocess_text,
+            clip_postprocess_text,
+            t5_postprocess_text,
+        )
+    )
+
+    # SD3 specific parameters
+    should_use_guidance: bool = False
+    guidance_scale: float = 7.0
+    use_precision_specific_weights = True
```
zhaochenyang20: `use_precision_specific_weights: bool = True`

python/sglang/multimodal_gen/runtime/pipelines/stable_diffusion_3.py
```diff
@@ -0,0 +1,120 @@
...
+        return batch
+
+    def _merge_embeddings(
+        self,
+        prompt_embeds_list: list[torch.Tensor],
+        pooled_embeds_list: list[torch.Tensor],
+        batch: Req,
+    ) -> Req:
+        if len(prompt_embeds_list) != 3:
+            return batch
```
zhaochenyang20: If not `3`, raise errors

zhaochenyang20 (changes requested):
The biggest issue is injecting model-specific logic into shared stages via isinstance — this will degrade as more pipelines are added. I recommend requiring the author to refactor text_encoding and denoising SD3 logic to be fully config/protocol-driven rather than type-checked.

python/sglang/multimodal_gen/runtime/pipelines/stable_diffusion_3.py
```diff
@@ -0,0 +1,120 @@
...
+        clip_merged = torch.nn.functional.pad(
+            clip_merged, (0, t5.shape[-1] - clip_merged.shape[-1])
+        )
+        batch.prompt_embeds = [torch.cat([clip_merged, t5], dim=-2)]
+        batch.pooled_embeds = [
+            torch.cat([pooled_embeds_list[0], pooled_embeds_list[1]], dim=-1)
+        ]
+        return batch
+
+    def _merge_negative_embeddings(
```
zhaochenyang20: `_merge_embeddings` and `_merge_negative_embeddings` violate DRY — identical logic duplicated. Extract the core merge logic into a pure function.

python/sglang/multimodal_gen/runtime/pipelines/stable_diffusion_3.py
zhaochenyang20: Silent return when `len(prompt_embeds_list) != 3` should be a ValueError — SD3 always requires exactly 3 embeddings.

python/sglang/multimodal_gen/runtime/pipelines_core/stages/denoising.py
```diff
@@ -1388,6 +1389,76 @@ def _predict_noise_with_cfg(
...
+                    pos_v = pos_cond_kwargs.get(key)
+                    neg_v = neg_cond_kwargs.get(key)
+                    if isinstance(pos_v, torch.Tensor) and isinstance(
+                        neg_v, torch.Tensor
+                    ):
+                        merged_kwargs[key] = torch.cat([neg_v, pos_v], dim=0)
+                    elif pos_v is not None:
+                        merged_kwargs[key] = pos_v
+                    elif neg_v is not None:
+                        merged_kwargs[key] = neg_v
```
zhaochenyang20: Missing `else` branch per style guide.

python/sglang/multimodal_gen/configs/pipeline_configs/stablediffusion3.py
```diff
@@ -0,0 +1,136 @@
...
+    vae_model_name = "diffusion_pytorch_model"
```
zhaochenyang20: `vae_model_name: str = `

python/sglang/multimodal_gen/runtime/models/dits/stablediffusion3.py
```diff
@@ -0,0 +1,203 @@
...
+        temb = self.time_text_embed(timestep, pooled_projections)
+        encoder_embeddings = self.context_embedder(encoder_embeddings)
+
+        for index_block, block in enumerate(self.transformer_blocks):
+            # Skip specified layers
+            is_skip = (
+                True
+                if skip_layers is not None and index_block in skip_layers
+                else False
+            )
```
zhaochenyang20: `is_skip = skip_layers is not None and index_block in skip_layers`

python/sglang/multimodal_gen/runtime/pipelines_core/stages/text_encoding.py
```diff
@@ -271,6 +292,26 @@ def encode_text(
             prompt_embeds = postprocess_func(outputs, text_inputs)
             if dtype is not None:
                 prompt_embeds = prompt_embeds.to(dtype=dtype)
+            batch_size = len(processed_text_list)
+            num_images_per_prompt = 1
```
zhaochenyang20: get it from the batch. Though right now alway have 1.

zhaochenyang20 (changes requested):

python/sglang/multimodal_gen/runtime/models/registry.py
```diff
@@ -33,7 +33,6 @@
 ]
 
 _IMAGE_ENCODER_MODELS: dict[str, tuple] = {
-    # "HunyuanVideoTransformer3DModel": ("image_encoder", "hunyuanvideo", "HunyuanVideoImageEncoder"),
```
alphabetc1, line 36: Why was this removed?

python/sglang/multimodal_gen/configs/pipeline_configs/stablediffusion3.py
gxlvera: My bad I will fix that….🥲

python/sglang/multimodal_gen/configs/pipeline_configs/stablediffusion3.py
gxlvera: is it really safe to import other model's function? What if someday we no longer support hunyuan...

python/sglang/multimodal_gen/configs/pipeline_configs/stablediffusion3.py
gxlvera: fixed

python/sglang/multimodal_gen/runtime/loader/utils.py
gxlvera: actually not entirely equivalent. The logic is that, except sd3, other models only have encoder (the index is 0) and encoder_2, so the earlier codes only normalize encoder_2. But sd3 has encoder_3, so I modified the codes to accommodate to encoder with any index.

python/sglang/multimodal_gen/runtime/models/dits/stablediffusion3.py
gxlvera: thanks! fixed

python/sglang/multimodal_gen/runtime/models/registry.py
gxlvera: sorry, it's supposed to be here. I have reverted it

python/sglang/multimodal_gen/runtime/pipelines/stable_diffusion_3.py
gxlvera: Nice catch! Fixed

python/sglang/multimodal_gen/configs/pipeline_configs/base.py
```diff
@@ -392,6 +394,23 @@ def post_denoising_loop(self, latents, batch):
...
+
+    def extract_pooled_output(self, encoder_index, encoder_outputs):
+        """Extract an optional pooled embedding from encoder outputs.
+
+        Return a tensor to append to `pooled_embeds_list`, or ``None`` to
+        skip.  Subclasses override to collect pooled outputs for specific
+        encoder indices.
+        """
+        return None
+
```
zhaochenyang20: These comments are too detailed. Only stress that "Subclasses can override to customise per-encoder masking behaviour
        (e.g. return ``None`` to disable masking for all encoders)."

And, make the comments concise.

python/sglang/multimodal_gen/configs/pipeline_configs/flux.py
```diff
@@ -86,6 +86,12 @@ class FluxPipelineConfig(ImagePipelineConfig):
         ]
     )
 
+    def extract_pooled_output(self, encoder_index, encoder_outputs):
+        """Flux v1 collects pooled output from the CLIP encoder (index 0)."""
+        if encoder_index == 0:
+            return encoder_outputs.pooler_output
+        return None
```
zhaochenyang20: else is missing.

zhaochenyang20 (changes requested):

python/sglang/multimodal_gen/configs/pipeline_configs/stablediffusion3.py
```diff
@@ -0,0 +1,137 @@
...
+from typing import Callable
+
+import torch
+
+from sglang.multimodal_gen.configs.models import DiTConfig, EncoderConfig, VAEConfig
+from sglang.multimodal_gen.configs.models.dits import StableDiffusion3TransformerConfig
+from sglang.multimodal_gen.configs.models.encoders import (
+    BaseEncoderOutput,
+    CLIPTextConfigForSD3,
+    T5ConfigForSD3,
```
Ratish1: These seem to be imported but I cant find the definitions for it?

python/sglang/multimodal_gen/registry.py
```diff
@@ -711,6 +717,12 @@ def _register_configs():
         hf_model_paths=["Qwen/Qwen-Image-Layered"],
         model_detectors=[lambda hf_id: "qwen-image-layered" in hf_id.lower()],
     )
+    register_configs(
+        sampling_param_cls=StableDiffusion3SamplingParams,
+        pipeline_config_cls=StableDiffusion3PipelineConfig,
+        hf_model_paths=["stabilityai/stable-diffusion-3-medium-diffusers"],
+        model_detectors=[lambda hf_id: "stable-diffusion-3" in hf_id.lower()],
```
Ratish1: should we instead do it like this?
```suggestion
        model_detectors=[lambda hf_id: "stable-diffusion-3-medium" in hf_id.lower()],
```

python/sglang/multimodal_gen/configs/pipeline_configs/stablediffusion3.py
gxlvera: Please refer to commit 1c7f7a0d774

gxlvera:
> There seems to be a CI failure related to this PR image



> There seems to be a CI failure related to this PR image

this has been fixed in commit 1c7f7a0

gxlvera:
> The isinstance checks in shared stages (text_encoding.py, denoising.py) are the biggest architectural issue. Every new pipeline will need more branches. The if i != 2 magic number is extremely fragile. The tokenizer kwargs should be fully config-driven — the fact that CLIPTextConfigForSD3.**post_init** sets tokenizer_kwargs but text_encoding.py overrides them again indicates a design conflict.

This has been fixed, thanks @zhaochenyang20

python/sglang/multimodal_gen/configs/pipeline_configs/stablediffusion3.py
```diff
@@ -0,0 +1,169 @@
...
+class StableDiffusion3PipelineConfig(SpatialImagePipelineConfig):
+    """Configuration for SD3 image generation pipeline.
+
+    This config intentionally relies on SD3-specific encoder configs to provide
+    tokenizer kwargs, instead of stage-level tokenizer overrides.
+    """
+
+    task_type: ModelTaskType = ModelTaskType.T2I
+
+    # Model configurations
```
zhaochenyang20: remove this comment.

python/sglang/multimodal_gen/configs/pipeline_configs/stablediffusion3.py
```diff
@@ -0,0 +1,169 @@
...
+        Callable[[BaseEncoderOutput, dict], torch.Tensor], ...
+    ] = field(
+        default_factory=lambda: (
+            sd3_clip_postprocess_text,
+            sd3_clip_postprocess_text,
+            t5_postprocess_text,
+        )
+    )
+
+    # SD3 specific parameters
```
zhaochenyang20: remove this comment

python/sglang/multimodal_gen/configs/sample/stablediffusion3.py
```diff
@@ -0,0 +1,20 @@
...
+
+from sglang.multimodal_gen.configs.sample.sampling_params import SamplingParams
+
+
+@dataclass
+class StableDiffusion3SamplingParams(SamplingParams):
+    """Sampling parameters for StableDiffusion3."""
+
+    # A single space ensures tokenizers produce valid (non-empty) input for CFG.
+    negative_prompt: str = " "
```
zhaochenyang20, line 16: why the default is `" "` but not `""`?

python/sglang/multimodal_gen/runtime/loader/component_loaders/text_encoder_loader.py
```diff
@@ -179,21 +180,30 @@ def load_customized(
...
+        if encoder_index >= num_encoder_configs:
+            raise IndexError(
+                f"Component '{component_name}' resolved to encoder index {encoder_index}, "
+                f"but only {num_encoder_configs} text_encoder_configs are provided."
+            )
+        if encoder_index >= num_encoder_precisions:
+            raise IndexError(
+                f"Component '{component_name}' resolved to encoder index {encoder_index}, "
+                f"but only {num_encoder_precisions} text_encoder_precisions are provided."
+            )
```
zhaochenyang20: These two `if -> raise` can be merged into one.

python/sglang/multimodal_gen/runtime/loader/component_loaders/text_encoder_loader.py
```diff
@@ -179,21 +180,30 @@ def load_customized(
...
 
-        # TODO(mick): had to throw an exception for different text-encoder arch
-        if not is_not_first_encoder(component_name):
```
zhaochenyang20, line 225: If this PR have not complete Mick's feedback, please keep his `TODO` here.

python/sglang/multimodal_gen/runtime/loader/component_loaders/text_encoder_loader.py
```diff
@@ -179,21 +180,30 @@ def load_customized(
...
-        else:
-            assert len(server_args.pipeline_config.text_encoder_configs) == 2
-            encoder_config = server_args.pipeline_config.text_encoder_configs[1]
-            encoder_config.update_model_arch(model_config)
-            encoder_dtype = server_args.pipeline_config.text_encoder_precisions[1]
+        encoder_config = server_args.pipeline_config.text_encoder_configs[encoder_index]
+        encoder_config.update_model_arch(model_config)
+
+        for key, value in diffusers_pretrained_config.__dict__.items():
+            setattr(encoder_config.arch_config, key, value)
```
zhaochenyang20: why should we use `setattr` here? Do not use dynamic attribute.

python/sglang/multimodal_gen/runtime/loader/component_loaders/vae_loader.py
```diff
@@ -81,9 +82,10 @@ def load_customized(
         vae_config = getattr(server_args.pipeline_config, pipeline_vae_config_attr)
         vae_precision = getattr(server_args.pipeline_config, pipeline_vae_precision)
         vae_config.update_model_arch(config)
-        if hasattr(vae_config, "post_init"):
+        post_init_fn = getattr(vae_config, "post_init", None)
+        if callable(post_init_fn):
             # NOTE: some post init logics are only available after updated with config
-            vae_config.post_init()
+            post_init_fn()
```
zhaochenyang20: Keep this part. Do not change it. I do not think this modify make sense.

python/sglang/multimodal_gen/runtime/loader/component_loaders/vae_loader.py
```diff
@@ -127,6 +129,12 @@ def load_customized(
             vae = vae_cls(vae_config).to(target_device)
 
         safetensors_list = _list_safetensors_files(component_model_path)
+
+        # Use pipeline-specific weight selection strategy if configured.
```
zhaochenyang20: remove the comments

python/sglang/multimodal_gen/runtime/loader/component_loaders/vae_loader.py
```diff
@@ -152,3 +160,34 @@ def load_customized(
...
+                )
+                target_files = [fp16_path] if os.path.exists(fp16_path) else []
+            else:
+                full_path = os.path.join(
+                    str(component_model_path), f"{base_name}.safetensors"
+                )
+                target_files = [full_path] if os.path.exists(full_path) else []
+            return target_files if target_files else safetensors_list
+
+        # Default: return all found safetensors files
```
zhaochenyang20: delete the comments here.

python/sglang/multimodal_gen/runtime/loader/utils.py
zhaochenyang20: good job.

python/sglang/multimodal_gen/runtime/models/dits/stablediffusion3.py
```diff
@@ -0,0 +1,180 @@
...
+        timestep: torch.LongTensor | None = None,
+        block_controlnet_hidden_states: list | None = None,
+        guidance: torch.Tensor | None = None,
+        joint_attention_kwargs: dict[str, Any] | None = None,
+        skip_layers: list[int] | None = None,
+    ) -> torch.Tensor:
+        if encoder_hidden_states is None:
+            raise ValueError("encoder_hidden_states must be provided.")
+        if pooled_projections is None:
+            raise ValueError("pooled_projections must be provided.")
```
zhaochenyang20, line 108: merge these two `if -> raise`.

python/sglang/multimodal_gen/runtime/models/encoders/clip.py
```diff
@@ -590,6 +590,51 @@ def load_weights(self, weights: Iterable[tuple[str, torch.Tensor]]) -> set[str]:
         return loaded_params
 
 
+class CLIPTextModelWithProjection(CLIPTextModel):
+    """
+    CLIP text encoder with projection head.
+
+    This mirrors transformers' CLIPTextModelWithProjection behavior for SD3:
```
zhaochenyang20: Do not use SD3 here. Please wirte in full name.

python/sglang/multimodal_gen/runtime/pipelines/stable_diffusion_3.py
```diff
@@ -0,0 +1,111 @@
...
+        by downstream stages (get_pos_prompt_embeds accesses index [0]).
+        """
+        if len(embeds_list) != 3:
+            raise ValueError(
+                f"SD3 requires exactly 3 prompt embedding tensors, got {len(embeds_list)}."
+            )
+        if len(pooled_list) < 2:
+            raise ValueError(
+                f"SD3 requires at least 2 pooled embedding tensors, got {len(pooled_list)}."
+            )
```
zhaochenyang20, line 55: 1. merge these two `if -> raise`.
2. I think it's better to use full name of SD3.

python/sglang/multimodal_gen/runtime/pipelines_core/stages/text_encoding.py
```diff
@@ -273,8 +275,11 @@ def encode_text(
                 prompt_embeds = prompt_embeds.to(dtype=dtype)
 
             embeds_list.append(prompt_embeds)
-            if is_flux_v1:
-                pooled_embeds_list.append(outputs.pooler_output)
```
zhaochenyang20: Are you sure? Are these changes really equivalent?

zhaochenyang20 (dismissed):

python/sglang/multimodal_gen/configs/pipeline_configs/stablediffusion3.py
gxlvera: fixed

python/sglang/multimodal_gen/configs/pipeline_configs/stablediffusion3.py
gxlvera: fixed

python/sglang/multimodal_gen/configs/sample/stablediffusion3.py
gxlvera: just as the comment indicates, we use " " as default to (not "") to avoid empty-input edge cases in CFG tokenization/encoding. In other words, i think it's just an engineering trick to make the CFG tokenization more robust.

python/sglang/multimodal_gen/runtime/loader/component_loaders/text_encoder_loader.py
gxlvera: make sense, modified as:
```
if (
            encoder_index >= num_encoder_configs
            or encoder_index >= num_encoder_precisions
        ):
            missing_targets: list[str] = []
            if encoder_index >= num_encoder_configs:
                missing_targets.append(
                    f"{num_encoder_configs} text_encoder_configs are provided"
                )
            if encoder_index >= num_encoder_precisions:
                missing_targets.append(
                    f"{num_encoder_precisions} text_encoder_precisions are provided"
                )
            raise IndexError(
                f"Component '{component_name}' resolved to encoder index {encoder_index}, "
                f"but only {' and '.join(missing_targets)}."
            )
```

python/sglang/multimodal_gen/runtime/loader/component_loaders/text_encoder_loader.py
gxlvera: indeed 😭 fixed, thanks!

python/sglang/multimodal_gen/configs/models/dits/stablediffusion3.py
```diff
@@ -0,0 +1,37 @@
+# SPDX-License-Identifier: Apache-2.0
+"""StableDiffusion3 Transformer model configuration"""
+
+from dataclasses import dataclass, field
+
+from sglang.multimodal_gen.configs.models.dits.base import DiTArchConfig, DiTConfig
+
+
+@dataclass
+class StableDiffusion3TransformerArchConfig(DiTArchConfig):
```
Rockdu, line 10: Since SD3 is a model family with many sizes, maybe we should specify that this conf denotes SD3-medium. Furthermore, we may need to think about how to support the whole model family elegantly

python/sglang/multimodal_gen/runtime/loader/component_loaders/text_encoder_loader.py
```diff
@@ -206,21 +207,38 @@ def load_customized(
...
-            encoder_config.update_model_arch(model_config)
-            for key, value in diffusers_pretrained_config.__dict__.items():
-                setattr(encoder_config.arch_config, key, value)
-            encoder_dtype = server_args.pipeline_config.text_encoder_precisions[0]
-        else:
-            assert len(server_args.pipeline_config.text_encoder_configs) == 2
-            encoder_config = server_args.pipeline_config.text_encoder_configs[1]
-            encoder_config.update_model_arch(model_config)
-            encoder_dtype = server_args.pipeline_config.text_encoder_precisions[1]
+        encoder_index = self._extract_encoder_index(component_name)
```
Rockdu, line 223: sacrificing the overall architecture for stable diffusion seems not an ideal choice, we need further discussion here

gxlvera: agreed, I will refactor this part to make it align with the origial implementation as much as possible i guess

python/sglang/multimodal_gen/runtime/pipelines_core/stages/text_encoding.py
```diff
@@ -282,8 +282,15 @@ def encode_text(
                 prompt_embeds = prompt_embeds.to(dtype=dtype)
 
             embeds_list.append(prompt_embeds)
+            is_sd3 = (
```
mickqian: avoid hardcode

python/sglang/multimodal_gen/runtime/loader/component_loaders/vae_loader.py
```diff
@@ -129,6 +132,19 @@ def load_customized(
             vae = vae_cls(vae_config).to(target_device)
 
         safetensors_list = _list_safetensors_files(component_model_path)
+        if (
```
mickqian: ditto

mickqian (commented):
TODO:
1. support combined cfg

python/sglang/multimodal_gen/runtime/loader/component_loaders/text_encoder_loader.py
```diff
@@ -219,21 +220,38 @@ def load_customized(
...
+                    f"{num_encoder_configs} text_encoder_configs are provided"
+                )
+            if encoder_index >= num_encoder_precisions:
+                missing_targets.append(
+                    f"{num_encoder_precisions} text_encoder_precisions are provided"
+                )
+            raise IndexError(
+                f"Component '{component_name}' resolved to encoder index {encoder_index}, "
+                f"but only {' and '.join(missing_targets)}."
+            )
```
Rockdu: we can squash this error log into a single line of assert

python/sglang/multimodal_gen/runtime/loader/component_loaders/text_encoder_loader.py
```diff
@@ -219,21 +220,38 @@ def load_customized(
...
+
+        encoder_config = server_args.pipeline_config.text_encoder_configs[encoder_index]
+        encoder_config.update_model_arch(model_config)
+
+        for key, value in diffusers_pretrained_config.__dict__.items():
```
Rockdu: let's align this with the original ver: only add these config to the first encoder

gxlvera: done

python/sglang/multimodal_gen/runtime/loader/component_loaders/text_encoder_loader.py
gxlvera: done

python/sglang/multimodal_gen/configs/models/dits/stablediffusion3.py
gxlvera: this class could be used by sd3/sd3.5-medium/large actually，i will add it to the docstring

python/sglang/multimodal_gen/runtime/pipelines_core/stages/text_encoding.py
gxlvera: I have deleted the hardcoded logics. I replaced hardcoded is_flux_v1/is_sd3 isinstance checks in TextEncodingStage with two PipelineConfig hooks, get_text_encoder_attention_mask and get_text_encoder_pooler_output added in base PipelineConfig, and they are overridden in FluxPipelineConfig, Flux2PipelineConfig, and StableDiffusion3PipelineConfig for model-specific handling.

python/sglang/multimodal_gen/runtime/pipelines_core/stages/text_encoding.py
```diff
@@ -282,8 +282,15 @@ def encode_text(
...
+                server_args.pipeline_config.__class__.__name__
+                == "StableDiffusion3PipelineConfig"
+            )
             if is_flux_v1:
                 pooled_embeds_list.append(outputs.pooler_output)
+            elif is_sd3 and i <= 1:
```
mickqian: could we use something like `use_pooler_output = is_sd3 or is_flux_v1`

mickqian (approved):

PR #19283, opened 2026-02-25
Model Layer Testing POC

akhilg-nv:
## Test Cost Comparison

sanity test with TP=8

`time python -m unittest registered.modeling.test_modeling_deepseek_v3 -k sanity`

real    1m1.416s
user    6m30.110s
sys     1m36.591s

`time python -m unittest registered.8-gpu-models.test_deepseek_v3_basic -v`

Initial Run:
real    22m18.176s
user    1m40.294s
sys     0m31.521s

Second run (model loads much faster, DeepGEMM JIT pre-compilation already complete):
real    2m26.941s
user    1m40.806s
sys     0m24.931s

## Checklist

## Review Process

### DIFF
--- a/test/registered/modeling/test_modeling_deepseek_v3.py
+++ b/test/registered/modeling/test_modeling_deepseek_v3.py
@@ -1,0 +1,153 @@
+import json
+import unittest
+
+import requests
+from parameterized import parameterized_class
+
+from sglang.srt.environ import envs
+from sglang.test.server_fixtures.default_fixture import DefaultServerBase
+
+MODEL = "deepseek-ai/DeepSeek-V3"
+MODEL_OVERRIDE_ARGS = json.dumps(
+    {
+        "num_hidden_layers": 2,
+        "first_k_dense_replace": 0,
+        "n_routed_experts": 8,
+    }
+)
+
+# Default values for all config dimensions
+DEFAULTS = {
+    "tp_size": 1,
+    "ep_size": 1,
+    "precision": "bf16",
+    "attention_backend": None,
+    "chunked_prefill": True,
+    "cuda_graph": False,
+}
+
+# Configs extended on top of defaults
+CONFIGS = [
+    # --- Initial baseline ---
+    {},
+    # --- Future configs ---
+    # {"tp_size": 2},
+    # {"tp_size": 2, "ep_size": 2},
+    # {"precision": "fp8"},
+    # {"tp_size": 2, "precision": "fp8"},
+    # {"attention_backend": "triton"},
+    # {"attention_backend": "fa3"},
+    # {"cuda_graph": True},
+    # {"chunked_prefill": False},
+]
+
+
+def _merge_config(overrides):
+    """Merge overrides with DEFAULTS."""
+    return {**DEFAULTS, **overrides}
+
+
+ALL_CONFIGS = [_merge_config(c) for c in CONFIGS]
+
+
+def _config_suffix(cfg):
+    """Generate a readable suffix from a config dict, showing only non-default values."""
+    parts = []
+    for key, val in cfg.items():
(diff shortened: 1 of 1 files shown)


test/registered/modeling/test_modeling_deepseek_v3.py
```diff
@@ -0,0 +1,303 @@
...
+import torch
+from parameterized import parameterized_class
+from transformers import AutoConfig, AutoModelForCausalLM, AutoTokenizer
+
+from sglang.srt.environ import envs
+from sglang.test.server_fixtures.default_fixture import DefaultServerBase
+
+MODEL = "deepseek-ai/DeepSeek-V3"
+MODEL_OVERRIDE_ARGS = {
+    "num_hidden_layers": 2,
```
nvpohanh: Can we run with 4 layers + `first_k_dense_replace` set to 2? That means, 2 MLP layers + 2 MoE layers so that we can cover the communications across different layer boundaries (MLP-MLP, MLP-MoE, MoE-MoE)? Thanks

test/registered/modeling/test_modeling_deepseek_v3.py
```diff
@@ -0,0 +1,303 @@
...
+    "first_k_dense_replace": 0,
+    "n_routed_experts": 8,
```
nvpohanh: Let's keep `n_routed_experts` `num_experts_per_tok` `n_group` `topk_group` as the original because lots of DSV3/DSR1 code paths have hardcoded check on these parameters.

test/registered/modeling/test_modeling_deepseek_v3.py
```diff
@@ -0,0 +1,303 @@
...
+        args.extend(["--attention-backend", cfg["attention_backend"]])
+
+    # Chunked prefill
+    if not cfg["chunked_prefill"]:
+        args.extend(["--chunked-prefill-size", "-1"])
+
+    # CUDA graph
+    if not cfg["cuda_graph"]:
+        args.append("--disable-cuda-graph")
+
```
nvpohanh, line 136: Let's also add:
- `--kv-cache-dtype fp8_e4m3` or no flag
- `--moe-dense-tp-size 1` or no flag

test/registered/modeling/test_modeling_deepseek_v3.py
```diff
@@ -0,0 +1,303 @@
...
+        "--trust-remote-code",
+        "--mem-fraction-static",
+        "0.5",
+    ]
+
+    if dummy_weights:
+        args.extend(["--load-format", "dummy"])
+
+    # Precision / quantization
+    precision = cfg["precision"]
```
nvpohanh: I think precision and quantization should be two separated fields because we could have different combinations like bf16+fp8 and fp16+fp8, etc. Usually, "precision" defines what the high-precision dtype part (like linear op output) should be and "quantization" defines what the low-precision dtype part (like linear op input) should be

test/registered/modeling/test_modeling_deepseek_v3.py
```diff
@@ -0,0 +1,303 @@
...
+                input_ids[0, i]
+            ].item()
+            diff = abs(sg_logprob - hf_token_logprob)
+            self.assertLess(
+                diff,
+                PREFILL_TOLERANCE,
+                f"Prefill logprob mismatch at position {i}: "
+                f"SGLang={sg_logprob:.6f}, HF={hf_token_logprob:.6f}, diff={diff:.6f}",
+            )
+
```
nvpohanh, line 450: Can we run one prefill step and one decode step? You can refer to the TRT-LLM test as an example: https://github.com/NVIDIA/TensorRT-LLM/blob/main/tests/unittest/_torch/modeling/test_modeling_llama.py#L207-L379

PR #19288, opened 2026-02-25
[PD-Disagg] Fix bootstrap server race condition when prefill workers not yet registered

hnyls2002:
## Summary

**Bootstrap server**
- Add `_is_ready()` check: track registered worker count, return 503 until all `dp_size * tp_size * pp_size` workers have registered
- Guard both metadata query (`-1,-1,-1`) and specific worker query with `_is_ready()`
- Catch `KeyError` on `prefill_port_table` lookup instead of crashing the server
- Remove unreachable `None` check after dict lookup (dead code)

**Decode client**
- Add retry logic in `ensure_parallel_info` (20 retries, 1s interval) so decode waits for prefill registration instead of failing immediately

## Error Log

From [CI failure](https://github.com/sgl-project/sglang/actions/runs/22377079442/job/64770066939?pr=19268):

```
File ".../sglang/srt/disaggregation/common/conn.py", line 675, in _handle_route_get
    info = PrefillServerInfo(
File ".../sglang/srt/disaggregation/common/conn.py", line 56, in __post_init__
    self.attn_tp_size = int(self.attn_tp_size)
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
```

## Root Cause

`CommonKVBootstrapServer` starts its HTTP server before any prefill worker has registered. Fields like `attn_tp_size` are `None` until a PUT arrives. A decode GET during this window crashes with `TypeError` in `PrefillServerInfo.__post_init__`.

This race always existed but was silently masked: previously `_handle_route_get` returned a plain dict (serializing `None` as JSON `null`), and the decode-side `int(None)` was swallowed by `except Exception`. After #19195 introduced `PrefillServerInfo` with `__post_init__`, the crash moved server-side and became visible.

## Repro Patch

Delay `register_to_bootstrap()` to widen the race window:

```python
def register_to_bootstrap(self):
    timer = threading.Timer(10, self._do_register_to_bootstrap)
    timer.daemon = True
    timer.start()
    return

def _do_register_to_bootstrap(self):
    # ... original body ...
```

## Test plan

- [ ] `test_disaggregation_basic.py` passes
- [ ] `test_disaggregation_dp_attention.py` passes

### DIFF
--- a/python/sglang/srt/disaggregation/common/conn.py
+++ b/python/sglang/srt/disaggregation/common/conn.py
@@ -135,13 +135,26 @@ class CommonKVManager(BaseKVManager):
         with self.failure_lock:
             self.failure_records[bootstrap_room] = failure_reason
 
-    def ensure_parallel_info(self, bootstrap_addr: str) -> bool:
+    def ensure_parallel_info(
+        self, bootstrap_addr: str, max_retries: int = 20, retry_interval: float = 1.0
+    ) -> bool:
         """Fetch and cache prefill parallel info if not yet available.
         Returns True if info is available (cached or freshly fetched).
+        Retries with backoff if the prefill server hasn't registered yet.
         """
         if bootstrap_addr in self.prefill_info_table:
             return True
-        info = self._fetch_prefill_server_info(bootstrap_addr)
+        info = None
+        for attempt in range(max_retries):
+            info = self._fetch_prefill_server_info(bootstrap_addr)
+            if info is not None:
+                break
+            if attempt < max_retries - 1:
+                logger.info(
+                    f"Prefill server info not available from {bootstrap_addr}, "
+                    f"retrying ({attempt + 1}/{max_retries})..."
+                )
+                time.sleep(retry_interval)
         if info is None:
             return False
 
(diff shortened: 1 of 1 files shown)


python/sglang/srt/disaggregation/common/conn.py
```diff
@@ -606,6 +598,12 @@ def __init__(self, host: str, port: int, dp_size: int = 1):
     def run(self):
         self.thread.start()
 
+    def _is_ready(self) -> bool:
+        if self.attn_tp_size is None or self.pp_size is None:
+            return False
+        expected = self.dp_size * self.attn_tp_size * self.pp_size
```
ShangmingCai, line 606: After this PR: https://github.com/sgl-project/sglang/pull/17213, we might need to take cp size into account in the future.

ShangmingCai (approved):
LGTM

python/sglang/srt/disaggregation/common/conn.py
hnyls2002: ~~We actually do not need to wait until all workers are registered. This is a guard for fetching the prefill's parallel info. For the bootstrap info fetching, the guard happened under this if branch.~~

Yes, we need to calculate how many workers are needed for this prefill server here.

PR #19290, opened 2026-02-25
feat: [2/2][DeepEP] Add waterfill load balancing for shared expert dispatch

xutizhou:
## Motivation

In DeepSeek V3/R1 with expert parallelism (EP), each rank processes a subset of routed experts plus the **full** shared expert. Since every token must visit the shared expert, it creates a fixed compute load on all ranks regardless of routed expert distribution. When routed expert load is imbalanced across ranks (common with skewed token distributions), the shared expert amplifies the bottleneck on already-overloaded ranks.
**Waterfill** addresses this by treating the shared expert as an additional routed expert (the "9th expert" per rank) and dispatching it to the **least-loaded** rank via DeepEP, effectively filling the load gap like water filling a container. This converts the shared expert from a fixed per-rank cost into a dynamic load-balancing lever.

coauthor: @AichenF 

Thanks to @ch-wan  for the discussion and suggestions.

This is the second half of the waterfill feature split (pending on https://github.com/sgl-project/sglang/pull/20089):

### Waterfill balancing algorithm

Waterfill treats the fused shared expert as one extra DeepEP dispatch slot per token.
It does not change the routed top-k experts selected by the router. It only chooses
which EP rank should execute the shared-expert slot.

#### Load score

For a source rank $s$, let $\mathcal{T}^{(s)}$ be the local tokens currently
held by that rank. Let $\tau_r = |\mathcal{T}^{(r)}|$ be source rank $r$'s local
batch size.

Waterfill first counts how many routed top-k slots land on each destination rank:

$$
A^{\mathrm{static}}_r =
\sum_{i \in \mathcal{T}^{(s)}} \sum_{k=1}^{K}
\mathbf{1}[\mathrm{rank}(e_{i,k}) = r]
$$

Static mode does not run a cross-rank collective, so it uses this local routed-load
view directly:

$$
L_r = A^{\mathrm{static}}_r
$$

Dynamic mode runs one EP-group collective. It collects the global routed-load vector
and the per-rank batch-size vector:

$$
A^{\mathrm{global}}_r =
\operatorname{AllReduceSum}(A^{\mathrm{static}}_r),
\qquad
\tau_r = |\mathcal{T}^{(r)}|
$$

The current SGLang implementation uses the following dynamic load score:

$$
L_r = A^{\mathrm{global}}_r + \tau_r
$$

#### Waterline

Each participating token adds one shared-expert slot. Therefore the number of new
shared-expert slots is:

$$
N =
\begin{cases}
|\mathcal{T}^{(s)}|, & \text{static mode} \\
\sum_r \tau_r, & \text{dynamic mode}
\end{cases}
$$

Waterfill computes a target waterline after adding these shared-expert slots:

$$
H = \left\lceil\frac{\sum_r L_r + N}{R}\right\rceil
$$

A rank below this waterline has room to absorb shared-expert work:

$$
S_r = \max(H - L_r, 0)
$$

#### Candidate ranks

The communication-conservative candidate set for token $i$ is:

$$
C_i = \{s\} \cup \{\mathrm{rank}(e_{i,k}) \mid k=1,\ldots,K\}
$$

This keeps the shared-expert slot on ranks already touched by the token's routed
top-k, plus the source rank as a fallback.

SGLang also has an all-rank mode. Static mode currently sets
`allow_all_ranks=True`, so it may choose any EP rank. Dynamic mode also allows all
ranks when `max_effective_load <= target_total`; otherwise it uses the conservative
candidate set. All-rank selection can improve balancing freedom, but it may add a
new per-token dispatch destination, so it should be treated as a communication
tradeoff.

#### Rank selection

For each candidate rank, Waterfill uses its slack as the sampling weight. The source
rank keeps full weight; non-source ranks are discounted by the local preference factor
${1}/{1.1}$:

$$
w_{i,r} =
\mathbf{1}[r \in C_i] \cdot S_r \cdot
\begin{cases}
1, & r=s \\
1/1.1, & r\ne s
\end{cases}
$$

If the total weight is positive, the implementation deterministically samples a
shared-expert rank from these normalized weights using a token-based seed.

If every candidate has zero slack, Waterfill falls back to a local-preference rule:
it starts from the source rank and only switches to another candidate rank when that
rank is clearly lighter. With the current `1.1` preference factor, a remote rank must
be about 10% lower-load before it replaces the current best rank.

#### Output

The output is one selected shared-expert rank per token. The routed top-k expert
choices and weights stay semantically unchanged. The adapter only appends the shared
expert column and materializes the selected rank into the DeepEP-expanded TopK layout.

## Modifications
**New file: `python/sglang/srt/layers/moe/deepep_waterfill.py` (514 lines)**
- Triton kernel `_count_routed_per_rank_kernel`: counts routed tokens per EP rank from topk_ids.
- Triton kernel `_waterfill_kernel`: assigns each token's shared expert to the least-loaded rank using a waterfill algorithm, with local-rank preference (1.1x bias).
- `DeepEPWaterfillBalancer` class: orchestrates the waterfill dispatch with two modes:
  - **Static mode**: Uses precomputed `rank_load` from EPLB expert distribution data (no runtime all_reduce).
  - **Dynamic mode**: Falls back to runtime all_reduce of per-rank routed counts when EPLB data is unavailable.
- `expand_topk()`: Expands topk output by one column (shared expert slot) with computed expert IDs and scaling factors.

**Modified files:**
- `models/deepseek_v2.py` (+112/-21): Creates `DeepEPWaterfillBalancer` in MoE init; calls `expand_topk()` after TopK selection; skips separate shared expert forward when waterfill is enabled (shared expert is now part of MoE dispatch/compute/combine).
- `layers/moe/fused_moe_triton/layer.py` (+40/-2): Adjusts weight loading to map checkpoint expert IDs correctly when waterfill expands the expert count by `ep_size` (one shared slot per rank).
- `eplb/expert_location.py` (+33/-1): Adds `rank_load` field to `ExpertLocationMetadata` and `_compute_rank_load()` to derive per-rank load from logical expert counts + physical-to-logical mapping.
- `server_args.py` (+29/-1): Adds `--enable-deepep-waterfill` flag with validation.
 ## Accuracy Tests
MMLU accuracy on DeepSeek-V3 (FP8, 2-node 16xH20, EP16, DP16):

| Configuration | MMLU Accuracy |
|---|---|
| Baseline (no waterfill) | 91.55% |
| Waterfill + static EPLB | **91.80%** |

No accuracy degradation observed.
 
 ## Benchmarking and Profiling
Throughput benchmark on DeepSeek-V3 (FP8, 2-node 16×H20, EP16, DP16, input_len≈~600 MMLU prompts, output_len=1, 256 concurrent requests × multiple rounds, 1000 prompts / round):

| Configuration | Output Throughput (tok/s) | Gain |
|---|---|---|
| Baseline + static EPLB (no waterfill) | ~29,700 | — |
| Waterfill + static EPLB | ~30,900 | **+4%** |

waterfill
```python
    --model-path /lustre/raplab/client/xutingz/workspace/model/DeepSeek-V3 \
    --tp 16 --dp-size 16 --nnodes 2 --node-rank {0|1} \
    --dist-init-addr 10.6.131.5:{PORT} \
    --host 0.0.0.0 --port 30000 --trust-remote-code \
    --moe-a2a-backend deepep --deepep-mode normal \
    --enable-dp-attention --mem-fraction-static 0.75 --max-running-requests 2048 \
    --watchdog-timeout 1800 --disable-radix-cache --disable-cuda-graph \
    --chunked-prefill-size -1 \
    --enable-deepep-waterfill \
    --init-expert-location /lustre/raplab/client/xutingz/workspace/bench/waterfill/mmlu_expert_dist/ep16_mmlu_logical_count.pt
```
Baseline
```python
    --model-path /lustre/raplab/client/xutingz/workspace/model/DeepSeek-V3 \
    --tp 16 --dp-size 16 --nnodes 2 --node-rank {0|1} \
    --dist-init-addr 10.6.131.5:{PORT} \
    --host 0.0.0.0 --port 30000 --trust-remote-code \
    --moe-a2a-backend deepep --deepep-mode normal \
    --enable-dp-attention --mem-fraction-static 0.75 --max-running-requests 2048 \
    --watchdog-timeout 1800 --disable-radix-cache --disable-cuda-graph \
    --chunked-prefill-size -1 \
    --init-expert-location /lustre/raplab/client/xutingz/workspace/bench/waterfill/mmlu_expert_dist/ep16_mmlu_logical_count.pt
```
## Checklist

## Review Process


python/sglang/srt/server_args.py
```diff
@@ -2214,6 +2215,17 @@ def _handle_a2a_moe(self):
             logger.warning(
                 f"DeepEP MoE is enabled. The expert parallel size is adjusted to be the same as the tensor parallel size[{self.tp_size}]."
             )
+            if self.enable_deepep_waterfill:
```
ch-wan, line 3147: Ensure `self.disable_shared_experts_fusion` is False here.

python/sglang/srt/models/deepseek_v2.py
ch-wan: The logic here is getting complicated and hard to maintain. We can modify it like this:
```python
disable_reason = None
if get_global_server_args().enable_deepep_waterfill:
    disable_reason = None
elif (
```

python/sglang/srt/layers/moe/deepep_waterfill.py
```diff
@@ -0,0 +1,514 @@
...
+        _LOCAL_PREF_DENOM,
+        target_total,
+        allow_all_ranks,
+        BLOCK_SIZE,
+    )
+
+    return expanded_topk_ids, expanded_topk_weights, local_shared_mask
+
+
+def expand_topk_with_shared_expert(
```
ch-wan, line 325: This function has several small kernels. Can we use `torch.compile` here?

python/sglang/srt/eplb/expert_location.py
```diff
@@ -564,9 +588,16 @@ def compute_initial_expert_location_metadata(
         logger.info(
             "init_expert_location from init_by_eplb using ServerArgs.init_expert_location"
         )
-        return ExpertLocationMetadata.init_by_eplb(
+        metadata = ExpertLocationMetadata.init_by_eplb(
             server_args, model_config, logical_count=data_dict["logical_count"]
         )
+        if metadata is not None:
```
ch-wan:
```suggestion
        if metadata is not None and get_global_server_args().enable_deepep_waterfill:
```

python/sglang/srt/eplb/expert_location.py
```diff
@@ -531,6 +533,28 @@ def from_model_config(model_config: ModelConfig):
             return None
 
 
+def _compute_rank_load(logical_count_raw, physical_to_logical_map, ep_size):
+    """Compute per-rank load (num_layers, ep_size) from logical counts and EPLB mapping."""
+    from sglang.srt.eplb.expert_distribution import compute_gpu_physical_count
+
+    if not isinstance(logical_count_raw, torch.Tensor):
```
ch-wan: How can `logical_coount_raw` has different data types / shapes? Add some comments here to explain this.

python/sglang/srt/layers/moe/fused_moe_triton/layer.py
```diff
@@ -555,6 +555,32 @@ def _map_global_expert_id_to_local_expert_id(self, expert_id: int) -> int:
         num_local_routed_experts = (
             self.num_local_experts - self.num_fused_shared_experts
         )
+
+        # Waterfill expands num_experts by ep_size (one shared slot per rank).
+        # Checkpoint IDs use the ORIGINAL layout, so we must map using
+        # old_experts_per_rank to avoid loading experts onto wrong EP ranks.
+        if (
+            get_global_server_args().enable_deepep_waterfill
+            and get_moe_a2a_backend().is_deepep()
```
ch-wan: all `get_moe_a2a_backend().is_deepep()` are redundant because it is ensured in server_args' init when waterfill is enabled.

python/sglang/srt/layers/moe/fused_moe_triton/layer.py
```diff
@@ -555,6 +555,32 @@ def _map_global_expert_id_to_local_expert_id(self, expert_id: int) -> int:
...
+                )
+                # Routed experts: map using original experts_per_rank
+                start_idx = self.moe_ep_rank * old_num_local_routed_experts
+                end_idx = (self.moe_ep_rank + 1) * old_num_local_routed_experts
+                if start_idx <= expert_id < end_idx:
+                    return expert_id - start_idx
+                # Shared expert: maps to old_num_local_routed_experts on ALL ranks
+                if expert_id >= old_num_global_routed_experts:
+                    return old_num_local_routed_experts
+                return -1
```
ch-wan: try reuse the excerpt below

python/sglang/srt/layers/moe/deepep_waterfill.py
```diff
@@ -0,0 +1,514 @@
+# Copyright 2023-2024 SGLang Team
```
ch-wan: 2026

python/sglang/srt/layers/moe/deepep_waterfill.py
```diff
@@ -0,0 +1,514 @@
...
+            total_tokens_global_t = total_routed_t // topk
+            total_effective_t = effective_load.sum()
+            max_effective_t = effective_load.max()
+            target_total = int(
+                (
+                    (total_effective_t + total_tokens_global_t + self.world_size - 1)
+                    // self.world_size
+                ).item()
+            )
+            allow_all_ranks = bool((max_effective_t <= target_total).item())
```
ch-wan: we can avoid these `item()`

python/sglang/srt/layers/moe/deepep_waterfill.py
```diff
@@ -0,0 +1,514 @@
...
+            return expand_topk_with_shared_expert(
+                topk_ids,
+                topk_weights,
+                self.num_routed_experts,
+                self.world_size,
+                self.rank,
+                self.shared_weight,
+            )
+
+        routed_counts_i64 = routed_counts.to(torch.int64)
```
ch-wan: is this necessary?

python/sglang/srt/layers/moe/deepep_waterfill.py
```diff
@@ -0,0 +1,514 @@
...
+        self.shared_weight = (
+            1.0 / routed_scaling_factor if routed_scaling_factor != 0 else 1.0
+        )
+        self.static_rank_load: Optional[Tensor] = None
+        self._counts_buf: Optional[Tensor] = None
+        self._eplb_map_data_ptr = None
+
+    def update_static_weights(self):
+        """Update static weights from EPLB metadata if layout changes."""
+        if os.environ.get("SGLANG_DISABLE_STATIC_WATERFILL", "0") == "1":
```
ch-wan: use `envs`

ch-wan (commented):
I left some comments. Also, is it possible to separate this PR into two? One for fusing shared experts under EP, and the other for waterfill.

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -830,9 +866,18 @@ def forward_deepep(
                     layer_id=self.layer_id,
                 ),
             )
+
+            # Waterfill: expand topk from 8 to 9 columns, routing shared expert
+            # to the least-loaded rank instead of the home rank.
+            if self._enable_deepep_waterfill:
+                topk_output = self.deepep_waterfill_balancer.expand_topk(
```
ch-wan, line 873: We'd better move this computation to `self.topk()`

python/sglang/srt/server_args.py
```diff
@@ -2877,6 +2878,22 @@ def _handle_a2a_moe(self):
...
```
ch-wan, line 2881: We can also set `enforce_shared_experts_fusion` here.

python/sglang/srt/models/deepseek_v2.py
ch-wan, line 1: For this feature, we'd better avoid modifying model files, as we may need to extend this feature to other models.

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -813,7 +845,11 @@ def forward_deepep(
         if hidden_states.shape[0] > 0:
             # router_logits: (num_tokens, n_experts)
             router_logits = self.gate(hidden_states, forward_batch=forward_batch)
-            if not sbo_enabled_flag and self.num_fused_shared_experts == 0:
+            if (
+                not sbo_enabled_flag
+                and not self._enable_deepep_waterfill
```
ch-wan, line 850: Is this true?
```
not _enable_deepep_waterfill => num_fused_shared_experts == 0
```

If yes, we don't need this hange.

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -830,9 +866,18 @@ def forward_deepep(
...
+            # Waterfill: expand topk from 8 to 9 columns, routing shared expert
+            # to the least-loaded rank instead of the home rank.
+            if self._enable_deepep_waterfill:
+                topk_output = self.deepep_waterfill_balancer.expand_topk(
+                    topk_output, hidden_states.shape[0]
+                )
         else:
             topk_output = self.topk.empty_topk_output(hidden_states.device)
-            if is_deepep_class_backend() and self.num_fused_shared_experts > 0:
+            if self._enable_deepep_waterfill:
```
ch-wan, line 878: similar here

python/sglang/srt/environ.py
```diff
@@ -395,6 +395,7 @@ class Envs:
     SGLANG_DEEPEP_NUM_MAX_DISPATCH_TOKENS_PER_RANK = EnvInt(128)
     SGLANG_DEEPEP_LL_COMBINE_SEND_NUM_SMS = EnvInt(32)
     SGLANG_BLACKWELL_OVERLAP_SHARED_EXPERTS_OUTSIDE_SBO = EnvBool(False)
+    SGLANG_DISABLE_STATIC_WATERFILL = EnvBool(False)
```
ch-wan, line 398: when should we set this env variable?

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -801,6 +830,9 @@ def forward_deepep(
         hidden_states: torch.Tensor,
         forward_batch: ForwardBatch,
     ) -> torch.Tensor:
+        if self._enable_deepep_waterfill:
+            self.deepep_waterfill_balancer.update_static_weights()
```
ch-wan, line 834: Can we follow EPLB's logic? They both need to update stats

python/sglang/srt/server_args.py
```diff
@@ -5404,6 +5421,16 @@ def add_cli_args(parser: argparse.ArgumentParser):
             "(e.g., --mooncake-ib-device mlx5_0,mlx5_1). "
             "Default is None, which triggers automatic device detection when Mooncake Backend is enabled.",
         )
+        parser.add_argument(
+            "--enable-deepep-waterfill",
```
ch-wan, line 5425: update server_arg.md

python/sglang/srt/server_args.py
xutizhou: --enable-deepep-waterfill is the user-facing switch. When it is enabled, it automatically enforces shared expert fusion because Waterfill relies on the fused
  shared expert layout.

python/sglang/srt/environ.py
xutizhou: when we use dynamic waterfill.

python/sglang/srt/layers/moe/topk.py
```diff
@@ -257,6 +255,7 @@ def __init__(
         num_expert_group: Optional[int] = None,
         renormalize: bool = True,
         num_fused_shared_experts: int = 0,
+        enable_deepep_waterfill: bool = False,
```
xutizhou: pass a class

python/sglang/srt/layers/moe/topk.py
```diff
@@ -290,6 +308,46 @@ def __init__(
...
+                        "DeepEP waterfill cannot infer num_routed_experts before "
+                        "the first non-empty TopK forward."
+                    )
+                self._deepep_waterfill_num_routed_experts = router_logits.shape[-1]
+
+            from sglang.srt.layers.moe.deepep_waterfill import (
+                DeepEPWaterfillBalancer,
+            )
+
+            self._deepep_waterfill_balancer = DeepEPWaterfillBalancer(
```
xutizhou: move to model runner

docs/advanced_features/server_arguments.md
```diff
@@ -332,6 +332,7 @@ Please consult the documentation below and [server_args.py](https://github.com/s
 | `--elastic-ep-backend` | Specify the collective communication backend for elastic EP. Currently supports 'mooncake'. | `none` | `none`, `mooncake` |
 | `--enable-elastic-expert-backup` | Enable elastic EP backend to backup expert weights in DRAM feature. Currently supports 'mooncake'.| `False` | bool flag (set to enable) |
 | `--mooncake-ib-device` | The InfiniBand devices for Mooncake Backend transfer, accepts multiple comma-separated devices (e.g., --mooncake-ib-device mlx5_0,mlx5_1). Default is None, which triggers automatic device detection when Mooncake Backend is enabled. | `None` | Type: str |
+| `--enable-deepep-waterfill` | Enable DeepEP Waterfill: dispatch the shared expert as the 9th routed expert to the least-loaded EP rank. Requires `--moe-a2a-backend deepep`, implicitly enables shared-expert fusion, and supports `--deepep-mode auto` or `normal`. Use `auto` for production decode so CUDA graph remains enabled. Supported on DeepSeek-V3/R1 with EP >= 2. By default, Waterfill uses static EPLB rank-load metadata when available; set `SGLANG_DISABLE_STATIC_WATERFILL=1` to force dynamic Waterfill with runtime EP all-reduce. | `False` | bool flag (set to enable) |
```
xutizhou: low latency

ch-wan: how about `low_latency`?

python/sglang/srt/eplb/expert_location.py
```diff
@@ -177,14 +180,21 @@ def init_by_eplb(
...
-        return ExpertLocationMetadata._init_raw(
+        metadata = ExpertLocationMetadata._init_raw(
             server_args=server_args,
             ep_size=common["ep_size"],
             physical_to_logical_map=physical_to_logical_map.to(server_args.device),
             logical_to_all_physical_map=logical_to_all_physical_map.to(
                 server_args.device
             ),
         )
+        if metadata is not None and server_args.enable_deepep_waterfill:
```
ch-wan: add a NOTE here

python/sglang/srt/eplb/expert_location.py
```diff
@@ -177,14 +180,21 @@ def init_by_eplb(
...
+            metadata.rank_load = _compute_rank_load(
```
xutizhou: add a note here , initial .pt or map

python/sglang/srt/eplb/expert_location.py
```diff
@@ -534,6 +544,30 @@ def from_model_config(model_config: ModelConfig):
             return None
 
 
+def _compute_rank_load(logical_count_raw, physical_to_logical_map, ep_size):
+    """Compute per-rank load (num_layers, ep_size) from logical counts and EPLB mapping."""
+    from sglang.srt.eplb.expert_distribution import compute_gpu_physical_count
+
+    # logical_count_raw comes from data_dict["logical_count"] loaded from .pt/.json:
+    # it may be Tensor/list, and shape may be [layers, experts] or [samples, layers, experts].
+    if not isinstance(logical_count_raw, torch.Tensor):
```
xutizhou: rm

ch-wan: it's always a tensor

python/sglang/srt/layers/moe/topk.py
```diff
@@ -274,6 +273,22 @@ def __init__(
...
+            if num_fused_shared_experts != 1:
+                raise ValueError("DeepEP waterfill expects exactly one shared expert.")
+            if layer_id is None:
+                raise ValueError("DeepEP waterfill requires layer_id.")
+            if not _is_cuda:
+                raise ValueError("DeepEP waterfill TopK is only supported on CUDA.")
+            if output_format not in (None, TopKOutputFormat.STANDARD):
+                raise ValueError("DeepEP waterfill requires STANDARD TopK output.")
+            # Waterfill appends the shared expert after routed TopK selection.
+            top_k -= num_fused_shared_experts
```
ch-wan, line 302:
``` suggestion
           waterfill_topk = top_k - num_fused_shared_experts
```

python/sglang/srt/layers/moe/topk.py
```diff
@@ -274,6 +273,22 @@ def __init__(
...
+        self.deepep_waterfill_balancer = None
+        if enable_deepep_waterfill:
+            if num_fused_shared_experts != 1:
+                raise ValueError("DeepEP waterfill expects exactly one shared expert.")
+            if layer_id is None:
+                raise ValueError("DeepEP waterfill requires layer_id.")
+            if not _is_cuda:
+                raise ValueError("DeepEP waterfill TopK is only supported on CUDA.")
+            if output_format not in (None, TopKOutputFormat.STANDARD):
+                raise ValueError("DeepEP waterfill requires STANDARD TopK output.")
```
xutizhou: rm

python/sglang/srt/layers/moe/topk.py
```diff
@@ -274,6 +273,22 @@ def __init__(
...
+            # Waterfill appends the shared expert after routed TopK selection.
+            top_k -= num_fused_shared_experts
+            num_fused_shared_experts = 0
+            output_format = TopKOutputFormat.STANDARD
+
```
xutizhou, line 305: todo refactor topk. shared fusion + topk fusion

python/sglang/srt/layers/moe/topk.py
```diff
@@ -274,6 +273,22 @@ def __init__(
             assert num_expert_group is not None and topk_group is not None
 
         self.layer_id = layer_id
+        self.enable_deepep_waterfill = enable_deepep_waterfill
+        self.deepep_waterfill_balancer = None
+        if enable_deepep_waterfill:
+            if num_fused_shared_experts != 1:
+                raise ValueError("DeepEP waterfill expects exactly one shared expert.")
```
ch-wan: redundant assertion

python/sglang/srt/layers/moe/topk.py
ch-wan, line 302: A refactor is needed. Leave a todo to cheng wan.

python/sglang/srt/model_executor/model_runner.py
```diff
@@ -1398,6 +1388,55 @@ def load_model(self):
                     f"TP rank {self.tp_rank} could finish the model loading, but there are other ranks that didn't finish loading. It is likely due to unexpected failures (e.g., OOM) or a slow node."
                 ) from None
 
+    def _prepare_moe_topk(self):
+        from sglang.srt.layers.moe.topk import TopK
```
ch-wan: move to the top of this file

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -2233,19 +2223,10 @@ def determine_num_fused_shared_experts(
...
-                "DeepEP shared expert fusion: fusing shared expert into MoE kernel "
-                "at home EP rank local slot (--enforce-shared-experts-fusion).",
-            )
-            self.num_fused_shared_experts = self.config.n_shared_experts
-            return
-
-        # Check all conditions that disable fusion.
         disable_reason = None
-        if is_sbo_enabled() or is_tbo_enabled():
+        if is_deepep_class_backend() and server_args.enforce_shared_experts_fusion:
```
xutizhou: rm is_deepep_class_backend()

ch-wan: only keep `enforce_xxx`

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -389,6 +384,13 @@ def __init__(
             is_deepep_class_backend() and self.num_fused_shared_experts > 0
         )
 
+        # DeepEP waterfill: shared expert dispatched to least-loaded rank via DeepEP.
+        # Uses same expert layout as fusion but routes shared expert dynamically
+        # instead of always sending to home rank.
+        _enable_deepep_waterfill = (
+            get_global_server_args().enable_deepep_waterfill and _is_deepep_fusion
```
xutizhou: rm

python/sglang/srt/models/deepseek_v2.py
```diff
@@ -461,6 +463,7 @@ def __init__(
             use_grouped_topk=True,
             num_expert_group=config.n_group,
             num_fused_shared_experts=self.num_fused_shared_experts,
+            enable_deepep_waterfill=_enable_deepep_waterfill,
```
xutizhou: rm it

python/sglang/srt/models/deepseek_v2.py
ch-wan: remove it

xutizhou: guard at server_args

python/sglang/srt/server_args.py
```diff
@@ -2984,6 +2878,23 @@ def _handle_a2a_moe(self):
...
+                    )
+                    self.disable_shared_experts_fusion = False
+                self.enforce_shared_experts_fusion = True
+                logger.info(
+                    "DeepEP Waterfill is enabled. Shared expert will be dispatched through DeepEP for load balancing."
+                )
+
+        # Validate enable_deepep_waterfill requires deepep backend
+        if self.enable_deepep_waterfill and self.moe_a2a_backend != "deepep":
+            raise ValueError(
```
ch-wan: do not raise error, modify a2a backend and throw warning

xutizhou: motify to deepep and warning

ch-wan (approved):

PR #19320, opened 2026-02-25
[HiCache] Optimize L2 mem allocation when cache miss in L3

beef9999:
## Motivation

When enabling HiCache, every L1 hit will trigger a write from L1 -> L2 -> L3, which leads to the L2 memory full at all times.

As a result, each `prefetch` operation almost inevitably triggers an `evict_host`. If the data does not exist in L3, this newly allocated L2 memory will be wasted.

This PR moves the host memory allocation from before `prefetch` submission, to after L3 `batch_exists` lookup completion. 

**The goal is to avoid unnecessary L2 evict, and improve end-to-end cache hit ratio, when L3 miss happens.**

## Modifications

- `cache_controller.py`
- `hiradix_cache.py`

Add a `prefetch_hit_queue` between the prefetch thread and the main scheduler thread, and delegate memory allocation to the main thread, since only the scheduler thread can invoke evict_host.

This queue is used to record `storage_hit_count`, which is the longest prefix return value of L3 `batch_exists`.

The actual mem allocation will be delayed to `drain_storage_control_queues` by the size of `storage_hit_count`.

### Test architecture

PD-Disaggregation, 2p 2d

### Test scenario

- Multi-round Test. 
- Set L2 --hicache-ratio=2. Total Size = 600GB
- Set L3 capacity to 800 GB. 
- Adjust the test parameter, make the total amount of write data to be 1.2 TB. So we can see L3 cache miss in the last few rounds.

### Test parameters

```python3 benchmark_code_agent.py --num-clients 512 --num-rounds 10 --request-rate 10 --max-parallel 100 --url <url> --model-path <model> --dataset-path <dataset> --disable-auto-run --disable-random-sample --enable-round-barrier```

### Result

| Mode | Cache Hit Rate | AVG TTFT | P90 TTFT | P99 TTFT |
|------|----------------|----------|----------|----------|
| Only L2 | 26% | 3.49 | 14.13 | 21.29 |
| L2 + L3 | 41% | 3.83 | 13.39 | 23.42 |
| L2 + L3 (Apply this patch) | 55% | 2.34 | 6.56 | 18.44 |

## Checklist

## Review Process

---
### CI States

Latest PR Test (Base): :x: [Run #29532771733](https://github.com/sgl-project/sglang/actions/runs/29532771733)
Latest PR Test (Extra): :hourglass_flowing_sand: [Run #29533823686](https://github.com/sgl-project/sglang/actions/runs/29533823686)

### CODE REVIEW

**Comment by beef9999** [comment]
@xiezhq-hermann @pansicheng Please take a look at this PR

**Comment by beef9999** [comment]
> Thanks for this PR. There’s one thing I’m not quite clear about: even if the L3 fetch fails, the free memory obtained by evicting from the host will soon be used by the next request, so it wouldn’t actually be wasted. Why does this PR lead to such a large improvement in cache hit rate (41% → 55%)?

Because a lot of L3 batch_exists query on the fly (cost milliseconds) will cause L2 to evict the equivalent amount of memory.

BTW, as to this PR, I'm still investigating on some boundary cases, will re-new the information later.

**Comment by xiezhq-hermann** [comment]
The changes in this PR look sound to me, and the motivation is clear. That said, I'm not yet convinced about the trade-off between the extra scheduler cycle introduced by deferring allocation until after the batch_exists check, versus the host memory that would otherwise be speculatively consumed.

Could you share some more detailed quantitative analysis? In particular, I'd find the following helpful:

1. Latency breakdown: the added delay between batch_exists completion and the actual KV fetch start, and how it varies with scheduler tick frequency and batch size.
2. Sensitivity sweeps: results across different L2/L3 ratios and L3 hit rates — especially regimes where L3 hit rate is high (where the old path rarely wasted allocations) to confirm there's no regression.
3. Eviction frequency: how often evict_host fires before vs. after the patch, to quantify the memory-pressure relief. The current benchmark convincingly shows the win in the L3-miss regime, but a fuller picture would help rule out regressions in the other operating points.

**Comment by zxpdemonio** [comment]
We discovered this exact issue last week while running multi-turn TTFT stability benchmarks with hierarchical cache (L1 GPU → L2 Host → L3 Mooncake Store over RDMA).
                                                                                                                                                                                     
  We applied this PR (combined with #26449) as a patch to our test environment and the results are significant:
                                                                                                                                                                                     
  Setup: Qwen3-8B, TP=2, hicache-ratio=1.5, 50 sessions × 10 turns (800 requests), prefetch_threshold=64
                                                                                                                                                                                     
  Key Results:                                                                                                                                                                       
                                                            
| Metric | Before | After | Change |                                                                                                                                               
|--------|--------|-------|--------|   
| Prefetch BLOCKED events | 414 | **0** | Eliminated |                                                                                                                             
| TTFT avg | 1.20s | 0.44s | **-63.5%** |                                                                                                                                          
| TTFT p99 | 7.23s | 2.67s | **-63.0%** |                                                                                                                                          
| Cache hit rate | 56.8% | 71.3% | **+14.5pp** |                                                                                                                                   
| Throughput | 0.62 req/s | 0.80 req/s | **+28.7%** |                                                                                                      
                                                            
  Per-round TTFT in later rounds (where host memory is saturated) improved by ~70%:                                                                                                  
  - Round 17: 3.11s → 0.87s                                 
  - Round 19: 4.16s → 1.26s                                                                                                                                                          
                                                            
  The deferred allocation approach completely eliminates the deadlock where mem_pool_host.alloc() fails under memory pressure, causing prefetch to silently skip without ever calling batch_exists. Great fix!

**Comment by beef9999** [comment]
I can only ensure wait_complete is correct, best_effort and timeout has too many corner cases, especially when TP synchronization is involved.

So is it OK I only apply this code to the wait_complete logic?

**Comment by xiezhq-hermann** [comment]
I think it makes sense to have this especially for `wait_complete` since it's designed for throughput-driven scenarios, we can proceed that way if the changes can be made clean

**Comment by xiezhq-hermann** [comment]
@beef9999 can we resolve the conflict and then proceed with making this default for `wait_complete` and `timeout` policy?

**Comment by xiezhq-hermann** [comment]
@stmatengss @huangtingwei9988 @alphabetc1 can you help taking a look and testing the latest update? I am making this `check then alloc` approach to be the default one, would appreciate some reliability validation

**Review by xiezhq-hermann — APPROVED** [review]

**Comment by stmatengss** [comment]
> @stmatengss @huangtingwei9988 @alphabetc1 can you help taking a look and testing the latest update? I am making this `check then alloc` approach to be the default one, would appreciate some reliability validation

Sure. I will add it to my todo list.

File: python/sglang/srt/mem_cache/hiradix_cache.py [hunk 1]
```diff
@@ -598,13 +600,35 @@ def _drain_queue(q, limit: Optional[int]):
...
+                    self._revoke_pending_prefetch(req_id)
+                    continue
+
+                # The L3 hit count is now known, so reserve exactly that much host
+                # memory (this is the whole point: no over-allocation up front).
+                host_indices = cc.mem_pool_host.alloc(operation.storage_hit_count)
+                if host_indices is None:
+                    self.evict_host(operation.storage_hit_count)
+                    host_indices = cc.mem_pool_host.alloc(operation.storage_hit_count)
+                if host_indices is None:
```
stmatengss, line 643: **[P2] Graceful memory-pressure fallback removed — all-or-nothing allocation**

The old code in `prefetch_from_storage` tried three strategies:
1. `alloc(prefetch_length)`
2. `evict_host` + `alloc(prefetch_length)`
3. **Reduce to available size** + `alloc(reduced)` — partial prefetch, better than nothing

This new code only tries strategies 1 and 2, then revokes entirely. Under memory pressure, a partial prefetch (loading some pages from L3) is better than no prefetch at all — the request still benefits from whatever hits were found.

Consider restoring the reduced-size fallback before the final revoke:
```python
if host_indices is None:
    available_size = cc.mem_pool_host.available_size()
    reduced = available_size - (available_size % self.page_size)
    if reduced >= self.prefetch_threshold and reduced < operation.storage_hit_count:
        operation.storage_hit_count = reduced
        operation.hash_value = operation.hash_value[:reduced // self.page_size]
        host_indices = cc.mem_pool_host.alloc(reduced)
if host_indices is None:
    self._revoke_pending_prefetch(req_id)
    continue
```

python/sglang/srt/mem_cache/hiradix_cache.py
```diff
@@ -1814,15 +1829,17 @@ def release_aborted_request(self, rid: str):
         if rid not in self.ongoing_prefetch:
             return
 
-        last_host_node, prefetch_key, host_indices, operation = self.ongoing_prefetch[
-            rid
-        ]
+        last_host_node, prefetch_key, operation = self.ongoing_prefetch[rid]
         if operation.host_indices is None:
```
stmatengss, line 1879: **[P2] Missing cleanup in abort path creates race window with `_drain_storage_hit`**

When `host_indices is None`, this sets `_terminated_flag` but does **not** pop from `ongoing_prefetch` or release the host node. Cleanup is deferred to `_drain_storage_hit`.

There's a subtle race: if `_drain_storage_hit` pops this operation from `prefetch_hit_queue` and sees `is_terminated() == True` but the `ongoing_prefetch` entry was already removed by a concurrent `_revoke_pending_prefetch` (from `check_prefetch_progress`'s P1 path), then `_drain_storage_hit` calls `_revoke_pending_prefetch` which returns early (`info is None`), and the host node is never released.

The P1 fix above (returning `False` instead of revoking when `host_indices is None`) would prevent this race from manifesting in practice. But consider also releasing the host node eagerly here:
```python
if operation.host_indices is None:
    self.cache_controller.terminate_prefetch(operation)
    self._revoke_pending_prefetch(rid)
    return
```

stmatengss (commented):
Strong optimization with clear benchmark wins (cache hit rate 41% → 55%, P90 TTFT -51%). The deferred-allocation design is sound.

One P1 behavioral issue and two P2s to address before merge:

**P1 (early termination):** `check_prefetch_progress` can kill an in-flight L3 query prematurely because `host_indices is None` is now a normal transient state (not an error).

**P2 (graceful fallback):** The old code had a reduced-size fallback when full allocation failed — a partial prefetch is better than none under memory pressure. That's gone.

**P2 (missing tests):** Two-thread synchronization with a new queue has subtle race conditions. The P1 issue would have been caught by a basic test.

xiezhq-hermann:
@hzh0425 the current changes is breaking the unified tree branch, should we guard unified tree from this or should we push unified tree to adopt the same policy? it seems a rather substantial change to support unified tree on this.

PR #19329, opened 2026-02-25
Bugfix: fix symm not enabled due to incorrect registration of comm

wangfakang:
CC @ShangmingCai @nvcastet @Fridge003 @BBuf @yizhang2077  @ch-wan    PTAL, thx.

## Motivation

fix symm not enabled due to incorrect registration of comm.

1. `get_local_dp_buffer` uses the group obtained by `get_tp_group` by default to register symm.  
2. The `attn_cp_all_gather_into_tensor` operation uses the group obtained by `get_attention_cp_group` to perform allgather.  
3. Steps 1 and 2 cause symm to fail when enabled because the two groups are inconsistent.

https://github.com/sgl-project/sglang/blob/671b59557045f13c13650c29b4e56a02874f7fcd/python/sglang/srt/layers/communicator_nsa_cp.py#L141-L163

https://github.com/sgl-project/sglang/blob/c4ef33862b30c6c41eeaefd1fdd4739243a838a6/python/sglang/srt/layers/dp_attention.py#L129-L130

https://github.com/sgl-project/sglang/blob/c4ef33862b30c6c41eeaefd1fdd4739243a838a6/python/sglang/srt/layers/dp_attention.py#L572-L573

https://github.com/sgl-project/sglang/blob/671b59557045f13c13650c29b4e56a02874f7fcd/python/sglang/srt/distributed/device_communicators/pynccl_allocator.py#L180-L184

https://github.com/sgl-project/sglang/blob/c4ef33862b30c6c41eeaefd1fdd4739243a838a6/python/sglang/srt/distributed/device_communicators/pynccl_allocator.py#L52-L58

## Modifications

Modify the functions `get_local_dp_buffer` and `get_global_dp_buffer` to add the `GroupOrdinator` parameter, which can make the comm group registered by symm consistent with the comm group of the collection operator.

## Checklist

## Review Process


python/sglang/srt/layers/dp_attention.py
```diff
@@ -126,8 +126,8 @@ def get_global_dp_buffer(cls) -> torch.Tensor:
         return buffer
 
     @classmethod
-    def get_local_dp_buffer(cls) -> torch.Tensor:
-        with use_symmetric_memory(get_tp_group(), disabled=not cls._dp_max_padding):
+    def get_local_dp_buffer(cls, group: GroupCoordinator) -> torch.Tensor:
+        with use_symmetric_memory(group, disabled=not cls._dp_max_padding):
```
ShangmingCai, line 140: Why not make `group: Optional[GroupCoordinator] = None`, if group==None, then we still use `get_tp_group()` by default.

wangfakang: > Why not make `group: Optional[GroupCoordinator] = None`, if group==None, then we still use `get_tp_group()` by default.

Using default values can easily lead to `group` inconsistency, so it's necessary to explicitly declare the correct `group` to ensure consistency with the communication operator's context.

wangfakang:
> Please, if possible use your own words to answer review comments instead of AI. It is easier when we have more direct and accurate back and forth. I don't think they are distinct issues. I think it would be better to fix the other issue first before merging this one since this one would give you the impression buffers are registered but they are actually still associated with the tp group. And before merging those fixes, we would need to re-run key configs for DSR1, and qwen to check if we did not break anything.

@nvcastet I apologize for the confusion. Would reverting to the previous design require another PR? Thank you.

wangfakang:
> > > Please, if possible use your own words to answer review comments instead of AI. It is easier when we have more direct and accurate back and forth. I don't think they are distinct issues. I think it would be better to fix the other issue first before merging this one since this one would give you the impression buffers are registered but they are actually still associated with the tp group. And before merging those fixes, we would need to re-run key configs for DSR1, and qwen to check if we did not break anything.
> > 
> > 
> > @nvcastet I apologize for the confusion. Would reverting to the previous design require another PR? Thank you.
> 
> @wangfakang: @merrymercy can shine in on that topic, but the issue with rolling back was the `snapshot()` API cost, see previous version at https://github.com/sgl-project/sglang/pull/12524/changes#diff-1857ea2e79f03309e0776136d7e45b432e0369c20ff8a57d418d68c764bb733f. We would need a new design decoupling registration without the cpu overhead of `snapshot()`.

@nvcastet @merrymercy I made a small refactoring of SymmPool that replace global MemPool with per-group MemPool dictionary #20153  . Each communication group now gets its own MemPool to ensure memory allocated for one group's comm is properly registered with that comm when reused from pool. This prevents cross-group memory allocation issues in multi-comm scenarios.

Now, Combining #19329 and #20153 can completely fix the two types of issues mentioned by @nvcastet. I can observe all the effects through the nccl tuning log when I enable symm locally.

wangfakang:
/rerun-failed-ci

[stage-b-test-2-npu-a2](https://github.com/sgl-project/sglang/actions/runs/25471283581/job/74735545829?pr=19329#logs) install dependencies failed.

wangfakang:
/rerun-failed-ci

retry some flaky.

wangfakang:
> @wangfakang Thanks for your PR! I believe, there is an issue here when multiple groups (communicators) are used with "use_symmetric_memory" context manager in the same run. The current behavior is:
> 
> * If the pool does not have enough inactive memory, it will allocate and register memory with the specified group.
> * If the pool has enough inactive memory, it just returns a memory block from the pool (without considering the specified group).
> 
> So we would need to come up with a design to address this issue: aka making sure the memory returned by the pool is properly registered with the correct group.
> 
> In a previous design, we decoupled registration from allocation and we could register or re-register memory segments at the exit of the context manager but now we perform the registration with the allocation since we did not want to pay the CPU cost of pytorch memory snapshot API. I guess we could come back to this earlier design but tracking the allocations ourself in C++ instead of going through memory snapshot. CC @merrymercy

Hello @nvcastet, following up on our previous discussion, the prerequisite [PR](https://github.com/sgl-project/sglang/pull/21392) has now been merged. We are ready to proceed with #19329. Could you please help review it and trigger the following test suites? thx.
* nightly-8-gpu-b200
* stage-c-test-4-gpu-b200

wangfakang:
hello @nvcastet @ShangmingCai @yizhang2077, All test cases related to symm have [passed](https://github.com/sgl-project/sglang/actions/runs/25698664201). Could you please help with a final round of review? thx.

image

nvcastet (approved):
Thanks for working hard to implement the correct fix!

PR #19340, opened 2026-02-25
[diffusion] model: Fix a performance bug in the Wan model about usp

triple-mu:
## Motivation

For non-cross-attention (self-attention), applying Ulysses all-to-all on q/k/v is correct and required.
Wan cross-attention is different: k/v come from conditioning branches (text/image), their sequence is relatively short, and they are not sequence-sharded before attention. In practice, cross-attention k/v are already full-sequence tensors on each rank. Applying _usp_input_all_to_all to them is therefore semantically wrong and causes runtime issues.

For the Wan model, there is currently an issue with the text and image inputs in cross-attention when sp > 1.
The sequence length produced by the text encoder is 257, while the sequence length produced by the image encoder is 512. I printed the tensor shape changes inside the cross-attention module under USP (sequence parallelism), as shown in the screenshot.

image

From the logs, we can see that after the all_to_all communication, the sequence lengths of k/v and k_img/v_img become sp times larger, which is incorrect.

This indicates that the cross-attention inputs for text and image are being improperly handled under sequence parallelism.

## Modifications

Explicitly separate self-attention and cross-attention paths:

1. Keep self-attention USP behavior unchanged
- Under Ulysses, q/k/v still use all-to-all.
- Under ring parallelism, still use ring_attn.
2. Add a dedicated cross-attention path
- q still goes through _usp_input_all_to_all (query-side parallel behavior preserved).
- k/v no longer use all-to-all; instead, select only the local head shard for the current rank via _ulysses_input_split.
3. Disable ring_attn for cross-attention to avoid entering an incompatible ring path.

## Checklist

## Review Process

### DIFF
--- a/python/sglang/multimodal_gen/runtime/layers/attention/layer.py
+++ b/python/sglang/multimodal_gen/runtime/layers/attention/layer.py
@@ -22,6 +22,7 @@ from sglang.multimodal_gen.runtime.layers.attention.backends.attention_backend i
 )
 from sglang.multimodal_gen.runtime.layers.attention.selector import get_attn_backend
 from sglang.multimodal_gen.runtime.layers.usp import (
+    _ulysses_input_split,
     _usp_input_all_to_all,
     _usp_output_all_to_all,
     ring_attn,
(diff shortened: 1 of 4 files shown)


BBuf:
Can you report wan2.2 benchmark with pr? thakns.

mickqian (approved):

PR #19350, opened 2026-02-25
[Logging] Fix prefill side logging in pd disagg

YAMY1234:
## Motivation

The disaggregation prefill path is missing scheduler metrics logging (prefill stats), which makes it inconsistent with the normal prefill path and harder to monitor prefill bugs and performance in disaggregated setups.

## Modifications

Add `log_prefill_stats` call in `process_batch_result_disagg_prefill` when `current_scheduler_metrics_enabled` is true, aligning the disaggregation prefill path with the standard prefill logging behavior.

Before:
image

After:
image

## Accuracy Tests

N/A - This change only adds optional metrics logging and does not affect model outputs.

## Benchmarking and Profiling

N/A - Logging is gated behind `current_scheduler_metrics_enabled` and has negligible overhead.

## Checklist

### DIFF
--- a/python/sglang/srt/disaggregation/prefill.py
+++ b/python/sglang/srt/disaggregation/prefill.py
@@ -535,6 +535,14 @@ class SchedulerDisaggregationPrefillMixin:
 
         self.maybe_send_health_check_signal()
 
+        if self.current_scheduler_metrics_enabled:
+            can_run_cuda_graph = getattr(result, "can_run_cuda_graph", False)
+            self.log_prefill_stats(
+                prefill_stats=batch.prefill_stats,
+                can_run_cuda_graph=can_run_cuda_graph,
+                dp_cooperation_info=batch.dp_cooperation_info,
+            )
+
     def process_disagg_prefill_inflight_queue(
         self: Scheduler, rids_to_check: Optional[List[str]] = None
     ) -> List[Req]:


ShangmingCai (commented):
CC @Kangyan-Zhou this issue has been added and removed over and over again. After tracing refactoring, do we still need this now? I remember you are working on similar issue early before, please help review

hnyls2002:
@ShangmingCai Good catch on the history — here's the trace:

The logging was originally handled at schedule-time via `maybe_prepare_mlp_sync_batch_and_log_stats`, which the disagg prefill path also called. In #18778, that function was renamed to `maybe_prepare_mlp_sync_batch` (logging responsibility removed), and `log_prefill_stats` was moved to result-processing-time inside `process_batch_result_prefill` and `process_batch_result_dllm`. However, `process_batch_result_disagg_prefill` was missed in that refactor.

Since the disagg prefill path bypasses the standard `process_batch_result` dispatcher entirely (calls `process_batch_result_disagg_prefill` directly), there's no other code path that provides this logging. Additionally, `log_prefill_stats` itself has dedicated `DisaggregationMode.PREFILL` branches (e.g. `#prealloc-req`, `#inflight-req`, PD transfer metrics), confirming it was designed to be called from the disagg path.

This PR correctly fixes the #18778 regression. LGTM.

PR #19358, opened 2026-02-25
Allow SGLang to still work if Triton is not installed.

jonahbernard:
## Motivation

Prevents SGLang from immediately crashing if Triton is not installed. This is useful for running on hardware that does not use Triton. Adapted from https://github.com/vllm-project/vllm/pull/6786.

## Modifications

Creates a mock Triton library that is used if Triton is not available in the environment. 

## Checklist

## Review Process


Jonahcb:
@ping1jing2 do you want me implement the second sglang-bot suggestion (lint checker)? 

The second one would be useful, but I don't know what the SGLang team's rules are for the lint checker so I want to confirm that you want me to implement the lint check.

python/sglang/test/test_cutlass_moe.py
```diff
@@ -1,14 +1,13 @@
...
 import torch
-import triton  # Added import
-import triton.testing  # Added import
 from transformers import AutoConfig
 
 from sglang.srt.layers.moe.cutlass_moe import cutlass_fused_experts_fp8
 from sglang.srt.layers.moe.fused_moe_triton.fused_moe import fused_experts
 from sglang.srt.layers.moe.moe_runner.base import MoeRunnerConfig
 from sglang.srt.layers.moe.topk import StandardTopKOutput
+from sglang.srt.triton_utils import triton  # Added import
```
ping1jing2, line 10: maybe you missed the `testing` here

jonahbernard: @ping1jing2 thank you for looking! `testing` was removed by the `pre-commit` because `testing` is accessed from the imported `triton'. `pre-commit` removed the `testing` in a lot of other cases as well.

scripts/lint/check_direct_triton_imports.py
```diff
@@ -0,0 +1,84 @@
...
+]
+
+
+def main() -> int:
+    errors: list[str] = []
+
+    for root in SCAN_ROOTS:
+        root_path = pathlib.Path(root)
+        if not root_path.exists():
+            continue
```
ping1jing2: 1. These directories must always exist. If not, we should throw a warning or error msg rather than continue.
2. maybe we should use `if not root_path.is_dir():` here

scripts/lint/check_direct_triton_imports.py
```diff
@@ -0,0 +1,84 @@
...
+            if any(str(p).startswith(a) for a in ALLOWED):
+                continue
+            try:
+                source = p.read_text(encoding="utf-8", errors="replace")
+            except OSError:
+                continue
+            try:
+                tree = ast.parse(source, filename=str(p))
+            except SyntaxError:
+                continue
```
ping1jing2, line 71: 1. we may use `p.read_byte()` here due to performance considerations, because I've checked that `source` can either be a normal string, a byte string, or an AST object in `ast.parse`
2. hope you can combine these two `try-except` into one.

scripts/lint/check_direct_triton_imports.py
```diff
@@ -0,0 +1,84 @@
...
+                        if alias.name == "triton" or alias.name.startswith("triton."):
+                            errors.append(
+                                f"{p}:{node.lineno}: direct `import {alias.name}` found"
+                            )
+                elif isinstance(node, ast.ImportFrom) and node.module and (
+                    node.module == "triton" or node.module.startswith("triton.")
+                ):
+                    errors.append(
+                        f"{p}:{node.lineno}: direct `from {node.module} import ...` found"
+                    )
```
ping1jing2: Actually, I am not faimilar with `AST` but I found `ast.NodeVisitor` class, and Gemini told me we can optimize it like the code below, please correct me if i am wrong.

```python
class TritonImportChecker(ast.NodeVisitor):
    def __init__(self, filename):
        self.filename = filename
        self.errors = []
        self._in_try_block = 0

    def visit_Try(self, node):
        # 进入 Try 块，标记内部导入为受保护
        self._in_try_block += 1
        self.generic_visit(node)
        self._in_try_block -= 1

    def _check_triton(self, name, lineno, is_from=False):
        if self._in_try_block > 0:
            return
        
        if name == "triton" or name.startswith("triton."):
            msg_type = f"from {name} import ..." if is_from else f"import {name}"
            self.errors.append(f"{self.filename}:{lineno}: direct `{msg_type}` found")

    def visit_Import(self, node):
        for alias in node.names:
            self._check_triton(alias.name, node.lineno)

    def visit_ImportFrom(self, node):
        if node.module:
            self._check_triton(node.module, node.lineno, is_from=True)


checker = TritonImportChecker(path_str)
checker.visit(tree)
all_errors.extend(checker.errors)
```

jonahbernard: @ping1jing2 thank you. this is a good improvement on the speed. done!

scripts/lint/check_direct_triton_imports.py
jonahbernard: @ping1jing2 done

scripts/lint/check_direct_triton_imports.py
jonahbernard: @ping1jing2 done, thank you

ping1jing2 (approved):

Issue #9853, opened 2025-09-01
RFC: Command-Line Interface Architecture for SGLang


vincentzed:
# RFD: Unified `sglang` Command-Line Interface

## Status

Draft

## Summary

SGLang’s current invocation pattern requires users to run Python modules directly, for example:

```bash
python -m sglang.launch_server
python -m sglang.bench_latency
```

This is workable but brittle: it is easy to mistype, hard to discover, and forces users to memorize internal module names that we may want to refactor. This RFD proposes a first-class `sglang` CLI with a small set of well-named commands, and an internal structure that can scale without turning into a bespoke CLI framework.

This document leans heavily on a close study of vLLM’s CLI, because vLLM is a good preview of what happens when an inference engine accumulates years of flags, modes, and “just one more option” requests. The goal here is not to imitate vLLM; it’s to learn from its successes and avoid inheriting its pain.

## Context and motivation

We repeatedly see users struggle with “what exact incantation do I need?” and “which module path is correct?” because SGLang exposes multiple operations as separate Python module entrypoints with no unified discovery mechanism. The net effect is that the interface leaks implementation details: users are coupled to our package structure.

A unified CLI should make common workflows obvious and stable. It should also give us a place to enforce consistent validation and error messages, rather than distributing that responsibility across multiple `__main__` blocks and scripts.

## Proposed interface

The entry point is `sglang`. The initial command surface is:

```text
sglang launch
sglang chat
sglang complete
sglang batch
sglang bench <subcommand>
```

Benchmarks are intentionally grouped as a subtree so they remain discoverable without polluting unrelated command help:

```text
sglang bench latency
sglang bench serving
sglang bench throughput
```

Whether the canonical verb is `launch` or `serve` is open; if we pick one, we should almost certainly provide the other as an alias for familiarity across ecosystems.

## Design principles

This CLI should be discoverable via `--help`, stable under internal refactors, and responsive for trivial operations like `--version` and `--help`. It should fail fast on invalid combinations of flags before expensive initialization. It should also be maintainable: the CLI will grow over time, and we should be wary of designs that force us into thousands of lines of plumbing or a proliferation of one-off parsing hacks.

The vLLM study is relevant here because vLLM demonstrates both ends of the spectrum: they have real rigor in validation and reuse, and they also show how a CLI can accumulate framework-wrestling code once it reaches a certain scale.

## Architecture

[...]

PR #19368, opened 2026-02-26
cli: Add lazy imports and fail-fast config validation (RFC #9853)

karthikreddy-02:
## Motivation
Partial implementation of RFC #9853. SGLang's CLI currently imports 
`serve` and `generate` at module level, causing PyTorch to initialize 
even for `sglang --help`. Additionally, invalid server configurations 
are not caught until deep into model initialization.

## Modifications
- **`cli/main.py`**: Removed top-level imports of `serve` and `generate`. 
  Both are now lazily imported inside their dispatch branches so 
  `sglang --help` no longer triggers PyTorch initialization.
- **`cli/config.py`** (new): Pydantic `ServerConfig` model that validates 
  `model_path` (non-empty), `port` (>= 1024), and `host` before any 
  model loading begins.
- **`cli/serve.py`**: Constructs `ServerConfig` immediately after argument 
  parsing. Invalid configs now fail fast before `run_server()` is called.

## Accuracy Tests
N/A — no changes to model forward code or kernel logic.

## Benchmarking and Profiling
N/A — this is a CLI parsing change. Running `sglang --help` will be 
noticeably faster as PyTorch is no longer imported eagerly.


python/sglang/test/test_cli_config.py
Kangyan-Zhou, line 1: Can we register this as a CPU test?

karthikreddy-02: Yes, this is a pure CPU test.

Kangyan-Zhou: Sry I meant https://github.com/sgl-project/sglang/pull/19518/changes#diff-83a9a991ee3c710a3b3313a371e9c2b445eed8cc19d822d715a89aedb3d44062R20

karthikreddy-02: @Kangyan-Zhou, wanted to check if there's anything I should be doing/fixing to get this PR approved. Please let me know. Thanks! :)

karthikreddy-02: Done — added test/registered/cli/test_cli_config.py with register_cpu_ci. Used unittest.TestCase directly instead of CustomTestCase since the test has no GPU/srt dependencies and want to avoid pulling in the heavy import chain.

Kangyan-Zhou: Can you remove this file?

Kangyan-Zhou (approved):

PR #19370, opened 2026-02-26
Fix parallel tool call test for speculative decoding variants

alisonshao:
## Summary

- Fix `_test_parallel` in `tool_call_test_runner.py` to reliably pass across all DeepSeek V3.2 variants (DP8, DP8+MTP, TP8, TP8+MTP)
- Use `tool_choice="auto"` with strict tools to get `structural_tag` constraint (DSML format), avoiding EAGLE+DP early JSON array termination with `tool_choice="required"`
- Use `temperature=0` for deterministic behavior and explicit prompt for reliable parallel tool call generation

## Root Cause

The parallel tool call test was failing for different variants depending on the constraint mode:

1. **`tool_choice="required"`** (original): Uses JSON schema `{"type": "array", "minItems": 1}`. EAGLE speculative decoding with DP attention would close the JSON array after 1 tool call since the grammar accepts `minItems=1`. Failed for DP8+MTP variant.

2. **`tool_choice="auto"` with non-strict tools** (first fix attempt): Produces **no grammar constraint at all**, leaving model behavior uncontrolled. Failed for TP8 variant due to non-deterministic generation at `temperature=0.1`.

The fix uses `tool_choice="auto"` with **strict tools** (`strict: true`), which activates the `structural_tag` constraint (DSML format). Combined with `temperature=0` and an explicit prompt, the test reliably produces 2+ parallel tool calls across all configurations.

## Test plan

- [ ] CI passes for all 4 DeepSeek V3.2 variants: DP8, DP8+MTP, TP8, TP8+MTP


alisonshao:
https://github.com/sgl-project/sglang/actions/runs/22425806167/job/64933520419

python/sglang/test/tool_call_test_runner.py
```diff
@@ -262,11 +267,17 @@ def _test_reasoning_usage(client, model):
 
 def _test_parallel(client, model):
     """Single request should return multiple tool calls."""
-    response = _call(
```
Kangyan-Zhou, line 265: should we just extend the interface for _call function?

PR #19391, opened 2026-02-26
[Qwen3.5] Enable MTP spec_v2 and add test for nvidia/Qwen3.5-397B-A17B-NVFP4

hlu1:
## Motivation
- Make MTP_v2 work for Qwen3.5 by passing `mm_input_embeds` to the MTP head.
- Add MTP_v1/v2 and non-MTP accuracy test for `nvidia/Qwen3.5-397B-A17B-NVFP4` and check acceptance length in the MTP tests. Note that it uses the eval harness `from sglang.test.run_eval import run_eval` which applies the chat_template. Without the chat_template, the accuracy is very bad. The sampling parameters are based on the official recommendation from https://huggingface.co/Qwen/Qwen3.5-397B-A17B-FP8
- Remove the two extra_buffer and mtp_v2 checks. These two checks are incorrect because mtp_v2 only requires extra_buffer when radix_cache is on. To make it more user friendly, I changed the default behavior of turning off radix-cache silently when spec decoding, no_buffer, and radix cache are enabled at the same time to raising an exception, in case the user want to enable spec decoding (v1 or v2) and radix cache but forgot to enable extra_buffer. 
- Remove duplicated code from server_args.py

co-author: @vincentzed https://github.com/sgl-project/sglang/pull/18906

## Accuracy
Without radix cache
```
gpqa:
Repeat: 8, mean: 0.866
Scores: ['0.859', '0.869', '0.874', '0.869', '0.884', '0.848', '0.869', '0.859']
```
With radix cache

```
Repeat: 8, mean: 0.861
Scores: ['0.848', '0.843', '0.874', '0.854', '0.864', '0.859', '0.894', '0.854']
```

## Benchmark
radix cache off

tps_vs_throughput_v1_vs_v2

## Checklist

## Review Process

### DIFF
--- a/python/sglang/srt/disaggregation/decode.py
+++ b/python/sglang/srt/disaggregation/decode.py
@@ -183,11 +183,6 @@ class HybridMambaDecodeReqToTokenPool(HybridReqToTokenPool):
             pre_alloc_size=pre_alloc_size,
         )
 
-        if envs.SGLANG_ENABLE_SPEC_V2.get() and not enable_mamba_extra_buffer:
-            raise ValueError(
-                "Spec v2 requires mamba scheduler strategy `extra_buffer` for mamba models. "
-                "Please set `--mamba-scheduler-strategy extra_buffer`."
-            )
         self.mamba_ping_pong_track_buffer_size = 2 if enable_overlap_schedule else 1
         self.enable_mamba_extra_buffer = enable_mamba_extra_buffer
         self.enable_memory_saver = enable_memory_saver
--- a/python/sglang/srt/mem_cache/memory_pool.py
+++ b/python/sglang/srt/mem_cache/memory_pool.py
@@ -465,11 +465,6 @@ class HybridReqToTokenPool(ReqToTokenPool):
             device=device,
             enable_memory_saver=enable_memory_saver,
         )
-        if envs.SGLANG_ENABLE_SPEC_V2.get() and not enable_mamba_extra_buffer:
-            raise ValueError(
-                "Spec v2 requires mamba scheduler strategy `extra_buffer` for mamba models. "
-                "Please set `--mamba-scheduler-strategy extra_buffer`."
-            )
 
         self.mamba_ping_pong_track_buffer_size = 2 if enable_overlap_schedule else 1
         self.enable_mamba_extra_buffer = enable_mamba_extra_buffer
--- a/python/sglang/srt/server_args.py
+++ b/python/sglang/srt/server_args.py
@@ -1888,10 +1888,11 @@ class ServerArgs:
                     self.disable_radix_cache = True
                     self.disable_overlap_schedule = False
             else:
-                logger.warning(
-                    f"Disabling radix cache since speculative decoding for {model_arch} is not supported with radix cache yet."
-                )
-                self.disable_radix_cache = True
+                if not self.disable_radix_cache:
+                    raise ValueError(
+                        f"Speculative decoding for {model_arch} is not compatible with radix cache when using --mamba-scheduler-strategy no_buffer."
+                        "To use radix cache with speculative decoding, please use --mamba-scheduler-strategy extra_buffer and set SGLANG_ENABLE_SPEC_V2=1."
+                    )
 
     def _handle_sampling_backend(self):
         if self.sampling_backend is None:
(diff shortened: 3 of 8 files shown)


python/sglang/srt/server_args.py
```diff
@@ -1778,10 +1777,11 @@ def _handle_mamba_radix_cache(
...
             else:
-                logger.warning(
-                    f"Disabling radix cache since speculative decoding for {model_arch} is not supported with radix cache yet."
-                )
-                self.disable_radix_cache = True
+                if not self.disable_radix_cache:
+                    raise ValueError(
+                        f"Speculative decoding for {model_arch} is not compatible with radix cache when using --mamba-scheduler-strategy no_buffer."
+                        "To use radix cache with speculative decoding, please use --mamba-scheduler-strategy extra_buffer and set SGLANG_ENABLE_SPEC_V2=1."
+                    )
```
ShangmingCai, line 1895: Since this change is inside the mamba rafix cache check, I think it is reasonable. Also, we disable radix cache by default for decode node in the PD scenerio, so the change LGTM.

python/sglang/srt/disaggregation/decode.py
```diff
@@ -184,11 +183,6 @@ def __init__(
             pre_alloc_size=pre_alloc_size,
         )
 
-        if envs.SGLANG_ENABLE_SPEC_V2.get() and not enable_mamba_extra_buffer:
```
yizhang2077, line 186: spec v2 can not be compatible with `no_buffer` radix cache， I think we need to add assertion here（but it should be compatible with disable radix cache here as well）

python/sglang/srt/mem_cache/memory_pool.py
yizhang2077, line 460: the same as above

test/registered/4-gpu-models/test_qwen35_models.py
```diff
@@ -0,0 +1,234 @@
...
+register_cuda_ci(est_time=1000, suite="stage-c-test-4-gpu-b200")
+
+QWEN35_FP4_MODEL = "nvidia/Qwen3.5-397B-A17B-NVFP4"
+
+ACC_THRESHOLDS = {
+    QWEN35_FP4_MODEL: {"gsm8k": 0.95},
+}
+
+
+class TestQwen35FP4(CustomTestCase):
```
yizhang2077, line 29: could we add qwen35 original model in this file as well?

python/sglang/srt/disaggregation/decode.py
hlu1: Why can't it be compatible with `no_buffer`? As I understand it, `extra_buffer` is only needed when you write the ssm/conv states to radix cache. Accuracy test with `no_buffer` and radix_cache off passes and benchmarks show speedup vs spec_v1.

test/registered/4-gpu-models/test_qwen35_models.py
hlu1: The gsm8k tests take a pretty long time when testing with chat_template and thinking on. As long as the you are ok with the added time, I can add the fp8 tests. Would you like it added for all three tests?

ShangmingCai (approved):
LGTM as long as the CI passes. CC: @yizhang2077 Please double check.

test/registered/4-gpu-models/test_qwen35_models.py
yizhang2077: I think we could add fp8 test without mtp

python/sglang/srt/disaggregation/decode.py
yizhang2077: if radix cache needs to be compatible with mtp, it needs to support bigram key, which only store keys and values in the range [0:len(key) - 1] (it is different from raw key range [0:len(key)]). `no_buffer` mode does not support this type since it can only use mamba value in the last position.

yizhang2077: wait, I think if we use speculative decoding in `no_buffer` mode, radix cache is close, so remove assertion here  is ok.

yizhang2077 (approved):

test/registered/4-gpu-models/test_qwen35_models.py
hlu1: I'll add them tomorrow in a follow-up PR.

hzh0425 (approved):
LGTM

hlu1:
The gb200 CI is temporarily disabled. I ran the tests locally and they all pass.

PR #19395, opened 2026-02-26
MFU metrics in Prometheus 

aramasethu:
## Motivation

This PR addresses issue #19286  : SGLang exposes throughput/latency metrics but lacks MFU-related estimated performance counters in Prometheus which vllm recently exposed - ([vllm-project/vllm@5cc7c44](https://github.com/vllm-project/vllm/commit/5cc7c44)) .

The goal is to add lightweight, opt-in observability signals for:
estimated per-GPU FLOPs
estimated per-GPU memory read bytes
estimated per-GPU memory write bytes
so operators can derive TFLOPS / bandwidth trends in production dashboards (similar to recent vLLM support), while keeping default serving behavior unchanged unless explicitly enabled.

## Modifications

1.  Added new Prometheus counters in `SchedulerMetricsCollector`:
`sglang:estimated_flops_per_gpu_total`
`sglang:estimated_read_bytes_per_gpu_total`
`sglang:estimated_write_bytes_per_gpu_total`

2. Added `--enable-mfu-metrics` server flag and gated emission logic so these metrics are exported only when both:
`--enable-metrics` and
`--enable-mfu-metrics`
are enabled.

3. Extended scheduler-side estimation in `scheduler_metrics_mixin.py` to include:
linear-layer FLOPs
attention dot-product FLOPs (context-dependent)
estimated weight read bytes
estimated activation read/write bytes
KV cache read/write bytes

4. Updated tests in `test/registered/metrics/test_metrics.py`:
gate-on path: MFU metrics are present and positive
gate-off path: MFU metrics are absent or remain zero

5. Updated docs:
`docs/advanced_features/server_arguments.md` (new flag)
`docs/references/production_metrics.md` (metric descriptions + PromQL + launch flags)

## Accuracy Tests

No model output path changes; accuracy impact is not expected (observability-only change).

### Benchmark (A100, Qwen/Qwen3-0.6B, 1000 prompts, input=512 output=128 tokens)

| Metric | A: main | B: branch gate-off | C: branch gate-on | B vs A | C vs B |
|---|---|---|---|---|---|
| Request throughput (req/s) | 72.96 | 75.48 | 75.55 | +3.5% | +0.1% |
| Output tok/s | 4707 | 4870 | 4874 | +3.5% | +0.1% |
| TTFT median (ms) | 3158 | 3063 | 3175 | -3.0% | +3.7% |
| E2E median (ms) | 11722 | 11145 | 10985 | -4.9% | -1.4% |

B vs A: Differences are within expected run-to-run variance (~3–5%); no clear regression with MFU metrics disabled.
C vs B: Enabling --enable-mfu-metrics shows no throughput regression and latency shifts remain within expected noise for this setup.
MFU estimation executes only in scheduler logging paths when --enable-mfu-metrics is enabled.

## Checklist

### DIFF
--- a/docs/advanced_features/server_arguments.md
+++ b/docs/advanced_features/server_arguments.md
@@ -185,6 +185,7 @@ Please consult the documentation below and [server_args.py](https://github.com/s
 | `--crash-dump-folder` | Folder path to dump requests from the last 5 min before a crash (if any). If not specified, crash dumping is disabled. | `None` | Type: str |
 | `--show-time-cost` | Show time cost of custom marks. | `False` | bool flag (set to enable) |
 | `--enable-metrics` | Enable log prometheus metrics. | `False` | bool flag (set to enable) |
+| `--enable-mfu-metrics` | Enable estimated MFU-related prometheus metrics. | `False` | bool flag (set to enable) |
 | `--enable-metrics-for-all-schedulers` | Enable --enable-metrics-for-all-schedulers when you want schedulers on all TP ranks (not just TP 0) to record request metrics separately. This is especially useful when dp_attention is enabled, as otherwise all metrics appear to come from TP 0. | `False` | bool flag (set to enable) |
 | `--tokenizer-metrics-custom-labels-header` | Specify the HTTP header for passing custom labels for tokenizer metrics. | `x-custom-labels` | Type: str |
 | `--tokenizer-metrics-allowed-custom-labels` | The custom labels allowed for tokenizer metrics. The labels are specified via a dict in '--tokenizer-metrics-custom-labels-header' field in HTTP requests, e.g., {'label1': 'value1', 'label2': 'value2'} is allowed if '--tokenizer-metrics-allowed-custom-labels label1 label2' is set. | `None` | List[str] |
--- a/python/sglang/srt/observability/scheduler_metrics_mixin.py
+++ b/python/sglang/srt/observability/scheduler_metrics_mixin.py
@@ -151,6 +151,9 @@ class SchedulerMetricsMixin:
             self.enable_mfu_metrics = bool(self.server_args.enable_mfu_metrics)
             if self.enable_mfu_metrics:
                 self._init_estimated_perf_constants()
+                self._mfu_log_flops = 0.0
+                self._mfu_log_read_bytes = 0.0
+                self._mfu_log_write_bytes = 0.0
 
             if ENABLE_METRICS_DEVICE_TIMER:
                 self.forward_pass_device_timer = DeviceTimer(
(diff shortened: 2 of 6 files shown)


sufeng-buaa:
This is a good feature. Can we also include it in the log output, controlled either by adding new server arguments or reusing existing ones?

sufeng-buaa:
Should we consider cases where the model's layer structures differ? For example, the latest Qwen3.5-397B-A17B: https://www.modelscope.cn/models/Qwen/Qwen3.5-397B-A17B-FP8. Of course, I think this issue can be addressed later.

python/sglang/srt/observability/scheduler_metrics_mixin.py
```diff
@@ -562,6 +732,32 @@ def report_decode_stats(
             self._emit_kv_metrics()
         self._publish_kv_events()
 
+    def log_decode_stats_every_iteration(
```
sufeng-buaa: I noticed that `log_decode_stats_every_iteration` isn’t being called anywhere else. Could this be caused by the recent rebase? Maybe it should be added to `report_decode_stats`?

aramasethu: Yes, this was indeed caused by the rebase. I've merged the logic from log_decode_stats_every_iteration directly into report_decode_stats's every-iteration block and removed the orphaned method.

docs/advanced_features/server_arguments.md
```diff
@@ -185,6 +185,7 @@ Please consult the documentation below and [server_args.py](https://github.com/s
 | `--crash-dump-folder` | Folder path to dump requests from the last 5 min before a crash (if any). If not specified, crash dumping is disabled. | `None` | Type: str |
 | `--show-time-cost` | Show time cost of custom marks. | `False` | bool flag (set to enable) |
 | `--enable-metrics` | Enable log prometheus metrics. | `False` | bool flag (set to enable) |
+| `--enable-mfu-metrics` | Enable estimated MFU-related prometheus metrics. | `False` | bool flag (set to enable) |
```
Kangyan-Zhou, line 188: @sufeng-buaa Does this metric have any side effect? 

If not should we just reuse the flag above?

sufeng-buaa: I think calculating it once per batch still costs time in every round. Even though the overlap scheduling mechanism may hide that overhead. We already have a lot of metric-related code, and if it keeps growing, it might eventually turn into a performance problem. Honestly, I think even the existing metrics should be managed in different levels.

aramasethu: @Kangyan-Zhou @sufeng-buaa any next steps on this? Would love to help out / contribute in any way

sufeng-buaa: I don't have any more comments. Does @Kangyan-Zhou  think anything else needs to be changed?

PR #19400, opened 2026-02-26
Fix missing StandardCombineInput import in BF16 flashinfer_trtllm MoE

alisonshao:
## Summary
- Add missing runtime import of `StandardCombineInput` in `fused_experts_none_to_flashinfer_trtllm_bf16()`, matching the pattern already used by the FP8 and FP4 variants in the same file
- Fixes `NameError: name 'StandardCombineInput' is not defined` that crashes both `test_flashinfer_trtllm_gen_attn_backend.py` and `test_flashinfer_trtllm_gen_moe_backend.py` in the nightly B200 CI
- may be related to #19266

## Error link
https://github.com/sgl-project/sglang/actions/runs/22422538348/job/64923364867

## Test plan
- [ ] Nightly B200 4-GPU tests pass (`test_flashinfer_trtllm_gen_attn_backend.py`, `test_flashinfer_trtllm_gen_moe_backend.py`)

### DIFF
--- a/python/sglang/srt/layers/moe/moe_runner/flashinfer_trtllm.py
+++ b/python/sglang/srt/layers/moe/moe_runner/flashinfer_trtllm.py
@@ -501,6 +501,8 @@ def fused_experts_none_to_flashinfer_trtllm_bf16(
     runner_config: MoeRunnerConfig,
 ) -> StandardCombineInput:
     # lazy import
+    from sglang.srt.layers.moe.token_dispatcher.standard import StandardCombineInput
+
     try:
         from flashinfer.fused_moe import trtllm_bf16_moe
     except ImportError as e:


python/sglang/srt/layers/moe/moe_runner/flashinfer_trtllm.py
samuellees: Quick Question: why not just remove this condition branch?

alisonshao: Good call — there's no circular import here, so the `TYPE_CHECKING` guard is unnecessary. Updated to make it a regular top-level import and removed the redundant lazy imports from all three functions (FP8, FP4, BF16).

alisonshao: Investigated — turns out there is a circular import if we remove the guard:

```
flashinfer_trtllm.py
→ token_dispatcher/__init__.py
→ deepep.py
→ topk.py
→ eplb/expert_location_dispatch.py
→ eplb/expert_location.py
→ eplb/eplb_algorithms/__init__.py
→ elastic_ep/elastic_ep.py
→ managers/schedule_batch.py
→ dllm/config.py
→ configs/model_config.py
→ layers/quantization (QUANTIZATION_METHODS)
→ compressed_tensors_w8a8_fp8_moe.py
→ flashinfer_trtllm.py ← circular!
```

Reverted to the original approach — keeping the `TYPE_CHECKING` guard and adding the lazy import inside the BF16 function (matching FP8/FP4 pattern).

alisonshao: error link: https://github.com/sgl-project/sglang/actions/runs/22458765782/job/65047037023

alisonshao: trying to create a cleaner fix...

alisonshao: Investigated and confirmed: the `TYPE_CHECKING` guard is required. Even importing from `token_dispatcher.standard` directly triggers `token_dispatcher/__init__.py` (Python always runs `__init__.py` for parent packages), which causes a circular import:

```
flashinfer_trtllm.py
→ token_dispatcher/__init__.py  (always runs for any submodule import)
→ deepep.py → topk.py → eplb → elastic_ep → schedule_batch
→ dllm/config.py → model_config.py
→ quantization/__init__.py (QUANTIZATION_METHODS)
→ compressed_tensors_w8a8_fp8_moe.py
→ flashinfer_trtllm.py ← circular!
```

Reverted to the minimal fix: lazy import inside the BF16 function, matching FP8/FP4 pattern.

alisonshao:
@b8zhong

b8zhong (approved):

PR #19412, opened 2026-02-26
[Bug Fix] Fix GPU memory leak in grammar-guided generation

dingzhiqiang:
## Motivation

Fixes #19410

When using grammar-guided generation features (structural_tag, json_schema, regex constraints), GPU memory continuously grows with each request until OOM occurs. Each grammar request leaks approximately 6-10 MiB of GPU memory.

## Modifications

This fix addresses three locations where grammar object references are not properly released:

1. **`stream_output_generation()`** in `scheduler_output_processor_mixin.py`:
   - Clear `req.grammar = None` when request finishes

2. **`filter_batch()`** in `sampling_batch_info.py`:
   - Filter `sampling_info.grammars` list when requests are removed from batch
   - Clear `vocab_mask` and `apply_mask_func` when grammars are cleared

3. **`copy_for_forward()`** in `sampling_batch_info.py`:
   - Clear grammars reference in the shallow copy to prevent `batch_record_buf` from holding stale references in overlap mode

## Benchmarking and Profiling

| Test | Requests | Concurrency | Memory Growth | Leak/Request |
|------|----------|-------------|---------------|--------------|
| Before fix | 100 | 1 | ~600 MiB | ~6 MiB |
| After fix | 2000 | 32 | 2 MiB | 0.001 MiB |

Tested with 2000 requests at 32 concurrency - memory stable at ~2 MiB total growth.

## Checklist

### DIFF
--- a/python/sglang/srt/managers/scheduler_output_processor_mixin.py
+++ b/python/sglang/srt/managers/scheduler_output_processor_mixin.py
@@ -955,6 +955,12 @@ class SchedulerOutputProcessorMixin:
                     # because of the one additional delayed token. This "continue" prevented the dummy output.
                     continue
                 req.finished_output = True
+
+                # Release grammar object to prevent memory leak.
+                # Grammar objects hold GPU tensor references that prevent GC from reclaiming memory.
+                if req.grammar is not None:
+                    req.grammar = None
+
                 if req.finished_len is None:
                     req.finished_len = len(req.output_ids)
                 should_output = True
--- a/python/sglang/srt/managers/scheduler_output_processor_mixin.py
+++ b/python/sglang/srt/managers/scheduler_output_processor_mixin.py
@@ -956,8 +956,7 @@ class SchedulerOutputProcessorMixin:
                     continue
                 req.finished_output = True
 
-                # Release grammar object to prevent memory leak.
-                # Grammar objects hold GPU tensor references that prevent GC from reclaiming memory.
+                # Release grammar object to free CPU resources held by the grammar state machine.
                 if req.grammar is not None:
                     req.grammar = None
 
--- a/python/sglang/srt/sampling/sampling_batch_info.py
+++ b/python/sglang/srt/sampling/sampling_batch_info.py
@@ -267,6 +267,30 @@ class SamplingBatchInfo:
         if self.logit_bias is not None:
             self.logit_bias = self.logit_bias[keep_indices_device]
 
+        # Filter grammars list to prevent memory leak.
+        # When requests are filtered out, their grammar references should also be removed.
+        # Note: grammars list length must match the batch size before filtering.
+        if self.grammars is not None and len(self.grammars) > max(
+            keep_indices, default=-1
+        ):
+            max_index = max(keep_indices, default=-1)
+            if len(self.grammars) > max_index:
+                self.grammars = [self.grammars[i] for i in keep_indices]
+                if not any(self.grammars):
+                    self.grammars = None
+            else:
+                # If grammars list is shorter than expected, clear it to avoid index errors
+                logger.warning(
+                    f"Inconsistent state: len(grammars)={len(self.grammars)} <= max(keep_indices)={max_index}. Clearing grammars."
+                )
+
+                self.grammars = None
+
+        # Clear vocab_mask when grammars are cleared to release GPU memory
+        if self.grammars is None and self.vocab_mask is not None:
+            self.vocab_mask = None
+            self.apply_mask_func = None
+
     def _filter_batch_custom_logit_processor(
         self, keep_indices: List[int], keep_indices_device: torch.Tensor
     ):
@@ -381,7 +405,10 @@ class SamplingBatchInfo:
     def copy_for_forward(self):
         # Accumulate the penalty into a pre-allocated buffer to get rid of the dependency of `penalizer_orchestrator` later
         self.update_penalties()
-        return dataclasses.replace(self, penalizer_orchestrator=None)
+        # Clear grammars reference in the copy to prevent memory leak.
+        # The grammars are only needed for filling vocab_mask, which has already been done
+        # in update_regex_vocab_mask() before this copy is created.
+        return dataclasses.replace(self, penalizer_orchestrator=None, grammars=None)
 
 
 def merge_bias_tensor(
--- a/python/sglang/srt/sampling/sampling_batch_info.py
+++ b/python/sglang/srt/sampling/sampling_batch_info.py
@@ -267,29 +267,14 @@ class SamplingBatchInfo:
         if self.logit_bias is not None:
             self.logit_bias = self.logit_bias[keep_indices_device]
 
-        # Filter grammars list to prevent memory leak.
-        # When requests are filtered out, their grammar references should also be removed.
-        # Note: grammars list length must match the batch size before filtering.
-        if self.grammars is not None and len(self.grammars) > max(
-            keep_indices, default=-1
-        ):
-            max_index = max(keep_indices, default=-1)
-            if len(self.grammars) > max_index:
+        # Filter grammars list to match the filtered batch.
+        if self.grammars is not None:
+            if not keep_indices:
+                self.grammars = None
+            else:
                 self.grammars = [self.grammars[i] for i in keep_indices]
                 if not any(self.grammars):
                     self.grammars = None
-            else:
-                # If grammars list is shorter than expected, clear it to avoid index errors
-                logger.warning(
-                    f"Inconsistent state: len(grammars)={len(self.grammars)} <= max(keep_indices)={max_index}. Clearing grammars."
-                )
-
-                self.grammars = None
-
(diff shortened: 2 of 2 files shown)


python/sglang/srt/managers/scheduler_output_processor_mixin.py
```diff
@@ -947,6 +947,12 @@ def stream_output_generation(
                     # because of the one additional delayed token. This "continue" prevented the dummy output.
                     continue
                 req.finished_output = True
+
+                # Release grammar object to prevent memory leak.
+                # Grammar objects hold GPU tensor references that prevent GC from reclaiming memory.
+                if req.grammar is not None:
```
DarkSharpness, line 960: I wonder why does grammar holds references to GPU memory? In my understanding, grammar should just maintain the inner CPU data structure.

python/sglang/srt/sampling/sampling_batch_info.py
```diff
@@ -267,6 +267,23 @@ def filter_batch(self, keep_indices: List[int], keep_indices_device: torch.Tenso
...
+                    self.grammars = None
+            else:
+                # If grammars list is shorter than expected, clear it to avoid index errors
+                self.grammars = None
+
+        # Clear vocab_mask when grammars are cleared to release GPU memory
+        if self.grammars is None and self.vocab_mask is not None:
+            self.vocab_mask = None
+            self.apply_mask_func = None
+
```
DarkSharpness: In my understanding, before this PR, we apply no change to `self.grammars`. After this PR, `self.grammars` become those remaining in batch? Will this break some latent invariants? The underlying state has change a lot (all grammars -> partial grammars).

In my understanding, if this PR is correct, it means that `self.grammars` may no longer be used within this batch, which implies that applying mask is already done. In this case, I guess we could simply drop all the grammars/vocab_mask/apply_mask_func safely after applying mask, as long as we won't reuse the grammars or vocab_mask.

dingzhiqiang: Thanks for the careful review! Your questions are excellent. Let me address them one by one:

### Q1: Why does grammar hold references to GPU memory?

Your understanding is correct — **grammar objects themselves only contain CPU data structures** (`GrammarMatcher`, `CompiledGrammar`, `accepted_tokens`, etc.).

The actual GPU memory is `SamplingBatchInfo.vocab_mask`, a GPU tensor allocated via `allocate_token_bitmask()`.

The root cause of the leak is **reference chains preventing GC**:

```
batch_record_buf (keeps last 2 batches in overlap mode)
  └── ModelWorkerBatch
      └── SamplingBatchInfo (shallow copy from copy_for_forward)
          ├── grammars list (not cleared in original code)
          └── vocab_mask (GPU tensor - the actual GPU memory!)
```

So the issue is not that grammar directly holds GPU memory, but that the shallow copy of `SamplingBatchInfo` is retained by `batch_record_buf`, preventing the entire object (including `vocab_mask`) from being GC'd.

### Q2: Will filtering `self.grammars` break latent invariants?

Actually, it's the opposite — **the original code was breaking the invariant**.

When `filter_batch()` is called, it filters `temperatures`, `top_ps`, `top_ks`, etc., but **does NOT filter `grammars`**. This means after filtering:
- `batch.reqs` length is reduced
- `grammars` list length remains unchanged
- The correspondence between `grammars[i]` and `batch.reqs[i]` is broken

The fix **restores** the correct invariant: `len(grammars) == len(batch)` and `grammars[i]` corresponds to `batch.reqs[i]`.

### Q3: Can we simply drop all grammars/vocab_mask after applying mask?

Your intuition is correct! Let me explain:

**`vocab_mask`**: Yes, it can be dropped after `apply_penalties()` since it's no longer needed in the current step.

**`grammars`**: It's slightly more nuanced:
- **`grammars` in the original `SamplingBatchInfo`**: Cannot be dropped because the next decode step's `update_regex_vocab_mask()` needs it to refill `vocab_mask` (grammar state changes after each token acceptance)
- **`grammars` in the copy created by `copy_for_forward()`**: Can be dropped because the copy only needs the pre-computed `acc_linear_penalties`, not the ability to recompute vocab_mask

The current fix clears `grammars=None` in `copy_for_forward()`, which:
1. Keeps `grammars` in the original object for the next step
2. Removes the reference from the copy, allowing GC to reclaim memory

**Regarding further optimization**: You're right, we could also clear `vocab_mask` in the copy:

```python
def copy_for_forward(self):
    self.update_penalties()
    return dataclasses.replace(
        self,
        penalizer_orchestrator=None,
        grammars=None,
        vocab_mask=None,       # could add this
        apply_mask_func=None   # could add this
    )
```

However, the current fix already resolves the leak (from ~6-10 MiB/request down to ~0.001 MiB/request). Adding these two lines would be a more thorough cleanup but isn't strictly necessary. I can add them if you think it makes the intent clearer.

DarkSharpness: You may give it a try. Please clean up this part thoroughly. It's still surprising to me that the original implementation was wrong... I need some more time to figure out the logic here.

python/sglang/srt/sampling/sampling_batch_info.py
```diff
@@ -267,6 +267,23 @@ def filter_batch(self, keep_indices: List[int], keep_indices_device: torch.Tenso
         if self.logit_bias is not None:
             self.logit_bias = self.logit_bias[keep_indices_device]
 
+        # Filter grammars list to prevent memory leak.
+        # When requests are filtered out, their grammar references should also be removed.
+        if self.grammars is not None:
+            max_index = max(keep_indices, default=-1)
+            if len(self.grammars) > max_index:
```
DarkSharpness: Can we simply compare `len(self.grammmars)` with `len(keep_indices)` here? The logic here is extremely hard to follow. There should be only 3 cases:

1. same length, do nothing
2. `len(keep_indices) == 0`, drop all
3. otherwise, slice the grammars

python/sglang/srt/sampling/sampling_batch_info.py
DarkSharpness: Logically, this should be correct. I would recommend reset `vocab_mask` and `apply_mask_func` right after we apply the mask (in function `apply_logits_bias`). This should avoid all memory leak in theory. @dingzhiqiang Please give it a try.

python/sglang/srt/managers/scheduler_output_processor_mixin.py
DarkSharpness: The problem here is: does grammar holds reference to mask? If not, then releasing grammar object can not prevent GPU memleak.

python/sglang/srt/sampling/sampling_batch_info.py
dingzhiqiang: Good suggestion! Simplified the logic to just 3 cases:

```python
if self.grammars is not None:
    if not keep_indices:
        self.grammars = None
    else:
        self.grammars = [self.grammars[i] for i in keep_indices]
        if not any(self.grammars):
            self.grammars = None
```

Also removed the `vocab_mask` cleanup from `filter_batch` since it's already handled by #20697 in `_preprocess_logits`.

python/sglang/srt/sampling/sampling_batch_info.py
dingzhiqiang: You were right! This has been fixed by #20697 which does exactly what you suggested — clearing `vocab_mask` in `_preprocess_logits()` right after `apply_logits_bias()`.

I also realized that setting `grammars=None` in `copy_for_forward()` was wrong — it would break grammar constraints in overlap mode because `_preprocess_logits()` calls `update_regex_vocab_mask()` on the copy, which needs grammars to recreate the vocab_mask. Reverted that change.

Now this PR only keeps:
1. Simplified grammar filtering in `filter_batch` (correctness fix for grammars/batch invariant)
2. `req.grammar = None` on finish (CPU resource cleanup)

python/sglang/srt/managers/scheduler_output_processor_mixin.py
dingzhiqiang: You are correct — grammar objects themselves are CPU-only and don't hold direct references to the GPU mask. The actual GPU memory leak was from `vocab_mask` being retained via reference chains (`batch_record_buf` → copy → `vocab_mask`), which is now fixed by #20697.

Updated the comment to reflect that releasing `req.grammar` is for freeing CPU resources (grammar state machine), not GPU memory.

hnyls2002:
Thanks @dingzhiqiang! One of the two files this patches, `python/sglang/srt/managers/scheduler_output_processor_mixin.py`, no longer exists (its logic is now in `managers/scheduler_components/batch_result_processor.py`), and the `vocab_mask` / `apply_mask_func` fields it clears were removed from `SamplingBatchInfo`. `sampling_info.grammars` is also rebuilt from `batch.reqs` on every forward now (`model_executor/forward_batch_info.py:774-780`), so the stale-reference window described here is much narrower. Closing as obsolete against current main - if you can still reproduce the growth, please reopen with a fresh repro and I'll take another look.

PR #19419, opened 2026-02-26
[Diffusion] [Wan] Skip USP for Cross-Attention with Replicated KV in Wan

AichenF:
## Motivation

In Wan Video's cross-attention layers, the KV tensors are derived from text encoder (512 tokens) or image encoder (257 tokens) outputs, which are identical (replicated) across all sequence-parallel ranks. The existing code unnecessarily applies the full USP pipeline to these replicated tensors:
* Ulysses all-to-all on replicated KV inflates the sequence dimension by ulysses_world_sizex (e.g., 512 tokens become 2048 on 4 GPUs), producing duplicated data and wasting both communication bandwidth and FLOPs on the subsequent attention computation.
* Ring Attention rotates identical KV blocks across ranks, gaining nothing since every rank already holds the full KV.

Since each rank's local Q shard only needs to attend to the full (and already locally available) KV, we can skip sequence parallelism entirely for cross-attention and compute local attention directly — producing mathematically identical results with less communication and less computation.

## Modifications

python/sglang/multimodal_gen/runtime/layers/attention/layer.py
Added skip_sp: bool = False parameter to USPAttention.forward().
When skip_sp=True, the method returns immediately after computing local attention, bypassing Ulysses all-to-all and Ring Attention.
python/sglang/multimodal_gen/runtime/models/dits/wanvideo.py
WanT2VCrossAttention.forward(): passes skip_sp=True to self.attn().
WanI2VCrossAttention.forward(): passes skip_sp=True to both self.attn() calls (image path and text path).

## Accuracy Tests

ring degree 4:
https://github.com/user-attachments/assets/4aa05019-fd08-44a0-85d4-652483647517

ring degree 4 skip cross attn:
https://github.com/user-attachments/assets/219484f3-b65f-4abf-a93a-c488ad4f75e6

## Benchmarking and Profiling
time per step/s on B200, achieves 7–15% speed up.
|  | baseline| cross attn skip sp |
| --- | --- | --- | 
| ring 4| 4.72 | 3.99  |
| ring2 uly 2 | 4.57 | 3.94 |
| uly 4|  4.01 | 3.68 |

## Checklist

## Review Process

### DIFF
--- a/python/sglang/multimodal_gen/runtime/layers/attention/layer.py
+++ b/python/sglang/multimodal_gen/runtime/layers/attention/layer.py
@@ -366,7 +366,7 @@ class USPAttention(nn.Module):
         ), "USPAttention does not support replicated_qkv."
         forward_context: ForwardContext = get_forward_context()
         ctx_attn_metadata = forward_context.attn_metadata
-        if get_sequence_parallel_world_size() == 1 or skip_sp: #skip_sp or
+        if skip_sp or get_sequence_parallel_world_size() == 1:
             out = self.attn_impl.forward(q, k, v, ctx_attn_metadata)
             return out
 
--- a/python/sglang/multimodal_gen/runtime/layers/attention/layer.py
+++ b/python/sglang/multimodal_gen/runtime/layers/attention/layer.py
@@ -347,11 +347,17 @@ class USPAttention(nn.Module):
         replicated_q: torch.Tensor | None = None,
         replicated_k: torch.Tensor | None = None,
         replicated_v: torch.Tensor | None = None,
+        skip_sp: bool = False,
     ) -> torch.Tensor:
         """
         Forward pass for USPAttention.
 
             q, k, v: [B, S_local, H, D]
+
+        Args:
+            skip_sp: If True, skip all sequence parallelism (both Ulysses
+                and Ring) and compute local attention directly. Use this for
+                cross-attention where KV is replicated across all ranks.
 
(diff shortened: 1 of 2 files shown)


triple-mu:
I previously submitted a PR implementing a similar functionality:
https://github.com/sgl-project/sglang/pull/19340

Your PR works correctly when ring > 1 and ulysses = 1. However, once ulysses > 1, it becomes incorrect. This is because q is already sharded along the sequence dimension at the beginning, meaning each worker only holds a partial sequence.

If cross-attention is skipped directly in this case, the attention computation will use a partial q sequence together with full k/v sequences, which leads to incorrect results.

AichenF:
> I previously submitted a PR implementing a similar functionality: #19340
> 
> Your PR works correctly when ring > 1 and ulysses = 1. However, once ulysses > 1, it becomes incorrect. This is because q is already sharded along the sequence dimension at the beginning, meaning each worker only holds a partial sequence.
> 
> If cross-attention is skipped directly in this case, the attention computation will use a partial q sequence together with full k/v sequences, which leads to incorrect results.

thanks for your reply. happy to see that we found the same issue. I think when uly>1 the result is also correct, as q is sharded per rank, and k/v is replicated on each rank, the result depends only on local q and full kv. I also checked the result with uly4-skipsp: 

https://github.com/user-attachments/assets/c382f94f-18c0-4039-9edd-7849bde00981

mickqian (approved):

PR #19468, opened 2026-02-27
fix[minimax]: support deepep with minimax models

ishandhanani:
1. Update DeepEP to support hidden size of 3072
2. Force `dtype` to be `bfloat16` in order to fix error when using deepep. Before the default was float16 which was cause a DeepEP assertion error

---

Via this cmd on 8xb200

```bash
python3 -m sglang.launch_server     --model-path /opt/model/minimax-m25-fp8     --served-model-name MiniMax-M2.5     --trust-remote-code      --kv-cache-dtype fp8_e4m3     --page-size 128     --tp 4     --ep-size 4     --attention-backend flashinfer     --moe-a2a-backend deepep     --moe-dense-tp-size 1     --deepep-mode normal     --mem-fraction-static 0.85     --context-length 136000     --chunked-prefill-size 32768     --max-prefill-tokens 32768     --cuda-graph-max-bs 512     --stream-interval 30     --watchdog-timeout 1000000     --host 0.0.0.0     --enable-metrics --dtype bfloat16
```

Before the dockerfile change we would hit

```bash
DeepEP/csrc/kernels/internode_ll.cu:391 'false and "Unsupported hidden"'
```

After dockerfile change (before `dtype` change) we hit

```bash
DeepEP/csrc/deep_ep.cpp:1102 'x.dim() == 2 and x.is_contiguous() and x.scalar_type() == torch::kBFloat16'
```

After change success :)


ishandhanani:
@ispobock - can I get a quick look?

---

@zhaochenyang20

In your original MiniMax PR you mention in the description that 

```
only tried on 8 * H200. If adding:

--moe-a2a-backend deepep \
--deepep-mode auto
will fail.
```

This should fix

ispobock (approved):

python/sglang/srt/server_args.py
```diff
@@ -1643,6 +1643,12 @@ def _handle_model_specific_adjustments(self):
                             "Use flashinfer_trtllm as MoE runner backend on sm100 for Glm4MoeForCausalLM"
                         )
 
+        elif model_arch in ["MiniMaxM2ForCausalLM"]:
```
Fridge003, line 2120: Do we have accuracy results for minimax model after shifting dtype to bf16
Not sure whether this change will affect accuracy

ishandhanani:
> I also happen accuracy error:
> 
> Prefill
> 
> ```shell
> CUDA_VISIBLE_DEVICES=0,1,2,3 GLOO_SOCKET_IFNAME=eth0 MODEL_LENGTH=131072 NCCL_MIN_NCHANNELS=24 NCCL_IB_QPS_PER_CONNECTION=8 python -m sglang.launch_server --model-path /data00/models/MiniMax-M2.5 --tp-size 4  --tool-call-parser minimax-m2 --reasoning-parser minimax-append-think --host 0.0.0.0 --trust-remote-code --port 31100 --mem-fraction-static 0.8 --attention-backend fa3 --disable-radix-cache --disaggregation-mode prefill --disaggregation-ib-device  "mlx5_1,mlx5_2,mlx5_3,mlx5_4" --page-size 64
> ```
> 
> Decode
> 
> ```shell
> CUDA_VISIBLE_DEVICES=4,5,6,7 GLOO_SOCKET_IFNAME=eth0 MODEL_LENGTH=131072 NCCL_MIN_NCHANNELS=24 NCCL_IB_QPS_PER_CONNECTION=8 python -m sglang.launch_server --model-path /data00/models/MiniMax-M2.5 --tp-size 4 --ep-size 4 --tool-call-parser minimax-m2 --reasoning-parser minimax-append-think --host 0.0.0.0 --trust-remote-code --port 31200 --mem-fraction-static 0.8 --attention-backend fa3 --disable-radix-cache --disaggregation-mode decode --disaggregation-ib-device "mlx5_1,mlx5_2,mlx5_3,mlx5_4"  --cuda-graph-max-bs 128 --cuda-graph-bs 1 2 4 8 16 32 64 128 --max-running-requests 128 --moe-a2a-backend deepep --deepep-mode low_latency --page-size 64
> ```
> 
> Result
> 
> ```shell
> curl -X POST http://localhost:8080/v1/chat/completions -H "Content-Type: application/json" -d '{
>     "model": "/data00/models/MiniMax-M2.5",
>     "messages": [
>         {
>             "role": "user",
>             "content": "给一份北京出行攻略"
>         }
>     ],
>     "max_tokens": 500,
>     "temperature": 0.6
> }'
> {"id":"d3fdf190852b424a8088019896b83c09","object":"chat.completion","created":1773728755,"model":"/data00/models/MiniMax-M2.5","choices":[{"index":0,"message":{"role":"assistant","content":"<think>The|fecha|tipo|tipo فرض|tipo|timezone|fecha|fecha|tipocomposer\u0002","reasoning_content":null,"tool_calls":null},"logprobs":null,"finish_reason":"stop","matched_stop":"NaN happened"}],"usage":{"prompt_tokens":43,"total_tokens":55,"completion_tokens":12,"prompt_tokens_details":null,"reasoning_tokens":0},"metadata":{"weight_version":"default"}}
> ```
> 
> my prefill is TP4 and decode is deepep4, so your code is error: prefill dtype is None，decode type is bfloat16
> 
> ```shell
> if self.moe_a2a_backend == "deepep":
>       # When using DeepEP, we need to make sure activation dtype is bf16 and not float16
>       # otherwise DeepEP will error due to activation dtype mismatch.
>       self.dtype = "bfloat16"
> ```
> 
> so we fix the bug use this code:
> 
> ```shell
> #if self.moe_a2a_backend == "deepep":
> # When using DeepEP, we need to make sure activation dtype is bf16 and not float16
> # otherwise DeepEP will error due to activation dtype mismatch.
> self.dtype = "bfloat16"
> ```

Are you saying to always set activation dtype to `bfloat16` regardless of whether or not we use deepep?

DaZhUUU:
you can fix like this. The dtypes of prefill and decode must also be kept consistent during pd disaggregation
```
    elif model_arch in ["MiniMaxM2ForCausalLM"]:
            if self.moe_a2a_backend == "deepep" or self.disaggregation_mode != "null":
                self.dtype = "bfloat16"
```

PR #19474, opened 2026-02-27
[VLM] Per-chunk ViT forward to reduce multi-image prefill OOM

ConnorLi96:
## Motivation

SGLang supports chunked prefill to process long sequences in smaller token chunks. However, the vision encoder (ViT) was not respecting this chunking for multi-image requests: even when the current chunk only covers a subset of images, ViT was still running a forward pass over all images at once.

For a request with 10 images (~1024 tokens each, ~10K total) and `chunked_prefill_size=8192`:

**Expected (after this PR):**
- Chunk 1: tokens [0, 8192) → ViT runs on images 1–8 (8 images)
- Chunk 2: tokens [8192, 10240) → ViT runs on images 9–10 (2 images)

**Actual (before this PR):**
- Chunk 1: tokens [0, 8192) → ViT runs on ALL 10 images ← unnecessary
- Chunk 2: tokens [8192, 10240) → ViT runs on ALL 10 images ← unnecessary

This meant `chunked_prefill_size` provided no ViT memory benefit for multi-image requests. With many high-resolution images, the redundant all-at-once ViT forward could OOM even when the token budget was otherwise fine.

The function `_get_chunked_prefill_embedding_for_chunked_items` was already written to solve this — it selects only the images whose token spans overlap the current chunk before running ViT. But it was never called: the call site in `get_embedding_and_mask` always fell through to the legacy `_get_chunked_prefill_embedding`, which runs ViT on all items.

> **Note on video / EVS:** requests using EVS (Efficient Video Sampling) still use the legacy all-frames path. EVS must process all frames simultaneously to compute per-frame retention scores, and returns an `EVSEmbeddingResult` that rewrites `input_ids`. Both properties are incompatible with per-chunk processing. The per-chunk path returns `None` in these cases, automatically falling back to the legacy path.

### 1. Route non-EVS requests through `_get_chunked_prefill_embedding_for_chunked_items` (`get_embedding_and_mask`)
- Add `_has_evs_items()` helper to detect EVS items.
- Non-EVS requests now go through the per-chunk ViT path first; if it returns `None` (e.g. expand failure), fall back to the legacy all-frames `_get_chunked_prefill_embedding`.

### 2. Fix off-by-one in `items_offset` semantics (`get_embedding_items_per_chunk_with_extra_padding`, `get_embedding_chunk_remove_extra_padding`)
- `items_offset` uses inclusive end `[start, end]` (matching `base_processor.py build_input_ids()`), but both functions treated `end` as exclusive `[start, end)`.
- Fixed: `start >= end` → `start > end`, `end > window_start` → `end >= window_start`, `seg_len = end - start` → `end - start + 1`, `overlap_end = min(end, chunk_end)` → `min(end + 1, chunk_end)`.

### 3. Fix caching in `_get_chunked_prefill_embedding_for_chunked_items`
- Pass `item_hashes` list (not the pre-combined hash) to `embedding_cache.get()`.
- Wrap raw tensor in `EmbeddingResult` before `embedding_cache.set()` (cache expects `EmbeddingResult`).
- On cache hit, extract `.embedding` from `EmbeddingResult` and move to current CUDA device (CPU after post-chunk offload).

### 4. Enable `get_embedding_chunk_remove_extra_padding` trimming (was commented out)
- Correctly handles chunk boundaries that fall mid-frame (e.g. 576-token frames, 8192-token chunks).

### 5. Expand bundled `MultimodalDataItem` for per-chunk selection
- When `SGLANG_ENABLE_MM_SPLITTING` is `False` (default), the processor bundles N images into one `MultimodalDataItem` with N offsets. The chunk-selection logic requires 1-to-1 item/offset mapping.
- Call `get_new_expanded_mm_items()` to expand; if expansion fails, return `None` to fall back to the legacy path.

## Accuracy Tests

Tested with Qwen3.5-397B-A17B-FP4 (TP=4, `chunked_prefill_size=16384`):
- Single image: ✅ correct response
- 3 images in one request (bundled item path): ✅ correct response

## Checklist

### DIFF
--- a/python/sglang/srt/managers/mm_utils.py
+++ b/python/sglang/srt/managers/mm_utils.py
@@ -496,10 +496,10 @@ def get_embedding_items_per_chunk_with_extra_padding(
     Assumptions:
         - len(embedding_items_per_req) == len(items_offset)
         - items_offset[j] = (start, end), meaning the multimodal tokens of the j-th
-        item correspond to [start, end) (left-closed, right-open) in the entire
-        token sequence
+          item occupy token positions [start, end] (both inclusive) in the sequence,
+          matching the format produced by base_processor.py build_input_ids().
         - The item order in embedding_items_per_req is one-to-one aligned with
-        items_offset
+          items_offset
 
     Args:
         embedding_items_per_req: all items of this modality under the current
@@ -531,12 +531,11 @@ def get_embedding_items_per_chunk_with_extra_padding(
     selected_items: List["MultimodalDataItem"] = []
 
     for item, (start, end) in zip(embedding_items_per_req, items_offset):
-        if start >= end:
+        if start > end:
             continue
 
-        # Check whether this item has overlap with [window_start, window_end)
-        # If has overlap, add the item into selected_item.
-        if end > window_start and start < window_end:
+        # Check whether this item [start, end] overlaps [window_start, window_end)
+        if end >= window_start and start < window_end:
             selected_items.append(item)
 
     return selected_items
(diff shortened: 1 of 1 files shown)


python/sglang/srt/managers/mm_utils.py
```diff
@@ -726,94 +718,146 @@ def get_embedding_chunk_remove_extra_padding(
...
         )
 
         if not embedding_items_per_chunk:
             continue
 
-        # 2) construct cache key
-        # embedding_items_hash = MultiModalStaticCache.combine_hashes(
-        #     embedding_items_per_chunk
-        # )
+        logger.info(
```
ConnorLi96: ignore these debugging logs, we can delete all of them later

yuan-luo:
This is a common lib, could you test more VLMs? Some VLM the behavior can be different.

ConnorLi96:
it's still buggy, let's close it firstly.

PR #19478, opened 2026-02-27
[Feature] support mimo w8a8-int8 mixd-precision-quantization

baoqian426:
## Motivation
I am adapting the MiMo-V2-Flash model on a domestic (China-made) chip, using int8 quantization with mixed precision. For the MoE layers, I left layers 5, 6, 7, and 8 unquantized to preserve accuracy. During adaptation on sglang, I found that under int8 precision, if the v tensor in mimo_v2_flash.py is multiplied by v_scale (the attention value scale), the attention results become incorrect, so I set v_scale to None.

## Modifications
modify get_quant_method
modify mimo_v2_flash.py

## Accuracy Tests
FLASHINFER_DISABLE_VERSION_CHECK=1 SGLANG_ENABLE_SPEC_V2=1 python3 -m sglang.launch_server \
    --model-path /ssd1/models/MiMo-V2-Flash-INT8-Dynamic \
    --quantization w8a8_int8 \
    --max-total-tokens 65535 \
    --disable-radix-cache \
    --decode-log-interval 1 \
    --host 0.0.0.0 \
    --port 8806 \
    --trust-remote-code \
    --tp-size 8 \
    --cuda-graph-max-bs 32 \
    --max-running-requests 32 \
    --disable-overlap-schedule \
    --attention-backend fa3 \
    --mem-fraction-static 0.9

if v_scale = 0.707 from config.json
b4387c8640961a8372c2c7ee66471c9d

set v_scale=None
6713c7cf8cd6680756ff70971abb17a1

## Benchmarking and Profiling
FLASHINFER_DISABLE_VERSION_CHECK=1 SGLANG_ENABLE_SPEC_V2=1 python3 -m sglang.launch_server \
    --model-path /ssd1/models/MiMo-V2-Flash-INT8-Dynamic \
    --quantization w8a8_int8 \
    --max-total-tokens 65535 \
    --disable-radix-cache \
    --decode-log-interval 1 \
    --host 0.0.0.0 \
    --port 8806 \
    --trust-remote-code \
    --tp-size 8 \
    --cuda-graph-max-bs 32 \
    --max-running-requests 32 \
    --disable-overlap-schedule \
    --attention-backend fa3 \
    --mem-fraction-static 0.9
image

## Checklist

## Review Process


baoqian426:
> @baoqian426 Have you quant the value scale into the weights?

no I didn't,do you mean I should quant value scale? I found that value scale doesn’t seem to be a weight parameter — it’s a hyperparameter written in the config file, and I’m not sure whether I should quantize this parameter.

baoqian426:
> > > @baoqian426 Have you quant the value scale into the weights?
> > 
> > 
> > no I didn't,do you mean I should quant value scale?
> 
> It's not needed. So why set it to None when CompressedTensorsConfig or W8A8Int8Config?

Honestly, I don’t even know why — it’s just something I noticed.

baoqian426:
The reason why we should not multiply by v_scale again when running an INT8 model is now clear. The explanation is as follows:

We can equivalently absorb v_scale into the weight matrix 𝑊𝑉 beforehand. After INT8 quantization, the effect of v_scale is effectively canceled out. Therefore, during inference, we should not multiply by v_scale again.

The derivation is as follows:
image

From the above formula, we can see that the factor v_scale cancels out during INT8 quantization.

baoqian426:
> The reason why we should not multiply by v_scale again when running an INT8 model is now clear. The explanation is as follows:
> 
> We can equivalently absorb v_scale into the weight matrix 𝑊𝑉 beforehand. After INT8 quantization, the effect of v_scale is effectively canceled out. Therefore, during inference, we should not multiply by v_scale again.
> 
> The derivation is as follows: image
> 
> From the above formula, we can see that the factor v_scale cancels out during INT8 quantization.

@BBuf @b8zhong Could you please review and help with starting CI

baoqian426:
@AniZpZ @Edwardf0t1 @FlamingoPg @HaiShaw @ch-wan  Could you please review and help with starting CI

PR #12531, opened 2025-11-03
[CPU] Add mrope kernel for Qwen3-vl

blzheng:
## Motivation

This pr adds mrope kernel for Qwen3-vl on CPU.

Test plan:
python test/srt/cpu/test_rope.py -k test_mrope

## Checklist

- [ ] Format your code according to the [Format code with pre-commit](https://docs.sglang.ai/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.ai/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.ai/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.ai/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.ai/developer_guide/contribution_guide.html#benchmark-the-speed).

PR #19484, opened 2026-02-27
[CPU] Add Qwen3.5 model optimization for CPU

jianan-gu:
This PR (work with @blzheng ) adds support for Qwen3.5 series with cpu optimized performance, including changes:

1. Dtype support in fusion of `fused_sigmoid_gating_delta_rule_update`
2. Continues support for fusion of `fused_qkvzba_split_reshape_cat_contiguous_cpu `
3. TP cases padding support for both bf16 and fp8
4. Refinements for logging CPU padding logic.

Note that this PR depends on previous CPU mrope kernel support https://github.com/sgl-project/sglang/pull/12531/


python/sglang/srt/layers/attention/vision.py
```diff
@@ -763,7 +764,10 @@ def __init__(
         self.tp_size = 1 if use_data_parallel else get_attention_tp_size()
         self.tp_rank = 0 if use_data_parallel else get_attention_tp_rank()
         self.dropout = dropout
-        self.head_size = embed_dim // num_heads
+        if head_size is not None:
+            self.head_size = head_size
+        else:
+            self.head_size = embed_dim // num_heads
```
mingfeima:
```suggestion
        self.head_size = head_size if head_size is not None else embed_dim // num_heads
```

align with style of first few lines, use if ... else...

python/sglang/srt/models/qwen2_moe.py
```diff
@@ -551,7 +552,11 @@ def forward(
             forward_batch
         )
 
-        hidden_states = self.mlp(hidden_states, forward_batch, use_reduce_scatter)
+        hidden_states = self.mlp(
+            hidden_states,
+            forward_batch=forward_batch,
+            use_reduce_scatter=use_reduce_scatter,
+        )
```
mingfeima: is this needed?

python/sglang/srt/models/qwen2_moe.py
```diff
@@ -131,6 +131,7 @@ def __init__(
     def forward(
         self,
         x,
+        forward_batch: Optional[ForwardBatch] = None,
```
mingfeima: is this needed?

python/sglang/srt/models/qwen3_5.py
```diff
@@ -398,7 +410,11 @@ def forward(
             )
         )
         if isinstance(self.mlp, Qwen2MoeSparseMoeBlock):
-            hidden_states = self.mlp(hidden_states, forward_batch, use_reduce_scatter)
+            hidden_states = self.mlp(
+                hidden_states,
+                forward_batch=forward_batch,
+                use_reduce_scatter=use_reduce_scatter,
+            )
```
mingfeima, line 642: is this needed?

python/sglang/srt/models/qwen3_5.py
```diff
@@ -640,14 +656,17 @@ def forward(
...
                 forward_batch
             )
         )
         if isinstance(self.mlp, Qwen2MoeSparseMoeBlock):
-            hidden_states = self.mlp(hidden_states, forward_batch, use_reduce_scatter)
+            hidden_states = self.mlp(
+                hidden_states,
+                forward_batch=forward_batch,
+                use_reduce_scatter=use_reduce_scatter,
+            )
```
mingfeima, line 902: is this needed?

python/sglang/srt/models/qwen3_moe.py
```diff
@@ -802,7 +802,10 @@ def forward(
         )
 
         hidden_states = self.mlp(
-            hidden_states, forward_batch, should_allreduce_fusion, use_reduce_scatter
+            hidden_states,
+            forward_batch=forward_batch,
+            should_allreduce_fusion=should_allreduce_fusion,
+            use_reduce_scatter=use_reduce_scatter,
```
mingfeima: is this needed?

python/sglang/srt/models/qwen3_moe.py
```diff
@@ -845,7 +848,9 @@ def op_comm_prepare_mlp(self, state):
 
     def op_mlp(self, state):
         hidden_states = state.pop("hidden_states_mlp_input")
-        state.hidden_states_mlp_output = self.mlp(hidden_states, state.forward_batch)
+        state.hidden_states_mlp_output = self.mlp(
+            hidden_states, forward_batch=state.forward_batch
+        )
```
mingfeima: is this needed?

python/sglang/srt/models/qwen3_next.py
```diff
@@ -566,7 +566,11 @@ def forward(
         use_reduce_scatter = self.layer_communicator.should_use_reduce_scatter(
             forward_batch
         )
-        hidden_states = self.mlp(hidden_states, forward_batch, use_reduce_scatter)
+        hidden_states = self.mlp(
+            hidden_states,
+            forward_batch=forward_batch,
+            use_reduce_scatter=use_reduce_scatter,
+        )
```
mingfeima: is this needed?

python/sglang/srt/models/qwen3_next.py
```diff
@@ -805,7 +809,11 @@ def forward(
...
```
mingfeima: is this needed?

python/sglang/srt/utils/common.py
```diff
@@ -3099,6 +3099,16 @@ def log_info_on_rank0(logger, msg):
         logger.info(msg)
 
 
+def log_debug_on_rank0(logger, msg):
```
mingfeima, line 2873:
```suggestion
def log_debug_on_rank0(logger, msg):
    """
    Log a debug message only on tensor model parallel rank 0.
    Falls back to logging if distributed is not initialized or error occurs.
    """
```

sgl-kernel/csrc/cpu/mamba/fla.cpp
```diff
@@ -1272,35 +1326,69 @@ at::Tensor fused_sigmoid_gating_delta_rule_update_cpu(
...
+          k_strideS,
+          k_strideH,
+          v_strideB,
+          v_strideS,
+          v_strideH,
+          use_qk_l2norm_in_kernel,
+          softplus_threshold);
+    } else {
+      TORCH_CHECK(false, "Unsupported A_log dtype");
+    }
```
mingfeima: use `CPU_DISPATCH_REDUCED_FLOATING_TYPES_EXT` to simplify the code.

sgl-kernel/csrc/cpu/mamba/fla.cpp
```diff
@@ -1327,15 +1415,29 @@ fused_gdn_gating_cpu(const at::Tensor& A_log, const at::Tensor& a, const at::Ten
...
+          a.data_ptr<scalar_t>(),
+          b.data_ptr<scalar_t>(),
+          dt_bias.data_ptr<scalar_t>(),
+          out.data_ptr<float>(),
+          beta.data_ptr<scalar_t>(),
+          batch,
+          num_heads);
+    } else {
+      TORCH_CHECK(false, "Unsupported A_log dtype");
+    }
```
mingfeima: same as above.

mingfeima (changes requested):

python/sglang/srt/models/qwen2_moe.py
blzheng: This is no longer needed because the accuracy issue has already been resolved in the main branch in #19070.

mingfeima (approved):

mingfeima:
@Kangyan-Zhou could you please help review this one?

this one relates to frontend change, need sglang guys to review!

this PR is the last PR for Qwen 3.5 series enabling on CPU backend. After this one is settled, we will post blogs on this topic.

python/sglang/srt/configs/update_config.py
```diff
@@ -152,60 +197,100 @@ def adjust_config_with_unaligned_cpu_tp(
...
             )
-            from sglang.srt.layers.vocab_parallel_embedding import pad_vocab_size
+            if num_heads % tp_size != 0:
+                from sglang.srt.layers.vocab_parallel_embedding import pad_vocab_size
 
-            pad_size = get_num_heads_padding_size(tp_size, weight_block_size)
-            model_config.hf_config.vision_config.num_attention_heads = pad_vocab_size(
-                model_config.hf_config.vision_config.num_attention_heads, pad_size
+                pad_size = get_num_heads_padding_size(
+                    tp_size, weight_block_size, head_dim
```
JustinTong0323, line 280: `head_dim` is only assigned inside the `if num_attention_heads % tp_size != 0 or ...` block above (lines 200–206). If text heads happen to be TP-aligned but a vision/audio config hits `num_heads % tp_size != 0`, this raises `UnboundLocalError` (e.g. Qwen3.5-VL with text=32, vision=12, tp=8).

Also, even when it is defined, `head_dim` here is the *text* model's — applying it to vision/audio padding math is semantically off since their head dims typically differ.

Consider computing per-config head_dim inside this loop, e.g. `getattr(sub_cfg, "head_dim", sub_cfg.hidden_size // num_heads)`.

python/sglang/srt/model_loader/weight_utils.py
```diff
@@ -1559,3 +1563,29 @@ def narrow_padded_param_and_loaded_weight(
...
+    # while in some TP cases like TP6, output_sizes will be padded, thus loaded_weight needs padding.
+    total_output_size = sum(output_sizes)
+    raw_output_size = loaded_weight.size(output_dim)
+    if total_output_size > raw_output_size:
+        loaded_weight_pad = []
+        weight_split_size = [
+            int(output_size / total_output_size * raw_output_size)
+            for output_size in output_sizes
+        ]
+        split_weight = loaded_weight.split_with_sizes(weight_split_size, dim=output_dim)
```
JustinTong0323, line 1580: This ratio-based split is only correct when raw shards are equal-sized and the ratios divide cleanly. With mixed shard sizes (e.g. raw gate=24, up=20, padded to 32/32), `int(24/64 * 44) = 16` — you'd cut 16 rows for gate when the raw gate had 24, and the extra 8 rows silently land in `up`. No error, wrong numerics.

If the caller knows the raw shard sizes, it'd be safer to take them explicitly; otherwise at minimum assert `sum(weight_split_size) == raw_output_size` and push the rounding remainder onto the last shard. Same concern applies to the copy of this logic in `qwen3_5.py:338` and `mamba.py:80-83`.

python/sglang/srt/models/qwen3_vl.py
```diff
@@ -1108,7 +1127,12 @@ def __init__(
                 prefix=add_prefix("model.language_model", prefix),
             )
             if self.pp_group.is_last_rank:
-                if self.pp_group.world_size == 1 and self.config.tie_word_embeddings:
+                if (
+                    self.pp_group.world_size == 1
+                    and self.config.tie_word_embeddings
+                    and not _is_cpu
+                    and not _is_cpu_amx_available
```
JustinTong0323, line 1134: I think this condition is slightly off. `_is_cpu_amx_available` implies `_is_cpu`, so `not _is_cpu and not _is_cpu_amx_available` is equivalent to just `not _is_cpu` — CPU without AMX also falls into the `else` branch. But the weight-copy fallback in `qwen3_5.py:1498-1508` only runs on `_is_cpu and _is_amx_available`, so CPU-without-AMX ends up with an uninitialized `lm_head` and returns garbage logits — no error.

I think the intent is "don't tie on CPU+AMX". Try `and not (_is_cpu and _is_cpu_amx_available)` here, and make the weight-copy fallback cover the non-AMX CPU case too.

python/sglang/srt/models/qwen3_5.py
```diff
@@ -1475,6 +1495,17 @@ def load_weights(self, weights: Iterable[Tuple[str, torch.Tensor]]):
 
                 weight_loader = getattr(param, "weight_loader", default_weight_loader)
                 weight_loader(param, loaded_weight)
+                if (
+                    self.config.tie_word_embeddings
+                    and name == "model.embed_tokens.weight"
+                    and _is_cpu
+                    and _is_amx_available
```
JustinTong0323, line 1502: This `_is_amx_available` gate plus the change in `qwen3_vl.py` means CPU-without-AMX with `tie_word_embeddings=True` goes untied here *and* never gets a copy — `lm_head.weight` stays uninitialized. Worth making the two sides symmetric.

python/sglang/srt/models/qwen3_5.py
```diff
@@ -321,7 +329,17 @@ def weight_loader(param, loaded_weight, loaded_shard_id=None):
...
                     split_dim = getattr(param, "output_dim", 0)
-                    chunks = loaded_weight.split(split_sizes, dim=split_dim)
+                    if _is_cpu:
+                        cpu_split_sizes = []
+                        split_size_sum = sum(split_sizes)
+                        target_size_sim = loaded_weight.size(split_dim)
+                        for i in range(len(split_sizes)):
+                            cpu_split_sizes.append(
+                                int(target_size_sim * split_sizes[i] / split_size_sum)
+                            )
```
JustinTong0323, line 339: Same silent-truncation pattern as `pad_loaded_weight` in `weight_utils.py`. `int(target_size_sim * split_sizes[i] / split_size_sum)` can produce splits that don't sum to `target_size_sim` (→ `split_with_sizes` throws), and when they do sum by coincidence, the boundaries may still land in the wrong raw shard. Would be good to assert `sum(cpu_split_sizes) == target_size_sim` and distribute the remainder onto the last chunk.

sgl-kernel/csrc/cpu/model/qwen3.cpp
```diff
@@ -113,3 +148,54 @@ std::tuple<at::Tensor, at::Tensor, at::Tensor, at::Tensor> fused_qkvzba_split_re
...
+  CHECK_INPUT(mixed_ba);
+  int64_t batch = mixed_qkvz.size(0);
+  int64_t k_tp = num_heads_qk * head_qk;
+  int64_t v_tp = num_heads_v * head_v;
+  int64_t qkv_dim = k_tp * 2 + v_tp;
+  int64_t ba_dim = num_heads_v * 2;
+  int64_t expected_dim = qkv_dim + v_tp;
+  CHECK_EQ(mixed_qkvz.size(1), expected_dim);
+  CHECK_EQ(mixed_ba.size(0), batch);
+  CHECK_EQ(mixed_ba.size(1), ba_dim);
```
JustinTong0323, line 175: `mixed_ba`'s dtype isn't checked. The dispatch uses `mixed_qkvz.scalar_type()` and then the impl reads `mixed_ba.data_ptr<scalar_t>()` with that same `scalar_t`. If a caller ever passes mismatched dtypes (fp32 vs bf16) the bytes are reinterpreted silently.

Suggest adding `TORCH_CHECK(mixed_ba.scalar_type() == mixed_qkvz.scalar_type(), "mixed_ba and mixed_qkvz must share dtype")` alongside the other CHECK_EQs.

python/sglang/srt/layers/linear.py
```diff
@@ -735,6 +735,14 @@ def _load_fused_module_from_checkpoint(
...
             shard_offsets.append((i, current_shard_offset, output_size))
             current_shard_offset += output_size
+        if _is_cpu:
+            from sglang.srt.model_loader.weight_utils import (
+                pad_loaded_weight,
+            )
+
+            loaded_weight = pad_loaded_weight(
+                loaded_weight, param.output_dim, self.output_sizes
+            )
```
JustinTong0323, line 745: Line 734 resolves `output_sizes = output_sizes or self.output_sizes` and uses the local for `shard_offsets`, but here we fall back to `self.output_sizes`. If a caller overrides `output_sizes` (the whole reason the arg exists), the padded shape no longer lines up with the shard offsets computed above, so the later `narrow` reads the wrong slice.

Should be `pad_loaded_weight(loaded_weight, param.output_dim, output_sizes)`.

python/sglang/srt/utils/common.py
```diff
@@ -2723,6 +2723,20 @@ def log_info_on_rank0(logger, msg):
...
+    """
+    Log a debug message only on tensor model parallel rank 0.
+    Falls back to logging if distributed is not initialized or error occurs.
+    """
+    from sglang.srt.distributed import get_tensor_model_parallel_rank
+
+    try:
+        if torch.distributed.is_initialized() and get_tensor_model_parallel_rank() == 0:
+            logger.debug(msg)
+    except:
```
JustinTong0323, line 2736: Bare `except:` catches `KeyboardInterrupt` / `SystemExit` and then logs on every rank, silently breaking the rank-0-only contract. Same pattern at `log_info_on_rank0` (L2722).

Prefer narrowing to `except RuntimeError as e:` and including the error in the fallback message so a misconfigured PG is visible rather than invisible.

python/sglang/srt/layers/attention/mamba/mamba.py
```diff
@@ -110,7 +110,7 @@ def loader(param: torch.Tensor, loaded_weight: torch.Tensor) -> None:
 
             # CPU logic of padding size for qwen3-next
             # TODO : make this common for all mamba.
-            if is_cpu() and loaded_weight.size(0) % tp_size != 0:
+            if is_cpu() and (loaded_weight.size(0) < full_dim_sum):
```
JustinTong0323, line 113: The change from `% tp_size != 0` to `< full_dim_sum` makes this branch silently fire for *any* undersized tensor, including partial/misconfigured checkpoints — those would be zero-padded with no warning. Could we add a `log_info_on_rank0` here with `(loaded_weight.size(0), full_dim_sum, weight_full_dim_list)` so the path is at least visible in logs?

test/srt/cpu/test_qwen3.py
```diff
@@ -53,6 +53,34 @@ def fix_query_key_value_ordering_reshape_cat(
...
+
+    # value / z reshape to (seq, num_v_heads/tp, head_v_dim)
+    value = value.reshape(value.size(0), -1, head_v_dim)
+    z = z.reshape(z.size(0), -1, head_v_dim)
+    query, key, value = map(lambda x: x.reshape(x.shape[0], -1), (query, key, value))
+    mixed_qkv = torch.cat((query, key, value), dim=-1)
+    return mixed_qkv, z, b, a
+
+
 class TestQwen3(CustomTestCase):
```
JustinTong0323, line 84: This file (and `test_mamba.py`) doesn't register with `register_cpu_ci(...)`, and `test/run_suite.py` only scans `test/registered/**`. So these tests don't currently run in `stage-a-test-cpu` — any regression they'd catch would slip through CI.

Could they be moved under `test/registered/cpu/` with a `register_cpu_ci(est_time=N, suite="stage-a-test-cpu")` at module level (literal values required by the AST parser)? Also, the new kernel test only exercises `attn_tp_size=1`; parameterizing over a couple of TP sizes would cover the shard wiring.

JustinTong0323 (commented):
Nice work overall — new kernel, fake registration, and the bf16 dispatch for `fla.cpp` all look clean. A few things worth flagging before merge, mostly around the TP-padding/weight-loader paths where a couple of silent correctness bugs seem likely, plus the tied-lm_head branch for CPU-without-AMX. Inline comments below.

test/srt/cpu/test_qwen3.py
jianan-gu: Thanks for the suggestion! Actually those tests of` test_qwen3.py` or `test_mamba.py` and others are not registered in Xeon CI suite: https://github.com/sgl-project/sglang/blob/main/test/srt/run_suite.py#L64-L69.

We definitely can do those refinements to make them be in scope of `stage-a-test-cpu` with `register_cpu_ci` and under `test/registered/`, and also adding cases like more TP sizes. Actually we are WIP similar refactoring things like https://github.com/sgl-project/sglang/pull/22677, and may I know if it is ok that we catch those in follow-up PRs? Thanks!

python/sglang/srt/configs/update_config.py
jianan-gu: Thanks for that catch! Yes, here have refined those codes with logic of `getattr ` in this loop, and we do have to get `head_dim` for  multimodal specifically.

python/sglang/srt/model_loader/weight_utils.py
jianan-gu: yes, thank for that, have added ` assert sum(weight_split_size) == raw_output_size` here and also `qwen3_5.py:338 `and `mamba.py:80-83`, to get avoid silently rounding issue.

python/sglang/srt/models/qwen3_vl.py
jianan-gu: Yes, good catch, we should make that with `and not (_is_cpu and _is_cpu_amx_available)`, have changed codes accordingly

python/sglang/srt/models/qwen3_5.py
jianan-gu: have added  assert `sum(cpu_split_sizes) == target_size_sim ` thanks.

sgl-kernel/csrc/cpu/model/qwen3.cpp
jianan-gu: yes, thanks for reminding that, the assumption should be checked, have added accordingly

python/sglang/srt/layers/attention/mamba/mamba.py
jianan-gu: yes, have added logging to catch that sizes changes info to avoid silent padding. Thanks!

python/sglang/srt/utils/common.py
jianan-gu: refined accordingly, thanks for the suggestion.

python/sglang/srt/layers/linear.py
jianan-gu: Yes, exactly, have corrected that, thanks!

mingfeima:
@jianan-gu fail at https://github.com/sgl-project/sglang/actions/runs/24720109083/job/72306535388?pr=19484#step:6:2541. rebase and test again.

JustinTong0323 (commented):
A few issues in the follow-up — the three new asserts use `assert (cond, msg,)` which Python parses as asserting a non-empty tuple, so they never fire (also triggers `SyntaxWarning: assertion is always true`). And the two new `TORCH_CHECK` lines in `qwen3.cpp` are missing trailing `;`.

python/sglang/srt/model_loader/weight_utils.py
```diff
@@ -1559,3 +1563,33 @@ def narrow_padded_param_and_loaded_weight(
...
+    if total_output_size > raw_output_size:
+        loaded_weight_pad = []
+        weight_split_size = [
+            int(output_size / total_output_size * raw_output_size)
+            for output_size in output_sizes
+        ]
+        assert (
+            sum(weight_split_size) == raw_output_size,
+            f"Padding the loaded weight failed due to sizes are not dividable cleanly from {output_sizes} to {raw_output_size}",
+        )
```
JustinTong0323: `assert (cond, msg,)` asserts a non-empty tuple — always truthy, never fires. Drop the outer parens:
```python
assert sum(weight_split_size) == raw_output_size, (
    f"Padding ... from {output_sizes} to {raw_output_size}"
)
```
Also nit: "dividable" → "divisible".

python/sglang/srt/models/qwen3_5.py
```diff
@@ -321,7 +329,21 @@ def weight_loader(param, loaded_weight, loaded_shard_id=None):
...
+                            assert (
+                                sum(cpu_split_sizes) == target_size_sim,
+                                f"Padding the loaded weight failed due to sizes are not dividable cleanly from {cpu_split_sizes} to {target_size_sim}",
+                            )
```
JustinTong0323: Two bugs here:

1. `assert (cond, msg,)` is tuple-form — non-empty tuple is always truthy, assert never fires.
2. Even once fixed, the assert is *inside* the `for i in range(len(split_sizes))` loop, so it'd check the partial sum after each `append` and fail for any multi-shard case.

Move it out of the loop and drop the parens:
```python
for i in range(len(split_sizes)):
    cpu_split_sizes.append(int(target_size_sim * split_sizes[i] / split_size_sum))
assert sum(cpu_split_sizes) == target_size_sim, (
    f"..."
)
```

python/sglang/srt/layers/attention/mamba/mamba.py
```diff
@@ -81,6 +90,10 @@ def loader(param: torch.Tensor, loaded_weight: torch.Tensor) -> None:
                 weight_full_dim_list.append(
                     int(full_dim / full_dim_sum * loaded_weight.size(0))
                 )
+            assert (
+                sum(weight_full_dim_list) == loaded_weight.size(0),
+                f"Padding the loaded weight failed due to sizes are not dividable cleanly from {weight_full_dim_list} to {loaded_weight.size(0)}",
+            )
```
JustinTong0323: Same tuple-form assert bug — always truthy, never fires. `assert sum(...) == loaded_weight.size(0), f"..."` (no outer parens).

sgl-kernel/csrc/cpu/model/qwen3.cpp
```diff
@@ -84,6 +119,7 @@ std::tuple<at::Tensor, at::Tensor, at::Tensor, at::Tensor> fused_qkvzba_split_re
   CHECK_EQ(mixed_qkvz.size(1), expected_dim);
   CHECK_EQ(mixed_ba.size(0), batch);
   CHECK_EQ(mixed_ba.size(1), ba_dim);
+  TORCH_CHECK(mixed_ba.scalar_type() == mixed_qkvz.scalar_type(), "mixed_ba and mixed_qkvz must share same dtype")
```
JustinTong0323: Missing trailing `;` (same thing on L177). `TORCH_CHECK` expands to a `do { ... } while(0)` and the call-site needs its own terminator — every other `TORCH_CHECK` / `CHECK_EQ` in this file ends with `;`.

Also — since this is just a scalar-type equality, you could reuse the existing `CHECK_EQ(mixed_ba.scalar_type(), mixed_qkvz.scalar_type());` macro to keep the style consistent with the surrounding checks.

python/sglang/srt/layers/attention/mamba/mamba.py
```diff
@@ -110,7 +123,12 @@ def loader(param: torch.Tensor, loaded_weight: torch.Tensor) -> None:
 
             # CPU logic of padding size for qwen3-next
             # TODO : make this common for all mamba.
-            if is_cpu() and loaded_weight.size(0) % tp_size != 0:
+            if is_cpu() and (loaded_weight.size(0) < full_dim_sum):
+                log_info_on_rank0(
+                    logger,
+                    f"Loaded_weight.dim(0) size:{loaded_weight.size(0)} is padding to {full_dim_sum}"
+                    f", where original sizes of {weight_full_dim_list} will be updated to  {full_dim_list}",
+                )
```
JustinTong0323: The log is loop-invariant (all referenced values are fixed before `for full_dim, ...`) but sits inside the shard-spec loop, so it fires 3× per mamba layer during weight load — a 48-layer model gets ~144 duplicate INFO lines on TP-odd startup. Consider hoisting it to before the loop, or gating with a `logged_once` flag.

Also the message has a double space before `{full_dim_list}`.

python/sglang/srt/utils/common.py
```diff
@@ -2793,8 +2793,22 @@ def log_info_on_rank0(logger, msg):
     try:
         if torch.distributed.is_initialized() and get_tensor_model_parallel_rank() == 0:
             logger.info(msg)
-    except:
-        logger.info(msg)
+    except RuntimeError as e:
+        logger.info(msg + f", and meeting Error with {e}")
```
JustinTong0323: Two things:

1. `msg + f", and meeting Error with {e}"` is awkward phrasing + mid-sentence `Error` capitalization. Something like `f"{msg} (rank-check failed: {e})"` reads cleaner.
2. The fallback still logs on *every* rank when `get_tensor_model_parallel_rank()` raises, which defeats the `_on_rank0` contract — not new to this PR but worth fixing while you're here. `logger.debug` for the fallback, or actually resolve rank via `torch.distributed.get_rank()` before giving up.

python/sglang/srt/configs/update_config.py
```diff
@@ -152,60 +197,105 @@ def adjust_config_with_unaligned_cpu_tp(
...
+            if num_heads % tp_size != 0:
+                from sglang.srt.layers.vocab_parallel_embedding import pad_vocab_size
 
-            pad_size = get_num_heads_padding_size(tp_size, weight_block_size)
-            model_config.hf_config.vision_config.num_attention_heads = pad_vocab_size(
-                model_config.hf_config.vision_config.num_attention_heads, pad_size
+                multimodal_head_dim = getattr(
+                    getattr(m_config, config_name),
+                    "head_dim",
+                    getattr(m_config, config_name).hidden_size // num_heads,
```
JustinTong0323: This duplicates the head_dim-resolution logic from lines 200-206 of this same function (which also handles `qk_head_dim` / `hf_text_config.head_dim` fallbacks). Worth extracting a small `resolve_head_dim(cfg, num_heads)` helper and using it in both spots — otherwise the two resolvers will drift.

python/sglang/srt/model_loader/weight_utils.py
jianan-gu: revised, thanks!

python/sglang/srt/models/qwen3_5.py
jianan-gu: revised as well, thanks!

python/sglang/srt/layers/attention/mamba/mamba.py
jianan-gu: yes, have reduced the log lines by moving out of the loop. Thanks.

python/sglang/srt/utils/common.py
jianan-gu: Thanks. Revised here for the fallback to be (1) if it is distributed, message only on rank0 (2) if distributed is not init, should only be like TP=1, message directly.

python/sglang/srt/configs/update_config.py
jianan-gu: yes, have revised with a `resolve_head_dim` to cover both.

python/sglang/srt/configs/update_config.py
```diff
@@ -53,6 +66,24 @@ def get_num_heads_padding_size(tp_size, weight_block_size, head_dim):
     return pad_size
 
 
+def resolve_head_dim(cfg, num_heads, is_text_config):
+    # default getting head_dim by hidden_size and num_heads
+    head_dim = cfg.hidden_size // num_heads
```
JustinTong0323: `resolve_head_dim` unconditionally does `cfg.hidden_size // num_heads` as its default, but the multimodal loop feeds `audio_config` in here (L281-283), and `Qwen3OmniMoeAudioEncoderConfig` uses `d_model=1280`, no `hidden_size` attribute (`python/sglang/srt/configs/qwen3_omni.py:16,34`). So on Qwen3-Omni with `encoder_attention_heads % tp_size != 0` (e.g. tp=3/6/7), this raises `AttributeError` before it can fall through to the `hasattr(cfg, "head_dim")` check.

Suggest:
```python
hidden_size = getattr(cfg, "hidden_size", getattr(cfg, "d_model", None))
head_dim = hidden_size // num_heads if hidden_size else None
```
and then let the `hasattr(cfg, "head_dim")` / `hf_config.head_dim` cascade override.

python/sglang/srt/utils/common.py
```diff
@@ -2793,8 +2793,30 @@ def log_info_on_rank0(logger, msg):
     try:
         if torch.distributed.is_initialized() and get_tensor_model_parallel_rank() == 0:
             logger.info(msg)
-    except:
-        logger.info(msg)
+    except RuntimeError as e:
+        if torch.distributed.is_initialized():
```
JustinTong0323, line 2866: Narrowing bare `except:` → `except RuntimeError` is a regression here. `get_tensor_model_parallel_rank()` calls `get_tp_group()` which raises **`AssertionError`** (not RuntimeError) when `_TP is None` (see `distributed/parallel_state.py:1463`). Previously bare-except swallowed it; now it propagates and kills the caller — exactly what the helper is supposed to shield against during early startup / shutdown.

Either `except Exception as e:` here (and in `log_debug_on_rank0`), or explicitly catch `(RuntimeError, AssertionError)`.

python/sglang/srt/layers/attention/mamba/mamba.py
```diff
@@ -81,6 +90,15 @@ def loader(param: torch.Tensor, loaded_weight: torch.Tensor) -> None:
...
                 )
+            assert sum(weight_full_dim_list) == loaded_weight.size(
+                0
+            ), f"Padding the loaded weight failed due to sizes are not divisible cleanly from {weight_full_dim_list} to {loaded_weight.size(0)}"
+            if loaded_weight.size(0) < full_dim_sum:
+                log_info_on_rank0(
+                    logger,
+                    f"Loaded_weight.dim(0) size:{loaded_weight.size(0)} is padding to {full_dim_sum}"
+                    f", where original sizes of {weight_full_dim_list} will be updated to {full_dim_list}",
+                )
```
JustinTong0323, line 99: Nit — zero-padding a checkpoint when `loaded_weight.size(0) < full_dim_sum` is a correctness-changing action (extra heads become permanent zeros). INFO is easy to miss during a chatty startup; `logger.warning` with the word `ZERO-PADDING` in the message would make accidental cases (user loaded a mismatched checkpoint) actually noticeable.

JustinTong0323 (commented):
Round-2 fixes all landed cleanly — asserts, TORCH_CHECK, head_dim resolver, rank-0 logging all look good. Two new things surfaced on a final pass, one of which is a crash on Qwen3-Omni with unaligned TP.

python/sglang/srt/configs/update_config.py
jianan-gu: Thanks, yes it is, have revised this part with the suggestion.

python/sglang/srt/utils/common.py
jianan-gu: revised with `except Exception as e:` here and also `log_debug_on_rank0`, thanks!

python/sglang/srt/layers/attention/mamba/mamba.py
jianan-gu: yes, make sense, have revised here with `logger.warning` (on rank0)

JustinTong0323 (approved):

mickqian (approved):

sgl-kernel/csrc/cpu/mamba/fla.cpp
fadara01, line 918: I think this should be:
```
float beta_val = 1 / (1 + std::exp(-b_ptr[bi * v_num_heads + ni]));
```
similar to `a_ptr` on L907

blzheng: Yes, you're correct. Thanks for pointing out this bug. It will be fixed in this PR. https://github.com/sgl-project/sglang/pull/26634

PR #19516, opened 2026-02-27
[Diffusion] add cuda graph support for Qwen-Image

zyksir:
## Motivation

For some of the profile, you can refer to [This document](https://uclabruins.feishu.cn/wiki/QXbtwSYfVipIiwkjsircPYZVnTh).

**Background:** Diffusion models typically have long sequence lengths. For some models (Wan, FLUX), kernels can be launched ahead of time, so CPU overhead is relatively small even without CUDA Graphs. In this case we will focus on Qwen-Image, where we observe ~30% cpu overhead on the profile. (For the profile, you can check [This document](https://uclabruins.feishu.cn/wiki/QXbtwSYfVipIiwkjsircPYZVnTh). )

image

## Modifications

There are several challenges of add cuda graph for diffusion models.

**Challenge 1:** In image generation, input shapes are tied to (width, height, prompt_length). In theory, users can pass a different (width, height) per request, making it impossible to pre-build graphs exhaustively. In practice, different customers have different size requirements, and occasionally someone wants to try a new size. prompt_length is a bit more manageable — most models cap it at a max length (e.g., Qwen-Image caps at 512).

** Challenge 2 **: Diffusion kernels are compute-bound, so padding on (width, height) has a significant performance cost. Padding on prompt_length has a relatively smaller impact on end-to-end performance.

** Current Solution **: 
- Current approach requires modifying `dit/<model>.py`, so it can't be generalized (given Background 1, if a new model shows no CPU overhead after profiling, CUDA Graph support isn't necessary anyway).
- The idea is to split the `img_process` and `txt_process` logic within a block into two separate functions, capture CUDA Graphs for each independently, and apply padding only to txt.

## Accuracy Tests

The command is used to test accuracy and end-to-end speed.
```shell
CUDA_VISIBLE_DEVICES=5 sglang generate --model-path Qwen/Qwen-Image \
    --prompt "A Cat Holding Logo With Bold Large Text: SGL Diffusion" \
    --num-inference-steps 50 \
    --guidance-scale 4.0 \
    --width 1024 --height 1024 \
    --dit-cpu-offload false \
    --text-encoder-cpu-offload false \
    --vae-cpu-offload false \
    --save-output \
    --seed 42 \
    --profile --num-profiled-timesteps 2 --enable-cuda-graph true
```

One of them is using cuda graph, one is not. They are the same
image
image

## Benchmarking and Profiling

We can see from profile that 
image

## Checklist

## Review Process

### DIFF
--- a/python/sglang/multimodal_gen/runtime/models/dits/qwen_image.py
+++ b/python/sglang/multimodal_gen/runtime/models/dits/qwen_image.py
@@ -51,6 +51,7 @@ from sglang.multimodal_gen.runtime.platforms import (
 )
 from sglang.multimodal_gen.runtime.utils.layerwise_offload import OffloadableDiTMixin
 from sglang.multimodal_gen.runtime.utils.logging_utils import init_logger
+from sglang.srt.utils.common import next_power_of_2
 
 logger = init_logger(__name__)  # pylint: disable=invalid-name
 _is_cuda = current_platform.is_cuda()
@@ -802,22 +803,9 @@ class QwenImageCrossAttention(nn.Module):
         return img_attn_output, txt_attn_output
 
 
-def next_power_of_2_bitwise(n):
-    if n <= 0:
-        return 1
-    n -= 1  # so that n=32 returns 32, not 64
-    n |= n >> 1
-    n |= n >> 2
-    n |= n >> 4
-    n |= n >> 8
-    n |= n >> 16
-    return n + 1
-
-
-def pad_tensor(tensor: torch.Tensor, dim: int) -> torch.Tensor:
-    assert dim == 0 or dim == 1, "dim must be 0 or 1"
+def pad_tensor_into_power_of_2(tensor: torch.Tensor, dim: int) -> torch.Tensor:
     seq_length = tensor.shape[dim]
-    padded_length = next_power_of_2_bitwise(seq_length)
+    padded_length = next_power_of_2(seq_length)
     padding_length = padded_length - seq_length
     if padding_length == 0:
         return tensor
@@ -1218,13 +1206,13 @@ class QwenImageTransformerBlock(nn.Module):
             img_modulated, img_gate1, img_mod2 = module(hidden_states, temb_img_silu)
 
         seq_len_txt = encoder_hidden_states.shape[1]
-        padded_seq_len_txt = next_power_of_2_bitwise(seq_len_txt)
+        padded_seq_len_txt = next_power_of_2(seq_len_txt)
         if not self._enable_cuda_graph_capture:
             txt_modulated, txt_gate1, txt_mod2 = self._txt_pre_attention_forward(
                 encoder_hidden_states, temb_txt_silu
             )
         else:
-            padded_encoder_hidden_states = pad_tensor(
+            padded_encoder_hidden_states = pad_tensor_into_power_of_2(
                 encoder_hidden_states, 1
             )  # [B, S, H]
             if (
@@ -1331,7 +1319,7 @@ class QwenImageTransformerBlock(nn.Module):
                 txt_attn_output, txt_mod2, txt_gate1, encoder_hidden_states
             )
         else:
-            padded_txt_attn_output = pad_tensor(txt_attn_output, dim=1)
+            padded_txt_attn_output = pad_tensor_into_power_of_2(txt_attn_output, dim=1)
             if (
                 "txt_post_attention_forward",
                 txt_attn_output.shape,
--- a/python/sglang/multimodal_gen/runtime/models/dits/qwen_image.py
+++ b/python/sglang/multimodal_gen/runtime/models/dits/qwen_image.py
@@ -802,7 +802,33 @@ class QwenImageCrossAttention(nn.Module):
         return img_attn_output, txt_attn_output
 
 
+def next_power_of_2_bitwise(n):
+    if n <= 0:
+        return 1
+    n -= 1  # so that n=32 returns 32, not 64
+    n |= n >> 1
+    n |= n >> 2
+    n |= n >> 4
+    n |= n >> 8
+    n |= n >> 16
+    return n + 1
+
+def pad_tensor(tensor: torch.Tensor, dim: int) -> torch.Tensor:
+    assert dim == 0 or dim == 1, "dim must be 0 or 1"
+    seq_length = tensor.shape[dim]
+    padded_length = next_power_of_2_bitwise(seq_length)
+    padding_length = padded_length - seq_length
+    if padding_length == 0:
+        return tensor
+    if dim == 0:
+        padding = torch.zeros(padding_length, *tensor.shape[1:], device=tensor.device, dtype=tensor.dtype)
+    else:
+        padding = torch.zeros(*tensor.shape[:dim], padding_length, *tensor.shape[dim+1:], device=tensor.device, dtype=tensor.dtype)
+    return torch.cat([tensor, padding], dim=dim)
+
 class QwenImageTransformerBlock(nn.Module):
+    GLOBAL_GRAPH_POOL_HANDLE = torch.cuda.graphs.graph_pool_handle()
+
     def __init__(
         self,
         dim: int,
@@ -902,6 +928,12 @@ class QwenImageTransformerBlock(nn.Module):
             }
             self.img_mlp = NunchakuFeedForward(self.img_mlp, **nunchaku_kwargs)
             self.txt_mlp = NunchakuFeedForward(self.txt_mlp, **nunchaku_kwargs)
+        
+        self._enable_cuda_graph_capture = False
+        self._cuda_graphs = {}
+    
+    def enable_cuda_graph_capture(self, enable: bool):
+        self._enable_cuda_graph_capture = enable
 
     def _modulate(
         self,
@@ -994,21 +1026,15 @@ class QwenImageTransformerBlock(nn.Module):
             else:
                 modulated = norm_module(x=x, shift=shift_result, scale=scale_result)
                 return modulated, gate_result
-
-    def forward(
+    
+    def _img_pre_attention_forward(
         self,
         hidden_states: torch.Tensor,
-        encoder_hidden_states: torch.Tensor,
-        encoder_hidden_states_mask: torch.Tensor,
         temb_img_silu: torch.Tensor,
-        temb_txt_silu: torch.Tensor,
-        image_rotary_emb: Optional[Tuple[torch.Tensor, torch.Tensor]] = None,
-        joint_attention_kwargs: Optional[Dict[str, Any]] = None,
         modulate_index: Optional[List[int]] = None,
-    ) -> Tuple[torch.Tensor, torch.Tensor]:
-        # Get modulation parameters for both streams
+    ) -> torch.Tensor:
+        # Get modulation parameters for image stream
         img_mod_params = self.img_mod[1](temb_img_silu)[0]  # [B, 6*dim]
-        txt_mod_params = self.txt_mod[1](temb_txt_silu)[0]  # [B, 6*dim]
 
         if (
             self.quant_config is not None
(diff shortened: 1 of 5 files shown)


python/sglang/multimodal_gen/runtime/models/dits/qwen_image.py
```diff
@@ -819,7 +803,30 @@ def forward(
...
+        )
+    else:
+        padding = torch.zeros(
+            *tensor.shape[:dim],
+            padding_length,
+            *tensor.shape[dim + 1 :],
+            device=tensor.device,
+            dtype=tensor.dtype,
+        )
+    return torch.cat([tensor, padding], dim=dim)
```
ping1jing2, line 824: shall we use [F.pad](https://docs.pytorch.org/docs/stable/generated/torch.nn.functional.pad.html) instead of `zeros` and `cat`

python/sglang/multimodal_gen/runtime/models/dits/qwen_image.py
```diff
@@ -819,7 +803,30 @@ def forward(
...
+
+
 class QwenImageTransformerBlock(nn.Module):
+    GLOBAL_GRAPH_POOL_HANDLE = torch.cuda.graphs.graph_pool_handle()
```
ping1jing2, line 828: could you please use `torch.get_device_module` instead of `torch.cuda`

python/sglang/multimodal_gen/runtime/models/dits/qwen_image.py
```diff
@@ -1058,46 +1046,56 @@ def forward(
...
+        self,
+        encoder_hidden_states: torch.Tensor,
+        temb_txt_silu: torch.Tensor,
+    ) -> torch.Tensor:
+        # Get modulation parameters for text stream
+        txt_mod_params = self.txt_mod[1](temb_txt_silu)[0]  # [B, 6*dim]
+
+        if (
+            self.quant_config is not None
+            and hasattr(self.quant_config, "get_name")
```
ping1jing2, line 1069: please avoid using `hasattr/setattr`

python/sglang/multimodal_gen/runtime/models/dits/qwen_image.py
```diff
@@ -1134,8 +1144,211 @@ def forward(
...
+                hidden_states.shape,
+                modulate_index.shape,
+            ) not in self._cuda_graphs:
+                self._cuda_graphs[
+                    (
+                        "img_pre_attention_forward",
+                        hidden_states.shape,
+                        modulate_index.shape,
+                    )
+                ] = torch.cuda.make_graphed_callables(
```
ping1jing2, line 1177: ditto

python/sglang/multimodal_gen/runtime/models/dits/qwen_image.py
```diff
@@ -1134,8 +1144,211 @@ def forward(
...
+            _img_pre_attention = functools.partial(
+                self._img_pre_attention_forward, modulate_index=None
+            )
+            if (
+                "img_pre_attention_forward",
+                hidden_states.shape,
+            ) not in self._cuda_graphs:
+                self._cuda_graphs[
+                    ("img_pre_attention_forward", hidden_states.shape)
+                ] = torch.cuda.make_graphed_callables(
```
ping1jing2, line 1198: ditto

python/sglang/multimodal_gen/runtime/models/dits/qwen_image.py
```diff
@@ -1134,8 +1144,211 @@ def forward(
...
+        else:
+            padded_encoder_hidden_states = pad_tensor_into_power_of_2(
+                encoder_hidden_states, 1
+            )  # [B, S, H]
+            if (
+                "txt_pre_attention_forward",
+                padded_seq_len_txt,
+            ) not in self._cuda_graphs:
+                self._cuda_graphs[("txt_pre_attention_forward", padded_seq_len_txt)] = (
+                    torch.cuda.make_graphed_callables(
```
ping1jing2, line 1223: ditto

python/sglang/multimodal_gen/runtime/models/dits/qwen_image.py
```diff
@@ -1134,8 +1144,211 @@ def forward(
...
+                img_attn_output.shape,
+                modulate_index.shape,
+            ) not in self._cuda_graphs:
+                self._cuda_graphs[
+                    (
+                        "img_post_attention_forward",
+                        img_attn_output.shape,
+                        modulate_index.shape,
+                    )
+                ] = torch.cuda.make_graphed_callables(
```
ping1jing2, line 1273: ditto

python/sglang/multimodal_gen/runtime/models/dits/qwen_image.py
```diff
@@ -1134,8 +1144,211 @@ def forward(
...
+            ) not in self._cuda_graphs:
+                self._cuda_graphs[
+                    ("img_post_attention_forward", img_attn_output.shape)
+                ] = torch.cuda.make_graphed_callables(
+                    _img_post_attention,
+                    (img_attn_output, img_mod2, img_gate1, hidden_states),
+                    pool=self.GLOBAL_GRAPH_POOL_HANDLE,
+                )
+                logger.debug(
+                    f"memory usage after captured img_post_attention_forward: {torch.cuda.memory_allocated() / 1024 / 1024 / 1024:.2f} GB"
```
ping1jing2, line 1310: maybe you can use `current_platform.get_available_gpu_memory()` here

python/sglang/multimodal_gen/runtime/models/dits/qwen_image.py
```diff
@@ -1134,8 +1144,211 @@ def forward(
...
+        self,
+        hidden_states: torch.Tensor,
+        encoder_hidden_states: torch.Tensor,
+        encoder_hidden_states_mask: torch.Tensor,
+        temb_img_silu: torch.Tensor,
+        temb_txt_silu: torch.Tensor,
+        image_rotary_emb: Optional[Tuple[torch.Tensor, torch.Tensor]] = None,
+        joint_attention_kwargs: Optional[Dict[str, Any]] = None,
+        modulate_index: Optional[List[int]] = None,
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
```
ping1jing2, line 1160: I am not sure if i am right, but i think there are some issues here:
1. It seems like you make `forward` more complicated.
2. The key of `self._cuda_graphs` is not user-friendly

python/sglang/multimodal_gen/runtime/pipelines_core/stages/denoising.py
```diff
@@ -129,6 +130,22 @@ def __init__(
...
+    def _maybe_enable_cuda_graph_capture(
+        self, server_args: ServerArgs, module: object
+    ) -> None:
+        """
+        Enable CUDA graph capture on the module when server_args.enable_cuda_graph is set.
+        No-op if disabled or the module does not support it (e.g. only QwenImageTransformer2DModel).
+        """
+        if not getattr(server_args, "enable_cuda_graph", False):
+            return
+        if not hasattr(module, "enable_cuda_graph_capture"):
```
ping1jing2, line 142: avoid using `gettar/hasattr`

ping1jing2:
please provide a performance comparison report refer to `docs\diffusion\contributing.md`

1.  **Baseline**: run the benchmark (for a single generation task)
    ```bash
    $ sglang generate --model-path <model> --prompt "A benchmark prompt" --perf-dump-path baseline.json
    ```

2.  **New**: run the same benchmark, without modifying any server_args or sampling_params
    ```bash
    $ sglang generate --model-path <model> --prompt "A benchmark prompt" --perf-dump-path new.json
    ```

3.  **Compare**: run the compare script, which will print a Markdown table to the console
    ```bash
    $ python python/sglang/multimodal_gen/benchmarks/compare_perf.py baseline.json new.json [new2.json ...]
    ### Performance Comparison Report
    ...
    ```
4. **Paste**: paste the table into the PR description

python/sglang/multimodal_gen/runtime/models/dits/qwen_image.py
```diff
@@ -1134,8 +1144,211 @@ def forward(
...
+
+        if not self._enable_cuda_graph_capture:
+            encoder_hidden_states = self._txt_post_attention_forward(
+                txt_attn_output, txt_mod2, txt_gate1, encoder_hidden_states
+            )
+        else:
+            padded_txt_attn_output = pad_tensor_into_power_of_2(txt_attn_output, dim=1)
+            if (
+                "txt_post_attention_forward",
+                txt_attn_output.shape,
```
BBuf, line 1325: `txt_post_attention_forward` already pads the text activations to a power-of-2 length, but the graph cache key still uses the original `txt_attn_output.shape`. That means different sequence lengths that fall into the same padded bucket will still capture separate graphs, which reduces the benefit of the bucketing logic and increases graph/memory overhead.

python/sglang/multimodal_gen/runtime/models/dits/qwen_image.py
```diff
@@ -939,6 +927,12 @@ def __init__(
             self.img_mlp = NunchakuFeedForward(self.img_mlp, **nunchaku_kwargs)
             self.txt_mlp = NunchakuFeedForward(self.txt_mlp, **nunchaku_kwargs)
 
+        self._enable_cuda_graph_capture = False
+        self._cuda_graphs = {}
```
BBuf, line 931: `_cuda_graphs` is an unbounded per-block cache with no eviction or reset path. Since it stores graphed callables rather than lightweight metadata, a long-running service can keep accumulating graph memory as it sees more batch/shape/text-bucket combinations.

python/sglang/multimodal_gen/runtime/models/dits/qwen_image.py
```diff
@@ -819,7 +803,30 @@ def forward(
         return img_attn_output, txt_attn_output
 
 
+def pad_tensor_into_power_of_2(tensor: torch.Tensor, dim: int) -> torch.Tensor:
```
BBuf, line 806: Could this function be moved into a shared utility file for CUDA graph support?

python/sglang/multimodal_gen/runtime/pipelines_core/stages/denoising.py
```diff
@@ -129,6 +130,22 @@ def __init__(
         self._cached_num_steps = None
         self._is_warmed_up = False
 
+    def _maybe_enable_cuda_graph_capture(
```
BBuf, line 133: nits: Can this check be moved up to the server_args stage?

python/sglang/multimodal_gen/runtime/models/dits/qwen_image.py
```diff
@@ -1134,8 +1144,211 @@ def forward(
         # Clip to prevent overflow for fp16
         if encoder_hidden_states.dtype == torch.float16:
             encoder_hidden_states = encoder_hidden_states.clip(-65504, 65504)
-        if hidden_states.dtype == torch.float16:
-            hidden_states = hidden_states.clip(-65504, 65504)
+
```
BBuf, line 1147: The qwen_image.py change is getting fairly large, and a lot of the new code is repeating the same graph-cache pattern. It may be worth extracting a small helper for “lookup/create graphed callable and run it”, so the main forward path stays focused on the actual model logic.

python/sglang/multimodal_gen/runtime/models/dits/qwen_image.py
```diff
@@ -1134,8 +1144,211 @@ def forward(
...
+        temb_txt_silu: torch.Tensor,
+        image_rotary_emb: Optional[Tuple[torch.Tensor, torch.Tensor]] = None,
+        joint_attention_kwargs: Optional[Dict[str, Any]] = None,
+        modulate_index: Optional[List[int]] = None,
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
+        if not self._enable_cuda_graph_capture:
+            img_modulated, img_gate1, img_mod2 = self._img_pre_attention_forward(
+                hidden_states, temb_img_silu, modulate_index
+            )
+        elif modulate_index is not None:
```
BBuf, line 1165: The img_pre/img_post paths currently split again on modulate_index is None vs not None, which adds quite a bit of duplication. I think this could be simplified by normalizing that argument in a wrapper and keeping the graph dispatch logic in one place.

python/sglang/multimodal_gen/runtime/models/dits/qwen_image.py
```diff
@@ -1134,8 +1144,211 @@ def forward(
...
+                ("img_post_attention_forward", img_attn_output.shape)
+            ]
+            hidden_states = module(img_attn_output, img_mod2, img_gate1, hidden_states)
+
+        if not self._enable_cuda_graph_capture:
+            encoder_hidden_states = self._txt_post_attention_forward(
+                txt_attn_output, txt_mod2, txt_gate1, encoder_hidden_states
+            )
+        else:
+            padded_txt_attn_output = pad_tensor_into_power_of_2(txt_attn_output, dim=1)
```
BBuf, line 1322: The text-side power-of-2 padding flow also seems like a good candidate for a small utility/helper. Right now `seq_len -> padded_seq_len -> pad -> run graph -> slice` back is embedded directly in `forward()`, which makes the control flow harder to read than it needs to be.

hnyls2002:
Thanks @zyksir! Diffusion CUDA graph support landed via #27436 (breakable CUDA graph for DiTs), including the Qwen-Image-specific padder that buckets/pads only the text sequence - `python/sglang/multimodal_gen/runtime/breakable_cuda_graph/model_padders/qwen_image.py:33,41` (registered at `:102`), with the DiT made BCG-aware at `runtime/models/dits/qwen_image.py:1216,1562`. The flag is `--enable-breakable-cuda-graph` (`runtime/server_args/server_args.py:332`). Closing as superseded - please reopen if I've missed something.

PR #19504, opened 2026-02-27
[PD] Support PD with context parallel after refactor

ShangmingCai:
## Motivation
Support CP in the PD module after #17213 

## Checklist

- [ ] Format your code according to the [Format code with pre-commit](https://docs.sglang.io/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [ ] Add unit tests according to the [Run and add unit tests](https://docs.sglang.io/developer_guide/contribution_guide.html#run-and-add-unit-tests).
- [ ] Update documentation according to [Write documentations](https://docs.sglang.io/developer_guide/contribution_guide.html#write-documentations).
- [ ] Provide accuracy and speed benchmark results according to [Test the accuracy](https://docs.sglang.io/developer_guide/contribution_guide.html#test-the-accuracy) and [Benchmark the speed](https://docs.sglang.io/developer_guide/contribution_guide.html#benchmark-the-speed).
- [ ] Follow the SGLang code style [guidance](https://docs.sglang.io/developer_guide/contribution_guide.html#code-style-guidance).

## Review Process

1. Ping Merge Oncalls to start the PR flow. See the [PR Merge Process](https://github.com/sgl-project/sglang/blob/main/.github/MAINTAINER.md#pull-request-merge-process).
2. Get approvals from [CODEOWNERS](https://github.com/sgl-project/sglang/blob/main/.github/CODEOWNERS) and other reviewers.
3. Trigger CI tests with [comments](https://docs.sglang.io/developer_guide/contribution_guide.html#how-to-trigger-ci-tests) or contact authorized users to do so.
   - `/tag-run-ci-label`, `/rerun-failed-ci`, `/tag-and-rerun-ci`
4. After green CI and required approvals, ask Merge Oncalls to merge.

PR #19548, opened 2026-02-28
fix: support PP2+CP8+TP8 (PP with context parallelism)

whybeyoung:
AS https://github.com/sgl-project/sglang/pull/19504#issuecomment-3976406376 mentioned:
i fixed it on H20 * 8 * 2

- scheduler_pp_mixin: only TP0+CP0 rank does pyobj send/recv to next PP stage; after TP broadcast, add CP broadcast so all CP ranks get data.
- server_args: set attn_cp_size=tp_size in NSA prefill CP path; allow PP with CP when enable_nsa_prefill_context_parallel is set.

Ref: 98e9ecbdf (fix pp new cp)

CC @ShangmingCai @Fridge003  @xu-yfei


python/sglang/srt/server_args.py
```diff
@@ -2116,7 +2117,10 @@ def _handle_context_parallelism(self):
             assert (
                 self.tp_size % (self.dp_size * self.attn_cp_size) == 0
             ), "tp_size must be divisible by dp_size * attn_cp_size"
-            assert self.pp_size == 1, "PP is not supported with context parallelism"
+            if not self.enable_nsa_prefill_context_parallel:
+                assert (
+                    self.pp_size == 1
+                ), "PP is not supported with context parallelism"
```
ShangmingCai: Should we just remove this?

ShangmingCai (approved):
LGTM for the broadcast part

ShangmingCai (commented):
need to fix lint

python/sglang/srt/server_args.py
```diff
@@ -2142,7 +2117,10 @@ def _handle_context_parallelism(self):
             assert (
                 self.tp_size % (self.dp_size * self.attn_cp_size) == 0
             ), "tp_size must be divisible by dp_size * attn_cp_size"
-            assert self.pp_size == 1, "PP is not supported with context parallelism"
+            if not self.enable_nsa_prefill_context_parallel:
+                assert (
```
whybeyoung: The assert blocks the case: context parallelism on (attn_cp_size > 1) with pipeline parallelism on (pp_size > 1) while NSA prefill context parallel is off (enable_nsa_prefill_context_parallel=False). That combination is not implemented/tested; removing the check would allow an invalid config to start.

python/sglang/srt/server_args.py
whybeyoung: The assert blocks the case: context parallelism on (attn_cp_size > 1) with pipeline parallelism on (pp_size > 1) while NSA prefill context parallel is off (enable_nsa_prefill_context_parallel=False). That combination is not implemented/tested; removing the check would allow an invalid config to start.

ShangmingCai: OK

python/sglang/srt/server_args.py
```diff
@@ -2142,7 +2117,8 @@ def _handle_context_parallelism(self):
             assert (
                 self.tp_size % (self.dp_size * self.attn_cp_size) == 0
```
yiakwy-xpu-ml-framework-team, line 2442: can we clarify the attn_cp_size ? Currently cp is used as attn_dp_size = tp / dp. It can be confusing .

python/sglang/srt/server_args.py
```diff
@@ -2142,7 +2117,8 @@ def _handle_context_parallelism(self):
...
             ), "tp_size must be divisible by dp_size * attn_cp_size"
-            assert self.pp_size == 1, "PP is not supported with context parallelism"
+            if not self.enable_nsa_prefill_context_parallel:
+                assert self.pp_size == 1, "PP is not supported with context parallelism"
```
yiakwy-xpu-ml-framework-team: This should be reomved.

with TP=2, we can support CP for H800/H20x8.

For H200, there is no such constraints.

yiakwy-xpu-ml-framework-team (approved):

PR #19568, opened 2026-02-28
Cleanup disagg decode prebuilt flow and add cross-stream sync in merge_batch

Baidu-AIAK:
## Summary

- Make the optimized prebuilt batch flow the default in disagg decode, removing the `SGLANG_ENABLE_OPT_PREBUILT_BATCH` env flag and the old non-optimized path.
- The new default flow calls `merge_batch` before `filter_batch`, which introduces a cross-stream data race (`running_batch.seq_lens` written on `forward_stream`, read on `default_stream`). Add unconditional `maybe_wait_verify_done()` in `merge_batch` to synchronize. For the regular scheduler path this is a no-op since `filter_batch` already syncs.
- Shallow-copy `reqs` list in `ScheduleBatch.copy()` to prevent in-place mutations from corrupting the snapshot used by `process_batch_result` in overlap mode.
- Add ITL constraint to disagg decode `bench_serving` test.

## Test plan

- [ ] Existing spec v2 tests (`test_eagle_infer_beta`, `test_eagle_infer_b`)
- [ ] Existing disagg DP attention tests (`test_disaggregation_dp_attention`)


lixiaolx:
@hnyls2002 Hello，In the PD split mode, merge_batch needs to use maybe_wait_verify_done to ensure data correctness. Below is the script and instructions for reproducing IMA using single-pressure decode.
badcase：
When multiple consecutive batches in the scheduler are prebuilt batches, the preceding filter batches become ineffective, causing the seq_lens value in the merge batch to be invalid, thus triggering IMA.

start_server

```
#!/usr/bin/env bash
set -e
export TP_SIZE=8
export DP_SIZE=8
export BATCH_SIZE=200
export MODEL_PATH=/huggingface.co/deepseek-ai/DeepSeek-V3/DeepSeek-V3.2-1201

export SGLANG_ENABLE_JIT_DEEPGEMM=True
export SGLANG_ENABLE_SPEC_V2=True
export SGLANG_DEEPEP_NUM_MAX_DISPATCH_TOKENS_PER_RANK=512
export NCCL_DEBUG=INFO
export SGLANG_SIMULATE_ACC_LEN=1

    --model-path "${MODEL_PATH}" \
    --disaggregation-ib-device mlx5_bond_0 \
    --prefill-round-robin-balance \
    --tp-size "${DP_SIZE}" \
    --dp-size "${DP_SIZE}" \
    --enable-dp-attention \
    --decode-log-interval 2 \
    --page-size 64 \
    --host 0.0.0.0 \
    --port 8000 \
    --trust-remote-code \
    --moe-dense-tp-size 1 \
    --disable-radix-cache \
    --watchdog-timeout 1000000 \
    --log-level info \
    --deepep-mode low_latency \
    --mem-fraction-static 0.835 \
    --max-total-tokens 600000 \
    --kv-cache-dtype fp8_e4m3 \
    --max-running-requests "$((${BATCH_SIZE} * DP_SIZE))" \
    --cuda-graph-max-bs "${BATCH_SIZE}" \
    --attention-backend nsa \
    --nsa-decode-backend flashmla_kv \
    --moe-a2a-backend deepep \
    --num-reserved-decode-tokens 102 \
    --enable-metrics \
    --enable-cache-report \
    --collect-tokens-histogram \
    --speculative-num-steps 1 \
    --speculative-eagle-topk 1 \
    --speculative-num-draft-tokens 2 \
    --enable-dp-lm-head \
    --disaggregation-decode-enable-fake-auto \
    --load-balance-method round_robin \
    --ep-dispatch-algorithm fake

